PR #99, opened 2019-07-01
Add expanding to report description

kapalex:
Report description cannot be fully viewed (unless entering edit), this keep the text truncated but adds ability to expand.


opencti-front/src/components/TruncatedMarkdown.js
```diff
@@ -0,0 +1,37 @@
...
+        <IconButton onClick={onClick}>
+          {expand ? <ExpandLess /> : <ExpandMore />}
+        </IconButton>
+      </Grid>
+    </Grid>
+  );
+};
+
+TruncatedMarkdown.propTypes = {
+  source: PropTypes.string,
```
richard-julien: isRequired too.

richard-julien (changes requested):
Thanks for your first contribution, it really makes sens. 
Just some remarks to take into account before merging.
Thanks.

kapalex:
Fixed all the issues pointed out.
Thank you.

PR #102, opened 2019-07-01
Typos

BFergerson:
Big Grakn fan here. Interesting project.

Noticed #96 after I already made changes. Includes an additional typo fix.


SamuelHassine:
Thanks for this PR! Even if it is a small one it is always useful!

PR #205, opened 2019-09-02
Fix missing / for deepscan.io badge

2xyo:


richard-julien:
Thanks for this PR. Appreciate you takes time for this.

Issue #212, opened 2019-09-07
Remove "waiting behavior" from entrypoint, let docker restart the containers


richard-julien:
## Problem to Solve

The current docker images implements a waiting for .... in order to wait for dependencies  like Grakn / Elasticsearch. Its lead to code complexity and extra dependencies

## Current Workaround

No

## Proposed Solution

Let docker do its jobs by restarting container if they fail to start. To do that:
- Group the schema, migration in the same process than the API.
- Exit 1 the process if any error occurred during the bootstrap
- Add a check in worker to test the API and fail to start if an error occurred (connect to rabbit or API)

richard-julien:
Take a look in #191 implementation

Issue #191, opened 2019-08-22
Reduce opencti/platform docker image size


2xyo:
## Problem to Solve

`opencti/platform` docker image is big:

``` 
$ docker images opencti/platform:1.1.1
REPOSITORY          TAG                 IMAGE ID            CREATED             SIZE
opencti/platform    1.1.1               6ba2d445f688        2 weeks ago         1.58GB

```
```
$ docker history opencti/platform:1.1.1
IMAGE               CREATED             CREATED BY                                      SIZE                COMMENT
6ba2d445f688        2 weeks ago         /bin/sh -c #(nop)  ENTRYPOINT ["/entrypoint.…   0B                  
<missing>           2 weeks ago         /bin/sh -c #(nop)  EXPOSE 4000/tcp              0B                  
<missing>           2 weeks ago         /bin/sh -c chmod +x /entrypoint.sh              1.09kB              
<missing>           2 weeks ago         /bin/sh -c #(nop) COPY file:28f787f6dc9089f6…   1.09kB              
<missing>           2 weeks ago         /bin/sh -c rm -Rf /opt/opencti-build            0B                  
<missing>           2 weeks ago         /bin/sh -c mv /opt/opencti-build/opencti-gra…   189MB               
<missing>           2 weeks ago         /bin/sh -c cd /opt/opencti-build/opencti-gra…   250MB               
<missing>           2 weeks ago         /bin/sh -c cd /opt/opencti-build/opencti-fro…   895MB               
<missing>           2 weeks ago         /bin/sh -c #(nop) COPY dir:257a895a13d3af22a…   689kB               
<missing>           2 weeks ago         /bin/sh -c #(nop) COPY dir:6a7dfa868771c828e…   3.14MB              
<missing>           2 weeks ago         /bin/sh -c DEBIAN_FRONTEND=noninteractive ap…   6.45MB              
<missing>           2 weeks ago         /bin/sh -c echo "deb https://dl.yarnpkg.com/…   47B                 
<missing>           2 weeks ago         /bin/sh -c curl -sS https://dl.yarnpkg.com/d…   14.8kB              
<missing>           2 weeks ago         /bin/sh -c DEBIAN_FRONTEND=noninteractive ap…   82.9MB              
<missing>           2 weeks ago         /bin/sh -c curl -sL https://deb.nodesource.c…   38MB                
[...]

PR #213, opened 2019-09-08
Reduce docker image size #191 + let docker restart the containers #212

richard-julien:


richard-julien:
See also https://github.com/OpenCTI-Platform/opencti/issues/212

opencti-worker/Dockerfile
```diff
@@ -1,23 +1,14 @@
...
-ENV LC_ALL en_US.UTF-8 
+FROM python:3.7.4-alpine3.10
 
 # Copy the worker
 COPY src /opt/opencti-worker
 
 # Install Python modules
-RUN cd /opt/opencti-worker && pip3 install -r requirements.txt
+RUN BUILD_WORKER=noninteractive \
+    apk add build-base && \
```
2xyo: `--no-cache` -> https://github.com/hadolint/hadolint/wiki/DL3019
version pinning -> https://github.com/hadolint/hadolint/wiki/DL3018
```suggestion
    apk add --no-cache build-base=0.5-r1 && \
```

richard-julien: Thanks for your comment. I will take some times to dig into hadolint and fix that.

SamuelHassine (approved):

Issue #236, opened 2019-09-24
Add version/build number and minimal system info in dashboard


8hur:
## Problem to Solve

When investigating issues or reporting bugs, finding out which version/build is deployed might be needed to avoid duplicates. Also, environment specific issues might be easer to diagnose if the main environment parameters are exposed.

## Current Workaround

Less detailed bug reports

## Proposed Solution

Add a "about" tab with the build number and other useful debugging infos to the settings menu.

maertv:
It seems that Grakn server does not expose its version.

PR #259, opened 2019-10-14
[api/frontend] Add an About tab with the version information in the Settings (#236)

maertv:


maertv:
I mentioned in the issue ([link](https://github.com/OpenCTI-Platform/opencti/issues/236#issuecomment-541112409)) that (as far as I know) the Grakn server does not expose its version. Therefore the current implementation has hard-coded Grakn version.

The Grakn Node.js client has an open ([link](https://github.com/graknlabs/client-nodejs/issues/47)) issue that includes the discussion about exposing the server version.

richard-julien:
Thanks again. I rework some part to have an API closer to the expected UI.

maertv:
Yes, it is much better with your modifications.

Issue #266, opened 2019-10-15
Add a tags field on creation forms


SamuelHassine:
## Problem to Solve

Tags can only be added/removed in the "view" of an entity (threat actor, intrusion sets, etc.).

## Current Workaround

None.

## Proposed Solution

Add a tags field in the creation forms of entities like the MarkingDefinition.

## Additional Information

None.

maertv:
It seems that following views are missing tag related functionality:
- Tools - details (overview) and update form missing tags.
- Vulnerabilities - details (overview) and update form missing tags.
- Observables - listing, details (overview) and update form missing tags.
- Reports - listing, details (overview) and update form missing tags.

Maybe it is clearer to create separate issues for each case if the view needs to handle tags.

Fred-certeu:
Hi, I confirm that tagging is crucial for reports
Thanks

SamuelHassine:
@Fred-certeu I confirm that tagging will be added for Tools/Vulnerabilities/Observables/Reports (#345)

PR #318, opened 2019-11-04
[api/frontend] Add a tags field on creation forms (#266)

maertv:


maertv:
See the comment in the issue #266 (not all views display tags).

Issue #410, opened 2020-01-15
Exception when importing STIX 2 Indicators due to invalid default type in valid_from


JasonKeirstead:
Please replace every line in curly brackets { like this } with an appropriate answer, and remove this line.

## Description

If you try to import a STIX 2 Indicator object with no valid_from, the code sets it to a millisecond long value when the database expects it to be a Date type, which throws an error.

## Environment

OpenCTI current source code

## Reproducible Steps

Steps to create the smallest reproducible scenario:

Step 1: Create a STIX 2 Indicator, without valid_from set
Step 2: Ensure it is valid STIX
Step 3: Attempt import to OpenCTI

## Expected Output

Expect it to import

## Actual Output
`
grakn.core.kb.concept.api.GraknConceptException: The value [1579035233199] of type [Long] must be of datatype [java.time.LocalDateTime]
    at grakn.core.kb.concept.api.GraknConceptException.create(GraknConceptException.java:55)
    at grakn.core.kb.concept.api.GraknConceptException.invalidAttributeValue(GraknConceptException.java:116)
    at grakn.core.concept.manager.ConceptManagerImpl.createAttribute(ConceptManagerImpl.java:271)
    at grakn.core.concept.manager.ConceptManagerImpl.createAttribute(ConceptManagerImpl.java:92)
    at grakn.core.concept.impl.AttributeTypeImpl.putAttribute(AttributeTypeImpl.java:120)
 `
## Additional information

Problematic generated insert statement

[...]

PR #411, opened 2020-01-15
Default valid_from in indicators creation (#410)

JasonKeirstead:
This is a fix for issue #410


opencti-platform/opencti-graphql/src/domain/indicator.js
```diff
@@ -117,7 +117,7 @@ export const addIndicator = async (user, indicator, createObservables = true) =>
   const indicatorToCreate = pipe(
     assoc('main_observable_type', indicator.main_observable_type.toLowerCase()),
     assoc('score', indicator.score ? indicator.score : 50),
-    assoc('valid_from', indicator.valid_from ? indicator.valid_from : Date.now()),
+    assoc('valid_from', indicator.valid_from ? indicator.valid_from : new Date()),
```
richard-julien, line 120: Function now() from database/grakn should be use instead.

JasonKeirstead: Not sure how to do that at this point in the code, previously it was using all the values from Node.

richard-julien: You can check addIncident in incident.js
```
import { now } rom '../database/grakn';
const currentDate = now();
```

richard-julien: Do you have some time to finish the PR or other question on my remark ? 
If you don't have the time I can modify it a little and merge it. Thanks

SamuelHassine: Because of the method `prepareAttribute` in grakn.js, I was indeed intended to create a new pure Date object in JS. Your PR is working and that's what we need. Thanks a lot!

PR #552, opened 2020-03-06
explain .env file

darcosion:
explain .env file for simplifications


SamuelHassine:
Thanks a lot for this PR!

PR #591, opened 2020-04-01
[frontend] Translate filtering in lists

SamuelHassine:


opencti-platform/opencti-front/src/components/list_cards/ListCards.js
```diff
@@ -143,7 +146,7 @@ class ListCards extends Component {
                     <Chip
                       key={filter[0]}
                       classes={{ root: classes.filter }}
-                      label={`${filter[0]}: ${
+                      label={`${t(`filter_${filter[0]}`)}: ${
```
richard-julien, line 149: test

richard-julien: test response

PR #681, opened 2020-05-04
[api] LDAP mapping error handling and logging

richard-julien:


kirintwn:
Hi, I am afraid that this line would cause every new user's email to be a mock email instead of their's even if they are not using LDAP authentication strategy? I think we shall preserve user's original email if they do exist.

https://github.com/OpenCTI-Platform/opencti/blob/d28639e3e47c8225bc651a8d134a9aae0570d075/opencti-platform/opencti-graphql/src/domain/user.js#L243

richard-julien:
Hi @kirintwn , thanks for you comment. The line is inside the function addPerson and not addUser. A person is an entity of the platform that cannot be used to log in, but by design its an inheritance of the user entity that require an email key. When a user log in through ldap he uses the method loginFromProvider. If you take a look, you will see that this method call addUser if the user is not found.

Issue #470, opened 2020-02-08
Prevent users from accidentally creating duplicate objects (e.g. threat actors).


graememeyer:
Please replace every line in curly brackets { like this } with appropriate answers, and remove this line.

## Problem to Solve

{ Prevent users from accidentally creating duplicate objects (e.g. threat actors). }

## Current Workaround

{ Basically have to manually make sure the threat actor / object doesn't already exist. }

## Proposed Solution

{ I would suggest a warning pop-up if you type something that already exists into the name field for creating threat actors, attack patterns etc. 

There's probably some more advanced stuff you could do around merging/modifying existing entries if a user attempts this, but a basic warning (or just straight-up not allowing the behaviour) would probably be a great feature. }

## Additional Information

{ Tested in the demo environment (2.1.4) with FF and Chrome on Win10 }

PR #729, opened 2020-05-27
[frontend] Detect potential duplicate entities at creation (#470)

SamuelHassine:


anhthiencao:
Can you give me the document or explain the reason that why detect duplicate entities including other entities? Thank you!

**Example in [ CampaignCreation.js](https://github.com/OpenCTI-Platform/opencti/blob/35b94879369a474fbefae6766d7f3dca6ce335b3/opencti-platform/opencti-front/src/private/components/threats/campaigns/CampaignCreation.js):**

`
detectDuplicate={[
  'Threat-Actor',
  'Intrusion-Set',
  'Campaign',
  'Malware'
 ]}
`

>

PR #958, opened 2020-12-21
An update to support more special characters in the chosen rabbitmq a…

### LINKED ISSUE #957: Support for more Special Characters in RabbitMQ Password
Please replace every line in curly brackets { like this } with appropriate answers, and remove this line.

## Problem to Solve

Havving an '@' in the rabbitmq user password causes a problem as it is misinterpreted as pre-host special character in "amqp://${user}:${pass}@${host}:${port}"

## Current Workaround

Not to use @ in rabbitmq user password

## Proposed Solution

An update to support more special characters in the chosen rabbitmq admin password, by passing the credentials as a connect socket option instead of passing them part of the connect URL

## Additional Information

Will submit a PR shortly

ayman-m:
An update that closes #957

### DIFF
--- a/opencti-platform/opencti-graphql/src/database/rabbitmq.js
+++ b/opencti-platform/opencti-graphql/src/database/rabbitmq.js
@@ -2,7 +2,7 @@ import amqp from 'amqplib';
 import axios from 'axios';
 import * as R from 'ramda';
 import conf from '../config/conf';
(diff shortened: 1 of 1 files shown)


SamuelHassine:
@ayman-m Thanks a lot for this PR! nice work!

PR #1017, opened 2021-01-22
[frontend] Show description on Vulnerablility overview

summershrimp:
It would be better to have description shown on Vulnearbility overview page to let user get more info.


SamuelHassine:
Thanks a lot for this PR!

PR #1172, opened 2021-03-18
[frontend] Allow query vulnerabilities by any CVSS score value

### LINKED ISSUE #1158: Search filter value input issue.
## Description

1. ~~search filters name is not translated by `Localization.js`, just show the translate key name such as `filter_created_start_date`~~
2. search filters dealing with self inputed value has some mistake

## Environment

1. https://demo.opencti.io/
2. also happens on  **opencti-docker 4.3.0** version with `docker compose up -d`.
3. chrome 89
## Reproducible Steps

1. Open any list search filter popup
2. ~~You can see the filter name is not translated.~~
3. Select any input box and input some text
4. cilck any blank area outside the search filter popup (search filter should dismiss)
5. reopen the search filter
6. you can see all the input box rather than date is fill with the same value as you input in step 3
7. change the values, you can see all the values are changed together.

## Expected Output
1. ~~search filter name should be well translated.~~
2. the input box won't accept any keyboard input value.

## Actual Output
1. ~~search filter name just same as translation key~~
2. the input box deal with input incorrect.
 
## Additional information

None

summershrimp:
Add `freeSolo` attribute to Filters' Autocomplete component so that
allowed to input any value to the input other than just select one
in the predefined lists.

Also temporary fix #1158

Signed-off-by: Yibai Zhang <xm1994@gmail.com>

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/common/lists/Filters.js
+++ b/opencti-platform/opencti-front/src/private/components/common/lists/Filters.js
@@ -173,6 +173,7 @@ class Filters extends Component {
         break;
       case 'x_opencti_base_score_gt':
         this.setState({
+          freeSolo: true,
           entities: {
             x_opencti_base_score_gt: union(
               this.state.entities,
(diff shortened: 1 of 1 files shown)


SamuelHassine:
Hello @summershrimp,

Thanks for your PR. We will add `freeSolo` in some filters to help users with more advanced filtering use cases, but for this specific need (CSVSS score), we've decided to just propose all possible levels (from 1 to 9) in the filters.

Kind regards,
Samuel

Issue #1217, opened 2021-03-29
[api/frontend] Note abstract property should not be required


maertv:
## Description

The current implementation requires that the _abstract_ (_attribute_abstract_ in the OpenCTI) property of a Note is filled. The _abstract_ property is optional by the STIX2 specification.

It would be nice to follow the STIX2 specification if there is no good reason to deviate from it. 

## Environment

1. OS (where OpenCTI server runs): Ubuntu 18.04.5 LTS
2. OpenCTI version: _master_
3. OpenCTI client: API/Frontend

## Reproducible Steps

Try creating a Note without an _abstract_ in the frontend.

## Expected Output

It should be possible to create a Note without providing the _abstract_.

## Actual Output

It is not possible to create a Note without the _abstract_.

PR #1218, opened 2021-03-29
[api/frontent] Make Note abstract property optional (#1217)

maertv:


SamuelHassine:
Sorry for double work @maertv, thanks for this PR anyway!

PR #1274, opened 2021-04-19
[frontend] Add a "related reporting" Correlation graph to the Knowledge UI in Analysis Reports

ckane:
Should be considered a Work In Progress right now, but it's about 80% of the way there - so I wanted to gather some feedback on it via this PR (and I'll keep updating my branch).

The basic idea is to add a view to Analysis Reports (Analysis->Reports in the UI), under "Knowledge" that renders a "correlation" graph of any reporting that is related to this report by any of the entities that are accessed via this report's `objects` property. I picked the name "correlation", since that's also the terminology MISP is using to render similar information in its Event view.

This is probably the most frequent visualization I find myself having to make in order to answer many CTI questions that come up within our team.

Since this is a WIP, it is based upon copying over code already used for the existing Knowledge "event graph", and then modifying a few key portions of that in order to diagram the relationships between the reports and the entities, as well as filter down the graph to only diagram entities with 2 or more unique reporting relationships. There's still some more work to do, such as:

- [x] fix rendering of 3d graphs
- <strike>clean up the code by consolidating shared parts of both graphs into common files (in `src/utils`?) to include</strike>
- [x] re-work the right-hand graph bar tools to provide actions that make the most sense in this context
- [x] make entity filters work
- [x] make time window selector work properly
- <strike>show a message when a graph is empty because it has no entities with other related reports</strike>
- [x] fix all other rendering issues that are (likely) due to the modifications I've made with these changes
- [x] fix the "2 args vs. 3 args" issue with `buildCorrelationData()` - not sure if I want to just eliminate the 3rd arg, or actually implement it
- [x] maybe the current "Graph" link should be renamed to "Entity Graph"?


ckane:
Example diagramming some reporting relationships from my MISP sync, which includes ScanBox, Sakula, and Black Vine reports from a few sources.
image

rhaist:
Thanks @ckane - this is great work and should prove really useful to most of our users.

ckane:
Hi @rhaist @richard-julien, I believe that the first run at this is complete and ready for review. I chose to not consolidate it with the code I copied, yet, and also chose not to display a message in the event of an empty graph. I feel both of those don't need to be blocking items on the new feature introduction, and that they can be done as separate items that would be later improvements. Right now it just maintains the same behavior as the other graphs: no message when there's nothing to render.

SamuelHassine:
Hello @ckane,

Thanks a lot for this PR. We will review it as soon as possible but it's very promising!

Kind regards,
Samuel

ckane:
> Hello @ckane,
> 
> Thanks a lot for this PR. We will review it as soon as possible but it's very promising!
> 
> Kind regards,
> Samuel

Thanks! Let me know if you have any recommended changes to it. This was my first attempt at working with both Ramda and GraphQL, so let me know if I should approach any problems differently than I have.

SamuelHassine (approved):

Issue #1521, opened 2021-08-16
OpenIDConnect Strategy doesn't support roles from claims


DeanF:
## Description

Today, role mappings are resolved through the decoded access_token. However, with many providers, this neither includes the mapped roles or groups. Instead, they are included in the `id_token` field. Or better yet, can be obtained by calling `tokenset.claims()`.

```
          const computeRolesMapping = () => {
            const rolesPath = mappedConfig.roles_management?.roles_path || [];
            const rolesMapping = mappedConfig.roles_management?.roles_mapping || [];
            const decodedUser = jwtDecode(tokenset.access_token);
            const availableRoles = R.flatten(rolesPath.map((path) => R.path(path.split('.'), decodedUser) || []));
            const rolesMapper = genConfigMapper(rolesMapping);
            return availableRoles.map((a) => rolesMapper[a]).filter((r) => isNotEmptyField(r));
          };
```

## Environment

1. OS (where OpenCTI server runs): Ubuntu 20.04
2. OpenCTI version: 4.5.5
3. OpenCTI client: all
4. Other environment details: OpenIDConnect

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Use AzureAD to map app roles to users.
2. No possible configuration of OpenCTI can map the roles.

## Expected Output

Roles to be fetched from the default "roles" claim.

## Actual Output

Login failed due to no role mapping.

PR #1522, opened 2021-08-16
[api] Use claims to resolve roles (#1521)

DeanF:
Use claims to resolve role/group instead of access_token.
Some services do not return all identity data in the access_token field, but rather in id_token/claims.
Remove jwt-decode dependency.

### DIFF
--- a/opencti-platform/opencti-graphql/package.json
+++ b/opencti-platform/opencti-graphql/package.json
@@ -68,7 +68,6 @@
     "iterall": "1.3.0",
     "jsondiffpatch": "0.4.1",
     "jsonwebtoken": "8.5.1",
-    "jwt-decode": "3.1.2",
     "lru-cache": "6.0.0",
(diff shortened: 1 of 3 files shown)


richard-julien:
Hi @DeanF ,

Thanks for your contribution!
Due to some timing problematic I directly push a different approach for this problem (https://github.com/OpenCTI-Platform/opencti/commit/6a61c1a5d768f5b8675197bf4537759ba910235f)
With this approach we can maintains the compatibility with system already using the access token to get the roles/groups.
For your use case you can setup in the configuration if you want to use the id_token instead (with token_reference)
```
        "roles_management": {
          "token_reference": "id_token",
          "roles_scope": "xxx",
          "roles_path": ["xxxx"],
          "roles_mapping": ["xxxx"]
        }
```
Same for the group.
Thanks again for the pull request even if this time i need to reject it.

Issue #1595, opened 2021-09-27
 Expose express server metrics for prometheus


debelyoo:
## Use case

Metrics are required to monitor properly the API. By metrics, I mean numbers about the following things:
- overall query rate
- failed query rate
- query rate by route
- response time

The metrics should be then used to feed monitoring dashboards (i.e. Grafana dashboards).

## Current Workaround

n/a

## Proposed Solution

I propose to expose metrics for prometheus and use https://www.npmjs.com/package/express-prometheus-middleware for that.

## Additional Information

n/a

## If the feature request is approved, would you be willing to submit a PR?

Yes

PR #1598, opened 2021-09-27
[api] Expose express server metrics for prometheus

debelyoo:
### Proposed changes

* Expose metrics for prometheus
* Use express-prometheus-middleware

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/1595

### Checklist

### Further comments

Exposing the metrics can be enabled by adding the following env variables to the container:

- `APP__EXPOSE_METRICS=true`


opencti-platform/opencti-graphql/config/default.json
```diff
@@ -2,6 +2,7 @@
   "app": {
     "port": 4000,
     "enabled": true,
+    "expose_metrics": false,
```
richard-julien: I think metrics is not only prometheus. We need to prepare the possibility to have more metrics providers exposed (lie https://github.com/open-telemetry/opentelemetry-js)

I think about a little more conplex configuration like

```
"telemetry": {
    "prometheus": {
        "enabled": false,
        "metrics_path": "/metrics"
    }
}
```

telemetry should be disable by default to prevent performance impact

opencti-platform/opencti-graphql/src/http/httpPlatform.js
```diff
@@ -83,6 +85,19 @@ const createApp = async (apolloServer) => {
   }
   app.use(securityMiddleware);
   app.use(compression({}));
+  const exposeMetrics = booleanConf('app:expose_metrics', false);
```
richard-julien: Change it to use the right configuration

opencti-platform/opencti-graphql/src/http/httpPlatform.js
```diff
@@ -83,6 +85,19 @@ const createApp = async (apolloServer) => {
...
+  if (exposeMetrics) {
+    logApp.info('Adding prometheus middleware (for metrics)...');
+    app.use(
+      promMid({
+        metricsPath: '/metrics',
```
richard-julien: Change it to use the right configuration.
Need also to be tested if its working well with basePath configured

richard-julien (changes requested):

opencti-platform/opencti-graphql/config/default.json
debelyoo: Yes, makes sense 👍

opencti-platform/opencti-graphql/src/http/httpPlatform.js
debelyoo: The `base_path` config property is not used by the prometheus middleware. The full path of the metrics endpoint has to be defined in the `metricsPath` property.

richard-julien: I think you have to concatenate yourself the basePath like we do for other middleware.  You can take a look at apolloServer.applyMiddleware

debelyoo: Ok. I updated the code.

richard-julien: Have you updated the PR?

debelyoo: Oups I forgot. Sorry for that.

Issue #1653, opened 2021-10-21
An irrelevant response with unauthenticated GraphQL requests


fdevillard:
## Description

An irrelevant response might be returned with unauthenticated requests.

## Environment

1. OS (where OpenCTI server runs): MacOS / Docker
2. OpenCTI version: 5.0.3
3. OpenCTI client: N/A
4. Other environment details: N/A

## Reproducible Steps

Run:

```
POST /graphql
{
  me {
    user_email
    id
    standard_id
    roles {
      name
    }
  	capabilities {
      id
      name
    }
  }
}
```

## Expected Output

```
{
  "errors": [
    {
      "message": "You must be logged in to do this.",
      "name": "AuthRequired",
      "time_thrown": "2021-10-21T12:56:55.561Z",
      "data": {
        "http_status": 401,
        "category": "technical"
      }
    }
  ],
  "data": {
    "me": null
  },
  "extensions": {
    "tracing": {
      "version": 1,
      "startTime": "2021-10-21T12:56:55.559Z",
      "endTime": "2021-10-21T12:56:55.579Z",
      "duration": 19969262,
      "execution": {
        "resolvers": [
          {
            "path": [
              "me"
            ],
            "parentType": "Query",
            "fieldName": "me",
            "returnType": "User",
            "startOffset": 1541350,
            "duration": 366186
          }
        ]
      }
    }
  }
}

```
(at least with in development mode)

## Actual Output

```
{
  "errors": [
    {
      "message": "The \"string\" argument must be of type string or an instance of Buffer or ArrayBuffer. Received undefined",
      "extensions": {
        "code": "INTERNAL_SERVER_ERROR",
        "exception": {
          "stacktrace": [...]
        }
      }
    }
  ]
}

```
 
## Additional information

No authentication scheme is used for the query.

An MR is in progress.

## Screenshots (optional)
N/A

richard-julien:
Im not able to reproduce it with the playground. What do you use to execute your graphql query? Thanks

fdevillard:
I'm using postman to call the API. 

I've reproduced it on the demo (is it what you meant by _playground_ ?). The output is smaller (I suppose due to the production mode):
```
{
    "errors": [
        {
            "message": "The \"string\" argument must be of type string or an instance of Buffer or ArrayBuffer. Received undefined",
            "extensions": {
                "code": "INTERNAL_SERVER_ERROR"
            }
        }
    ]
}
```

[...]

PR #1654, opened 2021-10-21
[fix] Empty variable exception

fdevillard:
This MR is about fixing an exception raised in the `loggerPlugin`. More
specifically, when `context.request.variables` is undefined, then computing its
size raises an exception. This behavior used to happen with
unauthenticated GraphQL request.

### Proposed changes
If the `context.request.variables` is undefined, then the variables' size is assumed to be `0`.

### Related issues
https://github.com/OpenCTI-Platform/opencti/issues/1653 

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/graphql/loggerPlugin.js
+++ b/opencti-platform/opencti-graphql/src/graphql/loggerPlugin.js
@@ -44,15 +44,16 @@ export default {
         if (!isCallError && !perfLog) {
           return;
         }
-        const size = Buffer.byteLength(JSON.stringify(context.request.variables));
+        const contextVariables = context.request.variables || {}
+        const size = Buffer.byteLength(JSON.stringify(contextVariables));
         const isWrite = context.operation && context.operation.operation === 'mutation';
         const contextUser = context.context.user;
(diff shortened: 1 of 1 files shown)


opencti-platform/opencti-graphql/src/graphql/loggerPlugin.js
```diff
@@ -44,15 +44,16 @@ export default {
         if (!isCallError && !perfLog) {
           return;
         }
-        const size = Buffer.byteLength(JSON.stringify(context.request.variables));
+        const contextVariables = context.request.variables
```
richard-julien: Maybe its simpler to just extract the variables and put a default value like this

`const contextVariables = context.request.variables || {}`

Then no need to changes the rest.

fdevillard: Yep, true. It'd be more concise.

In this case, the size of an empty variable set will be `2` rather than `0`. It seems a bit counter-intuitive. Would that be ok for you?

richard-julien: I thinks its fine yes

PR #1694, opened 2021-10-27
[api] add missing relationship `resolves-to`

axelfahy:
This PR adds a missing relationship: `resolves-to`

Currently, it is not possible to add a `resolves-to` relationship. A current alternative is to use the `resolves_to_refs` field. However, this is similar to the `belongs-to` relationship that do exist (which has a `belongs_to_refs` field as well).

### Proposed changes

* Add `resolves-to` relationship

### Related issues

N/A

### Checklist


SamuelHassine:
For the moment, only the STIX Cyber Observable relationships (resolves-to_refs) will be implemented.

axelfahy:
@SamuelHassine Then, I don't really understand why `belongs-to` is implemented, and is there a specific reason for using `belongs_to_refs` or `resolves_to_refs` instead of `belongs_to` / `resolves_to` ?

I thought that `resolves-to` was a STIX Cyber Observable relationships, where as `resolves_to_refs` was only an embedded relationship (only a property, not a relationship). 

Creating the relationship manually would allow specifying more details about the relation, for example the author.

PR #1798, opened 2021-12-28
[frontend] Fix Analysis Correlation Graph edge creation

ckane:
### Proposed changes
The current formula for creating edges in the "Correlation" view of the
Analysis report graphs creates an edge between every report node and
every non-report entity node on the graph, regardless of the database
contents. This change makes it so that edges will only be drawn between
entities and the reports they actually show up in, in the graph. The
current behavior is a bug.

### Checklist


ckane:
Correlation chart after fixes, demonstrating how only some entities are shared between the current report and another report in the database:
image

SamuelHassine (approved):

PR #2168, opened 2022-06-15
Support for IAM role usage when interacting with S3

### LINKED ISSUE #2024: Allow EC2 metadata credentials.
## Use case

We run in AWS, and best practices denote that we should not create users for running applications, but should instead create roles that the EC2 instance can assume.  This happens via the api using a fallback method of credentials.  Starting with environment vars -> ~/.aws/credentials file -> EC2 metadata endpoint.  The current minio client does not allow for this and instead insists on environment vars for credentials.

## Current Workaround

Create a user and add their creds to the deployment.

## Proposed Solution

Use the official AWS SDK for S3.  This should work with minio as described [here](https://docs.min.io/docs/how-to-use-aws-sdk-for-javascript-with-minio-server.html)

Javascript isn't my thing, but I can provide testing.  I'm also willing to give the JS a shot, just not certain on how to prioritise this with other work.

### LINKED ISSUE #2154: Allow MinIO Credentials to be fetched from AWS ECS attached IAM Execution Role
## Use case

Note this is similar to GitHub Issue: #2024 but for AWS ECS (Docker)

This is a request to allow OpenCTI Platform to fetch AWS Credentials for MinIO/S3 authentication from the metadata of the ECS Task.

To fetch this information the following command is performed:
`curl 169.254.170.2$AWS_CONTAINER_CREDENTIALS_RELATIVE_URI`

This will return the current role credentials that can be used.

Note, as IAM Roles expire after an hour typically, these credentials cannot be cached; its important that either the credentials are refetched for each MinIO call or a time based system is put in place to check if it has been 50 minutes, if so refetch credentials.

## Current Workaround

Make use of an IAM User Credentials which is not secure and is not recommended by AWS for service AWS Access OR make use of a ECS instance that is running MinIO.

## Proposed Solution

If the MinIO Access Key and Secret Key is not set then perform a check against the local metadata.

Happy to assist.

jake-walker:
### Proposed changes

* Supporting IAM role access to interact with S3 instead of requiring IAM users when a user wishes to use AWS S3 instead of MinIO server.

### Related issues

* Resolves #2154
* Resolves #2024 (I recognise this issue suggests a native AWS SDK implementation, it is something that I will add in a future pull request)

### Checklist

### Further comments

When using IAM credentials in AWS, the recommended practice is to use an attached role to an EC2 instance or ECS container rather than static credentials which do not expire. To solve this, I have used the AWS SDK to fetch the attached IAM role credentials which are then used in the MinIO client.

I have ensured this won't impact existing instances by adding a configuration option to opt-in to using this feature. For example, there may be an instance where a user is using an attached IAM role for logging but is interacting with MinIO through a set of static credentials. To ensure that IAM role credentials are not used in such a case, a default value is set to false which means the IAM role credentials won't be used by the MinIO client.

As IAM role credentials are temporary, a new MinIO client needs to be created every time.

This config option would require an update to the documentation as follows:

| Manual | Environment Variable | Default | Description |
| --- | --- | --- | --- |
| minio:use_aws_role | MINIO__USE_AWS_ROLE | false | Use attached IAM role for authentication for S3 |

### DIFF
--- a/opencti-platform/opencti-graphql/package.json
+++ b/opencti-platform/opencti-graphql/package.json
@@ -69,7 +69,7 @@
     "iterall": "1.3.0",
     "jsondiffpatch": "0.4.1",
     "jwt-decode": "3.1.2",
-    "lru-cache": "7.10.1",
+    "lru-cache": "^7.10.1",
     "merge-graphql-schemas": "1.7.8",
     "migrate": "1.8.0",
     "mime-types": "2.1.35",
--- a/opencti-platform/opencti-graphql/package.json
+++ b/opencti-platform/opencti-graphql/package.json
@@ -27,6 +27,7 @@
     "node": ">= 16.*"
   },
   "dependencies": {
+    "@aws-sdk/credential-providers": "^3.110.0",
     "@elastic/elasticsearch": "7.17.0",
     "@graphql-tools/schema": "8.3.14",
     "@graphql-tools/utils": "8.6.13",
--- a/opencti-platform/opencti-graphql/package.json
+++ b/opencti-platform/opencti-graphql/package.json
@@ -27,7 +27,7 @@
     "node": ">= 16.*"
   },
   "dependencies": {
-    "@aws-sdk/credential-providers": "^3.110.0",
+    "@aws-sdk/credential-providers": "3.110.0",
     "@elastic/elasticsearch": "7.17.0",
     "@graphql-tools/schema": "8.3.14",
     "@graphql-tools/utils": "8.6.13",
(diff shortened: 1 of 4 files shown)


richard-julien:
Not sure about the LRU Cache approach, seems a bit too much.
You should maybe just retains the last time credentials have been fetched. When getting the minio client, if now - time > refresh period, we just have to refetch the creds and get the updated minio client.
wdyt @jake-walker ?

opencti-platform/opencti-graphql/package.json
```diff
@@ -27,6 +27,7 @@
     "node": ">= 16.*"
   },
   "dependencies": {
+    "@aws-sdk/credential-providers": "^3.110.0",
```
richard-julien: Currently we decide to fix all the versions you depends on. Maybe we will change that in the future but please remove the ^ for now

richard-julien (changes requested):
One little comment. Everything all all good thanks

richard-julien (approved):

PR #2260, opened 2022-07-30
Switch object storage to use AWS S3 SDK

jake-walker:
### Proposed changes

The MinIO SDK has been switched out for the official S3 SDK which brings benefits when using EC2 instance profiles (as implemented in #2168) as well as working better with S3 (the correct regional endpoints are used for the region that the bucket is in).

### Related issues

* Although already resolved, this provides a better fix for #2024

### Checklist

### Further comments

As explained briefly above, using the AWS S3 SDK adds a number of benefits while still being completely compatible with MinIO. It should also work better with other 'S3 compatible' storages like GCP Cloud Storage, Azure Blob Storage, etc... as the AWS SDK is the 'standard'.

The original functions in `opencti-platform/opencti-graphql/src/database/minio.js` have been kept but completely rewritten with the new SDK using only async code (whereas before it was a bit of a mixture between async and promises) which makes the code much cleaner.

The fetching of credentials as implemented in #2168 is improved by using the AWS SDK. Now a custom ['credential provider'](https://aws.amazon.com/blogs/developer/credential-providers-credential-management-part-3/) which pulls credentials first from the regular configuration file or environment variables, then falls back to using an attached AWS IAM role if available. The AWS SDK does a better job at automatically checking when credentials expire so the hacky code to check for that (from the previous PR) has now been removed. ~~The configuration option for using the AWS role has been removed too, as it can automatically fall back if no configuration is set.~~

I would have changed the configuration section from 'minio' to 'object-storage' or 's3' but it could involve breaking changes. I think having backwards compatibily (i.e. load first from the 'object-storage' section, then if not use the 'minio' section) over complicates the code too much which is why I have just left it as 'minio'. I think this could be worth discussing because the standard for storage is S3, not MinIO anymore.


ReadyElbow:
One note: the 'use_aws_role' environment variable should be removed from the Notion documentation. 

https://luatix.notion.site/Configuration-a568604c46d84f39a8beae141505572a#:~:text=Accept%20all-,minio%3Ause_aws_role,-Show%20All

jake-walker:
> One note: the `use_aws_role` environment variable should be removed from the Notion documentation.
> 
> https://luatix.notion.site/Configuration-a568604c46d84f39a8beae141505572a#:~:text=Accept%20all-,minio%3Ause_aws_role,-Show%20All

As 'ChangeMe' could be used in some environments as well as the CI pipeline, I have re-added the variable from before which now skips loading the credentials from the configuration file and would use the AWS role instead.

opencti-platform/opencti-graphql/src/database/file-storage.js
```diff
@@ -0,0 +1,268 @@
...
+    }
+  }
+
+  return objects.reduce(async (items, item) => {
+    if (!isFileObjectExcluded(item.Key)) {
+      items.push(await loadFile(user, item.Key));
+    }
+    return items;
+  }, []);
+}
```
richard-julien: In case of multiple files this code will fail because items will be transformed as a promise.

Code proposal to replace.

return Promise.all(objects
        .filter((o) => !isFileObjectExcluded(o.Key))
        .map((item) => loadFile(user, item.Key)));

opencti-platform/opencti-graphql/src/database/file-storage.js
```diff
@@ -0,0 +1,268 @@
...
+
+export async function upload(user, path, fileUpload, metadata = {}) {
+  const { createReadStream, filename, mimetype, encoding = '', version = now() } = await fileUpload;
+  const readStream = createReadStream();
+
+  logAudit.info(user, UPLOAD_ACTION, { path, filename, metadata });
+
+  const key = joinPaths(path, filename);
```
richard-julien: I start some testing and looks like the upload is no longer working with minio
The file is no more created inside a hierarchy of directory
To be replace with

> const key = \`${path}/${filename}`;

richard-julien (changes requested):
Review and test still in progress

opencti-platform/opencti-graphql/tests/02-integration/01-database/object-storage-test.js
```diff
@@ -19,7 +19,7 @@ const exportFileName = '(ExportFileStix)_Malware-Paradise Ransomware_all.json';
 const exportFileId = (malware) => `export/Malware/${malware.id}/${exportFileName}`;
 const importFileId = `import/global/${exportFileName}`;
 
-describe('Minio file listing', () => {
+describe('Object storage file listing', () => {
```
richard-julien: File must be renamed to file-storage-test.js

opencti-platform/opencti-graphql/src/database/file-storage.js
jake-walker: I wasn't able to reproduce the problem on my end, but your suggestion should functionally be the same.

richard-julien (approved):

PR #2317, opened 2022-09-01
[api/frontend] Add organizations restrictions on top of markings to increase data segregation possibilities (#2188)

### LINKED ISSUE #2109: Create/Update notes and opinions specifying author with a different user
## Description

A user with only one role with the following permissions ...

image

Can specify the note author as admin or as organization:

Screenshot from 2022-05-25 07-11-35

The button to update a note that does not belong to the logged user is available and the edit panel shows up however it raises an error when we try to update it.

It is possible too for a user to create a new Opinion and specifying as author, for example, as the Admin and save it. Then, update the opinion to add a relation to another entity like a report.

I guess this is possible because Author is linked to Identifies which is not the same as OpenCTI users, however, should it be possible to users to change author of any entity? Or even if yes, should it be possible to create notes and opinions using another user name?

PS: I understand, it seems easy in first sight to solve the problem but conceptually is difficult due the mix between platform users/roles/permissions and stix author field as an Indentity.

## Environment

OpenCTI version: 5.2.4

### LINKED ISSUE #2188: Add organizations restrictions on top of markings to increase data segregation possibilities
## Use case

Today, when an entity has marking X:A and marking Y:N, it will be only available to the group that has both marking.

Use case: be able to share an entity with 2 different customers/teams that have X:A on one side and Y:N on the other side. 

## Current Workaround

Add groups/markings representing all possible permutations.

## Proposed Solution

Implement the ability to associated elements to specific organizations that will be only accessible for users that are part of these organizations

### LINKED ISSUE #2196: Finer access controls for Reports for feedback   - Separate "Opinions" as a knowledge creation access control under roles.
1) We would like the ability to use the "opinions" to give non-analyst users a way to give feedback. But allowing them to give an opinion on a report also allows them to create knowledge. Separating this access out gives the platform more flexibility in terms of user interaction
image
.

2) We would like to have the Opinion values and range (Agree, Disagree) to be configurable, so that we could define our own user/analyst feedback spectrum.
image (1)

### LINKED ISSUE #2483: Be able to use workflow status in the stream filters
## Use case

Be able to use workflow status in the stream filters.

richard-julien:


ckane:
Curious if, overall, the proposed changes would allow me to mark an IP Address (or any other entity, but just pulling IP Addr as an example) such that it is:
* TLP:AMBER+STRICT within my organization (as that might be what "internal restrictions" I might place on anything shared from an external source at TLP:GREEN with my org)
* TLP:GREEN within "CTI sharing group 1", where someone might have shared it at TLP:GREEN level to me
* TLP:RED within "CTI sharing group 2", where someone else might have independently shared it TLP:RED with my org as audience

richard-julien:
> Curious if, overall, the proposed changes would allow me to mark an IP Address (or any other entity, but just pulling IP Addr as an example) such that it is:
> 
> * TLP:AMBER+STRICT within my organization (as that might be what "internal restrictions" I might place on anything shared from an external source at TLP:GREEN with my org)
> * TLP:GREEN within "CTI sharing group 1", where someone might have shared it at TLP:GREEN level to me
> * TLP:RED within "CTI sharing group 2", where someone else might have independently shared it TLP:RED with my org as audience

To clarify a bit, its ongoing work, so need to re-define a bit the semantic :)
* Group = a set of marking. User can be assigned to multiple groups and so get a combination of granted markings
* Organization = company of a user. A user could be inside multiple organizations.
So the PR is more about adding Organization restrictions in top of Group restrictions.

Organization restrictions is another level on top of markings. For me TLP:AMBER+STRICT is a first STIX approach of this problem because it means AMBER **but for MY ORG only**.
For info we decide to do like markings, no organization = all organizations

The big difference between groups and organizations is that **group/markings are exclusive** and **organizations are inclusive**
For example:
**User definitions**
* USER01 is part of GROUP_GREEN (TLP:GREEN) - inside organizations COMPANY01
* USER02 is part of GROUP_RED (TLP:RED) -  inside organizations COMPANY01
* USER03 is part of GROUP_GREEN (TLP:GREEN) -  inside organizations RESTRICT

**Data**
Some use cases
* IP adress 8.8.8.8 created with TLP:GREEN
-> USER01, USER02 and USER03 and all have access.
* IP adress 8.8.8.8 created with TLP:GREEN + COMPANY01 orga restrictions
-> Only USER01 and USER02 have access.
* IP adress 8.8.8.8 created with TLP:RED + COMPANY01 orga restrictions
-> Only USER02 have access.

So for your "CTI sharing group 1" i think it will be modelize as an Organization (type circle/sharing)  that will have some sub organizations for example COMPANY01 + COMPANY02. So if you add the orga restrictions to "CTI sharing group 1", both organizations will have access to the IP.

Hope this help to understand our vision on this subject. :)

SamuelHassine:
Good pour moi cette PR.

SamuelHassine (approved):

opencti-platform/opencti-front/src/private/components/AuthBoundary.js
```diff
@@ -20,7 +20,10 @@ class AuthBoundaryComponent extends React.Component {
       const retroErrors = this.state.error.data?.res?.errors ?? [];
       const types = map((e) => e.name, [...baseErrors, ...retroErrors]);
       // If access is forbidden, just redirect to home page
-      if (includes('ForbiddenAccess', types)) window.location.href = '/';
+      if (includes('ForbiddenAccess', types)) {
+        return <RootPublic type="LOGIN" />;
```
RomuDeuxfois, line 24: Shouldn't the user be returned to a page call : " need more access to do that " ?

opencti-platform/opencti-front/src/private/components/arsenal/malwares/MalwareEditionOverview.js
```diff
@@ -17,37 +16,9 @@ import ConfidenceField from '../../common/form/ConfidenceField';
...
 import CommitMessage from '../../common/form/CommitMessage';
-import {
-  convertCreatedBy,
-  convertMarkings,
-  convertStatus,
-} from '../../../../utils/Edition';
+import { convertCreatedBy, convertMarkings, convertStatus } from '../../../../utils/Edition';
 import StatusField from '../../common/form/StatusField';
 
-const styles = (theme) => ({
```
RomuDeuxfois, line 28: We don't have style for this component anymore ?

opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectSharing.tsx
```diff
@@ -0,0 +1,305 @@
...
+import { commitMutation } from '../../../../relay/environment';
+import { useFormatter } from '../../../../components/i18n';
+import {
+  granted,
+  KNOWLEDGE_KNUPDATE_KNORGARESTRICT,
+  UserContext,
+} from '../../../../utils/Security';
+import { truncate } from '../../../../utils/String';
+
+// region types
```
RomuDeuxfois, line 28: I don't understand this comment.

opencti-platform/opencti-graphql/src/types/stix-common.d.ts
```diff
@@ -62,6 +63,10 @@ interface StixObject {
   };
 }
 
+interface StixInternalObject extends StixObject {
+  sequence: string;
```
RomuDeuxfois, line 67: What is a sequence ?

PR #2375, opened 2022-09-21
[rules] Add report objects related rules (#2341)

### LINKED ISSUE #2341: [rules] Add report objects related rules
## Use case

Data in a report need more inferences to enrich the current report.

## Proposed Solution

**First rule**
`If identity is in object_refs of a report and identity part-of identity, identity is in object_refs report (and the part-of relationship too)
`

** Second rule**
`If location is in object_refs of a report and location located-at location, location is in object_refs report (and the located-at relationship too)`

richard-julien:


SamuelHassine:
@richard-julien All good on my side, it just remains the computing of statistics. For instance, display a histogram of reports about France (displayed) and display a histogram of reports about Europe (not displayed).

SamuelHassine (approved):

PR #2444, opened 2022-10-12
[api] Filters support multiples keys to search on (#2435)

### LINKED ISSUE #2435: [api] Filters support multiples keys to search on
## Use case

Filters on "name" should also look into aliases

Kedae:
We can query multiple keys in GraphQL, for example filters: [{ key: [name, aliases], values: ["Alias1"] }]

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -6,7 +6,7 @@ steps:
   - name: api-tests
     image: node:16.10.0-alpine
     environment:
-      APP__BASE_URL: http://api-tests:4000/
+      APP__BASE_URL: http://api-tests:4010/
       APP__ADMIN__PASSWORD: admin
       APP__SYNC_RAW_START_REMOTE_URI: http://opencti-raw-start:4100/graphql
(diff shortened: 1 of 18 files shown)


opencti-platform/opencti-graphql/src/initialization.js
```diff
@@ -340,6 +340,7 @@ const platformInit = async (withMarkings = true) => {
       const reason = '[OPENCTI] Platform cant get the lock for initialization (can be due to other instance currently migrating/initializing)';
       throw LockTimeoutError({ participantIds: [PLATFORM_LOCK_ID] }, reason);
     } else {
+      logApp.error('[OPENCTI] Platform initialization fail', e);
```
richard-julien: Useful for testing but not in runtime that will duplicate the log.
For information e is also usually wrapped into error filed

richard-julien (changes requested):

richard-julien (approved):

PR #2445, opened 2022-10-12
[api] Introduce impersonate capability for bypass role (#1755)

### LINKED ISSUE #1755: Be able to select labels to import
## Use case

As an analyst, I want to able to select the labels I want and do not want in my opencti instance.

## Current Workaround

Delete the unwanted labels.

## Proposed Solution

Remove the right attributed to a connector to import labels.

richard-julien:
In the frame of "Be able to select labels to import #1755", the platform already have a capability to prevent a user to create labels.
The problem was more about a real impersonate of the user by the worker to fully respect the access rights of the connectors.

### DIFF
--- a/opencti-platform/opencti-graphql/src/domain/user.js
+++ b/opencti-platform/opencti-graphql/src/domain/user.js
@@ -636,12 +636,13 @@ export const authenticateUser = async (context, req, user, provider, token = '')
     if (isBypassUser(logged)) {
       const applicantUser = await resolveUserById(context, applicantId);
       if (isEmptyField(applicantUser)) {
-        throw UnsupportedError(`User ${applicantId} cant be impersonate (not exists)`);
+        logApp.warn(`User ${applicantId} cant be impersonate (not exists)`);
+      } else {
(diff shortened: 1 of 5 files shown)


opencti-platform/opencti-graphql/src/domain/user.js
```diff
@@ -625,9 +628,23 @@ export const userRenewToken = async (context, user, userId) => {
...
+  const logged = await buildCompleteUser(context, user);
   logAudit.info(userWithOrigin(req, user), LOGIN_ACTION, { provider });
-  const sessionUser = buildSessionUser(completeUser, provider);
+  let impersonate;
+  const applicantId = req.headers['opencti-applicant-id'];
+  if (isNotEmptyField(applicantId) && logged.id !== applicantId) {
+    if (isBypassUser(logged)) {
+      const applicantUser = await resolveUserById(context, applicantId);
+      if (isEmptyField(applicantUser)) {
+        throw UnsupportedError(`User ${applicantId} cant be impersonate (not exists)`);
```
Kedae: This error message is not accessible

Kedae (approved):

PR #2458, opened 2022-10-14
[api/front] Bump librairies + adapt front testing

richard-julien:


opencti-platform/opencti-front/jest/jest.file.transform.js
```diff
@@ -33,8 +33,9 @@ module.exports = {
           };
         }),
       };`;
+    } else {
+      code = `module.exports = ${assetFilename};`;
     }
-
-    return `module.exports = ${assetFilename};`;
-  },
+    return { code }
```
Kedae, line 39: Not sure how to test that, but not sure it is working anymore as you change the return type ? It was a string and is now a object ?

opencti-platform/opencti-front/src/private/components/Profile.js
```diff
@@ -31,40 +27,19 @@ const profileQuery = graphql`
...
-}
-
-Profile.propTypes = {
-  classes: PropTypes.object,
-  t: PropTypes.func,
+  );
 };
 
-export default R.compose(inject18n, withStyles(styles))(Profile);
+export default Profile;
```
Kedae: I think it would have been a great time to convert it to tsx :)

opencti-platform/opencti-front/src/__tests__/App.test.tsx
```diff
@@ -1,18 +1,112 @@
...
+});
+afterAll(() => {
+  console.error = originalError;
+});
+
+afterEach(cleanup);
+
+test('renders without crashing', async () => {
+  const environment = createMockEnvironment();
+  // eslint-disable-next-line arrow-body-style
```
Kedae: Should correct it instead of disabling

opencti-platform/opencti-front/jest/jest.file.transform.js
richard-julien: This is due to JEST new version

Kedae (approved):

PR #2468, opened 2022-10-18
[api] Add support of content_ref in Stix file for Artifact

Kedae:


richard-julien:
Early remarks
> UI modification for content_ref is missing 
> stixCyberObservableEdit > addRelation (del ..._ are not working. (not related to this task but could be fixed)

richard-julien (approved):

PR #2490, opened 2022-10-26
[frontend] Be able to expand all types of entity / relations in investigations (#1934)

### LINKED ISSUE #1934: Ability to expand to any kind of entity from Investigations Workspace
## Use case

In an Investigations workspace, users can expand to "all entities", "observables", and most/all SDOs. However, some individual SCOs aren't covered, such as ASN, URL, registry key, etc. It would be useful to include the ability to expand to all SCOs. 

## Current Workaround

Expand to all observables, then delete the ones that aren't wanted.

This is problematic because the Graph operations can be very memory intensive, and expanding to "all observables" sometimes isn't possible. 

## Proposed Solution

See above

## Additional Information

image

SamuelHassine:
### Proposed changes

* [frontend] Be able to expand all types of entity / relations in investigations (#1934)

### Related issues

* #1934


opencti-platform/opencti-front/src/private/components/common/form/TypesField.js
```diff
@@ -0,0 +1,162 @@
...
+    display: 'inline-block',
+    flexGrow: 1,
+    marginLeft: 10,
+  },
+  autoCompleteIndicator: {
+    display: 'none',
+  },
+}));
+
+const TypesField = (props) => {
```
Kedae: Props destructuration in method call is better and it will simplify our work for TS

opencti-platform/opencti-front/src/utils/Localization.js
```diff
@@ -897,6 +897,7 @@ const i18n = {
       'This sector does not have any description.':
         'Este sector no tiene ninguna descripción.',
       'Relationship type': 'Tipo de relación',
+      'Relationship types': 'Tipos de relaciones',
```
Kedae, line 917: Missing english ?

opencti-platform/opencti-front/src/utils/Security.tsx
```diff
@@ -32,14 +36,18 @@ export const SETTINGS_SETACCESSES = 'SETTINGS_SETACCESSES';
 export const SETTINGS_SETLABELS = 'SETTINGS_SETLABELS';
 
 interface SecurityProps {
-  children: React.ReactNode
-  needs: Array<string>
-  matchAll: boolean
+  children: React.ReactNode;
```
Kedae: You may have a wrong code style rule as the ; is useless here

opencti-platform/opencti-front/src/private/components/common/form/TypesField.js
```diff
@@ -0,0 +1,162 @@
...
+  const [dataTypes, setDataTypes] = useState([]);
+  const classes = useStyles();
+  const { t } = useFormatter();
+  const searchTypes = () => {
+    fetchQuery(typesFieldTypesQuery)
+      .toPromise()
+      .then((data) => {
+        let finaltypes = [];
+        if (types.includes('Stix-Domain-Object')) {
+          finaltypes = R.union(
```
RomuDeuxfois: Est-ce que ce n'est pas mieux d'utiliser la fonction javascript concat plutôt que le union de Ramda ?
Ou Ramda a un effet supplémentaire ?

richard-julien: Complete refactor done

opencti-platform/opencti-front/src/private/components/common/form/TypesField.js
richard-julien: done

opencti-platform/opencti-front/src/utils/Localization.js
richard-julien: Lot of missing translation on this screen.  Will be handle later.

Issue #2487, opened 2022-10-25
Empty channels type break the UI


richard-julien:
## Description

Empty channels type break the UI (frontend error)

+ Empty name for channel in data screen

PR #2492, opened 2022-10-26
[frontend] Fix channels list if types is empty (#2487)

SamuelHassine:
### Proposed changes

* [frontend] Fix channels list if types is empty (#2487)

### Related issues

* #2487


Kedae:
Apparently the Channel Type filter on the list is broken, could you fix it ?

Kedae (approved):

PR #2557, opened 2022-11-14
[api/frontend] Enhance search screen with table and number of reports (#2336, #2263)

### LINKED ISSUE #2263: Ability to search OpenCTI for a list of Observables (as opposed to one by one)
## Use case

It is useful to see if a list of IOCs have been observed in OpenCTI, and if so, in which Reports

## Current Workaround

Search every observable one by one

## Proposed Solution

An option by the global search bar (at the top right), to copy/paste a list of observables.
A global search would be performed on each one.
In order to avoid false positives, each search should be performed in as if there are double quotes around them.
Results should only be the direct hits, which should be clickable.

Ideally when hovering over the results, some basic information would pop up about in which situations the observable has been seen. Aka a list of reports, summary of relationships, etc

The results should be clickable (like a singular global search)

### LINKED ISSUE #2336: Bulk search of SDOs and SCOs
## Use case

Implement a way to easily search a bunch of SDOs or SCOs and know if they exist in the platform.

SamuelHassine:
### Proposed changes

* [api/frontend] Enhance search screen with table and number of reports ()
*

### Related issues

* #2336
* #2263

### DIFF
--- a/opencti-platform/opencti-front/src/utils/Localization.js
+++ b/opencti-platform/opencti-front/src/utils/Localization.js
@@ -64,7 +64,8 @@ const i18n = {
         '¿Quieres borrar esta fase de la «kill chain»?',
       'Do you want to delete this report?': '¿Quieres borrar este informe?',
       'Cascade delete': 'Eliminación en cascada',
-      'element(s) which are only in this report': 'elemento(s) que solo están en este informe',
+      'element(s) which are only in this report':
+        'elemento(s) que solo están en este informe',
       'Do you want to delete this grouping?':
         '¿Quieres eliminar esta agrupación?',
       'Do you want to kill this session?': '¿Quieres cerrar esta sesión?',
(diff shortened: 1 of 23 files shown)


opencti-platform/opencti-front/src/utils/Localization.js
```diff
@@ -2814,6 +2821,10 @@ const i18n = {
       IF: 'SI',
       THEN: 'ALORS',
       CREATE: 'CREER',
+      'Bulk search': 'Recherche de massage',
```
Kedae: De massage ? 🤭

SamuelHassine: J'avais vu hahaha. Je vais fixer plusieurs comportements dans la PR pour optimiser les recherches :)

Kedae:
Not sure to understand the bevavior. In the platform I have multiple objects with the same name (one file, one sector and one event) and if I bulk search by the name it only gets me the file ?

SamuelHassine:
> Not sure to understand the bevavior. In the platform I have multiple objects with the same name (one file, one sector and one event) and if I bulk search by the name it only gets me the file ?

Fixed.

PR #2575, opened 2022-11-17
Tool/relay hook template

Kedae:


opencti-platform/opencti-front/src/components/list_lines/ListLinesContent.js
```diff
@@ -32,8 +32,7 @@ class ListLinesContent extends Component {
   }
 
   componentDidUpdate(prevProps) {
-    const diff = R.symmetricDifferenceWith(
-      (x, y) => x.node.id === y.node.id,
+    const diff = !R.equals(
```
RomuDeuxfois, line 35: We can avoir Ramda by using

const diff = this.props.dataList.length === prevProps.dataList.length &&
   this.props.dataList..every(function (element) {
    return prevProps.dataList.includes(element);
  });

opencti-platform/opencti-front/src/private/components/common/form/ArtifactField.tsx
```diff
@@ -33,9 +33,9 @@ const ArtifactField: FunctionComponent<ArtifactFieldProps> = ({ attributeName, a
     .filter((f) => Boolean(f)) as StixCyberObservablesFiltering[];
   const data = useLazyLoadQuery<ArtifactFieldGetQuery>(artifactQuery, { filters });
 
-  const options = (data.stixCyberObservables?.edges ?? []).map((node) => ({
-    label: node?.node.observable_value ?? node?.node.id,
-    value: node?.node.id,
+  const options = (data.stixCyberObservables?.edges ?? []).map(({ node }) => ({
+    label: node.observable_value ?? node.id,
+    value: node.id,
```
RomuDeuxfois, line 38: I don't understand how do you handle the node.node to node

opencti-platform/opencti-front/src/private/components/entities/cities/CityCreation.tsx
```diff
@@ -0,0 +1,258 @@
...
+
+const CityCreation = ({ paginationOptions }: { paginationOptions: CitiesLinesPaginationQuery$variables }) => {
+  const classes = useStyles();
+  const { t } = useFormatter();
+
+  const [open, setOpen] = useState<boolean>(false);
+  const handleOpen = () => setOpen(true);
+  const handleClose = () => setOpen(false);
+
+  const [commit] = useMutation(cityMutation);
```
RomuDeuxfois, line 104: Why do we have this syntax ?

opencti-platform/opencti-front/src/private/components/entities/cities/CityEdition.tsx
```diff
@@ -0,0 +1,84 @@
...
+    overflow: 'auto',
+    transition: theme.transitions.create('width', {
+      easing: theme.transitions.easing.sharp,
+      duration: theme.transitions.duration.enteringScreen,
+    }),
+    padding: 0,
+  },
+}));
+
+const CityEdition = ({ cityId }: { cityId: string }) => {
```
RomuDeuxfois, line 34: Wy we don't have a FunctionComponent here ?

opencti-platform/opencti-front/src/private/components/entities/cities/CityKnowledge.tsx
```diff
@@ -0,0 +1,254 @@
...
+
+const cityKnowledgeFragment = graphql`
+  fragment CityKnowledge_city on City {
+    id
+    name
+    x_opencti_aliases
+  }
+`;
+
+const CityKnowledge = ({ cityData }: { cityData: CityKnowledge_city$key }) => {
```
RomuDeuxfois, line 34: FunctionComponent ?

opencti-platform/opencti-front/src/utils/hooks/useForceUpdate.ts
```diff
@@ -0,0 +1,11 @@
...
+
+const useForceUpdate = () => {
+  const [, setValue] = useState<object>({});
+
+  return useCallback((): void => {
+    setValue({});
+  }, []);
+};
+
+export default useForceUpdate;
```
RomuDeuxfois: Nice !

opencti-platform/opencti-front/src/utils/hooks/usePreloadedPaginationFragment.ts
```diff
@@ -0,0 +1,45 @@
...
+  const {
+    data,
+    hasNext,
+    loadNext,
+    isLoadingNext,
+  } = usePaginationFragment<QueryType, FragmentKey>(linesFragment, queryData);
+  useEffect(() => {
+    const deep_value = nodePath.reduce((a, v) => a[v as keyof object], data as object) as unknown[];
+    if (setNumberOfElements && deep_value) {
+      setNumberOfElements({ number: deep_value.length });
```
RomuDeuxfois, line 33: I don't understand why we do that

opencti-platform/opencti-front/src/components/list_lines/ListLinesContent.js
Kedae: Honestly I could but for deep equality Ramda is actually pretty good and I'm not sure we want to remove it (but the goal is still to remove Ramda as much as possible)

opencti-platform/opencti-front/src/private/components/entities/cities/CityEdition.tsx
Kedae: Because there is no need to, the component is quite simple so we don't need it.

opencti-platform/opencti-front/src/private/components/common/form/ArtifactField.tsx
Kedae: I destructure the node directly in the map

opencti-platform/opencti-front/src/private/components/entities/cities/CityCreation.tsx
Kedae: Hooks in react are always declared at the top of a component because they can't be optional, conditioned or called within an Element.

opencti-platform/opencti-front/src/private/components/entities/cities/CityKnowledge.tsx
Kedae: Same

opencti-platform/opencti-front/src/private/components/common/form/ArtifactField.tsx
richard-julien: deconstructed :)

opencti-platform/opencti-front/src/utils/hooks/usePreloadedPaginationFragment.ts
Kedae: Because the nb items of a list is shown in the Main component and the count is calculated in the Lines component. We need to share the info.
The useEffect is used to send the information everytime the list changes

opencti-platform/opencti-front/src/private/components/entities/Cities.tsx
```diff
@@ -0,0 +1,100 @@
...
+const Cities: FunctionComponent = () => {
+  const [viewStorage, _, storageHelpers] = useLocalStorage('view-cities', {
+    sortBy: 'name',
+    orderAsc: true,
+    searchTerm: '',
+    openExports: false,
+    filters: {},
+    numberOfElements: { number: 0, symbol: '' },
+  });
+  const finalFilters = convertFilters(viewStorage.filters) as unknown as Filters;
```
richard-julien, line 25: Any interest to directly convert the filters in the useLocalStorage hook?

opencti-platform/opencti-front/src/private/components/entities/Cities.tsx
```diff
@@ -0,0 +1,100 @@
...
+        filters={filters}
+        paginationOptions={paginationOptions}
+        numberOfElements={numberOfElements}
+        availableFilterKeys={[
+          'created_start_date',
+          'created_end_date',
+          'createdBy',
+        ]}
+      >
+        {queryRef && (
```
richard-julien, line 83: Need if suspense if used?

opencti-platform/opencti-front/src/private/components/entities/cities/CityCreation.tsx
```diff
@@ -0,0 +1,258 @@
...
+            onClick={handleClose}
+            size="large"
+            color="primary"
+          >
+            <Close fontSize="small" color="primary" />
+          </IconButton>
+          <Typography variant="h6">{t('Create a city')}</Typography>
+        </div>
+        <div className={classes.container}>
+          <Formik<CityAddInput>
```
richard-julien, line 161: useFormik ? :)

opencti-platform/opencti-front/src/private/components/entities/cities/CityEditionContainer.tsx
```diff
@@ -37,19 +38,41 @@ const styles = (theme) => ({
...
+    }
+  }
+`;
+const CityEditionContainer: FunctionComponent<CityEditionContainerProps> = ({ queryRef, handleClose }) => {
+  const classes = useStyles();
+  const { t } = useFormatter();
+
+  const queryData = usePreloadedQuery(cityEditionQuery, queryRef);
+
+  if (queryData.city) {
```
richard-julien: Can be undefined?

opencti-platform/opencti-front/src/private/components/entities/cities/CityEditionOverview.tsx
```diff
@@ -0,0 +1,390 @@
...
+      },
+      onCompleted: () => {
+        setSubmitting(false);
+        handleClose();
+      },
+    });
+  };
+
+  const initialValues = R.pipe(
+    R.assoc('createdBy', createdBy),
```
richard-julien: Could be cool to introduce a utils for this formik submit transforms

opencti-platform/opencti-front/src/private/components/entities/cities/Root.tsx
```diff
@@ -0,0 +1,193 @@
...
+
+  const link = `/dashboard/entities/cities/${cityId}/knowledge`;
+
+  const subConfig = useMemo<GraphQLSubscriptionConfig<RootCitiesSubscription>>(() => ({
+    subscription,
+    variables: { id: cityId },
+  }), [cityId]);
+  useSubscription(subConfig);
+
+  const data = useLazyLoadQuery<RootCityQuery>(cityQuery, { id: cityId });
```
richard-julien: Need to remain the old school way ? Any specific reason?

opencti-platform/opencti-front/src/private/components/entities/cities/Root.tsx
```diff
@@ -0,0 +1,193 @@
...
+  const { city, connectorsForExport } = data;
+
+  return (
+    <div>
+      <TopBar me={me || null} />
```
richard-julien, line 81: Could we just let TopBar handle undefined me?

opencti-platform/opencti-front/src/utils/hooks/useForceUpdate.ts
richard-julien: Personally dont get when we need to use this hook. Could you explain?

opencti-platform/opencti-front/src/utils/hooks/useQueryLoading.ts
```diff
@@ -0,0 +1,23 @@
...
+  variables: VariablesOf<T> = {},
+) => {
+  const [queryRef, loadQuery] = useQueryLoader<T>(query);
+  const varRef = useRef(variables);
+  if (!equals(variables, varRef.current)) {
+    varRef.current = variables;
+  }
+  // refetch when variables change
+  useEffect(() => {
+    loadQuery(variables);
```
richard-julien, line 17: Initial query should not be done in routing?

opencti-platform/opencti-front/src/utils/hooks/usePreloadedPaginationFragment.ts
richard-julien: +1 for the explanation :)

opencti-platform/opencti-front/src/utils/Store.js
```diff
@@ -11,7 +11,12 @@ export const isNodeInConnection = (payload, conn) => {
 export const insertNode = (store, key, filters, rootField) => {
   // Build record ids
   const record = store.get(store.getRoot().getDataID());
-  const conn = ConnectionHandler.getConnection(record, key, filters);
+
+  // TODO IT BREAKS CONNECTION ?
```
richard-julien: You can change that and put the explanation we talk about.

opencti-platform/opencti-front/src/utils/hooks/useForceUpdate.ts
Kedae: Actually I don't need it anymore.
We could have some case when we could need this but for now it is unused.

One case is : I use ref to store data because I don't want to be rendered everytime I update it, but some updates on this ref should render => useForceUpdate().

For now I'll just remove it.

opencti-platform/opencti-front/src/private/components/entities/Cities.tsx
Kedae: Done

opencti-platform/opencti-front/src/private/components/entities/Cities.tsx
Kedae: Seems so, the Suspens will handle the state when the "loadQuery" is called and the data not yet fetch but not the case when the ref doesn't exists yet. (As far as I've understood the doc).

Still, I will update this code a bit to have a loader when we have no ref.

opencti-platform/opencti-front/src/private/components/entities/cities/CityCreation.tsx
Kedae: As explained in the formik doc, useFormik is not yet the official way of doing stuff: https://formik.org/docs/api/useFormik

opencti-platform/opencti-front/src/private/components/entities/cities/CityEditionContainer.tsx
Kedae: City can be null

opencti-platform/opencti-front/src/private/components/entities/cities/Root.tsx
Kedae: It actually doesn't use `me` at all 😄

opencti-platform/opencti-front/src/utils/Store.js
Kedae: Done

opencti-platform/opencti-front/src/private/components/entities/cities/Root.tsx
Kedae: Actually it could be done the "proper" way, but it would require the main Root component to create all the refs of sub-rooting component and send refs. I've not found this very "readable" and chose this path, but we can discuss it.

opencti-platform/opencti-front/src/utils/hooks/useQueryLoading.ts
Kedae: Could be done. I've done it this way to have lighter code for now, we could change the initial-load behavior if we get performance/rendering issue in the future.

richard-julien (approved):

PR #2610, opened 2022-11-28
[front] Filters update to pure function

Kedae:


RomuDeuxfois:
If I understand properly, this is a split of the existing right ?

opencti-platform/opencti-front/src/private/components/common/lists/DialogFilters.js
```diff
@@ -0,0 +1,123 @@
...
+      </Tooltip>
+      <Dialog
+        PaperProps={{ elevation: 1 }}
+        open={open}
+        onClose={handleCloseFilters}
+        fullWidth={true}
+        maxWidth="md"
+      >
+        <DialogTitle>{t('Advanced search')}</DialogTitle>
+        <DialogContent style={{ paddingTop: 10 }}>
```
RomuDeuxfois, line 63: Should we have all the styles in the useStyles function ? (same questions for all components)

opencti-platform/opencti-front/src/private/components/common/lists/DialogFilters.js
```diff
@@ -0,0 +1,123 @@
...
+          {filters && !R.isEmpty(filters) && (
```
RomuDeuxfois: The implementation of Ramda is 
var isEmpty = _curry1(function isEmpty(x) {
  return x != null && equals(x, empty(x));
});

So, I am not sure we need the first filters condition

We can put this condition before rendering anything I think

opencti-platform/opencti-front/src/private/components/common/lists/FiltersElement.js
```diff
@@ -0,0 +1,157 @@
...
+                    onKeyDown={(event) => handleValidateDate(filterKey, event)}
+                    {...params}
+                  />
+                )}
+              />
+            </Grid>
+          );
+        }
+        let options = [];
+        if (['fromId', 'toId'].includes(filterKey)) {
```
RomuDeuxfois, line 91: Could be more readable to have a specific function for this code

opencti-platform/opencti-front/src/private/components/common/lists/ListFilters.js
```diff
@@ -0,0 +1,172 @@
...
+        }}
+        transformOrigin={{
+          vertical: 'top',
+          horizontal: 'center',
+        }}
+        elevation={1}
+      >
+        {filterElement}
+      </Popover>
+      {!noDirectFilters
```
RomuDeuxfois, line 100: Could be more readable to have a specific function for this code

opencti-platform/opencti-front/src/private/components/common/lists/Filters.js
```diff
@@ -1,1952 +1,205 @@
-import React, { Component } from 'react';
```
RomuDeuxfois, line 1: I agree with the new structure !

opencti-platform/opencti-front/src/private/components/common/lists/DialogFilters.js
Kedae: Yes, but no.
Yes, we should avoid inline style everytime we could and only use makeStyles.
But in some cases (like this one), we need to push style to a component so we can't rely on the computation from materialUi

opencti-platform/opencti-front/src/private/components/common/lists/DialogFilters.js
Kedae: Done

opencti-platform/opencti-front/src/private/components/common/lists/FiltersElement.js
Kedae: Not sure how we could make it more readable with a function ?

richard-julien:
Deepscan to fix.

+ List problem after selection

image

opencti-platform/opencti-front/src/private/components/common/lists/FiltersElement.js
RomuDeuxfois: I mean, to don't have it directly in the html/return part.

Kedae: Oh, yes we definitely should. For this task I choose not to update the way we do things because there is a lot to do and it was a side task from the OpenVocab one. But you're right, we should someday rewrite the Filters code to make it more friendly :)

opencti-platform/opencti-front/src/private/components/common/lists/ListFilters.js
Kedae: Same

PR #2630, opened 2022-12-06
[front/api] Add vocabulary as module

Kedae:


opencti-platform/opencti-front/src/private/components/settings/attributes/VocabularyEdition.tsx
```diff
@@ -42,43 +46,36 @@ const styles = (theme) => ({
...
-const attributeMutationUpdate = graphql`
-  mutation AttributeEditionUpdateMutation(
-    $id: String!
-    $previous: String!
-    $current: String!
+const vocabularyMutationUpdate = graphql`
+  mutation VocabularyEditionUpdateMutation(
+    $id: ID!
+    $category: VocabularyCategory!
+    $input: [EditInput!]
```
Archidoit: Compilation failed: mutation requires [EditInput!]! instead of [EditInput!]

Archidoit:
In Settings > Vocabularies, when we update a vocabulary : the aliases already entered are not written in the form (the aliases space is always empty).

Archidoit:
Unknown error occurs when trying to go to Data > Entities

Archidoit:
An unknown error occurs when trying to add a nested object to an observable.

opencti-platform/opencti-front/src/private/components/settings/attributes/VocabularyCreation.tsx
```diff
@@ -0,0 +1,214 @@
...
+  },
+  button: {
+    marginLeft: theme.spacing(2),
+  },
+}));
+
+const vocabularyAdd = graphql`
+  mutation VocabularyCreationMutation($input: VocabularyAddInput!) {
+    vocabularyAdd(input: $input) {
+      ...useVocabularyCategory_Vocabularynode
```
RomuDeuxfois, line 69: There are a reason why the fragment start by use ?

opencti-platform/opencti-front/src/utils/hooks/useDeletion.ts
```diff
@@ -0,0 +1,25 @@
+import { useState } from 'react';
+
+const useDeletion = ({ handleClose }: { handleClose: () => void }) => {
```
RomuDeuxfois, line 3: Should we maximize the use of hooks as best practices?

opencti-platform/opencti-front/src/utils/hooks/useEntityTypes.ts
```diff
@@ -0,0 +1,83 @@
...
+      [formatKey('stixCyberObservableTypes')]: optionBuilder('entity', data.stixCyberObservableTypes.edges.map(({ node }) => node)),
+    };
+  }
+
+  return (types?: string[]) => (types
+    ? types.flatMap((type) => EntityTypes[formatKey(type)])
+    : Object.values(EntityTypes).flat());
+};
+
+export default useEntityTypes;
```
RomuDeuxfois, line 83: This is never used

opencti-platform/opencti-graphql/src/modules/vocabulary/vocabulary-converter.ts
```diff
@@ -0,0 +1,16 @@
...
+
+const convertVocabularyToStix = (instance: StoreEntityVocabulary): StixVocabulary => {
+  const stixObject = buildStixObject(instance);
+  return {
+    ...stixObject,
+    name: instance.name,
+    description: instance.description,
+    category: instance.category,
+    aliases: instance.aliases ?? [],
+    builtIn: instance.builtIn || false,
```
RomuDeuxfois: Shouldn't we specify that this entity is an opencti extension?

extension_type : 'new-sdo'

opencti-platform/opencti-front/src/private/components/settings/attributes/VocabularyCreation.tsx
Kedae: Yep, because the vocabularyDefinitions is static, I added a hook to prevent to call them everytime we want it but only once. Therefore some of my Queries are in a hook

opencti-platform/opencti-front/src/utils/hooks/useDeletion.ts
Kedae: Totally. Hook = Life

opencti-platform/opencti-front/src/utils/hooks/useEntityTypes.ts
Kedae: I know :( But I don't want to throw some of the work I did, it will be used sometime when we will get further in converting elements

opencti-platform/opencti-graphql/src/modules/vocabulary/vocabulary-converter.ts
Kedae: I actually don't know :D, we will ask Julien

RomuDeuxfois:
When I edit a channel, I have the old Vocabularies -> fixed

image

PR #2646, opened 2022-12-14
Bugfix/release

RomuDeuxfois:
### Proposed changes

* Fix data source, data component and attack pattern icons
* Fix some traduction
* Re-fix filter converter errors

### Checklist

### Further comments

Some changes fixing filter conversion issues have been overwritten by a new commit.


opencti-platform/opencti-front/src/private/components/observations/StixCyberObservables.tsx
```diff
@@ -77,14 +76,21 @@ const StixCyberObservables: FunctionComponent = () => {
   Record<string, StixCyberObservableLine_node$data>
   >({});
 
-  const convertedFilters = convertFilters(
-    viewStorage.filters,
-  ) as unknown as Filters;
+  const finalType = () => {
```
Kedae: Can you remove the function and make it only a const ?

opencti-platform/opencti-front/src/private/components/observations/StixCyberObservables.tsx
```diff
@@ -77,14 +76,21 @@ const StixCyberObservables: FunctionComponent = () => {
...
+    if (types && types.length > 0) {
+      return types.map((n) => ({ id: n, value: n }));
+    }
+    return [{ id: 'Stix-Cyber-Observable', value: 'Stix-Cyber-Observable' }];
+  };
+  const exportFilters = {
+    entity_type: finalType(),
+    ...filters,
```
Kedae: You switch ordering, I don't think we want it

opencti-platform/opencti-front/src/private/components/observations/StixCyberObservables.tsx
Kedae: If it is not a function it could be added to the paginationMethod and prevent the exportFilters

Kedae (changes requested):

Issue #2681, opened 2022-12-26
[back] Prevent name update if already available in aliases


richard-julien:
## Use case

To prevent naming flickering, name must only be changed if not already available in aliases.

## Current
For example :

1. Create Name: MALWARE01, Aliases: ["MY ALIAS"]
2. Upsert Name: MY ALIAS, Aliases: ["MALWARE02"]

Result : MY ALIAS, Aliases: ["MY ALIAS", "MALWARE01", "MALWARE02"]

## Target

Result : MALWARE01, Aliases: ["MY ALIAS", "MALWARE02"]

PR #2682, opened 2022-12-26
[back] Prevent name update if already available in aliases (#2681)

richard-julien:


RomuDeuxfois (approved):

RomuDeuxfois:
Tested ❌:
- Non-functional with multiple words (maybe because of the capital letters -> normalize method)
[video](https://user-images.githubusercontent.com/53513584/209643468-69820ccb-50a0-4d3d-bd2b-f5f4222acf79.webm)


Comment : 
- There are no error or informational messages specifying this mechanism. The user could believe in a bug since the name does not update.

SamuelHassine (approved):

PR #2687, opened 2022-12-28
[front] Score is 'Unknown' if unset or equal to 0/100 if set to 0

### LINKED ISSUE #2654: x_opencti_score showing as "Unknown" is inconsistent
## Description

Observables with "x_opencti_score" key are not consistent. The web interface shows "Unknown" when the score is set to an IntType 0 value or a NoneType None value.

## Environment

1. OS (where OpenCTI server runs): Docker Desktop v4.15.0
2. OpenCTI version: OpenCTI 5.4.1
3. OpenCTI client: Web Interface

## Reproducible Steps

1. Create an artifact via the web interface
2. Create an observable via the web interface

## Expected Output

Observation's scores should be "Unknown" for either IntType value 0 OR NoneType value None but not for both. Would propose that IntType value 0 creates score of 0/100 while NoneType value None creates score of "Unknown".

## Actual Output

The artifact's x_opencti_score is set to None by default and the observable's x_opencti_score (being that it's required to be set) has to be an int (0) but both show score on the web interface of "Unknown"

## Additional information

Related bug: creating an observable via web interface requires a score value (IntType) but editing the observable allows for removal of score value (NoneType). Would propose that creation of observable score field is not set to required.

marieflorescontact:
### Proposed changes

* Creation of observable score field is not required
*  Score is 'Unknown' if unset or equal to 0/100 if set to 0

### Related issues

* 2654
*

### Checklist

### Further comments

### DIFF
--- a/opencti-platform/opencti-front/src/components/ItemScore.js
+++ b/opencti-platform/opencti-front/src/components/ItemScore.js
@@ -55,7 +55,7 @@ const inlineStyles = {
 const ItemScore = (props) => {
   const { score, classes, variant, t, theme } = props;
   const style = variant === 'inList' ? classes.chipInList : classes.chip;
-  if (!score) {
+  if (score === null) {
     return (
       <Chip
         classes={{ root: style }}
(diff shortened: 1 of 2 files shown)


opencti-platform/opencti-front/src/components/ItemScore.js
```diff
@@ -55,7 +55,7 @@ const inlineStyles = {
 const ItemScore = (props) => {
   const { score, classes, variant, t, theme } = props;
   const style = variant === 'inList' ? classes.chipInList : classes.chip;
-  if (!score) {
+  if (score === null) {
```
richard-julien: Could be interesting to use isEmptyField

richard-julien (approved):

PR #2693, opened 2023-01-02
[front/api] Fix on issue for relationTypes list in expand menu

Kedae:


RomuDeuxfois:
Can you provide some informations on which bug this pull request resolved ?
Thanks !

Kedae:
When go try to expand an entity on an investigation you can't filter the "Relation types" list and some items are in double.

RomuDeuxfois (approved):

PR #2759, opened 2023-01-19
[front] Rework of LocalStorage

Kedae:


opencti-platform/opencti-front/src/utils/filters/filtersUtils.tsx
```diff
@@ -3,6 +3,8 @@ export const FiltersVariant = {
   dialog: 'dialog',
 };
 
+export type BackendFilters = { key: string, values: string[], operator: string, filterMode: string }[];
```
RomuDeuxfois, line 6: It's only use for the backend and never show to the user ?

RomuDeuxfois (approved):

opencti-platform/opencti-front/src/private/components/analysis/Reports.tsx
```diff
@@ -63,45 +80,6 @@ const Reports: FunctionComponent<ReportsProps> = ({
...
   );
 
-  const { reportType } = match.params;
-
-  const reportFilterClass = reportType !== 'all' && reportType !== undefined
-    ? reportType.replace(/_/g, ' ')
-    : '';
-
-  const additionnalFilters = [];
-  if (reportFilterClass) {
```
RomuDeuxfois, line 73: Why are you removing that ?

RomuDeuxfois (approved):

opencti-platform/opencti-front/src/utils/filters/filtersUtils.tsx
Kedae: It's not about what is shown or not: it's the format. We don't have the same filter format for both front and back. I think there is a figma about that, you can ask Cathia.

opencti-platform/opencti-front/src/private/components/analysis/Reports.tsx
Kedae: I only removed the reportType part, the additionalFilters are above now.
The reportType is useless as it is part of the URL and filters and that's exactly what useLocalStorage is designed for, so it's redundant.

Issue #2760, opened 2023-01-19
BaseUri / BasePath is not always correctly set


richard-julien:
## Description

BaseUri / BasePath is not always correctly set.
We need to refactor a bit the method to be aligned everywhere in the code.
The idea is to always use the function *getBaseUrl* after adding correct basePath handling

# Hints

*access.ts*
> export const getBaseUrl = (req: Request): string
If baseUrl is defined, the basePath is not added.
Taxii root discovery with base path is not correctly defined because of it.

*externalReference.js*
> url: (externalReference) => (externalReference.fileId ? (baseUrl + externalReference.url) : externalReference.url),
BaseUrl is use directly without usage of basePath

*settings.js*
> platform_url: baseUrl,
Platform url dont take base path into account.

PR #2761, opened 2023-01-19
[back/front] BaseUri / BasePath is not always correctly set (#2760)

richard-julien:


opencti-platform/opencti-front/src/schema/relay.schema.graphql
```diff
@@ -1066,7 +1066,6 @@ type Settings implements InternalObject & BasicObject {
   platform_organization: Organization
   platform_title: String
   platform_favicon: String
-  platform_url: String
```
RomuDeuxfois, line 1069: It's still used in elastic.py file

RomuDeuxfois (approved):

opencti-platform/opencti-front/src/schema/relay.schema.graphql
richard-julien: Fixed

RomuDeuxfois (approved):

Issue #1400, opened 2021-06-08
Support Redis cluster


hieuit7:
## Problem to Solve

Redis is the only element in the stack that cannot be used in cluster.
Its due to the usage of multiple databases
Database 0 -> Stream and works
Database 1 -> User contexts and locks
Unfortunatly this feature is not available in Redis cluster. See https://redis.io/topics/cluster-spec

## Current Workaround

No

## Proposed Solution

Remove usage of Database 1 and use only the 0 to support cluster deployment.

richard-julien:
Hi @hieuit7 ,

For now Redis cluster is not supported.
Its basically due to the usage of multiple database to split concern that is not compatible with cluster mode (https://redis.io/topics/cluster-spec)

We plan to refactor and change that to support cluster mode but not short term.

richard-julien:
Content of the feature edited.

PR #2771, opened 2023-01-23
[back] Support Redis cluster (#1400)

richard-julien:
# Add support of the redis cluster

## Current situation

Redis is for now the only dependency that opencti doesnt support correctly when deployed in cluster.
This is due to multiple reasons:
* Client is not exactly the same to instanciate with ioredis when using single or cluster nodes.
* Use multiple database (that is not possible in cluster)
* Use keys scanning -> can be done but by changing the client api to scan all nodes in the cluster.
* Multi get of keys that can belong to different shards.

## Work to do
We need to change  a lot of aspects to support the redis clustering
* Client must be instanciate differently  to support single or cluster
* Keys must be collocated (locks) or distributed but without usage of multi get
* Scanning must be replace by SET with timeout mecanism to prevent multi node full scan.
* Locking keys must be collocated on the same shard (using {locks})
* Works key can be safely distributed

## Extra
* Redis clients will be not instanciate at starting
* Testing suite must be adapted to handle this new design and be fastest to execute.

### Deploy a easy redis cluster of 3 nodes locally

`docker run -e "IP=0.0.0.0" -p 7000-7005:7000-7005 grokzen/redis-cluster:latest`

### Configuration

Redis client in json configuration must now be defined like that for a cluster
```
  "redis": {
    "mode": "cluster",
    "hostnames": [
      { "host": "localhost", "port": 7000 },
      { "host": "localhost", "port": 7001 },
      { "host": "localhost", "port": 7002 }
    ]
  },
```


opencti-platform/opencti-graphql/src/database/redis.ts
```diff
@@ -44,15 +42,13 @@ import { telemetry } from '../config/tracing';
 import { filterEmpty } from '../types/type-utils';
 import type { ClusterConfig } from '../manager/clusterManager';
 
+export const REDIS_PREFIX = conf.get('redis:namespace') ? `${conf.get('redis:namespace')}:` : '';
 const USE_SSL = booleanConf('redis:use_ssl', false);
```
RomuDeuxfois, line 46: We still have redis:hostname property conf in cluster mode to be able to use ssl ?

opencti-platform/opencti-graphql/src/database/redis.ts
```diff
@@ -80,74 +73,197 @@ const redisOptions = (database: number): RedisOptions => ({
...
-      password: conf.get('redis:password'),
-      tls: USE_SSL ? { ...configureCA(REDIS_CA), servername: conf.get('redis:hostname') } : undefined,
-      lazyConnect: true,
-      maxRetriesPerRequest: 0,
-      showFriendlyErrorStack: DEV_MODE,
-    });
-    client.on('error', () => undefined /* do nothing */);
-    await client.get('test-key');
-    client.disconnect();
+    await getClientBase().get('test-key');
```
RomuDeuxfois, line 259: Why we have this key ?

opencti-platform/opencti-graphql/src/manager/connectorManager.js
```diff
@@ -94,7 +94,7 @@ const connectorHandler = async () => {
   let lock;
   try {
     // Lock the manager
-    lock = await lockResource([CONNECTOR_MANAGER_KEY]);
+    lock = await lockResource([CONNECTOR_MANAGER_KEY], { retryCount: 0 });
     running = true;
```
RomuDeuxfois, line 98: Why sometimes we need to set the retryCount ?

opencti-platform/opencti-graphql/builder/plugin/vitestMigrationPlugin.js
```diff
@@ -0,0 +1,26 @@
+import path from 'node:path';
+import fastGlob from 'fast-glob';
+
+const MIGRATIONS_IMPORT = 'import migrations, { filenames as migrationsFilenames } from \'../migrations/*.js\';'
+
+function vitestMigrationPlugin(opts = {}) {
+    return {
+        name: 'glob',
+        transform(code, id) {
```
RomuDeuxfois, line 9: Can you explain this code please ? I'm not sure to understand it.

SamuelHassine (approved):

RomuDeuxfois:
I am not sure it's working properly.

My conf file looks like this :
image

But, all the redis-node are used :
image

Explain by Julien -> Normal behavior

opencti-platform/opencti-graphql/builder/plugin/vitestMigrationPlugin.js
RomuDeuxfois: Explain by Julien -> re-write the glob plugin to be used by vitest

RomuDeuxfois (approved):

PR #2789, opened 2023-01-27
Improve pagination management + UI handling of disabled managers

richard-julien:


SamuelHassine (changes requested):
Functionally, not working as expected. 2 repro cases:

- Go to report, order by date
- Go to http://localhost:3000/dashboard/analysis/reports?sortBy=name&orderAsc=true
- It's now ordered by name
- Go to the home and then go back to reports, it's still ordered by date

--------

- Go to reports, order by name
- URL is rewritten and list ordered by name
- Go to home, go back to reports
- List is ordered by name but URL not rewritten

Issue #2735, opened 2023-01-12
Checkboxes / selection in all "Event" categories (incidents / sightings / observed data)


SamuelHassine:
## Use case

Checkboxes / selection in all "Event" categories (incidents / sightings / observed data)

PR #2819, opened 2023-02-06
[Front] Checkboxes / selection in all "Event" categories (incidents / sightings / observed data) #2735

Archidoit:
Add checkboxes / selection in the 3 "Event" categories:

- incidents
- sightings
- observed data

The functionnalities linked to selection should also be added:

- selection, deselection, selectAll
- export of selected elements only
- modification of selected elements via the toolbar

### Related issues

https://github.com/OpenCTI-Platform/opencti/issues/2735

### Associated PR in connectors
https://github.com/OpenCTI-Platform/connectors/pull/1016

### Associated page in Notion
https://www.notion.so/filigran/Checkboxes-selection-in-all-Event-categories-2735-0473e635161f41d3849c8614b20d9db5#67a6b74795f5456d97d3e755fe8261eb


opencti-platform/opencti-front/src/private/components/events/Incidents.js
```diff
@@ -67,6 +64,81 @@ class Incidents extends Component {
     this.setState({ openExports: !this.state.openExports });
   }
 
+  handleToggleSelectEntity(entity, event, forceRemove = []) {
```
RomuDeuxfois: Depending on the duration, this component should be migrated to use the useEntityToggle hook.
Or maybe, in the firstTime, just mutualize this code to facilitate future migration.

What do you think ?

opencti-platform/opencti-graphql/src/generated/graphql.ts
```diff
@@ -19921,6 +19921,8 @@ export type StixDomainObjectsFiltering = {
 
 export enum StixDomainObjectsOrdering {
   Score = '_score',
+  AttributeCount = 'attribute_count',
```
RomuDeuxfois: Why do we have these additions?

opencti-platform/opencti-front/src/private/components/events/ObservedDatas.js
```diff
@@ -65,6 +64,81 @@ class ObservedDatas extends Component {
     });
   }
 
+  handleToggleSelectEntity(entity, event, forceRemove = []) {
```
RomuDeuxfois: Same here

RomuDeuxfois:
Bug found : 
- Incident entity > The select all checkbox not update the selected elements number correctly
image
- Incident entity > The unselect all checkbox still check when we unselect a single chexkbox
image
- Incident entity > The select all checkbox not take all the elements on bulk actions
image

- Sighting entity > The unselect all checkbox still check when we unselect a single chexkbox

- Observed data entity > The unselect all checkbox still check when we unselect a single chexkbox
- Observed data entity > The select all checkbox not take all the elements on bulk actions
image

opencti-platform/opencti-graphql/src/generated/graphql.ts
Archidoit: it is taken into account in the export if the list is sorted by those columns. The export is not possible (raise error) if the column is not in StixDomainObjectsOrdering.

opencti-platform/opencti-front/src/private/components/events/ObservedDatas.js
Archidoit: same

opencti-platform/opencti-front/src/private/components/events/Incidents.js
Archidoit: yes good idea! what do you mean by 'mutualize' ? :)

RomuDeuxfois: I mean, put the logic in a file to use it, like the useEntityToggle file but in the old way.

Archidoit: ah yes ok I see !!

Archidoit:
The unselect all checkbox still check when we unselect a single chexkbox --> this behavior is present everywhere (special bug to fix maybe later in another issue)

opencti-platform/opencti-front/src/private/components/events/Incidents.js
Archidoit: waiting for the Incident refacto of Marie (we'll just have to add the useEntityToggle hook after the migration

RomuDeuxfois: So we waiting the merge of the Marie's PR
Good for me

RomuDeuxfois:
Bug founds : 
- The observed data screen is broken
image

Archidoit:
@RomuDeuxfois the error is not present for me...

RomuDeuxfois (approved):

Issue #2843, opened 2023-02-09
Address/Postal Code support on Position GUI


automated-tester:
## Use case

Positions include street_address and postal_code fields, but these are not currently editable through the GUI.

## Current Workaround

These fields can only be modified through the GraphQL resolvers.

## Proposed Solution

In PositionEditionOverview.js: Add two new Field components for street_address and postal_code. Add street_address and postal_code to positionValidation(), the render's initialValues, and to the GraphQL fragment.

## Additional Information

None.

## If the feature request is approved, would you be willing to submit a PR?

Yes

PR #2855, opened 2023-02-10
[Front] Address/Postal Code/City support on Position GUI (#2843)

Archidoit:
### PR content
- adding street_address and postal_code in the Position edition form
- adding street_address and postal_code in the Position creation form
- adding a 'Details' tab in Position>Overview
- adding missing translations for Position
- fixing Positions lines blinking (migrations to tsx)

### Related issues

https://github.com/OpenCTI-Platform/opencti/issues/2843

### Checklist


opencti-platform/opencti-front/src/private/components/locations/Positions.js
```diff
@@ -1,191 +0,0 @@
...
-    const params = buildViewParamsFromUrlAndStorage(
-      props.history,
-      props.location,
-      'view-positions',
-    );
-    this.state = {
-      sortBy: R.propOr('name', 'sortBy', params),
-      orderAsc: R.propOr(true, 'orderAsc', params),
-      searchTerm: R.propOr('', 'searchTerm', params),
-      view: R.propOr('lines', 'view', params),
```
RomuDeuxfois, line 33: The view was not used right ? This is why you remove it.
I don't see it on the master.

opencti-platform/opencti-front/src/private/components/locations/positions/PositionDetails.tsx
```diff
@@ -0,0 +1,238 @@
...
+interface PositionDetailsProps {
+  position: Position_position$data
+}
+
+const PositionDetails: FunctionComponent<PositionDetailsProps> = ({ position }) => {
+  const { t } = useFormatter();
+  const classes = useStyles();
+
+  return (
+    <QueryRenderer
```
RomuDeuxfois: It's will be better to use a suspense.

opencti-platform/opencti-front/src/private/components/locations/positions/PositionDetails.tsx
```diff
@@ -0,0 +1,238 @@
...
+        positionDetailsLocationRelationshipsLinesQuery
+      }
+      variables={{
+        count: 20,
+        elementId: [position.id],
+        relationship_type: 'located-at',
+      }}
+      render={({ props }: { props: PositionDetailsLocationRelationshipsLinesQueryLinesPaginationQuery$data }) => {
+        const targets = props?.stixCoreRelationships?.edges;
+        const cities = targets?.filter((n) => n?.node.to?.entity_type === 'City').map((n) => n?.node.to?.name);
```
RomuDeuxfois: If we have a relationship, the node, to, to.id and to.entity_type property should not be null, what do you think ?

opencti-platform/opencti-front/src/private/components/locations/positions/PositionEditionOverview.js
```diff
@@ -318,6 +322,32 @@ class PositionEditionOverviewComponent extends Component {
                 <SubscriptionFocus context={context} fieldName="longitude" />
               }
             />
+            <Field
```
RomuDeuxfois, line 325: We don't want this fields in creation as well ?

opencti-platform/opencti-front/src/private/components/locations/Positions.js
Archidoit: It was used but was always equal to 'lines' and it's value was never changed. There were no behavior for view != 'lines'. So it was not necessary to keep it.

RomuDeuxfois:
Tested : OK

We have no limit on the display of country, region, ect...
I don't know if it's possible to have a lot of them, but it can be a problem
image

--> the number of location to display is limited to 20 by security (even if it should never reach 20 in theory)

RomuDeuxfois:
Bug founds : I can't creation a Position right now

Kedae (approved):

PR #2861, opened 2023-02-10
[back] Improve live stream to continue to send Heartbeat during long resolutions 

### LINKED ISSUE #2860: Improve live stream to continue to send Heartbeat during long resolutions
### Problem
When live stream resolved refs and dependencies to ensure consistency, if the resolution is too long,  heartbeat can be too long and so the python client can consider it as dead socket.

We need to improve the live stream to continue to send Heartbeat during long resolutions.

+ Add some early stop when client lose the connection to stop as soon as possible current on going resolutions

### Proposal
Replace the current implementation that rely on the current processing to an separated interval that will continue to send heartbeat taking care of always sending the latest seen event id

richard-julien:


opencti-platform/opencti-graphql/src/graphql/sseMiddleware.js
```diff
@@ -190,26 +184,40 @@ const createSeeMiddleware = () => {
...
       },
     };
+    const heartTimer = () => {
+      if (lastEventId && !req.finished) {
+        const [idTime] = lastEventId.split('-');
+        const idDate = utcDate(parseInt(idTime, 10)).toISOString();
+        channel.sendEvent(lastEventId, 'heartbeat', idDate);
+      }
+    };
+    const heartbeatInterval = setInterval(heartTimer, 5000);
```
RomuDeuxfois: 5000 can be replaced by STREAM_BATCH_TIME ?

opencti-platform/opencti-graphql/src/graphql/sseMiddleware.js
```diff
@@ -190,26 +184,40 @@ const createSeeMiddleware = () => {
...
+        if (!req.finished) {
+          try {
+            res.end();
+          } catch (e) {
+            logApp.error('[STREAM] Failing to close client', { clientId: channel.userId, error: e });
+          }
         }
       },
     };
+    const heartTimer = () => {
```
RomuDeuxfois, line 290: I am not sure to understand this new logic.
Before, we sent Hearthbeat at the end of an action and now, we send it in interval ?

opencti-platform/opencti-graphql/src/graphql/sseMiddleware.js
richard-julien: Not the same purpose, however we could introduce a constant STREAM_HEARTBEAT

opencti-platform/opencti-graphql/src/graphql/sseMiddleware.js
richard-julien: Yes thats the point to detached the heartbeat from the processing, to prevent client kill the connection because nothing happens in the stream during long resolution

RomuDeuxfois (approved):

PR #2871, opened 2023-02-13
[back] Marking definition completion in entity history

### LINKED ISSUE #2869: Marking definition information leak in entity history
## Description

It looks like the history entries for changes to entities mostly properly obey the "marking definition" restrictions at the time the action occured. However, when deleting a marking definition in an entity, the "history" entry for the removal will be written using the "marking definition" visibility constraints after the operation has occurred. This results in the change, which includes the (potentially private) name of the marking definition, being visible by any user with a potentially lower level of access.

When a marking definition is deleted from an entity, the edit history entry for that operation should be stored with the access/visibility restriction that was in place prior to deletion of the marking definition.

## Environment

1. Ubuntu 22.04 LTS
2. OpenCTI version: 5.5.4
3. OpenCTI client: 5.5.4
4. Other environment details: Using `docker` containers

## Reproducible Steps

* Make "Group A" with permission to TLP:RED, TLP:AMBER, TLP:GREEN, TLP:CLEAR
* Make "Group B" with permission to TLP:CLEAR and TLP:GREEN
* Make "User 1" in "Group A"
* Make "User 2" in "Group B"
* Have "User 1" log in and create a new IP address 5.5.5.5
* Have "User 1" give this the "TLP:CLEAR" and "TLP:AMBER" markings. This will make it so that the IP is only accessible to members of "Group A" but not "Group B"
* Validate this is the case with "User 2"
* Go back to "User 1" and edit the IP 5.5.5.5: Delete the "TLP:AMBER" marking definition, leaving the "TLP:CLEAR" one alone
* Using "User 2" go to the 5.5.5.5 IP address entity and look at the history
* "User 2" should be able to see an entry like `User 1 removes TLP:AMBER in object_marking_refs` despite still not having access to the "TLP:AMBER" marking definition

## Expected Output

The change saying `User 1 removes TLP:AMBER in object_marking_refs` should only be visible to users who have both the "TLP:CLEAR" and "TLP:AMBER" permissions (in my example, members of Group 1)

## Actual Output

Anyone with TLP:CLEAR permission will be able to see that "User 1" removed the "TLP:AMBER" restriction from the entity. The information that the entity was previously marked as "TLP:AMBER", who changed it, and potentially the name of the marking itself, should (IMO) be considered information restricted to users with that permission level.

richard-julien:


opencti-platform/opencti-graphql/src/manager/historyManager.ts
```diff
@@ -80,18 +82,26 @@ export const eventsApplyHandler = async (context: AuthContext, events: Array<Sse
       const updateEvent: UpdateEvent = event.data as UpdateEvent;
       contextData.commit = updateEvent.commit?.message;
       contextData.external_references = updateEvent.commit?.external_references ?? [];
+      // Previous markings must be kept to ensure data visibility restrictions
+      const { newDocument: previous } = jsonpatch.applyPatch(R.clone(stix), updateEvent.context.reverse_patch);
+      const previousMarkingRefs = (previous.object_marking_refs ?? [])
+        .map((stixId) => markingsById.get(stixId)?.internal_id)
+        .filter((o) => isNotEmptyField(o));
+      eventMarkingRefs.push(...previousMarkingRefs);
```
RomuDeuxfois, line 90: I am not able to find where the change is effective

Kedae (approved):

Issue #2856, opened 2023-02-10
Delete a vocabulary let the dialog opened (and redirect instead of removing the node from the store)


SamuelHassine:
## Description

In Settings => Labels & Attributes => Vocabs => One of them

When deleting a value, the confirm dialog stays open. When it works (sometimes), it redirects instead of removing the node from the store.

PR #2880, opened 2023-02-14
[Front] Vocabulary deletion bug fix (#2856)

Archidoit:
### Related issues

https://github.com/OpenCTI-Platform/opencti/issues/2856

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/settings/attributes/VocabularyPopover.tsx
+++ b/opencti-platform/opencti-front/src/private/components/settings/attributes/VocabularyPopover.tsx
@@ -76,7 +76,6 @@ const VocabularyPopover: FunctionComponent<VocabularyPopoverProps> = ({ vocab, p
     setDisplayUpdate(false);
     refetch();
   };
-
   let deleteLabel = t('Delete');
   const deletable = !vocab.builtIn && (!vocab.category.fields.some(({ required }) => required) || vocab.usages.length === 0);
   if (!deletable) {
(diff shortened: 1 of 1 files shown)


opencti-platform/opencti-front/src/private/components/settings/attributes/VocabularyPopover.tsx
```diff
@@ -96,18 +95,19 @@ const VocabularyPopover: FunctionComponent<VocabularyPopoverProps> = ({ vocab, p
   } = useDeletion({ handleClose });
   const [commit] = useMutation(VocabularyPopoverDeletionMutation);
 
-  const submitDelete = () => {
+  const submitDelete = (vocabulary: useVocabularyCategory_Vocabularynode$data) => {
```
RomuDeuxfois: It's supposed to work with the vocab property no ?
It seems strange to me to have to pass this property via this method

Archidoit: no it doesn't work, since vocab takes the value of an other vocab of the list once we've deleted the current one...

Archidoit: the bug comes from that point

Archidoit: you're right !

Archidoit: fixed

Kedae (approved):

PR #2887, opened 2023-02-14
[back] Disable cache for dataloader (#2878)

### LINKED ISSUE #2878: Memory leak issue due to misuse of the dataloader
## Description

I have several connectors. I keep them pushing "_indicator_" data to MQ. There are 4 workers fetching the data from MQ and sending the `indicatorAdd` request to the opencti. Within one hour, there are around 58000 "_indicator_" bundles in total sent to the opencti, i.e., that is 58000 `indicatorAdd` requests. (According to the opencti worker logic, it is one "_indicator_" data in one bundle).

The memory keeps going up from the beginning. A couple of hours later, the memory can reach like 3-4G, and eventually will be killed by OOM killer.

Found in version 5.3.6, 5.4.0, 5.5.4.

## Root cause

The "query" part of the `indicatorAdd` request is as follows,
```graphql
mutation IndicatorAdd($input: IndicatorAddInput) {
    indicatorAdd(input: $input) {
        id
        standard_id     
        entity_type     
        parent_types    
        observables {   
            edges {         
                node {          
                    id              
                    standard_id     
                    entity_type     
                }               
            }
        }
    }
}
```

(The handler/resolver of the `indicatorAdd` query is the function `addIndicator()` (in `opencti-platform/opencit-graphql/src/domain/indicator.js`. This function has no problem.)

According to the query above, the opencti (precisely, [the GraphQL framework](https://github.com/graphql/graphql-js)) will put the following fields of the added indicator to the response,
```
id
standard_id
entity_type
parent_types
observables
```

The problem is right in the handler/resolver of the field `observables`, see below,
```js
  Indicator: {
    killChainPhases: (indicator, _, context) => killChainPhasesLoader.load(indicator.id, context, context.user),
    observables: (indicator, _, context) => batchObservablesLoader.load(indicator.id, context, context.user),
  },
```
In which, the `batchObservablesLoader` is a global variable, see below,
```js
const batchObservablesLoader = batchLoader(batchObservables);
```
In which, the `batchLoader` is a function, see below,
```js
export const batchLoader = (loader) => {
  const dataLoader = new DataLoader(
    (objects) => {
      const { context, user, args } = R.head(objects);
      const ids = objects.map((i) => i.id);
      return loader(context, user, ids, args);
    },
    { maxBatchSize: MAX_BATCH_SIZE }
  );
  return {
    load: (id, context, user, args = {}) => {
      return dataLoader.load({ id, context, user, args });
    },
  };
};
```
In which, a [DataLoader](https://github.com/graphql/dataloader) instance is created, and is held all the time from the opencti starts to the opencti ends.

There is a memoization cache in the DataLoader. After check the code of DataLoader, I found there is only the code to put data to the cache, no code to delete data from the cache, which means the cache can grow as much as possible until use out of the memory. And this is where the memory leak happens.

But I won't blame the DataLoader. Because in the [README.md](https://github.com/graphql/dataloader/blob/v2.2.2/README.md) of the DataLoader, you can find such words as follows,
> DataLoader provides a memoization cache for all loads which occur **_in a single request_** to your application. After .load() is called once with a given key, the resulting value is cached to eliminate redundant loads.

and this

> As mentioned above, DataLoader is intended to be used **_as a per-request cache_**. Since requests are short-lived, DataLoader uses an infinitely growing [Map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) as a memoization cache. This should not pose a problem as most requests are short-lived and the entire cache can be discarded after the request completes.
>
> **_However this memoization caching strategy isn't safe when using a long-lived DataLoader, since it could consume too much memory._** If using DataLoader in this way, you can provide a custom Cache instance with whatever behavior you prefer, as long as it follows the same API as [Map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map).

These words are saying
1. The lifespan of a dataloader can only be within a request;
2. Otherwise, it will be able to cause memory leak.

But the way the opencti uses the dataloader is
1. the dataloader is held by a global variable, which makes it possible for the cache to grow infinitively;
2. and, the cache key that the opencti set is `{ id, context, user, args }`, which makes the cache unlikely to be hit, because the `context` contains a _session-created-time_, which differs from request to request.

These two usage together make the memory leak happen.

Actually, For any GraphQL query, as long as its handler/resolver is like the `batchLoader().load()`, it will cause the memory leak issue.

## Resolution

Use the DataLoader without cache. Since the cache can never be hit, it doesn't make sense to use the dataloader with cache.
```diff
   const dataLoader = new DataLoader(
     (objects) => {
       const { context, user, args } = R.head(objects);
       const ids = objects.map((i) => i.id);
       return loader(context, user, ids, args);
     },
-    { maxBatchSize: MAX_BATCH_SIZE }
+    { maxBatchSize: MAX_BATCH_SIZE, cache: false }
   );
```

richard-julien:
Because dataloader use a simple Map as cache.
Cache is not needed as we always fetch a list of uniq ids

### DIFF
--- a/opencti-platform/opencti-graphql/src/database/middleware.js
+++ b/opencti-platform/opencti-graphql/src/database/middleware.js
@@ -258,7 +258,7 @@ export const batchLoader = (loader) => {
       const ids = objects.map((i) => i.id);
       return loader(context, user, ids, args);
     },
-    { maxBatchSize: MAX_BATCH_SIZE }
+    { maxBatchSize: MAX_BATCH_SIZE, cache: false }
   );
   return {
     load: (id, context, user, args = {}) => {


RomuDeuxfois:
Is it not the return of the load method that is cached? Base on the list of ids as a key.

ckyoog:
@RomuDeuxfois 

It is the return of the load method that is cached. The key is not the list of ids, the key is `{id, context, user, args}`.

RomuDeuxfois (approved):

PR #2888, opened 2023-02-15
[back/front] Cumulate creator_id when upserting an entity

### LINKED ISSUE #2886: Cumulate creator_id when upserting an entity
## Use case

When an entity is created, the `creator_id` is set with the ID of the user creating the entity. If another user tries to create the same entity after and nothing is updated, the information is lost. Even if something is modified, technically the information is lost (no filter is possible) except in the history.

We need to handle `creator_id` as a list of strings, and when a creation occurs and the entity already exists, insert the user ID in this list (if not already in it). 

Impacts to take into account:

* Dynamic resolver in GraphQL for `creator` should be a list of "Creator" (instead of a Creator)
* Be sure that filters are working when filtering on multiple technical creators:
  * Example: Be able to list all indicators created by AlienVault and CrowdStrike

Thanks to this feature, when filtering a list of entities with a technical creator, ALL entities created by this user will be listed, including the entities that already existed because they were created by another user.

Also, it will allow organizations to create TAXII collections / CSV feeds / stream with the source(s) they want without missing anything.

richard-julien:


opencti-platform/opencti-front/src/private/components/analysis/Groupings.tsx
```diff
@@ -119,7 +119,7 @@ const Groupings: FunctionComponent<GroupingsProps> = (
         isSortable: isRuntimeSort ?? false,
       },
       creator: {
-        label: 'Creator',
+        label: 'Creators',
```
RomuDeuxfois, line 125: Miss I18n

opencti-platform/opencti-graphql/src/domain/user.js
```diff
@@ -138,13 +138,21 @@ export const batchGroups = async (context, user, userId, opts = {}) => {
   return batchListThroughGetTo(context, user, userId, RELATION_MEMBER_OF, ENTITY_TYPE_GROUP, opts);
 };
 
-export const batchCreators = async (context, user, userIds) => {
-  const internalUserIds = Object.keys(INTERNAL_USERS);
+const internalUserIds = Object.keys(INTERNAL_USERS);
+export const batchCreator = async (context, user, userIds) => {
```
RomuDeuxfois, line 142: Now, only workspace can have one creator ?

PR #2899, opened 2023-02-17
[back] Improve cluster configuration + platform stops when redis is not available (#2875)

### LINKED ISSUE #2875: Improve Redis cluster configuration + platform stops when redis is not available
## Description

Improve Redis cluster configuration > Hostnames must be converted to a simple list of string.
`"hostnames": ["localhost:7000", "localhost:7001", "localhost:7002"]`
and for nat mapping if needed

`nat_map: ['10.0.1.230:30001>203.0.113.73:30001', '10.0.1.231:30001>203.0.113.73:30002']`

When Redis becomes unavailable, platform stops in some conditions:

```
MaxRetriesPerRequestError: Reached the max retries per request limit (which is 10). Refer to "maxRetriesPerRequest" option for details.
    at Socket.<anonymous> (/var/data/app/opencti-platform/opencti-graphql/build/node_modules/ioredis/built/redis/event_handler.js:177:37)
    at Object.onceWrapper (node:events:628:26)
    at Socket.emit (node:events:513:28)
    at TCP.<anonymous> (node:net:313:12)
```

richard-julien:


opencti-platform/opencti-graphql/src/database/redis.ts
```diff
@@ -70,11 +70,22 @@ const redisOptions = (): RedisOptions => ({
   showFriendlyErrorStack: DEV_MODE,
 });
 
-const generateNatMap = (mappings: [{ nat: string; host: string; port: number; }]): Record<string, { host: string; port: number; }> => {
+// From "HOST:PORT" to { host, port }
+export const generateClusterNodes = (nodes: string[]): { host: string; port: number; }[] => {
```
lhviet, line 74: I'm wondering if using the delimiter `-` may wrongly parse the host like `a-b-c.redis.com`. Do you think using ` ` (space) or some non-qualified-url character will work also?
I think to avoid the below chars are recommended.

> In general URIs as defined by [RFC 3986](https://www.rfc-editor.org/rfc/rfc3986/) (see [Section 2: Characters](https://www.rfc-editor.org/rfc/rfc3986#section-2)) may contain any of the following 84 characters:
`ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789-._~:/?#[]@!$&'()*+,;=`

Ref. https://stackoverflow.com/questions/1547899/which-characters-make-a-url-invalid

richard-julien: Sure, good comment. we can use a whitespace or other seperator for the nap part.

"HOST:PORT HOST:PORT"

or

"HOST:PORT>HOST:PORT"

richard-julien: Changed to "HOST:PORT>HOST:PORT"

lhviet (approved):
Great job.

RomuDeuxfois (approved):

PR #2913, opened 2023-02-21
[Front] Fix not update input values in some filters

RomuDeuxfois:
### Proposed changes

* Fix not update input values in some filters like stream/taxi creation


opencti-platform/opencti-front/src/private/components/common/lists/Filters.js
```diff
@@ -109,6 +109,7 @@ const Filters = ({
   } else {
     handleUpdateDateInputValues(usedFilters);
```
marieflorescontact: Don't you need to change the condition in the  `handleUpdateDateInputValues` method because it will always be true as you give a default value to  `usedFilters = {}` ?
```suggestion
handleUpdateDateInputValues(usedFilters)
    if (filtersToApply !== {} ) {
```

Issue #2928, opened 2023-02-24
[backend] X-TAXII-Date-Added-First/Last response headers are broken


rlynch-ironnet:
## Description

In the data sharing stix/taxii feed, the `X-TAXII-Date-Added-First/Last` response headers are always null/undefined.

## Environment

1. OS (where OpenCTI server runs): AWS fargate
2. OpenCTI version: 5.5.4
3. OpenCTI client: backend
4. Other environment details: n/a

## Reproducible Steps

Add a stix object.
Make a feed.
Curl any of the commands below.

## Actual/Expected Output

```
✗ curl_args=("-qs" "-H" "Accept: application/taxii+json;version=2.1" "--basic" "-u" "admin@opencti.io:password")
✗ collection_id="7b7b5633-0ca1-42cf-93e4-540fd8702046"
✗ object_id="indicator--45aedf95-8195-5b4f-a034-070df3198211"
✗ taxii_url="http://127.0.0.1:8080/taxii2/root/collections/$collection_id"

✗ curl $curl_args $taxii_url/objects | jq ".objects[] | select (.id == \"${object_id}\")"
{
  "id": "indicator--45aedf95-8195-5b4f-a034-070df3198211",
  "extensions": {
    "extension-definition--ea279b3e-5c71-4632-ac08-831c66a786ba": {
      "created_at": "2023-02-24T16:25:40.503Z",
      "updated_at": "2023-02-24T16:25:40.523Z",
      ...
    }
  },
  "created": "2023-02-24T16:24:30.766Z",
  "modified": "2023-02-24T16:25:40.523Z",
  ...
}

# === /manifest ===
# Actual
✗ curl --head $curl_args $taxii_url/manifest | grep -i X-TAXII
X-TAXII-Date-Added-First: undefined
X-TAXII-Date-Added-Last: undefined

# Expected
✗ curl --head $curl_args $taxii_url/manifest | grep -i X-TAXII
X-TAXII-Date-Added-First: 2023-02-24T16:25:27.997Z
X-TAXII-Date-Added-Last: 2023-02-24T16:25:40.523Z

# === /objects ===
# Actual
✗ curl --head $curl_args $taxii_url/objects | grep -i X-TAXII
X-TAXII-Date-Added-First: undefined
X-TAXII-Date-Added-Last: undefined

# Expected
✗ curl --head $curl_args $taxii_url/objects | grep -i X-TAXII
[...]

PR #2929, opened 2023-02-24
[back] Fix X-TAXII-Date-Added-First/Last headers

rlynch-ironnet:
### Proposed changes

* Fix how the X-TAXII-Date-Added-First/Last headers are populated

### Related issues

* #2928

### Checklist

### Further comments

This needs discussion on fixing how the /manifest endpoint will end up needing 2 queries without making further changes.

### DIFF
--- a/opencti-platform/opencti-graphql/src/http/httpTaxii.js
+++ b/opencti-platform/opencti-graphql/src/http/httpTaxii.js
@@ -3,6 +3,7 @@ import * as R from 'ramda';
 import { authenticateUserFromRequest, TAXIIAPI } from '../domain/user';
 import { basePath, getBaseUrl } from '../config/conf';
 import { AuthRequired, ForbiddenAccess, UnsupportedError } from '../config/errors';
+import { STIX_EXT_OCTI } from '../types/stix-extensions';
 import {
   restAllCollections,
   restCollectionManifest,
@@ -44,6 +45,9 @@ const rebuildParamsForObject = (id, req) => {
   const { spec_version, version } = match;
   const argsMatch = { id, spec_version, version };
   return { added_after, limit, next, match: argsMatch };
+};
+const getUpdatedAt = (obj) => {
+  return obj?.extensions?.[STIX_EXT_OCTI]?.updated_at;
 };
 
 const initTaxiiApi = (app) => {
@@ -111,8 +115,8 @@ const initTaxiiApi = (app) => {
       const context = executionContext('taxii');
       const user = await extractUserFromRequest(context, req, res);
       const manifest = await restCollectionManifest(context, user, id, req.query);
-      res.set('X-TAXII-Date-Added-First', R.head(manifest.objects)?.updated_at);
-      res.set('X-TAXII-Date-Added-Last', R.last(manifest.objects)?.updated_at);
+      res.set('X-TAXII-Date-Added-First', R.head(manifest.objects)?.version);
+      res.set('X-TAXII-Date-Added-Last', R.last(manifest.objects)?.version);
       res.json(manifest);
     } catch (e) {
       const errorDetail = errorConverter(e);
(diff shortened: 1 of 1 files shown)


rlynch-ironnet:
Leaving this as a WIP until we figure out if domain/taxii.js should be updated, or if this is a good enough fix for now.

opencti-platform/opencti-graphql/src/http/httpTaxii.js
```diff
@@ -45,6 +45,9 @@ const rebuildParamsForObject = (id, req) => {
   const argsMatch = { id, spec_version, version };
   return { added_after, limit, next, match: argsMatch };
 };
+const getUpdatedAt = (obj) => {
+  return obj?.extensions?.['extension-definition--ea279b3e-5c71-4632-ac08-831c66a786ba']?.updated_at;
```
richard-julien: Can you use the constant STIX_EXT_OCTI ? Thanks

richard-julien (changes requested):

rlynch-ironnet:
@richard-julien regarding `/manifest` doing a double query, should this PR refactor the functions to handle this, or deal with it in a later release?

richard-julien:
> @richard-julien regarding `/manifest` doing a double query, should this PR refactor the functions to handle this, or deal with it in a later release?

Can you explain what you mean by `/manifest doing a double query`

Thanks

rlynch-ironnet:
Previously, to get the manifest list, `/manifest` would call `const manifest = await restCollectionManifest(context, user, id, req.query);` which does a backend search. Each manifest object only has the dateAdded, not any dateModified, so we need to pull the stix objects as well, which performs another search via `const stix = await restCollectionStix(context, user, id, req.query);`

A better fix, would modify restCollectionManifest/Stix to accept a list of search objects, so only one search is performed instead.

richard-julien:
> Previously, to get the manifest list, `/manifest` would call `const manifest = await restCollectionManifest(context, user, id, req.query);` which does a backend search. Each manifest object only has the dateAdded, not any dateModified, so we need to pull the stix objects as well, which performs another search via `const stix = await restCollectionStix(context, user, id, req.query);`
> 
> A better fix, would modify restCollectionManifest/Stix to accept a list of search objects, so only one search is performed instead.

Oh yes i miss read the file.

I think you can use directly data fetch from the manifest
```

  return {
    id: data.standard_id,
    date_added: data.created_at,
    version: data.updated_at,
    media_type: STIX_MEDIA_TYPE,
  };
```

The updated_at information is contains in version field

rlynch-ironnet:
That should do the trick then.

opencti-platform/opencti-graphql/src/http/httpTaxii.js
```diff
@@ -156,9 +160,10 @@ const initTaxiiApi = (app) => {
       const args = rebuildParamsForObject(object_id, req);
       const stix = await restCollectionStix(context, user, id, args);
       const data = R.head(stix.objects);
-      res.set('X-TAXII-Date-Added-First', data?.updated_at);
-      res.set('X-TAXII-Date-Added-Last', data?.updated_at);
-      const versions = data ? [data.updated_at] : [];
+      const updatedAt = getUpdatedAt(data);
+      res.set('X-TAXII-Date-Added-First', updatedAt);
+      res.set('X-TAXII-Date-Added-Last', updatedAt);
```
richard-julien, line 165: Already problematic before your modification, but this should not be taken for last element?

rlynch-ironnet: In theory, we'd want to do something like this
```
      const versions = stix.objects.map(o => getUpdatedAt(o));
      res.set('X-TAXII-Date-Added-First', R.head(versions));
      res.set('X-TAXII-Date-Added-Last', R.last(versions));
      res.json({ versions });
```
That is, if OpenCTI supported versioned objects.

richard-julien (approved):

PR #2933, opened 2023-02-27
[back] - Improve of the query for huge graphs

Kedae:


opencti-platform/opencti-graphql/src/domain/container.js
```diff
@@ -68,7 +68,17 @@ export const containersObjectsOfObject = async (context, user, { id, types, filt
...
+  const containerTypedObjectsRelationships = {};
+  containersObjectsRelationships.forEach(({ node }) => {
+    if (!containerTypedObjectsRelationships[node.toType]) {
+      containerTypedObjectsRelationships[node.toType] = [];
+    }
+    if (!containerTypedObjectsRelationships[node.toType].includes(node.toId)) {
+      containerTypedObjectsRelationships[node.toType].push(node.toId);
+    }
+  });
+  const containersObjects = R.flatten(await Promise.all(Object.entries(containerTypedObjectsRelationships)
```
richard-julien: Need to be discuss about how that can solve the problem.
And also the usage of await Promise.map(entries, function, { concurrency: ES_MAX_CONCURRENCY });

richard-julien (approved):

PR #2982, opened 2023-03-06
[back/front] - Fix sur marking searching on the platform

### LINKED ISSUE #2930: When more than 200 markings exists in the system, user build is failing
## Description

When more than 200 markings exists in the system, we have 
- User build is failing > login is impossible
- Marking filter is not complete, some markings are not selectable.
This is due to a default limitation that only fetch the first 200 elements.

We need to change in the user builder to always fetch all the markings and adapt the combobox.

Kedae:


opencti-platform/opencti-graphql/src/domain/user.js
```diff
@@ -182,7 +182,7 @@ const getUserAndGlobalMarkings = async (context, userId, userGroups, capabilitie
   const groupIds = userGroups.map((r) => r.id);
   const userCapabilities = map((c) => c.name, capabilities);
   const shouldBypass = userCapabilities.includes(BYPASS) || userId === OPENCTI_ADMIN_UUID;
-  const allMarkingsPromise = allMarkings(context, SYSTEM_USER).then((data) => R.map((i) => i.node, data.edges));
+  const allMarkingsPromise = allMarkings(context, SYSTEM_USER, { first: MAX_SEARCH_SIZE }).then((data) => R.map((i) => i.node, data.edges));
```
richard-julien: AllMarkings should use a list all entities instead ?

richard-julien: Or better in the cache manager ?

richard-julien (approved):

PR #3001, opened 2023-03-09
[back] Adapt rel merging + migration

richard-julien:


RomuDeuxfois:
what is the purpose of this change?

Issue #2555, opened 2022-11-13
Customizable confidence scale


SamuelHassine:
## Use case

Customizable confidence scale.

Confidence = 0 to 100

Currently, the confidence labels "low / medium / high etc" are hardcoded. Be able to customize the confidence labels in the settings of the platform for intervals:

* 1 to 10: xxx
* 10 to 30: xxx

etc.

richard-julien:
Also take a look to https://github.com/OpenCTI-Platform/opencti/issues/2772

PR #3006, opened 2023-03-10
[back/front] Add Confidence Scale Configuration (#2555)

SarahBocognano:
### Proposed changes

* Add scalable confidence in entitySetting (attributes configuration)
* Change Item Confidence Field in StixDomainObjectOverview
* Change Confidence Field Globally (From a Select Box to an Input/Slider)
* Change default scale for the one specified in [Stix spec](https://docs.oasis-open.org/cti/stix/v2.1/os/stix-v2.1-os.html#_axjijf603msy)

Capture d’écran 2023-03-10 à 16 21 29

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/2555

### Checklist

### Further comments


opencti-platform/opencti-front/src/components/InputSliderField.tsx
```diff
@@ -0,0 +1,127 @@
...
+        />
+        <div style={{ paddingTop: '20px' }}>
+          <Slider
+            value={value}
+            min={min}
+            max={max}
+            marks={sliderMarks}
+            onChange={(_, v) => setFieldValue(name, v.toString())}
+            onChangeCommitted={(_, v) => onSubmit?.(name, v.toString())}
+            sx={{
```
RomuDeuxfois: Can we have this style in a make style ?
Or can we mutualize with line 109 ?

opencti-platform/opencti-front/src/components/ItemConfidence.tsx
```diff
@@ -0,0 +1,52 @@
...
+  confidence: number | null,
+  variant?: string,
+  entityType?: string,
+}
+
+const ItemConfidence: FunctionComponent<ItemConfidenceProps> = ({ confidence, variant, entityType }) => {
+  const { t } = useFormatter();
+  const classes = useStyles();
+
+  const { level: confidenceLevel } = useLevel(entityType ?? null, 'confidence', confidence);
```
RomuDeuxfois: Why we don't pass direclty entityType and handle the null or undefined case in use level method ?

opencti-platform/opencti-front/src/private/components/analysis/groupings/Grouping.js
```diff
@@ -74,6 +74,7 @@ export default createFragmentContainer(GroupingComponent, {
     fragment Grouping_grouping on Grouping {
       id
       standard_id
+      entity_type
```
RomuDeuxfois, line 77: Are you using this property ? It's maybe not necessary.

Same comment for other files

opencti-platform/opencti-front/src/private/components/analysis/groupings/GroupingEditionOverview.js
```diff
@@ -208,11 +208,9 @@ const GroupingEditionOverviewComponent = (props) => {
               }
             />
             <ConfidenceField
-              name="confidence"
               onFocus={editor.changeFocus}
-              onChange={handleSubmitField}
-              label={t('Confidence')}
-              fullWidth={true}
+              onSubmit={handleSubmitField}
+              entityType="Grouping"
```
RomuDeuxfois, line 213: We can use grouping.entity_type on this one.
If you add entity_type in the GroupingEditionOverview_grouping fragment (I missed this with the yup schema)

Same comment for other files

opencti-platform/opencti-front/src/private/components/analysis/groupings/GroupingEditionOverview.js
```diff
@@ -208,11 +208,9 @@ const GroupingEditionOverviewComponent = (props) => {
               }
             />
             <ConfidenceField
-              name="confidence"
               onFocus={editor.changeFocus}
-              onChange={handleSubmitField}
-              label={t('Confidence')}
-              fullWidth={true}
+              onSubmit={handleSubmitField}
```
RomuDeuxfois, line 212: Why we have onSubmit instead of onChange ?

opencti-platform/opencti-graphql/config/schema/opencti.graphql
```diff
@@ -3066,6 +3066,7 @@ enum OpinionsOrdering {
   createdBy
   objectMarking
   x_opencti_workflow_id
+  confidence
```
RomuDeuxfois, line 3069: We can order on confidence ?

opencti-platform/opencti-graphql/src/domain/attribute.ts
```diff
@@ -4,62 +4,92 @@ import { schemaAttributesDefinition } from '../schema/schema-attributes';
...
 import type { BasicStoreEntityEntitySetting } from '../modules/entitySetting/entitySetting-types';
 
-interface MandatoryAttribute {
+interface ScaleAttribute {
   name: string
-  builtIn: boolean
+  scale: string
+}
+
+interface AttributeConfigMeta {
```
RomuDeuxfois, line 17: Why we have the prefix Meta ?

opencti-platform/opencti-graphql/src/domain/attribute.ts
```diff
@@ -4,62 +4,92 @@ import { schemaAttributesDefinition } from '../schema/schema-attributes';
...
   });
 
+  // override with stored attributes configuration in entitySettings
   const userDefinedAttributes = getAttributesConfiguration(entitySetting);
   userDefinedAttributes?.forEach((userDefinedAttr) => {
-    const customizableAttr = customizableAttributes.find((a) => a.name === userDefinedAttr.name);
+    const customizableAttr = attributesConfiguration.find((a) => a.name === userDefinedAttr.name);
     if (customizableAttr) {
-      customizableAttr.mandatory = userDefinedAttr.mandatory;
+      if (customizableAttr.mandatoryType === 'customizable') {
```
RomuDeuxfois: We should add the check ' && userDefinedAttr.mandatory' like you did on line 77 to retrieve the mandaoty property only if it's exist on database

opencti-platform/opencti-front/src/utils/hooks/useScale.ts
```diff
@@ -0,0 +1,105 @@
...
+const useScale = (entityType: string | null, attributeName: string): ScaleConfig | null => {
+  if (!entityType) {
+    // return default configuration scale if entity type is not defined (ex: relationships)
+    return defaultScale;
+  }
+  const entitySetting = useEntitySettings(entityType)
+    .find((node) => node.scaleAttributes !== null);
+  const scaleAttribute = entitySetting?.scaleAttributes.find((a) => a.name === attributeName);
+  if (!scaleAttribute || !scaleAttribute.scale) {
+    return null;
```
RomuDeuxfois: It's maybe better to return defaultScale here
If we have (later) two scale attributes like confidence and likelihood but only confidence is configure, we can return null on likelihood and have a bug.

yassine-ouaamou:
Hey,
For the moment, here are some first remarks:
- [ ] Only in english, a capital letter is missing in the saving confirmation message: "confidence scale configuration has been successfully updated"
- [x] In spanish, same for "add a tick" => Añadir rotulador
- [ ] In my opinion, it is confusing that the "Add tick" text isn't a link
- [ ] The fact that we need to refresh the app in order to apply the changes is not optimal. If we can avoid this, it could be great ;)

opencti-platform/opencti-front/src/private/components/settings/sub_types/SubType.tsx
```diff
@@ -60,43 +64,50 @@ const SubType = ({ data }: { data: SubType_subType$key }) => {
   const classes = useStyles();
   const subType = useFragment(subTypeFragment, data);
   const statuses = (subType.statuses?.edges ?? []).map((edge) => edge.node);
-  const queryRef = useQueryLoading<EntitySettingQuery>(entitySettingQuery, {
-    targetType: subType.id,
-  });
+
+  const subTypeSettingsId = subType.settings?.id;
+  if (subTypeSettingsId) {
+    const config = useMemo<GraphQLSubscriptionConfig<SubTypeEntitySettingSubscription>>(
```
RomuDeuxfois, line 70: Seems not working on scale configuration

opencti-platform/opencti-front/src/private/components/settings/sub_types/EntitySettingScaleTickLine.tsx
```diff
@@ -0,0 +1,105 @@
...
+    alignItems: 'end',
+    height: '100%',
+  },
+}));
+
+interface EntitySettingScaleTickLineProps {
+  tick: Tick | UndefinedTick,
+  tickLabel: string,
+  deleteEnabled?: boolean,
+  validation: (t: (v: string) => string) => void,
```
RomuDeuxfois: Should return a shape

opencti-platform/opencti-front/src/private/components/settings/sub_types/EntitySettingScaleTickLine.tsx
```diff
@@ -0,0 +1,105 @@
...
+    display: 'flex',
+    justifyContent: 'flex-end',
+    alignItems: 'end',
+    height: '100%',
+  },
+}));
+
+interface EntitySettingScaleTickLineProps {
+  tick: Tick | UndefinedTick,
+  tickLabel: string,
```
RomuDeuxfois, line 24: We can't use the tick.label property ?

opencti-platform/opencti-front/src/private/components/settings/sub_types/EntitySettingScale.tsx
```diff
@@ -0,0 +1,221 @@
...
+            return !(!!value && (value < tickDefinition.min.value || value > tickDefinition.max.value));
+          }),
+      });
+  };
+
+  const minMaxValidation = () => {
+    return Yup.object()
+      .shape({
+        label: Yup.string().required(t('This field is required')),
+        color: Yup.string().required(t('This field is required')),
```
RomuDeuxfois: We can add the required pattern defined in back-end

opencti-platform/opencti-front/src/private/components/settings/sub_types/EntitySettingScale.tsx
```diff
@@ -0,0 +1,221 @@
...
+                }}>
+                <div className="rail">
+                  <span className="railSpan" style={{
+                    backgroundColor: `${tick.color ? tick.color : '#00b1ff'}`,
+                    height: '5px',
+                    width: '100%',
+                    display: 'block',
+                  }}></span>
+                </div>
+                <div className="valueAndLabel" style={{ display: 'flex', padding: '10px 0' }}>
```
RomuDeuxfois: This classname is not used

RomuDeuxfois:
Tests API :
- The available options are not manage when I get the attributesDefinitions property. I should not have the scale of the confidence property.
image
- What's the effects of changing better_side property ?
- I can have min equals to max

Tests front :
- I have a missing translation on confidence scale. And the Low/faible it's translated but it's a user configuration so.. I am not sure it's a good idea to translate it
image
- I can't fill the confidence field to null

opencti-platform/opencti-front/src/components/InputSliderField.tsx
```diff
@@ -0,0 +1,127 @@
...
+import { useLevel } from '../utils/hooks/useScale';
+import { Option } from '../private/components/common/form/ReferenceField';
+import { useFormatter } from './i18n';
+
+interface InputSliderFieldProps {
+  label: string;
+  variant?: string;
+  onSubmit?: (name: string, value: string | number | number[] | Option) => void;
+  onFocus?: (name: string, value: string) => void;
+  editContext?: unknown;
```
Kedae: :(

opencti-platform/opencti-front/src/components/InputSliderField.tsx
```diff
@@ -0,0 +1,127 @@
...
+import TextField from './TextField';
+import { SubscriptionFocus } from './Subscription';
+import { useLevel } from '../utils/hooks/useScale';
+import { Option } from '../private/components/common/form/ReferenceField';
+import { useFormatter } from './i18n';
+
+interface InputSliderFieldProps {
+  label: string;
+  variant?: string;
+  onSubmit?: (name: string, value: string | number | number[] | Option) => void;
```
Kedae: Don't think you can take option here. Why string ?

opencti-platform/opencti-front/src/components/InputSliderField.tsx
```diff
@@ -0,0 +1,127 @@
...
+}) => {
+  const { level: { color }, marks } = useLevel(entityType, attributeName, value);
+  const { t } = useFormatter();
+
+  const min = marks.length > 0 ? marks[0].value : 0;
+  const max = marks.length > 0 ? marks[marks.length - 1].value : 0;
+
+  const sliderMarks = marks.map((mark) => ({
+    value: mark.value,
+    label: mark.label ? t(mark.label) : mark.value,
```
Kedae: t(mark.label ?? mark.value)

opencti-platform/opencti-front/src/components/ItemConfidence.tsx
Kedae: I would add that we shouldn't have a null entityType here. ItemConfidence should only exists with an entityType

opencti-platform/opencti-front/src/private/components/analysis/groupings/GroupingEditionOverview.js
Kedae: Because we are in edition, we don't want to trigger a mutation each time we change the value but only once the user at the one he wants.

opencti-platform/opencti-front/src/private/components/cases/incidents/IncidentCreation.tsx
```diff
@@ -129,6 +129,7 @@ export const CaseCreationForm: FunctionComponent<IncidentFormProps> = ({ updater
       description: values.description,
       severity: values.severity,
       priority: values.priority,
+      confidence: parseInt(String(values.confidence), 10),
```
Kedae, line 132: What ?
You are parsing a stringify of the value ? Why not use directly the value ?

opencti-platform/opencti-front/src/private/components/common/form/ConfidenceField.tsx
```diff
@@ -0,0 +1,69 @@
...
+  },
+  message: {
+    width: '100%',
+    overflow: 'hidden',
+  },
+}));
+
+interface ConfidenceFieldProps {
+  variant?: string
+  onSubmit?: (name: string, value: string | number | number[] | Option) => void;
```
Kedae: No Option here

opencti-platform/opencti-front/src/private/components/common/form/ConfidenceField.tsx
```diff
@@ -0,0 +1,69 @@
...
+  onFocus?: (name: string, value: string) => void;
+  editContext?: unknown;
```
Kedae: Still :(

opencti-platform/opencti-front/src/private/components/events/incidents/IncidentEditionOverview.tsx
```diff
@@ -214,7 +214,7 @@ IncidentEditionOverviewProps
       },
     });
   };
-  const handleSubmitField = (name: string, value: string | string[] | null) => {
+  const handleSubmitField = (name: string, value: string | string[] | number | number[] | Option | null) => {
```
Kedae: Why Option here ?

opencti-platform/opencti-front/src/private/components/settings/sub_types/EntitySettingAttributesConfigurationScale.tsx
```diff
@@ -0,0 +1,146 @@
...
+const EntitySettingAttributesConfigurationScale = ({
+  entitySettingsData,
+}: {
+  entitySettingsData: SubType_subType$data['settings'];
+}) => {
+  const { t } = useFormatter();
+  const classes = useStyles();
+
+  const entitySetting = useFragment<EntitySetting_entitySetting$key>(entitySettingFragment, entitySettingsData);
+  if (!entitySetting) {
```
Kedae, line 65: How can we not have a entitySetting here ?

opencti-platform/opencti-front/src/private/components/settings/sub_types/EntitySettingAttributesConfigurationScale.tsx
```diff
@@ -0,0 +1,146 @@
...
+    ? JSON.parse(entitySetting.attributes_configuration)
+    : [];
+
+  const [commit] = useMutation(entitySettingsPatch);
+  const [submitting, setSubmitting] = useState(false);
+
+  const handleSubmitAttributeScale = (attributeName: string, scaleConfig: ScaleConfig) => {
+    setSubmitting(true);
+    const scale: Scale = { local_config: scaleConfig };
+    const saveConfiguration = [...attributesConfiguration];
```
Kedae, line 89: Why this const ?

opencti-platform/opencti-front/src/private/components/settings/sub_types/EntitySettingAttributesConfigurationScale.tsx
```diff
@@ -0,0 +1,146 @@
...
+  const getScaleConfig = (attributeScale: string) => {
+    const scale = JSON.parse(attributeScale) as Scale;
+    return scale.local_config;
+  };
+
+  const hasAttributeConfiguration = entitySetting.availableSettings.includes(
+    'attributes_configuration',
+  );
+
+  if (hasAttributeConfiguration) {
```
Kedae: Do this check at the top of the file like the other ones

opencti-platform/opencti-front/src/private/components/settings/sub_types/EntitySettingScale.tsx
```diff
@@ -0,0 +1,221 @@
...
+}
+
+const isTickDefinitionValid = (tickDefinition: ScaleConfig) => {
+  let isValid = true;
+  if (tickDefinition.min.value > tickDefinition.max.value) {
+    MESSAGING$.notifyError('The min value cannot be greater than max value');
+    return false;
+  }
+  tickDefinition.ticks.forEach((tick) => {
+    if (!!tick.value && (tick.value < tickDefinition.min.value || tick.value > tickDefinition.max.value)) {
```
Kedae, line 32: Not sure why a tick without a value is valid ?

opencti-platform/opencti-front/src/private/components/analysis/groupings/Grouping.js
SouadHadjiat: we need the entity_type property to know in which entity we are in all components that include ItemConfidence and ConfidenceField. 
See for example StixDomainObjectOverview :
```<ItemConfidence confidence={stixDomainObject.confidence} entityType={stixDomainObject.entity_type} />```

File: opencti-platform/opencti-front/src/private/components/analysis/groupings/GroupingEditionOverview.js (hunk 5 above)
Comment (SouadHadjiat) [reply opencti-platform/opencti-front/src/private/components/analysis/groupings/GroupingEditionOverview.js:212]: The input has changed from a selector to a text field input, and it's onSubmit method that is used on this input.

File: opencti-platform/opencti-front/src/private/components/settings/sub_types/EntitySettingScaleTickLine.tsx (hunk 12 above)
Comment (SouadHadjiat) [reply opencti-platform/opencti-front/src/private/components/settings/sub_types/EntitySettingScaleTickLine.tsx:24]: it's to be able to handle "min", "max", and tick label. The tick object can be a tick from the array of ticks, or "min" and "max" ticks, but we can't know that it's "min" or "max" in the tick object alone, that's why we added tickLabel as a property

File: opencti-platform/opencti-front/src/components/InputSliderField.tsx (hunk 17 above)
Comment (SouadHadjiat) [reply opencti-platform/opencti-front/src/components/InputSliderField.tsx]: you want to translate a number value ? there is only `mark.label` that can be translated.

File: opencti-platform/opencti-graphql/src/modules/entitySetting/entitySetting-utils.ts [hunk 26]
```diff
@@ -33,6 +33,26 @@ export const defaultEntitySetting: Record<string, typeAvailableSetting> = {
   attributes_configuration: JSON.stringify([]),
 };
 
+export const defaultScale = JSON.stringify({
```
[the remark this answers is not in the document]
SarahBocognano: We actually following the stix documentation (screenshot in the PR)

opencti-platform/opencti-front/src/private/components/settings/sub_types/EntitySettingScale.tsx
SouadHadjiat: If you add a tick, it doesn't have a value at first. 
If we remove the check, we get an error `'tick.value' is possibly 'undefined'`

opencti-platform/opencti-front/src/components/ItemConfidence.tsx
SouadHadjiat: There is currently one component where ItemConfidence doesn't get the entityType : `StixCoreRelationshipOverview.js` we might want to change it and set the entityType to `'stix-core-relationship'`

opencti-platform/opencti-graphql/config/schema/opencti.graphql
SouadHadjiat: yes, we added this to fix a current bug on production. If you try to order opinions on confidence, it will fail with an error :  
https://demo.opencti.io/dashboard/analysis/opinions?sortBy=created&orderAsc=false&searchTerm=&filters=%7B%22confidence_lte%22%3A%5B%7B%22id%22%3A%2250%22%2C%22value%22%3A%2250%22%7D%5D%7D&deSelectedElements=null

opencti-platform/opencti-graphql/src/domain/attribute.ts
SouadHadjiat: It's a name I choose for this object, because it's only used for the resolver and there is already an type named `AttributeConfiguration`

opencti-platform/opencti-front/src/components/InputSliderField.tsx
Kedae: No, I just want the code to be simplier. The formatter won't translate something if is not a given key so you can pass him anything you don't have to make conditionnal

opencti-platform/opencti-front/src/private/components/settings/sub_types/EntitySettingScale.tsx
Kedae: I hear you, yet I don't think we should consider an undefined tick value as a "valid" possibility. You may want to add another check before this condition

opencti-platform/opencti-graphql/config/schema/opencti.graphql
RomuDeuxfois: Yes I understand for the filter, but where it's possible to order ?

yassine-ouaamou:
As Romuald said, the tick labels are translated which is valid for the default values (Low, High, Out of range, ...) but should not be done for custom values.
Ex: if I put "Data source" as a tick label, it will be translated.

opencti-platform/opencti-front/src/private/components/settings/sub_types/EntitySettingScale.tsx
SarahBocognano: Done

opencti-platform/opencti-front/src/private/components/analysis/groupings/GroupingEditionOverview.js
SarahBocognano: Why would you change it ?

opencti-platform/opencti-front/src/components/InputSliderField.tsx
SarahBocognano: Done

opencti-platform/opencti-graphql/config/schema/opencti.graphql
SarahBocognano: You can order a confidence in Sightings for exemple :)

opencti-platform/opencti-graphql/src/domain/attribute.ts
SarahBocognano: Done

opencti-platform/opencti-front/src/private/components/settings/sub_types/EntitySettingScaleTickLine.tsx
SarahBocognano: What do you mean by that ?

opencti-platform/opencti-front/src/utils/hooks/useScale.ts
SarahBocognano: Done. But we wil probably not have a bug because we are filtering by attributeName

opencti-platform/opencti-front/src/private/components/settings/sub_types/EntitySettingScale.tsx
SarahBocognano: I don't understand, they are the same

opencti-platform/opencti-front/src/components/InputSliderField.tsx
SarahBocognano: Do you know the type of editContext ? Because it's seems that it's unknown everywhere in the platform...

opencti-platform/opencti-front/src/components/InputSliderField.tsx
SouadHadjiat: I think it's already a simple code `label: mark.label ? t(mark.label) : mark.value,` Why should it be simplified more, and pass a value that we know won't work with the translator ? But there is a question about translating the mark labels, so maybe we should answer it first.

opencti-platform/opencti-front/src/components/InputSliderField.tsx
SarahBocognano: We removed the Option, but we need the string because the Slider component send a number and for the mutation to append, we need to pass string

opencti-platform/opencti-front/src/private/components/cases/incidents/IncidentCreation.tsx
SarahBocognano: Because it's written like that everywhere in the platform

opencti-platform/opencti-front/src/private/components/common/form/ConfidenceField.tsx
SarahBocognano: Done

opencti-platform/opencti-front/src/private/components/common/form/ConfidenceField.tsx
SarahBocognano: Same answer ... :(

opencti-platform/opencti-front/src/private/components/events/incidents/IncidentEditionOverview.tsx
SarahBocognano: Done

opencti-platform/opencti-front/src/components/InputSliderField.tsx
Kedae: Then it's a mistake everywhere in the platform. The type should be defined by relay and you should be able to use it.

SouadHadjiat: Maybe we could do it in another PR ? it's not in the scope of this one, which is already huge.

SouadHadjiat: We might have found the correct type, we are testing

SarahBocognano: Ok it's done :)

opencti-platform/opencti-front/src/private/components/common/form/ConfidenceField.tsx
SarahBocognano: Done

opencti-platform/opencti-front/src/private/components/settings/sub_types/EntitySettingAttributesConfigurationScale.tsx
SarahBocognano: It's just for security. Do you want to remove it ?

opencti-platform/opencti-front/src/private/components/settings/sub_types/EntitySettingAttributesConfigurationScale.tsx
SouadHadjiat: it's exactly the same method as in `EntitySettingAttributesConfiguration`, we wrote it together with @RomuDeuxfois

opencti-platform/opencti-front/src/private/components/settings/sub_types/EntitySettingAttributesConfigurationScale.tsx
SarahBocognano: Done

opencti-platform/opencti-front/src/private/components/settings/sub_types/EntitySettingScale.tsx
SouadHadjiat: I don't see how I could add another check before the condition, if I remove `!!tick.value`, the rest of my condition doesn't work. Do you have a suggestion that works ?

SarahBocognano:
> 

I think it's important to have the Default Values translated, but for the custom ones, the probability to put a word that is already translated in the platform is closed to 0... We can't know in advance if it's a custom or a default label.

opencti-platform/opencti-front/src/private/components/settings/sub_types/EntitySettingScale.tsx
SarahBocognano: > Tests API :
> 
> * The available options are not manage when I get the attributesDefinitions property. I should not have the scale of the confidence property.
>   image

---> You ask for the scale in your query. If you don't ask for it, you won't have the scale

> * What's the effects of changing better_side property ?
---> It's not used for now but was requested for an eventual V2

> * I can have min equals to max
---> The min cannot be greater than max but that's the only condition

> 
> Tests front :
> 
> * I have a missing translation on confidence scale. And the Low/faible it's translated but it's a user configuration so.. I am not sure it's a good idea to translate it
>   image

--> same answer as for Yassine. We need the default label to be translated, but we can't translate a custom one because we cannot know in advance what it will be. The probability to be the same as words translated in the platform are very very low.

> * I can't fill the confidence field to null

--> Yes it's not possible for now, it will be in a V2 (we plan to open issues after the merge)

SarahBocognano:
Tests API :

The available options are not manage when I get the attributesDefinitions property. I should not have the scale of the confidence property.
image
---> You ask for the scale in your query. If you don't ask for it, you won't have the scale

What's the effects of changing better_side property ?
---> It's not used for now but was requested for an eventual V2
I can have min equals to max
---> The min cannot be greater than max but that's the only condition
Tests front :

I have a missing translation on confidence scale. And the Low/faible it's translated but it's a user configuration so.. I am not sure it's a good idea to translate it
image
--> same answer as for Yassine. We need the default label to be translated, but we can't translate a custom one because we cannot know in advance what it will be. The probability to be the same as words translated in the platform are very very low.

I can't fill the confidence field to null
--> Yes it's not possible for now, it will be in a V2 (we plan to open issues after the merge)

SarahBocognano:
Tests API :

The available options are not manage when I get the attributesDefinitions property. I should not have the scale of the confidence property.
image
---> You ask for the scale in your query. If you don't ask for it, you won't have the scale

What's the effects of changing better_side property ?
---> It's not used for now but was requested for an eventual V2
I can have min equals to max
---> The min cannot be greater than max but that's the only condition
Tests front :

I have a missing translation on confidence scale. And the Low/faible it's translated but it's a user configuration so.. I am not sure it's a good idea to translate it
image
--> same answer as for Yassine. We need the default label to be translated, but we can't translate a custom one because we cannot know in advance what it will be. The probability to be the same as words translated in the platform are very very low.

I can't fill the confidence field to null
--> Yes it's not possible for now, it will be in a V2 (we plan to open issues after the merge)

opencti-platform/opencti-front/src/private/components/settings/sub_types/SubType.tsx
SouadHadjiat: Yes, we saw this together, and I think it's because we use a state in EntitySettingScale. I don't know how I could reset the state when scaleConfig changes. Is there a method to allow us to know when there is a change ? I'm searching about it.

opencti-platform/opencti-front/src/private/components/analysis/groupings/GroupingEditionOverview.js
RomuDeuxfois: Just to use a property of the object instead of hard code.
for me, it makes the code more dynamic

opencti-platform/opencti-graphql/config/schema/opencti.graphql
RomuDeuxfois: Oh, nice catch ! I missed this one.
So, if you add it here we need to add it in the UI.
If not, it's not necessary to add it on graphQL

opencti-platform/opencti-front/src/utils/hooks/useScale.ts
RomuDeuxfois: And you are right !

opencti-platform/opencti-front/src/private/components/settings/sub_types/SubType.tsx
RomuDeuxfois: Yes I remember, I was hoping that you would find a solution :D
But your explanation makes sense.. I don't have any idea for now, so we can let it like this (it's not critical at all).

opencti-platform/opencti-front/src/private/components/settings/sub_types/EntitySettingScaleTickLine.tsx
RomuDeuxfois: Thank you for the clear answer

opencti-platform/opencti-front/src/private/components/settings/sub_types/EntitySettingScaleTickLine.tsx
RomuDeuxfois: Sorry, I did not explain myself.
The validation method in Formik use a shape in input. So we need to return a shape when we build the validation method to be compliant with the formik contract.

Like in the hook useYupSchemaBuilder, we return a BaseSchema.

opencti-platform/opencti-front/src/private/components/settings/sub_types/EntitySettingScale.tsx
RomuDeuxfois: Yes, but if we add it here, the Yup validation can throw an exception here and we don't have to call the back-end to have it.
This removes an extra call to the server.

opencti-platform/opencti-front/src/private/components/settings/sub_types/EntitySettingAttributesConfigurationScale.tsx
RomuDeuxfois: It's just to be more readable.
Maybe we can simplify this by working direclty with attributesConfiguration.
But, it's more clear like this for me.

RomuDeuxfois:
@SarahBocognano 
Thank you for your answer. I can't wait for the V2 !

RomuDeuxfois:
Design details : the trash icon are not align with the labels and bottom buttons
image

opencti-platform/opencti-front/src/private/components/settings/sub_types/SubType.tsx
SouadHadjiat: I investigated and didn't find a solution for now

opencti-platform/opencti-front/src/private/components/settings/sub_types/EntitySettingScale.tsx
SouadHadjiat: Ah you mean the regex validation ! ok we can add it here

opencti-platform/opencti-front/src/private/components/analysis/groupings/GroupingEditionOverview.js
SouadHadjiat: I don't really see the benefit of adding `entity_type` in all the fragments, knowing that we already are in an entity type specific component. The `OpenVocabField` has already this kind of behavior with `type` and `name` parameters which have hardcoded values.

opencti-platform/opencti-front/src/private/components/settings/sub_types/EntitySettingScaleTickLine.tsx
SarahBocognano: Done :)

opencti-platform/opencti-front/src/private/components/settings/sub_types/EntitySettingScale.tsx
SarahBocognano: Done

SarahBocognano:
> Design details : the trash icon are not align with the labels and bottom buttons image

The trash icon is actually align but only when hover

opencti-platform/opencti-front/src/components/InputSliderField.tsx
```diff
@@ -0,0 +1,118 @@
...
+  entityType,
+  attributeName,
+  disabled,
+}) => {
+  const { level: { color }, marks } = useLevel(entityType, attributeName, value);
+
+  const min = marks.length > 0 ? marks[0].value : 0;
+  const max = marks.length > 0 ? marks[marks.length - 1].value : 0;
+
+  const sliderMarks = marks.map((mark) => ({
```
SouadHadjiat: maybe we don't need anymore this block and can directly use `marks` instead of `sliderMarks` (since we removed the label translation) @SarahBocognano

SarahBocognano: Done :)

Kedae (approved):

PR #3032, opened 2023-03-21
[back/front] Settings routes protection (#issue/2985)

SarahBocognano:
### Proposed changes

* Regulate settings access problems
* Manage display access menu for markings (if SET_MARKINGS) and redirect to marking
* Realign backend security access with frontend

### Checklist

### Further comments

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/nav/TopMenuSettings.js
+++ b/opencti-platform/opencti-front/src/private/components/nav/TopMenuSettings.js
@@ -5,7 +5,10 @@ import Button from '@mui/material/Button';
 import makeStyles from '@mui/styles/makeStyles';
 import { useFormatter } from '../../../components/i18n';
 import Security from '../../../utils/Security';
-import { SETTINGS_SETACCESSES, SETTINGS_SETLABELS } from '../../../utils/hooks/useGranted';
+import {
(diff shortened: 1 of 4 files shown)


Kedae (approved):

opencti-platform/opencti-front/src/private/components/nav/TopMenuSettings.js
```diff
@@ -35,7 +39,7 @@ const TopMenuSettings = () => {
                 classes={{ root: classes.button }}>
                 {t('Parameters')}
             </Button>
-            <Security needs={[SETTINGS_SETACCESSES]}>
+            <Security needs={[SETTINGS_SETMARKINGS]}>
```
RomuDeuxfois, line 42: I don't understand this security.
If I don't have the SETTINGS_SETMARKINGS, so the button is not display, but the others routes inside has not always the security SETTINGS_SETMARKINGS (exemple: dashboard/settings/accesses/roles)

So I don't be bale to go to roles if I don't have the security for marking ?

Kedae: Right, we missed some stuff and it's not working that way. I'll correct it

Issue #3033, opened 2023-03-21
Can't update end date of an event


Archidoit:
## Description

Can't update end date of an event

## Reproducible Steps

Go to an event. Edit the field 'end date'. The modification is not saved.

PR #3043, opened 2023-03-22
[Front] event end_date edition bug fix (#3033)

Archidoit:
Can't update end date of an event.

Not: idem for Last seen of a Malware

Reproductible steps:

Go to an event. Edit the field 'end date'. The modification is not saved.

### Related issues

https://github.com/OpenCTI-Platform/opencti/issues/3033

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/entities/events/EventEditionOverview.js
+++ b/opencti-platform/opencti-front/src/private/components/entities/events/EventEditionOverview.js
@@ -82,8 +82,9 @@ const EventEditionOverviewComponent = (props) => {
     event_types: Yup.array().nullable(),
     start_time: Yup.date().typeError(t('The value must be a datetime (yyyy-MM-dd hh:mm (a|p)m)')).nullable(),
     stop_time: Yup.date()
+      .typeError(t('The value must be a datetime (yyyy-MM-dd hh:mm (a|p)m)'))
       .min(Yup.ref('start_time'), "The end date can't be before start date")
-      .typeError(t('The value must be a datetime (yyyy-MM-dd hh:mm (a|p)m)')).nullable(),
+      .nullable(),
     references: Yup.array(),
     x_opencti_workflow_id: Yup.object(),
   };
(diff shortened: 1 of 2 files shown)


opencti-platform/opencti-front/src/private/components/entities/events/EventEditionOverview.js
```diff
@@ -134,17 +135,12 @@ const EventEditionOverviewComponent = (props) => {
...
-        .then(() => {
-          editor.fieldPatch({
-            variables: {
-              id: event.id,
-              input: { key: name, value: finalValue ?? '' },
-            },
-          });
-        })
-        .catch(() => false);
+      editor.fieldPatch({
```
RomuDeuxfois, line 138: It's not a good solution to remove the validation here because all the others property will not be validated.
We need to re-work the stop_time validator.

RomuDeuxfois: I find that
`    stop_time: Yup.date()
      .when('start_time', (startTime, field) => {
        return startTime ? field.min(Yup.ref('start_time'), "The end date can't be before start date") : field;
      })
      .typeError(t('The value must be a datetime (yyyy-MM-dd hh:mm (a|p)m)'))
      .nullable(),`
      
Seems to working as expected

RomuDeuxfois: If it's ok for you, we need to propagate this changes on MalwareEditionDetails

Archidoit: the properties seems validated anyway because the validator was present in the Formik...
but i'm going to try your solution and keep it if it works :)

Archidoit: --> fist solution kept because not redondant

RomuDeuxfois (approved):

PR #3089, opened 2023-03-29
[frontend] Add case container management in Workbench (#3088)

### LINKED ISSUE #3088: Add case container management in Workbench
## Use case

Add case container management in Workbench

richard-julien:


Kedae (commented):
Doesn't seem to be working ? I've tried with an export from Demo. The Workbench correctly shows the case but no creation is done when approved.

richard-julien:
> Doesn't seem to be working ? I've tried with an export from Demo. The Workbench correctly shows the case but no creation is done when approved.

Lets talk about it

Kedae (approved):

PR #3096, opened 2023-03-29
[backend] Enforce merging to update elements base on their current index (#3095)

### LINKED ISSUE #3095: Enforce merging to update elements base on their current index
## Use case

Merging currently rely on _index that is inferred from the element type instead of the real element index.
This situation can lead to error if the index of the element to merge is not the latest one (rollover)
We need to get index of the elements to merge through a internal finds. That will impact a bit the performance of the merge but its mandatory to have a merge independant of the indices.

richard-julien:


opencti-platform/opencti-graphql/src/database/middleware.js
```diff
@@ -1218,11 +1218,13 @@ const loadMergeEntitiesDependencies = async (context, user, entityIds) => {
   for (let entityIndex = 0; entityIndex < entityIds.length; entityIndex += 1) {
     const entityId = entityIds[entityIndex];
     // Internal From
-    const listFromCallback = (elements) => {
+    const listFromCallback = async (elements) => {
+      const findArgs = { toMap: true, baseData: true };
+      const relTargets = await internalFindByIds(context, user, elements.map((rel) => rel.toId), findArgs);
```
Kedae, line 1223: Do you need to be vigilant about perf inside the mergeEntities or not ? Because adding queries inside it might make it slower ?

yassine-ouaamou (approved):

RomuDeuxfois (approved):

PR #3099, opened 2023-03-29
[backend] Improve engine to handle large amount of text to search (#3098)

### LINKED ISSUE #3098: Improve engine to handle large amount of text to search
## Use case

When using a large text to find information in the platform, the engine generate a too large amount of nested terms. 
Engine must be improved to use a more appropriate approach

richard-julien:


Archidoit:
When doing a Search with the following text (50 words), I have an error (no error when doing the same Search in Demo).

“It is used by cybercriminals to conduct ransomware attacks against multiple sectors and organisations worldwide, including Australia. Once gaining access to a victim’s environment, cybercriminals use this ransomware for similar purposes as other variants such as encrypting their data, and extorting a ransom to return access to the sensitive files.”

image

--> Notice: no error for texts with 29 words or less. Error for texts with more than 29 words.

Kedae (approved):

PR #3105, opened 2023-03-31
[backend/frontend]  User overview modification (#2849)

### LINKED ISSUE #2849: User overview modification
## Use case

In an overview of a user:

1. Fix the user operations history chart which does not work (line is always 0)
2. Change the layout of the bottom with a 50/50 grid:
  a. Left = Operations chart
  b. Right = History

Archidoit:
### Related issue
https://github.com/OpenCTI-Platform/opencti/issues/2849

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/settings/users/User.js
+++ b/opencti-platform/opencti-front/src/private/components/settings/users/User.js
@@ -460,7 +460,7 @@ class UserComponent extends Component {
           container={true}
           spacing={3}
           classes={{ container: classes.gridContainer }}
-          style={{ marginTop: 30 }}
+          style={{ marginTop: 35 }}
         >
           <Grid item={true} xs={6} style={{ marginBottom: 25 }}>
             <Typography variant="h4" gutterBottom={true}>
--- a/opencti-platform/opencti-front/src/private/components/settings/users/User.js
+++ b/opencti-platform/opencti-front/src/private/components/settings/users/User.js
@@ -460,9 +460,9 @@ class UserComponent extends Component {
           container={true}
           spacing={3}
           classes={{ container: classes.gridContainer }}
-          style={{ marginTop: 20 }}
+          style={{ marginTop: 30 }}
         >
-          <Grid item={true} xs={12}>
+          <Grid item={true} xs={6} style={{ marginBottom: 25 }}>
             <Typography variant="h4" gutterBottom={true}>
               {t('Operations')}
             </Typography>
@@ -513,8 +513,10 @@ class UserComponent extends Component {
               />
             </Paper>
           </Grid>
+          <Grid item={true} xs={6} style={{ paddingTop: 8 }}>
+            <UserHistory userId={user.id} />
+          </Grid>
         </Grid>
-        <UserHistory userId={user.id} />
         <Fab
           onClick={this.handleOpenUpdate.bind(this)}
           color="secondary"
--- a/opencti-platform/opencti-front/src/private/components/settings/users/UserHistory.js
+++ b/opencti-platform/opencti-front/src/private/components/settings/users/UserHistory.js
@@ -2,20 +2,10 @@ import React, { Component } from 'react';
 import * as PropTypes from 'prop-types';
 import Typography from '@mui/material/Typography';
 import { compose } from 'ramda';
-import withStyles from '@mui/styles/withStyles';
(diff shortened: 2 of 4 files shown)


opencti-platform/opencti-graphql/src/database/middleware.js
```diff
@@ -498,7 +499,9 @@ const convertAggregateDistributions = async (context, user, limit, orderingFunct
 export const timeSeriesEntities = async (context, user, types, args) => {
   const timeSeriesArgs = buildEntityFilters({ types, ...args });
   const { startDate, endDate, interval } = args;
-  const histogramData = await elHistogramCount(context, user, args.onlyInferred ? READ_DATA_INDICES_INFERRED : READ_DATA_INDICES, timeSeriesArgs);
+  const dataIndices = args.onlyInferred ? READ_DATA_INDICES_INFERRED : READ_DATA_INDICES;
+  const indexName = types.includes(ENTITY_TYPE_HISTORY) ? INDEX_HISTORY : dataIndices;
```
Kedae: Shouldn't the check be that the types only contains History ? I'm not sur to understand the choice that has been made of getting only History (after reading the brainstorm). 

Additionnaly, why INDEX_HISTORY and not READ_INDEX_HISTORY ?

I thought we could do something like types.includes(HISTORY) ? dataIndices : READ_PLATFORM_INDICES ?

Archidoit: right for READ_INDEX_HISTORY instead of INDEX_HISTORY

not sure to understand the rest, we can have a quick talk

Archidoit: understood, fixed

opencti-platform/opencti-front/src/private/components/settings/users/User.js
```diff
@@ -460,9 +460,9 @@ class UserComponent extends Component {
           container={true}
           spacing={3}
           classes={{ container: classes.gridContainer }}
-          style={{ marginTop: 20 }}
+          style={{ marginTop: 30 }}
         >
-          <Grid item={true} xs={12}>
+          <Grid item={true} xs={6} style={{ marginBottom: 25 }}>
```
RomuDeuxfois, line 465: Why do we have this margin bottom?

opencti-platform/opencti-front/src/private/components/settings/users/User.js
```diff
@@ -513,8 +513,10 @@ class UserComponent extends Component {
               />
             </Paper>
           </Grid>
+          <Grid item={true} xs={6} style={{ paddingTop: 8 }}>
```
RomuDeuxfois, line 516: Proposition: Isn't it more readable with more space ?

paddingTop: 40 px on the left grid

image

opencti-platform/opencti-front/src/private/components/settings/users/UserHistory.js
```diff
@@ -31,29 +21,29 @@ class UserHistory extends Component {
     const { t, userId } = this.props;
     const { entitySearchTerm } = this.state;
     return (
-      <div style={{ marginTop: 50 }}>
+      <div>
         <Typography
           variant="h4"
           gutterBottom={true}
-          style={{ float: 'left', marginTop: 12 }}
+          style={{ float: 'left', margin: '16px 0 0 0' }}
```
RomuDeuxfois, line 28: you can use marginTop directly

opencti-platform/opencti-front/src/private/components/settings/users/User.js
Archidoit: when passing History and Operations in the same line, there was a difference at the bottom (Operations didn't respect the margin bottom)

opencti-platform/opencti-front/src/private/components/settings/users/UserHistory.js
Archidoit: no I can't, if I do this, there is a gap between History title and the History box (see screenshot below)

image

opencti-platform/opencti-front/src/private/components/settings/users/User.js
Archidoit: you're right! maybe not that much, but I will increase it :)

RomuDeuxfois (approved):

PR #3134, opened 2023-04-11
[frontend/backend] Fix on settings page (#2512bis)

Kedae:


Archidoit:
Bugs of RBAC refacto fixed:
- Settings hidden types access forbidden without SETTINGS_SETACCESS rights
- Entity Setting overview 'hidden in roles' display forbidden wihtout SETTINGS_SETACCESS rights

PR #3137, opened 2023-04-11
[front] Fix on "Top Labels" on smaller screens (#3136)

### LINKED ISSUE #3136: "Top Labels" text overflows on small screens
## Description

On smaller screen sizes, the text of "Top Labels" is too large and the resulting output is visually unpleasant and can be difficult to decipher.

## Environment

The default environment currently running on the [demo](https://demo.opencti.io/dashboard).

## Reproducible Steps

1. Open the demo link
2. Log in
3. Shrink your window to below 900px wide
4. Observe the "Top Labels" box

## Expected Output

Text that's easily read and visually appealing:

Expected Output

## Actual Output

This is what it currently looks like:

Actual Output

ajmeese7:
I haven't had the opportunity to get connectors working in my local setup, so I wasn't able to import data to test this out. If someone could proof-check this for me to make sure the media query applies as expected, I would appreciate it!

closes: #3136

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ To help you get you familiar with our contribution process, we have a list of [b
 
 ### Development
 
-If you want to actively help OpenCTI, we created a [dedicated documentation](https://filigran.notion.site/Environment-setup-606996f36d904fcf8d434c6d0eae4a00) about the deployment of a development environement and how to start the source code modification.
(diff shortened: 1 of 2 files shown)


SamuelHassine:
Hello @ajmeese7,

Thanks a lot for your contribution! We will handle this problem but we don't want to use a CSS approach here. Someone in the team will open a PR in the current milestone.

Have a great day!

Kind regards,
Samuel

Issue #3116, opened 2023-04-03
Group default membership and auto marking must be initialized to allow correct sorting/filtering


richard-julien:
## Use case

When groups are sorted/filtered by default membership or auto new marking, sometimes the groups are not listed. This is due to the fact that this 2 booleans are not initialized to false by default.

2 Things to do:
- Change creation method to initalize the 2 values.
- Add a migration to initialize the missing boolean to false if needed

Must be done after https://github.com/OpenCTI-Platform/opencti/issues/2512

PR #3155, opened 2023-04-14
[frontend/backend] Group default membership and auto marking must be initialized (#3116)

Archidoit:
## Related issue
https://github.com/OpenCTI-Platform/opencti/issues/3116

## Notion page
https://www.notion.so/filigran/Group-default-membership-and-auto-marking-must-be-initialized-to-allow-correct-sorting-filtering-3-9e54df103a2448ffa0dd95166fbbeda2

### DIFF
--- a/opencti-platform/opencti-graphql/src/domain/group.js
+++ b/opencti-platform/opencti-graphql/src/domain/group.js
@@ -54,19 +54,11 @@ export const batchRoles = async (context, user, groupIds) => {
 };
 
 export const addGroup = async (context, user, group) => {
-  let groupWithDefaultValues = group;
-  if (groupWithDefaultValues.default_assignation === undefined) {
-    groupWithDefaultValues = {
-      ...groupWithDefaultValues,
-      default_assignation: false,
-    };
-  }
-  if (groupWithDefaultValues.auto_new_marking === undefined) {
-    groupWithDefaultValues = {
-      ...groupWithDefaultValues,
-      auto_new_marking: false,
-    };
-  }
+  const groupWithDefaultValues = {
+    ...group,
+    default_assignation: group.default_assignation ?? false,
+    auto_new_marking: group.auto_new_marking ?? false
+  };
   const created = await createEntity(context, user, groupWithDefaultValues, ENTITY_TYPE_GROUP);
   return notify(BUS_TOPICS[ENTITY_TYPE_GROUP].ADDED_TOPIC, created, user);
 };
--- a/opencti-platform/opencti-graphql/src/domain/group.js
+++ b/opencti-platform/opencti-graphql/src/domain/group.js
@@ -54,7 +54,20 @@ export const batchRoles = async (context, user, groupIds) => {
 };
 
 export const addGroup = async (context, user, group) => {
-  const created = await createEntity(context, user, group, ENTITY_TYPE_GROUP);
+  let groupWithDefaultValues = group;
+  if (groupWithDefaultValues.default_assignation === undefined) {
+    groupWithDefaultValues = {
+      ...groupWithDefaultValues,
+      default_assignation: false,
+    };
+  }
+  if (groupWithDefaultValues.auto_new_marking === undefined) {
+    groupWithDefaultValues = {
+      ...groupWithDefaultValues,
+      auto_new_marking: false,
+    };
+  }
+  const created = await createEntity(context, user, groupWithDefaultValues, ENTITY_TYPE_GROUP);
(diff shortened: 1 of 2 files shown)


opencti-platform/opencti-front/src/private/components/settings/groups/GroupCreation.js
```diff
@@ -98,10 +98,15 @@ class GroupCreation extends Component {
   }
 
   onSubmit(values, { setSubmitting, resetForm }) {
+    const finalValues = {
```
RomuDeuxfois: Must be done in back-end too, in case of group creation by API

opencti-platform/opencti-graphql/src/migrations/1681460290302-group-initialization.js
```diff
@@ -0,0 +1,31 @@
+import { executionContext, SYSTEM_USER } from '../utils/access';
+import { logApp } from '../config/conf';
+import { findAll, groupEditField } from '../domain/group';
+
+export const up = async (next) => {
+  const context = executionContext('migration');
+  const groupsEdges = await findAll(context, SYSTEM_USER);
+  const groups = groupsEdges.edges.map((n) => n.node);
+  const groupEditionPromises = groups.map((group) => {
+    if (group.auto_new_marking === undefined) {
```
RomuDeuxfois: It could be simplified

Const updateFields = [];
if (group.auto_new_marking === undefined) {
   { key: 'auto_new_marking', value: ['false'] }
}
if (group.default_assignation === undefined) {   
   { key: 'default_assignation', value: ['false'] }
}

if (updateFields.length > 0) {
   return groupEditField(context, SYSTEM_USER, group.id, updateFields);
}

What do you think ?

Archidoit: you're right! it's better this way!

opencti-platform/opencti-graphql/src/domain/group.js
```diff
@@ -54,7 +54,12 @@ export const batchRoles = async (context, user, groupIds) => {
 };
 
 export const addGroup = async (context, user, group) => {
-  const created = await createEntity(context, user, group, ENTITY_TYPE_GROUP);
+  const groupWithDefaultValues = {
+    ...group,
+    default_assignation: false,
```
RomuDeuxfois: ⚠️ In this order, the default_assignation and auto_new_marking property will be overriden if they exist in the group object.

Archidoit: owww right !!

opencti-platform/opencti-graphql/src/migrations/1681825498443-group-initialization.js
```diff
@@ -0,0 +1,26 @@
+import { executionContext, SYSTEM_USER } from '../utils/access';
+import { logApp } from '../config/conf';
+import { findAll, groupEditField } from '../domain/group';
+
+export const up = async (next) => {
+  const context = executionContext('migration');
+  const groupsEdges = await findAll(context, SYSTEM_USER);
```
Kedae: You should fetch only the groups that matches your need (with undefined values) and not with findAll cause you might miss some items.

Archidoit: I try to use listAllEntities with a filter, but I can't set the filtering value to undefined...
image

opencti-platform/opencti-graphql/src/domain/group.js
```diff
@@ -54,7 +54,20 @@ export const batchRoles = async (context, user, groupIds) => {
...
+      ...groupWithDefaultValues,
+      default_assignation: false,
+    };
+  }
+  if (groupWithDefaultValues.auto_new_marking === undefined) {
+    groupWithDefaultValues = {
+      ...groupWithDefaultValues,
+      auto_new_marking: false,
+    };
+  }
```
SouadHadjiat: I think it could be simplified, maybe like this :
```
const groupWithDefaultValues = {
    ...group,
    default_assignation: group.default_assignation ?? false,
    auto_new_marking: group.auto_new_marking ?? false,
  };
```
What do you think @Archidoit ?

Archidoit: yes you're right!! I'll modify this!

Archidoit: --> done by @richard-julien

PR #3200, opened 2023-04-18
[backend] Implementation of user roles batch loading (#3196)

### LINKED ISSUE #3196: List of users is broken in the backend because not enriched
## Use case

The list of users is broken in the backend:

```
query {
  users {
    edges {
      node {
        roles {
          id
        }
      }
    }
  }
}
```

```
          "stacktrace": [
            "Error: Cannot return null for non-nullable field User.roles.",
            "    at completeValue 
```

Most of the "mandatory" attributes of a user are filled by the `completeUser` function. But the list is directly an Elastic pagination query...

richard-julien:
[X] User role batch loading
[X] User allowed_marking to be removed
[X] Check roleSessionRefresh that seems broken
[X] Refactor migration to care about all roles and users + reapply it

### DIFF
--- a/opencti-platform/opencti-graphql/src/database/middleware-loader.ts
+++ b/opencti-platform/opencti-graphql/src/database/middleware-loader.ts
@@ -57,6 +57,7 @@ interface EntityFilters<T extends BasicStoreCommon> extends ListFilter<T> {
 }
 
 export interface EntityOptions<T extends BasicStoreCommon> extends EntityFilters<T> {
+  ids?: Array<string>;
   indices?: Array<string>;
 }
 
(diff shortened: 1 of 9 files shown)


opencti-platform/opencti-graphql/src/database/middleware-loader.ts
```diff
@@ -57,6 +57,7 @@ interface EntityFilters<T extends BasicStoreCommon> extends ListFilter<T> {
 }
 
 export interface EntityOptions<T extends BasicStoreCommon> extends EntityFilters<T> {
+  ids?: Array<string>;
```
SouadHadjiat, line 60: how does `ids` filter work ? Doesn't it need to be handled in `buildEntityFilters` ?

SouadHadjiat (approved):
I tested the API, but not the migration (when I started the app locally, it ran the new migration but there wasn't any user->role relationship left already)

PR #3228, opened 2023-04-24
[backend] Default Group for Connector at initialization (#3224)

### LINKED ISSUE #3224: Create a default Group for Connector at initialization
## Use case

We have  a "Connector" role initialized at creation.

At initialization, we should create a "Connectors" (with s) group with:

* Role connector
* Auto new marking
* All default markings associated

RomuDeuxfois:
### Related issues

* [issues 3224](https://github.com/OpenCTI-Platform/opencti/issues/3224)

### DIFF
--- a/opencti-platform/opencti-graphql/src/initialization.js
+++ b/opencti-platform/opencti-graphql/src/initialization.js
@@ -263,7 +263,7 @@ export const createBasicRolesAndCapabilities = async (context) => {
     description: 'Administrator role that bypass every capabilities',
     capabilities: [BYPASS],
   });
-  await addRole(context, SYSTEM_USER, {
+  const connectorRole = await addRole(context, SYSTEM_USER, {
     name: 'Connector',
     description: 'Connector role that has the recommended capabilities',
     capabilities: [
@@ -291,6 +291,17 @@ export const createBasicRolesAndCapabilities = async (context) => {
     relationship_type: 'has-role',
   };
   await groupAddRelation(context, SYSTEM_USER, defaultGroup.id, defaultRoleRelationInput);
+  // Create connector group with connector role
+  const connectorGroup = await addGroup(context, SYSTEM_USER, {
+    name: 'Connectors',
+    description: 'Connector group',
+    auto_new_marking: true,
+  });
+  const connectorRoleRelationInput = {
+    toId: connectorRole.id,
+    relationship_type: 'has-role',
+  };
+  await groupAddRelation(context, SYSTEM_USER, connectorGroup.id, connectorRoleRelationInput);
 };
 
 const createVocabularies = async (context) => {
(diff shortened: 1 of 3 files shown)


opencti-platform/opencti-graphql/src/initialization.js
```diff
@@ -291,6 +291,27 @@ export const createBasicRolesAndCapabilities = async (context) => {
...
+  await groupAddRelation(context, SYSTEM_USER, connectorGroup.id, connectorRoleRelationInput);
+  // Markings
+  const markings = await findAll(context, SYSTEM_USER, { connectionFormat: false });
+  for (let i = 0; i < markings.length; i += 1) {
+    const marking = markings[i];
+    const connectorMarkingRelationInput = {
+      toId: marking.id,
+      relationship_type: 'accesses-to',
+    };
+    await groupAddRelation(context, SYSTEM_USER, connectorGroup.id, connectorMarkingRelationInput);
```
SouadHadjiat: I don't understand how this works, why do we need to add this relationship here since it's already done in `addMarkingDefinition` ?

opencti-platform/opencti-graphql/src/initialization.js
```diff
@@ -320,11 +341,11 @@ const initializeDefaultValues = async (context, withMarkings = true) => {
     platform_language: 'auto',
   });
   await createDefaultStatusTemplates(context);
-  await createBasicRolesAndCapabilities(context);
-  await createVocabularies(context);
   if (withMarkings) {
     await createMarkingDefinitions(context);
   }
+  await createBasicRolesAndCapabilities(context);
```
SouadHadjiat: If we create groups after markings, how does the relationships added between groups and markings will work in `addMarkingDefinition` ?

opencti-platform/opencti-graphql/tests/02-integration/01-database/elasticSearch-test.js
```diff
@@ -505,7 +505,7 @@ describe('Elasticsearch pagination', () => {
     let data = await elPaginate(testContext, ADMIN_USER, READ_RELATIONSHIPS_INDICES);
     expect(data).not.toBeNull();
     const groupByIndices = R.groupBy((e) => e.node._index, data.edges);
-    expect(groupByIndices[`${ES_INDEX_PREFIX}_internal_relationships-000001`].length).toEqual(25);
+    expect(groupByIndices[`${ES_INDEX_PREFIX}_internal_relationships-000001`].length).toEqual(33);
```
SouadHadjiat, line 507: what are the relationships added ? I can count one "has-role" between the new connectors group and connector role, then five "accesses-to" between the connectors group and the five default markings, which would be +6, what are the last 2 that I didn't count ?

RomuDeuxfois: There are three new marking-definition in the DATA-TEST-STIX2_v2 file
: 
- TLP:WHITE
- TLP:TEST
- Copyright 2017, The MITRE Corporation

But the TLP:WHITE is resolved by the same id as TLP:CLEAR in identifier.js file

So, we have two new markings ingested and two new relations

SouadHadjiat (approved):

PR #3266, opened 2023-05-02
[backend/frontend] Add policies for local password (#3262)

### LINKED ISSUE #3262: Add policies for local password
## Use case

Nothing prevent today any user to setup a password with only one char.
Password policy should be possible to specificy a complexity.
- min length
- max length
- number of min symbols
- number of min numbers
- number of words
- number of lowercase
- number of uppercase

richard-julien:


opencti-platform/opencti-front/src/private/components/common/form/PasswordPolicies.tsx
```diff
@@ -0,0 +1,60 @@
...
+    password_policy_min_words
+    password_policy_min_lowercase
+    password_policy_min_uppercase
+  }
+`;
+
+const PasswordPolicies: FunctionComponent<{ style?: object }> = ({ style }) => {
+  const classes = useStyles();
+  const { t } = useFormatter();
+  const settings = useContext(UserContext).settings as unknown as PasswordPolicies_policies$key;
```
Kedae: I thought I already discussed that with someone but apparently no 😱
Could you provide a useSettings hook (or use the useAuth one) as it is also used in LocationMiniMap & LocationMiniMapTarget

opencti-platform/opencti-front/src/private/components/common/form/PasswordPolicies.tsx
```diff
@@ -0,0 +1,60 @@
...
+  const policies = useFragment(passwordPoliciesFragment, settings);
```
Kedae: Not sure why you get policies here, then destructure it, then use it directly line 49. Could you aligne everything with one usage ? 

const { 
password_... , 
password_... ,
password_... ,
} = useFragment(); for example

opencti-platform/opencti-front/src/private/components/profile/ProfileOverview.js
```diff
@@ -435,16 +430,15 @@ const ProfileOverviewComponent = (props) => {
           )}
         </div>
         <div className="clearfix" />
-        <Formik
+        { external === false && <Formik
```
Kedae: !external ? Or an undefined/null external has a different meaning ?

opencti-platform/opencti-front/src/private/components/settings/AccessSettings.tsx
```diff
@@ -0,0 +1,207 @@
...
+  password_policy_min_length: Yup.number(),
+  password_policy_min_symbols: Yup.number(),
+  password_policy_min_numbers: Yup.number(),
+  password_policy_min_words: Yup.number(),
+  password_policy_min_lowercase: Yup.number(),
+  password_policy_min_uppercase: Yup.number(),
+});
+
+const AccessSettings: FunctionComponent<AccessSettingsComponentProps> = ({ queryRef }) => {
+  const { settings } = usePreloadedQuery<AccessSettingsQuery>(accessSettingsQuery, queryRef);
```
Kedae: Why not using context ?

opencti-platform/opencti-front/src/private/components/settings/AccessSettings.tsx
```diff
@@ -0,0 +1,207 @@
...
+  const classes = useStyles();
+  const { t } = useFormatter();
+  const handleSubmitField = (name: string, value: unknown) => {
+    settingsValidation().validateAt(name, { [name]: value }).then(() => {
+      commitMutation({
```
Kedae: Better to use useMutation ?

opencti-platform/opencti-front/src/private/components/settings/AccessSettings.tsx
```diff
@@ -0,0 +1,207 @@
...
+    password_policy_min_numbers: settings.password_policy_min_numbers,
+    password_policy_min_words: settings.password_policy_min_words,
+    password_policy_min_lowercase: settings.password_policy_min_lowercase,
+    password_policy_min_uppercase: settings.password_policy_min_uppercase,
+  };
+  return <div className={classes.container}>
+    <AccessesMenu />
+    <Grid container={true} spacing={3}>
+      <Grid item={true} xs={12}>
+          <Formik onSubmit={() => {}} enableReinitialize={true} initialValues={initialValues} validationSchema={settingsValidation()}>
```
Kedae: ={true} not needed

opencti-platform/opencti-front/src/private/components/settings/AccessSettings.tsx
```diff
@@ -0,0 +1,207 @@
...
+                  </div>
+                  <div style={{ marginTop: 20 }}>
+                    <Typography variant="h4" gutterBottom={true}>
+                      {t('Local password policies')}
+                    </Typography>
+                    <Paper classes={{ root: classes.paper }} variant="outlined">
+                      <Field component={TextField} type="number" variant="standard"
+                          name="password_policy_min_length" label={t('Number of chars must be greater or equals to')} fullWidth={true}
+                          onSubmit={(name: string, value: string) => {
+                            return handleSubmitField(name, value !== '' ? value : '0');
```
Kedae, line 138: Wouldn't it be better to handle the value stuff in the handleSubmitField because it is applied everywhere, and then the onSubmit would be onSubmit={handleSubmitField}

opencti-platform/opencti-front/src/private/components/settings/AccessSettings.tsx
```diff
@@ -0,0 +1,207 @@
...
+                  </div>
+                </Form>
+            )}
+          </Formik>
+      </Grid>
+    </Grid>
+  </div>;
+};
+
+const Root = () => {
```
Kedae: Usually we use a XXXComponent and XXX naming, with export default XXX;

opencti-platform/opencti-front/src/private/components/settings/Root.js
```diff
@@ -77,6 +78,15 @@ const Root = () => (
             </Security>
           )}
         />
+        <BoundaryRoute
+          exact
+          path="/dashboard/settings/accesses/settings"
+          render={() => (
+            <Security needs={[SETTINGS_SETACCESSES]} placeholder={<Redirect to={'/dashboard/settings'} />}>
```
Kedae, line 85: In the graphql we are behing SETTINGS, is that normal ?

opencti-platform/opencti-front/src/private/components/common/form/PasswordPolicies.tsx
richard-julien: Ok

opencti-platform/opencti-front/src/private/components/common/form/PasswordPolicies.tsx
richard-julien: Ok

opencti-platform/opencti-front/src/private/components/profile/ProfileOverview.js
richard-julien: Ok

opencti-platform/opencti-front/src/private/components/settings/AccessSettings.tsx
richard-julien: ok

opencti-platform/opencti-front/src/private/components/settings/Root.js
richard-julien: For now we have only 1 api to patch platform settings and have some explicit checking for access in resolvers. Should be refactor later to split the API

opencti-platform/opencti-front/src/private/components/settings/AccessSettings.tsx
richard-julien: To discuss about this approach that prevent to have some strange conversion in the handleSubmitField function

opencti-platform/opencti-front/src/private/components/settings/AccessSettings.tsx
richard-julien: ok

opencti-platform/opencti-front/src/private/components/settings/AccessSettings.tsx
richard-julien: Ok

Kedae (approved):

Issue #3290, opened 2023-05-05
Can't update is family field on malware


RomuDeuxfois:
## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Go to a malware
2. Update is family field to true
3. The modification is present
4. Update is family field to false
5. The modification is not present

PR #3297, opened 2023-05-09
[frontend] Can't update is family field on malware (#3290)

Archidoit:
https://github.com/OpenCTI-Platform/opencti/issues/3290


SarahBocognano (commented):
-

SarahBocognano (approved):

PR #3321, opened 2023-05-11
[frontend/backend] Notifications for relationships involving restricted entities (#3318)

### LINKED ISSUE #3318: Notifications for relationships involving restricted entities
## Description

Some notifications concerning relationships involving a restricted entity are displayed with the name of the restricted entity for users with restricted rights.

## Environment

4. Other environment details:

## Reproducible Steps

- Log with a user Y with no access to RED marking.
- Create a trigger on event Update/Create/Delete,  with outcome = UI, and without filters.
- Log in a private window as admin.
- Create a relationship: from an entity with marking=red, to X.
- Go back to the platform logged with user Y.
There is a notification "creation of a relationship from 'name' to X.
The name of the source entity is displayed whereas the user has no access to this entity.
If we click on the notification line: Error (instead of the overview of the relationship 'from restricted to X').

## Expected Output

Write 'Restricted entity' instead of the restricted entity name.
If we click on the notification line: redirect to the overview of the relationship 'from restricted to X'.

## Screenshots (optional)
- notification line:
image

- whereas the source entity is restricted:
image

Archidoit:
### PR content
- For notifications on relationships/sightings, write 'Restricted' instead of a restricted element name
- Handle redirection when clicking on a notification line of a relationship/sighting involving restricted element(s)

### Github issue
https://github.com/OpenCTI-Platform/opencti/issues/3318

### Notion page
https://www.notion.so/filigran/Notifications-for-relationships-involving-restricted-entities-3318-f72abfe191cc4735ab4d65058ec8fe56


opencti-platform/opencti-front/src/private/components/events/stix_sighting_relationships/StixSightingRelationshipOverview.js
```diff
@@ -281,44 +298,55 @@ class StixSightingRelationshipContainer extends Component {
...
                     top: 20,
                     right: 20,
                   }}
                 >
                   <div
                     className={classes.itemHeader}
                     style={{
-                      borderBottom: `1px solid ${itemColor(to.entity_type)}`,
+                      borderBottom: `1px solid ${itemColor(
+                        !toRestricted ? to.entity_type : 'Unknown',
```
SouadHadjiat: why 'Unknown' here instead of 'Restricted' ?

Archidoit: @SouadHadjiat  it's not a displayed name, it's the color applied in the platform for such restricted entities (see StixCoreRelationshipOverview).
The color is Restricted for relationships lines, and Unknown in Relationships overview... --> I'm going to uniformize this

opencti-platform/opencti-graphql/src/manager/publisherManager.ts
```diff
@@ -31,6 +94,212 @@ const OUTCOME_TYPE_UI = 'UI';
...
+const extractStixRepresentative = async (context: AuthContext, user: NotificationUser, stix: StixObject): Promise<string> => {
+  const entityType = stix.extensions[STIX_EXT_OCTI].type;
+  // region Modules
+  const convertFn = getStixRepresentativeConverters(entityType);
+  if (convertFn) {
+    return convertFn(stix);
+  }
+  // endregion
+  // region Sighting
+  if (isStixSightingRelationship(entityType)) {
```
RomuDeuxfois: Why did you move this method? It deals with stix, I think it makes sense to have it in the stix file

RomuDeuxfois (approved):

opencti-platform/opencti-graphql/src/manager/publisherManager.ts
Archidoit: I put it in an other file because of dependencies cycles error in imports.

Archidoit: Indeed we had this cycle : middleware.js > engine.js > identifier.js > stix-converter.js > stixCoreObject.js > middleware.js (since I added the relation stix-converter > stixCoreObject by using the findById method in stix-converter)

PR #3326, opened 2023-05-12
[frontend] Fix on Workbench and Vocabulary usage

### LINKED ISSUE #3311: Observable handling in Workbench
- In the workbench we are missing the observable_value handling (in format.js in the back) -> User Account observable is missing the default value.

- The vocab attributes are not handled at all, we need to extends the Entity.ts file to handle those -> ⁉️This will probably mean that we have to get all related files into pure function

Additionnal knowledge: this tasks is **deeply** linked to some work of @RomuDeuxfois, you need to sync with him regarding default values.
https://github.com/OpenCTI-Platform/opencti/issues/3211

Kedae:


marieflorescontact:
After test:

- Default value is not unknown any more 👌
- Is it normal to create an observable via Workbench without any value? 🤔

RomuDeuxfois:
Need to discuss about the next step for useAttribute file

SarahBocognano (approved):

PR #3333, opened 2023-05-16
[frontend] Loadquery fix

Kedae:


opencti-platform/opencti-front/src/utils/hooks/useEntityTypes.ts
SarahBocognano, line 1: Why did you remove this file ? Just curious

opencti-platform/opencti-front/src/private/Root.tsx
```diff
@@ -55,16 +59,25 @@ const Root = () => {
...
-        <ConnectedThemeProvider settings={settings}>
-          <ConnectedIntlProvider settings={settings}>
-            <Index />
-          </ConnectedIntlProvider>
-        </ConnectedThemeProvider>
-      </StyledEngineProvider>
-    </UserContext.Provider>
+    <React.Suspense fallback={<Loader variant={LoaderVariant.inElement} />}>
+      {(vocabularyCategoriesQueryRef) && (
+        <QueryContext.Provider value={{ vocabularyCategoriesQueryRef }}>
```
SarahBocognano, line 68: What is the use of that ?

PR #3338, opened 2023-05-16
[backend] lazy-loading spec_version when fetching stix Refs

### LINKED ISSUE #3330: Export broken for Data components
## Environment

2. OpenCTI version: 5.7
3. OpenCTI client: frontend
4. Other environment details:

## Reproducible Steps

Export a Data component containing a reference to a data source (json export)

## Expected Output

Be able to download the generated export file

## Actual Output

```
{"timestamp": "2023-05-15T12:37:12.760685Z", "level": "ERROR", "name": "pycti.api", "message": "Cannot return null for non-nullable field StixRefRelationship.spec_version."}
{"timestamp": "2023-05-15T12:37:13.652370Z", "level": "ERROR", "name": "root", "message": "Error in message processing, reporting error to API", "exc_info": "Traceback (most recent call last):\n  File \"C:\\Users\\yassi\\IdeaProjects\\client-python\\pycti\\connector\\opencti_connector_helper.py\", line 167, in _process_message\n    json_data[\"internal\"][\"work_id\"], message_task.result()\n  File \"C:\\Users\\yassi\\scoop\\apps\\python310\\current\\lib\\concurrent\\futures\\thread.py\", line 58, in run\n    result = self.fn(*self.args, **self.kwargs)\n  File \"C:\\Users\\yassi\\IdeaProjects\\connectors\\internal-export-file\\export-file-stix\\src\\export-file-stix.py\", line 103, in _process_message\n    bundle = self.helper.api_impersonate.stix2.export_list(\n  File \"C:\\Users\\yassi\\IdeaProjects\\client-python\\pycti\\utils\\opencti_stix2.py\", line 2111, in export_list\n    entity_bundle = self.prepare_export(\n  File \"C:\\Users\\yassi\\IdeaProjects\\client-python\\pycti\\utils\\opencti_stix2.py\", line 1611, in prepare_export\n    stix_nested_ref_relationships = self.opencti.stix_nested_ref_relationship.list(\n  File \"C:\\Users\\yassi\\IdeaProjects\\client-python\\pycti\\entities\\opencti_stix_nested_ref_relationship.py\", line 132, in list\n    result = self.opencti.query(\n  File \"C:\\Users\\yassi\\IdeaProjects\\client-python\\pycti\\api\\opencti_api_client.py\", line 348, in query\n    raise ValueError(\nValueError: {'name': 'Cannot return null for non-nullable field StixRefRelationship.spec_version.', 'message': 'Cannot return null for non-nullable field StixRefRelationship.spec_version.'}"
```
## Additional information

Stix requires that entites set the spec version field in relations which is not the case for Data components

yassine-ouaamou:
### Proposed changes

* Add a resolver for the spec_version attribute  in the Stix Ref Relationship API
* Return a default value ("2.1") for spec_version if it is not set

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/3330
*

### Checklist

Test cases will be included the issue of malware analysis and will consist of tests the resolver of stixRefRelationships

### Further comments

### DIFF
--- a/opencti-platform/opencti-graphql/src/domain/stixSightingRelationship.js
+++ b/opencti-platform/opencti-graphql/src/domain/stixSightingRelationship.js
@@ -46,9 +46,8 @@ export const findById = (context, user, stixSightingRelationshipId) => {
   return storeLoadById(context, user, stixSightingRelationshipId, STIX_SIGHTING_RELATIONSHIP);
 };
 
-export function lazyLoadSpecVersion({ spec_version }) {
-  return spec_version ?? STIX_SPEC_VERSION;
-}
+export const lazyLoadSpecVersion = ({ spec_version }) => spec_version ?? STIX_SPEC_VERSION;
+
 export const stixSightingRelationshipsNumber = (context, user, args) => ({
   count: elCount(context, user, READ_INDEX_STIX_SIGHTING_RELATIONSHIPS, assoc('types', [STIX_SIGHTING_RELATIONSHIP], args)),
   total: elCount(
--- a/opencti-platform/opencti-graphql/src/domain/stixSightingRelationship.js
+++ b/opencti-platform/opencti-graphql/src/domain/stixSightingRelationship.js
@@ -36,6 +36,7 @@ import {
   stixObjectOrRelationshipAddRefRelation,
   stixObjectOrRelationshipDeleteRelation
 } from './stixObjectOrStixRelationship';
+import { STIX_SPEC_VERSION } from '../database/stix';
 
 export const findAll = async (context, user, args) => {
   return listRelations(context, user, STIX_SIGHTING_RELATIONSHIP, args);
@@ -45,6 +46,9 @@ export const findById = (context, user, stixSightingRelationshipId) => {
   return storeLoadById(context, user, stixSightingRelationshipId, STIX_SIGHTING_RELATIONSHIP);
 };
 
+export function lazyLoadSpecVersion({ spec_version }) {
+  return spec_version ?? STIX_SPEC_VERSION;
+}
 export const stixSightingRelationshipsNumber = (context, user, args) => ({
   count: elCount(context, user, READ_INDEX_STIX_SIGHTING_RELATIONSHIPS, assoc('types', [STIX_SIGHTING_RELATIONSHIP], args)),
   total: elCount(
--- a/opencti-platform/opencti-graphql/src/domain/stixSightingRelationship.js
+++ b/opencti-platform/opencti-graphql/src/domain/stixSightingRelationship.js
@@ -36,7 +36,6 @@ import {
   stixObjectOrRelationshipAddRefRelation,
   stixObjectOrRelationshipDeleteRelation
 } from './stixObjectOrStixRelationship';
-import { STIX_SPEC_VERSION } from '../database/stix';
 
 export const findAll = async (context, user, args) => {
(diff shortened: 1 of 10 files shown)


opencti-platform/opencti-graphql/src/resolvers/stixRefRelationship.js
```diff
@@ -44,7 +45,8 @@ const stixRefRelationshipResolvers = {
     opinions: (rel, _, context) => opinionsLoader.load(rel.id, context, context.user),
     creators: (rel, _, context) => creatorsLoader.load(rel.creator_id, context, context.user),
     editContext: (rel) => fetchEditContext(rel.id),
-    datable: (rel) => isDatable(rel.fromType, rel.relationship_type)
+    datable: (rel) => isDatable(rel.fromType, rel.relationship_type),
+    spec_version: (rel) => { return rel.spec_version ?? STIX_SPEC_VERSION; }
```
Kedae: spec_version: ({ spec_version }) => spec_version ??  STIX_SPEC_VERSION

RomuDeuxfois: Should we extend this behavior to SDO and SRO?

SarahBocognano:
Tested 👍🏼

opencti-platform/opencti-graphql/src/resolvers/stixRefRelationship.js
yassine-ouaamou: We agreed to extend it to SRO (stix core and stix sighting)

opencti-platform/opencti-graphql/src/domain/stixSightingRelationship.js
```diff
@@ -45,6 +46,9 @@ export const findById = (context, user, stixSightingRelationshipId) => {
   return storeLoadById(context, user, stixSightingRelationshipId, STIX_SIGHTING_RELATIONSHIP);
 };
 
+export function lazyLoadSpecVersion({ spec_version }) {
```
Kedae: This is not a lazyLoad function, a defaultValue one maybe.
And this is definitely not how we define function in our code. 

Finally I'm not sure there is any gain from defining that rather than just inlining it the way you used to do it

tomibennett: > This is not a lazyLoad function, a defaultValue one maybe.

I agree with you because it better describes the function internal behavior. However *lazyLoad* is the term proposed and used all along in our discussion **to describe the behavior in the system** (i.e. get rid off `spec_version` in stored relations objects and only set it for graphql responses because some our clients need it, starting with `pycti` and probably externals too), hence the *lazy* in the code. 

Do you consider this term misleading? Totally fine to change the name if it makes more sense to you.

>  And this is definitely not how we define function in our code.

It will be changed.

> Finally I'm not sure there is any gain from defining that rather than just inlining it the way you used to do it

I agree it makes little sense for a oneline method, however, I consider it a good practice as it adds lisiblity. With the function definition, I can focus on the *what* is done instead of *how* it is done. The intent is explicit and I can have a better high-level understanding of what is going on, instead of having to make sense of all the details.

Kedae: As discussed today : It seems that spec_version is missing from StixRelationship type. We need to check with @richard-julien.

In that case we need to update only the resolver/domain of stixRelationship and rename the method.

richard-julien (approved):

PR #3368, opened 2023-05-22
[frontend] fix broken Delete button in Analyst Workbench interface (#3363)

### LINKED ISSUE #3363: Broken Delete button in Analyst Workbench interface
## Description

The "inline" Delete Button to delete an entity, an observable, a container or a relationship in the analyst Workbench interface is broken. Nothing happen when it is clicked, even if the related checkbox is checked. 
image

To note: the Delete button in the toolbar is working fine.

## Environment

4. Other environment details:

## Reproducible Steps

1. { e.g. Run ... }
2. { e.g. Click ... }
3. { e.g. Error ... }

SouadHadjiat:
### Proposed changes

* fix call to function onClick (the argument was the event object)
* minor fix for dialog open attribute which should be a boolean (errors in console)

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/3363

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/common/files/workbench/WorkbenchFileContent.js
+++ b/opencti-platform/opencti-front/src/private/components/common/files/workbench/WorkbenchFileContent.js
@@ -3686,7 +3686,7 @@ const WorkbenchFileContentComponent = ({
       {currentTab === 2 && renderRelationships()}
       {currentTab === 3 && renderContainers()}
       <Dialog
-        open={deleteObject}
+        open={!!deleteObject}
         PaperProps={{ elevation: 1 }}
         keepMounted
         TransitionComponent={Transition}
@@ -3703,7 +3703,7 @@ const WorkbenchFileContentComponent = ({
           </Button>
           <Button
             color="secondary"
-            onClick={submitDeleteObject}
+            onClick={() => submitDeleteObject()}
           >
             {t('Remove')}
           </Button>


opencti-platform/opencti-front/src/private/components/common/files/workbench/WorkbenchFileContent.js
```diff
@@ -3686,7 +3686,7 @@ const WorkbenchFileContentComponent = ({
       {currentTab === 2 && renderRelationships()}
       {currentTab === 3 && renderContainers()}
       <Dialog
-        open={deleteObject}
+        open={!!deleteObject}
```
Kedae, line 3689: Why is that necessary ? It seems that deleteobject is either null, undefined or an object ?

SouadHadjiat: `open` parameter should be a boolean, I had warnings in the console

Kedae: Oh, I see. In that case we might want to enforce a rule between this usage and the Boolean(deleteObject) one ? What do you like the most ?

Archidoit:
Behavior tested --> OK

marieflorescontact (approved):

opencti-platform/opencti-front/src/private/components/common/files/workbench/WorkbenchFileContent.js
SouadHadjiat: a rule ? I don't understand

Issue #2540, opened 2022-11-05
Be able to add images in overview of some entity types


SamuelHassine:
This feature should be designed as generic in the backend (any types of entity can have images) but in the frontend (overview / cards), only those first types of entities are concerned:

- Threat Actor Individual
- Threat Actor Group
- Intrusion Set
- Malware
- Tool
- Individual
- Organization

**In Data tab :**

- [ ] **`MVP`** Adding a “Pictures” section
    - each uploaded picture (through the “uploaded files” section) is represented by a line in the “Pictures management” section with:
        - a miniature of the Picture
        - the description
        - order
    - the “in Caroussel” checkbox
    - a burger button to manage or delete the picture
    - At the top of the “Pictures management” section are 2 Combo boxes:
        - One to choose which picture will be displayed in the Card view
        - `Nice-to-have` One to choose the caroussel mode : Static=the caroussel does not rotate automatically, Dynamic= the caroussel rotates automatically
    - By default, the list is sorted by Order
- [ ] **`MVP`**Managing a Picture:
    - when clicking on the burger menu of a Picture, it opens an “Manage the Picture” form:
        - field description
        - field Order (with an Information : “Order for the caroussel display)
        - a toggle “in Caroussel” to define if the Picture is in the caroussel or not

!https://share.balsamiq.com/c/cBxJq4zLwKm42bizBjYFin.png

**In Overview:**

- [ ] **`MVP`** The caroussel contain all picture with the “in Caroussel” toggled on.
- [ ] **`MVP`** The caroussel default position is set to display the Picture with the lowest Order value
- [ ] **`MVP`** When hovering a picture in the caroussel, you get the description in a tooltip
- [ ] `Nice-to-have` If no picture has been added, a placeholder is displayed in the caroussel
- [ ] `Nice-to-have` With the caroussel, 3 buttons:
    - remove the displayed picture from caroussel (toggle off the “in Overview”)
    - modify the displayed picture: open the edit form for the picture
    - add a picture to the Overview

!https://share.balsamiq.com/c/bmdgTsZ2yJY8HL754gaPSq.png

**In Card view:**

- [ ] **`MVP`** Picture chosen to be “Picture for Card” in the “Data/Picture management” section is displayed in the card
    - ⚠️ It must be able to handle portait or landscapes pictures.

!https://share.balsamiq.com/c/2vC5sq1t8fEv9jx6ghh1pU.png

Issue #2349, opened 2022-09-13
Adding Image to entity profile


AAA10CR7:
I would like to request a new feature for different entities in OpenCTI where we can upload image that represent logo or flag of the entity .

For example, enable user to

1) upload photo of the individuals (actual photo)
2) add logo of the organization (Microsoft, Google ...etc)
3) upload Threat actor/intrusionSet image (IRGC, GRU, APT34)
4) flag image of the country (US, Italy, UK ...)

Also, it would be great if we can substitute current entity logo (repetitive for same type) with these images in different graphs ( investigation, knowledge)

SamuelHassine:
Duplicate of https://github.com/OpenCTI-Platform/opencti/issues/2540

PR #3371, opened 2023-05-22
[frontend/backend] Added Photo Carousel to Threat Actor Details page

ParamConstructor:
### Proposed changes

Aligns with part of the open Issues -  #2540 and #2349. 

Added the ability to support a Photo Carousel on the Threat Actor Details page. A user can upload and attach multiple photos (filtered based on size and types) in the code. Excerpt:
````
    const SUPPORTED_FORMATS = ['image/bmp', 'image/gif', 'image/jpg', 'image/jpeg', 'image/png'];
    const FILE_SIZE = 10000000;
```` 
The photo is uploaded using a new page modal and allows for the selection of the photo and a comment to go with the photo. The photo and comment are stored using the existing OpenCTI FileUploader (opencti-front/src/private/components/common/files/FileUploader.tsx) and OpenCTI File Object with a new field added of "comment"

````
type File {
  id: ID!
  name: String!
  size: Int
  lastModified: DateTime
  lastModifiedSinceMin: Int
  metaData: FileMetadata
  uploadStatus: State!
  works: [Work]
  comment: String
}
````

The photo/file is added to the "data" tab of the Threat Actor. It is automatically linked to the photo carousel using a new field added to the ThreatActor (x_opencti_photo_refs: [String]). Deleting the photo is done by deleting the file in the "Data Tab". If you want/need to update the comment, the user can re-upload the photo again and it will replace the comment.
 
NOTE: If no photo is assigned - a default of (Man_Silhouette_clip_art_large.png) is displayed - You may have a different photo you would like to replace this with, as the licensing of this photo is unknown.

### Related issues

There is a Dependency cycle with these two files and imports within them. Some rework is required to decompose these functions/imports and it was felt to be outside of this PR to resolve completely.
````
opencti-platform/opencti-graphql/src/database/file-storage.js
  18:1  error  Dependency cycle detected  import/no-cycle
````

### Checklist

- [X] I consider the submitted work as finished (for the initial intent - further comments will elaborate)
- [X] I wrote test cases for the relevant uses case
     -  ThreatActorDetails.js converted to a TSX file.

### Further comments

- The way this was implemented should support adoption and application onto other request object types from the open issues #2540 and  #2349.
- Currently, ordering of the photos is based on file date/add date from the Data tab. It may be requested to be able to re-order these based on a number method. However, ordering of them would require additional storage keys or could be based on the array order within x_opencti_photo_refs.
- Currently, there is no designated "primary" photo or way to flag one if you have multiple photos.


ParamConstructor:
@SamuelHassine / @Kedae - This PR can be rejected / closed - it will be rebaselined and resubmitted at a later date.

PR #3383, opened 2023-05-24
[backend/frontend] Add a new capability to restrict the access to users list (#3151)

### LINKED ISSUE #3151: Add a new capability to restrict the access to users list (assignees)
## Use case

Add a new capability to restrict the access to users list (assignees).

Problem: on a shared platform with read only access, everyone can see the list of all users / customers in Reports => Filters => Assignees.

SouadHadjiat:
### Proposed changes

* Add a new capability "Assign users" under "Knowledge update"
* Use aggregations for creators and assignee filters : filters will only return actual creators and assignees for the entity type searched.
* Retrict creators / assignees / members APIs : creators and members endpoints will return the full list only for "knowledge update" capability and assignees is restricted with the newly added "knowledge update assign" capability

#### To test the issue : 
- log with user that doesn't have bypass capability
- In a list of entities (for example reports), in the autocomplete filters "assignee" and "technical creator" you should have only actual assignees and creators of the searched entities (instead of all the users of the platform)
- When you select an entity (like a report), and try to edit it, you should not be able to see "assign to" field if you don't have the new capability "Assign users"

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/3151

### Checklist


opencti-platform/opencti-front/src/private/components/common/form/ObjectAssigneeField.jsx
```diff
@@ -101,34 +103,36 @@ class ObjectAssigneeField extends Component {
...
-              <div className={classes.icon}>
-                <ItemIcon type={option.type} />
-              </div>
-              <div className={classes.text}>{option.label}</div>
-            </li>
-          )}
-          classes={{ clearIndicator: classes.autoCompleteIndicator }}
-        />
-      </div>
+        <Security needs={[KNOWLEDGE_KNUPDATE_KNASSIGN]}>
```
marieflorescontact: Maybe  you could migrate the component into pure function? 😇

opencti-platform/opencti-graphql/src/database/engine.js
```diff
@@ -857,13 +857,10 @@ export const elFindByIds = async (context, user, ids, opts = {}) => {
...
-      const id = workingIds[i];
-      for (let indexType = 0; indexType < elementTypes.length; indexType += 1) {
-        const elementType = elementTypes[indexType];
-        const term = { [`${elementType}.keyword`]: id };
-        idsTermsPerType.push({ term });
-      }
+    for (let indexType = 0; indexType < elementTypes.length; indexType += 1) {
+      const elementType = elementTypes[indexType];
+      const terms = { [`${elementType}.keyword`]: workingIds };
+      idsTermsPerType.push({ terms });
```
SarahBocognano, line 863: Why did you rework that part ?

opencti-platform/opencti-graphql/src/database/engine.js
```diff
@@ -1507,6 +1501,64 @@ export const elAggregationRelationsCount = async (context, user, indexName, opti
       throw DatabaseError('[SEARCH] Fail processing AggregationRelationsCount', { error: e });
     });
 };
+
+export const elAggregationsList = async (context, user, indexName, aggregations = [], options = {}) => {
```
SarahBocognano: Just curious about what this method doing exactly

SarahBocognano (commented):
Tested ✅

opencti-platform/opencti-graphql/src/database/engine.js
SouadHadjiat: yes it's a good question, I had to use this method to resolve entities by ids (ids that I get from the aggregation query) and I saw that we used "term" query with multiple values instead of a "terms" query.

opencti-platform/opencti-graphql/src/migrations/1685090556916-add-assign-users-capability.js
```diff
@@ -0,0 +1,19 @@
+import { executionContext, SYSTEM_USER } from '../utils/access';
+import { logApp } from '../config/conf';
+import { addCapability } from '../domain/grant';
+
+export const up = async (next) => {
```
RomuDeuxfois: Now, we neeed to have the capability KNOWLEDGE_KNUPDATE_KNASSIGN to see some fields
Should we add this capa for all existing users ? To have the same behaviors as before this changes.

opencti-platform/opencti-front/src/private/components/events/incidents/IncidentsCards.tsx
```diff
@@ -88,6 +88,7 @@ const IncidentsCards: FunctionComponent<IncidentsCardsProps> = ({
     linesQuery: incidentsCardsAndLinesPaginationQuery,
     linesFragment: IncidentsCardsAndLinesFragment,
     queryRef,
+    nodePath: ['incidents', 'pageInfo', 'globalCount'],
```
RomuDeuxfois, line 91: Does that fix a bug ?

opencti-platform/opencti-front/src/utils/filters/useSearchEntities.tsx
```diff
@@ -243,6 +243,9 @@ const useSearchEntities = ({
   }));
   const searchEntities = (
     filterKey: string,
+    searchContext: { entityType: string },
+    cacheEntities: Record<string, { label: string, value: string, type: string }[]>,
```
RomuDeuxfois, line 277: Good idea !

opencti-platform/opencti-graphql/src/database/engine.js
```diff
@@ -1507,6 +1501,64 @@ export const elAggregationRelationsCount = async (context, user, indexName, opti
...
+      body,
+    };
+    const searchType = `Aggregations (${aggregations.map((agg) => agg.field)?.join(', ')})`;
+    const data = await elRawSearch(context, user, searchType, query).catch((err) => {
+      throw DatabaseError('[SEARCH] Aggregations list fail', { error: err, query });
+    });
+    const aggsMap = Object.keys(data.aggregations);
+    const aggsValues = R.uniq(R.flatten(aggsMap.map((agg) => data.aggregations[agg].buckets?.map((b) => b.key))));
+    const aggsElements = resolveByIds
+      ? await elFindByIds(context, user, aggsValues, { baseData: true, baseFields: ['internal_id', 'name', 'entity_type'] })
```
RomuDeuxfois: The base field name is not present for all the entity
Maybe it is better to parametize it ?

opencti-platform/opencti-graphql/src/database/engine.js
RomuDeuxfois: I think the way to do it is good, but there are two queries.
I don't know if that lead to performance issues ?

opencti-platform/opencti-graphql/src/domain/user.js
```diff
@@ -139,6 +148,24 @@ export const findAll = (context, user, args) => {
   return listEntities(context, user, [ENTITY_TYPE_USER], args);
 };
 
+export const findCreators = (context, user, args) => {
+  const { onlyUsed = false, entityType = null } = args;
+  if (!onlyUsed && isUserHasCapability(user, KNOWLEDGE_UPDATE)) {
```
RomuDeuxfois: I don't understand the usage of onlyUsed

PR #3468, opened 2023-06-06
[frontend] Add the lasso selection tool in graphs (#3109)

### LINKED ISSUE #3109: Implement lasso selection in graph in addition to rectangle selection
## Use case

Implement lasso selection in graph in addition to rectangle selection

Kedae:
Be careful while testing it to use as many items in graph as possible as I had to change the render behavior of the Graph. We could have some corner cases I missed.

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/analysis/reports/ReportKnowledgeCorrelation.jsx
+++ b/opencti-platform/opencti-front/src/private/components/analysis/reports/ReportKnowledgeCorrelation.jsx
@@ -1,35 +1,23 @@
-import React, { Component } from 'react';
+import withTheme from '@mui/styles/withTheme';
 import * as PropTypes from 'prop-types';
 import * as R from 'ramda';
-import SpriteText from 'three-spritetext';
-import { debounce } from 'rxjs/operators';
-import { Subject, timer } from 'rxjs';
-import { graphql, createFragmentContainer } from 'react-relay';
+import React, { Component } from 'react';
 import ForceGraph2D from 'react-force-graph-2d';
 import ForceGraph3D from 'react-force-graph-3d';
-import withTheme from '@mui/styles/withTheme';
-import { withRouter } from 'react-router-dom';
 import RectangleSelection from 'react-rectangle-selection';
+import { createFragmentContainer, graphql } from 'react-relay';
+import { withRouter } from 'react-router-dom';
+import { Subject, timer } from 'rxjs';
+import { debounce } from 'rxjs/operators';
+import SpriteText from 'three-spritetext';
 import inject18n from '../../../../components/i18n';
 import { commitMutation, fetchQuery } from '../../../../relay/environment';
-import {
-  buildCorrelationData,
-  computeTimeRangeInterval,
-  computeTimeRangeValues,
-  decodeGraphData,
-  encodeGraphData,
-  linkPaint,
-  nodeAreaPaint,
-  nodePaint,
-  nodeThreePaint,
-} from '../../../../utils/Graph';
-import {
-  buildViewParamsFromUrlAndStorage,
-  saveViewParameters,
-} from '../../../../utils/ListParameters';
-import ReportKnowledgeGraphBar from './ReportKnowledgeGraphBar';
-import { reportMutationFieldPatch } from './ReportEditionOverview';
 import { hexToRGB } from '../../../../utils/Colors';
+import { buildCorrelationData, computeTimeRangeInterval, computeTimeRangeValues, decodeGraphData, encodeGraphData, linkPaint, nodeAreaPaint, nodePaint, nodeThreePaint } from '../../../../utils/Graph';
+import LassoSelection from '../../../../utils/graph/LassoSelection';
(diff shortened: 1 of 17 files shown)


Archidoit:
- Select things with lasso
- Select things with rectangle selection
- Some weird behaviors occur

image

image


---
@Kedae : Done

SarahBocognano:
Except the weird behavior that @Archidoit pointed out, everything seems to work.

opencti-platform/opencti-front/src/private/components/analysis/reports/ReportKnowledgeCorrelation.jsx
```diff
@@ -284,7 +271,9 @@ class ReportKnowledgeCorrelationComponent extends Component {
     );
     this.state = {
       mode3D: R.propOr(false, 'mode3D', params),
-      selectModeFree: R.propOr(false, 'selectModeFree', params),
+      selectRectangleModeFree: R.propOr(false, 'selectRectangleModeFree', params),
+      selectModeFree: !!params.selectModeFree,
```
richard-julien: Not useful

opencti-platform/opencti-front/src/private/components/analysis/reports/ReportKnowledge.jsx
```diff
@@ -123,7 +123,7 @@ class ReportKnowledgeComponent extends Component {
     const params = buildViewParamsFromUrlAndStorage(
       props.history,
       props.location,
-      `view-report-knowledge-${props.report.id}`,
+      `view-report-${props.report.id}-knowledge`,
```
richard-julien: Impact needs to be checked

opencti-platform/opencti-front/src/private/components/analysis/reports/ReportKnowledgeCorrelation.jsx
```diff
@@ -397,8 +411,24 @@ class ReportKnowledgeCorrelationComponent extends Component {
...
+
   handleToggleSelectModeFree() {
-    this.setState({ selectModeFree: !this.state.selectModeFree }, () => this.saveParameters());
+    document.removeEventListener('mousedown', this.startFreeHandFunction);
+    document.removeEventListener('mouseup', this.stopFreeHandFunction);
+    this.setState({
+      selectModeFree: !this.state.selectModeFree,
+      selectRectangleModeFree: false,
+    }, () => this.saveParameters());
+    if (!this.state.selectModeFree) {
```
richard-julien: Can be simplify

opencti-platform/opencti-front/src/utils/Graph.js
```diff
@@ -1198,3 +1198,78 @@ export const parseDomain = (data) => [
...
+      // Invert odd
+      odd = !odd;
+    }
+    j = i;
+  }
+  // If the number of crossings was odd, the point is in the polygon
+  return odd;
+};
+
+// TEST FOR UTILITY
```
richard-julien: To be cleaned

richard-julien (changes requested):

richard-julien (commented):
Comment

richard-julien:
Looks like sometimes click handler are not correctly canceled.
Like @Archidoit already mentionned.
I also be able to do something like this.

image

richard-julien:
Another question also, do we really want to complete the drag in case of open selection?

--- 
@Kedae I honesly thinks that it is much easier to use and to understand that way. May be we should go this way for a V1, and then update the algorithm if some users struggle with it ? @Jipegien Any thoughts on that ?

opencti-platform/opencti-front/src/utils/Graph.js
Kedae: Sorry, I told you garbage yesterday. Those methods are actually the ones used to make the lasso work. So I'll simply modify the comments to explain that.

opencti-platform/opencti-front/src/private/components/analysis/reports/ReportKnowledgeCorrelation.jsx
Kedae: Done

opencti-platform/opencti-front/src/private/components/analysis/reports/ReportKnowledge.jsx
Kedae: Done, misread the code, reverted those

opencti-platform/opencti-front/src/private/components/analysis/reports/ReportKnowledgeCorrelation.jsx
Kedae: Done

Jipegien:
@richard-julien @Kedae  about drag completion. As a user,  i will be upset to see my entire and complex selection being cancelled because I miss to cross my drag from 3 pixels. Plus, I think if we autocomplete the drag, it will not bother users that used to complete their drags, as it is straightforward to drag again even if elements are selected.

SarahBocognano (approved):
Tested after changes ✅

RomuDeuxfois (approved):

PR #3483, opened 2023-06-07
[backend] fix dashboard entities list filtered by related entity (#3476)

### LINKED ISSUE #3476: Some list are broken on dashboard
## Reproducible Steps

1. Create a Dashboard
2. Add a list based with filters : Entity Type Report and Related Entity: Anything
3. Kaboom

## Screenshots (optional)
image
image

SouadHadjiat:
### Proposed changes

* add missing elementId filter key
* fix stixCoreObject findAll method : handle elementId in filters
* Minor fix : deduplicate filters types

#### To test
* Add a new widget in a dashboard with entities list, a filter on an entity type and a filter on "related entity" (try to find first a report for example with a related entity like an organization)
=> There should be no error, and you should find your report

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/3476

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/database/middleware-loader.ts
+++ b/opencti-platform/opencti-graphql/src/database/middleware-loader.ts
@@ -375,7 +375,7 @@ export const buildEntityFilters = <T extends BasicStoreCommon>(args: EntityFilte
   }
   // endregion
   // Override some special filters
-  builtFilters.types = [...(types ?? []), ...entityTypes, ...relationshipTypes];
+  builtFilters.types = R.uniq([...(types ?? []), ...entityTypes, ...relationshipTypes]);
   builtFilters.filters = customFilters;
   return builtFilters;
 };
--- a/opencti-platform/opencti-graphql/src/domain/stixCoreObject.js
+++ b/opencti-platform/opencti-graphql/src/domain/stixCoreObject.js
@@ -72,23 +72,16 @@ export const findAll = async (context, user, args) => {
     throw UnsupportedError('Cant find stixCoreObject only based on relationship type, elementId is required');
   }
   let filters = args.filters ?? [];
-  const elementIds = isNotEmptyField(args.elementId) ? [args.elementId] : [];
-  const filterElementId = filters.find((f) => f.key.includes('elementId'));
-  if (filterElementId) {
-    elementIds.push(...filterElementId.values);
-    // remove elementId filter key
-    filters.splice(filters.indexOf(filterElementId), 1);
-  }
-  if (elementIds.length > 0) {
+  if (isNotEmptyField(args.elementId)) {
     // In case of element id, we look for a specific entity used by relationships independent of the direction
     // To do that we need to lookup the element inside the rel_ fields that represent the relationships connections
     // that are denormalized at relation creation.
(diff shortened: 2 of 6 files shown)


opencti-platform/opencti-graphql/src/domain/stixCoreObject.js
```diff
@@ -72,16 +72,23 @@ export const findAll = async (context, user, args) => {
     throw UnsupportedError('Cant find stixCoreObject only based on relationship type, elementId is required');
   }
   let filters = args.filters ?? [];
-  if (isNotEmptyField(args.elementId)) {
+  const elementIds = isNotEmptyField(args.elementId) ? [args.elementId] : [];
```
Kedae: Not sure why we have all those changes.
The code in stixDomainObject is much simplier, can you check that either your code is too complex or stixDomainObject doesn't work as expected ?

opencti-platform/opencti-graphql/src/database/middleware-loader.ts
```diff
@@ -368,7 +368,7 @@ export const buildEntityFilters = <T extends BasicStoreCommon>(args: EntityFilte
   }
   // endregion
   // Override some special filters
-  builtFilters.types = [...(types ?? []), ...entityTypes, ...relationshipTypes];
+  builtFilters.types = R.uniq([...(types ?? []), ...entityTypes, ...relationshipTypes]);
```
SarahBocognano, line 378: Is there a way to not use Ramda ?

SarahBocognano: Maybe using a new Set object

SarahBocognano:
Tested ✅

opencti-platform/opencti-graphql/src/domain/stixCoreObject.js
SouadHadjiat: "elementId" is sent in the filters, and it wasn't handled at all, I had to change it and this is one way I found to make it work, did you test it ?

Kedae: I think you missed my point. Either:

- Your code works but is too complex
- The code in src/domain/stixDomainObject doesn't work

Because the goal is the same. Can you check ?

Kedae: After a discussion with Julien it seems that the correct path to handle the situation is:

- Keep the existing behavior of args.elementId and args.relationship_type
- Add the support of elementId as a Filter in StixCoreObject
- Update the stixCoreObject resolver to resolve the key and provide the relationKey
- Cherry: Add a new params for elementId filter to handle the relationship_type, see with @richard-julien

SouadHadjiat: updated and it works with StixCoreObjectsFilter, but I didn't add the new param to handle the relationship_type

SouadHadjiat:
PR status :
- Updated with @Kedae suggestions
- Could you test again @SarahBocognano ?

SarahBocognano (approved):
Tested ✅

PR #3517, opened 2023-06-14
[frontend] Add Note resolution in TimeLine (#3486)

### LINKED ISSUE #3486: Resolve Note's name in the Timeline view of a Report & Cases
## Use case

Note can be used to add unstructured context in a Report or a Case. But Note's name are not displayed in the Timeline View.
image

Being able to add text (through Note) in those view participate to the capacity to automaticaly generate outcomes from OpenCTI Knowledge.

That's because the Note contained the Report/Case in STIX model. Thus, the name of the note is not resolved.

## Current Workaround

None

## Proposed Solution

create a resolver for the Note's name

RomuDeuxfois:
### Proposed changes

* Add missing inline fragments for Note in different TimeLine

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/3486

### Checklist

### Further comments

### DIFF
--- a/opencti-platform/opencti-front/src/utils/graph/LassoSelection.tsx
+++ b/opencti-platform/opencti-front/src/utils/graph/LassoSelection.tsx
@@ -79,7 +79,7 @@ const LassoSelection: FunctionComponent<LassoSelectionProps> = ({
     coord = reposition(e);
     currentContext.moveTo(coord.x, coord.y);
     currentContext.lineWidth = 1;
-    currentContext.setLineDash([0.5, 1]);
+    currentContext.setLineDash([1, 3]);
     currentContext.lineCap = 'round';
     currentContext.strokeStyle = (theme.palette.warning as SimplePaletteColorOptions)?.main ?? theme.palette.common.white;
     currentContext.beginPath();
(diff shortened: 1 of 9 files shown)


RomuDeuxfois:
Reflexion : @Kedae could we have a reusable fragment containing all the definitions of our entities ? To mutalize the fragment present in the timelines for example.


---
@Kedae : Yes we could, we probably should and may be we will, but that's a reflexion we will have globally with Julien, some similar works has been done on report I think. We will think about that in the future.

opencti-platform/opencti-front/src/utils/graph/LassoSelection.tsx
```diff
@@ -79,7 +79,7 @@ const LassoSelection: FunctionComponent<LassoSelectionProps> = ({
     coord = reposition(e);
     currentContext.moveTo(coord.x, coord.y);
     currentContext.lineWidth = 1;
-    currentContext.setLineDash([0.5, 1]);
+    currentContext.setLineDash([1, 3]);
```
SarahBocognano, line 82: Why did you change that line ? (curious)

SarahBocognano (approved):
Tested ✅

opencti-platform/opencti-front/src/utils/graph/LassoSelection.tsx
RomuDeuxfois: @Kedae is the real author of this

Kedae: Because I love css :) It's juste adjustment

PR #3526, opened 2023-06-15
[frontend] Hide add in container button for unusable types

### LINKED ISSUE #3164: Add an icon in mass operation toolbar to quickly add to container (and create container on the fly)
## Use case

Current workflow is:

* Check a box
* Click on modify
* Select add
* Add in container
* Select the container or create on the fly

Desired workflow:

* Have a button direct "Add in container"
* Be able to create a container on the fly
* Validate

Kedae:


SarahBocognano (commented):
It's still possible to add in container when select all. But the rest is ok ✅

RomuDeuxfois:
Can you give more informations about where is possible to test this scenario ? Thank you

Archidoit:
@RomuDeuxfois go to the list of a type that shouldn't be added to a container ('Label', 'Vocabulary', 'Case-Template')
You shouldn't have the 'add in container' icon in the toolbar.

Archidoit:
The disabled 'merge' icon is present for 'select all Labels',  and for vocabularies... should'nt we hide it like the add-in-container icon?
image

--- 
@Kedae: Yes, we chose to let the disabled icon because I think you have a tooltip on it saying that you have selected too many instances. The difference between disabled and hidden is : 
- Disabled: you can use it with a correct configuration
- Hidden: you should never use it with the selected types

--- 
mmm no, I selected only 1 element... and merge should never be used for Labels?

---
Well, that's the opposite in that case, you can't merge only one element :) You can definitely merge Labels and Vocabularies.

---
What's strange is that I have the icon when I select 1 vocabulary but not when I select 1 label...
No merge icon when selecting labels:
image

Archidoit:
Problems with the merge icon:

- In a screen (labels or reports etc) : if I select all the instances, then deselect instances to keep only 2 instances (so not too much to do a merge) : merge icon is disabled:
image


- If I select manually 2 labels: no merge icon
If I select manually 2 vocabularies: I have a merge icon

- When the merge icon is disabled: no tooltip explaining why

PR #3577, opened 2023-06-26
[frontend/backend] Dashboard persistance and update in server (#issue/3163)

### LINKED ISSUE #3163: Be able to define default dashboards in groups / organizations
## Use case

Administrator needs to be able to define default Dashboard in Settings for Groups and Organizations. Default Dashboards need to be listed in Overview of Users, Group and Organizations (As this feature requires to implement an overview of an organization in the settings panel, need sync with https://github.com/OpenCTI-Platform/opencti/issues/3520).

Users without admin (or dedicated) capability should not be able to modify the content of Group, Organization default Dashboards.

For a user when login, this order should apply for the default dashboard:
- User defined (already implemented)
- Organization defined
- Group defined

tomibennett:
### Proposed changes
The choice of the dashboard to be displayed on the landing page is now persisted and updated accordingly on the server. 

Previously to this PR, it was only managed in the browser of the logged user.

### Related issues
[issue/3163](https://github.com/OpenCTI-Platform/opencti/issues/3163)

### Checklist

### Further comments

This is a first step to have finer control on those dashboards. The full story is to allow managers or other groups/organizations admins to change default dashboards for their teammates. See issue mentioned above


opencti-platform/opencti-front/src/private/components/Dashboard.jsx
```diff
@@ -908,25 +905,47 @@ const CustomDashboard = ({ dashboard, timeField }) => {
...
+    changeDashboard({
+      variables: {
+        input: [{
+          key: 'dashboard_id',
+          value: (event.target.value === 'default' ? [null] : event.target.value),
+        }],
+      },
+    });
+
+    saveLocalDashboardPreferences({ dashboard: event.target.value, timeField });
```
yassine-ouaamou: I don't think it is needed to add Local to the name of the function. Maybe just saveDashboardPreferences

yassine-ouaamou:
image
Capital letter to Dashboard and maybe call it Default Dashboard?

opencti-platform/opencti-front/src/private/components/Dashboard.jsx
```diff
@@ -908,25 +905,47 @@ const CustomDashboard = ({ dashboard, timeField }) => {
...
+
 const Dashboard = () => {
   const classes = useStyles();
-  const [view, saveView] = useViewStorage('view-dashboard');
-  const { dashboard = 'default', timeField = 'technical' } = view;
-  const handleChangeTimeField = (event) => saveView({ dashboard, timeField: event.target.value });
-  const handleChangeDashboard = (event) => saveView({ dashboard: event.target.value, timeField });
+  const { me: { dashboard_id: dashboardId } } = useContext(UserContext);
+  const [changeDashboard] = useMutation(changeDashboardPref);
+  const [localDashboardPreferences, saveLocalDashboardPreferences] = useLocalStorage('view-dashboard', {});
```
Kedae: Ok, I now understand why we didn't understood each other yesterday. You were right, we don't need that anymore.

opencti-platform/opencti-front/src/private/components/Dashboard.jsx
```diff
@@ -908,25 +905,47 @@ const CustomDashboard = ({ dashboard, timeField }) => {
...
+  mutation DashboardPrefMutation($input: [EditInput]!) {
+    meEdit(input: $input) {
+      dashboard_id
+    }
+  }
+`;
+
 const Dashboard = () => {
   const classes = useStyles();
-  const [view, saveView] = useViewStorage('view-dashboard');
```
Kedae, line 913: Can you delete the declaration from ListParameters if it is no longer used ?

opencti-platform/opencti-graphql/tests/04-sync/sync-utils.js
```diff
@@ -61,7 +61,7 @@ export const SYNC_START_QUERY = `mutation SynchronizerStart($id: ID!) {
     }
   `;
 
-export const UPLOADED_FILE_SIZE = 36926;
+export const UPLOADED_FILE_SIZE = 37891;
```
RomuDeuxfois: Why we have this changes ?

opencti-platform/opencti-graphql/src/modules/attributes/internalObject-registrationAttributes.ts
```diff
@@ -78,6 +78,7 @@ const internalObjectsAttributes: { [k: string]: Array<AttributeDefinition> } = {
   ],
   [ENTITY_TYPE_USER]: [
     { name: 'user_email', type: 'string', mandatoryType: 'external', multiple: false, upsert: false },
+    { name: 'dashboard_id', type: 'string', mandatoryType: 'external', multiple: false, upsert: true },
```
RomuDeuxfois: dashboard_id is not mandatory if we return null is some case
What do you think ?

opencti-platform/opencti-front/src/private/components/Dashboard.jsx
```diff
@@ -908,25 +905,47 @@ const CustomDashboard = ({ dashboard, timeField }) => {
...
-  const { dashboard = 'default', timeField = 'technical' } = view;
-  const handleChangeTimeField = (event) => saveView({ dashboard, timeField: event.target.value });
-  const handleChangeDashboard = (event) => saveView({ dashboard: event.target.value, timeField });
+  const { me: { dashboard_id: dashboardId } } = useContext(UserContext);
```
Kedae: useAuth()

RomuDeuxfois:
Maybe not relatif to this PR, but :

The app crash when I change the date
[screen-capture (2).webm](https://github.com/OpenCTI-Platform/opencti/assets/53513584/442c74ed-39c8-42df-bb44-a6641906eccb)
The field is not fill by the dashboard name
[screen-capture (3).webm](https://github.com/OpenCTI-Platform/opencti/assets/53513584/ccaa5b02-69a4-472b-ac0e-72a1c378f321)

opencti-platform/opencti-front/src/private/components/Dashboard.jsx
tomibennett: would you like to give me a little bit of info on why it is preferable to use `useAuth()` here?

opencti-platform/opencti-graphql/src/modules/attributes/internalObject-registrationAttributes.ts
tomibennett: I do and I also do reckon I need more info on those options. Would you point me a resource on that matter, or take a little bit of time to explain me directly?

opencti-platform/opencti-graphql/tests/04-sync/sync-utils.js
tomibennett: It was a mistake, it will be reverted

opencti-platform/opencti-front/src/private/components/Dashboard.jsx
Kedae: That's a choice we made to hide the UserContext as we use it a lot of time in the app and add some security.

tomibennett:
> Maybe not relatif to this PR, but :
> 
> The app crash when I change the date [screen-capture (2).webm](https://github.com/OpenCTI-Platform/opencti/assets/53513584/442c74ed-39c8-42df-bb44-a6641906eccb) The field is not fill by the dashboard name [screen-capture (3).webm](https://github.com/OpenCTI-Platform/opencti/assets/53513584/ccaa5b02-69a4-472b-ac0e-72a1c378f321)

I can't reproduce the bug on my side. Can we take a moment to see it together?

opencti-platform/opencti-front/src/private/components/Dashboard.jsx
tomibennett: I didn't see `view-dashboard` defined anywhere other that here. Maybe I've missed it though, or misunderstood your request.

Do you refers to **params** declared in `opencti-platform/opencti-front/src/utils/ListParameters.js`?

Kedae: useViewStorage is no longer used I think you can remove it completely from the code

opencti-platform/opencti-front/src/private/components/Dashboard.jsx
tomibennett: There is still the date persisted (technical vs. functional) though. 

I see two possible way:
- leave it as it is now
- limit the the local storage to this value
- manage it like the dashboard (persist it on the back along with the dashboard preference)

I favor the second option, which seems to be the simplest wdyt

Kedae: Yep, let's only store the date.

opencti-platform/opencti-front/src/schema/relay.schema.graphql
```diff
@@ -1244,7 +1244,8 @@ type User implements BasicObject & InternalObject {
   theme: String
   language: String
   external: Boolean
-  dashboard: String
+  dashboard_id: ID
+  dashboard: Workspace
```
SouadHadjiat: I'm wondering why do we need to expose `dashboard_id` and `dashboard`, can't we just use `dashboard` ?

Kedae: We need it because we will have a bit of logic with those dashboards : 
- Show in admin what dashoard a user has as default
- Compute at login a list of default dashboards for a user (coming from groups and orgas)

So dashboard_id will still be the "default dashboard choosen by user" and dashboard will be dashboards: [Workspace] representing the list of default dashboards assigned to him through groups and orga

SouadHadjiat: Ok, I still don't understand why `dashboard_id` and not `dashboard`, which can get us the `name`. In the admin, we show the name of the default dashboard, we don't just use the id.

Kedae: Dashboard_id = default dashboard choosen by user
Dashboard = default dashboard for user (will be computed from other stuff)

SouadHadjiat: A user can have two different default dashboards then ?

SouadHadjiat: I can see in user resolver :
```
dashboard: (current, _, context) => {
      return current.dashboard_id ? findWorskpaceById(context, context.user, current.dashboard_id) : null;
```

So dashboard and dashboard_id are the same.

Kedae: Nop. He will have a list of default dashboards defined by its organization and groups and only one **active** default dashboard that will be either the first one of the list or the one stored in dashboard_id because he has choosen that one

Kedae: > I can see in user resolver :
> 
> ```
> dashboard: (current, _, context) => {
>       return current.dashboard_id ? findWorskpaceById(context, context.user, current.dashboard_id) : null;
> ```
> 
> So dashboard and dashboard_id are the same.

Yes because we don't have groups/orgas yet

SouadHadjiat: wow, this is really not easy to understand by just looking at the code

SouadHadjiat: I still don't really get why in the `me` query, we get `dashboard_id` instead of `dashboard { id }`

Kedae: > I still don't really get why in the `me` query, we get `dashboard_id` instead of `dashboard { id }`

Yep, we could do that. But in anycase it will be modified in next step.

opencti-platform/opencti-front/src/private/components/Dashboard.jsx
tomibennett: My intention is to explicit that the *save* operation is not an update to the server, which can be confusing in this context because we are doing both (the dashboard is updated but the timefield is saved locally). I used this verbosity as a way to have this distinction.

That being said, I may have missed better way to handle this problem, in which case, I welcome suggestion.

opencti-platform/opencti-front/src/private/components/Dashboard.jsx
tomibennett: I realize that `useLocalStorage` is typed with `LocalStorage` interface, which does not seem to be designed for this kind of purpose, but to help with list pagination and various other elements manipulation.

It works as it, partly because the component is in `jsx` and partly because all fields in `useLocalStorage` are optional, but as soon as the component move to typescript, the transpiler will surely complain.

I can **create another hook**, for this special case that allow any kind of object to be stored.
Or **add another field to be handled by `LocalStorage` interface** (but that seems not a good way to resolve our problem),

do you have any other idea?

opencti-platform/opencti-graphql/src/modules/attributes/internalObject-registrationAttributes.ts
RomuDeuxfois: I can explain to you directly, because we don't have clear documentation on this

RomuDeuxfois:
> > Maybe not relatif to this PR, but :
> > The app crash when I change the date [screen-capture (2).webm](https://github.com/OpenCTI-Platform/opencti/assets/53513584/442c74ed-39c8-42df-bb44-a6641906eccb) The field is not fill by the dashboard name [screen-capture (3).webm](https://github.com/OpenCTI-Platform/opencti/assets/53513584/ccaa5b02-69a4-472b-ac0e-72a1c378f321)
> 
> I can't reproduce the bug on my side. Can we take a moment to see it together?

Yes

Kedae (approved):

PR #3599, opened 2023-06-29
[backend/frontend] Export investigation in JSON as STIX report (#1374)

### LINKED ISSUE #1374: Export investigation in JSON or CSV format
Please replace every line in curly brackets { like this } with appropriate answers, and remove this line.

## Problem to Solve

It would be useful to be able to export investigations made from the investigation panel into JSON (and maybe CSV), in order to be able to share it with others. 

The envisioned workflow is the following : 
1/ several reports are integrated to OpenCTI
2/ An analyst investigate the knowledge from the investigation tab
3/ Based on the result of the investigation, the analyst write a short analysis/commentary, and join the JSON of his/her OpenCTi investigation in order to share it with partners, constituents, clients etc. 
4/ The JSON can be directly integrated in tools of the partners.

## Current Workaround

Make an investigation from the knowledge tab of a report in order to be able to export it.

## Proposed Solution
The export could be a button at the top right, or a dedicated tab for files, just like it exists in other spaces today.
Only the objects displayed on the graph should be exported, with the displayed relations. This should take into account active filters. 
Attributes and parameters of the object should be exported (source, dates, confidence level).
A useful option could be to be able to add for each object and each relation the list of related reports.

SouadHadjiat:
### Proposed changes

* Add toStixReportBundle query, which returns the bundle of a workspace (only for investigation type) with all the investigation objects (stix entities and relationships) and creates a new report containing all these objects.
* Add a button "Download as STIX report" in investigation header : calls toStixReportBundle and creates a file with the bundle to download.

With this file, you can import it as a workbench, then validate it to create the report generated from the investigation.

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/1374

### Checklist

### Further comments


Archidoit:
Tested, everything is Ok :) 
Just a question to be sure: the relations of type refs are not exported? (only relationships and sightings are)
Ex: in the capture below, only 3 relationships are exported and found in the generated report (the 2 'targets' and 1 'uses')
image

SouadHadjiat:
> Tested, everything is Ok :) Just a question to be sure: the relations of type refs are not exported? (only relationships and sightings are) Ex: in the capture below, only 3 relationships are exported and found in the generated report (the 2 'targets' and 1 'uses') image

Yes, only stix relationships are exported, and the "refs" are not stix relationships. I exported first all relationships but I had an error (because refs couldn't be converted in stix) and @RomuDeuxfois helped me undertstand it.

Thank you for your tests @Archidoit !

opencti-platform/opencti-graphql/src/modules/workspace/workspace-domain.ts
```diff
@@ -77,6 +82,30 @@ export const objects = async (context: AuthContext, user: AuthUser, workspaceId:
...
+  if (workspace.type !== 'investigation') {
+    throw FunctionalError('You can only export investigation objects as a stix report bundle');
+  }
+  const types = ['Stix-Core-Object', 'Stix-Core-Relationship', 'Stix-Sighting-Relationship'];
+  const filters = [{ key: buildRefRelationKey(RELATION_HAS_REFERENCE), values: [workspace.id] }];
+  const stixObjects = await stixLoadByFilters(context, user, types, { filters });
+  const reportData = {
+    name: workspace.name,
+    published: workspace.created_at,
+    type: 'report',
```
RomuDeuxfois: You can use the constant
ENTITY_TYPE_CONTAINER_REPORT.toLowerCase()

opencti-platform/opencti-graphql/src/modules/workspace/workspace-domain.ts
```diff
@@ -77,6 +82,30 @@ export const objects = async (context: AuthContext, user: AuthUser, workspaceId:
   return listThings(context, user, types, finalArgs);
 };
 
+export const exportObjectsAsStixReportBundle = async (context: AuthContext, user: AuthUser, workspace: BasicStoreEntityWorkspace): Promise<string> => {
```
RomuDeuxfois: Is it really an export?

I would rather have thought for toStixBundleReport

opencti-platform/opencti-graphql/src/modules/workspace/workspace-domain.ts
```diff
@@ -77,6 +82,30 @@ export const objects = async (context: AuthContext, user: AuthUser, workspaceId:
...
+  if (workspace.type !== 'investigation') {
```
RomuDeuxfois, line 87: We don't have an enum to handle it ?

opencti-platform/opencti-front/src/private/components/workspaces/WorkspaceHeader.jsx
```diff
@@ -139,6 +149,19 @@ const WorkspaceHeader = ({
...
   const handleCloseManageAccess = () => setDisplayManageAccess(false);
 
+  const handleDownloadAsStixReport = () => {
+    fetchQuery(workspaceHeaderToStixReportBundleQuery, { id: workspace.id })
+      .toPromise()
+      .then((data) => {
+        const toStixBundleData = data?.workspace?.toStixReportBundle;
+        if (toStixBundleData) {
+          const blob = new Blob([toStixBundleData], { type: 'text/json' });
+          const fileName = `${workspace.name}-export-stix-report`;
```
RomuDeuxfois: It's maybe better for the name to be close to the actual export file (connector side) :
image

opencti-platform/opencti-graphql/src/modules/workspace/workspace-domain.ts
```diff
@@ -77,6 +82,30 @@ export const objects = async (context: AuthContext, user: AuthUser, workspaceId:
...
+    throw FunctionalError('You can only export investigation objects as a stix report bundle');
+  }
+  const types = ['Stix-Core-Object', 'Stix-Core-Relationship', 'Stix-Sighting-Relationship'];
+  const filters = [{ key: buildRefRelationKey(RELATION_HAS_REFERENCE), values: [workspace.id] }];
+  const stixObjects = await stixLoadByFilters(context, user, types, { filters });
+  const reportData = {
+    name: workspace.name,
```
RomuDeuxfois, line 94: Can overlap an existing report at import (maybe is a KO limit)

opencti-platform/opencti-graphql/src/modules/workspace/workspace-domain.ts
```diff
@@ -77,6 +82,30 @@ export const objects = async (context: AuthContext, user: AuthUser, workspaceId:
...
+    spec_version: STIX_SPEC_VERSION,
+  };
+  const report = { ...reportData, id: generateStandardId(ENTITY_TYPE_CONTAINER_REPORT, reportData) };
+  const allObjects = [...stixObjects, report];
+  const bundle = {
+    id: `bundle--${uuidv4()}`,
+    type: 'bundle',
+    objects: allObjects,
+  };
+  return JSON.stringify(bundle);
```
RomuDeuxfois, line 107: @richard-julien Do we want to export a json rather than a string so it can be used directly via the API?

RomuDeuxfois:
Tested ✅

opencti-platform/opencti-graphql/src/modules/workspace/workspace-domain.ts
SouadHadjiat: I didn't find it, and I thought about creating one, but for now it's the only place we use it, so it didn't seem very useful to create an enum for one condition.

opencti-platform/opencti-graphql/src/modules/workspace/workspace-domain.ts
SouadHadjiat: This was requested in the brainstorm to name the report after the investigation

opencti-platform/opencti-graphql/src/modules/workspace/workspace-domain.ts
SouadHadjiat: I could even use this method : `convertTypeToStixType(ENTITY_TYPE_CONTAINER_REPORT)`

marieflorescontact:
Tested, looks good

opencti-platform/opencti-graphql/src/modules/workspace/workspace-domain.ts
RomuDeuxfois: @Kedae maybe an opinion on that ?

SouadHadjiat:
PR status : ready for more reviews / waiting for approval

RomuDeuxfois (approved):

PR #3602, opened 2023-06-30
Make the platform supports opinion custom vocab / scale (#3003)

### LINKED ISSUE #3003: Make the platform supports opinion custom vocab / scale
## Use case

Currently, opinion scale is:

image

When changing the vocab in the opinions_ov, the scale remains the same.

### LINKED ISSUE #3587: Users should not be able to express multiple opinions on the same Entity, only update it
## Use case

Currently, a user can create multiple opinions about the same Entity, even if this is an opinion with the same value if the description is different.

## Current Workaround

None

## Proposed Solution

One opinion per user per Entity

RomuDeuxfois:
### Proposed changes

* Add order property for ALL vocab (not required but fill to 0 at creation)
* Use opinion vocab in Opinions Radar component
* Fix create/update opinion in Opinions Radar component
* Fix capabilities in Opinions Radar component
* Remove opinion on Hidden Types list (opinion is not display in the entire platform)

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/3003

### Checklist

### Further comments


Archidoit:
- Can't sort opinion vocab by usages or order
image

--> fixed ✅ 

- opinion vocab sorting by name seems incorrect
image

--> fixed ✅ 

- what about adding the 'delete' icon in the toolbar for vocabs?
image

--> not handle here

- in the toolbar, the 'add in container' appears if we do a select all, but not if we select some vocabs manually
image

--> fixed ✅

Archidoit:
Update an opinion:
- pre-created opinions are not completely displayed even if there are some place
image

opencti-platform/opencti-front/src/utils/Localization.js
```diff
@@ -1994,9 +1994,11 @@ const i18n = {
       Reset: 'Reiniciar',
       'Do you want to clear the works of this connector?':
         '¿Quieres eliminar todas las tareas de este conector?',
+      'strongly-disagree': 'muy en desacuerdo',
```
Archidoit: these translations are not seen in the UI (in 'update an opinion, in the Opinion Radar, and in the vocab lists...

Archidoit:
missing translations in vocab lines: Aliases, Usages, Create a vocabulary, Some fields in usage are mandatory
image

image

image

--> fixed ✅

opencti-platform/opencti-front/src/utils/Localization.js
Archidoit: are they used somewhere?

RomuDeuxfois:
> missing translations in vocab lines: Aliases, Usages, Create a vocabulary, Some fields in usage are mandatory image
> 
> image
> 
> image

--> Fixed ✅

opencti-platform/opencti-front/src/private/components/analysis/opinions/StixCoreObjectOpinions.jsx
```diff
@@ -1,40 +1,29 @@
...
-    const { title, variant, height, marginTop, stixCoreObjectId, field } = this.props;
-    const opinionsDistributionVariables = {
-      objectId: stixCoreObjectId,
-      field: field || 'opinion',
-      operation: 'count',
-      limit: 8,
-    };
-    return (
+const StixCoreObjectOpinions = (props) => {
+  const { stixCoreObjectId, variant, height, marginTop } = props;
```
Kedae: Can you inline the props ?

opencti-platform/opencti-front/src/private/components/analysis/opinions/StixCoreObjectOpinionsRadar.jsx
```diff
@@ -38,93 +39,73 @@ const styles = () => ({
...
-  'disagree',
-  'neutral',
-  'agree',
-  'strongly-agree',
-];
+const StixCoreObjectOpinionsRadarComponent = (props) => {
+  const classes = useStyles();
+  const { t } = useFormatter();
+  const theme = useTheme();
+  const { stixCoreObjectId, data, variant, height, marginTop, paginationOptions } = props;
```
Kedae: Same

RomuDeuxfois:
> * Can't sort opinion vocab by usages or order

--> Fixed order, but we don't want to sort on usages for now

>   image

--> Open a new issue for that

> * opinion vocab sorting by name seems incorrect
>   image

--> Open a new issue for that

> * what about adding the 'delete' icon in the toolbar for vocabs?
>   image

--> Open a new issue for that

> * in the toolbar, the 'add in container' appears if we do a select all, but not if we select some vocabs manually
>   image

--> Open a new issue for that

RomuDeuxfois:
PR ready to be reviewed

marieflorescontact:
Why is there no `delete`? 
Capture d'écran 2023-07-07 101033

Archidoit:
@marieflorescontact  it's because the opinion is used in at least one entity --> we can't delete it.

helene-nguyen:
[Frontend]

- Can add negative order

image

-> Can be used like this

- Cannot merge more than 4 opinions ?

image

- When merging, nothing happens except a pop-up message. What is the expected output?

helene-nguyen:
[Frontend]

- Can add negative value for confidence level

image

- What is the impact of confidence level?

Archidoit:
@helene-nguyen when you merge entities, there are reunited in a single one (but it's a background task, you have to wait a bit to see the effect, it is not immediate)
--> but indeed the merge of vocabularies doesn't seem to work.

Archidoit:
it's possible to have several opinion_ov with the same order. Is it normal? How to decide which is displayed first then?

RomuDeuxfois:
Bugs : we can't update confidence level on entity with enforce reference activated

Archidoit:
When entering a float number (ex: 1.2) for an opinion order, and validate: the opinion order is put with an integer (ex: 1 here) --> the user shouldn't be able to enter numbers with comma (see status template)

-> Fixed ✅

helene-nguyen:
When creating a new opinion on the entity, you can add an opinion to this existing opinion, and another user can add another opinion on it => it creates opinions in opinions

opinion

RomuDeuxfois:
> it's possible to have several opinion_ov with the same order. Is it normal? How to decide which is displayed first then?

It's order by alphabetical order in this case

opencti-platform/opencti-graphql/src/migrations/1688710489709-add_order_to_opinion_ov.js
```diff
@@ -0,0 +1,63 @@
...
+
+  const updateVocabulary = async (vocabulary) => {
+    const defaultVocabulary = defaultVocabularies.get(vocabulary.name);
+    if (defaultVocabulary) {
+      const patch = { order: defaultVocabulary.order };
+      await elReplace(vocabulary._index, vocabulary.id, { doc: patch });
+    }
+  };
+
+  await Promise.map(vocabularies, updateVocabulary, { concurrency: ES_MAX_CONCURRENCY });
```
SouadHadjiat, line 55: what is the difference between updateEntityQuery and updateVocabulary ? Why do we need both for the migration ?

SouadHadjiat: Would it work if we get the defaultVocabularies and use them as a param in the script above ? then we set the order with either the default vocabulary order, or 0

RomuDeuxfois: We need to add, by default, 0 to all existing vocabulary.
And  for opinion, we need to have a valid order to display opinion in radar correctly (good ordering).

SouadHadjiat: Ok I understand now, still I think we could do everything in the same script, something like :

```
script: {
      params: { defaultVocabulariesList },
      source: `ctx._source.order = 0; 
          if (ctx._source.category == 'opinion_ov') { 
             for (defaultVocab in params.defaultVocabularies) { 
                 if (defaultVocab.name == ctx._source.name)) {
                    ctx._source.order = defaultVocab.order;
                 } 
             } 
         }`,
}
```

But maybe it's not really important, if the migration works already, I guess it's not necessary to change it.

RomuDeuxfois:
From Sarah

- The order doesn't work at creation of a new vocab (it works on update though)

Otherwise, everything works perfectly 🕺🏼 

--> Fixed 🆗

opencti-platform/opencti-graphql/src/modules/vocabulary/vocabulary-utils.ts
```diff
@@ -110,16 +110,16 @@ export const openVocabularies: Record<VocabularyCategory, Array<{ key: string, d
...
   // C
   case_severity_ov: [
-    { key: '01 - low', description: 'Low impact', aliases: ['low'] },
-    { key: '02 - medium', description: 'Medium impact', aliases: ['medium'] },
-    { key: '03 - high', description: 'High impact', aliases: ['high'] },
-    { key: '04 - critical', description: 'Critical impact', aliases: ['critical'] },
+    { key: 'low', description: 'Low impact', aliases: ['low'], order: 1 },
+    { key: 'medium', description: 'Medium impact', aliases: ['medium'], order: 2 },
+    { key: 'high', description: 'High impact', aliases: ['high'], order: 3 },
+    { key: 'critical', description: 'Critical impact', aliases: ['critical'], order: 4 },
```
SouadHadjiat, line 116: why did the key change ? since we have default orders here too, don't we want them to be updated in the database ?

SarahBocognano (approved):

PR #3612, opened 2023-07-03
[frontend] avoid re-render when click in the settings hidden entity types list (#3611)

### LINKED ISSUE #3611: In settings, when clicking the first time on the "hidden entities", re-render and force to reclick
## Description

1. Go in Settings => Parameters
2. Click on the dropdown selector for hidden entities

=> The first time, it triggers a full re-render and the user is forced to click a second time.

Archidoit:
Github issue: https://github.com/OpenCTI-Platform/opencti/issues/3611


marieflorescontact:
Tested locally, looks good !

Issue #3844, opened 2023-07-25
Dashboard crashes


Klaas-Vaak:
## Prerequisites

- [x] I read the [Deployment and Setup](https://www.notion.so/OpenCTI-Public-Knowledge-Base-d411e5e477734c59887dad3649f20518) section of the OpenCTI documentation as well as the [Troubleshooting](https://www.notion.so/Troubleshooting-ebc8fb04137d495aad917bc20340b9a6) page and didn't find anything relevant to my problem.
- [x] I went through old GitHub issues and couldn't find anything relevant
- [x] I googled the issue and didn't find anything relevant

## Environment

1. OS (where OpenCTI server runs): Ubuntu 20.04
2. OpenCTI version: OpenCTI 5.9.4
3. OpenCTI client: frontend
5. Other environment details: docker deployment. Server 16 cores and 64GB RAM

## Additional information

The TOP LABELS (3 LAST MONTHS) widget doesn't load and eventually an error appears:

Error
An unknown error occurred. Please contact your administrator or the OpenCTI maintainers.

Just before this error appears I see this entry in the logs:

[...]

PR #3996, opened 2023-08-07
[backend/frontend] improve relationships distribution query (#3844)

SouadHadjiat:
### Proposed changes

* Fix targeted country dashboard query (we didn't return only countries ids, but also entities ids related to the countries, and return 100 results instead of 20)
* Set query results to 0 for aggregations (we only need aggregations), and use the limit param for aggregations size
* Add toRole param to have a more accurate filter (instead of `*_to`)

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/3844

There is a performance issue with relations distribution query which can result to a timeout

### Checklist

- [x] I tested the code for its functionality => works locally, but I can't test the perf since I only have a small db

### Further comments


helene-nguyen:
Tested locally and all good :)

helene-nguyen (approved):

richard-julien (approved):

Issue #3986, opened 2023-08-05
In investigation graph, when selecting a marking or label, displaying a not found error


SamuelHassine:
## Use case

In investigation graph, when selecting a marking or label, displaying a not found error.

PR #4003, opened 2023-08-08
[frontend] Stix Meta Object details in Investigation graphs (#3986)

Archidoit:
Github issue: https://github.com/OpenCTI-Platform/opencti/issues/3986

To test the issue:
- Go to an investigation graph.
- Click on an entity that has meta relations (labels, markings, kill chain phase or external reference).
- Expand the entity (via a button in the toolbar)
- Click on a meta object. A details panel opens and it has an error.


SamuelHassine:
@Archidoit I think this is also the case for kill chain phases and external references (all meta objects). Whether we find a way to display them properly or to discard them in a generic approach.

Archidoit:
@SamuelHassine indeed. I'm going to create a detail vue for the meta objects. Maybe I can also create a generic query (and a module) for stixMetaObjects. This will enable to do a single query for all the meta objects, and then display their attribute in the details part.

SarahBocognano (commented):
Tested ✅ Works fine

Archidoit:
PR ready for review

yassine-ouaamou:
All good for me 👍

yassine-ouaamou (approved):

Issue #3922, opened 2023-08-03
Possible to lockout workspaces (dashboards / investigation) in a case


SamuelHassine:
## Description

Possible to lockout workspaces (dashboards / investigation) in a case:

Repro:

1. Create a dashboard with a specific user, allow "All users" with "can manage"
2. Delete the user that created the dashboard
3. Login with any other user
4. Go to the dashboard, remove the "All users" permission

Dashboard will be no longer accessible from anyone.

PR #4014, opened 2023-08-09
[backend] Workspace should have at least one valid admin (#3922)

Archidoit:
Github issue: https://github.com/OpenCTI-Platform/opencti/issues/3922

demo: see link in Teams

### DIFF
--- a/opencti-platform/opencti-graphql/src/utils/authorizedMembers.ts
+++ b/opencti-platform/opencti-graphql/src/utils/authorizedMembers.ts
@@ -1,8 +1,19 @@
+import { uniq } from 'ramda';
 import type { AuthContext, AuthUser } from '../types/user';
-import type { BasicStoreEntity } from '../types/store';
+import type { BasicGroupEntity, BasicOrganizationEntity, BasicStoreEntity } from '../types/store';
 import type { MemberAccess } from '../generated/graphql';
-import { AuthorizedMember, validateUserAccessOperation } from './access';
-import { findAllMembers } from '../domain/user';
+import {
+  AuthorizedMember,
+  BYPASS,
+  MEMBER_ACCESS_ALL,
+  MEMBER_ACCESS_RIGHT_ADMIN,
+  SYSTEM_USER,
+  validateUserAccessOperation
+} from './access';
+import { findAllMembers, findById as findUser } from '../domain/user';
+import { findById as findGroup } from '../domain/group';
+import { findById as findOrganization } from '../domain/organization';
+import { RELATION_MEMBER_OF, RELATION_PARTICIPATE_TO } from '../schema/internalRelationship';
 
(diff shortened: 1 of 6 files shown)


Archidoit:
Comment: If a user is deleted and he is the only admin of a workspace, the workspace will be unadministrable. This is a more global problem, that will be resolved in a future issue: https://github.com/OpenCTI-Platform/opencti/issues/4026

yassine-ouaamou (approved):

yassine-ouaamou:
Behaviour tested, code checked, PR approved!

opencti-platform/opencti-graphql/src/utils/authorizedMembers.ts
```diff
@@ -29,3 +40,37 @@ export const getAuthorizedMembers = async (
...
+export const containsValidAdmin = async (
+  context: AuthContext,
+  authorized_members: Array<AuthorizedMember>,
+) => {
+  const adminIds = authorized_members
+    .filter((n) => n.access_right === MEMBER_ACCESS_RIGHT_ADMIN)
+    .map((e) => e.id);
+  if (adminIds.includes(MEMBER_ACCESS_ALL)) { // everyone is admin
+    return true;
+  }
```
SouadHadjiat, line 59: maybe add a condition here that checks that there is at least one member with admin right, before trying to resolve the admin ids, otherwise return false 
like this :
```
if (adminIds.length == 0) { 
   return false 
} 
```

It's the condition we had before.

opencti-platform/opencti-graphql/src/utils/authorizedMembers.ts
```diff
@@ -29,3 +40,37 @@ export const getAuthorizedMembers = async (
...
+  const groups = (await Promise.all(adminIds.map((id) => findGroup(context, SYSTEM_USER, id))))
+    .filter((n) => n) as BasicGroupEntity[];
+  const organizations = (await Promise.all(adminIds.map((id) => findOrganization(context, SYSTEM_USER, id))))
+    .filter((n) => n) as BasicOrganizationEntity[];
+  const groupsMembersIds = uniq(groups.map((group) => group[RELATION_MEMBER_OF]).flat()) as string[];
+  const organizationsMembersIds = uniq(organizations.map((o) => o[RELATION_PARTICIPATE_TO]).flat());
+  const userIds = adminIds
+    .filter((id) => !groups.map((o) => o.id).includes(id)
+      && !organizations.map((o) => o.id).includes(id))
+    .concat(groupsMembersIds, organizationsMembersIds);
```
SouadHadjiat, line 69: groups.map((o) => o.id) and organizations.map((o) => o.id) are done inside the filter, maybe we could extract it before ? have organizationIds and groupIds, then use the include ?

Issue #3848, opened 2023-07-27
Elasticsearch ApiKey authentication not working


fraekfyr76:
## Description

According to https://github.com/OpenCTI-Platform/opencti/blob/master/opencti-platform/opencti-graphql/src/database/engine.js#L142 you can use an apikey to auth with your search engine (Elasticsearch) for example with an environment variable like `ELASTICSEARCH__API_KEY=<base64 string>`.

This does not work because the default search engine client (Opensearch) does not support apikey authentication.

## Environment

1. OS (where OpenCTI server runs): Docker
2. OpenCTI version: OpenCTI 5.9.6

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Run Elasticsearch and create an apikey
2. Run OpenCTI with the environment variable `ELASTICSEARCH__API_KEY=<your apikey>`
3. See `[OPENCTI] Platform start fail` error message

## Expected Output

Not getting the `[OPENCTI] Platform start fail` error message

## Actual Output

```
{"category":"APP","error":{"context":{"category":"technical","error":"security_exception: [security_exception] Reason: missing authentication credentials for REST request [/]","http_status":500,"reason":"[SEARCH] Search engine seems down"},"message":"A configuration error has occurred","name":"ConfigurationError","stack":"ConfigurationError: A configuration error has occurred\n    at error (/opt/opencti/build/src/config/errors.js:8:10)\n    at ConfigurationError (/opt/opencti/build/src/config/errors.js:61:53)\n    at /opt/opencti/build/src/database/engine.js:185:15\n    at processTicksAndRejections (node:internal/process/task_queues:95:5)\n    at searchEngineInit (/opt/opencti/build/src/database/engine.js:175:3)\n    at checkSystemDependencies (/opt/opencti/build/src/initialization.js:138:3)\n    at platformStart (/opt/opencti/build/src/boot.js:179:5)"},"level":"error","message":"[OPENCTI] Platform start fail","timestamp":"2023-07-27T12:11:18.106Z","version":"5.9.6"}
```

## Additional information

Opensearch client not supporting apikeys: 
https://github.com/opensearch-project/opensearch-js/blob/main/lib/Connection.js#L329

richard-julien:
Hi @fraekfyr76 , thanks for the report. 
As its impossible to check/detect the client on this case, we will add a new configuration option to have auto detection of the client or directly defined the correct one.

ckyoog:
[...]

PR #4023, opened 2023-08-09
[backend] Add engine_selector option to support elk specific auth strategy (#3848)

richard-julien:


yassine-ouaamou:
All good for me!
Just need to update the documentation with the new added parameter

yassine-ouaamou (approved):

PR #4042, opened 2023-08-10
[backend/frontend] Fix ingestion creation (#4020)

### LINKED ISSUE #4020: Ingestion page is broken
Reproduction : 

1. Create a new ingestion on a platform without any ingestion
2. Reload
3. Kaboom

Steps:

1. Create a new attribute current_state_date in Synchronizer
2. Delete by query every synchronizer that have {} as current_state
3. Migrate existing current_state in the new attribute IF DATE FORMAT
4. Change the registrationAttributes into "date"

SouadHadjiat:
### Proposed changes

* migrate current_state field to current_state_date with date type (schema definition)
* fix prepareElementsForIndexing method to better handle objects

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/4020

### Checklist

### Further comments


opencti-platform/opencti-graphql/src/migrations/1691677872733-sync-from-current-state-to-current-state-date.js
```diff
@@ -0,0 +1,42 @@
...
+    {
+      script: {
+        source: 'if (ctx._source.current_state == params.emptyCurrentState) {'
+          + 'ctx._source.current_state_date = params.defaultDate'
+          + '} else {'
+          + 'ctx._source.current_state_date = ctx._source.current_state'
+          + '}'
+          + 'ctx._source.remove("current_state")',
+        params: {
+          emptyCurrentState: {},
```
richard-julien: Should not be as string here? '{}'

opencti-platform/opencti-graphql/src/migrations/1691677872733-sync-from-current-state-to-current-state-date.js
```diff
@@ -0,0 +1,42 @@
+import { elUpdateByQueryForMigration } from '../database/engine';
+import { READ_INDEX_INTERNAL_OBJECTS } from '../database/utils';
+import { DatabaseError } from '../config/errors';
+
+export const up = async (next) => {
+  const defaultDate = new Date();
+  defaultDate.setHours(0, 0, 0, 0);
```
richard-julien: Using now() is not sufficient? Why round on hour?

tomibennett: I guess it is, though we replicated this behavior from the creation process for compliance.

opencti-platform/opencti-graphql/src/migrations/1691677872733-sync-from-current-state-to-current-state-date.js
tomibennett: We tested it against problematic stored document and the migration executed flawlessly. What benefit do you see to make it a string?

SouadHadjiat: no it's an empty object, not a string

opencti-platform/opencti-graphql/src/migrations/1691677872733-sync-from-current-state-to-current-state-date.js
SouadHadjiat: yes we used the same default date as the one used in SyncCreation (`dayStartDate()`)

SarahBocognano (commented):
Tested : ✅ Everything works fine for the ingestion page.

I have just notice some minor stuff : 
- The banner message is not tranlasted
- There is no gap between "stream ID" and "running", I think it's not really pretty (even if it was already the case)
- The button "start" and "stop" does not have the right color 

Capture d’écran (1)
Capture d’écran (2)

Issue #3838, opened 2023-07-24
Add/update Opinion requires 2 clicks to validate


Lhorus6:
Once the Opinion pop-up is open, we need to press the "Update" button twice to validate and close the pop-up window.

PR #4062, opened 2023-08-11
[frontend] Fix double click on add/update opinion in entities

RomuDeuxfois:
### Need help

* I try to fix this issues for the second time. I am not able to reproduce in local, so it's complicated to find the fix. If someone have some ideas to help me on it, I would appreciate.

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/3838

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/analyses/opinions/StixCoreObjectOpinionsRadar.tsx
+++ b/opencti-platform/opencti-front/src/private/components/analyses/opinions/StixCoreObjectOpinionsRadar.tsx
@@ -181,11 +181,10 @@ StixCoreObjectOpinionsRadarProps
           id: alreadyExistingOpinion,
           input: finalValues,
         },
+        updater: () => fetchQuery(),
         onCompleted: () => {
-          fetchQuery();
           setSubmitting(false);
           resetForm();
-          handleClose();
         },
       });
     } else {
@@ -194,11 +193,10 @@ StixCoreObjectOpinionsRadarProps
         variables: {
           input: inputValues,
         },
+        updater: () => fetchQuery(),
         onCompleted: () => {
-          fetchQuery();
           setSubmitting(false);
           resetForm();
-          handleClose();
         },
       });
     }


yassine-ouaamou:
I will try to reproduce

opencti-platform/opencti-front/src/private/components/analyses/opinions/StixCoreObjectOpinionsRadar.tsx
```diff
@@ -182,10 +182,10 @@ StixCoreObjectOpinionsRadarProps
           input: finalValues,
         },
         onCompleted: () => {
```
Kedae, line 185: updater: () => fetchQuery(),
        onCompleted: () => {
          setSubmitting(false);
          resetForm();
        },

opencti-platform/opencti-front/src/private/components/analyses/opinions/StixCoreObjectOpinionsRadar.tsx
```diff
@@ -195,10 +195,10 @@ StixCoreObjectOpinionsRadarProps
           input: inputValues,
         },
         onCompleted: () => {
```
Kedae, line 197: updater: () => fetchQuery(),
        onCompleted: () => {
          setSubmitting(false);
          resetForm();
        },

Kedae (changes requested):

Kedae:
Testable with slow 3G throttling.

The issue is that we fetch the query (i.e try to re-render the component) while mutating the relay store = Kaboom.
Updater function is called when the local relay store HAS been updated, not while completion.

Kedae (dismissed):

yassine-ouaamou:
I still reproduce this bug using the 3G throttling
throttling

SarahBocognano:
Tested: Still happening is Slow 3G, we first have the "create" button appearing, then when clicking, the "update" button finally appears"

SouadHadjiat:
I still have the bug too with "fast 3G" or less. The dialog keeps displaying even though the "open" variable is false. You can't close it with "Cancel" button, you need to update again or change any value. The only way I found to close the dialog is to close it at the beggining of `onSubmit`, because it looks like in `onComplete `or `updater`, we don't really have the same scope anymore, because of the refetch query.

PR #4072, opened 2023-08-14
[frontend] retro compatibility for entities relations histories (#4012)

### LINKED ISSUE #4012: In the history of a sector, the relationship historic is not correct
## Description

image

Originally:

* Only contains create / delete of relationships

Even if we would like to display also the modifications of the relationships:

* Icon is incorrect
* Message is not clear about which relationship has been modified so most of those messages are useless

Archidoit:
issue: https://github.com/OpenCTI-Platform/opencti/issues/4012

Explanation: 
in the history of an entity, the right panel shows the history of the relations of the entity. It should only show creations and deletions (not updates).

New history entities (after june 2023): 'create' or 'delete' is contained in event_scope
old history entities: 'create' or 'delete' is contained in event_type

The modifications here enable to take both new and old history entities into account (retro compatibility)

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectHistory.jsx
+++ b/opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectHistory.jsx
@@ -120,8 +120,12 @@ class StixCoreObjectHistory extends Component {
                     operator: 'wildcard',
                   },
                   {
+                    key: 'event_scope',
+                    values: ['create', 'delete', null], // if event_scope is null, event_type is not
+                  },
+                  {
                     key: 'event_type',
-                    values: ['mutation', 'create', 'update', 'delete', 'merge'],
+                    values: ['create', 'delete', 'mutation'], // retro-compatibility
                   },
                 ],
                 first: 20,


opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectHistory.jsx
```diff
@@ -70,7 +70,7 @@ class StixCoreObjectHistory extends Component {
               filters: [
                 { key: 'entity_id', values: [stixCoreObjectId] },
                 {
-                  key: 'event_type',
+                  key: 'event_scope',
```
SouadHadjiat: I'm not sure to understand why we should use `event_scope `instead of `event_type`. When I request `logs` in the demo graphql playground, `event_scope` is empty, while `event_type` contains the right values. Example with this report : https://demo.opencti.io/dashboard/analyses/reports/3a5e57e9-fdab-4421-b903-c7edeaf627d1

```
query {
  logs(
    first: 10
    filters: [{key: [entity_id], values: ["3a5e57e9-fdab-4421-b903-c7edeaf627d1"]}]
  ) {
    edges {
      node {
        id
        event_type
        event_scope
        entity_type
        timestamp
      }
    }
  }
}
```

Archidoit: When you request the History objects in elastic, you can see that event_type contains a global type (like 'mutation'), and event_scope more precise types (like 'delete', 'update', etc).

Archidoit: image

Archidoit: event_type is updated with the value of event_scope, but not at the moment of the query I modified : 
image
(capture of the resolver log.js)

SouadHadjiat: but `event_scope` has been introduced recenlty, has it been migrated to have a value for older logs ? From what I can see in the demo, `event_scope` is null for old logs, so the filter will get 0 results for this case

Archidoit: ow i'm going to check that

Archidoit: fixed

Archidoit:
PR ready for review

SarahBocognano (commented):
Tested for recent history✅
(I cannot test history from before May 2023 because I didn't have any data stored)

SouadHadjiat (commented):
looks good, tested locally ✅

SouadHadjiat (approved):

PR #4076, opened 2023-08-15
[Backend] Refactor ingestion preparation  to rely only on model definition

richard-julien:
The idea of this refactor is to enforce the model consistency and prevent any data in incorrect format.
Every field will be prepared depending of the attribute definition and the value will be checked/converted accordingly.
A new type 'object' was introduced to have a difference between complex object nature:
- json type: stored in string
- object type: stored in object
- Dictionary type: stored in  object but with content limitation.


opencti-platform/opencti-graphql/src/database/engine.js
```diff
@@ -2083,35 +2084,116 @@ export const elDeleteElements = async (context, user, elements, stixLoadById) =>
...
+      throw UnsupportedError('Invalid data for json preparation', { type: typeof value, attribute: key, value });
+    }
+    return value.trim();
+  }
+  // Come from internal complex object only
+  if (attribute.type === 'object') { // For complex object, prepare inner elements
+    if (!isObject(value)) {
+      throw UnsupportedError('Invalid data type for object preparation', { type: entityType, attribute, value });
+    }
+    return value;
```
SouadHadjiat, line 3032: Where do we prepare the inner elements of the object ?

opencti-platform/opencti-graphql/src/modules/attributes/internalObject-registrationAttributes.ts
```diff
@@ -150,18 +150,18 @@ const internalObjectsAttributes: { [k: string]: Array<AttributeDefinition> } = {
...
-    { name: 'last_execution_date', type: 'string', mandatoryType: 'no', multiple: false, upsert: false },
+    { name: 'task_processed_number', type: 'numeric', mandatoryType: 'no', multiple: false, upsert: false },
+    { name: 'task_expected_number', type: 'numeric', mandatoryType: 'no', multiple: false, upsert: false },
+    { name: 'last_execution_date', type: 'date', mandatoryType: 'no', multiple: false, upsert: false },
     { name: 'completed', type: 'boolean', mandatoryType: 'no', multiple: false, upsert: false },
   ],
   [ENTITY_TYPE_RETENTION_RULE]: [
     { name: 'name', type: 'string', mandatoryType: 'external', multiple: false, upsert: false },
     { name: 'filters', type: 'string', mandatoryType: 'external', multiple: false, upsert: false },
     { name: 'max_retention', type: 'string', mandatoryType: 'external', multiple: false, upsert: false },
```
yassine-ouaamou: max_retention is numeric

opencti-platform/opencti-graphql/src/modules/attributes/internalObject-registrationAttributes.ts
```diff
@@ -109,7 +109,7 @@ const internalObjectsAttributes: { [k: string]: Array<AttributeDefinition> } = {
   ],
   [ENTITY_TYPE_RULE_MANAGER]: [
     { name: 'lastEventId', type: 'string', mandatoryType: 'no', multiple: false, upsert: false },
-    { name: 'errors', type: 'string', mandatoryType: 'no', multiple: false, upsert: false }
+    { name: 'errors', type: 'string', mandatoryType: 'no', multiple: true, upsert: false }
```
SouadHadjiat: Errors looks like an object, here is the mapping I have locally : 

```
"errors": {
          "properties": {
            "message": {
              "type": "text",
              "fields": {
                "keyword": {
                  "type": "keyword",
                  "ignore_above": 512,
                  "normalizer": "string_normalizer"
                }
              }
            },
            "source": {
              "type": "text",
              "fields": {
                "keyword": {
                  "type": "keyword",
                  "ignore_above": 512,
                  "normalizer": "string_normalizer"
                }
              }
            },
            "timestamp": {
              "type": "date"
            }
          }
        }
```

But I can't find in the code where we store the errors

Archidoit:
Need to upgrade the number of the migration

opencti-platform/opencti-graphql/tests/02-integration/02-resolvers/user-test.js
```diff
@@ -443,6 +443,7 @@ describe('User resolver standard behavior', () => {
 describe('User list members query behavior', () => {
   it('Should user lists all members', async () => {
     const queryResult = await editorQuery({ query: LIST_MEMBERS_QUERY, variables: { first: 20 } });
+    console.log(JSON.stringify(queryResult));
```
tomibennett: left debug?

opencti-platform/opencti-graphql/tests/01-unit/database/engine-test.js
```diff
@@ -1,27 +1,64 @@
 import { describe, expect, it } from 'vitest';
-import { prepareElementForIndexing } from '../../../src/database/engine';
+import { prepareDataFromSchemaDefinition } from '../../../src/database/engine';
+import { ENTITY_TYPE_CONTAINER_OPINION, ENTITY_TYPE_INDICATOR } from '../../../src/schema/stixDomainObject';
+import '../../../src/modules/index';
+import { ENTITY_HASHED_OBSERVABLE_ARTIFACT } from '../../../src/schema/stixCyberObservable';
+import { ENTITY_TYPE_CONNECTOR, ENTITY_TYPE_GROUP } from '../../../src/schema/internalObject';
 
 describe('prepareElementForIndexing testing', () => {
```
tomibennett: should be renamed as well?

opencti-platform/opencti-graphql/src/database/engine.js
```diff
@@ -2155,25 +2235,25 @@ const prepareRelation = (thing) => {
...
 const prepareEntity = (thing) => {
-  return R.pipe(R.dissoc(INTERNAL_TO_FIELD), R.dissoc(INTERNAL_FROM_FIELD))(thing);
+  return R.pipe(
+    R.dissoc(INTERNAL_TO_FIELD),
+    R.dissoc(INTERNAL_FROM_FIELD)
+  )(thing);
 };
-const prepareIndexing = async (elements) => {
-  return Promise.all(
-    R.map(async (element) => {
```
tomibennett: out of curiosity, why this closure was async?

opencti-platform/opencti-graphql/src/modules/attributes/internalObject-registrationAttributes.ts
```diff
@@ -109,7 +109,7 @@ const internalObjectsAttributes: { [k: string]: Array<AttributeDefinition> } = {
   ],
```
SouadHadjiat, line 188: Bookmarks field in User type is also an object [{id, type}] : 
{ name: 'bookmarks', type: 'json', mandatoryType: 'no', multiple: true, upsert: false },

opencti-platform/opencti-graphql/src/modules/attributes/internalObject-registrationAttributes.ts
```diff
@@ -170,11 +170,21 @@ const internalObjectsAttributes: { [k: string]: Array<AttributeDefinition> } = {
...
+  [ENTITY_TYPE_HISTORY]: [
+    { name: 'event_type', type: 'string', mandatoryType: 'internal', multiple: false, upsert: false },
+    { name: 'event_scope', type: 'string', mandatoryType: 'internal', multiple: false, upsert: false },
+    { name: 'user_id', type: 'string', mandatoryType: 'internal', multiple: false, upsert: false },
+    { name: 'applicant_id', type: 'string', mandatoryType: 'no', multiple: false, upsert: false },
+    { name: 'group_ids', type: 'string', mandatoryType: 'internal', multiple: true, upsert: false },
+    { name: 'organization_ids', type: 'string', mandatoryType: 'internal', multiple: true, upsert: false },
+    { name: 'timestamp', type: 'date', mandatoryType: 'internal', multiple: false, upsert: false },
+    { name: 'context_data', type: 'dictionary', mandatoryType: 'internal', multiple: false, upsert: false },
+  ]
```
SouadHadjiat: I think there are also `ENTITY_TYPE_WORK`, `ENTITY_TYPE_ACTIVITY`, `ENTITY_TYPE_FEED` that are missing attributes definition

opencti-platform/opencti-graphql/src/database/engine.js
```diff
@@ -2083,35 +2084,116 @@ export const elDeleteElements = async (context, user, elements, stixLoadById) =>
...
+};
+
+/**
+ * prepareElementFromDefinition
+ * @param entityType string
+ * @param attribute AttributeDefinition
+ * @param value object
+ * @returns {object}
+ */
+export const prepareElementFromDefinition = (entityType, attribute, value) => {
```
tomibennett, line 2970: From a design point of view, there is multiple things going on and duplication in this method. 

I suggest separating the checking step that may discard the attribute from the one that prepare its value. Maybe, the first step could be done sooner, when the element is retrieved from the request for instance.

The preparation would then, be a lot ligther and easily integrated into future refinement of the codebase.

opencti-platform/opencti-graphql/src/database/middleware.js
```diff
@@ -886,11 +886,11 @@ export const updatedInputsToData = (instance, inputs) => {
     }
     return { [key]: val };
   }, inputs);
-  return mergeDeepRightAll(...inputPairs);
+  const dataObject = mergeDeepRightAll(...inputPairs);
+  return prepareDataFromSchemaDefinition({ entity_type: instance.entity_type, ...dataObject });
```
tomibennett, line 884: Why is it needed here? Also, why `prepareElementForIndexing` was not?

opencti-platform/opencti-graphql/src/database/engine.js
```diff
@@ -2142,8 +2224,6 @@ const prepareRelation = (thing) => {
   });
   return R.pipe(
     R.assoc('connections', connections),
-    R.dissoc(INTERNAL_TO_FIELD),
-    R.dissoc(INTERNAL_FROM_FIELD),
```
RomuDeuxfois, line 2786: Why removing this part ?

opencti-platform/opencti-graphql/tests/02-integration/02-resolvers/user-test.js
RomuDeuxfois: forgot code?

opencti-platform/opencti-graphql/src/resolvers/stix.js
```diff
@@ -25,6 +26,7 @@ const stixResolvers = {
       /* istanbul ignore next */
       return 'Unknown';
     },
+    spec_version: getSpecVersionOrDefault,
```
RomuDeuxfois, line 29: Should we remove the method in stix Relationship resolver and have a hiearchy between StixRelationship & StixObject ?

opencti-platform/opencti-graphql/src/modules/threatActorIndividual/threatActorIndividual.ts
```diff
@@ -79,6 +80,9 @@ const THREAT_ACTOR_INDIVIDUAL_DEFINITION: ModuleDefinition<StoreEntityThreatActo
     { name: 'primary_motivation', type: 'string', mandatoryType: 'no', multiple: false, upsert: true },
     { name: 'secondary_motivations', type: 'string', mandatoryType: 'no', multiple: true, upsert: true },
     { name: 'personal_motivations', type: 'string', mandatoryType: 'no', multiple: true, upsert: false },
+    { name: 'confidence', type: 'numeric', mandatoryType: 'no', multiple: false, upsert: false },
```
RomuDeuxfois: Already defined for StixDomainObject in stixDomainObject-registrationAttributes

opencti-platform/opencti-graphql/src/modules/threatActorIndividual/threatActorIndividual.ts
```diff
@@ -62,6 +62,7 @@ const THREAT_ACTOR_INDIVIDUAL_DEFINITION: ModuleDefinition<StoreEntityThreatActo
   attributes: [
     { name: 'name', type: 'string', mandatoryType: 'external', multiple: false, upsert: true },
     { name: 'description', type: 'string', mandatoryType: 'customizable', multiple: false, upsert: true },
+    { name: 'aliases', type: 'string', mandatoryType: 'no', multiple: true, upsert: true },
```
RomuDeuxfois: Already added by the module mechanism

  if (definition.type.aliased) {
    attributes.push(...[resolveAliasesField(definition.type.name), iAliasedIds]);
  }

opencti-platform/opencti-graphql/src/domain/stixCoreObject.js
```diff
@@ -422,7 +423,13 @@ export const stixCoreObjectImportDelete = async (context, user, fileId) => {
     await deleteFile(context, user, fileId);
     // Patch the updated_at to force live stream evolution
     const files = (previous.x_opencti_files ?? []).filter((f) => f.id !== fileId);
-    await elUpdateElement({ _index: previous._index, internal_id: entityId, updated_at: now(), x_opencti_files: files });
+    await elUpdateElement({
+      _index: previous._index,
+      entity_type: previous.entity_type,
```
RomuDeuxfois, line 549: Why adding entity_type here ?

opencti-platform/opencti-graphql/src/modules/threatActorIndividual/threatActorIndividual.ts
RomuDeuxfois: Add  aliased: true in contract

opencti-platform/opencti-graphql/src/database/engine.js
SouadHadjiat: I think it's because `prepareRelation` is called only in `prepareElementIndexing` and inside we already call `prepareEntity` first which does this same operation

opencti-platform/opencti-graphql/src/domain/stixCoreObject.js
SouadHadjiat: I think it's because we need `entity_type` in `prepareDataFromSchemaDefinition`

opencti-platform/opencti-graphql/src/database/engine.js
richard-julien: I remove that part for now has inner object elements are not defined by schema yet.

opencti-platform/opencti-graphql/src/modules/attributes/internalObject-registrationAttributes.ts
richard-julien: Is id,type but stored in string, so json type for me

opencti-platform/opencti-graphql/src/resolvers/stix.js
richard-julien: This should be removed everywhere. But this work will also modify the python lib, for now  i remove the attribute of the database and put a dynamic resolution.

opencti-platform/opencti-graphql/src/database/engine.js
richard-julien: True. I think we have to work to prepare the data is expected format and check sooner in the chain. But the target here is to introduce a defensive code to prevent any bad behavior waiting for a better structural approach

opencti-platform/opencti-graphql/src/database/middleware.js
richard-julien: At update we dynamically apply inputs on already fetched elements to not fetch again after save and so gain performance. As inputs are not in the correct type, the instance need to be data prepared before sending back to graphql that have the correct type definition and so fail if the type are not aligned.

opencti-platform/opencti-graphql/src/database/engine.js
richard-julien: No reason i think.

opencti-platform/opencti-graphql/src/modules/threatActorIndividual/threatActorIndividual.ts
richard-julien: Interesting subject. I think I will add a new control code in the registerAttributes function to throw exception when we try to add an attribute already defined by a parent

opencti-platform/opencti-graphql/src/database/middleware.js
tomibennett: 🙏

opencti-platform/opencti-graphql/src/modules/attributes/internalObject-registrationAttributes.ts
SouadHadjiat: I don't see it stored as String, here is the mapping I have locally : 

```
"bookmarks": {
  "properties": {
    "id": {
      "type": "text",
          "fields": {
        "keyword": {
          "type": "keyword",
              "ignore_above": 512,
              "normalizer": "string_normalizer"
        }
      }
    },
    "type": {
      "type": "text",
          "fields": {
        "keyword": {
          "type": "keyword",
              "ignore_above": 512,
              "normalizer": "string_normalizer"
        }
      }
    }
  }
},
```

And here is an example where it is stored :  

```
"bookmarks": [
  {
    "id": "b446835d-66f5-4a46-b57b-de11de93a68a",
    "type": "Threat-Actor-Individual"
  },
  {
    "id": "fe16663c-6b94-4857-9640-e53573560f02",
    "type": "Intrusion-Set"
  },
  {
    "id": "1ee7de1c-c6bc-48e5-9766-4de1f2f59f8b",
    "type": "Threat-Actor-Group"
  }
]

```

opencti-platform/opencti-graphql/src/modules/attributes/internalObject-registrationAttributes.ts
richard-julien: Yes thanks. Its done in taskManager.js with script function.

`source += `ctx._source.errors.add(["timestamp": params.received_time, "id": "${error.id}", "message": "${error.message}"]); `;`

I need to change the definition

richard-julien:
Closing, was covered in another PR/task

PR #4080, opened 2023-08-16
[frontend] Handle null scope on task

RomuDeuxfois:
### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/4079

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/data/tasks/TasksList.jsx
+++ b/opencti-platform/opencti-front/src/private/components/data/tasks/TasksList.jsx
@@ -30,7 +30,6 @@ import { commitMutation, MESSAGING$ } from '../../../../relay/environment';
 import Security from '../../../../utils/Security';
 import { KNOWLEDGE_KNUPDATE_KNDELETE } from '../../../../utils/hooks/useGranted';
 import TaskScope from '../../../../components/TaskScope';
-import computeLabel from './TasksUtils';
 
 const interval$ = interval(FIVE_SECONDS);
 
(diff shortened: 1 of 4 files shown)


Archidoit:
Tested and works well
But maybe you can change the color for scope Rule in the Background tasks list, because grey is not really lisible and is rather a color for errors/uncompleted fields etc
image

opencti-platform/opencti-front/src/private/components/data/tasks/TasksList.jsx
```diff
@@ -373,12 +374,14 @@ class TasksListComponent extends Component {
...
-                        {t('Scope')}
-                      </Typography>
-                      <TaskScope scope={task.scope} label={t(task.scope)} />
-                    </Grid>
+                    {computeLabel(task.type, task.scope)
+                      && <Grid item={true} xs={2}>
+                        <Typography variant="h3" gutterBottom={true}>
+                          {t('Scope')}
+                        </Typography>
+                        <TaskScope scope={task.scope} label={t(computeLabel(task.type, task.scope))} />
```
yassine-ouaamou: can't we just do something like: t(task.scope ?? task.type) instead of creating a function just for that?

RomuDeuxfois: I prefer to separate the logic from the purely UI part. But, we can discuss it.

SarahBocognano (approved):
Tested ✅

Archidoit (approved):

PR #4081, opened 2023-08-16
[backend] Added Financial Accounts and Assets

Bonsai8863:
### Proposed changes

* Added Financial Data category to track things like financial accounts and assets.
* Added relationships and backend support for these new data.

### Checklist

- [x] Localization


richard-julien:
After first quick review, some main points to discuss.
- Ok to modelized Account as a model entity.
- Not Ok to have an Asset entity, I strongly think it should be a **observable** "financial account" that will replace BankAccount and CryptoWallet (types need to be migrated)
- Not Ok with elUpdateMany approach. Looks like not used at all so should be simply removed from the base code.
- financial_account_balances must be changed to use the same approach than Height and Weight.

As the modifications impact the model, i think we need to discuss directly about it.

PR #4094, opened 2023-08-18
[backend/frontend] Fix available relationships for Administrative-Area

### LINKED ISSUE #4092: No relationship type available when creating a relationship from an Area
No relationship type available when creating a relationship from an Area

## Steps to reproduce
Go to an Area. Go to the Knowledge tab.
Select 'Incidents' in the right tab.
Select an incident.
You can't create the relationship because there is no relationship type available.

image

It is the same from relationship from Area to other entity types (Threat actor, campaigns etc)

RomuDeuxfois:
### Proposed changes

* Fix pain point : https://www.notion.so/filigran/Technical-Process-pains-65d4613846a7499392f748bd73087a30?pvs=4#601d33abdab9451dbaa9ea7860f32981

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/4092

### Checklist

### Further comments


SarahBocognano (commented):
Tested ✅  Everything works fine. It's good to remodel the relation.js file like that !

opencti-platform/opencti-front/src/private/Root.tsx
```diff
@@ -97,6 +97,10 @@ const rootPrivateQuery = graphql`
         }
       }
     }
+    schemaRelationsTypesMapping: schemaRelationsTypesMapping {
```
SouadHadjiat: Since it's the same name, could we just have `schemaRelationsTypesMapping {` ?

opencti-platform/opencti-front/src/private/Root.tsx
```diff
@@ -128,11 +132,12 @@ interface RootComponentProps {
 
 const RootComponent: FunctionComponent<RootComponentProps> = ({ queryRef }) => {
   const queryData = usePreloadedQuery(rootPrivateQuery, queryRef);
-  const { me, settings, entitySettings, schemaSCOs, schemaSDOs, schemaSROs } = queryData;
+  const { me, settings, entitySettings, schemaSCOs, schemaSDOs, schemaSROs, schemaRelationsTypesMapping } = queryData;
   const schema = {
     scos: schemaSCOs.edges.map((sco) => sco.node),
     sdos: schemaSDOs.edges.map((sco) => sco.node),
     sros: schemaSROs.edges.map((sco) => sco.node),
+    schemaRelationsTypesMapping,
```
SouadHadjiat: Could it be a map instead of a list ? like this :  
`new Map(schemaRelationsTypesMapping.map((n) => [n.key, n.values]));`
So that in `resolveRelationsTypes`, we could direcly get the values by key

opencti-platform/opencti-graphql/src/domain/stixRelationship.js
```diff
@@ -18,3 +18,22 @@ export const stixRelationshipDelete = async (context, user, stixRelationshipId)
 };
 
 export const getSpecVersionOrDefault = ({ spec_version }) => spec_version ?? STIX_SPEC_VERSION;
+
+export const schemaRelationsTypesMapping = (entityTypes = []) => {
+  const entries = Object.entries(stixCoreRelationshipsMapping);
+  let filterEntries;
+  if (entityTypes.length > 0) {
```
SouadHadjiat: where do we call this method with entityTypes argument ?

opencti-platform/opencti-front/src/private/Root.tsx
RomuDeuxfois: well seen

opencti-platform/opencti-graphql/src/domain/stixRelationship.js
RomuDeuxfois: Never, it's just to prepare the next if we need it
But, it's maybe overkill

RomuDeuxfois:
PR comments fixed

opencti-platform/opencti-front/src/private/components/common/stix_core_relationships/StixCoreRelationshipCreation.jsx
```diff
@@ -351,37 +352,44 @@ class StixCoreRelationshipCreation extends Component {
...
+                      defaultConfidence={confidence}
+                      defaultStartTime={startTime}
+                      defaultStopTime={stopTime}
+                      defaultCreatedBy={defaultCreatedBy}
+                      defaultMarkingDefinitions={defaultMarkingDefinitions}
+                  />
+                </>
+            );
+          }}
+        </UserContext.Consumer>
```
tomibennett, line 392: For curiosity, can you tell me what is `UserContext.Consumer` tag, and why embedding the rest of the code in it?

tomibennett:
I don't know if there is a strong indent rule but some of your modifications have 4 spaces instead of 2. That may be something to take care of.

opencti-platform/opencti-front/src/private/components/common/stix_core_relationships/StixCoreRelationshipCreation.jsx
RomuDeuxfois: This allows the application context to be used in React classes. It's the old way before hooks.

https://legacy.reactjs.org/docs/context.html

PR #4096, opened 2023-08-18
[frontend] Entity details in correlation views (#4065)

### LINKED ISSUE #4065: Ability to select objects in the Report's Correlation View
## Use case

Correlation view is very useful. Once identifying correlated objects, user should be able to view and/or open that object. Instead the user must search for that object  in the search bar

## Current Workaround

Search for the correlated object in the search bar

## Proposed Solution

Provide a link / method to open objects presented in the Correlation view. The same as the function currently available in the standard Knowledge Graph view

Archidoit:
issue: https://github.com/OpenCTI-Platform/opencti/issues/4065

## PR content
In correlation views (of Grouping, Report, RFT, RFI and Incidents) :
- add the possibility to view the entities details when clicking on an object.
- possibility to view a basic-relationship details when clicking on an arrow
- disable basic relationship edition button
- fix bottom bar design (no more overlaps on the left)

## Comment
If you can't populate the correlation view, it's normal, the view is not working well... It may be hard to test if you can't populate, the issue will be resolved later.


opencti-platform/opencti-front/src/private/components/analyses/reports/ReportKnowledgeGraphBar.jsx
```diff
@@ -385,6 +385,7 @@ class ReportKnowledgeGraphBar extends Component {
         && numberOfSelectedNodes === 0
         && numberOfSelectedLinks === 1
         && selectedLinks.length === 1
+        && selectedLinks[0].entity_type !== 'basic-relationship'
```
RomuDeuxfois, line 388: Only in correlation view

Archidoit: the KnowledgeGraphBar is used for all the views in the Report Knowledge tab... so also for the Correlation view

SarahBocognano:
Still impossible to test locally on my part :(

yassine-ouaamou:
All good for me! 
Able to view details of the entities and the relationships 👍 
No edition possible 👍 
No overlap 👍

yassine-ouaamou (approved):

PR #4134, opened 2023-08-22
[frontend] fix relationship stop_time edition (#4133)

### LINKED ISSUE #4133: Can't update relationship stop_time
Go to a relationship.
Edit the stop_time.
The update is not taken into account.

Archidoit:
Github issue: https://github.com/OpenCTI-Platform/opencti/issues/4133

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/common/stix_core_relationships/StixCoreRelationshipEditionOverview.tsx
+++ b/opencti-platform/opencti-front/src/private/components/common/stix_core_relationships/StixCoreRelationshipEditionOverview.tsx
@@ -251,7 +251,7 @@ Omit<StixCoreRelationshipEditionOverviewProps, 'queryRef'>
       .nullable(),
     stop_time: Yup.date()
       .typeError(t('The value must be a datetime (yyyy-MM-dd hh:mm (a|p)m)'))
-      .min(Yup.ref('stop_time'), "The end date can't be before start date")
+      .min(Yup.ref('start_time'), "The end date can't be before start date")
       .nullable(),
     description: Yup.string().nullable(),
     references: Yup.array(),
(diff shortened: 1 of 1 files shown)


opencti-platform/opencti-front/src/private/components/common/stix_core_relationships/StixCoreRelationshipEditionOverview.tsx
```diff
@@ -275,6 +275,18 @@ Omit<StixCoreRelationshipEditionOverviewProps, 'queryRef'>
     stixCoreRelationshipValidator,
   );
 
+  // necessary for stop_time because the validator includes a reference to another value (start_time)
```
tomibennett, line 278: If the validator refers to the wrong value, should it be fixed?

Archidoit: it's not a wrong value, you need it to compare start_time and stop_time

opencti-platform/opencti-front/src/private/components/common/stix_core_relationships/StixCoreRelationshipEditionOverview.tsx
```diff
@@ -383,7 +395,7 @@ Omit<StixCoreRelationshipEditionOverviewProps, 'queryRef'>
                 component={DateTimePickerField}
                 name="stop_time"
                 onFocus={editor.changeFocus}
-                onSubmit={editor.changeField}
+                onSubmit={handleSubmitFieldStopTime}
```
SouadHadjiat, line 398: why doesn't it work like the other fields ? why did you need to add `handleSubmitFieldStopTime` ?

Archidoit: the other fields don't call an other field for their validation

Archidoit: the origin of the problem is the same as for an old issue whose brainstorm is here : https://www.notion.so/filigran/Can-t-update-end-date-of-an-event-3033-ae8c2ab25c564973b974dfb3c980c90d#3e295d32596d4da391830a004964244e

SarahBocognano (approved):
Tested ✅

PR #4154, opened 2023-08-24
[backend] Improve stix loader and stream relation resolution

### LINKED ISSUE #4087: Removing Sharing Organization removes relationships
## Description

There's something weird going on when streaming data with a user that has an organization attached to it (let's say OrgX in this case) and there's data that has has been shared with OrgX.

A relationship link between two things disappears in a certain condition. It cannot be found with the internal / standard ids after it has disappeared. 

## Environment

1. OS (where OpenCTI server runs): Debian 10 (docker)
2. OpenCTI version: 5.9.6
3. OpenCTI client: frontend
4. Other environment details:

## Reproducible Steps

1. Create an organization OrgX
2. Add a user with streaming capabilities and add it to OrgX
3. Create an observable, indicator and a relation (easiest to add e.g. a domain-name and then enable "Create an indicator from this observable")
4. Share all 3 to OrgX ("Share with an organization" button). You should see all 3 in the stream now.
5. Remove sharing to OrgX from the indicator. After this point the relationship should still exist.
6. Stream the data using the abovementioned user.
7. After this the relationship has disappeared from the system

## Expected Output

Removing the sharing should not delete the relationship between the two objects.

## Actual Output

The relationship has been removed. It cannot be found in the system with the internal ID or standard ID.

## Additional information

Streaming at step 6. as an admin or another user, which has not OrgX affiliations does not have the effect and the relationship remains.

I tried searching the history index for deletions but it shows just some old stuff.

richard-julien:
There is conditions where an auto sanity mode think that the relation should be deleted. 
Its due to our process that doesnt not use system user to resolve to be sure about the situation.
Also improve the stream to not publish the relation if one of the from/to is not accessible.

### DIFF
--- a/opencti-platform/opencti-graphql/src/database/middleware.js
+++ b/opencti-platform/opencti-graphql/src/database/middleware.js
@@ -472,7 +472,7 @@ const loadElementsWithDependencies = async (context, user, elements, opts = {}) 
         const message = `From ${element.fromId} is ${validFrom}, To ${element.toId} is ${validTo}`;
         logApp.warn(`Auto delete of invalid relation ${element.id}. ${message}`);
         // Auto deletion of the invalid relation
-        // await elDeleteElements(context, SYSTEM_USER, [element], storeLoadByIdWithRefs);
+        await elDeleteElements(context, SYSTEM_USER, [element], storeLoadByIdWithRefs);
       } else {
         const from = R.mergeRight(element, { ...rawFrom, ...depsElementsMap.get(element.fromId) });
         const to = R.mergeRight(element, { ...rawTo, ...depsElementsMap.get(element.toId) });
--- a/opencti-platform/opencti-graphql/src/database/middleware.js
+++ b/opencti-platform/opencti-graphql/src/database/middleware.js
@@ -453,7 +453,8 @@ const loadElementsWithDependencies = async (context, user, elements, opts = {}) 
   const depsPromise = loadElementMetaDependencies(context, user, elementsToDeps, opts);
   if (targetsToResolved.length > 0) {
     const args = { toMap: true, connectionFormat: false };
-    fromToPromise = elFindByIds(context, user, targetsToResolved, args);
+    // Load with System user, access rights will be dynamically change after
+    fromToPromise = elFindByIds(context, SYSTEM_USER, targetsToResolved, args);
   }
   const [fromToMap, depsElementsMap] = await Promise.all([fromToPromise, depsPromise]);
   const loadedElements = [];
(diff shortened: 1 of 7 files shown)


opencti-platform/opencti-graphql/src/graphql/sseMiddleware.js
```diff
@@ -350,10 +355,18 @@ const createSseMiddleware = () => {
...
-    }).flat();
+    const missingAllPerIds = missingInstances.map((m) => [m.internal_id, m.standard_id, ...(m.x_opencti_stix_ids ?? [])].map((id) => ({ id, value: m }))).flat();
     const missingMap = new Map(missingAllPerIds.map((m) => [m.id, m.value]));
+    if (stixData.type === STIX_TYPE_RELATION || stixData.type === STIX_TYPE_SIGHTING) {
+      // Check for a relation that the from and the to is correctly accessible.
+      const fromId = stixData.source_ref ?? stixData.sighting_of_ref;
+      const toId = stixData.target_ref ?? stixData.where_sighted_refs[0];
+      const hasFrom = missingMap.has(fromId) || cache.has(fromId);
+      const hasTo = missingMap.has(toId) || cache.has(toId);
+      if (!hasFrom || !hasTo) {
```
RomuDeuxfois, line 366: I am not sur this is exactly what we want
If there are in a platform
- a sector : marking RED
- a sector : marking GREEN
- a relationship between this two sectors : marking GREEN
- user A : marking GREEN

If user A modify the relationship, the modification is not displayed in the stream

opencti-platform/opencti-graphql/src/database/middleware.js
```diff
@@ -471,7 +472,7 @@ const loadElementsWithDependencies = async (context, user, elements, opts = {})
         const message = `From ${element.fromId} is ${validFrom}, To ${element.toId} is ${validTo}`;
         logApp.warn(`Auto delete of invalid relation ${element.id}. ${message}`);
         // Auto deletion of the invalid relation
-        await elDeleteElements(context, SYSTEM_USER, [element], storeLoadByIdWithRefs);
+        // await elDeleteElements(context, SYSTEM_USER, [element], storeLoadByIdWithRefs);
```
RomuDeuxfois: Why removing this ?
There are some side effects ?

opencti-platform/opencti-graphql/src/graphql/sseMiddleware.js
richard-julien: Its exactly what is expected. If both part of the relation is not accessible the relation must not be published.

opencti-platform/opencti-graphql/src/database/middleware.js
richard-julien: Just forgot to remove the comment after some testing

PR #4159, opened 2023-08-24
[backend] Improve objects resolutions for container (#4153)

### LINKED ISSUE #4153: Creating a note on observed data crashes the website
## Description

When creating a note on observed datas, the observed data loose its name value and crash the whole tab.

## Environment

1. OS (where OpenCTI server runs): Demo and Kubernetes v1.27
2. OpenCTI version: OpenCTI 5.9.6
3. OpenCTI client: frontend
4. Other environment details:

## Reproducible Steps

1. Go to observed data
2. Create a note on the observed data
3. Refresh the page

## Expected Output

Seeing the note on the observed data like so : 
Capture d'écran 2023-08-24 083441

## Actual Output

The whole tab of observed data crashes :  
Capture d'écran 2023-08-24 083711

The whole data/entities tab also crashes:
Capture d'écran 2023-08-24 083800

## Error output:
```
["Error: Cannot return null for non-nullable field ObservedData.name."
````

## Additional information

* Only solution so far is to delete the note, wait for the background task to end, and then refresh the page.
* Creating a note on other SDO doesn't create such a bug

### PR DESCRIPTION
**Pull request by richard-julien**

### CODE REVIEW

**Comment by SouadHadjiat** [comment]
When I create an ObservedData from a domain name, it has the name of the domain name, then I add a note, there is no error anymore but the ObservedData name is now "empty".

Capture d'écran 2023-08-24 213228

===> update
I began to debug `objects` method which returns an empty list, and inside `relations` returns only one relationship with the domain name, and `elements` returns the note only. I guess the method should return both ?
=> it returns only one element because `first` argument is 1 in `resolveName`. So we only query 1 relationship and 1 element, but both are not the same.

File: opencti-platform/opencti-graphql/src/domain/container.js [hunk 1]
```diff
@@ -52,27 +48,37 @@ export const numberOfContainersForObject = (context, user, args) => {
...
-  const key = buildRefRelationKey(RELATION_OBJECT, '*');
   const types = args.types ? args.types : ['Stix-Core-Object', 'stix-relationship'];
+  const relationArgs = { fromId: containerId, toTypes: types, baseData: true };
+  const objectRelationsPromise = listAllRelations(context, user, RELATION_OBJECT, relationArgs);
+  const key = buildRefRelationKey(RELATION_OBJECT, '*');
   const filters = [{ key, values: [containerId], operator: 'wildcard' }, ...(args.filters || [])];
-  const data = args.all ? await paginateAllThings(context, user, types, R.assoc('filters', filters, args))
-    : await listThings(context, user, types, R.assoc('filters', filters, args));
+  const queryFilters = { ...args, filters, connectionFormat: false };
+  const elementsPromise = args.all ? listAllThings(context, user, types, queryFilters) : listThings(context, user, types, queryFilters);
```
RomuDeuxfois, line 57: Why are we retrieve elements from another query rather than retrieve them from the relations ?
We can have const toIds = relations.map((rel) => rel.toId) and resolve toIds, no ?

RomuDeuxfois:
> When I create an ObservedData from a domain name, it has the name of the domain name, then I add a note, there is no error anymore but the ObservedData name is now "empty".
> 
> Capture d'écran 2023-08-24 213228
> 
> ===> update I began to debug `objects` method which returns an empty list, and inside `relations` returns only one relationship with the domain name, and `elements` returns the note only. I guess the method should return both ? => it returns only one element because `first` argument is 1 in `resolveName`. So we only query 1 relationship and 1 element, but both are not the same.

I test this case and it's fixed now

RomuDeuxfois:
Tested 🆗

opencti-platform/opencti-graphql/src/domain/container.js
richard-julien: Its a general problem of the current model. In order to support case of pagination/sorting/filtering we use the denormalized id of elements stored in rel_. Unfortunalty we miss the fact that storing also the direction was needed and so when we list through that part we have sometime to much elements.

RomuDeuxfois (approved):

PR #4187, opened 2023-08-28
[frontend] Create DrawerOpenCTI component

### LINKED ISSUE #4011: When having a platform with header, creating live trigger / digest, padding top is too low
## Use case

image

When a banner in security policies in set, right menu in container => content is not correct:

Padding top is too low:

image

RomuDeuxfois:
### Proposed changes

* Create an OpenCTI Drawer component
* Use it in 3 components to confirm the usage

### Next steps

* Use it in all the necessary forms in order to remove a lot of code and homogenize the UI
* Change this style in the DrawerOpenCTI component
```
container: {
      padding: '10px 20px 20px 20px',
    }
```
```
padding: `${bannerHeightNumber}px`
```
* Remove spacing in all forms using the DrawerOpenCTI component
~~<Form style={{ margin: '20px 0 20px 0' }}>~~
* Add style in StixDomainObjectCreation
```
<Select
          value={type}
          onChange={(event) => setType(event.target.value)}
          fullWidth={true}
          size="small"
          **style={{ marginBottom: '20px' }}**
        >
```

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/4011

### Checklist

### Further comments

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/common/drawer/DrawerOpenCTI.tsx
+++ b/opencti-platform/opencti-front/src/private/components/common/drawer/DrawerOpenCTI.tsx
@@ -49,6 +49,7 @@ const DrawerOpenCTI: FunctionComponent<DrawerOpenCTIProps> = ({
 }) => {
   const classes = useStyles();
 
+  const handleClose = () => setOpen(false);
   return (
     <Drawer
       open={open}
@@ -56,12 +57,12 @@ const DrawerOpenCTI: FunctionComponent<DrawerOpenCTIProps> = ({
       elevation={1}
       sx={{ zIndex: 1202 }}
       classes={{ paper: classes.drawerPaper }}
-      onClose={() => setOpen(false)}
+      onClose={handleClose}
     >
       <div className={classes.header} >
         <IconButton
           aria-label="Close"
-          onClick={() => setOpen(false)}
+          onClick={handleClose}
           size="large"
           color="primary"
         >
--- a/opencti-platform/opencti-front/src/private/components/common/drawer/DrawerOpenCTI.tsx
+++ b/opencti-platform/opencti-front/src/private/components/common/drawer/DrawerOpenCTI.tsx
@@ -1,0 +1,79 @@
+import React, { FunctionComponent } from 'react';
+import Drawer from '@mui/material/Drawer';
+import makeStyles from '@mui/styles/makeStyles';
+import IconButton from '@mui/material/IconButton';
+import { Close } from '@mui/icons-material';
+import Typography from '@mui/material/Typography';
+import { Theme } from '../../../../components/Theme';
+import useAuth from '../../../../utils/hooks/useAuth';
+
+const useStyles = makeStyles<Theme>((theme) => {
+  const {
+    bannerSettings: { bannerHeightNumber },
+  } = useAuth();
+  return ({
+    drawerPaper: {
+      minHeight: '100vh',
+      width: '50%',
+      position: 'fixed',
+      overflow: 'auto',
+      transition: theme.transitions.create('width', {
+        easing: theme.transitions.easing.sharp, duration: theme.transitions.duration.enteringScreen,
+      }),
+      paddingTop: `${bannerHeightNumber}px`,
+    },
+    header: {
+      backgroundColor: theme.palette.background.nav,
+      padding: '10px 0',
+      display: 'inline-flex',
+      alignItems: 'center',
+    },
+    container: {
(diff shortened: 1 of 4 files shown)


SarahBocognano (commented):
I have a question: Why did you change the Drawer by the DrawerOpenCti only in Reports/Notification/Malwares ? :)

RomuDeuxfois:
> I have a question: Why did you change the Drawer by the DrawerOpenCti only in Reports/Notification/Malwares ? :)

To not have a huge PR at first. It's better to verify if everyone is agreed with a tiny code :)

SarahBocognano (commented):
Tested on Report/Malware/Notifier Creation and works well ! ✅

opencti-platform/opencti-front/src/private/components/common/drawer/DrawerOpenCTI.tsx
```diff
@@ -0,0 +1,79 @@
+import React, { FunctionComponent } from 'react';
+import Drawer from '@mui/material/Drawer';
+import makeStyles from '@mui/styles/makeStyles';
```
marieflorescontact, line 3: this is an excellent inititative!!!! 
makstyle seems to be [deprecated](https://mui.com/system/styles/basics/) (incompatible with React 18).
I don't fully understand the documentation and can't say whether it's very useful, but there seems to be a topic on the use of `makestyle` hook.

Kedae: Yep, deprecated, but we are not on React 18 yet and not prepared to make the jump. 
That being said, we could try to use the correct method from now on : import { styled } from "@mui/system";

opencti-platform/opencti-front/src/private/components/common/drawer/DrawerOpenCTI.tsx
```diff
@@ -0,0 +1,79 @@
...
+  const classes = useStyles();
+
+  return (
+    <Drawer
+      open={open}
+      anchor="right"
+      elevation={1}
+      sx={{ zIndex: 1202 }}
+      classes={{ paper: classes.drawerPaper }}
+      onClose={() => setOpen(false)}
```
Kedae: This should be a handleClose

opencti-platform/opencti-front/src/private/components/common/drawer/DrawerOpenCTI.tsx
```diff
@@ -0,0 +1,79 @@
...
+});
+
+interface DrawerOpenCTIProps {
+  title: string
+  open: boolean
+  setOpen: (value: boolean) => void
+  children: React.ReactNode
+}
+
+const DrawerOpenCTI: FunctionComponent<DrawerOpenCTIProps> = ({
```
Kedae, line 44: How costly is it to add the <Fab> inside the drawer ?

Kedae: As discussed with Romuald today, we will add the Fab with a variant inside the Drawer. The only specific case is TriggerCreation that will not have the new custom drawer.

opencti-platform/opencti-front/src/private/components/common/drawer/DrawerOpenCTI.tsx
RomuDeuxfois: https://stackoverflow.com/questions/71948818/no-result-using-makestyles-material-ui-in-react-18

RomuDeuxfois: I made an attempt and it works very well. But there is a paradigm shift to be made and we are not ready to do it as it stands.
There are a snippet in the page named process pain (notion)

opencti-platform/opencti-front/src/private/components/common/drawer/DrawerOpenCTI.tsx
RomuDeuxfois: Is not possible for now :
Style in ReportCreation ->
```
 createButton: {
    position: 'fixed',
    bottom: 30,
    right: 30,
  },

```
Style in NotifierCreation -> 
```
createButton: {
    position: 'fixed',
    bottom: 30,
    right: 230,
  },
```


Because of the right bar
image

And no fab in notifierEdition
image

RomuDeuxfois:
All comments have been taken into account.
Ready for a second round

PR #4192, opened 2023-08-29
[frontend/backend] Notifiers edition validation (#4179bis)

### LINKED ISSUE #4179: Unable to modify notifiers in Settings => Customizations => Notifiers
## Description

Go to the section, try to change the name of a notifier, click on save. Nothing happen.

Archidoit:
Go to Customization > Notifiers.
Edit a notifier by removing mandatory fields (ex: in a notifier with Connector = Simple Mailer, remove the field 'title')
The edition is taken into account.
If we test the notifier, an error occurs.

Solution: add some validation checking for notifiers edition

PR content:
- notifier edition validation
- notifier test validation
- correct bug at Simple mail testing
- reset test results in notifier test (front)


helene-nguyen:
Tested locally, required field already empty **before** changes applied still send notification

notifier

But when update the notifier and fill the empty field again then re-update, it works => the field is required

Archidoit:
@helene-nguyen since it is possible to remove mandatory fields of a notifier only after a commit pushed yesterday, this bug is not in the previous release and the clients can't have such empty fields. So I don't think it is necessary to do a migration.

opencti-platform/opencti-front/src/private/components/settings/notifiers/NotifierEdition.tsx
```diff
@@ -126,19 +127,37 @@ const NotifierEdition: FunctionComponent<NotifierEditionComponentProps> = ({
...
-    }
+  const submitForm = (
+    setSubmitting: FormikHelpers<NotifierEditionValues>['setSubmitting'],
+    setErrors: FormikHelpers<NotifierEditionValues>['setErrors'],
+    values: NotifierEditionValues,
+    current: FormRef,
+  ) => {
+    notifierValidation(t)
+      .validate(values)
+      .then(() => {
```
RomuDeuxfois, line 138: You shouldn't have its lines since the validator is declared on the formik side
validationSchema={notifierValidation(t)}
You need to disabled the button if the form is not valid

opencti-platform/opencti-front/src/private/components/settings/notifiers/NotifierEdition.tsx
```diff
@@ -167,7 +186,7 @@ const NotifierEdition: FunctionComponent<NotifierEditionComponentProps> = ({
             onSubmit={() => {}}
             onClose={onClose}
           >
-            {({ values, setFieldValue, setSubmitting, isSubmitting }) => (
+            {({ values, setFieldValue, setSubmitting, setErrors, isSubmitting }) => (
               <Form style={{ margin: '20px 0 20px 0' }}>
```
RomuDeuxfois, line 190:
```suggestion
            {({ values, setFieldValue, setSubmitting, setErrors, isSubmitting, isValid, dirty }) => (
              <Form style={{ margin: '20px 0 20px 0' }}>
```

opencti-platform/opencti-front/src/private/components/settings/notifiers/NotifierEdition.tsx
```diff
@@ -227,8 +246,7 @@ const NotifierEdition: FunctionComponent<NotifierEditionComponentProps> = ({
                     variant="contained"
                     color="secondary"
                     onClick={() => {
-                      submitForm(setSubmitting, values, formRef.current);
-                      onClose();
+                      submitForm(setSubmitting, setErrors, values, formRef.current);
                     }}
                     disabled={isSubmitting}
```
RomuDeuxfois, line 251:
```suggestion
                    disabled={isSubmitting || !isValid || !dirty}
```

opencti-platform/opencti-front/src/private/components/settings/notifiers/NotifierEdition.tsx
RomuDeuxfois: Same in nofifierCreation file

opencti-platform/opencti-graphql/src/modules/notifier/notifier-domain.ts
```diff
@@ -55,6 +55,12 @@ export const notifierGet = (context: AuthContext, user: AuthUser, notifierId: st
 };
 
 export const notifierEdit = async (context: AuthContext, user: AuthUser, notifierId: string, input: EditInput[]) => {
+  const fieldsToValidate = {
+    name: '',
```
RomuDeuxfois: Why do you need the name property ?

Archidoit: it's for types reasons: validateNotifier input should have a 'name'

SarahBocognano (commented):
I don't understand if it's on my side or not but when I empty the field, the test doesn't show an error, but if I click on "test", the app crash

opencti-platform/opencti-front/src/private/components/settings/notifiers/NotifierEdition.tsx
Archidoit: It's not what is done in the platform... If we want to create/edit an entity without entering the mandatory fields, the 'create' button is not disabled but we have a message in red

Archidoit: image

Archidoit:
> I don't understand if it's on my side or not but when I empty the field, the test doesn't show an error, but if I click on "test", the app crash


@SarahBocognano the test doesn't show error (normal, it's a choice), but if you click on 'create' (or 'save' for edition), a message error is indicating.
For the app crash when we do a test, I'm going to have a look.

opencti-platform/opencti-front/src/private/components/settings/notifiers/NotifierEdition.tsx
RomuDeuxfois: This is indeed the case for creation (my mistake), but for editing has started to be implemented.
Example in ReportEditionOverview file.

opencti-platform/opencti-graphql/src/modules/notifier/notifier-domain.ts
RomuDeuxfois: I am not really comfortable to add some unused property to satisfy a method contract.
It's better to change the method contract by saying
`validateNotifier = (notifier: {notifier_connector_id: string, notifier_configuration: string} )`

opencti-platform/opencti-front/src/private/components/settings/notifiers/NotifierEdition.tsx
Archidoit: yes but the notifier editor is constructed like a creation form: there is a Cancel and a Save button. Whereas entities edition forms are constructed without such submission button (update is done at field changing).

opencti-platform/opencti-graphql/src/modules/notifier/notifier-domain.ts
Archidoit: right! fixed :)

opencti-platform/opencti-front/src/private/components/settings/notifiers/NotifierEdition.tsx
Archidoit: seen with Romu

Archidoit:
Remaining bugs:

- problem at a simple mailer notifier testing:
image

- when clicking on 'Test' the initial test result displayed is the last test run

Archidoit:
All the bugs and comment have been fixed :)

RomuDeuxfois (approved):

SarahBocognano (approved):

PR #4198, opened 2023-08-30
[frontend] Rework radar opinions component (#3838)

### LINKED ISSUE #3838: Add/update Opinion requires 2 clicks to validate

RomuDeuxfois:
### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/3838

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/analyses/opinions/StixCoreObjectOpinions.tsx
+++ b/opencti-platform/opencti-front/src/private/components/analyses/opinions/StixCoreObjectOpinions.tsx
@@ -1,7 +1,8 @@
 import React, { FunctionComponent, useCallback, useEffect } from 'react';
-import { useQueryLoader } from 'react-relay';
+import { graphql, PreloadedQuery, usePreloadedQuery, useQueryLoader } from 'react-relay';
+import Typography from '@mui/material/Typography';
 import StixCoreObjectOpinionsRadar, {
-  stixCoreObjectOpinionsRadarFragmentQuery,
+  stixCoreObjectOpinionsRadarDistributionQuery,
 } from './StixCoreObjectOpinionsRadar';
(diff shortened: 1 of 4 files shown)


yassine-ouaamou:
It seems like each edit creates a new opinion. It doesn't update the previous opinion

opencti-platform/opencti-front/src/private/components/analyses/opinions/StixCoreObjectOpinions.tsx
```diff
@@ -44,27 +77,57 @@ StixCoreObjectOpinionsProps
     () => fetchLoadQuery(variables, { fetchPolicy: 'store-and-network' }),
     [],
   );
+
+  const height = 260;
```
SarahBocognano, line 81: Why just didn't define it directly in the style line ?

SarahBocognano (commented):
Isn't it possible to modify our opinion ? If I want to change my opinion, it create a new one but doesn't change mine

RomuDeuxfois:
The bug found has been fixed -> is not possible to update my opinion
PR ready to be re-tested

opencti-platform/opencti-front/src/private/components/analyses/opinions/StixCoreObjectOpinions.tsx
RomuDeuxfois: Because it's used in different places

```
<StixCoreObjectOpinionsRadar
            queryRef={queryRef}
            height={height}
            opinionOptions={opinionOptions}
          />
```

SarahBocognano (approved):

yassine-ouaamou (approved):

Issue #4217, opened 2023-09-01
Entity alias edition fields not working in graphs


Archidoit:
Go to an investigation.
Add an organization.
Click on the organization and click on 'edit' in the toolbar.
There are 2 fields 'aliases' and the first one is not editable.

--> Same for all the entities with x_opencti_aliases (ex: position, individuals, etc), same in all the graphs (knowledge graphs, etc).

![image](https://github.com/OpenCTI-Platform/opencti/assets/75783086/c4b68b5b-8ff7-427e-9919-8dc1813c2294)

![image](https://github.com/OpenCTI-Platform/opencti/assets/75783086/39c2058e-5309-4d0c-8edf-2fd9da7167d5)

PR #4218, opened 2023-09-01
[frontend] Fix entity alias edition field in graphs (#4217)

Archidoit:
issue: https://github.com/OpenCTI-Platform/opencti/issues


SarahBocognano (approved):

opencti-platform/opencti-front/src/private/components/common/stix_domain_objects/StixDomainObjectEditionOverview.jsx
```diff
@@ -359,7 +359,7 @@ const StixDomainObjectEditionContainer = (props) => {
                   }
                 />
               )}
-              {'aliases' in stixDomainObject && (
```
RomuDeuxfois, line 362: How can we have both ?
Do we have a problem with the declaration of attributes which means that we have the two fields: aliases & x_opencti_aliases?

Archidoit:
other bug to fix: when adding an entity to an investigation graph, the entities already present in the graph are not crossed in the list.

opencti-platform/opencti-front/src/private/components/common/stix_domain_objects/StixDomainObjectEditionOverview.jsx
Archidoit: seen

PR #4245, opened 2023-09-06
[backend] configure docker volumes for elasticsearch

SouadHadjiat:
### Proposed changes

* configure docker volumes to persist data when we update elasticsearch image (otherwise all data is lost when container is updated)

### Checklist

### Further comments


SarahBocognano:
Question: Do we have a way to test it ?

sbocahu:
As you also added a volume for snapshots, I would recommend setting the path.repo config that would permit to use the snapshot feature when needed. (env path.repo=...)

helene-nguyen:
Locally tested:

- recreate new containers with new images
- add some data
- re play docker compose with another version of Elasticsearch image

Seems that all good :)

helene-nguyen (approved):

PR #4274, opened 2023-09-07
[frontend] Complete refactor of TopBar (#2796)

### LINKED ISSUE #2796: Refactor TopBar
We always use two topbar at the same place for entities as the main TopBar doesn't work. It needs to be redone completely to prevent multiple renders with no reason

### LINKED ISSUE #4237: When putting a custom dashboard as the home dashboard, not possible to come back on the default one
## Description

1. Create a custom dashboard
2. Put the custom as home dashboard
3. Go back in "Automatic"

=> Never possible to come back on the default one.

Kedae:


opencti-platform/opencti-front/src/private/Root.tsx
```diff
@@ -33,6 +33,9 @@ const rootPrivateQuery = graphql`
         id
         name
       }
+      default_dashboard {
```
marieflorescontact, line 36: is that in order to solve the [default dashboard bug](https://github.com/OpenCTI-Platform/opencti/issues/4237)?

opencti-platform/opencti-front/src/private/Index.jsx
```diff
@@ -25,20 +25,17 @@ import Message from '../components/Message';
...
-import TopBar from './components/nav/TopBar';
 import RootCases from './components/cases/Root';
 import SystemBanners from '../public/components/SystemBanners';
 import TimeoutLock from './components/TimeoutLock';
 import useAuth from '../utils/hooks/useAuth';
-import SettingsMessagesBanner, {
-  useSettingsMessagesBannerHeight,
-} from './components/settings/settings_messages/SettingsMessagesBanner';
+import SettingsMessagesBanner, { useSettingsMessagesBannerHeight } from './components/settings/settings_messages/SettingsMessagesBanner';
+import TopBar from "@components/nav/TopBar";
```
marieflorescontact, line 33: is this notation `@` supposed to be used for imports from now on?

marieflorescontact:
locally tested 👌

opencti-platform/opencti-front/src/private/components/nav/TopBar.tsx
```diff
@@ -194,6 +177,79 @@ const topBarQuery = graphql`
   }
 `;
 
+const routes = {
+  // ME
+  '/dashboard/profile/me': () => <TopMenuProfile />,
```
RomuDeuxfois, line 182: We should think about path duplication : based direclty on route declaration.

And

We can have a TopBar service where we can push specific TopBar (like TopMenuProfile) when we are in a specific component to have dependency inversion.

RomuDeuxfois:
Not working for all cases menus and grouping menu

image

opencti-platform/opencti-front/src/private/Root.tsx
RomuDeuxfois: locally tested 👌

SarahBocognano (commented):
Tested ✅ Same remarks as @RomuDeuxfois : does not work for all cases section, and for groupings. The rest is ok

opencti-platform/opencti-front/src/private/Root.tsx
Kedae: Partially yes :)

opencti-platform/opencti-front/src/private/Index.jsx
Kedae: It will, but it's not mandatory to do something about it.

opencti-platform/opencti-front/src/private/components/nav/TopBar.tsx
Kedae: Let's think about that later 😆

RomuDeuxfois (approved):

PR #4276, opened 2023-09-07
[backend] Improve merging and deleting for self refs (#4226)

### LINKED ISSUE #4226: Merging a report into another report lead to have the same report in object_refs
## Description

1. Create 2 reports
2. Merge the first one in the other one

Result: 

image

In entities and the knowledge graph, the report seems to be in the object_refs of itself.

This behavior should be discussed with @richard-julien because I thought we have a low level prevention on relationship between the same source and target (and this ref is obviously not respecting this rule).

richard-julien:
See https://github.com/OpenCTI-Platform/opencti/issues/4226

### DIFF
--- a/opencti-platform/opencti-graphql/src/database/middleware.js
+++ b/opencti-platform/opencti-graphql/src/database/middleware.js
@@ -1029,11 +1029,19 @@ export const hashMergeValidation = (instances) => {
 // region mutation update
 const ed = (date) => isEmptyField(date) || date === FROM_START_STR || date === UNTIL_END_STR;
 const noDate = (e) => ed(e.first_seen) && ed(e.last_seen) && ed(e.start_time) && ed(e.stop_time);
-const filterTargetByExisting = (targetEntity, sources, targets) => {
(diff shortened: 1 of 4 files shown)


opencti-platform/opencti-graphql/src/database/middleware.js
```diff
@@ -1029,26 +1029,31 @@ export const hashMergeValidation = (instances) => {
...
   const cache = [];
+  const sources = sourcesDependencies[`i_relations_${redirectSide}`];
+  const targets = targetDependencies[`i_relations_${redirectSide}`];
   for (let index = 0; index < sources.length; index += 1) {
     const source = sources[index];
+    // If the relation source is already in target = filtered
+    const finder = (t) => {
+      const sameTarget = t.internal_id === source.internal_id;
+      const sameRelationType = t.i_relation.entity_type === source.i_relation.entity_type;
+      return sameRelationType && sameTarget && noDate(t.i_relation);
```
RomuDeuxfois, line 1043: Why we are checking the date here ?

RomuDeuxfois:
Repro case ✅

Issue #4255, opened 2023-09-07
Killchain is Unknown is list widget in dashboards


SamuelHassine:
## Description

1. Create a custom dashboard
2. Create a widget with list distribution (numbers)
3. Type = Indicator, distribute on Kill Chain Phase

https://filigran.octi.filigran.io/dashboard/workspaces/dashboards/6c25e9a7-19f0-4e4e-badd-33adaeefbdd9

![image](https://github.com/OpenCTI-Platform/opencti/assets/1334279/f3fef92a-ef86-4c34-8fdb-624c4f25b140)

PR #4289, opened 2023-09-08
[backend/frontend] Improv representative management (#4255)

RomuDeuxfois:
### Proposed changes

* https://github.com/OpenCTI-Platform/opencti/issues/4255

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectsDistributionList.jsx
+++ b/opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectsDistributionList.jsx
@@ -13,7 +13,6 @@ import ListItemText from '@mui/material/ListItemText';
 import { QueryRenderer } from '../../../../relay/environment';
 import { useFormatter } from '../../../../components/i18n';
 import { convertFilters } from '../../../../utils/ListParameters';
-import { defaultValue } from '../../../../utils/Graph';
 import { resolveLink } from '../../../../utils/Entity';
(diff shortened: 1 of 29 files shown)


SarahBocognano (commented):
Behavior works well

opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectsDistributionList.jsx
```diff
@@ -99,148 +98,27 @@ const stixCoreObjectsDistributionListDistributionQuery = graphql`
...
-          opinion
-        }
+          ... on StixObject {
+              id
+              entity_type
+              representative {
+                  main
+              }
+          }
+          ... on StixRelationship {
```
richard-julien, line 108: Could b interesting to introduce the basic object abstraction?

RomuDeuxfois: Totally agree

RomuDeuxfois: This would give representation to internal entities like User, Role, Connector, ect.
I'm not sure that these are entities that we want to manage.

RomuDeuxfois: Not handle now

SarahBocognano (approved):

PR #4290, opened 2023-09-08
[backend] Fix merge of vocabularies

### LINKED ISSUE #4258: Merging of vocabularies is broken
## Description

1. Go to Settings / Taxonomies / Vocabularies
2. Go to attack_motivation_ov
3. Select personal-gain and personal-satisfaction

image

4. Merge them

Background task finished, the 2 vocabularies are still in the list.

Kedae:


opencti-platform/opencti-graphql/src/domain/backgroundTask-common.js
```diff
@@ -168,5 +169,5 @@ export const createListTask = async (context, user, input) => {
 };
 
 export const isTaskEnabledEntity = (entityType) => {
```
RomuDeuxfois, line 180: We can add an information message for the user where the entity passed as a parameter is not managed on the back side.
What do you think ?

Archidoit:
Sugestions for design in Background task lists:
- 'Settings' translation
- set 'Settings' scope color to sth else than grey (not lisible and a color associated to deprecated/not-important)
image

Archidoit:
Behavior OK 👍

opencti-platform/opencti-front/src/components/TaskScope.tsx
```diff
@@ -42,6 +42,14 @@ const TaskScope: FunctionComponent<TaskScopeProps> = ({ label, scope }) => {
   const classes = useStyles();
   const style = classes.chip;
   switch (scope) {
+    case 'SETTINGS':
+      return (
+        <Chip
+          classes={{ root: style }}
+          style={inlineStyles.rose}
```
Archidoit: same color as scope = 'RULE'

SarahBocognano (commented):
Tested ✅ Works fine

opencti-platform/opencti-front/src/components/TaskScope.tsx
Kedae: Done

opencti-platform/opencti-graphql/src/domain/backgroundTask-common.js
Kedae: The API handles that, we just didn't add it in front and it could hide the fact that we put the toolbar somewhere we don't want to.

PR #4305, opened 2023-09-11
In TA and Sectors, inferred relationships located-at should not have a button "unlink / delete" (#issue/4220)

### LINKED ISSUE #4220: In TA, inferred relationships located-at should not have a button "unlink / delete"
## Bug

In Threat Actor (all) => Overview:

Quick located-at is displaying the button "unlink" on inferred relationships (no effect).

image

SarahBocognano:
### Related issues
 
* https://github.com/OpenCTI-Platform/opencti/issues/4220

(Don't forget to enable rules propagation for located-at and part-of to repro cases)

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/entities/sectors/SectorSubSectors.jsx
+++ b/opencti-platform/opencti-front/src/private/components/entities/sectors/SectorSubSectors.jsx
@@ -109,7 +109,6 @@ const SectorSubSectors = createFragmentContainer(SectorSubSectorsComponent, {
       name
       parent_types
       entity_type
-      isSubSector
       subSectors {
         edges {
           node {
--- a/opencti-platform/opencti-front/src/private/components/entities/sectors/SectorSubSectors.jsx
+++ b/opencti-platform/opencti-front/src/private/components/entities/sectors/SectorSubSectors.jsx
@@ -57,6 +57,7 @@ class SectorSubSectorsComponent extends Component {
         <List style={{ marginTop: -10 }}>
           {sector.subSectors.edges.map((subSectorEdge) => {
             const subSector = subSectorEdge.node;
+            const isInferred = subSector.is_from_relation_inferred;
             return (
               <ListItem
                 key={subSector.id}
@@ -71,13 +72,19 @@ class SectorSubSectorsComponent extends Component {
                 </ListItemIcon>
                 <ListItemText primary={subSector.name} />
                 <ListItemSecondaryAction>
-                  <IconButton
-                    aria-label="Remove"
-                    onClick={this.removeSubSector.bind(this, subSectorEdge)}
-                    size="large"
-                  >
-                    <LinkOff />
-                  </IconButton>
+                  {isInferred ? (
+                    <></>
+                  ) : (
+                    <Security needs={[KNOWLEDGE_KNUPDATE]}>
+                      <IconButton
+                        aria-label="Remove"
+                        onClick={this.removeSubSector.bind(this, subSectorEdge)}
+                        size="large"
+                      >
+                        <LinkOff />
+                      </IconButton>
+                    </Security>
+                  )}
                 </ListItemSecondaryAction>
               </ListItem>
             );
@@ -102,6 +109,7 @@ const SectorSubSectors = createFragmentContainer(SectorSubSectorsComponent, {
       name
       parent_types
       entity_type
+      isSubSector
       subSectors {
         edges {
           node {
@@ -109,6 +117,7 @@ const SectorSubSectors = createFragmentContainer(SectorSubSectorsComponent, {
             name
             description
             parent_types
+            is_from_relation_inferred
           }
         }
       }
--- a/opencti-platform/opencti-front/src/private/components/entities/sectors/SectorSubSectors.jsx
+++ b/opencti-platform/opencti-front/src/private/components/entities/sectors/SectorSubSectors.jsx
@@ -56,8 +56,8 @@ class SectorSubSectorsComponent extends Component {
         <div className="clearfix" />
         <List style={{ marginTop: -10 }}>
           {sector.subSectors.edges.map((subSectorEdge) => {
+            const { types } = subSectorEdge;
             const subSector = subSectorEdge.node;
-            const isInferred = subSector.is_from_relation_inferred;
             return (
               <ListItem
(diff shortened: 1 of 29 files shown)


marieflorescontact:
locally tested. 👌

2 observations that are probably not related to this bug fix:
- fully refresh is required to display/delete inferred relationships when adding a relationship via the quick `add` button. 
- tooltip missing on the quick `add` button and on the `unlink / delete` button .


@SarahBocognano : Good call on the tooltip missing ! And you are right, the fully refresh is still required in the app to display/delete inferred relationships. That behavior will not be change on this issue.

@SarahBocognano : @marieflorescontact After investigation, there is no 'add' button or' unlike / delete' button who have a tooltip on the platform. So to not change the global behavior, I will not add them.

opencti-platform/opencti-front/src/private/components/entities/sectors/SectorSubSectors.jsx
```diff
@@ -102,13 +109,15 @@ const SectorSubSectors = createFragmentContainer(SectorSubSectorsComponent, {
       name
       parent_types
       entity_type
+      isSubSector
```
SouadHadjiat: Where is it used `isSubSector` ?

opencti-platform/opencti-graphql/src/database/middleware.js
```diff
@@ -307,6 +307,12 @@ const batchListThrough = async (context, user, sources, sourceSide, relationType
     targetIds = targetIds.filter((id) => elementIds.includes(id));
   }
   const targets = await elFindByIds(context, user, targetIds, opts);
+  // For each target, search if the relation is inferred
+  const findAllTargets = targets.map((target) => {
+    const findRelation = relations.find((relation) => (relation.toId === target.id) || (relation.fromId === target.id));
```
SouadHadjiat: could it happen that there are multiple relations for the same target ? since we are in a batch method, we return results for a list of entities. So we could have entityA that has relations, one inferred with targetA, and entityB that also has a relation with targetA that is not inferred, so what would happen ?

Archidoit:
Don't know if it's a good idea, but maybe we can add an 'inferred' indication/icon instead of the removed 'unlink' button, to indicate why we can't manage the relation (concern the 'reg1' line in the screen short below):
image

@SarahBocognano : It could be a good idea to add an icon like in the "last created relationship" component where we display a particular icon for inferred relationship with a tooltip indicating why. But right now, this behavior is not present in other part of the platform (in IntrusionSet for instance) so I'm not sure if we should do it now. What is your thought on that @Jipegien ?

Jipegien:
> Don't know if it's a good idea, but maybe we can add an 'inferred' indication/icon instead of the removed 'unlink' button, to indicate why we can't manage the relation (concern the 'reg1' line in the screen short below): image
> 
> @SarahBocognano : It could be a good idea to add an icon like in the "last created relationship" component where we display a particular icon for inferred relationship with a tooltip indicating why. But right now, this behavior is not present in other part of the platform (in IntrusionSet for instance) so I'm not sure if we should do it now. What is your thought on that @Jipegien ?

Same comportment for TA and IS. For IntrusionSet, it is not a "located-at" relationship but a "originates-from". For having the same comportment, we need to add an inference rule to propagates the originates-from to parents (region).

RomuDeuxfois:
Hey,
What is the status of this work ?

Kedae:
> Hey, What is the status of this work ?

Waiting for review

opencti-platform/opencti-front/src/private/components/entities/sectors/SectorSubSectors.jsx
```diff
@@ -71,13 +72,19 @@ class SectorSubSectorsComponent extends Component {
...
                 <ListItemText primary={subSector.name} />
                 <ListItemSecondaryAction>
-                  <IconButton
-                    aria-label="Remove"
-                    onClick={this.removeSubSector.bind(this, subSectorEdge)}
-                    size="large"
-                  >
-                    <LinkOff />
-                  </IconButton>
+                  {isInferred ? (
```
lndrtrbn: Instead of using a ternary with an empty React Fragment you can wrap `ListItemSecondaryAction` with a simple condition `!isInferred` like this:
```javascript
{!isInferred && (
  <ListItemSecondaryAction>
    <Security needs={[KNOWLEDGE_KNUPDATE]}>
      <IconButton
        aria-label="Remove"
        onClick={this.removeSubSector.bind(this, subSectorEdge)}
        size="large"
      >
        <LinkOff />
      </IconButton>
    </Security>
  </ListItemSecondaryAction>
)}
``` 

Same for the 2 other components

opencti-platform/opencti-graphql/src/modules/case/case-rfi/case-rfi.graphql
```diff
@@ -96,6 +96,7 @@ type CaseRfi implements BasicObject & StixObject & StixCoreObject & StixDomainOb
   objectAssignee: AssigneeConnection
   avatar: OpenCtiFile
   # Container
+  is_from_relation_inferred: Boolean
```
richard-julien: Not sure about the placement of this. We need to check if it could be more interesting to put the information in the edge type

opencti-platform/opencti-graphql/src/database/middleware.js
```diff
@@ -305,16 +305,27 @@ const batchListThrough = async (context, user, sources, sourceSide, relationType
...
   const elGrouped = R.groupBy((e) => e[`${sourceSide}Id`], relations);
   if (paginate) {
     return ids.map((id) => {
       const values = elGrouped[id];
       const edges = [];
       if (values) {
-        const data = first ? R.take(first, values) : values;
-        const filterIds = (data || []).map((i) => i[`${opposite}Id`]);
-        const filteredElements = targets.filter((t) => filterIds.includes(t.internal_id));
+        const currentSourceRelations = first ? R.take(first, values) : values;
```
richard-julien: I need more information about why we need to filter information here

richard-julien (changes requested):

SarahBocognano:
Update: Changes requested have been made. Waiting for another review

SamuelHassine:
@Kedae can we push a little on this one after 4 months?

richard-julien (approved):

PR #4311, opened 2023-09-11
[frontend] fix trigger digest time with right timezone (#4310)

### LINKED ISSUE #4310: Digest trigger doesn't send the digest for the right timezone
## Description

If you configure a daily digest trigger with a specific time to send the notification, you'll get your notification one hour later then the configured time, during daylight savings time.

## Environment

2. OpenCTI version: 5.10.1
3. OpenCTI client: frontend
4. Other environment details:

## Reproducible Steps

1. Create a live trigger
2. Create a digest based on this live trigger, set the period to daily, and set a time
3. The digest will send the notification one hour later (you can also check the time saved for the digest in your database)

## Screenshots (optional)

Capture d'écran 2023-09-11 192638
Capture d'écran 2023-09-11 192649

SouadHadjiat:
### Proposed changes

* Set today's date along with digest trigger time : we used to set a fixed date, which doesn't get us the right timezone, with daylight savings time
* refactor TimePickerField to remove some duplicate code

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/4310

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/components/TimePickerField.jsx
+++ b/opencti-platform/opencti-front/src/components/TimePickerField.jsx
@@ -66,69 +66,25 @@ const TimePickerField = (props) => {
       onSubmit(name, value ? parse(value).toISOString() : null);
     }
   }, [setTouched, onSubmit, name]);
-  if (withSeconds) {
-    return (
-      <TimePicker
-        {...fieldToTimePicker(props)}
-        variant="inline"
-        disableToolbar={false}
-        autoOk={true}
-        allowKeyboardControl={true}
-        onAccept={internalOnAccept}
-        onChange={internalOnChange}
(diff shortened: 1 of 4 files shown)


marieflorescontact:
locally tested:
it works correctly using the clock, but the problem persists when directly modifying the input text

SouadHadjiat:
> locally tested: it works correctly using the clock, but the problem persists when directly modifying the input text

The problem with the input should be solved when we'll upgrade to the latest version of mui, you can see here that the input is well handled on the examples provided : https://mui.com/x/react-date-pickers/timezone/, which is not the case for 5.x versions : https://v5.mui.com/x/react-date-pickers/getting-started/

marieflorescontact (approved):

PR #4329, opened 2023-09-14
[frontend] Use internal Drawer in Cases creation

RomuDeuxfois:
### Proposed changes

* Use internal Drawer component in Incident response, Case RFT, Case RFI and Task creation


Archidoit (approved):

Archidoit:
Works well 🆗

PR #4338, opened 2023-09-15
[backend] Fix Unknown value for Software creation

### LINKED ISSUE #4336: Software name displayed as unknown
- Go to the observables tab
- Create a software by filling ONLY the filed 'CPE'
- Its name is displayed as 'unknown'

Same in Reports > Observables.

image

marieflorescontact:
### Proposed changes

* Fix `Unkown` value for Software creation
* Add mandatory attributes (name || cpe || swid ) when creating a software => it was possible to create a software with just a version 

### Related issues

* [issue/4336](https://github.com/OpenCTI-Platform/opencti/issues/4336)

### Checklist

### Further comments

### DIFF
--- a/opencti-platform/opencti-graphql/src/utils/format.js
+++ b/opencti-platform/opencti-graphql/src/utils/format.js
@@ -95,7 +95,7 @@ export const observableValue = (stixCyberObservable) => {
     case ENTITY_PROCESS:
       return stixCyberObservable.pid || stixCyberObservable.command_line || 'Unknown';
     case ENTITY_SOFTWARE:
-      return stixCyberObservable.name || 'Unknown';
+      return stixCyberObservable.name || stixCyberObservable.cpe || stixCyberObservable.swid || 'Unknown';
     case ENTITY_USER_ACCOUNT:
       return stixCyberObservable.account_login || stixCyberObservable.user_id || 'Unknown';
     case ENTITY_BANK_ACCOUNT:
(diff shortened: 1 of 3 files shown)


opencti-platform/opencti-graphql/src/utils/syntax.js
```diff
@@ -168,6 +169,9 @@ export const checkObservableSyntax = (observableType, observableData) => {
     case C.ENTITY_MAC_ADDR:
       if (!macAddrChecker.test(observableData.value)) return 'Valid MAC address';
       break;
+    case C.ENTITY_SOFTWARE:
+      if (!observableData.name && !cpeChecker.test(observableData.cpe) && !observableData.swid) return 'Valid Software attributes';
```
Goumies, line 173: We could extract this condition to a constant `areValidSoftwareAttributes` to explicit it

opencti-platform/opencti-graphql/src/utils/format.js
```diff
@@ -95,7 +95,7 @@ export const observableValue = (stixCyberObservable) => {
     case ENTITY_PROCESS:
       return stixCyberObservable.pid || stixCyberObservable.command_line || 'Unknown';
     case ENTITY_SOFTWARE:
-      return stixCyberObservable.name || 'Unknown';
+      return stixCyberObservable.name || stixCyberObservable.cpe || stixCyberObservable.swid || 'Unknown';
```
Goumies, line 98: We could create a function returning the truthy submitted argument or the second argument as the default value.

Something like : 
```
function handleValue(attributes: Array<Attribute>, defaultValue: string): Attribute | string {
  return attributes.find(attribute => !!attribute) || defaultValue
}
```
And use it for every switch case.

Further, refactoring could be the use of this function in Map with switch cases as keys.

RomuDeuxfois, line 98: You missed this TODO

// TODO for now this list is duplicated in Front, think about updating it aswell

You need to update the Entity.ts file in front or find a way to retrieve the observableValue from the back-end

RomuDeuxfois:
Tested 🆗

opencti-platform/opencti-graphql/src/utils/format.js
marieflorescontact: thanks @RomuDeuxfois , i've missed that 🙏

marieflorescontact: thanks for the suggestion, the refacto will be made later when we have found a way to no longer have the list duplicated in the front and back.

RomuDeuxfois (approved):

Issue #2686, opened 2022-12-28
Full refactor of filtering and searching using grouping / operators / fuzzy filters


Archidoit:
## Use case

Adding the possibility to switch between the AND and the OR operators in filters

## Solution

When some filters are selected, clicking on the 'OR'/'AND' button between filters would switch to the AND/OR operation between those filters.

## Comment
Global filter refacto also done in this issue.

## Notion page
https://www.notion.so/filigran/Switch-AND-OR-operators-in-filters-change-filters-format-d05fe9af797c4751a293e3a7046ef89d#23a6b0fb69c94d2a8b837b50184c60b8

ristowee:
This would be nice.
Currently there's no way to allow unmarked objects in the filter, e.g. "CLEAR **OR** GREEN **OR** unmarked". I tried to go around it by creating a negated filter forbidding AMBER, AMBER+STRICT and RED with a "**NOT**(AMBER **OR** AMBER+STRICT **OR** RED)" but that wasn't possible because in negated filter the combiner operator was **AND**.

Archidoit:
Will also be resolved in this issue : https://github.com/OpenCTI-Platform/opencti/issues/4323 (Filters not working for Threat Actors Individual)

PR #4396, opened 2023-09-19
[frontend/backend] Filters operator switching + filters refacto (#2686)

Archidoit:
## PR content
- filters refacto to prepare the future version of filters
- enable filters operator switching between filters of different groups and withing a filter group

## Dynamic filters
- back filters refacto, new logic implementation (imbrication, modes switching etc)
- front filters refacto
- backend tests for filters complex combinations
- operator switching in front: enable operator switching by clicking on the 'AND/OR'
- filter value resolution in the back
- rename all the local storage keys (not to use local storage containing old filters anymore)
- remove filterMode option
- handle redirection from an url with filters in the old format

## General
- rename filter keys that were different between front and back
- filter type generalization to supress filter enums
- filter keys checking and conversion in the back according to the schema definition
- generalize some args in filters
- introduce 'nil' and 'not_nil' operators (filter empty / not empty fields)
- cache: add an object containing all the existing attributes in the schema definitions name with their type (for filters key checking)
- list special filter keys (special behaviors)
- fix 'some filters cannot have NOT operator' (https://github.com/OpenCTI-Platform/opencti/issues/4878)
- client-python filter types adaptation: see associated PR
- connectors filter types adaptation: see associated PR

## Stored filters
- query background tasks: back & front adaptation
- feed and taxii: back & front adaptation
- write backend tests with complex filtering for taxii collections
- streams, playbooks, triggers: back & front adaptation
- migration of stored filters (feeds, taxii collections, triggers, not finished query background tasks, workspaces, playbooks, rules)
- remove unncessary filter ids resolution in backend
- new filter key 'connectedToId' for instance triggers
- new encoded logic for stix and events: isStixMatchFilters / isEventMatchFilters
- store filter keys as array in the back, handle them as string in the front

## Ressources

Notion page: https://www.notion.so/filigran/Switch-between-AND-and-OR-operators-between-filters-change-filter-schema-d05fe9af797c4751a293e3a7046ef89d#23a6b0fb69c94d2a8b837b50184c60b8

Github issue: https://github.com/OpenCTI-Platform/opencti/issues/2686  and https://github.com/OpenCTI-Platform/opencti/issues/4878 (bug)

Associated PR in client-python: https://github.com/OpenCTI-Platform/client-python/pull/461

Associated PR in documentation: https://github.com/OpenCTI-Platform/docs/pull/70

Associated PR in connectors: https://github.com/OpenCTI-Platform/connectors/pull/1544


SouadHadjiat:
I have this error when trying to see Reports list (and all sub screens "Groupings", "Notes", ...)

Capture d'écran 2023-09-20 172525

I guess it's linked to my local storage which might not be compatible anymore (I think we should handle this, otherwise all clients would have this issue when we'll deploy). 
I tried to open the app in a "private mode" browser page, and I still have the same error for "Notes", but not anymore for the other "Analyses" screens. All other list views have an error, even if I don't have anything in my local storage and there is no filter applied.

SouadHadjiat:
`filtersRepresentatives` query is called even for filters that won't return any result (ex: created)

Capture d'écran 2023-09-20 174407

Capture d'écran 2023-09-20 174416

Archidoit:
> I have this error when trying to see Reports list (and all sub screens "Groupings", "Notes", ...)
> 
> Capture d'écran 2023-09-20 172525
> 
> I guess it's linked to my local storage which might not be compatible anymore (I think we should handle this, otherwise all clients would have this issue when we'll deploy). I tried to open the app in a "private mode" browser page, and I still have the same error for "Notes", but not anymore for the other "Analyses" screens. All other list views have an error, even if I don't have anything in my local storage and there is no filter applied.

@SouadHadjiat It's normal that you have an error for Notes since the filter types have not been adapted for this entity type (they have for the other Analyses screen).
For the local storage, it will be handled later.

Archidoit:
> `filtersRepresentatives` query is called even for filters that won't return any result (ex: created)
> 
> Capture d'écran 2023-09-20 174407
> 
> Capture d'écran 2023-09-20 174416

fixed 🌞

opencti-platform/opencti-graphql/tests/02-integration/02-resolvers/grouping-test.js
```diff
@@ -288,10 +288,16 @@ describe('Grouping resolver standard behavior', () => {
...
-          filters: [{
-            key: 'type',
-            value: 'investigation'
-          }]
+          filters: {
+            mode: 'and',
+            filters: [{
+              key: 'type',
+              value: 'investigation',
+              operator: 'eq',
```
RomuDeuxfois: We loose the default value for the operator?

opencti-platform/opencti-graphql/config/schema/opencti.graphql
```diff
@@ -80,6 +80,18 @@ enum FilterMode {
   or
 }
 
+input Filter {
+  key: String
+  values: [String]
```
RomuDeuxfois: This should be mandatory no ?

opencti-platform/opencti-graphql/config/schema/opencti.graphql
```diff
@@ -80,6 +80,18 @@ enum FilterMode {
...
+  operator: String
+  mode: String
+}
+
+type FilterRepresentative {
```
RomuDeuxfois: I am not sur to understand the usage of this one

opencti-platform/opencti-graphql/config/schema/opencti.graphql
```diff
@@ -1748,11 +1760,16 @@ enum ExternalReferencesFilter {
   creator
   created
 }
-input ExternalReferencesFiltering {
+input ExternalReferencesGroupFiltering {
+  mode: FilterMode!
+  filters: [ExternalReferencesFilterFiltering]!
+  filterGroups: [ExternalReferencesGroupFiltering]!
```
RomuDeuxfois:
```suggestion
  filters: [ExternalReferencesFilterFiltering!]!
  filterGroups: [ExternalReferencesGroupFiltering!]!
```

https://graphql-dotnet.github.io/docs/getting-started/lists-non-null/

opencti-platform/opencti-graphql/config/schema/opencti.graphql
```diff
@@ -2157,11 +2174,16 @@ enum StixDomainObjectsFilter {
   result_name
   operatingSystem
 }
-input StixDomainObjectsFiltering {
+input StixDomainObjectsGroupFiltering {
+  mode: FilterMode!
+  filters: [StixDomainObjectsFilterFiltering]!
```
RomuDeuxfois: Same as above for the array

https://graphql-dotnet.github.io/docs/getting-started/lists-non-null/

opencti-platform/opencti-graphql/tests/02-integration/02-resolvers/report-test.js
```diff
@@ -462,11 +462,17 @@ describe('Report resolver standard behavior', () => {
...
-            value: 'investigation'
-          }]
-        }
+          filters: {
+            mode: 'and',
+            filters: [{
+              key: 'type',
+              value: 'investigation',
+              operator: 'eq',
+              mode: 'or',
```
RomuDeuxfois: Is not default values anymore ?
I think so in engine file

opencti-platform/opencti-graphql/config/schema/opencti.graphql
RomuDeuxfois: Regarding the tests, filterGroups can be null instead of empty array a lot of times

opencti-platform/opencti-graphql/src/modules/malwareAnalysis/malwareAnalysis.graphql
```diff
@@ -178,11 +178,16 @@ enum MalwareAnalysesFilter {
     installedSoftware
 }
 
-input MalwareAnalysesFiltering {
+input MalwareAnalysesGroupFiltering {
+    mode: FilterMode!
+    filters: [MalwareAnalysesFilterFiltering]!
```
RomuDeuxfois: We re defined the [entity]GroupFiltering because of the enum filters of each entity..
I don't know if it's possible to have a generic type and handle enum filters in an other way

opencti-platform/opencti-graphql/src/utils/filtering.js
```diff
@@ -98,49 +102,51 @@ export const resolvedFiltersMapForUser = async (context, user, filters) => {
   return resolveUserMap;
 };
 
-export const convertFiltersFrontendFormat = async (context, user, filters) => {
-  // Grab all values that are internal_id that needs to be converted to standard_ids
-  const resolvedMap = await getEntitiesMapFromCache(context, SYSTEM_USER, ENTITY_TYPE_RESOLVED_FILTERS);
-  // Remap the format of specific keys
+const convertFiltersFrontendFormatContent = async (context, user, mainFilterGroup, resolvedMap) => {
```
RomuDeuxfois: The arg name : mainFilterGroup is confusing for me
The name filterGroup is enough I think

opencti-platform/opencti-graphql/src/utils/filtering.js
```diff
@@ -98,49 +102,51 @@ export const resolvedFiltersMapForUser = async (context, user, filters) => {
...
-      const workingKey = key.replace('_lte', '');
-      adaptedFilters.push({ key: workingKey, operator: 'lte', values });
-    } else if (key.endsWith('_not_eq')) {
-      const workingKey = key.replace('_not_eq', '');
-      adaptedFilters.push({ key: workingKey, operator: 'not_eq', values, filterMode: 'and' });
-    } else {
-      adaptedFilters.push({ key, operator: 'eq', values, filterMode: 'or' });
-    }
+    adaptedFilters.push({ ...filter, values });
   }
```
RomuDeuxfois: Why we don't need all this cases ?

opencti-platform/opencti-graphql/tests/02-integration/01-database/filterGroup-test.js
```diff
@@ -0,0 +1,641 @@
+import { describe, expect, it } from 'vitest';
```
RomuDeuxfois, line 1: It’s so great to have these tests, thank you!

opencti-platform/opencti-graphql/tests/02-integration/01-database/filterGroup-test.js
```diff
@@ -0,0 +1,641 @@
...
+    queryResult = await queryAsAdmin({ query: LIST_QUERY,
+      variables: {
+        first: 10,
+        filters: {
+          mode: 'and',
+          filters: [],
+          filterGroups: [],
+        },
+      } });
+    // expect(queryResult2.data.reports.edges.length).toEqual(4); TODO uncomment
```
RomuDeuxfois: If we have no key, we have all the reports or instead we have no report ?

Just to understand this behavior

opencti-platform/opencti-graphql/src/domain/stixDomainObject.js
```diff
@@ -91,26 +91,30 @@ export const stixDomainObjectAvatar = (stixDomainObject) => {
...
+  const filteringArgs = { search, orderBy, orderMode, filters, filterMode, relationship_type, elementId };
   const filtersOpts = stixDomainObjectOptions.StixDomainObjectsFilter;
   const ordersOpts = stixDomainObjectOptions.StixDomainObjectsOrdering;
-  let newArgsFiltersFilters = argsFilters.filters;
+  let newArgsFiltersFilters = filteringArgs.filters?.filters ?? [];
   const initialParams = {};
-  if (argsFilters.filters && argsFilters.filters.length > 0) {
-    if (argsFilters.filters.filter((n) => n.key.includes('elementId')).length > 0) {
-      initialParams.elementId = R.head(R.head(argsFilters.filters.filter((n) => n.key.includes('elementId'))).values);
+  if (filteringArgs.filters?.filters && filteringArgs.filters.filters.length > 0) {
```
RomuDeuxfois, line 139: Can we have recursivity instead of specific imbrication ?

opencti-platform/opencti-graphql/src/domain/stixCoreObject.js
```diff
@@ -459,3 +459,106 @@ export const stixCoreObjectEditContext = async (context, user, stixCoreObjectId,
...
+
+// filters representatives
+export const findFilterRepresentative = async (context, user, filter) => {
+  const { key, values } = filter;
+  let data;
+  switch (key) {
+    case 'toSightingId':
+      data = await Promise.all(values.map(async (id) => {
+        const result = await storeLoadById(context, user, id, ENTITY_TYPE_IDENTITY_INDIVIDUAL);
+        return {
```
RomuDeuxfois: Ultimately, you can't use extractEntityRepresentativeName in this method ?

I think, you can use it 
Instead of 
```
        return {
          id,
          value: result?.name,
        };
```
You can have 
```
        return {
          id,
          value: result ? extractEntityRepresentativeName(result) : undefined,
        };
```

opencti-platform/opencti-front/src/components/FilterIconButtonContainer.tsx
```diff
@@ -0,0 +1,187 @@
...
+    overflow: 'hidden',
+    textOverflow: 'ellipsis',
+  },
+}));
+
+interface FilterIconButtonContainerProps {
+  displayedFilters: Filter[];
+  globalMode: string;
+  handleRemoveFilter?: (key: string, op?: string) => void;
+  handleSwitchGlobalMode?: () => void;
```
RomuDeuxfois, line 75: Maybe : handleSwitchGroupOpMode

opencti-platform/opencti-front/src/components/FilterIconButtonContainer.tsx
```diff
@@ -0,0 +1,187 @@
...
+  handleSwitchLocalMode?: (filter: Filter) => void;
```
RomuDeuxfois, line 76: Maybe : handleSwitchFilterOpMode

opencti-platform/opencti-graphql/tests/02-integration/02-resolvers/grouping-test.js
Archidoit: the value by default is still 'eq', I wrote it here to be explicit, but we can decide to never write it when it's 'eq'

opencti-platform/opencti-graphql/config/schema/opencti.graphql
Archidoit: yes, this will be changed
future schema for all the entity types:
image

opencti-platform/opencti-graphql/config/schema/opencti.graphql
Archidoit: this will also be changed after the meeting with Julien yesterday
We want the back to return the type below when we ask for filterRepresentatives (query to fetch the values of the filters ids) --> open to discussion for the filterRepresentatives query returned type
image

opencti-platform/opencti-graphql/config/schema/opencti.graphql
Archidoit: right!!! I modify this

opencti-platform/opencti-graphql/config/schema/opencti.graphql
Archidoit: fixed

opencti-platform/opencti-graphql/tests/02-integration/02-resolvers/report-test.js
Archidoit: yes default values are still present, I will remove operator and mode specifications when they are equal to the default values

opencti-platform/opencti-graphql/tests/02-integration/02-resolvers/grouping-test.js
Archidoit: fixed

opencti-platform/opencti-graphql/src/modules/malwareAnalysis/malwareAnalysis.graphql
Archidoit: yes it has been decided with Julien: the approch will be generic, we'll handle filters key checking in the back with the schemaDefinition of the entity

opencti-platform/opencti-graphql/src/utils/filtering.js
Archidoit: it's because we define filterGroups below... 
image

opencti-platform/opencti-graphql/src/utils/filtering.js
Archidoit: it's because the key and operator are now already correct (we pass the correct format from the front)

opencti-platform/opencti-graphql/tests/02-integration/01-database/filterGroup-test.js
Archidoit: yes it's so useful to now the code is protected 😃

opencti-platform/opencti-graphql/tests/02-integration/01-database/filterGroup-test.js
Archidoit: no key = no filtering --> all the reports

opencti-platform/opencti-graphql/src/domain/stixDomainObject.js
Archidoit: I don't understand what you mean

opencti-platform/opencti-graphql/src/domain/stixCoreObject.js
Archidoit: yes! this part about filter values resolution in back will be changed completely

opencti-platform/opencti-front/src/components/FilterIconButtonContainer.tsx
Archidoit: maybe we can have a single hook for handleSwitchGlobalMode and handleSwitchLocalMode, the hook will know in which case we are according if we pass a filter argument or not

opencti-platform/opencti-graphql/src/modules/malwareAnalysis/malwareAnalysis.graphql
RomuDeuxfois: Awesome !!!

opencti-platform/opencti-front/src/components/FilterIconButtonContainer.tsx
RomuDeuxfois: Yes, it's a better approach

Issue #1865, opened 2022-02-01
Number of entities to be extended


SamuelHassine:
## Problem to Solve
 
On every element of the investigation graph that can be extended, it would be nice to have the number (next to the icon) of how many elements there are to expand it

## Current Workaround

None

PR #4400, opened 2023-09-19
[frontend/backend] Investigation graph: number of entities to be extended (Part 1) (#1865)

lndrtrbn:
### Proposed changes

* In investigation graph, display next to an entity how many entities will be displayed if we extend this entity.

### Related issues

* #1865

### Checklist

### Further comments

Adding the number of connected entities allowed to find some issues about updating the investigation graph data.
Some of those issues have been fixed directly in this PR. Nonetheless, others are not as it implies a certain time of analysis.

Known issue not fixed: counting `indicates` relationships is not working so the count for elements that have those type of relationship is inexact.


opencti-platform/opencti-front/src/private/components/workspaces/investigations/InvestigationGraph.jsx
```diff
@@ -341,6 +341,22 @@ const investigationGraphStixSightingRelationshipQuery = graphql`
     }
 `;
 
+// To count the number of relationships for MetaObjects
+const investigationGraphStixMetaRelCountQuery = graphql`
```
lndrtrbn: For meta objects we cannot use the same process as for core objects to compute the number of relationships.
So we had to use a specific query for those.

opencti-platform/opencti-front/src/private/components/workspaces/investigations/InvestigationGraph.jsx
```diff
@@ -954,6 +972,9 @@ class InvestigationGraphComponent extends Component {
       params,
     );
     const timeRangeInterval = computeTimeRangeInterval(this.graphObjects);
+
+    this.fetchMetaObjectRelCounts(this.graphObjects);
```
lndrtrbn: We fetch asynchronously the number of relationships of meta objects to avoid waiting too long to display the graph as the query can take some times.
We display a placeholder in the graph until the query is resolved.

opencti-platform/opencti-front/src/private/components/workspaces/investigations/InvestigationGraph.jsx
```diff
@@ -1497,78 +1567,60 @@ class InvestigationGraphComponent extends Component {
   }
 
   async handleDeleteSelected() {
```
lndrtrbn, line 1585: Some rework in this function to make one query with all the ids of entities to remove instead of multiple queries.
It seems that making multiple queries created issues as some were not executed (ie. when removing the selected node, both the ids of the node entity and associated relationship were sent for removal but either one or the other was not removed in ES, sometimes the node, sometimes the relationship from a test to an other).

opencti-platform/opencti-front/src/utils/Graph.js
```diff
@@ -1112,6 +1031,119 @@ export const buildGraphData = (objects, graphData, t) => {
...
+        color: n.x_opencti_color || n.color || itemColor(n.entity_type, false),
+        parent_types: n.parent_types,
+        entity_type: n.entity_type,
+        relationship_type: n.relationship_type,
+        fromId: n.from?.id,
+        fromType: n.from?.entity_type,
+        toId: n.to?.id,
+        toType: n.to?.entity_type,
+        isObservable: !!n.observable_value,
+        numberOfConnectedElement,
```
lndrtrbn, line 1122: Only this property has changed in all this block (see `let numberOfConnectedElement` above the return).
The block was moved after computing links to use the Map above.

opencti-platform/opencti-graphql/src/domain/stixCoreObject.js
```diff
@@ -252,6 +254,13 @@ export const stixCoreObjectsMultiNumber = (context, user, args) => {
   }));
 };
 
+export const stixCoreObjectsConnectedNumber = (stixCoreObject) => {
```
lndrtrbn, line 305: Counting core objects relationships using the number of properties starting with `rel_`

RomuDeuxfois: Well done

opencti-platform/opencti-front/src/private/components/workspaces/investigations/InvestigationGraph.jsx
```diff
@@ -341,6 +341,24 @@ const investigationGraphStixSightingRelationshipQuery = graphql`
     }
 `;
 
+// To count the number of relationships for MetaObjects and Identities.
+// /!\ It counts only rels that point towards given ids. So the value may
+// not be exactly the total number of rels in some cases.
```
RomuDeuxfois: Can you write the specific cases to be explicit on that point ?
Thank you

RomuDeuxfois:
Maybe I missed something in the demo
But, I have a report with 99 and more relationships, I expand it and I have this :
image

There are some other relationships (54) not displayed by the UI.

RomuDeuxfois:
Why we choose to use the wave icon istead of the relationship icon ? 
image
image

lndrtrbn:
> Maybe I missed something in the demo But, I have a report with 99 and more relationships, I expand it and I have this : image
> 
> There are some other relationships (54) not displayed by the UI.

Yep we also faced this case during development, my opinion on this one was that the count is incorrect on Report due to some kind of relations pointing toward other relations. But I can have a look again on it to try better understand what's going on here.

lndrtrbn:
> Why we choose to use the wave icon istead of the relationship icon ? image image

In fact the ~ is here to tell that it is an approximation count

RomuDeuxfois:
> > Maybe I missed something in the demo But, I have a report with 99 and more relationships, I expand it and I have this : image
> > There are some other relationships (54) not displayed by the UI.
> 
> Yep we also faced this case during development, my opinion on this one was that the count is incorrect on Report due to some kind of relations pointing toward other relations. But I can have a look again on it to try better understand what's going on here.

Great, you can use the insert:dev script to have the same report : Poison Ivy

lndrtrbn:
> > Yep we also faced this case during development, my opinion on this one was that the count is incorrect on Report due to some kind of relations pointing toward other relations. But I can have a look again on it to try better understand what's going on here.
> 
> Great, you can use the insert:dev script to have the same report : Poison Ivy

After analysis, It could be an issue with limit of results returned instead of a wrong count.
The resolution of this issue is out of scope of this PR and I will do my best to fix it with the second PR about this topic 💪

opencti-platform/opencti-front/src/private/components/workspaces/investigations/InvestigationGraph.jsx
lndrtrbn: Done

RomuDeuxfois (approved):

labo-flg:
Tested OK and code seems good (as far as I understand the platform yet 😇)

PR #4406, opened 2023-09-19
[frontend/backend] paginated objects in container + ContainerStixDomainObjects migration (#4328)

### LINKED ISSUE #4328: Only 25 items available in report--> observables and report --> entities
## Description

As you can see in the first image, there are thousands of observables and entities in the report.

thousands

But only 25 are shown in report --> Observables and report --> entities

## Environment

3 node cluster on Ubuntu bare metal installation
OpenCTI version 5.10.2

## Reproducible Steps

Go to any report with more than 25 observables/entities and move to the corresponding tab
We have tested with reports generated by us, and from other sources (AlienVault for instance)

image

## Expected Output

To be able to see all entities/observables

## Actual Output

Only 25 are shown

Archidoit:
issue: https://github.com/OpenCTI-Platform/opencti/issues/4328

PR content: 
- ContainerStixDomainObjects migration to tsx
- paginated objects in container to enable query refetch


opencti-platform/opencti-graphql/src/domain/container.js
```diff
@@ -72,15 +72,16 @@ export const objects = async (context, user, containerId, args) => {
...
+  for (let index = 0; index < elements?.length; index += 1) {
     const element = elements[index];
     const relationTypes = relationsMap.get(element.id);
     if (relationTypes) {
       const edge = { types: relationTypes, node: element, sort: element.sort };
       resultNodes.push(edge);
     }
   }
-  return buildPagination(args.first, args.after, resultNodes, resultNodes.length);
+  return buildPagination(pageInfo?.startCursor, pageInfo?.endCursor, resultNodes, pageInfo?.globalCount);
```
SouadHadjiat: why `pageInfo?.startCursor` instead of `args.first` ? and same for after ?
The first argument of `buildPagination` is `limit`, should be the number of elements we get for each page, I don't think it's `startCursor`

Archidoit: fix

opencti-platform/opencti-front/src/utils/hooks/useLocalStorage.ts
```diff
@@ -58,6 +58,8 @@ export interface UseLocalStorageHelpers {
     symbol?: string;
     original?: number;
   }) => void;
+  handleToggleTypes: (type: string) => void;
+  handleClearTypes: () => void;
```
SouadHadjiat, line 62: `handleToggleTypes` and `handleClearTypes` are declared twice

PR #4418, opened 2023-09-20
[frontend] Add classes checking lint rule to remove unused classes

Kedae:
This PR creates the first custom esLint rule to prepare future.
It also allows us to know if some classes are unused in a component and should be deleted.
The lint rule is fixable with --fix (to remove the class, but it need a review to check if we still need Theme, MakeStyles and so on)


opencti-platform/opencti-front/eslint-plugin-custom-rules/lib/rules/classes-rule.js
```diff
@@ -0,0 +1,49 @@
...
+                node: node,
+                message: "MakeStyles is used with no class declared",
+                data: {
+                  identifier: node.name
+                }
+              });
+            }
+          } else {
+            node.body.properties.forEach((v) => {
+              if (!sourceCode.text.includes("classes." + v.key.name)) {
```
lndrtrbn, line 51: Should we explicitly write in Notion that we need to use the keyword `classes` otherwise the rule is bypassed ?

lndrtrbn (commented):
Tested locally, ok for me 👍but my IDE does not underline the unused property, instead it underlines a parenthesis of `makeStyle`, I don't know if it's easy to change this, maybe it's just on my config.

image

opencti-platform/opencti-front/eslint-plugin-custom-rules/lib/rules/classes-rule.js
Kedae: We could write it but I think everyone (that has already done some front) knows that we have that convention :)

lndrtrbn: It's a possibility that yesterday I was close to use `styles` as a keyword... 😶 But yep I understand what you mean

lndrtrbn (approved):

PR #4467, opened 2023-09-28
[frontend] Cases background tasks: take filters into account (#4466)

### LINKED ISSUE #4466: Background tasks on Cases don't take filtering into account
- Go to the Case Rfis list (same for Case Rft, Case Incidents, Tasks, Feedbacks)
- Filter by a label
- Do a 'select all' via the checkbox
- Do a background task on these cases (ex: delete them)
- All the cases have been deleted (whereas only the filtered one (= the ones with the selected label in the example) should have been deleted)

Archidoit:
issue: https://github.com/OpenCTI-Platform/opencti/issues/4466


SarahBocognano (approved):
Works fine ✅

PR #4487, opened 2023-10-03
fix: prevent too_many_clauses error in query builder

alexkornitzer:
### Proposed changes

* Use `term` queries in the query builder for nested ID look ups.

### Related issues

* Certain queries would fail where there were more than `1024` relationships due to the `TooManyClause` limit error. This had major damaging effects for streams as if a stream encounters `any` error it will get stuck due to its infinite retry policy.

### Checklist

### Further comments

By default the elQueryBodyBuilder will use match_phrase for nested query building. This is fine for data that has fewer than 1024 clauses but anything larger than that will cause a TooManyClauses error. This is a problem for objects that have a large number of relationships (>1024). The fix is to make use of term based queries for keyword fields, which is what this commit does. For now this will only apply to the core ID fields as there is no list of all keywordable fields.

Furthermore terms_set can only be used on ES >8.10.0 as this adds in the `minimum_should_match` field.

### DIFF
--- a/opencti-platform/opencti-graphql/src/database/engine.js
+++ b/opencti-platform/opencti-graphql/src/database/engine.js
@@ -160,6 +160,7 @@ const searchConfiguration = {
 const elasticSearchClient = new ElkClient(searchConfiguration);
 const openSearchClient = new OpenClient(searchConfiguration);
 let isRuntimeSortingEnable = false;
+let isTermsSetUsable = false;
 let engine = openSearchClient;
 
 // The OpenSearch/ELK Body Parser (oebp)
@@ -221,6 +222,7 @@ export const searchEngineInit = async () => {
   }
   // Setup the platform runtime field option
   isRuntimeSortingEnable = enginePlatform === ELK_ENGINE && semver.satisfies(engineVersion, '>=7.12.x');
+  isTermsSetUsable = enginePlatform === ELK_ENGINE && semver.satisfies(engineVersion, '>=8.10.0');
   const runtimeStatus = isRuntimeSortingEnable ? 'enabled' : 'disabled';
   logApp.info(`[SEARCH] ${enginePlatform} (${engineVersion}) client selected / runtime sorting ${runtimeStatus}`);
   // Everything is fine, return true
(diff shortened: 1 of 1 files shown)


opencti-platform/opencti-graphql/src/database/engine.js
```diff
@@ -1336,25 +1338,43 @@ const elQueryBodyBuilder = async (context, user, options) => {
...
+          // TODO: Get full list of keywordable fields but for now IDs are the most important use case
+          const keywordable = [ID_INTERNAL, ID_STANDARD, IDS_STIX, 'connections.internal_id'];
+          if (nestedOperator !== 'wildcard' && !(localFilterMode !== 'or' && isTermsSetUsable === false) && keywordable.indexOf(nestedFieldKey) !== -1) {
+            const keywordKey = `${nestedFieldKey}.keyword`;
+            let query;
+            if (localFilterMode === 'or') {
+              query = { terms: { [keywordKey]: nestedValues.map((v) => v.toString()) } };
             } else {
-              nestedShould.push({ match_phrase: { [nestedFieldKey]: nestedSearchValues } });
+              query = { terms_set: { [keywordKey]: { terms: nestedValues.map((v) => v.toString()), minimum_should_match: nestedValues.length } } };
```
SouadHadjiat, line 1351: terms_set doesn't seem to work with `minimum_should_match` : 
`[terms_set] query does not support [minimum_should_match]`

According to the [documentation](https://www.elastic.co/guide/en/elasticsearch/reference/8.10/query-dsl-terms-set-query.html) : 

> The required number of matching terms is defined in the minimum_should_match_field or minimum_should_match_script parameter.

alexkornitzer: Ah damn, did they get their release notes wrong or did I mess up the version check? (https://www.elastic.co/guide/en/elasticsearch/reference/current/release-notes-8.10.0.html).

I will take that query offline and test it against local elasticsearch releases and see what I find.

SouadHadjiat: Oh I can see it has been added recently https://github.com/elastic/elasticsearch/pull/96082, but the documentation has not been updated it seems.

alexkornitzer: Yeah they sort of ninja added it, took me a while to find it when I was hunting for a solution in the first place.

SouadHadjiat: I tested it on my local elasticsearch 8.9.1, not on 8.10, that's why I got the error. I'm going to test it on 8.10,

alexkornitzer: So my version check is not working? (line 255)

SouadHadjiat: So the fix will only work on elasticsearch >= 8.10 ? otherwise we use the old match_phrase query ?

SouadHadjiat: > So my version check is not working? (line 255)

no no, I tested directly by quering my database, not with the code.

alexkornitzer: So it won't work on opensearch either until I can work out what version they have added it in as well, but that can be a follow up. Assuming you guys are happy with the PR.

alexkornitzer: I think OpenSearch adds the feature in `2.1` would you like me to check that and get that in the PR too?

SouadHadjiat: I wonder if it could be possible to use `minimum_should_match_script` which is already supported by older elasticsearch versions and opensearch : 
https://www.elastic.co/guide/en/elasticsearch/reference/8.10/query-dsl-terms-set-query.html#terms-set-query-script
https://opensearch.org/docs/2.0/opensearch/query-dsl/term/#terms-set

like this : 
```
"minimum_should_match_script": {
  "source": "params.num_terms"
}
```

alexkornitzer: Its doable (I think) but I don't know what the performance impact would be. I could try and implement it as a fallback for old elastics?

SouadHadjiat: I think we could use only "minimum_should_match_script" without having a test on elasticsearch/opensearch version, the code will be simpler

alexkornitzer: I would need to go and dig around the elastic source. But I dont like the idea of script evaluation each query which is why the feature was added. And would say the complexity is better to prevent performance impact. But it's not my product so am happy to do as directed.

SouadHadjiat: Well, I didn't spend enough time on it to be able to tell what is the best solution, I only thought that it's better to implement a fix that works for all database versions. What performance issue does "minimum_should_match_script" have ?

alexkornitzer: I am quite busy this week, but I will take this offline and get back to you with any findings.

alexkornitzer: @SouadHadjiat based on this [ES Docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/scripts-and-search-speed.html#_improving_search_speed). I don't think its a good idea to use the script solution as a default. I still think versioning is a better solution but also acknowledge that the `and` use case is probably a rare one. Basically in our deployment the `too_many_clauses` is a huge issue for objects with many relations which are `or` based queries.

richard-julien:
First of all thanks for the problem raising and your PR @alexkornitzer.
Based on your work we decide to do a bit differently on the resolution. You can see the work in https://github.com/OpenCTI-Platform/opencti/pull/4744.
Basically using a better function to resolve the ids preventing this limitation. 
Closing your PR in favor of this one.
Thanks.

alexkornitzer:
@richard-julien, no worries at all, I don't know the underlying data well enough so just took a stab. I should really have raised an issue first to prevent wasted effort. Thanks for getting a fix in place.

PR #4492, opened 2023-10-04
[frontend] Added padding to toolbar to not overlap with FAB

Bonsai8863:
### Proposed changes

* Added padding to toolbars so as to not overlap with the floating action button

### Checklist


richard-julien:
We would like to have a better approach where the button is moved when the bar is opened / closed.
Do you think you can take a look to modify the PR to target this behavior?
Thanks

PR #4524, opened 2023-10-09
[frontend/backend] Investigation graph: number of entities to be extended (Part 2) (#1865)

lndrtrbn:
### Proposed changes

* In investigation graph, modify the expand form to :
  * Display only types of targets and relationships that have something to expand,
  * Display the number of elements of those types that will be added to the graph if expanded,
  * For the UI, use checkboxes instead of a select input,
  * Remove the field `limit` as now we can see how much we will expand.

### Related issues

* #1865 

### Checklist

### Further comments

To separate concerns, the part of the code about the expand form has been moved from `InvestigationGraph.jsx` into its own component `InvestigationExpandForm.tsx`.

Deep dive into `InvestigationExpandForm.tsx`:

- In this component we make 3 different queries based on `stixRelationshipsDistribution`:
  - One to count the number of elements targeted by a relation where the source is one of the given ids, the result is multiple counts grouped by type of elements,
  - Second to count the number of elements targeting one of the given ids (ie. sources of relations where the target is one of given ids), the result is multiple counts grouped by type of elements,
     > A first approach was to get the counts by doing only one query that fetch both directions of a relation at the same time.
     > But we had issues when there is relations directly between the elements given as parameters of the query (double counts or total exclusion depending on the variables) 
  - Third to count directly the relations themselves instead of extremities, grouped by type of relation.
- In the component we have 3 `useEffects`:
  - The first one is listening to changes on links already displayed on the graph and base don those links compute 2 Maps that are here to help us filter the counts returned by the backend with what is already in the graph,
  - The second one is responsible of fetching the count on relations and filter the result when the selected nodes change,
  - The last one is responsible of fetching the count on relations targets and filter the result when the selected nodes change.
- The component use the new `CheckboxesField` component to display computed targets and relationships.

--

Remove component `TytpesField` that was used only in the old expand form.

--

Some changes into `Graph.js` to improve what is displayed in the bullet next to the nodes in the graph.

--

Adding a new option `aggregateOnConnections` on query to count relationships to be able to count directly the number of relationships instead of extremities.


opencti-platform/opencti-front/src/components/CheckboxesField.tsx
```diff
@@ -0,0 +1,93 @@
...
+  value: string
+};
+
+type Props = FieldProps<CheckboxesItem[]> & {
+  label: string
+  items: CheckboxesItem[]
+  maxHeight?: string
+};
+
+export default function CheckboxesField({
```
Kedae: Not the choosen way of declaring a function

opencti-platform/opencti-front/src/components/CheckboxesField.tsx
```diff
@@ -0,0 +1,93 @@
...
+  form,
+  field,
+  label,
+  items,
+  maxHeight = '300px',
+}: Props) {
```
Kedae: Props should have the ComponentName prefixing it

opencti-platform/opencti-front/src/components/CheckboxesField.tsx
```diff
@@ -0,0 +1,93 @@
...
+  const { t } = useFormatter();
+
+  const { setFieldValue } = form;
+  const { name, value } = field;
+
+  function isChecked(val: CheckboxesItem) {
```
Kedae: Same

opencti-platform/opencti-front/src/utils/Graph.js
```diff
@@ -1190,13 +1190,15 @@ export const nodePaint = (
...
+    ctx.arc(x + 4, y - 5, 4, 0, 2 * Math.PI, false);
     ctx.fill();
     ctx.fillStyle = '#000';
     let numberLabel = '?';
     if (numberOfConnectedElement !== undefined) numberLabel = numberOfConnectedElement;
-    if (numberOfConnectedElement > 99) numberLabel = 99;
-    if (numberLabel !== '?') numberLabel = `${numberLabel}~`;
+    if (numberLabel !== '?') {
+      if (numberOfConnectedElement > 99) numberLabel = '99+';
+      else numberLabel = `~${numberLabel}`;
```
Goumies: numberLabel = numberOfConnectedElement > 99
  ? `99+`
  : `~${numberLabel}`;

opencti-platform/opencti-front/src/private/components/workspaces/investigations/InvestigationExpandForm.tsx
```diff
@@ -0,0 +1,403 @@
...
+    setExistingRels(countRels);
+  }, [links, setExistingRels, setExistingTargets]);
+
+  // Fetch all possible relationships linked to the nodes to expand.
+  useEffect(() => {
+    async function fetchRelationships() {
+      if (selectedNodes.length === 0) return;
+      const nodeIds = Array.from(selectedNodes).map((node) => node.id);
+
+      const { stixRelationshipsDistribution } = await fetchQuery(
```
Kedae: I don't think this is the best way of doing it. We can discuss it if you want but yo should use a relay hook to perform a query from a ref already declared

opencti-platform/opencti-front/src/components/CheckboxesField.tsx
```diff
@@ -0,0 +1,93 @@
+import { FieldProps } from 'formik';
+import FormControl from '@mui/material/FormControl';
+import { ButtonGroup, FormLabel } from '@mui/material';
+import FormGroup from '@mui/material/FormGroup';
+import FormControlLabel from '@mui/material/FormControlLabel';
```
guillaumejparis, line 5:
```suggestion
import { FieldProps } from '';
import FormControl from '@mui/material/';
import { ButtonGroup, FormLabel } from '@mui/';
import FormGroup from '@mui/material/';
import FormControlLabel from '@mui/material/';
```

opencti-platform/opencti-front/src/components/CheckboxesField.tsx
lndrtrbn: Done

opencti-platform/opencti-front/src/components/CheckboxesField.tsx
lndrtrbn: Done

opencti-platform/opencti-front/src/components/CheckboxesField.tsx
lndrtrbn: Done

opencti-platform/opencti-front/src/utils/Graph.js
lndrtrbn: I have made the changes. But I think it will be interesting to have a discussion between us (all devs) to determine what we expect in term of level of granularity in code review. 😊

Kedae:
Remove bubble from Knowledge graph

lndrtrbn:
### PR Status

**Reviewed by :** Romy, Adrien, Guillaume
**In test by :** Frederic

**To be done :**
Remove bubble from Knowledge graph -> ready
Rework Relay queries -> tomorrow morning

opencti-platform/opencti-front/src/components/CheckboxesField.tsx
```diff
@@ -0,0 +1,89 @@
...
+import Checkbox from '@mui/material/Checkbox';
+import Button from '@mui/material/Button';
+import { useFormatter } from './i18n';
+
+export type CheckboxesItem = {
```
Kedae: Option type is already declared somewhere :D

Kedae (approved):

guillaumejparis (approved):

PR #4526, opened 2023-10-09
[backend] Restrict organization access capability check failing (#4509)

### LINKED ISSUE #4509: Restrict organization access capability check failing
## Description

Users with "Restrict organization access" are unable to perform organization sharing. When attempted the message "You are not allowed to do this appears" instead of success.

## Environment

1. OS Ubuntu
2. OpenCTI version: OpenCTI 5.10.3
3. OpenCTI client: front-end
4. Other environment details:

## Reproducible Steps

1. Standup fresh instance of OpenCTI with docker-compose
2. Create role with 'Restrict organization access' capability, add it to a group, create a user and add to that group.
3. Create an organization
4. Create a report
5. Attempt to share the report to the organization

## Expected Output

Successful share

## Actual Output

A message appears "You are not allowed to do this"

## Additional information

After some digging it appears the capability check at:
https://github.com/OpenCTI-Platform/opencti/blob/5.10.3/opencti-platform/opencti-graphql/src/database/middleware.js#L1973
is failing.

The call to !userHaveCapability(user, KNOWLEDGE_ORGANIZATION_RESTRICT)
is passing 'KNOWLEDGE_KNUPDATE_KNORGARESTRICT' imported from opencti-graphql/src/utils/access.ts

but the userHaveCapability function is splitting up the values in user.capabilities such that it is trying to match that string against
one of:
['KNOWLEDGE',      'KNUPDATE',        'KNORGARESTRICT']
after the calls to split('_') and flatten().

If KNOWLEDGE_ORGANIZATION_RESTRICT is imported from opencti-graphql/src/schema/general.js
it will be set to 'KNORGARESTRICT' which will successfully match and this appears to work if I make the change locally. I don't believe it would interfere with other uses of that constant in middleware.js.

marieflorescontact:
### Proposed changes

* Use isUserHasCapability instead of userHaveCapability
*

### Related issues

* [issue/4509](https://github.com/OpenCTI-Platform/opencti/issues/4509)
*

### Checklist

### Further comments

### DIFF
--- a/opencti-platform/opencti-graphql/src/database/middleware.js
+++ b/opencti-platform/opencti-graphql/src/database/middleware.js
@@ -177,7 +177,19 @@ import {
 } from '../utils/format';
 import { checkObservableSyntax } from '../utils/syntax';
 import { deleteAllFiles, storeFileConverter, upload } from './file-storage';
-import { BYPASS, BYPASS_REFERENCE, executionContext, INTERNAL_USERS, isBypassUser, isUserCanAccessStoreElement, KNOWLEDGE_ORGANIZATION_RESTRICT, RULE_MANAGER_USER, SYSTEM_USER, userFilterStoreElements, validateUserAccessOperation } from '../utils/access';
+import {
+  BYPASS_REFERENCE,
+  executionContext,
+  INTERNAL_USERS,
+  isBypassUser,
+  isUserCanAccessStoreElement,
+  isUserHasCapability,
+  KNOWLEDGE_ORGANIZATION_RESTRICT,
+  RULE_MANAGER_USER,
+  SYSTEM_USER,
+  userFilterStoreElements,
+  validateUserAccessOperation
+} from '../utils/access';
 import { isRuleUser, RULES_ATTRIBUTES_BEHAVIOR } from '../rules/rules';
 import { instanceMetaRefsExtractor, isSingleRelationsRef, } from '../schema/stixEmbeddedRelationship';
 import { createEntityAutoEnrichment } from '../domain/enrichment';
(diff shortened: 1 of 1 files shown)


opencti-platform/opencti-graphql/src/database/middleware.js
```diff
@@ -1933,7 +1945,7 @@ export const updateAttribute = async (context, user, id, type, inputs, opts = {}
         }
       } else {
         // Special access check for RELATION_GRANTED_TO meta
-        if (relType === RELATION_GRANTED_TO && !userHaveCapability(user, KNOWLEDGE_ORGANIZATION_RESTRICT)) {
+        if (relType === RELATION_GRANTED_TO && !isUserHasCapability(user, KNOWLEDGE_ORGANIZATION_RESTRICT)) {
```
Goumies, line 1948: Can we refactor `isUserHasCapability` to `doesUserHasCapability` ?

marieflorescontact: I don't know because the method has many uses and also because in the platform we have several different methods based on this "isUser..." model such as `isUserHasAccessToElemen`t or `isUserHasCapability`. But you're right, it should be `does` instead of `is`

guillaumejparis: Mmm problem is we have this syntax on others variables for example : 
  isBypassUser
  isUserCanAccessStoreElement
  isUserHasCapability
If we want to harmonize everything it will be a full issue I suppose

marieflorescontact: yes I agree with you @guillaumejparis  that's what it seems to me too

RomuDeuxfois (approved):

Issue #2513, opened 2022-10-28
Be able to add Analystics pixels tracking systems for demographics


SamuelHassine:
## Use case

Be able to add Analystics pixels tracking systems for demographics in platforms.

PR #4530, opened 2023-10-09
[frontend] Add analytics basic system + google analytics v4 (#2513)

guillaumejparis:


opencti-platform/opencti-front/src/private/Root.tsx
```diff
@@ -150,6 +155,19 @@ const RootComponent: FunctionComponent<RootComponentProps> = ({ queryRef }) => {
...
   const bannerSettings = computeBannerSettings(settings);
   const platformModuleHelpers = platformModuleHelper(settings);
+
+  useEffect(() => {
+    setAnalyticsInstance(Analytics({
+      app: 'opencti',
+      plugins: [
+        googleAnalytics({
+          measurementIds: [queryData.settings.analytics_google_analytics_v4],
+          enabled: !!queryData.settings.analytics_google_analytics_v4,
```
guillaumejparis, line 165: At first we have only one plugin : google analytics, this is optin if no measurementId is supplied the plugin is inactive

opencti-platform/opencti-front/src/private/components/nav/TopBar.tsx
```diff
@@ -292,6 +294,9 @@ const TopBarComponent: FunctionComponent<TopBarProps> = ({
       sub.unsubscribe();
     };
   });
+  useEffect(() => {
+    page();
+  }, [location.pathname]);
```
guillaumejparis, line 299: We put the tracker in the TopBar so at every page change we can do a "page view" action, it permits to track the user journey trough the app

opencti-platform/opencti-front/src/private/components/settings/Settings.jsx
```diff
@@ -307,7 +311,8 @@ const Settings = () => {
                       style={{ marginTop: 15 }}
                     >
                       <Formik
-                        onSubmit={() => {}}
+                        onSubmit={() => {
+                        }}
```
guillaumejparis, line 315: When I use our Project.xml definition for code styling it style like this ¯\_(ツ)_/¯

opencti-platform/opencti-front/src/private/components/settings/settings_analytics/SettingsAnalytics.tsx
```diff
@@ -0,0 +1,104 @@
...
+                onFocus={(name: string) => handleChangeFocus(id, name)}
+                onSubmit={(name: string, value: string | null) => handleSubmitField(id, name, value)}
+                variant="standard"
+                helperText={
+                  <SubscriptionFocus
+                    context={editContext}
+                    fieldName="analytics_google_analytics_v4"
+                  />
+                }
+              />
```
guillaumejparis, line 97: only a field at the moment because we only supports google analytics v4

opencti-platform/opencti-front/src/utils/global.d.ts
```diff
@@ -0,0 +1,24 @@
...
+  export const AnalyticsConsumer: Context<AnalyticsInstance>['Consumer'];
+  export const AnalyticsContext: Context<AnalyticsInstance>;
+
+  export function AnalyticsProvider(props: {
+    instance: AnalyticsInstance;
+    children: ReactNode;
+  }): JSX.Element;
+}
+
+declare module '@analytics/google-analytics' ;
```
guillaumejparis, line 24: adding global typings because these libs do not have typîngs

opencti-platform/opencti-front/src/private/Root.tsx
```diff
@@ -1,6 +1,9 @@
 import { StyledEngineProvider } from '@mui/material/styles';
-import React, { FunctionComponent } from 'react';
+import React, { FunctionComponent, useEffect, useState } from 'react';
 import { graphql, usePreloadedQuery, PreloadedQuery } from 'react-relay';
+import { AnalyticsProvider } from 'use-analytics';
+import googleAnalytics from '@analytics/google-analytics';
+import Analytics from 'analytics';
```
guillaumejparis, line 6: analytics lib works with a plugin system so if needed we can later add more analytiucs providers

Goumies (approved):

opencti-platform/opencti-front/src/private/Root.tsx
```diff
@@ -150,6 +155,19 @@ const RootComponent: FunctionComponent<RootComponentProps> = ({ queryRef }) => {
   // TODO : Use the hook useHelper when all project is pure function //
   const bannerSettings = computeBannerSettings(settings);
   const platformModuleHelpers = platformModuleHelper(settings);
+
+  useEffect(() => {
+    setAnalyticsInstance(Analytics({
+      app: 'opencti',
+      plugins: [
+        googleAnalytics({
+          measurementIds: [queryData.settings.analytics_google_analytics_v4],
```
lndrtrbn, line 164: If we don't have settings for google analytics then `measurementIds` equals `[undefined]` is it an issue ?

opencti-platform/opencti-front/src/private/components/settings/Settings.jsx
lndrtrbn: We can check together if you want

opencti-platform/opencti-front/src/private/Root.tsx
guillaumejparis: Nop because the plugin is not enabled in this case -> enabled: !!queryData.settings.analytics_google_analytics_v4,

lndrtrbn (approved):

Issue #4540, opened 2023-10-11
KillChain view of Attack Pattern Knowledge : error occurs


Archidoit:
- Go to Event > Incidents
- Select an incident
- Go to the Knowledge tab
- Go to the 'kill chain view'
- Add an attack pattern if there is none
- An error occurs

PR #4541, opened 2023-10-11
[frontend] StixDomainObjectAttackPatternsKillChain bug fix (#4540)

Archidoit:
issue: https://github.com/OpenCTI-Platform/opencti/issues/4540


opencti-platform/opencti-front/src/private/components/common/stix_domain_objects/StixDomainObjectAttackPatternsKillChain.jsx
```diff
@@ -272,7 +272,7 @@ class StixDomainObjectAttackPatternsKillChainComponent extends Component {
               entityLink={entityLink}
               paginationOptions={paginationOptions}
               onDelete={this.props.relay.refetch.bind(this)}
-              searchTerm={searchTerm}
+              searchTerm={searchTerm ?? ''}
```
RomuDeuxfois: Shouldn't this be handled by the view storage of the parent component's StixDomainObjectAttackPatterns?
Or I'm missing something.

Archidoit: It's not handled by view storage, but can be handled by the parent right!

Archidoit: fixed

RomuDeuxfois (approved):

PR #4545, opened 2023-10-11
[frontend/backend] Have an administrator for an organization (#3637)

### LINKED ISSUE #3637: Be able to have an administrator for an organization who is able to manage users inside its organization
## Use case

Be able to have an administrator for an organization who is able to manage users inside its organization.

yassine-ouaamou:
### Proposed changes

* Add new (virtual) capability of organization admin
* A plateform can promote a member of an organization as an "organization admin"
* An organization admin is capable of viewing the organizations he is admin of and its members
* An organization admin is able to create a new user who will be part of his organization 
* An organization admin is able to edit members of his organization

### Edited APIs (to check please)
- [ ]  users
- [ ]  user
- [ ]  userAdd
- [ ] userEdit
- [ ]  groups
- [ ]  group
- [ ]  organization{members}

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/3637


opencti-platform/opencti-front/src/components/list_lines/ListLinesContent.jsx
```diff
@@ -190,6 +192,8 @@ class ListLinesContent extends Component {
             contentMapping={contentMapping}
             contentMappingData={contentMappingData}
             bypassEditionRestriction={bypassEditionRestriction}
+            isOrganizationAdmin={isOrganizationAdmin}
```
SouadHadjiat: where do you need this property ? isOrganizationAdmin, same for organization

yassine-ouaamou: organization is needed in the line component (SettingsOrganizationUserLine).
isOrganizationAdmin is not used anymore. I will delete it

SouadHadjiat: still, it's a bit weird to have `organization` in a generic component used for all entities list. I wonder if there is another way to have this information in SettingsOrganizationUserLine

SouadHadjiat: but I see that there are already other spectific properties too...

opencti-platform/opencti-graphql/src/modules/attributes/internalObject-registrationAttributes.ts
```diff
@@ -101,6 +101,7 @@ const internalObjectsAttributes: { [k: string]: Array<AttributeDefinition> } = {
     { name: 'default_time_field', type: 'string', mandatoryType: 'no', multiple: false, upsert: false },
     { name: 'account_status', type: 'string', mandatoryType: 'external', multiple: false, upsert: false },
     { name: 'account_lock_after_date', type: 'date', mandatoryType: 'no', multiple: false, upsert: false },
+    { name: 'administrated_organizations', type: 'string', mandatoryType: 'no', multiple: true, upsert: false },
```
SouadHadjiat: do we need to declare it as attribute ? since it's not stored, only get through graphql API

opencti-platform/opencti-front/src/private/components/common/form/GroupField.jsx
```diff
@@ -53,18 +53,22 @@ class GroupField extends Component {
...
-          pathOr([], ['groups', 'edges']),
-          map((n) => ({
-            label: n.node.name,
-            value: n.node.id,
-          })),
-        )(data);
-        this.setState({ groups });
-      });
+    if (this.props.predefinedGroups) {
+      this.setState({ groups: this.props.predefinedGroups });
```
SouadHadjiat, line 57: does it mean that there is no search when we have predefinedGroups ? it will always display the same list ?

opencti-platform/opencti-graphql/src/domain/user.js
```diff
@@ -126,11 +98,30 @@ const extractTokenFromBasicAuth = async (authorization) => {
 
 export const findById = async (context, user, userId) => {
   const data = await storeLoadById(context, user, userId, ENTITY_TYPE_USER);
+  if (!isUserHasCapability(user, SETTINGS_SET_ACCESSES) && user.id !== userId) {
```
SouadHadjiat, line 115: does it mean that an orga admin can administer itself ?

opencti-platform/opencti-front/src/components/list_lines/ListLinesContent.jsx
yassine-ouaamou: Yep! I agree that it is not the best way

opencti-platform/opencti-graphql/src/domain/user.js
yassine-ouaamou: Since we use findById to get the logged user (and get its capas, organizations, administrated_organizations through buildCompleteUser), this condition was added to not check that the fetched user has common orgas with the logged user if it's himself.
In the line 103, user.administrated_organizations was empty and we ended up with ForbiddenAccess each time

opencti-platform/opencti-front/src/private/components/common/form/GroupField.jsx
yassine-ouaamou: Yep. You can still search but only within the pre defined list of groups

opencti-platform/opencti-front/src/components/list_lines/ListLinesContent.jsx
Kedae: Stepping into : you should use the generic entityId attribute of listLines for that

opencti-platform/opencti-graphql/src/domain/user.js
```diff
@@ -425,6 +416,13 @@ export const addUser = async (context, user, newUser) => {
   if (existingUser) {
     throw FunctionalError('User already exists', { email: userEmail });
   }
+  if (!isUserHasCapability(user, SETTINGS_SET_ACCESSES)) {
+    // user is Organization Admin
+    const groupIds = R.uniq(user.administrated_organizations.map((orga) => orga.grantable_groups).flat());
+    if (!newUser.groups.every((group) => groupIds.includes(group))) {
+      throw ForbiddenAccess();
```
SouadHadjiat, line 442: don't we check that the `newUser` is also part of an administrated organization ? I see only a check on grantable groups, but I guess we could check first the organizations then test these organizations grantable groups

SouadHadjiat: also how does it behave if newUser.groups is empty ? can an admin orga create a new user without any group ? and any orga ? (via the API)

opencti-platform/opencti-graphql/src/domain/user.js
SouadHadjiat: so the admin orga can administer itself ? I mean he can edit his own user, since userEdit is accessible with orga admin capability.

opencti-platform/opencti-graphql/src/modules/organization/organization-resolver.ts
```diff
@@ -36,6 +37,10 @@ const organizationResolvers: Resolvers = {
     subOrganizations: (organization, _, context) => subOrganizationsLoader.load(organization.id, context, context.user),
     parentOrganizations: (organization, _, context) => parentOrganizationsLoader.load(organization.id, context, context.user),
     default_dashboard: (current, _, context) => loadByIdLoader.load(current.default_dashboard, context, context.user),
+    grantable_groups: (organization, _, context) => findGrantableGroups(context, context.user, organization),
```
SouadHadjiat, line 40: I think it needs to be batched, otherwise organization list will be very slow to load. and we also get roles in each groups, which needs to be batched too (group resolver does that, but here we don't)

SouadHadjiat: Oh I see that SettingsOrganizationLine_node doesn't try to get grantable_groups, it was just slow for me locally, not sure then if it's necessary

SouadHadjiat:
I can't access user admin view (from users list) as orga admin if enterprise edition is enabled, I get a forbidden error for this query :
```
data: {auditsTimeSeries: null}
errors: [{message: "You are not allowed to do this.", name: "ForbiddenAccess",…}]
```
Capture d'écran 2023-10-13 111829

opencti-platform/opencti-graphql/src/domain/user.js
yassine-ouaamou: I fixed this and now we check if groups or orgas are not empty and if not, we check that it's eligible groups/orgas

opencti-platform/opencti-graphql/src/modules/attributes/internalObject-registrationAttributes.ts
yassine-ouaamou: I agree

SouadHadjiat:
As an organization admin, I can add to myself (and to any other user, even a user that I cannot administer) any group with this mutation :
 
```
mutation {
  userEdit(id: "USER_ID") {
    relationAdd(
      input: {
        toId: "GROUP_ID",
        relationship_type: "member-of"
    	}
    ) {
      id
    }
  }
}
```

SouadHadjiat (changes requested):
Address security issues with userEdit

yassine-ouaamou:
> I can't access user admin view (from users list) as orga admin if enterprise edition is enabled, I get a forbidden error for this query :
> 
> ```
> data: {auditsTimeSeries: null}
> errors: [{message: "You are not allowed to do this.", name: "ForbiddenAccess",…}]
> ```
> 
> Capture d'écran 2023-10-13 111829

fixed!

yassine-ouaamou:
> Address security issues with userEdit

The following security issues have been adressed:
- adding groups that are not in grantable_groups to a user
- adding non administrated organizations to a user
- delete relation between user and administration
- updating a field of a user that's who is in a different organization

opencti-platform/opencti-graphql/src/domain/user.js
yassine-ouaamou: I added some security checks (in the API) and now, an organization admin can administrate itself but won't be able to change group or organization

PR #4549, opened 2023-10-12
[frontend] Update the display of new widget (#4529)

### LINKED ISSUE #4529: Wrong location for new dashboard widgets
## Description

When a new widget is created on a dashboard, it is not placed in the desired/logical location (i.e. the end of the dashboard). But in the middle of the other widgets, more precisely, on the second line on the far left.

## Environment

OpenCTI 5.10.3

## Reproducible Steps

1. Create multiple widgets to create a dashboard with multiple rows and columns
2. Add a new widget
3. See where it has been placed in the dashboard

## Expected Output

The new widget should be placed at the end of the dashboard/last position - i.e. on the last line and on the right OR on a new line below.

## Actual Output

The newly created widget is placed in the middle of the other widgets, more precisely, on the second row on the far left.

## Screenshots 

Dashboard_before_new_widget
Dashboard_after_new_widget

Kedae:


marieflorescontact (approved):
Locally tested 👌

opencti-platform/opencti-front/src/private/components/workspaces/dashboards/Dashboard.jsx
```diff
@@ -140,7 +133,28 @@ const DashboardComponent = ({ workspace, noToolbar }) => {
...
+    );
+    return maxY;
+  };
+  const getMaxX = () => {
+    const y = getMaxY();
+    const maxX = Object.values(manifest.widgets).filter((n) => n.layout.y === y).reduce(
+      (max, n) => (n.layout.x > max ? n.layout.x : max),
+      0,
+    );
+    return maxX + 4;
```
lndrtrbn, line 148: The hard coded `4` works only when the widget on the left of the one added is 4 cell wide.
If the widget is wider then the new goes below it. If it is narrower there is a space between both widgets.

I think we need to use `layout.w` in the reduce function to find correct `maxX`.

And when the widget is added on aa new line there is an offset on the left also.

opencti-platform/opencti-front/src/private/components/workspaces/dashboards/Dashboard.jsx
```diff
@@ -140,7 +133,28 @@ const DashboardComponent = ({ workspace, noToolbar }) => {
     }
     saveManifest(newManifest);
   };
+  const getMaxY = () => {
+    const maxY = Object.values(manifest.widgets).reduce(
+      (max, n) => (n.layout.y > max ? n.layout.y : max),
+      0,
+    );
+    return maxY;
```
lndrtrbn:
```suggestion
    return Object.values(manifest.widgets).reduce(
      (max, n) => (n.layout.y > max ? n.layout.y : max),
      0,
    );
```

Eslint was not happy

lndrtrbn (changes requested):

opencti-platform/opencti-front/src/private/components/workspaces/dashboards/Dashboard.jsx
lndrtrbn: I added screen captures but they disappeared 😒

lndrtrbn: Narrower
image

Wider
image

New line
image

Issue #4501, opened 2023-10-05
A table too large is cut off in the pdf export


Lhorus6:
## Description

Exporting files containing large tables does not work correctly (i.e. tables are cut off).

## Reproducible Steps

1. Create a report, then go to its "Content" tab
2. Create an HTML file and add a table with several columns
3. Fill in the columns to make the table very large
4. Export the file in pdf format
=> The table not fully displayed (see screenshot attached)

<img width="962" alt="Table_in_OCTI" src="https://github.com/OpenCTI-Platform/opencti/assets/70956706/23979f15-af44-4927-89de-b35522a5c89b">
<img width="652" alt="Table_cut" src="https://github.com/OpenCTI-Platform/opencti/assets/70956706/e4ece5c5-bbbe-456d-b8c8-7d868ae2480c">

marieflorescontact:
to solve this bug properly, the user needs to be able to set the format and orientation of the pdf before exporting it. We need to discuss this @Jipegien  and @nino-filigran 💐

nino-filigran:
@Lhorus6 we'll provide a quick fix for this: basically if your content is larger than an A4 page in protrait mode, we'll switch to landscape mode when you want to print it. However if the text is bigger than the landscape, it would still be cropped. This is an easy fix before we do create a more sustainable feature around this. We'll track the implementation of this feature in here https://github.com/OpenCTI-Platform/opencti/issues/4862 and we'll close your bug as soon as the PR will be merged.

Lhorus6:
That's great thank you!

PR #4632, opened 2023-10-17
[frontend] Fix tables wider than page width when exporting to pdf (#4501)

marieflorescontact:
### Proposed changes

adjusts page orientation according to element width
### Related issues

* [issue/4501](https://github.com/OpenCTI-Platform/opencti/issues/4501)

### Checklist

### Further comments

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/common/stix_domain_objects/StixDomainObjectContent.jsx
+++ b/opencti-platform/opencti-front/src/private/components/common/stix_domain_objects/StixDomainObjectContent.jsx
@@ -428,10 +428,10 @@ class StixDomainObjectContentComponent extends Component {
       )(ret.images);
 
       let pdfElementMaxWidth = 0;
-      // We need to get the children width in order to know which mode we should save the PDF
+      // We need to get tables width inside ckeditor in order to know which mode we should save the PDF
       const elementCkEditor = document.querySelector('.ck-content.ck-editor__editable.ck-editor__editable_inline');
       if (elementCkEditor) {
-        Array.from(elementCkEditor.children).forEach((c) => {
+        Array.from((elementCkEditor.querySelectorAll('figure.table') ?? [])).forEach((c) => {
           if (c.offsetWidth > pdfElementMaxWidth) {
             pdfElementMaxWidth = c.offsetWidth;
           }
--- a/opencti-platform/opencti-front/src/private/components/common/stix_domain_objects/StixDomainObjectContent.jsx
+++ b/opencti-platform/opencti-front/src/private/components/common/stix_domain_objects/StixDomainObjectContent.jsx
@@ -430,16 +430,19 @@ class StixDomainObjectContentComponent extends Component {
       let pdfElementMaxWidth = 0;
       // We need to get the children width in order to know which mode we should save the PDF
       const elementCkEditor = document.querySelector('.ck-content.ck-editor__editable.ck-editor__editable_inline');
-      Array.from(elementCkEditor.children).forEach((c) => {
-        if (c.offsetWidth > pdfElementMaxWidth) {
-          pdfElementMaxWidth = c.offsetWidth;
-        }
-      });
+      if (elementCkEditor) {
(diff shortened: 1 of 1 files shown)


opencti-platform/opencti-front/src/private/components/common/stix_domain_objects/StixDomainObjectContent.jsx
```diff
@@ -365,6 +365,7 @@ class StixDomainObjectContentComponent extends Component {
       const pdfData = {
         content: ret.content,
         images,
+        pageSize: 'A2',
```
SarahBocognano: Is A2 the largest we can provide ?

SouadHadjiat: So A2 will be the size by default ? I'm not sure it's something we want, if you write a simple text report, it will be exported in A2 format, do we want that ?

SouadHadjiat: this is what the page look like in A2 :
Capture d'écran 2023-10-18 162816

comparing to default A4 : 
Capture d'écran 2023-10-18 162801

marieflorescontact: It's the only solution I've found, but if you have a better one, I'd love to have you're advice. On an A2 the advantage is that the table is fully rendered, the document can then be resized, cropped etc.
@SarahBocognano the widest we can supply is A0, which seems excessive to me.

opencti-platform/opencti-front/src/private/components/common/stix_domain_objects/StixDomainObjectContent.jsx
```diff
@@ -362,9 +362,20 @@ class StixDomainObjectContentComponent extends Component {
           ])),
         R.fromPairs,
       )(ret.images);
+
+      let pdfElementMaxWidth = 0;
+      // We need to get the children width in order to know which mode we should save the PDF
+      const elementCkEditor = document.querySelector('.ck-content.ck-editor__editable.ck-editor__editable_inline');
```
jpkha, line 432: We could not find a better way to determine the width of the element. `CKEditor` did not provide the width of the element, and neither did `htmlToPdfmake`.  Moreover, we used a querySelector to retrieve the content inside CKEditor and checked the children's width to determine the maxWidth. Depending on the result, we choose between landscape or portrait for the PDF printing

labo-flg:
No easy tradeoff, but this solution seems good to me !

labo-flg (approved):

opencti-platform/opencti-front/src/private/components/common/stix_domain_objects/StixDomainObjectContent.jsx
```diff
@@ -362,9 +362,20 @@ class StixDomainObjectContentComponent extends Component {
...
+      Array.from(elementCkEditor.children).forEach((c) => {
```
SouadHadjiat: Are we sure here that `elementCkEditor` is defined ? document.querySelector can return a null value, can we add a test to make sure we don't have an error ?

lndrtrbn (commented):
My exported PDF seems to be always in landscape. Any idea?

image

HTML Code:
```
<h2>Welcome to My Website</h2>
<figure class="table">
    <table>
        <tbody>
            <tr>
                <td>efkjlhwebfkjhbgfe</td>
                <td>ewbfiewfkfjbwef</td>
                <td>efdbfwqeb</td>
                <td>wefbqewlbfw</td>
                <td>ewfbfeebf</td>
            </tr>
        </tbody>
    </table>
</figure>
```

marieflorescontact:
> My exported PDF seems to be always in landscape. Any idea?

the breakpoint between portrait and landscape mode is set for a size of 680 pixels:
`const maxContentForPortraitMode = 680;`

maybe it's too small, I don't really know what size to choose...

lndrtrbn (approved):

Issue #4342, opened 2023-09-15
Add option to choose the maximum number of elements to push on CSV feeds


richard-julien:
## Use case

Add option to choose the maximum number of elements to push on CSV feeds.
That could be useful on specific use cases. Be aware that extending the number of default elements (5000) will have impact performance.

PR #4639, opened 2023-10-17
[backend] Improve CSV feed rolling max results size to not rely on pagination limitation

### LINKED ISSUE #4636: Why is there a limit of 5000 entities in a csv feed?
## Description

Limit of 5.000 entities in csv feed. Why is there a limit on entities in a csv feed?

This is probably for performance but I would like to know if we can increase this?

## Environment

2. OpenCTI version: 5.11.2
3. OpenCTI client: front-end

## Reproducible Steps

1. Data-> Data Sharing->CSv-feed
2. Create a feed with more than 5.000 entities
3. it will only show till 5.000

## Solution
Introduce an option to increase the size

`data_sharing:max_csv_feed_result`

richard-julien:
Related to:
-  https://github.com/OpenCTI-Platform/opencti/issues/4636
- https://github.com/OpenCTI-Platform/opencti/issues/4342


opencti-platform/opencti-graphql/src/database/engine.js
```diff
@@ -1913,10 +1913,11 @@ export const elPaginate = async (context, user, indexName, options = {}) => {
     );
 };
 export const elList = async (context, user, indexName, options = {}) => {
-  const { first = MAX_SEARCH_SIZE, infinite = false } = options;
+  const { first = MAX_SEARCH_SIZE, maxSize = undefined, infinite = false } = options;
```
SouadHadjiat: is `infinite` option used somewhere ? I can't find it anywhere in the code except here in this method.

There is also a very similar method `elList` in `middleware-loader.ts` , with exactly the same logic, can we deduplicate ?

richard-julien: Nice catch, will take a look

richard-julien: So i refactor to keep only 1 implementation. To do that and prevent circle dependencies i was forced to create engine-loader (like middleware-loader) to start extract some logic and convert that to TS. Due to time limitation i dont really spend some time to correctly convert to TS.

labo-flg: I can have a look at that !

labo-flg: Wow. 1k+ lines of code. It might take a while 😅

richard-julien: Its more code move than implementation :)

opencti-platform/opencti-graphql/src/database/engine-loader.ts
```diff
@@ -0,0 +1,1108 @@
...
+  `${ATTRIBUTE_ABSTRACT}^5`,
+  `${ATTRIBUTE_EXPLANATION}^5`,
+  // Add all other attributes
+  '*',
+];
+
+const elasticSearchClient = new ElkClient(searchConfiguration);
+const openSearchClient = new OpenClient(searchConfiguration);
+let isRuntimeSortingEnable = false;
+let engine: any = openSearchClient;
```
labo-flg, line 107: The correct type would be :

`let engine: OpenClient | ElkClient = openSearchClient;`

This will causes errors on each call to the `engine` for one reason: the 2 clients API are _similar_ but not _compatible_ from TS standpoint.
The functions `info` or `search` exist, but hey do not have the same signature.

Solving this issue is not trivial, because we need to reconsider completely this _one-object-fits-all_ strategy in the code. This is an easy and quick solution in JS, but when you think of it it's a dangerous strategy generally speaking. In our case here, risk is under control as we handle properly the results depending on which engine is used in reality.

I would advise not to change anything for now, it's too big.
We would have to refactor entirely so we have typeguards (or `if (engine instanceof ... etc`) and proper types according to elk/opensearch generic typings, like:

`client.search<OurSpecificDocument>(...)`

opencti-platform/opencti-graphql/src/database/engine-loader.ts
```diff
@@ -0,0 +1,1108 @@
...
+  ];
+  if (includeAuthorities) {
+    const roleIds = user.roles.map((r) => r.id);
+    const owners = [...userAccessIds, ...capabilities, ...roleIds];
+    authorizedFilters.push({ terms: { 'authorized_authorities.keyword': owners } });
+  }
+  return [{ bool: { should: authorizedFilters } }];
+};
+
+export const elRawSearch = (context: AuthContext, user: AuthUser, types: string[], query: any) => {
```
labo-flg, line 383: The functions that query the database share all the same problem regarding typescript conversion: we need to specify dynamically the input and output. Here, we use any and hope we find the right result according to our input.

The functions on the elastic client side are all generic, and what we do is some sort of wrapping around them to expose our own API ( `elPaginate`, `elQueryBuilder` etc.). Thus our functions should also be generic, and the code that calls them must define the input and output types locally, depending on the context.

Here again, this is a huge change, this time across all the backend. Probably not worth it right now.

Fortunately we have the graphql layer that is strictly typed, so the backend won't send something it's not supposed to send to the frontend. We could detect some problems that way (problems that would not be possible with proper TS typing).

opencti-platform/opencti-graphql/src/database/engine.js
labo-flg: This is definitely a good start, I don't see much _eslint-disable_ and the like... the next stage would be to actually convert the _any_ everywhere.

I started to dig deeper in that direction, to check the impact (see my comments in the code).
IMO as long as converting the any requires changes in the code logic, it's too risky to do it _on the side_, not worth it.
The engine code is too critical to do this lightly.

What we could do iteratively though, is to handle the types locally, where we use `elPaginate` and the like.
We would cast the `any` returned by our engine, and locally the code would be proper TS with a real added value.
This seems feasible...

What do you think ?

richard-julien: Yes i think its the right direction. The target of this PR is to deduplicate code and not resolving the TS migration of engine.js.

labo-flg (approved):

richard-julien:
Closed in favor of https://github.com/OpenCTI-Platform/opencti/pull/5061
Due to some big recent change, no more big bang of moving code

PR #4645, opened 2023-10-17
[backend] Improve dictionary upsert management (#4644)

### LINKED ISSUE #4644: Improve dictionary upsert management
## Use case

Improve the way we handle dictionary upsert like hashes in StixFile
- Prevent error when receiving upsert with empty data (should do nothing in standard but not produced any error)
- Handle full sync mode to have perfect match between 2 platforms.
- Allow full cleanup when empty value is received in full sync mode.

richard-julien:
See https://github.com/OpenCTI-Platform/opencti/issues/4644


helene-nguyen:
Why these changes ?

richard-julien:
> Why these changes ?

More information directly in the ticket https://github.com/OpenCTI-Platform/opencti/issues/4644.
Not a lot of description but should be sufficient.

Issue #4694, opened 2023-10-20
Cannot save changes to report content. Update button is inactive


rattat0r:
## Description

Cannot save changes to report content. Update button is inactive.

## Environment

1. OS: Ubuntu 20.04.6
2. OpenCTI version: OpenCTI 5.10.1
3. OpenCTI client: frontend

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Go to Report->Knowledge->Content mapping view
4. Switch to Edition mode
5. Make some changes
6. The Update button is not clickable

## Expected Output

Update button is clickable.

## Actual Output

Update button is not clickable.

## Screenshots
![image](https://github.com/OpenCTI-Platform/opencti/assets/122795003/f31ec9bb-0326-4563-8a60-e3a6d2559f38)

labo-flg:
Bug confirmed in latest version (5.11.6) ✅

This happens when the option _Enforce References_ is enabled for the report entity type by an admin.

**Workaround**
* If you only need to update the _description_, you can go through the burger menu (vertical dots next to the report name) and change it from the edition panel. The "Update" buttons works fine there.
* Unfortunately I did not find any workaround for the _Content_ field, as it is only present in the main view.
* disabling the option _Enforce References_ for the reports would also solve the issue but I understand it might be not suitable for you

**RCA**
Once the option _Enforce References_ is enabled, it enables the `<CommitMessage />` inside `<ContainerContent />`
--> see [here](https://github.com/OpenCTI-Platform/opencti/blob/88ebeb32a6a9c50a1d5bdf70284ba4e31ad68919/opencti-platform/opencti-front/src/private/components/common/containers/ContainerContent.jsx#L389)

The button stays disabled here because of the form validation. 
It expects 2 fields that are not present in the form : `name` and `published`.

The form validator comes from the utility function `useYupSchemaBuilder` 
--> see [here](https://github.com/OpenCTI-Platform/opencti/blob/88ebeb32a6a9c50a1d5bdf70284ba4e31ad68919/opencti-platform/opencti-front/src/utils/hooks/useEntitySettings.tsx#L70)

**Additional information**
Same bug when editing a **Grouping** or **RFI** with _Enforce References_. 
* RFI --> the form expects `created` and `name` fields,  not  present in this form.
* Grouping --> the form expects `context` and `name` fields,  not  present in this form.

rattat0r:
@labo-flg Thank you very much!

PR #4720, opened 2023-10-23
[frontend] fix enforced references for Reports/Rfi/Groupings

labo-flg:
### Proposed changes

This PR addresses #4694 where the _enforced references_ mechanism does not work when editing a RFI, Reports or Grouping from the Content mapping view.

Changes:
* When editing, change the form validation to accept partial changes (not all required field from schema)
* Repair the CommitMessage
  * close dialog after submitting ref+commit message
  * fix query variable mapping

### Related issues

#4694

### Checklist


opencti-platform/opencti-front/src/private/components/common/containers/ContainerContent.jsx
```diff
@@ -128,21 +128,49 @@ const ContainerContentComponent = ({ containerData }) => {
...
+    const { setSubmitting } = options;
     const { message, references, ...otherValues } = values;
     const commitMessage = message ?? '';
     const commitReferences = (references ?? []).map(({ value }) => value);
     editor.fieldPatch({
       variables: {
         id: containerData.id,
-        input: otherValues,
+        // rebuild an array of {key: x, value: y } objects for the query
+        input: Object.keys(otherValues).map(((k) => ({ key: k, value: otherValues[k] }))),
```
labo-flg, line 158: The query could never work before; we passed an object like `{ description: "blabla" ,...}` and not this key-value format

opencti-platform/opencti-front/src/private/components/common/containers/ContainerContent.jsx
```diff
@@ -256,22 +273,23 @@ const ContainerContentComponent = ({ containerData }) => {
     description: description || '',
     content: content || '',
   };
+
   return (
     <div className={classes.container}>
       <Grid
-        container={true}
+        container
```
labo-flg, line 280: quick refactor: the `={true}` is implicit in such case

opencti-platform/opencti-front/src/private/components/common/containers/ContainerContent.jsx
```diff
@@ -256,22 +273,23 @@ const ContainerContentComponent = ({ containerData }) => {
...
           <Typography
             variant="h4"
-            gutterBottom={true}
+            gutterBottom
             style={{ float: 'left' }}
           >
             {t('Content')}
           </Typography>
-          <Tooltip title={t('Edition mode')}>
+          <Tooltip title={editionMode ? t('Turn off edition mode') : t('Turn on edition mode')}>
```
labo-flg, line 292: suggestion of UI improvement ;

opencti-platform/opencti-front/src/private/components/common/form/CommitMessage.tsx
```diff
@@ -50,12 +50,25 @@ const CommitMessage: FunctionComponent<CommitMessageProps> = ({
...
+
   const validateReferences = (
     references: ExternalReferencesValues | undefined,
   ) => !!references && references.length > 0;
+
+  const onSubmitFromDialog = async () => {
+    await submitForm();
+    handleClose?.();
+    handleCloseControl(); // make sure the dialog is closed now
+  };
```
labo-flg, line 66: dialog was never closed at the end of query

opencti-platform/opencti-front/src/utils/hooks/useEntitySettings.tsx
```diff
@@ -73,17 +73,24 @@ export const useYupSchemaBuilder = (
   isCreation: boolean,
   exclusions?: string[],
 ): ObjectSchema<{ [p: string]: unknown }> => {
+  // simplest case: we're in update mode, so we do not need all mandatory fields
+  if (!isCreation) {
+    return Yup.object().shape(existingShape);
+  }
```
labo-flg, line 79: This is what caused the initial problem described in the bug.

I'd really like a review on this, this is a rough change.
I assumed that this function had 2 modes, edition and creation, and we actually want to return a yup validation computed from schema ONLY on creation, because in update mode nothing is actually required, the query might patch any field.

opencti-platform/opencti-front/src/private/components/common/containers/ContainerContent.jsx
Archidoit: you should also add the translations in Localization.js :)

labo-flg: done.

opencti-platform/opencti-front/src/private/components/common/containers/ContainerContent.jsx
helene-nguyen: These changes will patch both field when only one field is changed, do we want that ?

Screenshot 2023-10-23 211646

Issue #4505, opened 2023-10-06
CSV column mapping when a column can have different representations


Jipegien:
## Use case

A CSV contains a list of different objects. One of the column represent the type of object. According to the type, mapping of other column will differ.

## Current Workaround

Split the csv file

## Proposed Solution

In the CSV mapper, be able to define the column representing the objects' type and the value to use for the current representation mapping.
Example : Representation for "Threat Actor" when column Index "D" with value "equals/not equals" "threat_actor"

securitiz:
@Jipegien was this feature implemented? I'm having a difficult time figuring out how to define the proposed columns.

My understanding is there should be one column, we'll call it "Value" - and it can contain all sorts of object types - files, IPv4s, URLs, etc. There is another column where we specify what type of object each value should be ingested as. Is my understanding correct?

Jipegien:
This is clearly not implemented. Title of the issue has not been changed I think. Sorry about that. I confirm you cannot map such type of csv at the moment. @romain-filigran: to include in the long list of improvements for the CSV mapper.

nino-filigran:
closing this one since it's a duplicate of https://github.com/OpenCTI-Platform/opencti/issues/4935

PR #4742, opened 2023-10-25
[Frontend/Backend] Minor improvement for CSVMapper

jpkha:
### Proposed changes

* Manage CSVMapperTest Error on API Call 
* Display default message if no csv mappers in connector list, leading the users to Settings or contact administrator
*  Modifying test result panel to include some global information about the result
* Add some missing i18n
### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/4505 

### Checklist

### Further comments


opencti-platform/opencti-front/src/private/components/data/csvMapper/CsvMapperTestDialog.tsx
```diff
@@ -25,6 +25,19 @@ interface CsvMapperTestDialogProps {
   configuration: string;
 }
 
+interface CsvMapperResult {
+  value: string;
+  nbRelationships: number;
+  nbEntities: number;
+}
+
+interface ResultCSVMapper {
```
Kedae: Not sure to undersand the use of that, shouldn't it be the type comming from relay ?

opencti-platform/opencti-front/src/private/components/data/csvMapper/CsvMapperTestDialog.tsx
```diff
@@ -58,14 +71,21 @@ const CsvMapperTestDialog: FunctionComponent<CsvMapperTestDialogProps> = ({
       .then((data) => {
         const resultTest = (data as CsvMapperTestDialogQuery$data)
           .csvMapperTest;
-        setResult(JSON.stringify(resultTest, null, '  '));
+        setResult({
+          value: JSON.stringify(resultTest, null, '  '),
+          nbEntities: resultTest.filter((obj: ResultCSVMapper) => !obj.relationship_type).length,
+          nbRelationships: resultTest.filter((obj: ResultCSVMapper) => !!obj.relationship_type).length,
```
Kedae: Why !! ?

opencti-platform/opencti-front/src/private/components/data/csvMapper/CsvMapperTestDialog.tsx
```diff
@@ -113,13 +133,26 @@ const CsvMapperTestDialog: FunctionComponent<CsvMapperTestDialogProps> = ({
           </Button>
           {loading && (
             <div style={{ marginLeft: 10 }}>
-              <Loader variant={LoaderVariant.inElement} />
+              <Loader variant={LoaderVariant.inElement}/>
             </div>
           )}
         </div>
+        {result
+          && <div style={{
```
Kedae: Use classes instead of inline style. And check if you can use a MUI component instead of redefining something like that ?

opencti-platform/opencti-front/src/private/components/data/csvMapper/CsvMapperTestDialog.tsx
```diff
@@ -113,13 +133,26 @@ const CsvMapperTestDialog: FunctionComponent<CsvMapperTestDialogProps> = ({
...
+            paddingTop: 8,
+            fontSize: '1rem',
+            gap: 8,
+            justifyContent: 'center',
+            display: 'flex',
+          }}>
+            <span>{t('Objects found')} : </span>
+            <span><span style={{ fontWeight: 'bold' }}>{result.nbEntities} </span> {t('Entities')}</span>
```
Kedae: Same

opencti-platform/opencti-front/src/private/components/import/ImportContent.jsx
```diff
@@ -171,17 +172,18 @@ const importValidation = (t, configurations) => {
   return Yup.object().shape(shape);
 };
 
+const initState = {
```
Kedae: I don't think we want to initialize that outside of the react render

opencti-platform/opencti-front/src/private/components/import/ImportContent.jsx
```diff
@@ -199,7 +201,7 @@ class ImportContentComponent extends Component {
   }
 
   handleCloseImport() {
-    this.setState({ fileToImport: null });
+    this.setState({ ...initState });
```
Kedae: No need for spreding

opencti-platform/opencti-front/src/private/components/import/ManageImportConnectorMessage.tsx
```diff
@@ -0,0 +1,32 @@
...
+interface ManageImportConnectorMessageProps {
+  name: string;
+}
+
+const ManageImportConnectorMessage: FunctionComponent<ManageImportConnectorMessageProps> = ({ name }) => {
+  const { t } = useFormatter();
+  const isCsvMapperUpdater = useGranted([TAXIIAPI_SETCSVMAPPERS]);
+  switch (name) {
+    case 'ImportCsv':
+      return <div style={{ paddingTop: 6 }}>
```
Kedae: Same for inline styling

opencti-platform/opencti-front/src/private/components/data/csvMapper/CsvMapperTestDialog.tsx
jpkha: The type from relay is any. I could not get any good typing so I need a customize one.

opencti-platform/opencti-front/src/private/components/data/csvMapper/CsvMapperTestDialog.tsx
jpkha: We need to display : 
- number of entity found
- number of relationship found
So I thought it's was better to provide the information here. 

How should I handle it ?

opencti-platform/opencti-front/src/private/components/data/csvMapper/CsvMapperTestDialog.tsx
jpkha: Ok sure. All the component have inline-style so I try to keep the styling consistent. So I will refactor and use MUI component and class.

opencti-platform/opencti-front/src/private/components/import/ImportContent.jsx
jpkha: I think if we put inside the react render, we will initialize each time the component is render. What do you think ?

opencti-platform/opencti-front/src/private/components/data/csvMapper/CsvMapperTestDialog.tsx
Kedae: In that case I think we need to modify relay configuration to handle JSON type the way we want in front.

Kedae: @richard-julien Any thougth ?

opencti-platform/opencti-front/src/private/components/import/ImportContent.jsx
jpkha: Ok will remove it. Thank you.

opencti-platform/opencti-front/src/private/components/data/csvMapper/CsvMapperTestDialog.tsx
Kedae: Just asking about the double negation on the relationship_type.

opencti-platform/opencti-front/src/private/components/import/ImportContent.jsx
Kedae: Yes, but I don't think we want a object to be living with some value even when the component as been destroyed. Not really confident about the best practices with react. @guillaumejparis a thougth ?

guillaumejparis: No difference for me, the state will be initialized with this object but after the state will be handle internally no action will be done on this object after and it will be destroyed when the js garbage collector passes.



> I think if we put inside the react render, we will initialize each time the component is render. What do you think ?

No danger here, we are in old component so no re render of the constructor

opencti-platform/opencti-front/src/private/components/data/csvMapper/CsvMapperTestDialog.tsx
jpkha: For me having a double negation would show my intention to return a boolean value. But I can remove it if you prefer it to be written without double negation.

Kedae: Ok, was just checking that I was not missing something about the obj type :) Good for me.

opencti-platform/opencti-front/src/private/components/data/csvMapper/CsvMapperTestDialog.tsx
jpkha: I use Box from MUI in order to style the div container, is it okay or you prefer to a Styled-component or MakeStyle instead ?

opencti-platform/opencti-front/src/private/components/import/ImportContent.jsx
jpkha: Finally, I get back to the initial configuration.

opencti-platform/opencti-front/src/private/components/import/ImportContent.jsx
jpkha: I changed it as I explained above. But I try without the spreading, and it did not rerender and we are blocked we the modal open. I think using the object with spread, create the same object properties with a different reference and make the component rerender.

labo-flg: yes, spreading will just shallow copy the properties in a new object, so the ref does change and a rerender is triggered

opencti-platform/opencti-front/src/private/components/import/ManageImportConnectorMessage.tsx
```diff
@@ -0,0 +1,33 @@
...
+  name: string;
+}
+
+const ManageImportConnectorMessage: FunctionComponent<ManageImportConnectorMessageProps> = ({ name }) => {
+  const { t } = useFormatter();
+  const isCsvMapperUpdater = useGranted([TAXIIAPI_SETCSVMAPPERS]);
+  switch (name) {
+    case 'ImportCsv':
+      return <Box sx={{ paddingTop: '8px' }}>
+        {t('There are not any configurations set yet')}
```
labo-flg: probably put this message in red like an error message?

opencti-platform/opencti-front/src/private/components/data/csvMapper/CsvMapperTestDialog.tsx
jpkha: I modified the graphql in order to get the information from the back and so the typing that's coming with.

@richard-julien 
For information, I could not transform the object to string from the back because it removes the indentation/structure of the JSON so it became unreadable.
We can change it when we change the layout of the result.

opencti-platform/opencti-front/src/utils/Localization.js
```diff
@@ -8909,6 +8980,14 @@ const i18n = {
...
       update: 'Modification events',
       delete: 'Deletion events',
+      Processing: 'Processing',
+      Automation: 'Automation',
+      'My CSV file contains headers': 'My CSV file contains headers',
+      'CSV separator': 'CSV separator',
+      'Objects found': 'Objects found',
+      'Please contact an administrator': 'Please contact an administrator',
+      'Create a CSV Mapper configuration': 'Create a CSV Mapper configuration',
+      'There are not any configurations set yet': 'There are not any configurations set yet',
```
SouadHadjiat: there is no need to translate these messages in english here, we usually use the key as the english message, unless the key is different from the message.

opencti-platform/opencti-front/src/private/components/data/csvMapper/CsvMapperTestDialog.tsx
jpkha: Thanks to Julien, I find the correct way to handle the Stringify in the back.

opencti-platform/opencti-front/src/private/components/import/ManageImportConnectorMessage.tsx
jpkha: done

opencti-platform/opencti-front/src/utils/Localization.js
jpkha: I removed these lines

opencti-platform/opencti-front/src/private/components/import/ManageImportConnectorMessage.tsx
```diff
@@ -0,0 +1,37 @@
...
+import Box from '@mui/material/Box';
+import { styled } from '@mui/material/styles';
+import { useFormatter } from '../../../components/i18n';
+import useGranted, { TAXIIAPI_SETCSVMAPPERS } from '../../../utils/hooks/useGranted';
+
+interface ManageImportConnectorMessageProps {
+  name: string;
+}
+const WarningText = styled('span')({
+  color: '#f44336',
```
Kedae: You should use the feme

opencti-platform/opencti-front/src/private/components/import/ManageImportConnectorMessage.tsx
```diff
@@ -0,0 +1,37 @@
+import React, { FunctionComponent } from 'react';
+import { Link } from 'react-router-dom';
+import Box from '@mui/material/Box';
+import { styled } from '@mui/material/styles';
+import { useFormatter } from '../../../components/i18n';
+import useGranted, { TAXIIAPI_SETCSVMAPPERS } from '../../../utils/hooks/useGranted';
+
+interface ManageImportConnectorMessageProps {
+  name: string;
```
Kedae: no ';' please

jpkha: done

opencti-platform/opencti-front/src/private/components/import/ManageImportConnectorMessage.tsx
jpkha: done

Archidoit:
Missing translation :
image

Archidoit:
Missing translation: 
image

labo-flg:
I tested with the following data and mapper, and it fails with `"Error: Cannot return null for non-nullable field CsvMapperTestResult.objects."`

Note that it works if you keep only the first entity representation and remove the second one.

[country-incident.csv](https://github.com/OpenCTI-Platform/opencti/files/13237402/country-incident.csv)

image

image

labo-flg:
false alarm, it works fine.

opencti-platform/opencti-front/src/private/components/data/csvMapper/CsvMapperTestDialog.tsx
```diff
@@ -112,17 +126,31 @@ const CsvMapperTestDialog: FunctionComponent<CsvMapperTestDialogProps> = ({
...
+            sx={{
+              paddingTop: '8px',
+              fontSize: '1rem',
+              gap: '8px',
+              justifyContent: 'center',
+              display: 'flex',
+            }}>
+            <span>{t('Objects found')} : </span>
+            <span><strong>{result?.csvMapperTest?.nbEntities} </strong> {t('Entities')}</span>
+            <span><strong>{result?.csvMapperTest?.nbRelationships}</strong> {t('Relationships')}</span>
```
labo-flg, line 147: when there is an error these values are "". 
Maybe print 0 instead

opencti-platform/opencti-front/src/private/components/import/ManageImportConnectorMessage.tsx
```diff
@@ -0,0 +1,37 @@
...
+        <div>
+          {
+            isCsvMapperUpdater
+              ? <Link to="/dashboard/data/processing/csv_mapper">{t('Create a CSV Mapper configuration')}</Link>
+              : <WarningText>{t('Please contact an administrator')}</WarningText>
+          }
+        </div>
+      </Box>;
+    case undefined: // In case there isn't any connector selected
+      return <></>;
```
labo-flg: is there a reason to return an empty fragment instead of null ?

jpkha: No reason, is it better to use null or false instead ?

opencti-platform/opencti-front/src/private/components/data/csvMapper/CsvMapperTestDialog.tsx
labo-flg: well it should never happen in fact, so it's ok.

opencti-platform/opencti-front/src/private/components/import/ManageImportConnectorMessage.tsx
labo-flg: It's common to use null to return an element that has nothing to display. At least in my experience.

labo-flg: https://legacy.reactjs.org/docs/conditional-rendering.html#preventing-component-from-rendering

labo-flg:
So I finally fount out why I got this validation error  `Cannot return null for non-nullable field CsvMapperTestResult.objects.`

I happens only if the type of entity you are trying to extract from CSV has a specific customization: **enforced references**.

The CSV bundler calls `validateInputCreation` which in that case expects a mandatory field `externalReferences` that is missing.

I suggest we open another bug as it's far from the initial scope of this PR (and the bug is present from beginning in our CSV mappers)

opencti-platform/opencti-front/src/private/components/import/ManageImportConnectorMessage.tsx
jpkha: Thank you. Nice to know.

labo-flg:
> I suggest we open another bug as it's far from the initial scope of this PR (and the bug is present from beginning in our CSV mappers)

opened https://github.com/OpenCTI-Platform/opencti/issues/4839 to tackle this problem separately.

labo-flg (approved):

Issue #4774, opened 2023-10-26
Incorrect redirection in data / relationships for hyper relations (relation to relation)


SamuelHassine:
## Description

1. Create a targets between an intrusion set and a sector
2. Create a located-at between the relation in 1. and a country
3. Go to data / relationships and click on the eye

PR #4791, opened 2023-10-27
[frontend] Fix Incorrect redirection in data / relationships for hyper relations (#4774)

marieflorescontact:
### Proposed changes

* Add a check on `node.from` to know if it is a relationship.

### Related issues

* [issue/4774](https://github.com/OpenCTI-Platform/opencti/issues/4774)

### Checklist

### Further comments

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/data/relationships/RelationshipsStixCoreRelationshipLine.jsx
+++ b/opencti-platform/opencti-front/src/private/components/data/relationships/RelationshipsStixCoreRelationshipLine.jsx
@@ -19,7 +19,7 @@ import { useFormatter } from '../../../../components/i18n';
 import ItemIcon from '../../../../components/ItemIcon';
 import ItemMarkings from '../../../../components/ItemMarkings';
 import { defaultValue } from '../../../../utils/Graph';
-import { resolveLink } from '../../../../utils/Entity';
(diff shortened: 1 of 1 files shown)


SamuelHassine:
Hello,

We need to resolve the correct relationship from / to of the hyper relation and redirect properly.

SamuelHassine:
Ok understood the source code, but the hyper relation can also be on the to.

marieflorescontact:
Is it possible that `from` and `to` are both relationships? @SamuelHassine

SamuelHassine:
It is, but you can ignore this case as it is not used.

marieflorescontact:
@SamuelHassine , 
`const remoteNode = (node.from && !node.from.relationship_type) ? node.from : node.to;`

We either take `to` if `from` is a relationship, or we take `from` if it is not a relationship, which will work if `to` is a relationship, as `from` and `to` cannot, for now, be both relationships.
 Is it correct ? Otherwise what do you suggest?

Archidoit:
When I click on the line of a relationship between a relationship and an entity (an opinion in my case), I'm redirected to this : 
image

SarahBocognano (commented):
Tested ✅  The redirect works now. As discussed together, just need to have a information/confirmation about the redirection page

Issue #4789, opened 2023-10-27
Priority and Severity filters are not handled in Live streams


labo-flg:
## Description
In streams, we are able to select a filter on the severity and priority.
Once live, the stream won't take into account these filters.

## Environment

1. OS (where OpenCTI server runs):  N/A
2. OpenCTI version: 5.11.11
3. OpenCTI client: frontend

## Reproducible Steps

1. create a stream with entity filter "Incident" and severity filter "low"
2. open the stream feed on the side
3. create an incident with severity "critical"
4. create an incident with severity "low"

## Expected Output

Only the event on the creation of incident with low severity is output in stream feed.

## Actual Output

Both events are output.

## Additional information

Quick RCA leads in backend in `filtering.js:isStixMatchFilter()`.
This function handles all filter attributes one by one, and severity/priority are simply not handled.

Issue #4536, opened 2023-10-10
Modifying Schema for filter of feed/taxii/stream


Jipegien:
## Use case

Apply the refactor of filters done in 5.11 (https://www.notion.so/filigran/Switch-between-AND-and-OR-operators-between-filters-change-filter-schema-d05fe9af797c4751a293e3a7046ef89d) when processing stream events against the stored filters

PR #4794, opened 2023-10-27
[backend] handle priority and severity attributes in filters

labo-flg:
* used by streams, triggers and playbooks

### Proposed changes

All changes inside `isStixMatchFilter` to handle the `priority` and `severity` attributes filters. 
There is 2 things to consider
* comparison should be case insensitive, as the filter can be saved as priority `p2` where stix value is `P2`
* user is able to select "no priority" or "no severity" and the filter is that case is `{ id: '', value '' }`

### Related issues

* #4789 

### Checklist


labo-flg:
I did not try to refactor the code while working on this bug, because I'm already in the middle of a complete revamp of this code while working on #4536.

I've tested multiple cases with incidents, RFIs, etc. on Live Streams.

PR #4810, opened 2023-10-30
[frontend] Status filter display (#4800)

### LINKED ISSUE #4800: Filter on status is not grouped correctly
## Use case

Multiple groups for case incident response, without real reason for this neither alphabetical order etc.

image

Archidoit:
issue: https://github.com/OpenCTI-Platform/opencti/issues/4800


SarahBocognano (approved):
Tested ✅

PR #4818, opened 2023-10-31
[frontend] Fix correlated cases is not working

### LINKED ISSUE #4803: Correlated cases is not working
## Description

As an example (working):

1. Create Report 1 and add Observable A in the report
2. Create Report 2 and add Observable A in the report

In both report, "Correlated reports" display the other one.

Do the same thing with Case 1 and Case 2.

=> Nothing in Correlated cases.

Megafredo:
### Proposed changes

* Updating the ternary operator in the "CaseIncidentDetail" file
* Adding the ternary operator in the "ReportDetail" file to replace empty space with '-'

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/4803

### Checklist

### Further comments

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/analyses/reports/ReportDetails.jsx
+++ b/opencti-platform/opencti-front/src/private/components/analyses/reports/ReportDetails.jsx
@@ -170,12 +170,12 @@ const ReportDetailsComponent = (props) => {
         <List>
           {(report.relatedContainers?.edges ?? []).length >= 2
             ? R.take(expanded ? 200 : 5, report.relatedContainers.edges)
-            .filter(
-              (relatedContainerEdge) => relatedContainerEdge.node.id !== report.id,
-            )
-            .map((relatedContainerEdge) => {
-              const relatedContainer = relatedContainerEdge.node;
-              return (
+              .filter(
+                (relatedContainerEdge) => relatedContainerEdge.node.id !== report.id,
+              )
+              .map((relatedContainerEdge) => {
+                const relatedContainer = relatedContainerEdge.node;
+                return (
                 <ListItem
                   key={report.id}
                   dense={true}
@@ -211,8 +211,8 @@ const ReportDetailsComponent = (props) => {
                     />
                   </div>
                 </ListItem>
-              );
-            }) : '-'}
+                );
+              }) : '-'}
         </List>
         {expandable && (
           <Button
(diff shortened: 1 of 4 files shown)


opencti-platform/opencti-front/src/private/components/cases/case_incidents/CaseIncidentDetails.tsx
```diff
@@ -233,7 +233,7 @@ const CaseIncidentDetails: FunctionComponent<CaseIncidentDetailsProps> = ({
           {t('Correlated cases')}
         </Typography>
         <List classes={{ root: classes.relatedContainers }}>
-          {expandable
+          {(data.relatedContainers?.edges ?? []).length >= 2
```
jpkha: I think you can write something like this instead :
`data.relatedContainers?.edges?.length >= 2`

opencti-platform/opencti-front/src/private/components/analyses/reports/ReportDetails.jsx
```diff
@@ -168,7 +168,8 @@ const ReportDetailsComponent = (props) => {
           {t('Correlated reports')}
         </Typography>
         <List>
-          {R.take(expanded ? 200 : 5, report.relatedContainers.edges)
+          {(report.relatedContainers?.edges ?? []).length >= 2
```
jpkha: I think you can write something like this instead :
data.relatedContainers?.edges?.length >= 2

jpkha (approved):

opencti-platform/opencti-front/src/private/components/cases/case_incidents/CaseIncidentDetails.tsx
```diff
@@ -233,7 +233,7 @@ const CaseIncidentDetails: FunctionComponent<CaseIncidentDetailsProps> = ({
           {t('Correlated cases')}
         </Typography>
         <List classes={{ root: classes.relatedContainers }}>
-          {expandable
+          {(data.relatedContainers?.edges && data.relatedContainers.edges.length >= 2)
```
Kedae: Not sure, let's talk about that

Kedae (changes requested):

opencti-platform/opencti-front/src/private/components/cases/case_incidents/CaseIncidentDetails.tsx
Kedae: You need to extract the filtering of the list before the return and check that this computed list as at least one element

PR #4830, opened 2023-11-02
On manual logout, invalidate referrer

### LINKED ISSUE #4823: URL not reset after disconnection
## Description

When logging out, the url is not reset, i.e. if I am at the time of disconnection on the page https://[OpenCTI]/dashboard/settings/accesses/organizations/57bcf037-8d44-4eba-b52e -bced46605d0c, this url will still be present after I log out while I am on the platform connection page.

If I log back in, I'll find myself back on the page https://[OpenCTI]/dashboard/settings/accesses/organizations/57bcf037-8d44-4eba-b52e -bced46605d0c (rather than on the dashboard as I suppose it should be).

This is strange behavior and can cause certain bugs as described in the second point of this issue : https://github.com/OpenCTI-Platform/opencti/issues/4822

## Environment

OpenCTI 5.11.12

## Reproducible Steps

1. Go to a platform page (other than the dashboard)
2. Log out
3. Look at the URL of your browser while you are on the login portal
4. When you reconnect, you will return to the page where you were before disconnection

## Screenshots

Screenshot 2023-10-31 145201
Screenshot 2023-10-31 145209

labo-flg:
When a user logs our using the top bar button, we redirect to the signin page. However, the referrer is kept and when a user logs in again, he is redirected to this URL.

This is not suitable as the user who logs in after the first might not be the same, and might not have access to this specific URL (for instance, an admin logs out from some deep configuration in the platform settings ; next user is not admin and does not have the right capabilities).

### Proposed changes

* frontend: on logout from the top bar, invalidate referrer in the link to `/logout`
* backend: on logout, if no referrer is given, redirect to '/'

### Related issues

* #4823

### Checklist

### Further comments

To clarify the UI, I suggest we use the secondary color for the link text
image

Logging out is not like a redirection to profile or something "light"... Let's highlight it in the menu.

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/nav/TopBar.tsx
+++ b/opencti-platform/opencti-front/src/private/components/nav/TopBar.tsx
@@ -11,6 +11,7 @@ import Menu from '@mui/material/Menu';
 import Divider from '@mui/material/Divider';
 import MenuItem from '@mui/material/MenuItem';
 import Tooltip from '@mui/material/Tooltip';
+import MuiLink from '@mui/material/Link';
 import { graphql, PreloadedQuery, usePreloadedQuery, useSubscription } from 'react-relay';
 import { useTheme } from '@mui/styles';
 import makeStyles from '@mui/styles/makeStyles';
@@ -496,8 +497,17 @@ const TopBarComponent: FunctionComponent<TopBarProps> = ({
                 {t('Profile')}
               </MenuItem>
               <MenuItem onClick={handleOpenDrawer}>{t('Feedback')}</MenuItem>
-              <MenuItem id="logout-button" onClick={() => handleLogout()}>
-                {t('Logout')}
+              <MenuItem id="logout-button">
+                <MuiLink
+                  underline="none"
+                  color="secondary"
+                  href="/logout"
+                  // user will be redirected to login, but we do not want to keep the referrer url
+                  // on next login, the user might not be the same and might not have access to this url
+                  rel="noreferrer"
+                >
+                  {t('Logout')}
+                </MuiLink>
               </MenuItem>
             </Menu>
           </div>
--- a/opencti-platform/opencti-front/src/private/components/nav/TopBar.tsx
+++ b/opencti-platform/opencti-front/src/private/components/nav/TopBar.tsx
@@ -133,10 +133,6 @@ const useStyles = makeStyles<Theme>((theme) => ({
   },
 }));
 
-export const handleLogout = () => {
-  window.location.pathname = '/logout';
-};
-
 const topBarNotificationNumberSubscription = graphql`
   subscription TopBarNotificationNumberSubscription {
     notificationsNumber {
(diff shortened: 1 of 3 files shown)


opencti-platform/opencti-graphql/src/http/httpPlatform.js
```diff
@@ -257,7 +257,7 @@ const createApp = async (app) => {
   // Logout
   app.get(`${basePath}/logout`, async (req, res, next) => {
     try {
-      const referer = extractRefererPathFromReq(req);
+      const referer = extractRefererPathFromReq(req) ?? '/';
```
labo-flg, line 260: otherwise you could be redirected to `/undefined` when referrer is unset

opencti-platform/opencti-front/src/private/components/nav/TopBar.tsx
```diff
@@ -496,8 +497,17 @@ const TopBarComponent: FunctionComponent<TopBarProps> = ({
                 {t('Profile')}
               </MenuItem>
               <MenuItem onClick={handleOpenDrawer}>{t('Feedback')}</MenuItem>
-              <MenuItem id="logout-button" onClick={() => handleLogout()}>
```
jpkha, line 499: You can maybe delete the "handleLogout" function at the same time ?

labo-flg: it's actually a global function exported and used elsewhere. 
it's a simple `window.location = "/logout"`

opencti-platform/opencti-front/src/private/components/nav/TopBar.tsx
```diff
@@ -496,8 +497,17 @@ const TopBarComponent: FunctionComponent<TopBarProps> = ({
...
-                {t('Logout')}
+              <MenuItem id="logout-button">
+                <MuiLink
+                  underline="none"
+                  color="secondary"
+                  href="/logout"
+                  // user will be redirected to login, but we do not want to keep the referrer url
```
lndrtrbn:
```suggestion
                  // user will be redirected to login screen after logout, but we do not want to keep the referrer url
```
No ?

opencti-platform/opencti-front/src/private/components/nav/TopBar.tsx
lndrtrbn: Actually used only in `TimeoutLock.tsx`, maybe move it here?

opencti-platform/opencti-front/src/private/components/nav/TopBar.tsx
labo-flg: hm I meant he's redirected to /logout ... which shows him the login screen.

opencti-platform/opencti-front/src/private/components/nav/TopBar.tsx
labo-flg: yeah let's do that

opencti-platform/opencti-front/src/private/components/nav/TopBar.tsx
lndrtrbn: Ahhhh! Suggestion updated

lndrtrbn (approved):

PR #4834, opened 2023-11-02
[frontend] Fix Empty TAXII Ingester User Reverts to Previous User instead of System (#4813)

### LINKED ISSUE #4813: Empty TAXII Ingester User Reverts to Previous User instead of System
## Description

The "User responsible for data creation (empty = System)" reverts to the previously set user when changing the field blank. There doesn't appear to be a way to set the User back to "System".

## Environment

1. OS (where OpenCTI server runs): Ubuntu 22.04
2. OpenCTI version: 5.11.12
3. OpenCTI client: frontend
4. Other environment details:

## Reproducible Steps

1. Create a TAXII Ingester. Set the user to an existing user. 
2. Click Update
3. Set the user field to blank and hit the "X" to close the UI element
4. Hit Update again to view the User field and notice that the value is no longer blank or set to System. It is the previous user.

## Expected Output

When the User field is blank, the value becomes System.

## Actual Output

The User field becomes the previous user value.

## Screenshots (optional)
Changing value to empty:
image

Value after change (reverted to previous user "admin"):
image

helene-nguyen:
### Proposed changes

* Handle empty value

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/4813

### Checklist

### Further comments

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/common/form/CreatorField.tsx
+++ b/opencti-platform/opencti-front/src/private/components/common/form/CreatorField.tsx
@@ -73,7 +73,11 @@ const CreatorField: FunctionComponent<CreatorFieldProps> = ({
           label: n?.node.name,
           value: n?.node.id,
         }));
-        const templateValues = [...creators, ...NewCreators];
+        const emptyValue = {
+          label: '',
+          value: '',
+        };
+        const templateValues = [emptyValue, ...creators, ...NewCreators];
         // Keep only the unique list of options
         const uniqTemplates = templateValues.filter((item, index) => {
           return (
--- a/opencti-platform/opencti-front/src/private/components/common/form/CreatorField.tsx
+++ b/opencti-platform/opencti-front/src/private/components/common/form/CreatorField.tsx
@@ -73,11 +73,7 @@ const CreatorField: FunctionComponent<CreatorFieldProps> = ({
           label: n?.node.name,
           value: n?.node.id,
         }));
-        const emptyValue = {
-          label: '',
-          value: '',
-        };
-        const templateValues = [emptyValue, ...creators, ...NewCreators];
+        const templateValues = [...creators, ...NewCreators];
         // Keep only the unique list of options
         const uniqTemplates = templateValues.filter((item, index) => {
           return (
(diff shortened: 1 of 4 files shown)


SouadHadjiat:
When I empty the input, I can now see an empty user in the list, and if I just empty the input and close the update panel, when I reopen it it still displays my old user.

The only way I can set back to System user, is to select in the list the empty user.

Capture d'écran 2023-11-03 094700

Also, When I create a new ingester, and try to set a user, I will always see this empty user in the list first

opencti-platform/opencti-front/src/private/components/common/form/CreatorField.tsx
```diff
@@ -73,7 +73,11 @@ const CreatorField: FunctionComponent<CreatorFieldProps> = ({
           label: n?.node.name,
           value: n?.node.id,
         }));
-        const templateValues = [...creators, ...NewCreators];
+        const emptyValue = {
+          label: '',
+          value: '',
+        };
+        const templateValues = [emptyValue, ...creators, ...NewCreators];
```
SarahBocognano: Exact same comments as @SouadHadjiat. I also have that empty field in the autocomplete drawer and if I put an empty field, when I reopen the update, my old user is still here.
Maybe add a "null" in your onInputChange={searchCreators || null} ?

helene-nguyen: @SarahBocognano @SouadHadjiat, after more investigations on that, the issue comes from validation schema for the creator only.

It's ready for another review :)

opencti-platform/opencti-front/src/private/components/data/ingestionTaxii/IngestionTaxiiEdition.jsx
```diff
@@ -39,7 +39,16 @@ const ingestionTaxiiValidation = (t) => Yup.object().shape({
...
+  user_id: Yup.lazy((value) => {
+    switch (typeof value) {
+      case 'object':
+        return Yup.object().nullable(); // schema for object
+      case 'string':
+        return Yup.string().nullable();
+      default:
+        return Yup.mixed();
+    }
+  }),
```
SouadHadjiat: I'm wondering if we could use Yup.mixed().nullable() :

`user_id: Yup.mixed().nullable(),`

I tested it and it works, I can see it used in other components, but I'm not sure of the best practice in this case. What do you think @Kedae ?

Also, we have the same issue for RSS and sync (SyncEdition, IngestionRSSEdition)

Kedae: I think the interest of lazy with a switch is to have a list of constrained possibilities.

In that case, since we have a default as mixed, a global mixed should be enough

helene-nguyen: Done :)

SouadHadjiat (approved):

Issue #3242, opened 2023-04-27
Export/Import Custom Dashboard (and Widget)


Arcuza:
## Use case

Export/Import of Custom Dashboard would allow the community to create and share Custom Dashboards. Might as well be a handy way to create backups of Custom Dashboards.

## Current Workaround

None that I'm aware of, without sharing sensitive data.

## Proposed Solution

Make a serialization of the Custom Dashboard and enable export to XML/JSON or whatever format suitable. Also enable import and allow duplication of dashboards.

## Additional Information

No

## If the feature request is approved, would you be willing to submit a PR?

No

PR #4837, opened 2023-11-02
[frontend/backend] Export and import dashboard config (#issue/3242)

Goumies:
### Proposed changes

* First lot (1/3) dedicated to custom dashboard
* Export
* Import

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/3242

### Checklist

- [x] I wrote test cases for the relevant use case
- [ ] I added/updated the relevant documentation (either on github or on notion)

### Further comments
Since those features will be delivered in the 5.12.0 release, the minimal compatible version is : 5.12.0.
`openCTI_version` is equal to the platform used for the export.
So, to test nominal cases, in this PR, you will have to manually change the `openCTI_version` to 5.12.0 and more.


Kedae:
Might not be you, but can you fix that ? 
image

Kedae:
Could be usefull to have the export button in the Dashboard list


@SarahBocognano : What do you mean by that ? Put an export button in the lines or in the popover ?

Kedae:
Could you format the file to have a correctly indented JSON instead of a full line ?

opencti-platform/opencti-front/src/components/ExportButtons.jsx
```diff
@@ -182,6 +204,13 @@ class ExportButtons extends Component {
               <FilePdfBox fontSize="small" color="primary" />
             </ToggleButton>
           </Tooltip>
+          {isCustomDashBoard && (
+            <Tooltip title={t('Export to JSON')}>
+              <ToggleButton onClick={this.handleExportJson.bind(this, workspace)}>
```
marieflorescontact: Not a big deal but `<ToggleButton/>` component is supposed to have a `value` prop (The value to associate with the button when selected in a ToggleButtonGroup)

opencti-platform/opencti-front/src/components/ItemIcon.jsx
```diff
@@ -314,6 +314,7 @@ const iconSelector = (type, variant, fontSize, color, isReversed) => {
         <SafetyCheckOutlined style={style} fontSize={fontSize} role="img" />
       );
     case 'dashboard':
+    case 'workspace':
```
marieflorescontact, line 320: is it normal to have `dashboard`: nothing ?
(`dashboard` is entity Type `workspace`)

opencti-platform/opencti-front/src/private/components/workspaces/WorkspaceCreation.jsx
```diff
@@ -46,12 +72,33 @@ const sharedUpdater = (store, userId, paginationOptions, newEdge) => {
...
+      },
+    });
+  };
+
+  const onSubmit = (values, { setSubmitting, resetForm }) => {
     commitMutation({
       mutation: workspaceMutation,
       variables: {
-        input: assoc('type', this.props.type, values),
+        input: assoc('type', type, values),
```
marieflorescontact: if you have time, could you remove Ramda? 💐

marieflorescontact (commented):
locally tested 👌

opencti-platform/opencti-front/src/components/ItemIcon.jsx
SarahBocognano: Yes for that case, it's due to the stream part. When you go to "Activity" in Entreprise Edition, you can access a stream and the icon was missing because the stream take the name of the entity type in the back, in our case workspace to define an icon. But in other part of the platform the icon will take 'dashboard' as a reference. 
So we put them both. In the case/switch mode, the dashboard will take the same property as the one below.

helene-nguyen:
Everything shrinks when you add or delete a graph

add-graph

helene-nguyen:
The dashboard I want to export is:

image


But when importing, everything also shrinks, is it the behavior expected ?

export-import-dash

opencti-platform/opencti-front/src/components/ExportButtons.jsx
```diff
@@ -168,7 +169,10 @@ class ExportButtons extends Component {
       containerId,
       investigationAddFromContainer,
       history,
+      location,
+      workspace,
```
SouadHadjiat, line 172: this component would need some day a refactoring, the properties we inject are so mixed (csv, investigation, history, workspace ...)

opencti-platform/opencti-graphql/src/modules/workspace/workspace-domain.ts
```diff
@@ -135,3 +142,45 @@ export const workspaceEditContext = async (context: AuthContext, user: AuthUser,
   return storeLoadById(context, user, workspaceId, ENTITY_TYPE_WORKSPACE)
     .then((workspaceToReturn) => notify(BUS_TOPICS[ENTITY_TYPE_WORKSPACE].EDIT_TOPIC, workspaceToReturn, user));
 };
+
+function checkDashboardConfigurationImport(parsedData: any) {
```
SouadHadjiat: we usually don't declare functions with `function`, more with 
`const myfunction = (args) => { }`

opencti-platform/opencti-graphql/src/modules/workspace/workspace-domain.ts
```diff
@@ -135,3 +142,45 @@ export const workspaceEditContext = async (context: AuthContext, user: AuthUser,
...
+  if (parsedData.version !== WORKSPACE_VERSION) {
+    throw FunctionalError(`Invalid version. Your workspace version must match the current one. Actual version : ${WORKSPACE_VERSION}`);
+  }
+}
+
+export const workspaceImport = async (context: AuthContext, user: AuthUser, file: Promise<FileHandle>) => {
+  const uploadedFile = await file;
+  const readStream = uploadedFile.createReadStream();
+  let fileContent = '';
+  await new Promise<void>((resolve, reject) => {
```
SouadHadjiat: I think you can use `streamToString` function in file-storage.js

opencti-platform/opencti-graphql/src/modules/workspace/workspace-domain.ts
```diff
@@ -135,3 +142,45 @@ export const workspaceEditContext = async (context: AuthContext, user: AuthUser,
...
+      resolve();
+    });
+    readStream.on('error', (error) => {
+      reject(error);
+    });
+  });
+  const parsedData = JSON.parse(fileContent);
+  checkDashboardConfigurationImport(parsedData);
+
+  const importWorkspaceCreation = await createEntity(context, user, parsedData, ENTITY_TYPE_WORKSPACE);
```
SouadHadjiat: which authorized members have this workspace ?

opencti-platform/opencti-graphql/tests/02-integration/02-resolvers/workspace-test.js
```diff
@@ -201,7 +204,63 @@ describe("Workspace resolver standard behavior", () => {
...
+        }
+      `,
+      variables: {
+        file: upload,
+      },
+    });
+
+    expect(queryResult.errors[0].message).toEqual(
+      "Invalid version. Your workspace version must match the current one. Actual version : 1.0.0",
+    );
```
SouadHadjiat, line 263: do we have a test with a valid json ?

Goumies: No, I forgot

opencti-platform/opencti-graphql/src/modules/workspace/workspace-domain.ts
Goumies: This property is not managed in the duplication process.
Since it only includes : name, type, description, version, and manifest.

opencti-platform/opencti-front/src/components/ExportButtons.jsx
SarahBocognano: If we finish the feature (the 3 slots) completely for the due date we will think about it, but for now, it's not prior

opencti-platform/opencti-graphql/src/modules/workspace/workspace-domain.ts
SouadHadjiat: Yes we can't copy the authorized members, but I think it should have the default authorized members we put when someone creates a dashboard, otherwise the dashboard is public, everyone can manage it. @Jipegien

opencti-platform/opencti-graphql/tests/02-integration/02-resolvers/workspace-test.js
SarahBocognano: The test with a valid json has been added

SarahBocognano:
Update: All comments are resolved, except for the question of auhtorized_members. Waiting for @Jipegien for infos

opencti-platform/opencti-front/src/components/ExportButtons.jsx
```diff
@@ -182,6 +186,13 @@ class ExportButtons extends Component {
               <FilePdfBox fontSize="small" color="primary" />
             </ToggleButton>
           </Tooltip>
+          {isCustomDashBoard && (
```
SouadHadjiat, line 188: why not test on workspace property instead of location path ? it can be present only if we are in dashboards view, right ?

opencti-platform/opencti-front/src/private/components/workspaces/workspaceExportHandler.ts
```diff
@@ -0,0 +1,24 @@
...
+interface workspaceToExport {
+  name: string
+  type?: string
+  manifest?: string
+}
+
+const handleExportJson = (workspace: workspaceToExport) => {
+  const dashboardName = workspace.name;
+  const dashboardConfig = JSON.stringify({
+    version: '1.0.0',
```
SouadHadjiat: where is `openCTI_version` ?

Goumies: Changes are coming

opencti-platform/opencti-front/src/components/ExportButtons.jsx
Goumies: Right

opencti-platform/opencti-front/src/private/components/workspaces/workspaceExportHandler.ts
```diff
@@ -0,0 +1,27 @@
+import fileDownload from 'js-file-download';
+import pjson from '../../../../package.json';
```
SouadHadjiat, line 2: So openCTI_version is the version we have in package.json of opencti-front ?

opencti-platform/opencti-graphql/src/modules/workspace/workspace-domain.ts
Jipegien: Yes. By default authorized_members. Need to be clearly documented in the official Doc.

Goumies: To make sure, I get it right : we want to add the user that importing the dashboard configuration as the only / first authorized member, @Jipegien.

opencti-platform/opencti-front/src/private/components/workspaces/workspaceExportHandler.ts
Goumies: Yes the one from the export platform

marieflorescontact (commented):
locally tested:

I found a weird behaviour:

- Create a dashboard with a user who has workspace management rights (admin). 
- Give access to  a specific group.
- create a user who is **not** part of this  specific group. 
- export the dashboard. 
- log with your new user
- new user managed to import the dashboard (he should not because he is not part of the group)
- we now have 2 identical workspaces with 2 different creators.

opencti-platform/opencti-graphql/src/modules/workspace/workspace-domain.ts
Jipegien: yes

Goumies:
Update : authorized members are managed. Thanks for your feedback

marieflorescontact (commented):
Locally tested 👌

marieflorescontact (approved):

opencti-platform/opencti-front/src/private/components/workspaces/WorkspaceCreation.jsx
```diff
@@ -46,12 +71,36 @@ const sharedUpdater = (store, userId, paginationOptions, newEdge) => {
...
+        );
+      },
+      onError: (error) => {
+        handleError(error);
+      },
+    });
+  };
+
+  const onSubmit = (values, { setSubmitting, resetForm }) => {
     commitMutation({
```
SouadHadjiat: sorry I just saw this when reviewing again, but since you refactored this component, we should now use `useMutation` as you did with `importMutation` instead of `commitMutation` for `workspaceMutation`

SouadHadjiat: Or maybe it was not possible to use it ? I can see now the updater part, so I guess it was too complicated to refactor for now.

Goumies: Let me check this out

opencti-platform/opencti-graphql/src/modules/workspace/workspace.graphql
```diff
@@ -95,6 +95,7 @@ type Mutation {
     workspaceEditAuthorizedMembers(id: ID!, input:[MemberAccessInput!]!): Workspace @auth(for: [EXPLORE_EXUPDATE])
     workspaceContextPatch(id: ID!, input: EditContext!): Workspace
     workspaceContextClean(id: ID!): Workspace
+    configurationImport(file: Upload!): String!
```
SouadHadjiat: I wonder about the name of this mutation `configurationImport`, it doesn't explain well that it's a workspace import. In the code we declare it here in workspace.graphql but when we deploy, it won't be declared specifically for workspace domain (you can see it in graphql.ts, it's defined right after cityEdit), and we can use it directly through the API without knowing that it's linked to workspace.

I suggest `workspaceImport` what do you think ?

opencti-platform/opencti-graphql/src/modules/workspace/workspace-domain.ts
```diff
@@ -135,3 +146,49 @@ export const workspaceEditContext = async (context: AuthContext, user: AuthUser,
...
+  await publishUserAction({
+    user,
+    event_type: 'mutation',
+    event_scope: 'create',
+    event_access: 'extended',
+    message: `import ${importWorkspaceCreation.name} workspace`,
+    context_data: { id: workspaceId, entity_type: ENTITY_TYPE_WORKSPACE, input: importWorkspaceCreation }
+  });
+  await notify(BUS_TOPICS[ENTITY_TYPE_WORKSPACE].ADDED_TOPIC, importWorkspaceCreation, user);
+  return workspaceId;
```
SouadHadjiat, line 193: also, to be more consistent with `workspaceAdd` mutation, shouldn't we return the workspace object, not only workspaceId ?

opencti-platform/opencti-front/src/private/components/workspaces/WorkspaceCreation.jsx
Goumies: I have refactored it in the new way.
It takes the same argument after initialization with the mutation.

opencti-platform/opencti-graphql/src/modules/workspace/workspace-domain.ts
SarahBocognano: No it was the point to return just the id

Goumies: We have chosen this approach based on our need and spotted upload in the codebase

opencti-platform/opencti-graphql/src/modules/workspace/workspace.graphql
Goumies: `workspaceImport` implies the import of data and configuration.
We will use `workspaceConfigurationImport` to explicit the intent.

Goumies:
Update, we have processed all your feedback.
Thanks.

PR #4838, opened 2023-11-02
[Frontend] CSVMapper - fix missing value in relationship related-to a…

### LINKED ISSUE #4797: CSV Mapper prevent related-to relationships + missing revoked by
## Description

This is 2 problems on generic relationships.

**related to**
When trying to create a csv mapper with related-to relationships, the available options have weird behavior and prevent the configuration of this kind of relation.

**revoked-by**
When trying to create a csv mapper with related-to relationships, nothing is available for selection

## Environment

1. OpenCTI version: 5.11.12
2. OpenCTI client: frontend

## Reproducible Steps

1. Create a fresh CSV Mapper
2. Add first entity Area
3. Add second entity Cryptocurrency-wallet
4. Create a relationship representation "related-to"
5. Try to link both entity representation

## Expected Output

related-to is compatible with everything so we should be able to configure Area-related to-Crypto or Crypto-related to-Area
This must be also possible for any types available.

## Actual Output

Strange behavior. Option can be available in the second box but not in the first one. When one part is selected, there is no more option in the first one.

image

## Code hints

For missing handling of revoked-by, Relation.ts

`if (values.includes(relationshipType) || relationshipType === 'related-to') {`

jpkha:
### Proposed changes

* Fixed relationship related-to and revoked-by

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/4797

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/data/csvMapper/representations/attributes/CsvMapperRepresentationAttributeRefForm.tsx
+++ b/opencti-platform/opencti-front/src/private/components/data/csvMapper/representations/attributes/CsvMapperRepresentationAttributeRefForm.tsx
@@ -80,25 +80,11 @@ CsvMapperRepresentationAttributeRefFormProps
 
   const { schema } = useAuth();
   const { schemaRelationsTypesMapping, schemaRelationsRefTypesMapping } = schema;
-  const relationshipTypes = resolveTypesForRelationship(
-    schemaRelationsTypesMapping,
-    entityType,
-    attribute.key,
-    fromType,
-    toType,
-  );
-  const relationshipRefTypes = resolveTypesForRelationshipRef(
-    schemaRelationsRefTypesMapping,
-    entityType,
-    attribute.key,
-  );
-  const options = representations
+
+  const filteredOptions = ((representationsOptions: Representation[]) => representationsOptions
     .filter((r) => {
-      return [...relationshipTypes, ...relationshipRefTypes].includes(
-        r.target.entity_type,
-      );
+      return r.id !== representation.id;
     })
-    .filter((r) => r.id !== representation.id)
     .filter((r) => {
       if (attribute.key === 'from' && toId) {
         return r.id !== toId;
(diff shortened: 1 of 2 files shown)


opencti-platform/opencti-front/src/private/components/data/csvMapper/representations/attributes/CsvMapperRepresentationAttributeRefForm.tsx
```diff
@@ -107,7 +93,31 @@ CsvMapperRepresentationAttributeRefFormProps
         return r.id !== fromId;
       }
       return true;
-    });
+    }));
+
+  let options;
+  if (entityType === 'related-to' || entityType === 'revoked-by') {
+    options = filteredOptions(representations);
+  } else {
```
lndrtrbn, line 104: Maybe add a comment to explain why we are doing this? 
From an external point of view (me) this is black magic 🧙

jpkha: Yes, good point @lndrtrbn

jpkha: I add this comment : 
```
// For both entity types, whether they are related-to or revoked-by,
// we don't need to resolve the different types, as they can link any entity between them.
```

Is it clearer with this comment ?

opencti-platform/opencti-front/src/private/components/data/csvMapper/representations/attributes/CsvMapperRepresentationAttributeRefForm.tsx
```diff
@@ -107,7 +93,34 @@ CsvMapperRepresentationAttributeRefFormProps
         return r.id !== fromId;
       }
       return true;
-    });
+    }));
+
+  let options;
+  
+// For both entity types, whether they are related-to or revoked-by,
+// we don't need to resolve the different types, as they can link any entity between them.
```
lndrtrbn:
```suggestion
  // For both entity types, whether they are related-to or revoked-by,
  // we don't need to resolve the different types, as they can link any entity between them.
```
To fix eslint errors

lndrtrbn (approved):

PR #4841, opened 2023-11-03
[backend/frontend] Add EE identity support

### LINKED ISSUE #4765: Display a "By Filigran" logo on login page and on all pages of the app
## Use case

Ability for our comminutiy users to clearly see that Filigran developped the tool and for Entreprise Edition users to remove it, in order to easily create a whitelabled solution
## Proposed Solution

Add a Filigran logo on the login page and on all pages of the app.
Create a toogle within settings to remove this logo (only avaiable to Entreprise Edition users).

Kedae:


opencti-platform/opencti-front/jsconfig.json
```diff
@@ -1,7 +0,0 @@
-{
```
SarahBocognano, line 1: Why did you delete this file ?

SarahBocognano (commented):
Capture d'écran 2023-11-03 110416

Maybe you can close the border ? I find it weird that this is open

opencti-platform/opencti-front/jsconfig.json
Kedae: Because it is not working and confusing :D

Kedae:
> Capture d'écran 2023-11-03 110416
> 
> Maybe you can close the border ? I find it weird that this is open

I can, but it is actually the same border as above separating the items ? @nino-filigran a thought ?

Kedae:
Localization not done, waiting for validation on semantic

nino-filigran:
At first, I was not bothered by it. But indeed, after a 2nd thought, having it closed might be better, I agree with you @Kedae.
From a scalability point of view, if we are to add more toggles specific to EE, then we would also have this "closed" and dedicated space, which is why I'm even more in favor of your suggestion.

Unless also @Jipegien you would like something different?

opencti-platform/opencti-front/src/private/components/nav/LeftBar.jsx
```diff
@@ -502,6 +503,31 @@ const LeftBar = () => {
           </Security>
         </MenuList>
       </Security>
+      {!platform_whitemark && (
+        <MenuItem
+          dense={true}
+          style={{ marginBottom: bannerHeightNumber }}
+          classes={{
+            root: navOpen ? classes.menuLogoOpen : classes.menuLogo,
+          }}
```
labo-flg, line 518: This menu item is clickable but does nothing. Open the filigran website in a new window ?

```suggestion
          }}
          onClick={() => window.open('https://filigran.io/', '_blank')}
```

labo-flg:
I suggest to make these 2 buttons consistent in the UI (same position)

image

image

Kedae:
> I suggest to make these 2 buttons consistent in the UI (same position)
> 
> image
> 
> image

Ah oui tiens, j'ai cassé.

labo-flg:
How far this PR should go in terms of identifying the EE features inside the app ?
We have this nice chip, tooltips and specific color, but what should happen if EE is enabled?

For instance, the "share with organization" button is still in EE color, but the automation (playbooks) screen does not display any clear labelling or color or chip to tell the user he's using some EE cool feature :)

Kedae:
> How far this PR should go in terms of identifying the EE features inside the app ? We have this nice chip, tooltips and specific color, but what should happen if EE is enabled?
> 
> For instance, the "share with organization" button is still in EE color, but the automation (playbooks) screen does not display any clear labelling or color or chip to tell the user he's using some EE cool feature :)

That's a mistake, every 'EE' UI item should disappear when under EE.

nino-filigran:
280243469-b93458be-086a-40ae-a3bf-4c800da8ace4


After talking with  @Jipegien, we agreed that the divider bar in grey + the violet bar is a bit weird.
Therefore, is it possible to do as the image above? (sorry it's ugly but want to ensure it's clear. The blue bar in the image is simply there to show that the violet one should not exist)
- have the divider bar directly in violet and remove the existing violet upper bar? 
- do not close the space between the 2 bars

Kedae:
Every comment as been taken care of (still missing Localization)

opencti-platform/opencti-front/src/private/components/nav/LeftBar.jsx
labo-flg: @Kedae wdyt?

yassine-ouaamou:
Since the organization admin feature is EE, it would be better to block this part also:
image

Kedae:
> Since the organization admin feature is EE, it would be better to block this part also: image

Good spot, will do.

yassine-ouaamou:
In the Activity section, I couldn't click the agree checkbox when I click on "Enable EE" from the pop-up. It works fine from the other one.
Here are some shots to make it clear:
Working:
ee_agree_bug_working_path

Not working:
ee_agree_bug

Archidoit:
The 'share' button seems to be present in the overview of every entity... Except for entities of type Data Sources 
image

Archidoit:
In the pop-up to activate EE, the 'associated blog post' link redirects without opening a new tab and we 'loose' the openCTI page (whereas a new tab is openend when we click on the other links).

image

yassine-ouaamou:
@Kedae Can you update the branch with master so I can do another test and approve this PR, please?
Merci

yassine-ouaamou:
ee_activitiy_clicked
ee_file_indexing_clic
File indexing should be in purple as Activity

yassine-ouaamou (approved):

PR #4847, opened 2023-11-03
[backend] new logic engine for resolving FilterGroups for event streams (#4536)

labo-flg:
### Proposed changes

This code is completely independent and is only used by unit tests for now.
It implements our new filtering logic with nested, recursive Filter groups, in the case of stream events.

It's closely related to @Archidoit's work and will require some integration, so things might change on the edges.

I strongly suggest to read the code in parallel with unit tests, and to start "from the bottom"
1. boolean-logic-engine.ts
2. stix-testers.ts
3. stix-filtering.ts

### Related issues

* #4536 

### Checklist

### Further comments

This implementation separates the responsibilities properly:

* New boolean logic engine
  * takes a FilterGroup and some data, to test the filter against the data
  * Is independent of the actual model of the data
  * provides functions to test a Filter against dates, strings, booleans, numbers
  
* A set of individual testers functions, one for each filter key currently possible
  * knows the stix model and how to extract the data
  * has no filtering logic, only "stix logic"

* final function `isStixMatchFilterGroups` 
  * meant to replace `isStixMatchFilter`,  the one used for streams, triggers and playbooks when we want to check a filter against a stix object
  * only handles the context (user, filter adaptation..) but no filtering logic


opencti-platform/opencti-graphql/src/schema/stixEmbeddedRelationship.ts
```diff
@@ -27,6 +27,10 @@ const RELATIONS_EMBEDDED_STIX_ATTRIBUTES = [
 ];
 // eslint-disable-next-line
 export const stixRefsExtractor = (data: any, idGenerator?: (key: string, data: unknown) => string) => {
+  if (!data.extensions?.[STIX_EXT_OCTI]?.type) {
+    return [];
+  }
```
labo-flg, line 32: this is a bugfix,  we might get an undefined below otherwise

opencti-platform/opencti-graphql/src/utils/stix-filtering/stix-testers.ts
```diff
@@ -0,0 +1,369 @@
...
+  // refs
+  aggregatedStixValues.push(...stixRefsExtractor(stix, generateStandardId));
+
+  return testStringFilter(filter, aggregatedStixValues);
+};
+
+/**
+ * Gives the right tester function according to the filter key.
+ * If the key is not handled, returns a function that always return false.
+ * TODO: make it dependent on the schema.
```
labo-flg: The whole code in this file could be replaced by a more generic approach:
* update the schema to add the stix paths, value type and multiplicity
* extract the data thanks to the schema, according to the attribute key

And we would remove this ugly part of the code.
Now, it's far easier to say than done, so this version is MVP : it is based on the current filters available in the UI and handle each case specifically. We expect no regression. 

In a next version we could work on this complex task involving schema.

opencti-platform/opencti-graphql/tests/02-integration/01-database/stix-filtering-test.ts
```diff
@@ -0,0 +1,104 @@
+import { describe, expect, it } from 'vitest';
+
+import { ADMIN_USER, buildStandardUser, testContext } from '../../utils/testQuery';
+import { isStixMatchFilterGroup } from '../../../src/utils/stix-filtering/stix-filtering';
+import type { FilterGroup } from '../../../src/utils/stix-filtering/filter-group';
+
+import stixReports from '../../data/stream-events/stream-event-stix2-reports.json';
```
labo-flg, line 7: I could not rely on the data file `DATA-TEST-STIX2_v2.json` as it's exported data and not what we would find in stream events (that contains more data as it seems).

I've added a whole set of stix objects in tests/data/stream-events on which I base my tests.

opencti-platform/opencti-graphql/src/utils/stix-filtering/stix-filtering.ts
```diff
@@ -0,0 +1,71 @@
...
+import type { Filter, FilterGroup } from './filter-group';
+import { getEntitiesMapFromCache } from '../../database/cache';
+import { isUserCanAccessStixElement, SYSTEM_USER } from '../access';
+import { ENTITY_TYPE_RESOLVED_FILTERS } from '../../schema/stixDomainObject';
+import type { AuthContext, AuthUser } from '../../types/user';
+import type { StixId, StixObject } from '../../types/stix-common';
+import { extractStixRepresentative } from '../../database/stix-representative';
+
+// TODO: changed by Cathia for #2686, to integrate properly next
+const LABEL_FILTER = 'objectLabel';
```
Archidoit, line 19: Note: other keys have been changed (ex: objectMarking instead of markedBy)

opencti-platform/opencti-graphql/tests/02-integration/01-database/stix-filtering-test.ts
```diff
@@ -0,0 +1,104 @@
...
+const stixReport = stixReports[0]; //  confidence 3, revoked=false, labels=report, TLP:TEST
+const stixIndicator = stixIndicators[0]; // confidence 75, revoked=true, no label
+
+const TLP_CLEAR_ID = 'marking-definition--613f2e26-407d-48c7-9eca-b8e91df99dc9';
+const WHITE_TLP = { standard_id: TLP_CLEAR_ID, internal_id: '' };
+
+describe('Stix Filtering', () => {
+  it('matches stix objects with basic filter groups', async () => {
+    let filterGroup: FilterGroup = {
+      mode: 'AND',
```
Archidoit: mode is in lower case

opencti-platform/opencti-graphql/tests/02-integration/01-database/stix-filtering-test.ts
```diff
@@ -0,0 +1,104 @@
...
+      filters: [{
+        key: ['entity_type'],
+        mode: 'OR',
```
Archidoit: same for local modes

opencti-platform/opencti-graphql/tests/02-integration/01-database/stix-filtering-testers-test.ts
```diff
@@ -0,0 +1,670 @@
...
+import type { Filter } from '../../../src/utils/stix-filtering/filter-group';
+
+describe('Stix filter testers', () => {
+  describe('by Markings (key=markedBy)', () => {
+    const stixWithMarkings = stixIndicators[0];
+    const stixWithoutMarkings = stixReports[0];
+
+    it('should test positive for a stix object with matching filter', () => {
+      let filter: Filter = {
+        key: ['markedBy'],
```
Archidoit: objectMarking

opencti-platform/opencti-graphql/tests/02-integration/01-database/stix-filtering-testers-test.ts
```diff
@@ -0,0 +1,670 @@
...
+    });
+  });
+
+  describe('by Entity Type (key=entity_type)', () => {
+    const report = stixReports[0];
+    const stixWithExtType = stixIndicators[0];
+
+    it('should test positive for a stix object with matching filter, using generated internal type', () => {
+      let filter: Filter = {
+        key: ['entity_type'],
```
Archidoit, line 49: Remark: for me (feed, dynamic filters etc), backend tests are also working if the key is a string (instead of an array of strings)

opencti-platform/opencti-graphql/tsconfig.json
```diff
@@ -10,6 +10,7 @@
     "noImplicitAny": true,
     "noFallthroughCasesInSwitch": true,
     "moduleResolution": "node",
+    "resolveJsonModule": true,
```
Archidoit, line 13: why? should we keep this?

opencti-platform/opencti-graphql/tests/02-integration/01-database/stix-filtering-testers-test.ts
labo-flg: The type for keys on your side is `string | string[]` ?

opencti-platform/opencti-graphql/tsconfig.json
labo-flg: it's to import .json files (the data file for the tests) in TS.
So far, all test files were in JS.

opencti-platform/opencti-graphql/tests/02-integration/01-database/stix-filtering-testers-test.ts
Archidoit: The type is an array 
image

Archidoit: but string are also handled, thus my tests are working...

Archidoit: (old behavior)

Archidoit: You can keep your behavior, it was just a remark, for now, the back handle both (from old behavior)

opencti-platform/opencti-graphql/src/utils/stix-filtering/stix-filtering.ts
labo-flg: I've checked and I updated all the keys I'm using for now (the ones that were handled by isStixMatchFilter).
Obviously, when we will extend to other keys I'll make sure to use the new keys.

opencti-platform/opencti-graphql/src/utils/stix-filtering/boolean-logic-engine.ts
```diff
@@ -0,0 +1,187 @@
...
+  return false;
+};
+
+/**
+ * Implementation of testGenericFilter for string values.
+ * String comparison is insensitive to case.
+ */
+export const testStringFilter = (filter: FilterExcerpt, stixCandidates: string[]) => {
+  const filterValuesLowerCase = filter.values.map((v) => v.toLowerCase());
+  const stixValuesLowerCase = stixCandidates.map((v) => v.toLowerCase());
```
labo-flg: Worth noting : we're case insensitive by default.

opencti-platform/opencti-graphql/src/utils/stix-filtering/boolean-logic-engine.ts
```diff
@@ -0,0 +1,187 @@
...
+  if (mode === 'and') {
+    // we need to find all of them or none of them
+    return (operator === 'eq' && adaptedFilterValues.every((v) => stixCandidates.includes(v)))
+      || (operator === 'not_eq' && adaptedFilterValues.every((v) => !stixCandidates.includes(v)))
+
+      // In real cases, there is only 1 filter value with the next operators (not much sense otherwise)
+      || (operator === 'lt' && adaptedFilterValues.every((v) => stixCandidates.some((c) => c < v)))
+      || (operator === 'lte' && adaptedFilterValues.every((v) => stixCandidates.some((c) => c <= v)))
+      || (operator === 'gt' && adaptedFilterValues.every((v) => stixCandidates.some((c) => c > v)))
+      || (operator === 'gte' && adaptedFilterValues.every((v) => stixCandidates.some((c) => c >= v)));
```
jpkha, line 55: What about replacing it by 
```suggestion
      || (['lt', 'lte', 'gt', 'gte'].includes(operator) && adaptedFilterValues.every((v) => stixCandidates.some((c) => c < v)))
```

opencti-platform/opencti-graphql/src/utils/stix-filtering/boolean-logic-engine.ts
```diff
@@ -0,0 +1,187 @@
...
+  }
+
+  if (mode === 'or') {
+    // we need to find one of them or at least one is not found
```
jpkha, line 59: Is it the same logic as `and` condition ? If yes, we could refactor this part.

opencti-platform/opencti-graphql/src/utils/stix-filtering/boolean-logic-engine.ts
labo-flg: there are all different comparison operators, why only use `c < v` ? it's only for "lower than"

opencti-platform/opencti-graphql/src/utils/stix-filtering/boolean-logic-engine.ts
jpkha: I don't understand, why are we doing the mapping if it worth nothing ?

opencti-platform/opencti-graphql/src/utils/stix-filtering/boolean-logic-engine.ts
labo-flg: we are using `some` instead of `every` as only one matching filter value is sufficient.

Maybe there are some optimization to do, but generally speaking I favored readability.

opencti-platform/opencti-graphql/src/utils/stix-filtering/boolean-logic-engine.ts
```diff
@@ -0,0 +1,187 @@
...
+  const filterValuesAsBooleans = filter.values.map((v) => v.toLowerCase() === 'true' || v.toLowerCase() === 'yes' || v === '1');
+  return testGenericFilter<boolean>(filter, filterValuesAsBooleans, toValidArray(stixCandidate));
+};
+
+/**
+ * Implementation of testGenericFilter for numerical values.
+ * Filter values are parsed as integers.
+ */
+export const testNumericFilter = (filter: FilterExcerpt, stixCandidate: number | null | undefined) => {
+  const filterValuesAsNumbers = filter.values.map((v) => parseInt(v, 10)).filter((n) => !Number.isNaN(n));
```
jpkha: By default the base is 10 for the parseInt. So it's not mandatory to add the second argument but we can let's it like that if you want to explicitly say the value.

opencti-platform/opencti-graphql/src/utils/stix-filtering/boolean-logic-engine.ts
labo-flg: worth noting. Like "worth to note" :)

opencti-platform/opencti-graphql/src/utils/stix-filtering/boolean-logic-engine.ts
labo-flg: it's made mandatory by a linter rule https://eslint.org/docs/latest/rules/radix
Maybe an outdated rule, but still commonly used.

opencti-platform/opencti-graphql/src/utils/stix-filtering/boolean-logic-engine.ts
jpkha: I tried to read it many time but did not saw it.

opencti-platform/opencti-graphql/src/utils/stix-filtering/boolean-logic-engine.ts
jpkha: The only thing I think for a solution is to create a function and pass every` or `some` as an argument.

opencti-platform/opencti-graphql/src/utils/stix-filtering/stix-testers.ts
```diff
@@ -0,0 +1,378 @@
...
+ * @param key
+ */
+export const getStixTesterFromFilterKey = (key: string): TesterFunction => {
+  switch (key) {
+    case MARKING_FILTER:
```
jpkha: What about a mapping instead of a switch case ?
Like : 
```
{
  MARKING_FILTER: testMarkingFilter,
  TYPE_FILTER: testEntityType
 [...]
}

```

opencti-platform/opencti-graphql/src/utils/stix-filtering/stix-testers.ts
```diff
@@ -0,0 +1,378 @@
...
+};
+
+/**
+ * INSTANCE TYPE (elementId)
+ * - elementId is id in stix extension
+ * - useSideEventMatching arg to optionally test against relations and refs of the object
+ */
+export const testInstanceType = (stix: any, filter: Filter, useSideEventMatching = false) => {
+  const stixValues: string[] = toValidArray(stix.extensions?.[STIX_EXT_OCTI]?.id);
+  if (!useSideEventMatching) {
```
labo-flg: The case `useSideEventMatching` will be handled by a new filter key as we do not want exotic behavior like this (we try to have all filtering techniques embedded as filter keys.

opencti-platform/opencti-graphql/src/utils/stix-filtering/boolean-logic-engine.ts
labo-flg: yes, but honestly I prefer a little bit of (controlled) code duplication to make the code straightforward and simple to read for anyone.

opencti-platform/opencti-graphql/src/utils/stix-filtering/stix-testers.ts
labo-flg: yes, probably nicer. I'll do that !

opencti-platform/opencti-graphql/src/utils/stix-filtering/boolean-logic-engine.ts
```diff
@@ -0,0 +1,187 @@
...
+  const stixDate = moment(new Date(stixCandidate));
+  if (!stixDate.isValid()) {
+    // This is actually an error case that should not happen (invalid stix)
+    return false;
+  }
+
+  if (mode === 'and') {
+    // NOTE: equality is very strict (milliseconds)
+    return (operator === 'eq' && filterValuesAsDates.every((v) => stixDate.isSame(v)))
+      || (operator === 'not_eq' && filterValuesAsDates.every((v) => !stixDate.isSame(v)))
```
labo-flg, line 130: isSame is maybe too strict as the filter selected in UI are only dates, no time.

labo-flg: probably use isSameDay

opencti-platform/opencti-graphql/src/utils/stix-filtering/stix-testers.ts
labo-flg: new filter key `connectedToId` encapsulate this case now

opencti-platform/opencti-graphql/src/utils/stix-filtering/boolean-logic-engine.ts
labo-flg: this has to be discussed a bit more : what's the expected granularity for date filtering. UI wise it's day, but API wise it could be millisec.

opencti-platform/opencti-graphql/src/utils/stix-filtering/stix-testers.ts
labo-flg: for the record, I think it could be like the stix `converter` or the `representative` inside an entity in the schema. 

A field `stixFilter` in `AttributeDefinition` for the the individual tester logic ?

opencti-platform/opencti-graphql/src/utils/stix-filtering/boolean-logic-engine.ts
```diff
@@ -0,0 +1,187 @@
...
+      || (operator === 'gt' && adaptedFilterValues.some((v) => stixCandidates.some((c) => c > v)))
+      || (operator === 'gte' && adaptedFilterValues.some((v) => stixCandidates.some((c) => c >= v)));
+  }
+
+  return false;
+};
+
+/**
+ * Implementation of testGenericFilter for string values.
+ * String comparison is insensitive to case.
```
labo-flg: also trim the values to compare !

opencti-platform/opencti-graphql/src/utils/stix-filtering/boolean-logic-engine.ts
```diff
@@ -0,0 +1,187 @@
...
+export const testBooleanFilter = (filter: FilterExcerpt, stixCandidate: boolean | null | undefined) => {
+  const filterValuesAsBooleans = filter.values.map((v) => v.toLowerCase() === 'true' || v.toLowerCase() === 'yes' || v === '1');
+  return testGenericFilter<boolean>(filter, filterValuesAsBooleans, toValidArray(stixCandidate));
+};
+
+/**
+ * Implementation of testGenericFilter for numerical values.
+ * Filter values are parsed as integers.
+ */
+export const testNumericFilter = (filter: FilterExcerpt, stixCandidate: number | null | undefined) => {
```
labo-flg, line 97: to be more resilient we need to parse float instead

opencti-platform/opencti-graphql/src/utils/stix-filtering/boolean-logic-engine.ts
```diff
@@ -0,0 +1,187 @@
...
+ * @param data data to test
+ * @param filterGroup complex filter group object with nested groups and filters
+ * @param getTesterFromFilterKey function that gives a function to test a filter, according to the filter key
+ *                               see unit tests for an example.
+ */
+export const testFilterGroup = (data: any, filterGroup: FilterGroup, getTesterFromFilterKey: (key: string) => TesterFunction) : boolean => {
+  if (filterGroup.mode === 'and') {
+    const results: boolean[] = [];
+    if (filterGroup.filters.length > 0) {
+      // note that we are not compatible with multiple keys yet, so we'll always check the first one only
```
labo-flg, line 168: in case of multiple keys, we shall warn and reject this case completly

opencti-platform/opencti-graphql/src/utils/stix-filtering/stix-testers.ts
```diff
@@ -0,0 +1,377 @@
...
+  [PRIORITY_FILTER]: testPriority,
+  [REVOKED_FILTER]: testRevoked,
+  [SEVERITY_FILTER]: testSeverity,
+  [SCORE_FILTER]: testScore,
+  [TYPE_FILTER]: testEntityType,
+  [WORKFLOW_FILTER]: testWorkflow,
+  [INSTANCE_FILTER]: testInstance,
+
+  // special keys (more complex behavior)
+  [CONECTED_TO_FILTER]: testConnectedTo,
```
labo-flg: missing N

opencti-platform/opencti-graphql/src/utils/stix-filtering/stix-testers.ts
```diff
@@ -0,0 +1,377 @@
...
+/**
+ * Gives the right tester function according to the filter key.
+ * If the key is not handled, returns a function that always return false.
+ * @param key
+ */
+export const getStixTesterFromFilterKey = (key: string): TesterFunction => {
+  if (!FILTER_KEY_TESTERS_MAP[key]) {
+    logApp.warn(`Unrecognized filter key when matching stix object: [${key}]`);
+    return () => false;
+  }
```
labo-flg: better to validate filter keys (no multiple + exist in map) and throw error on top before matching.
Warn is too lazy.

opencti-platform/opencti-graphql/src/utils/stix-filtering/boolean-logic-engine.ts
labo-flg: seen with @richard-julien : our storage granularity is the millisec, our API when testing equality should be granular in millisecs too. We'll keep everything that way.

The UI shall be adapted to this fact, and not suggest to use the `eq / not_eq` operators when it comes to date (cc @jpkha).

As an evolution, let's think about a new operator `intersects`. 
`<created_at> intersects "2023-08-24"` would mean "creation date is in the **period** "24th August 2023".

labo-flg:
all comments addressed in last commit.

opencti-platform/opencti-graphql/tests/02-integration/01-database/stix-filtering-testers-test.ts
```diff
@@ -0,0 +1,693 @@
...
+  describe('by Markings (key=objectMarking)', () => {
+    const stixWithMarkings = stixIndicators[0];
+    const stixWithoutMarkings = stixReports[0];
+
+    it('should test positive for a stix object with matching filter', () => {
+      let filter: Filter = {
+        key: ['objectMarking'],
+        mode: 'or',
+        operator: 'eq',
+        values: ['<some-id>', 'marking-definition--613f2e26-407d-48c7-9eca-b8e91df99dc9']
```
labo-flg, line 27: no it's cheating, the filter contains the id and not the standard id

opencti-platform/opencti-graphql/src/utils/stix-filtering/stix-testers.ts
labo-flg: We'll tackle this later, once a first working version is out. That's the priority.

opencti-platform/opencti-graphql/tests/02-integration/01-database/stix-filtering-testers-test.ts
labo-flg: fixed in last commits, we're properly mapping the filter values when required.

labo-flg:
Last commits address most of the changes requested by @richard-julien after our pair review.
* resolution map is built from cache in nominal case, but an intermediate function allows to mock this mapping in unit tests
* trim string before comparison
* parse numbers as float instead of just integers
* validate filters and throw errors if not compatible (multi key, unrecognized key)

The following changes will be done AFTER merge with Cathia's feature branch
* move the filter key validation from the isStixMatchFilterGroup function to the creation/update of saved filters
* all plumbing works to call the new code, check the filter key renamings, etc.

labo-flg:
Last round validated by @richard-julien 
* we no longer need to handle the filter key `elementId` in streams et al. The only case handled is the new `connectedToId` which is basically `elementId` with `useSideEventMatching=true`. We'll add this if the feature is requested (ability to filter a stream on a unique instance id)

Archidoit (approved):

PR #4853, opened 2023-11-06
[Frontend] Put configurations options into dialog (#4505)

jpkha:
### Proposed changes

* CSV mappers : Put configurations options into dialog

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/4505

### Checklist

### Further comments

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/data/csvMapper/representations/attributes/CsvMapperRepresentationAttributeForm.tsx
+++ b/opencti-platform/opencti-front/src/private/components/data/csvMapper/representations/attributes/CsvMapperRepresentationAttributeForm.tsx
@@ -11,6 +11,10 @@ import {
 } from '@components/data/csvMapper/representations/attributes/Attribute';
 import { useFormikContext } from 'formik';
 import { CsvMapper } from '@components/data/csvMapper/CsvMapper';
+import CsvMapperRepresentationDialogOption
+  from '@components/data/csvMapper/representations/attributes/CsvMapperRepresentationDialogOption';
+import CsvMapperRepresentionAttributeSelectedConfigurations
+  from '@components/data/csvMapper/representations/attributes/CsvMapperRepresentionAttributeSelectedConfigurations';
 import { useFormatter } from '../../../../../../components/i18n';
 import { isEmptyField } from '../../../../../../utils/utils';
 
@@ -21,6 +25,7 @@ const useStyles = makeStyles(() => ({
     gridTemplateColumns: '1fr 1fr 1fr',
     alignItems: 'center',
     marginTop: '10px',
+    gap: '10px',
   },
   inputError: {
     '& fieldset': {
@@ -138,11 +143,21 @@ CsvMapperRepresentationAttributeFormProps
         />
       </div>
       <div>
-        <CsvMapperRepresentationAttributeOptions
-          attribute={attribute}
-          indexRepresentation={indexRepresentation}
-        />
+        {
+          (attribute.type === 'date' || attribute.multiple)
+          && <CsvMapperRepresentationDialogOption attribute={attribute}>
+            <CsvMapperRepresentationAttributeOptions
+                attribute={attribute}
+                indexRepresentation={indexRepresentation}
+            />
+          </CsvMapperRepresentationDialogOption>
+        }
       </div>
+        <CsvMapperRepresentionAttributeSelectedConfigurations
+          configuration={ formikContext.values.representations[indexRepresentation]
+            .attributes[indexAttribute]?.column?.configuration}
+        />
+
     </div>
   );
 };
--- a/opencti-platform/opencti-front/src/private/components/data/csvMapper/representations/attributes/CsvMapperRepresentationDialogOption.tsx
+++ b/opencti-platform/opencti-front/src/private/components/data/csvMapper/representations/attributes/CsvMapperRepresentationDialogOption.tsx
@@ -1,0 +1,58 @@
+import React, { FunctionComponent, ReactNode } from 'react';
+import Dialog from '@mui/material/Dialog';
+import Button from '@mui/material/Button';
+import DialogTitle from '@mui/material/DialogTitle';
+import DialogContent from '@mui/material/DialogContent';
+import DialogActions from '@mui/material/DialogActions';
+import { CogOutline } from 'mdi-material-ui';
+import IconButton from '@mui/material/IconButton';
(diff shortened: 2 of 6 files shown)


opencti-platform/opencti-front/src/private/components/data/csvMapper/representations/attributes/CsvMapperRepresentationAttributeForm.tsx
```diff
@@ -138,11 +143,21 @@ CsvMapperRepresentationAttributeFormProps
         />
       </div>
       <div>
-        <CsvMapperRepresentationAttributeOptions
-          attribute={attribute}
-          indexRepresentation={indexRepresentation}
-        />
+        {
+          (attribute.type === 'date' || attribute.multiple)
```
labo-flg, line 147: maybe put this test in a dedicated function `hasOptions` to encapsulate this logic

opencti-platform/opencti-front/src/private/components/data/csvMapper/representations/attributes/CsvMapperRepresentationDialogOption.tsx
```diff
@@ -0,0 +1,57 @@
...
+  const handleClickOpen = () => {
+    setOpen(true);
+  };
+
+  const handleClose = () => {
+    setOpen(false);
+  };
+  return (
+        <>
+            <IconButton
```
labo-flg: this file seems to mix indent 4 (jsx) and 2 (the rest). 
This is strange 🤔

labo-flg:
I'm wondering if we need to make some options mandatory.
For instance, if we set a column index for a list, but the list separator is not set. Should we mark the field as invalid ? (setting icon outlined in red + mapper marked as invalid). I'm inclined to think this way 🤔

Wdyt ? 

For the date pattern, there is a default behavior (iso format) so the `date_pattern` is indeed optional.

EDIT: semantically that would make these "options" more like "parameters", these parameters being optional or not. But I'm being too picky maybe 😄

opencti-platform/opencti-front/src/private/components/data/csvMapper/representations/attributes/CsvMapperRepresentationAttributeForm.tsx
jpkha: I could but I decide to not do it because in the next chunk, I will remove this condition as the dialog should be available to each representation with on option of adding a default value.

jpkha:
@labo-flg I think, you are right, I think we could make as mandatory.
But I think we need to rethink the UX. 
IMO the main problem would be having a mandatory field within the dialog.
It won't be obvious to click on the icon to define the field or at least we should add a text to give a direction.

opencti-platform/opencti-front/src/private/components/data/csvMapper/representations/attributes/CsvMapperRepresentationAttributeForm.tsx
labo-flg: ok, clear !

Jipegien:
> I'm wondering if we need to make some options mandatory. For instance, if we set a column index for a list, but the list separator is not set. Should we mark the field as invalid ? (setting icon outlined in red + mapper marked as invalid). I'm inclined to think this way 🤔
> 
> Wdyt ?
> 
> For the date pattern, there is a default behavior (iso format) so the `date_pattern` is indeed optional.
> 
> EDIT: semantically that would make these "options" more like "parameters", these parameters being optional or not. But I'm being too picky maybe 😄

We don't need to have this kind of control. The use case of the CSV mappers is for an administrator to define a parser to be used by end users. Administrators must know what they are doing and what type of files they are handling. So it is not mandatory to have this type of control. Plus, how can you know that the defined column index will indeed be a list of multiple elements? The administrator may also want to only keep the first element...

opencti-platform/opencti-front/src/private/components/data/csvMapper/representations/attributes/CsvMapperRepresentationDialogOption.tsx
```diff
@@ -0,0 +1,58 @@
...
+  };
+  return (
+    <>
+      <IconButton
+        color="primary"
+        aria-label={t('Settings')}
+        onClick={handleClickOpen}
+        size="large"
+        disabled={!attribute.column?.column_name}
+      ><CogOutline/></IconButton>
```
labo-flg:
```suggestion
      >
        <CogOutline/>
      </IconButton>
```

opencti-platform/opencti-front/src/private/components/data/csvMapper/representations/attributes/CsvMapperRepresentionAttributeSelectedConfigurations.tsx
```diff
@@ -0,0 +1,40 @@
...
+  return <div className={classes.attributeOptionsContainer}>
+        {
+            configuration.pattern_date
+          && <div>{t('Date pattern')}: <span className={classes.selectedOption}>{configuration.pattern_date}</span></div>
+        }
+        {
+            configuration.separator
+          && <div>{t('List separator')}: <span className={classes.selectedOption}>{configuration.separator}</span></div>
+        }
+    </div>;
```
labo-flg:
```suggestion
  return (
    <div className={classes.attributeOptionsContainer}>
    {
      configuration.pattern_date && (
        <div>{t('Date pattern')}: <span className={classes.selectedOption}>{configuration.pattern_date}</span></div>
      )
    }
    {
      configuration.separator && (
        <div>{t('List separator')}: <span className={classes.selectedOption}>{configuration.separator}</span></div>
      )
    }
    </div>
  )
```

labo-flg: sorry to bother you with this, but it's very inconsistent in terms of indentation and code styling in general.

jpkha: I added the option on the IDE to reformat code automatically on save, I hope it will solve this issue.
Thanks for the review, and sorry to let you spend time on this type of issue

jpkha:
Here the latest display of the feature
image

labo-flg (approved):

PR #4871, opened 2023-11-09
[frontend/backend] Fix changing entity name to same as alias

### LINKED ISSUE #4798: Cannot modify the name of an entity if already present in the alias of the same entity
## Description

Cannot modify the name of an entity if already present in the alias of the same entity.

## Reproducible Steps

1. Create an intrusion set "Intrusion Set A"
2. Add the alias "Intrusion Set B" and "Intrusion Set A"
3. Try to modify the name with "Intrusion Set B"

Megafredo:
### Proposed changes

* Fix changing entity name to same as alias 
* Fix creating alias name to same as entity

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/4798

### Checklist

### Further comments

Behavior before fix :
OpenCTI_alias_before_fix

Behavior after fix : 
OpenCTI_alias_after_fix

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/common/stix_domain_objects/StixDomainObjectHeader.jsx
+++ b/opencti-platform/opencti-front/src/private/components/common/stix_domain_objects/StixDomainObjectHeader.jsx
@@ -246,6 +246,7 @@ const StixDomainObjectHeader = (props) => {
       setOpenCommitCreate(false);
       setNewAlias('');
       resetForm();
+      MESSAGING$.notifyError('You can\'t add the same alias as the name');
       return;
     }
     if (
--- a/opencti-platform/opencti-front/src/private/components/common/stix_domain_objects/StixDomainObjectHeader.jsx
+++ b/opencti-platform/opencti-front/src/private/components/common/stix_domain_objects/StixDomainObjectHeader.jsx
@@ -239,6 +239,15 @@ const StixDomainObjectHeader = (props) => {
 
   const onSubmitCreateAlias = (values, { resetForm, setSubmitting }) => {
     const currentAliases = getCurrentAliases();
+    const normalizeNameEntity = (stixDomainObject.name).toLowerCase().trim();
+    const normalizeNewAlias = newAlias.toLowerCase().trim();
+    if (normalizeNameEntity === normalizeNewAlias) {
+      setOpenAlias(false);
+      setOpenCommitCreate(false);
+      setNewAlias('');
+      resetForm();
+      return;
+    }
     if (
       (currentAliases === null || !currentAliases.includes(newAlias))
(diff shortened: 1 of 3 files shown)


labo-flg:
Would it make more sense for the user (in terms of UX) to have a validation in the form instead of "not doing the requested changes" being his back?

I mean, you edit the name, put one of its alias, the form is red with proper error like "Cannot rename to an alias".

IMO disallowing bad behavior with explanation is better UX (educating user).

Megafredo:
Yes indeed, having a visual way to communicate why adding the alias does not work can be a good approach (in terms of UX).

opencti-platform/opencti-front/src/private/components/common/stix_domain_objects/StixDomainObjectHeader.jsx
```diff
@@ -240,6 +240,15 @@ const StixDomainObjectHeader = (props) => {
...
   const onSubmitCreateAlias = (values, { resetForm, setSubmitting }) => {
     const currentAliases = getCurrentAliases();
+    const normalizeNameEntity = (stixDomainObject.name).toLowerCase().trim();
+    const normalizeNewAlias = newAlias.toLowerCase().trim();
+    if (normalizeNameEntity === normalizeNewAlias) {
+      setOpenAlias(false);
+      setOpenCommitCreate(false);
+      setNewAlias('');
+      resetForm();
+      return;
```
SouadHadjiat, line 250: I suggest we display an error message here to notify the user that the alias can't be named the same as the entity name.

labo-flg (approved):
Considering we leave the UX improvements for later, it looks ok to me !

opencti-platform/opencti-graphql/src/database/middleware.js
```diff
@@ -1801,8 +1802,19 @@ const updateAttributeRaw = async (context, user, instance, inputs, opts = {}) =>
 export const updateAttributeMetaResolved = async (context, user, initial, inputs, opts = {}) => {
   const { locks = [], impactStandardId = true } = opts;
   const updates = Array.isArray(inputs) ? inputs : [inputs];
+
+  const filteredUpdates = updates.map((item) => {
+    if (item.key === ATTRIBUTE_ALIASES || item.key === ATTRIBUTE_ALIASES_OPENCTI) {
+      const filteredValues = item.value.filter(e => normalizeName(e) !== normalizeName(initial.name));
```
SouadHadjiat, line 1808: are we sure item.value is defined and is a list ?

richard-julien:
Just thinking with Souad about a little missing piece at creation.
Alias should also be purged of the name when creating the entity.
Dont hesitate to join me to have more information @Megafredo

Issue #4863, opened 2023-11-08
Adding a tag to a dashboard is not working


SamuelHassine:
## Description

1. Create a dashboard
2. Go in the dashboard
3. Try to add a tag on the top right "+"

=> Nothing happens.

PR #4875, opened 2023-11-09
[frontend] Fix Adding a tag to a dashboard is not working (#4863)

SarahBocognano:
### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/4863


helene-nguyen:
Adding one tag works fine :) But when adding multiple tags, the `MORE...` button appears and when click on it, nothing happens

image

It would be great to align the `WORK...` button with the others :)

What do you mean by the `WORK...` button ?  @helene-nguyen

Archidoit:
When I add several tags, there is an overlap: 
image

But if I add enough tags, there is no more overlap:  
image


@Archidoit Yep I'm trying to work on it. Since OpenCTI is not responsive it's overlapping on those one -_-'

SarahBocognano:
Update: After a talk with the product team. Knowing that the app is not really responsive, we've decided to put the tags under the buttons. All comments have been resolved

SamuelHassine:
Adding more than 5 tags displays the "More" button which is not working in this branch.

Also, I'm not sure about putting the tags below, which is not so user friendly for dashboards. I would prefer we hide any overflow on the line, even if only one tag is displayed. And have a button to open the rest.

SarahBocognano:
Update: style problem fix, and I align the "More" tags button with the ones in aliases in entities

helene-nguyen:
Canwe add more than 5 tags? The 'plus' button disappears and the 'more' button doesn't work

image

@helene-nguyen  You're not supposed to have this visual. The more button is not there anymore, it is manage like the aliases.

SamuelHassine:
And the "More" button should have a margin right to avoid to be closed to the orange button. Also alignment is not correct we should margin top negative to align with current right buttons position.

@SamuelHassine  It's not a more button anymore, it is manage like the aliases. I'm not sure that the comment up matter because it's not align with the last version of this PR.

Kedae:
image
Aliases or tag ?

opencti-platform/opencti-front/src/private/components/workspaces/WorkspaceHeader.jsx
```diff
@@ -102,52 +117,68 @@ const WorkspaceHeader = ({
...
+  const openTagsCreate = false;
   const [openTag, setOpenTag] = useState(false);
+  const [newTag, setNewTag] = useState('');
   const [openTags, setOpenTags] = useState(false);
   const userCanManage = workspace.currentUserAccessRight === 'admin';
   const userCanEdit = userCanManage || workspace.currentUserAccessRight === 'edit';
   const [displayDuplicate, setDisplayDuplicate] = useState(false);
   const handleCloseDuplicate = () => setDisplayDuplicate(false);
   const [duplicating, setDuplicating] = useState(false);
+  const tags = R.propOr([], 'tags', workspace);
```
marieflorescontact:
```suggestion
  const tags = workspace && workspace.tags ? workspace.tags : [];
```

opencti-platform/opencti-front/src/private/components/workspaces/WorkspaceHeader.jsx
```diff
@@ -102,52 +117,68 @@ const WorkspaceHeader = ({
...
+        && newTag !== ''
     ) {
       commitMutation({
         mutation: workspaceMutation,
         variables: {
           id: workspace.id,
           input: {
             key: 'tags',
-            value: R.append(data.new_tag, currentTags),
+            value: R.append(newTag, currentTags),
```
marieflorescontact:
```suggestion
            value: [...currentTags, newTag],

```

marieflorescontact (commented):
locally tested 👌

helene-nguyen:
Locally tested too, everything seems good :)

helene-nguyen (approved):

PR #4892, opened 2023-11-14
[frontend] Fix error in container knowledge views (#3242)

SouadHadjiat:
### Proposed changes

* fix error when workspace is undefined : replace workspace prop with handleExportDashboard so we keep all workspace logic in its dedicated component

### Related issues

* Error in testing platform : unknown error occured for all containers knowledge views (reports, cases, ...)

### Checklist

### Further comments

### DIFF
--- a/opencti-platform/opencti-front/src/components/ExportButtons.jsx
+++ b/opencti-platform/opencti-front/src/components/ExportButtons.jsx
@@ -12,7 +12,6 @@ import Tooltip from '@mui/material/Tooltip';
 import ToggleButtonGroup from '@mui/material/ToggleButtonGroup';
 import ToggleButton from '@mui/material/ToggleButton';
 import { withRouter } from 'react-router-dom';
-import handleExportJson from '../private/components/workspaces/workspaceExportHandler';
 import themeLight from './ThemeLight';
 import themeDark from './ThemeDark';
 import { commitLocalUpdate } from '../relay/environment';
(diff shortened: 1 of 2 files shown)


opencti-platform/opencti-front/src/components/ExportButtons.jsx
```diff
@@ -165,13 +164,12 @@ class ExportButtons extends Component {
...
       csvFileName,
-      handleDownloadAsStixReport,
       containerId,
+      handleDownloadAsStixReport,
+      handleExportDashboard,
       investigationAddFromContainer,
       history,
-      workspace,
     } = this.props;
-    const isCustomDashBoard = workspace.type === 'dashboard';
```
SarahBocognano, line 174: Maybe the previous way of checking with the location path was better? Less code and less risk of having bugs

SouadHadjiat: the new check is now with `type === 'dashboard'`, we already have a similar check with investigations, and I think it's better to check with the parameters we provide.

labo-flg: My two cents on this :
This component is meant to be used in different places, so in my opinion we should keep a "generic" approach.
Having one prop like `type` that conditions what is the behavior seems a good fit.

the only thing arguable would be what to give to this component in terms of handlers: should it encapsulate all possible handlers and just run them according to type, or give all handlers as props.

I really advise **not** to condition the behavior based on location, generally speaking.
Routing is a specific matter that should live on its own. We might want to move things around, change location names, etc. Let's keep components out of this business or we're prone to side-effects (do not consider location as fixed forever).

labo-flg (approved):

PR #4898, opened 2023-11-14
[backend] validate saved filters on create/edit (#4536)

labo-flg:
We need to validate each and every filter saved into the platform  with respect to isStixMatchFilterGroup capabilities:
* no multi filter keys
* only a subset of filter keys compatible

### Proposed changes

* improve the validator function in stix-filtering.ts so the error gives enough details (what key is bad, what's available)
* call this validator on edit and create resolvers for: Stream, Feed, Taxii, PlaybookNode, notif Trigger

### Related issues

* #4536
* #2686 

### Further comments

Only for internal review in the "filter team".


opencti-platform/opencti-graphql/src/domain/feed.ts
```diff
@@ -36,6 +37,11 @@ const checkFeedIntegrity = (input: FeedAddInput) => {
       }
     });
   }
+
+  // our stix matching is currently limited, we need to validate the input filters
+  if (input.filters && input.filters !== '{}') {
```
Archidoit, line 42: can input.filters be equal to {} ?  I think it's undefined or of type FilterGroup (mode is compulsory etc)

labo-flg: it can be apparently. I tested it, but don't remember the actual test case.

Well in fact this check is not useful, the validation would throw an error properly anyway..

labo-flg:
Reviewed lived and merged.

Issue #4693, opened 2023-10-20
Table text does not change color after switching from dark theme to light theme


rattat0r:
## Description

The text in the report content edit table does not change color after changing from a dark theme to a light theme. Because of this, the text is not visible.

## Environment

1. OS: Ubuntu 20.04.6
2. OpenCTI version: OpenCTI 5.10.1
3. OpenCTI client: frontend

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Go to Report->Knowledge->Content mapping view
2. Switch to Edition mode
3. Add a table and enter text into it
4. Change the theme to light
5. The text remains white, so it is not visible

## Expected Output

After changing the theme color, the text will change to the opposite color.

## Actual Output

After changing the theme color, the text remains the same color, so it is not visible

## Screenshots
Before
![image](https://github.com/OpenCTI-Platform/opencti/assets/122795003/fce9f38a-d3fc-4f53-bfac-9b92419cf326)
After
![image](https://github.com/OpenCTI-Platform/opencti/assets/122795003/c9f75bea-258d-461d-83ff-a2ae331e660a)

lndrtrbn:
Until this is fix, you can manually change the color of the text using the tool in the screenshot below
![image](https://github.com/OpenCTI-Platform/opencti/assets/33682259/59994965-4868-443e-a1fa-0774aa5e1a1d)

rattat0r:
@lndrtrbn I discovered that this issue occurs when using the "Inter" font. When using the font "Arial" everything works fine.

lndrtrbn:
After some tests, I don't think it's font family related. I could reproduce the issue no matter the font I chose.

However, it seems that the issue is that rich text editor theme is not reloaded correctly when changing the theme in settings.
If I reproduce the steps of the issue, and refresh the page then I have correct font colors and text in the table is so visible.

PR #4906, opened 2023-11-15
[frontend] Keep html body attribute 'data-theme' in sync when changing theme (#4693)

lndrtrbn:
### Proposed changes

* Update the attribute 'data-theme' of the body when we change mode (light, dark)
* Move `Index.jsx` in TypeScript

> How this fix the initial issue? CKEditor uses this body attribute to determine which style to apply.

### Related issues

* #4693 

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/private/Index.tsx
+++ b/opencti-platform/opencti-front/src/private/Index.tsx
@@ -4,6 +4,7 @@ import makeStyles from '@mui/styles/makeStyles';
 import Box from '@mui/material/Box';
 import CssBaseline from '@mui/material/CssBaseline';
 import { useTheme } from '@mui/styles';
+import { BoundaryRoute, NoMatch } from '@components/Error';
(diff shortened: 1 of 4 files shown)


opencti-platform/opencti-front/src/private/Index.tsx
```diff
@@ -52,10 +58,29 @@ const Index = ({ settings }) => {
...
+    const body = document.querySelector('body');
+    if (body) {
+      const bodyMode = body.getAttribute('data-theme');
+      const themeMode = `${theme.palette.mode}`;
+      if (bodyMode !== themeMode) {
+        body.setAttribute('data-theme', themeMode);
+      }
+    }
+  }, [theme]);
+
```
lndrtrbn, line 75: The code about fixing the issue.
The rest is to be able to use TypeScript.

jpkha (approved):

Issue #3181, opened 2023-04-17
Extend authorized_members ACL system to Feedbacks


SamuelHassine:
## Use case

Be able activate/deactivate Access restriction on instance of Feedback cases and defined specific access rights for users, groups, organizations.
Be able to define default access restrictions and specific access rights in the Settings of the Feedback Entity type

SamuelHassine:
Be careful: think about https://github.com/OpenCTI-Platform/opencti/issues/3720.

What we do with cases? Not really possible to delete them.

richard-julien:
@SamuelHassine @Jipegien we needs to rediscuss about this one. Looks like its not possible to handle the situation like this because it will force the assignee to be part of authorized members and so take over the instance.

SamuelHassine:
@richard-julien @Jipegien @Kedae: when starting working on this, please don't forget NYC request: "be able to put a 'default authorized member'" by entity type in the default values of the type in settings".

Jipegien:
@SamuelHassine Questions on this:
- Do they want the possibility to add several members as default ?
- Do they want the possibility to define rights when selecting these users?

Jipegien:
Yes and Yes.
Users, Groups, Organizations.
Possibility to define that the creator of the entity is authorized_member by default
Priority on Case

Jipegien:
rescoped for an implementation to only Feedback in 5.12. Will be implemented to Case after that, then refactored to include "Assignee/participant" definition

PR #4918, opened 2023-11-16
[frontend/backend] Authorized members for Feedback (#3181)

lndrtrbn:
### Proposed changes

* [front] Refactoring of form to manage authorized members in workspace to be reusable
* [back] Add new attribute `authorized_members` in Feedback entity
* [back] Add new capability to allow managing authorized members on an entity
* [back] Add migration for the new capability
* [back] Enrich the Feedback API to be able to fetch and edit authorized members
* [front] In Settings > Customization > Feedback, manage the case to edit default authorized members
* [front] In Feedback details, manage new access and add a dialog to manage authorized members
* [back] Update tests
* [back] Refactoring of `attributeDefinition` to not use the value of `mandatoryType` to determine if the attribute can have a default value or not (see [this commit](https://github.com/OpenCTI-Platform/opencti/pull/4918/commits/280897fb491d0f085efd537eec71214da7baab48))

### Related issues

* #3181

### Checklist

### Further comments

You can find manual testing process in Notion page _Extend autorized_member system to Feedback_ in block _Tests cases_.

**Why refactoring `attributeDefinition` ?**
Until now, we determined if an attribute should be returned to the frontend so we can edit its defaults value using the property `mandatoryType` and it worked well because we didn't have attribute which can be editable and not mandatory. So the condition to return the attribute was "is the attribute mandatory or not ?".
But in our case, authorized members can have a default value and is never mandatory so the condition was broken. Instead of just add a tricky condition based on the name of the attribute we decided to revamp, adding a new property `editDefault` that tell if the attribute can be edited or not.


opencti-platform/opencti-graphql/src/utils/authorizedMembers.ts
```diff
@@ -42,23 +43,18 @@ export const getAuthorizedMembers = async (
...
+  requiredCapabilities: string[] = []
 ) => {
   const adminIds = authorized_members
     .filter((n) => n.access_right === MEMBER_ACCESS_RIGHT_ADMIN)
     .map((e) => e.id);
   if (adminIds.length === 0) { // no admin
     return false;
   }
-  if (adminIds.includes(MEMBER_ACCESS_ALL)) { // everyone is admin
+  if (adminIds.includes(MEMBER_ACCESS_ALL) || adminIds.includes(MEMBER_ACCESS_CREATOR)) { // everyone  or creator is admin
```
SouadHadjiat, line 61: what is MEMBER_ACCESS_CREATOR exactly ? you can have this 'CREATOR' in the list of authorized members, who does it grant access to ?

lndrtrbn: `CREATOR` is a key used as ID in default values for `authorized_members`. The goal is to define, in default values, a specific role to apply to the user that will create feedbacks. At feedback creation, this `CREATOR` fake ID is replaced by the ID of the real user making the request of creation

lndrtrbn:
Rebase ongoing

Kedae:
Need to fix the missing creator as admin when getting into manage access
Need to find a way of displaying the EE agreement when clicking on the attributes line in settings

SarahBocognano (commented):
When I add a group (Feedback Manager), it displays when adding, but if I close the form and open it again, this happen : 
Capture d'écran 2023-11-29 100645

Missing Trad : 
Capture d'écran 2023-11-29 100851

- If I give manage access to the group "Feedback Manager" in Feedback entity settings, my  user (Jean) who is in Feedback Manager, doesn't have access to the manage button in his own comment. But if I modify the access directly in the feedback of Jean and give him access, it's working.
- If Jean (who has manage access) give manage access permission to Anne (in Anne's feedback), when I go to the feedback with Anne profile, the manage access permission isn't here, I can only see it, not edit or manage. --> Maybe we can find a way to inform the user that if Anne only has default group access (in our case read knowledge capa only) then the manage access won't work. It can be confusing to not know why it's not working

lndrtrbn:
@Kedae 
Both points have been done.

@SarahBocognano 
1. Fixed,
2. Fixed,
3. Yep that is not a bug, default values are applied at entity creation (for any attributes) so if you modify the default value it has no impact on already existing entities,
4. We didn't find an easy solution to do this, that's why we explain it inside the info message below the switch input in the authorized members dialog form. But the info message changes depending on if the switch is ON or OFF, maybe that's not a good idea

opencti-platform/opencti-graphql/src/migrations/1699521799493-capability-authorized-members.js
```diff
@@ -0,0 +1,16 @@
+import { executionContext, SYSTEM_USER } from '../utils/access';
+import { addCapability } from '../domain/grant';
+
+export const up = async (next) => {
```
SouadHadjiat: you need to update the timestamp of this migration file, otherwise it won't run when merging. The most recent migration is `1700492747686-update-manager-configuration.js`

lndrtrbn: Oh, yes, I'll change that when the PR is approved so I'll be sure there is not an other added between 😅

SarahBocognano:
Tested again, good for me with the explications of the switch button. I think we can do better in terms of user experience (like maybe only display user/groups/organisations who have manage access in their capabilities in the access restriction management) but for now, the message is clear. 
We can add that information to the documentation for now to avoid confusion/questions.

✅ Good to go

opencti-platform/opencti-graphql/src/database/generate-message.js
```diff
@@ -56,12 +58,26 @@ export const generateDeleteMessage = (instance) => {
...
 
-export const generateUpdateMessage = (entityType, inputs) => {
+export const generateUpdateMessage = async (context, entityType, inputs) => {
   const inputsByOperations = R.groupBy((m) => m.operation ?? UPDATE_OPERATION_REPLACE, inputs);
   const patchElements = Object.entries(inputsByOperations);
   if (patchElements.length === 0) {
     throw UnsupportedError('[OPENCTI] Error generating update message with empty inputs');
   }
+
+  const authorizedMembersIds = patchElements.slice(0, 3).flatMap(([,operations]) => {
```
SouadHadjiat, line 68: this whole block looks a bit too complicated for just history logging, do we really need this ?

lndrtrbn: It's to be able to resolve name of users affected by the update of authorized members otherwise we have a dirty history message with unreadable ids. But I can add some comment to explain what is going on

opencti-platform/opencti-graphql/src/database/generate-message.js
```diff
@@ -56,12 +58,26 @@ export const generateDeleteMessage = (instance) => {
...
+    return operations.slice(0, 3).flatMap(({ key, value }) => {
+      return key === 'authorized_members' ? (value ?? []).map(({ id }) => id) : [];
+    });
+  });
+  let members = [];
+  if (authorizedMembersIds.length > 0) {
+    members = await internalFindByIds(context, SYSTEM_USER, authorizedMembersIds, {
+      baseData: true,
+      baseFields: ['internal_id', 'name']
```
SouadHadjiat, line 77: nice thing to return only the fields we need 👍

I guess patchElements can have mutliple operations with authorized_members ?

opencti-platform/opencti-graphql/src/manager/historyManager.ts
```diff
@@ -117,6 +118,7 @@ const eventsApplyHandler = async (context: AuthContext, events: Array<SseEvent<S
       applicant_id: event.data.origin?.applicant_id,
       timestamp: eventDate,
       context_data: contextData,
+      authorized_members: stix.extensions[STIX_EXT_OCTI].authorized_members,
```
SouadHadjiat, line 121: why do we want to save authorized members in history index ?

opencti-platform/opencti-graphql/src/database/generate-message.js
lndrtrbn: I'm not sure to understand the questions, you can change/add multiple authorized members at once yes

opencti-platform/opencti-graphql/src/manager/historyManager.ts
lndrtrbn: To keep same behavior as for marking and organization sharing systems

opencti-platform/opencti-graphql/src/database/generate-message.js
SouadHadjiat: the question was why do this search for members by id in this block and not in below : 
```
 else if (key === 'authorized_members') {
          message = value.map(({ id, access_right }) => {
            const member = members.find(({ internal_id }) => internal_id === id);
            return `${member?.name ?? id} (${access_right})`;
          }).join(', ');
        }
```

lndrtrbn: Ah ! At first we wanted to put it here directly. But this block is inside 2 `{array}.map(() => ...)` and so making an async call inside would have been difficult

SouadHadjiat: ah yes ok, that is why I was wondering if we really need to log the authorized members name in history.

SarahBocognano (approved):

Issue #4735, opened 2023-10-24
Relationship popup window closes by itself when reverse the direction


Lhorus6:
## Description

When changing the direction of a relationship for an observable, the popup window closes by itself.

## Environment

5.11.8

## Reproducible Steps

1. Go to the "Knowledge" tab of an observable file
2. Click on the "+" to add a relationship
3. Search for malware to create a relationship with it
4. Click on "Reverse" and you'll see the window close by itself (see screenshot) and no relation created.

## Screenshots

<img width="1699" alt="Relation_creation_bug" src="https://github.com/OpenCTI-Platform/opencti/assets/70956706/931694d5-5fe8-4383-8800-b429f7dd2ed8">

SarahBocognano:
Just a quick question: Did you try to do it with other entities?
I've been able to reproduce the bug, but only with malware

----> Nevermind, I tested with all entities, and the same behavior is happening on other entities. Thanks for your contribution, I'm on it !

SarahBocognano:
After discussion with @Lhorus6 , we've notice that the pop up close when the reserved relationship isn't possible. We're thinking about disabling the reversed button if the relationship is not reversible. What do you think @Jipegien @SamuelHassine ?

Jipegien:
From STIX 2.1 Doc:"STIX also allows relationships from any SDO or SCO to any SDO or SCO that have not been defined in this specification. These relationships MAY use the related-to relationship type or MAY use a user-defined relationship type." 

So, if there IS no SRO available between 2 SDO or SRO in OpenCTI. It is a problème. At least "related-to" SRO should be available. Except if the contexte is a specific view in Knowledge that limits choices. In this case, disable the button

PR #4954, opened 2023-11-22
[frontend] Relationship popup window closes by itself when reverse the direction (#4735)

SarahBocognano:
### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/4735


lndrtrbn (approved):
I was a bit confused about not resetting the form when clicking cancel but it seems to work and not keeping the old state when opening again so 🤷‍♂️

Issue #4965, opened 2023-11-23
In Knowledge views, clicking on the "Entities view" button or the "Relationships view" when active make the UI disappeared


Jipegien:
## Description

In Knowledge views, clicking on the "Entities view" button or the "Relationships view" when active make the UI disappeared
![image](https://github.com/OpenCTI-Platform/opencti/assets/104078945/71a268a4-f527-4f56-847d-7cea7fa02bf7)

Then

![image](https://github.com/OpenCTI-Platform/opencti/assets/104078945/593dceb4-7a65-4037-b85e-50bf3f1d6854)

PR #4971, opened 2023-11-23
[frontend] Fix changing view in Knowledge (#4965)

lndrtrbn:
### Proposed changes

* A check was made before to ensure the value of the new view is valid, but in a recent commit this check was removed, add it again.

### Related issues

* #4965 

### Checklist


lndrtrbn:
Something is bothering me, even if this fix prevent from facing this issue again, if you already faced it it won't fix it as the issue corrupts the local storage with a null value. And so, even with this fix, your local storage still has a null value that leads you into a blank page.

You can have a look at `EntityStixCoreRelationshipsIndicators.tsx` to see why.

Issue #3720, opened 2023-07-11
Delete cascade user => triggers / notifications / workspaces


SamuelHassine:
## Use case

When deleting a user we should also:

* Delete all workspaces where the user is the only one with admin rights
* Delete all notifications of the user
* Delete all the triggers & digests of the user

Archidoit:
We should also delete the user background tasks, the background tasks the user is the only admin of.

PR #4976, opened 2023-11-24
[backend/frontend] delete investigation, dashboard and notification of a user when user is deleted (#3720)

aHenryJard:
> Ready for review, there is one feature that is missing and will be in a follow up PR:
>   * Delete Workspace user's Investigation and Dashboard shared to other, but where the user is the only 'admin'.

### Proposed changes

* Improve the delete popup message to explain the consequences of a user deletion.
* Delete user's Notification, Digests and Triggers (both are Triggers)
* Delete user's Investigation and Dashboard (both are Workspace) that are not shared to another 'admin'.
* User workspace where the user is 'admin' and having other users is deleted.
* Only one audit log is create for the user deletion. No audit log for Notification, Triggers, Workspace deletion.

### Related issues

* #3720

### Screenshots

popup_delete

### Checklist

### Further comments

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/settings/users/UserPopover.jsx
+++ b/opencti-platform/opencti-front/src/private/components/settings/users/UserPopover.jsx
@@ -146,7 +146,7 @@ class UserPopover extends Component {
                 <li>{t('All notifications, triggers and digests associated with the user will be deleted.')}</li>
                 <li>{t('All investigations and dashboard where the user is the only admin, will be deleted.')}</li>
               </ul>
-                  {t('If you want to save this information, we recommend deactivating the user instead.')}
+                  {t('If you want to keep the associated information, we recommend deactivating the user instead.')}
           </DialogContent>
           <DialogActions>
             <Button
--- a/opencti-platform/opencti-front/src/private/components/settings/users/UserPopover.jsx
+++ b/opencti-platform/opencti-front/src/private/components/settings/users/UserPopover.jsx
@@ -9,10 +9,10 @@ import IconButton from '@mui/material/IconButton';
 import Dialog from '@mui/material/Dialog';
 import DialogActions from '@mui/material/DialogActions';
 import DialogContent from '@mui/material/DialogContent';
-import DialogContentText from '@mui/material/DialogContentText';
 import Slide from '@mui/material/Slide';
 import MoreVert from '@mui/icons-material/MoreVert';
 import { withRouter } from 'react-router-dom';
+import DialogTitle from '@mui/material/DialogTitle';
 import inject18n from '../../../../components/i18n';
 import { commitMutation, QueryRenderer } from '../../../../relay/environment';
 import UserEdition from './UserEdition';
@@ -140,10 +140,13 @@ class UserPopover extends Component {
           TransitionComponent={Transition}
           onClose={this.handleCloseDelete.bind(this)}
         >
-          <DialogContent>
-            <DialogContentText>
-              {t('Do you want to delete this user?')}
-            </DialogContentText>
+          <DialogTitle>{t('Do you want to delete this user?')}</DialogTitle>
+          <DialogContent dividers>
+              <ul>
+                <li>{t('All notifications, triggers and digests associated with the user will be deleted.')}</li>
+                <li>{t('All investigations and dashboard where the user is the only admin, will be deleted.')}</li>
+              </ul>
+                  {t('If you want to save this information, we recommend deactivating the user instead.')}
           </DialogContent>
           <DialogActions>
             <Button
--- a/opencti-platform/opencti-graphql/src/domain/user.js
+++ b/opencti-platform/opencti-graphql/src/domain/user.js
@@ -751,9 +751,9 @@ export const deleteAllNotificationByUser = async (userId) => {
  * Delete a user and related data:
  * - Delete relation
  * - Delete user's Notification, Digests and Triggers (both are Triggers)
- * - Delete user's Investigation and Dashboard (both are Workspace) that are not shared to another 'admin' or 'edit'.
+ * - Delete user's Investigation and Dashboard (both are Workspace) that are not shared to another 'admin'.
  *
- * User workspace where the user is 'admin' and having other users with 'view' only are deleted too.
+ * User workspace where the user is 'admin' and having other users are deleted too.
  * Only one audit log is create for the user deletion. No audit log for Notification, Triggers, Workspace deletion.
  * @param context
  * @param user the user that is user to call delete
--- a/opencti-platform/opencti-graphql/src/domain/user.js
+++ b/opencti-platform/opencti-graphql/src/domain/user.js
@@ -14,14 +14,19 @@ import {
   ACCOUNT_STATUSES,
   OPENCTI_SESSION
 } from '../config/conf';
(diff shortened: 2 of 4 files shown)


opencti-platform/opencti-graphql/src/domain/user.js
```diff
@@ -4,21 +4,40 @@ import * as R from 'ramda';
 import { uniq } from 'ramda';
 import { v4 as uuid } from 'uuid';
 import {
-  BUS_TOPICS,
```
aHenryJard: arf my bad, bad configuration of my IDE... I will revert all import changes.

opencti-platform/opencti-graphql/tests/utils/testQuery.ts
```diff
@@ -150,7 +150,7 @@ const TEST_ORGANIZATION: Organization = {
 };
 
 // Users
-interface User {
+export interface User {
```
aHenryJard:
```suggestion
interface User {
```

opencti-platform/opencti-graphql/tests/utils/testQuery.ts
```diff
@@ -365,7 +365,7 @@ const USER_CREATION_MUTATION = `
     }
   }
 `;
-const createUser = async (user: User) => {
+export const createUser = async (user: User) => {
```
aHenryJard:
```suggestion
const createUser = async (user: User) => {
```

opencti-platform/opencti-graphql/src/modules/workspace/workspace-domain.ts
```diff
@@ -46,6 +46,8 @@ export const findAll = (context: AuthContext, user: AuthUser, args: QueryWorkspa
 };
 
 export const editAuthorizedMembers = async (context: AuthContext, user: AuthUser, workspaceId: string, input: MemberAccessInput[]) => {
+  console.log('*** editAuthorizedMembers, input', input);
+
```
aHenryJard:
```suggestion
```

opencti-platform/opencti-graphql/src/modules/workspace/workspace-domain.ts
```diff
@@ -103,6 +105,8 @@ const initializeAuthorizedMembers = (authorizedMembers: InputMaybe<MemberAccessI
   return initializedAuthorizedMembers;
 };
 export const addWorkspace = async (context: AuthContext, user: AuthUser, input: WorkspaceAddInput) => {
+  console.log('######### input:', input);
+
```
aHenryJard:
```suggestion
```

opencti-platform/opencti-graphql/src/domain/user.js
```diff
@@ -681,6 +686,68 @@ export const meEditField = async (context, user, userId, inputs, password = null
   return userEditField(context, user, userId, inputs);
 };
 
+export const deleteAllWorkspaceForUser = async (userId) => {
+  console.log('##### DELETE Workspaces for userId', userId);
+
```
aHenryJard:
```suggestion
```

opencti-platform/opencti-graphql/src/domain/user.js
```diff
@@ -681,6 +686,80 @@ export const meEditField = async (context, user, userId, inputs, password = null
   return userEditField(context, user, userId, inputs);
 };
 
+export const deleteAllWorkspaceForUser = async (userId) => {
+  return await elRawDeleteByQuery({
```
SouadHadjiat, line 690: I'm wondering if having these `elRawDeleteByQuery` in a domain class is thre right place or if it should be kepts in engine.js ? I don't really know, engine is already so big, and I understand the logic to keep specific functions in the dedicated domain class.

aHenryJard: I have no strong opinion on that, we could move it but as you said engine is already so big. And I don't think we want to reuse it anywhere else, it's quite specific.

opencti-platform/opencti-front/src/private/components/settings/users/UserPopover.jsx
```diff
@@ -140,10 +140,13 @@ class UserPopover extends Component {
...
-            <DialogContentText>
-              {t('Do you want to delete this user?')}
-            </DialogContentText>
+          <DialogTitle>{t('Do you want to delete this user?')}</DialogTitle>
+          <DialogContent dividers>
+              <ul>
+                <li>{t('All notifications, triggers and digests associated with the user will be deleted.')}</li>
+                <li>{t('All investigations and dashboard where the user is the only admin, will be deleted.')}</li>
+              </ul>
+                  {t('If you want to save this information, we recommend deactivating the user instead.')}
```
lndrtrbn:
```suggestion
                  {t('If you want to keep the associated information, we recommend deactivating the user instead.')}
```

Just a detail, but it felt strange to me to use `save` as the data is already saved 😅

opencti-platform/opencti-graphql/src/domain/user.js
lndrtrbn: I had the same interrogation, but I saw that @richard-julien did similar stuff for "works" in `domain/work.js l150` very recently so..

opencti-platform/opencti-graphql/tests/02-integration/01-database/user-delete-cleanup-test.ts
```diff
@@ -0,0 +1,163 @@
...
+  const createdUser = await addUser(adminContext, adminUser, simpleUser);
+
+  // TODO what is the type here ??
+  return createdUser;
+};
+
+/**
+ * Create a AuthUser from user information that have been used to create the user in elastic.
+ * // TODO there is probably a better way to have a AuthUser from a User...
+ * @param user
```
marieflorescontact, line 53: Maybe this isn't relevant, but have you tried using the createUser method in` utils/testQuery.ts`?

marieflorescontact (commented):
locally tested ok 👌

opencti-platform/opencti-front/src/private/components/settings/users/UserPopover.jsx
lndrtrbn: Also change the Localization file 😁

CelineSebe: Fixed !

opencti-platform/opencti-graphql/src/domain/user.js
richard-julien: For me there is no issue to use engine.js in domain files. Should be limited because lot of stuff is handled by middleware.js but when you need to do low level api, it seems ok for me. Engine.js is not the place to have specific function linked to specific entity behavior so i dont think its a good idea to move the function here. I dont really see a good alternative but if you have some idea, feel free to propose :)

marieflorescontact (approved):

opencti-platform/opencti-graphql/tests/02-integration/01-database/user-delete-cleanup-test.ts
aHenryJard: I think thar createUser from testQuery uses graphQL requests , and since tests are domain & engine level only it's an issue because it could lead to unconsistency. That's why we rewrote a createUser in the file. But this domain level one could be in a common file. We can do that in a follow up PR.

PR #4978, opened 2023-11-24
Updated Worker with support for custom auth objects

Obdam:
### Proposed changes

* Added the `auth` headers to the instantiation of the OpenCTIApiClient;
* Created support for custom headers in the worker configuration/environment variables.

### Related issues

* This change is related to the addition of an authentication class to the client-python repository, which can be found here: https://github.com/OpenCTI-Platform/client-python/pull/483

### Checklist

### Further comments
N/A.

### DIFF
--- a/opencti-worker/src/worker.py
+++ b/opencti-worker/src/worker.py
@@ -24,7 +24,9 @@ from opentelemetry.sdk.metrics import MeterProvider
 from opentelemetry.sdk.resources import SERVICE_NAME, Resource
 from pika.adapters.blocking_connection import BlockingChannel
 from prometheus_client import start_http_server
+from pycti import __version__
 from pycti import OpenCTIApiClient
+from pycti import OpenCTIApiAuthentication
 from pycti.connector.opencti_connector_helper import (
     create_mq_ssl_context,
     get_config_variable,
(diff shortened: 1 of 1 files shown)


opencti-worker/src/worker.py
```diff
@@ -439,6 +440,16 @@ def __post_init__(self) -> None:
         self.log_level = get_config_variable(
             "WORKER_LOG_LEVEL", ["worker", "log_level"], config
         )
+        # Custom authentication
+        self.custom_headers = get_config_variable(
+            "CUSTOM_HEADERS", ["opencti", "custom_headers"], config
```
aHenryJard, line 445: Could you give a use case example of custom header ? I don't see "CUSTOM_HEADERS" usage in OpenCTI, so I have difficulties to see how to tests this part.

Obdam: Hi @aHenryJard,

Thanks for your reply. A potential use case is as follows:

- You're hosting the OpenCTI Platform and its Workers on a 'serverless' environment at a cloud vendor, e.g. Google Cloud's Cloud Run or Azure's Container Apps.
- Serverless environments may contain multiple authentication methods/settings, for instance an Identity-Aware Proxy.
- For instance, GCP Cloud Run expects an extra header beside the traditional `Authorization: Bearer ID_TOKEN`, namely the `X-Serverless-Authorization: Bearer ID_TOKEN`. This is further described [here](https://cloud.google.com/run/docs/authenticating/service-to-service#acquire-token).
- Therefore, you may want to influence the way how the Worker sends HTTP request to the Platform.

Obdam:
@labo-flg What's the reason for closure?

richard-julien:
Hi @Obdam , will be done with https://github.com/OpenCTI-Platform/client-python/issues/844.
Changing the concept of a global auth override to specify custom headers on connector and worker.

Issue #4930, opened 2023-11-19
Page "group" not updated when adding users


SamuelHassine:
## Description

Page "group" not updated when adding users

## Reproducible Steps

1. Go to settings / security / groups
2. Create a group
3. Edit the group and go to the members tab
4. Add a few members

=> The members do not appear in the list of members, you should reload the page to make them appear.

CelineSebe:
The bug has been fixed with updaters in GroupEditionUsers

PR #4979, opened 2023-11-24
[frontend] Fix page group updated when adding users

CelineSebe:
### Related issues

#4930

### Checklist

### Further comments

### DIFF
--- a/opencti-platform/opencti-graphql/src/database/file-storage.js
+++ b/opencti-platform/opencti-graphql/src/database/file-storage.js
@@ -10,7 +10,7 @@ import { SemanticAttributes } from '@opentelemetry/semantic-conventions';
 import conf, { booleanConf, ENABLED_FILE_INDEX_MANAGER, logApp } from '../config/conf';
 import { now, sinceNowInMinutes } from '../utils/format';
 import { DatabaseError, FunctionalError } from '../config/errors';
-import { createWork, deleteWorkForFile, loadExportWorksAsProgressFiles } from '../domain/work';
+import { createWork, deleteWorkForFile, deleteWorkForSource, loadExportWorksAsProgressFiles } from '../domain/work';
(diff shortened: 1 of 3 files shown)


opencti-platform/opencti-front/src/private/components/settings/users/GroupUsersLines.tsx
```diff
@@ -94,14 +96,15 @@ const GroupUsersLines: FunctionComponent<GroupUsersLinesProps> = ({
...
+
   const membersData = data.group?.members;
   return (
     <ListLinesContent
       initialLoading={!data}
       isLoading={isLoadingMore}
       loadMore={loadMore}
       hasMore={hasMore}
-      dataList={membersData?.edges ?? []}
+      dataList={groupData.edges ?? []}
```
lndrtrbn: You can't call `groupData.edges` and type `groupData: string[]`, TypeScript won't be happy with that 👿

But more globally, I'm not sure this is the approach we want to fix the issue. By giving members in props you are bypassing Relay refetchable fragment defined in this file. Or if we want to do this that way, you need to remove query that is not used  anymore in this file. @Kedae any sharing on this?

lndrtrbn (changes requested):

opencti-platform/opencti-front/src/private/components/settings/users/GroupUsersLines.tsx
Kedae: Well, it's working but I think we can update the connection from an outside component. I'll check because I think we have done that for tasks. I'll get back to you.

Kedae: As discussed with @CelineSebe > We need to use insertNode and deleteNode but we need to find a way to propagate the paginationOptions

PR #5001, opened 2023-11-28
[frontend/backend] Filters bug fixes 2 (#2686)

Archidoit:


opencti-platform/opencti-front/src/components/TasksFilterValueContainer.tsx
```diff
@@ -21,6 +48,26 @@ const TasksFilterValueContainer = ({ filters }: { filters: FilterGroup }) => {
...
+                <InformationOutline
+                  fontSize="small"
+                  color="secondary"
+                  style={{ cursor: 'default' }}
+                />
+              </Tooltip>
+            </>
+          }
+        />
+      }
```
labo-flg: On a second thought, we might as well remove this as we already convert it on the fly to new format. This information has no effect on the user, he does not care and won't do anything about it.

what do you think @Archidoit ?

Archidoit: It was to warn the user this filters won't be accessible anymore in 1 or 2 versions

Archidoit: But indeed we can remove this chip, the 2 solutions are ok

labo-flg: ok, I do it

labo-flg (approved):

PR #5003, opened 2023-11-28
[backend] entity_type filter with different operators and modes fix

Archidoit:
ewemple of a query that is not working well: entity_type = Malware AND Software


Archidoit:
@labo-flg maybe add the 2 following test cases:
- Report OR Container
- Report AND Container

labo-flg:
> @labo-flg maybe add the 2 following test cases:
> 
> * Report OR Container
> * Report AND Container

It's equivalent to my tests Malware _and_ SDO / Malware _or_ SDO (SDO is the immediate parent type of Malware, like Container is the immediate parent type of Report)

@labo-flg yes :)

labo-flg (approved):

PR #5006, opened 2023-11-29
[backend/frontend] Improve auditing to split file read and file download (#5005)

### LINKED ISSUE #5005: Improve auditing to split file read and file download
## Use case

Improve auditing to split file read and file download

richard-julien:
See #5005

### DIFF
--- a/opencti-platform/opencti-graphql/src/http/httpPlatform.js
+++ b/opencti-platform/opencti-graphql/src/http/httpPlatform.js
@@ -16,7 +16,7 @@ import contentDisposition from 'content-disposition';
 import { basePath, booleanConf, DEV_MODE, logApp, OPENCTI_SESSION } from '../config/conf';
 import passport, { empty, isStrategyActivated, STRATEGY_CERT } from '../config/providers';
 import { authenticateUser, authenticateUserFromRequest, loginFromProvider, userWithOrigin } from '../domain/user';
-import { downloadFile, getFileContent, loadFile } from '../database/file-storage';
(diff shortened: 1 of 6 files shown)


opencti-platform/opencti-graphql/src/http/httpPlatform.js
```diff
@@ -51,6 +51,19 @@ const extractRefererPathFromReq = (req) => {
...
+  const data = buildContextDataForFile(entity, file.id, filename);
+  await publishUserAction({
+    user: auth,
+    event_type: 'file',
+    event_access: 'extended',
+    event_scope: 'download',
+    context_data: data
+  });
+};
+
```
lndrtrbn, line 66: Can't we just have one function taking the scope `download` or `read` in parameter to avoid duplicates ? In a same way `checkFileAccess` has been made

```suggestion
const publishFile = async (executeContext, auth, file, event_scope) => {
  const { filename, entity_id } = file.metaData;
  const entity = entity_id ? await internalLoadById(executeContext, auth, entity_id) : undefined;
  const data = buildContextDataForFile(entity, file.id, filename);
  await publishUserAction({
    user: auth,
    event_type: 'file',
    event_access: 'extended',
    event_scope,
    context_data: data
  });
};

```

Kedae (approved):

opencti-platform/opencti-graphql/src/http/httpPlatform.js
richard-julien: Thanks for the comment. I will merge as it is and will be part of future refactor.

PR #5040, opened 2023-12-01
[frontend] Updated required fields for TA Individual to include External References

ParamConstructor:
Yup() basicShape configuration modified for Threat Actor Individual to allow Customized Configured Fields on Threat Actor Individual to denote "*" stars on the required fields. External References added as a field than can be enabled as required.

Customized required fields is an existing capability per object type off the Admin console.

### Proposed changes

* "*" Stars shown next to labels on required fields for Threat Actor Individual form. (Could be extended to all platform forms)
* uses Yup() to track required fields marking

### Related issues

* This is only a modification to the Threat Actor Individual forms - but could be applied sitewide.

### Checklist

### Further comments

N/A

### DIFF
--- a/opencti-platform/opencti-front/src/utils/hooks/useEntitySettings.tsx
+++ b/opencti-platform/opencti-front/src/utils/hooks/useEntitySettings.tsx
@@ -67,6 +67,19 @@ export const useIsEnforceReference = (id: string): boolean => {
   );
 };
 
+export const useIsMandatoryAttribute = (id: string) => {
+  const entitySettings = useEntitySettings(id).at(0);
+  if (!entitySettings) {
+    throw Error(`Invalid type for setting: ${id}`);
+  }
+  const mandatoryAttributes = [...entitySettings.mandatoryAttributes];
+  // In creation, if enforce_reference is activated, externalReferences is required
+  if (entitySettings.enforce_reference === true) {
+    mandatoryAttributes.push('externalReferences');
+  }
+  return { entitySettings, mandatoryAttributes };
+};
+
 export const useYupSchemaBuilder = (
   id: string,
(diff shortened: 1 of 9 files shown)


opencti-platform/opencti-front/src/utils/hooks/useEntitySettings.tsx
```diff
@@ -95,15 +95,24 @@ export const useYupSchemaBuilder = (
...
       .filter((attr) => !(exclusions ?? []).includes(attr))
       .map((attrName: string) => {
+        let validator: Schema;
         if (existingKeys.includes(attrName)) {
-          const validator: Schema = (existingShape[attrName] as Schema)
+          const fieldInfo = Yup.object().shape(existingShape).describe().fields[attrName] as Yup.SchemaLazyDescription;
+          validator = (existingShape[attrName] as Schema)
+            .transform((v) => ((Array.isArray(v) && v.length === 0) ? undefined : v))
+            .required(t('This field is required'))
+            .label(`${fieldInfo?.label}*`)
```
labo-flg: This is a bit hacky IMHO :)

I suggest you rely on the "required" prop in MUI Field instead, to let the UI (potentially themed and customized) decide how to display required fields markers.

We could have a simple utility function that tells the value of such prop taking the modified yup schema in parameter. What do you think ?

richard-julien: Agree with @labo-flg , i would also prefer to split the concern and not modify all the yup form. We need a utility function that take the entity type and the attribute to compute the require and use it in the material component.

ParamConstructor:
Closing and replacing with different branch that has required fields update across whole platform, versus just in TA area.

PR #5044, opened 2023-12-02
[backend/frontend] Improve id conversion between platforms (#3242)

richard-julien:
See #3242
Export > transform internal to standard
Import > transform standard to internal


opencti-platform/opencti-front/src/private/components/workspaces/WorkspaceCreation.jsx
```diff
@@ -74,6 +74,7 @@ const WorkspaceCreation = ({ paginationOptions, type }) => {
     commitImportMutation({
       variables: { file: importedFile },
       onCompleted: (data) => {
+        inputRef.current.value = ''; // Reset the input uploader ref
```
Kedae: You should put that in an updater in case of an error.
Plus I would set to null instead of '', at the input is a file input (not really important)

opencti-platform/opencti-front/src/private/components/workspaces/dashboards/WidgetConfig.jsx
```diff
@@ -423,6 +423,9 @@ const WidgetConfig = ({ workspace, widget, onComplete, closeMenu }) => {
           dashboardManifest,
         },
       },
+      onCompleted: () => {
+        inputRef.current.value = ''; // Reset the input uploader ref
```
Kedae: Same

Kedae (approved):

PR #5080, opened 2023-12-04
[Frontend] Some console errors fixes on settings page

lndrtrbn:
### Proposed changes

* Diverse small fixes to clear a bit errors in dev console

### Related issues

* No related issues, just QoL for developers

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/components/ColorPickerField.jsx
+++ b/opencti-platform/opencti-front/src/components/ColorPickerField.jsx
@@ -62,13 +62,15 @@ const ColorPickerField = (props) => {
     }
   };
 
+  const { value, ...otherProps } = fieldToTextField(props);
+
   return (
     <>
       <MuiTextField
-        {...fieldToTextField(props)}
+        {...otherProps}
+        value={value ?? ''}
         error={!isNil(meta.error)}
         helperText={!isNil(meta.error) ? meta.error : props.helperText}
-        ref={anchorEl}
         onChange={internalOnChange}
         onFocus={internalOnFocus}
         onBlur={internalOnBlur}
--- a/opencti-platform/opencti-front/src/components/MarkdownField.jsx
+++ b/opencti-platform/opencti-front/src/components/MarkdownField.jsx
@@ -63,7 +63,7 @@ const MarkdownField = (props) => {
         {label}
       </InputLabel>
       <ReactMde
-        value={field.value}
+        value={field.value ?? ''}
         readOnly={disabled}
         onChange={(value) => setFieldValue(name, value)}
         selectedTab={controlledSelectedTab || selectedTab}
--- a/opencti-platform/opencti-front/src/components/SelectField.jsx
+++ b/opencti-platform/opencti-front/src/components/SelectField.jsx
@@ -5,6 +5,7 @@ import MuiSelect from '@mui/material/Select';
 import InputLabel from '@mui/material/InputLabel';
 import FormControl from '@mui/material/FormControl';
 import FormHelperText from '@mui/material/FormHelperText';
+import { fieldToTextField } from 'formik-mui';
 
 const fieldToSelect = ({
   disabled,
@@ -18,7 +19,6 @@ const fieldToSelect = ({
   return {
     disabled: disabled ?? isSubmitting,
     error: showError,
-    formError: showError ? fieldError : undefined,
     onBlur: () => {},
     onChange: fieldOnChange ?? (() => {}),
     onClose: onClose ?? (async (e) => {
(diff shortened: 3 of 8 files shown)


opencti-platform/opencti-front/src/components/ColorPickerField.jsx
```diff
@@ -62,13 +62,15 @@ const ColorPickerField = (props) => {
...
   };
 
+  const { value, ...otherProps } = fieldToTextField(props);
+
   return (
     <>
       <MuiTextField
-        {...fieldToTextField(props)}
+        {...otherProps}
+        value={value ?? ''}
```
lndrtrbn, line 71: Reason: `value` can be null. But an input should not be initialized with null or undefined value, instead use an empty string. By so, React can properly track changes with the input.

opencti-platform/opencti-front/src/components/MarkdownField.jsx
```diff
@@ -63,7 +63,7 @@ const MarkdownField = (props) => {
         {label}
       </InputLabel>
       <ReactMde
-        value={field.value}
+        value={field.value ?? ''}
```
lndrtrbn, line 66: Reason: `value` can be null. But an input should not be initialized with null or undefined value, instead use an empty string. By so, React can properly track changes with the input.

opencti-platform/opencti-front/src/components/SelectField.jsx
```diff
@@ -79,7 +81,8 @@ const SelectField = (props) => {
         {props.label}
       </InputLabel>
       <MuiSelect
-        {...fieldToSelect(props)}
+        {...otherProps}
+        value={value ?? ''}
```
lndrtrbn, line 84: Reason: `value` can be null. But an input should not be initialized with null or undefined value, instead use an empty string. By so, React can properly track changes with the input.

opencti-platform/opencti-front/src/components/TextField.jsx
```diff
@@ -40,9 +40,12 @@ const TextField = (props) => {
     [onSubmit, setTouched, name],
   );
   const [, meta] = useField(name);
+  const { value, ...otherProps } = fieldToTextField(props);
+
   return (
     <MuiTextField
-      {...fieldToTextField(props)}
+      {...otherProps}
+      value={value ?? ''}
```
lndrtrbn, line 48: Reason: `value` can be null. But an input should not be initialized with null or undefined value, instead use an empty string. By so, React can properly track changes with the input.

opencti-platform/opencti-front/src/private/components/settings/Settings.jsx
```diff
@@ -484,7 +484,7 @@ const Settings = () => {
                                 <ItemBoolean
                                   variant="inList"
                                   neutralLabel={
-                                    settings.platform_cluster.instances_number
+                                    `${settings.platform_cluster.instances_number}`
```
lndrtrbn, line 487: Reason: `neutralLabel` accepts a string and not a number

opencti-platform/opencti-front/src/private/components/settings/roles/CapabilitiesList.tsx
```diff
@@ -59,7 +59,7 @@ const CapabilitiesList: FunctionComponent<CapabilitiesListProps> = ({
             );
           }
         }
-        return <div key="none" />;
+        return <div key={i} />;
```
lndrtrbn, line 62: Reason: always set a unique key to elements inside a loop. By so, React can correctly track each individual element.

opencti-platform/opencti-front/src/private/components/settings/sub_types/SubTypesLine.tsx
```diff
@@ -197,39 +197,34 @@ export const SubTypeLineDummy = ({
...
+                  animation="wave"
+                  variant="rectangular"
+                  width="90%"
+                  height={20}
+                />
+              </div>
+            ))}
+          </div>
+        }
+      />
```
lndrtrbn, line 227: Just removed the imbricated `ListItem` as it was not required.

Reason: `li` inside `li` was generated.

opencti-platform/opencti-front/src/private/components/settings/roles/CapabilitiesList.tsx
labo-flg: could you find something based on the data ? edge.node.id maybe ?

Using index is a last-resort strategy that can lead to issues (see https://robinpokorny.com/blog/index-as-a-key-is-an-anti-pattern/)

lndrtrbn: Well, this return is called when the data inside the map is undefined, so I can't really use any stuff linked to the data

labo-flg: ok then

labo-flg (approved):

Issue #5093, opened 2023-12-05
Incorrect filterGroup when editing Cases


labo-flg:
## Description

An error occurs when you edit a Case (any case), when leaving the drawer after edition.

## Environment

octi 5.12.3

## Reproducible Steps

1. open a RFI
2. hit '+' on origin of the case
3. leave the drawer (do not even edit the info in the form)
4. an error occurs

PR #5094, opened 2023-12-05
[frontend] fix filtergroup not set correctly for Cases (#5093)

labo-flg:
### Proposed changes

The filters stored in the component state are incorrectly reset when leaving the edition form.

### Related issues

* #5093

### Checklist

### Further comments

Quick refacto of the filter utils on this occasion, for a clearer naming.

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/common/containers/ContainerAddStixCoreObjects.jsx
+++ b/opencti-platform/opencti-front/src/private/components/common/containers/ContainerAddStixCoreObjects.jsx
@@ -22,7 +22,7 @@ import {
   constructHandleAddFilter,
   constructHandleRemoveFilter,
   filtersAfterSwitchLocalMode,
-  initialFilterGroup,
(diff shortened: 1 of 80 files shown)


opencti-platform/opencti-front/src/private/components/common/containers/ContainerAddStixCoreObjects.jsx
```diff
@@ -464,13 +467,8 @@ const ContainerAddStixCoreObjects = (props) => {
...
                 || targetStixCoreObjectTypes.includes('Stix-Cyber-Observable')
             )
-        ? {
-          key: 'entity_type',
-          values: targetStixCoreObjectTypes,
-          operator: 'eq',
-          mode: 'or',
-        }
-        : initialFilterGroup,
+        ? targetEntityTypesFilterGroup
```
labo-flg, line 470: bug was here.

Issue #5035, opened 2023-12-01
Migrate status filter to use underlying template to simplify filter usage


Archidoit:
## Description

Status display several times by entity in status filter

## Reproducible Steps
- Edit a trigger filters
- Click on the status filter
- Statuses are displayed by entity type, and so displayed several times sometimes
![image](https://github.com/OpenCTI-Platform/opencti/assets/75783086/4e892c31-9eea-4eb5-a280-7c9899fb4855)

Jipegien:
Expected output : one single list with all status possible.

PR #5098, opened 2023-12-05
[frontend/backend] Migrate status filter to use underlying template to simplify filter usage (#5035)

Archidoit:
Github issue: https://github.com/OpenCTI-Platform/opencti/issues/5035


opencti-platform/opencti-graphql/src/database/engine.js
```diff
@@ -1893,6 +1897,44 @@ const adaptFilterToSourceReliabilityFilterKey = async (context, user, filter) =>
...
+    const newFilter = {
+      ...filter,
+      key: ['x_opencti_workflow_id'], // we just have to change the key
+    };
+    return { newFilter, newFilterGroup: undefined };
+  }
+  // case 2: operator = eq or no_eq
+  if (operator === 'eq' || operator === 'not_eq') {
+    // worflow_id filter values can be both status ids and templates ids of a status
+    // get all the statuses
```
Goumies: The code is self-explanatory.
Can we remove those comments?

opencti-platform/opencti-graphql/src/database/engine.js
```diff
@@ -1893,6 +1897,44 @@ const adaptFilterToSourceReliabilityFilterKey = async (context, user, filter) =>
...
+    let statuses = await getEntitiesListFromCache(context, user, ENTITY_TYPE_STATUS);
+    // keep the statuses with their id corresponding to the filter values, or with their template id corresponding to the filter values
+    statuses = statuses.filter((status) => values.includes(status.id) || values.includes(status.template_id));
+    // we construct a new filter that matches against the status internal_id with a template id in the filters values
```
Goumies, line 1926: Same here

opencti-platform/opencti-graphql/src/database/engine.js
```diff
@@ -1893,6 +1897,44 @@ const adaptFilterToSourceReliabilityFilterKey = async (context, user, filter) =>
...
+      key: ['x_opencti_workflow_id'],
+      values: statusIds,
+      mode,
+      operator,
+    };
+
+    return { newFilter, newFilterGroup: undefined };
+  }
+  throw Error(`The operators supported for a filter with key=workflow_id are: eq, not_eq, nil, not_nil: ${operator} is not supported.`);
+};
```
Goumies, line 1938: Can we rewrite this to an early exit function: throwing the error as soon as possible.
For example:
`if(!['eq', 'not_eq', 'nil', 'not_nil'].includes(operator)){ ... }`

opencti-platform/opencti-graphql/src/migrations/1701859469408-rename-workflow-filter-key.js
```diff
@@ -0,0 +1,321 @@
...
+    const { filters, filterGroups } = parsedFilters;
+    const newFiltersContent = [];
+    const newFilterGroups = [];
+    filters.forEach((filter) => {
+      const { key } = filter;
+      const arrayKeys = Array.isArray(key) ? key : [key];
+      if (arrayKeys.includes('x_opencti_workflow_id')) {
+        const newKeys = arrayKeys.filter((k) => k !== 'x_opencti_workflow_id');
+        newKeys.push('workflow_id');
+        newFiltersContent.push({ ...filter, key: uniq(newKeys) });
```
Goumies, line 34: I have noticed the usage of this Ramda function.
Since we are trying to remove Ramda, I suggest using native Javascript, `Set`.
For example:
```
const set = new Set(newKeys)
const key = Array.from(set)
```

opencti-platform/opencti-graphql/src/migrations/1701859469408-rename-workflow-filter-key.js
```diff
@@ -0,0 +1,321 @@
...
+    [ENTITY_TYPE_PLAYBOOK],
+  );
+
+  let playbooksDefinitionConvertor = {};
+  playbooks
+    .forEach((playbook) => {
+      const playbookDefinition = JSON.parse(playbook.playbook_definition);
+      const definitionNodes = playbookDefinition.nodes;
+      const newDefinitionNodes = [];
+      for (let i = 0; i < definitionNodes.length; i += 1) {
```
Goumies: A forEach on definitionNodes was not suggested by ESlint ?
I have encountered it in one of my previous PR's checks.
The second argument is the index.

opencti-platform/opencti-graphql/src/utils/filtering/filtering-resolution.ts
```diff
@@ -72,6 +83,19 @@ const resolveFilter = (filter: Filter, resolutionMap: FilterResolutionMap): Filt
...
+  // 2. handle the special case of workflow filter
+  if (key.includes(WORKFLOW_FILTER)) {
+    // get all the statuses
+    let statuses = await getEntitiesListFromCache(context, user, ENTITY_TYPE_STATUS) as BasicWorkflowStatus[];
+    // keep the statuses with their id corresponding to the filter values, or with their template id corresponding to the filter values
+    statuses = statuses.filter((status) => values.includes(status.id) || values.includes(status.template_id));
+    const statusIds = statuses.length > 0 ? statuses.map((status) => status.internal_id) : ['<no-status-matching-filter>'];
+    // replace filter values with the statusIds
+    // !!! it works to do the mode/operator filter on the status (and not on the template)
+    // because a status can only have a single template and because the operators are full-match operators (eq/not_eq) !!!
```
Goumies, line 95: Besides those last comments that explicit the why, we could remove the others.

opencti-platform/opencti-graphql/src/database/engine.js
Archidoit: fixed

opencti-platform/opencti-graphql/src/migrations/1701859469408-rename-workflow-filter-key.js
Archidoit: no it was not suggested... I fix this :)

opencti-platform/opencti-graphql/src/database/engine.js
Archidoit: the function should return something, so I put an error at the end, to handle both the fact that return is compulsory and the mentionned operators are not handled

Archidoit: and because there is only 2 comparisons done before the 'throw', it's not compulsorty to put the error as soon as possible

opencti-platform/opencti-graphql/src/migrations/1701859469408-rename-workflow-filter-key.js
Archidoit: when it complicates the code too much, we still use Ramda I think (it's the case for uniq and dissoc for instance)

opencti-platform/opencti-graphql/src/database/engine.js
```diff
@@ -1958,6 +1992,16 @@ const completeSpecialFilterKeys = async (context, user, inputFilters) => {
...
+    } else if (arrayKeys.includes(WORKFLOW_FILTER) || arrayKeys.includes(X_OPENCTI_WORKFLOW_ID)) {
+      // in case we want to filter by status template (template of a workflow status) or status
+      // we need to find all statuses filtered by status template and filter on these statuses
+      const { newFilter, newFilterGroup } = await adaptFilterToWorkflowFilterKey(context, user, filter);
+      if (newFilter) {
+        finalFilters.push(newFilter);
+      }
+      if (newFilterGroup) {
+        finalFilterGroups.push(newFilterGroup);
+      }
```
labo-flg: this would never happen, right ? We  can just return a newFilter and use it

Archidoit: fixed

labo-flg:
all tested ok, LGTM

labo-flg (approved):

PR #5131, opened 2023-12-06
[backend/frontend] Refactor files management to rely on internal engine instead of S3 (#4583)

### LINKED ISSUE #4583: Full refactor of files storage and fetching to avoid querying S3 when not necessary
## Use case

Today, a lot of components in the OpenCTI frontend (UI, Python library) are requesting the files of an entity each time we read it. This leads a lot of organizations to have extra-cost on their S3 buckets and this has a lot of limitation:

* No real sorting / pagination
* No search on file
* Performance and latency
* etc.

To be clear: all features implemented in the ElasticSearch engine are not available when it comes to handle files (including mass deletion, etc.). This is a high priority for a lot of OpenCTI users and customers.

Solution discussed with @Kedae and @richard-julien: implement files representation in ElasticSearch, maintain synchronization between S3 and Elastic and fetch S3 content only when needed (file read or download).

richard-julien:
See #4583


opencti-platform/opencti-graphql/src/migrations/1701354091174-files-registration.js
```diff
@@ -0,0 +1,23 @@
...
+import { rawFilesListing } from '../database/file-storage';
+import { logApp } from '../config/conf';
+import { buildFileDataForIndexing } from '../modules/internal/document/document-domain';
+import { elIndexElements } from '../database/engine';
+import { INDEX_INTERNAL_OBJECTS } from '../database/utils';
+
+export const up = async (next) => {
+  const context = executionContext('migration');
+  logApp.info('[MIGRATION] Starting 1701354091161-files-registration.js');
+  const files = await rawFilesListing(context, SYSTEM_USER, '/', { recursive: true });
```
SouadHadjiat, line 11: won't rawFileListing take too much time ?

richard-julien: Thats the initial loading for migration, do you have an alternative in mind?

SouadHadjiat: not yet

opencti-platform/opencti-graphql/src/migrations/1701354091174-files-registration.js
```diff
@@ -0,0 +1,23 @@
...
+  logApp.info(`[MIGRATION] ${files.length} files to register in index`);
+  const elements = files.map((file) => {
+    return { _index: INDEX_INTERNAL_OBJECTS, ...buildFileDataForIndexing(file) };
+  });
+  await elIndexElements(context, SYSTEM_USER, 'Migration files registration', elements);
```
SouadHadjiat, line 16: elIndexElements seem to index in bulk, but without any chunk, it looks like it will index everything in one bulk request

opencti-platform/opencti-graphql/src/modules/internal/document/document-domain.ts
```diff
@@ -0,0 +1,165 @@
...
+  const lastIndexedFiles = await elSearchFiles(context, SYSTEM_USER, searchOptions);
+  const lastIndexedFile = lastIndexedFiles?.length > 0 ? lastIndexedFiles[0] : null;
+  return lastIndexedFile ? moment(lastIndexedFile.uploaded_at).toISOString() : FROM_START_STR;
+};
+
+export const buildFileDataForIndexing = (file: any) => {
+  const standardId = generateFileIndexId(file.id);
+  const fileData = R.dissoc('id', file);
+  return {
+    ...fileData,
```
SouadHadjiat, line 44: what will we have exactly in fileData ?

opencti-platform/opencti-graphql/src/migrations/1701354091174-files-registration.js
SouadHadjiat: What do we need more than what we get with simpleFilesListing ? I guess entityId and mimeType, or something else ? We could use simpleFilesListing and load only files for which we don't have the mimeType. entityId can easily be found in the file path.

opencti-platform/opencti-graphql/src/database/file-storage.js
```diff
@@ -366,73 +369,33 @@ export const upload = async (context, user, path, fileUpload, opts) => {
...
-    size: readStream.bytesRead,
+    size: existingFile.size,
     information: '',
     lastModified: new Date(),
     lastModifiedSinceMin: sinceNowInMinutes(new Date()),
     metaData: { ...fullMetadata, messages: [], errors: [] },
     uploadStatus: 'complete'
   };
+  // Register in elastic
+  await indexFileToDocument(file);
```
SouadHadjiat, line 384: we should probably think about elastic mapping for this object (which fields will we need to query, and which can we just store, for example metaData could be a flat object)
and maybe we don't need to store everything, for example "information" is empty

opencti-platform/opencti-graphql/src/migrations/1701354091174-files-registration.js
richard-julien: The concept is to index all the files contains in minio, not just a subset.
In case of migration i  think its not really a problem if this listing take 10 or more secs

opencti-platform/opencti-graphql/src/migrations/1701354091174-files-registration.js
richard-julien: Yes. One bulk its perfectly fine i think

opencti-platform/opencti-graphql/src/database/file-storage.js
richard-julien: As we do a lot of search/filtering on the inside elements, we unfortunatly need a concrete object
However you right defining mapping can still be a good idea

opencti-platform/opencti-graphql/src/modules/internal/document/document-domain.ts
richard-julien: You right, using any was a fast approach but not the expected one, need to rework that

PR #5132, opened 2023-12-06
[backend] Customize time range of relationship de-deduplication (#3387)

### LINKED ISSUE #3387: Be able to customize the time range of relationship de-deduplication mechanism
## Use case

Be able to customize the time range of relationship de-deduplication mechanism.
Be able to add (as an option) the author as a deduplication key.

Currently, we deduplicate relationships on:

- source ID
- target ID
-  -1 month < start_time  < +1 month
- -1 month < stop_time  < +1 month

We need to propose the ability to change the resolved range for de-deplucation.

- [ ] Add static configuration in platform config
- [ ] Adapt engine algorithm to take into account the configuration

This will enable a powerful capability for the platform to be more "precise" as a threat intelligence team but also to de-duplicate everything based on 12 months range for instance for less mature / less resourced team that just need the "higher level" insights.

richard-julien:
See #3387

### DIFF
--- a/opencti-platform/opencti-graphql/config/default.json
+++ b/opencti-platform/opencti-graphql/config/default.json
@@ -91,6 +91,11 @@
   "no_proxy": "",
   "https_proxy_ca": [],
   "https_proxy_reject_unauthorized": false,
+  "relations_deduplication": {
+    "past_days": 30,
+    "next_days": 30,
+    "created_by_based": false
+  },
   "data_sharing": {
     "max_csv_feed_result": 5000
   },
--- a/opencti-platform/opencti-graphql/src/database/middleware.js
+++ b/opencti-platform/opencti-graphql/src/database/middleware.js
@@ -2522,7 +2522,7 @@ const buildRelationDeduplicationFilter = (input) => {
     created_by_based: false,
     types_overrides: {}
   };
-  const config = deduplicationConfig.types_overrides[relationshipType] ?? deduplicationConfig;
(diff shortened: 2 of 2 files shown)


opencti-platform/opencti-graphql/config/default.json
```diff
@@ -91,6 +91,11 @@
   "no_proxy": "",
   "https_proxy_ca": [],
   "https_proxy_reject_unauthorized": false,
+  "relations_deduplication": {
+    "past_days": 30,
+    "next_days": 30,
+    "created_by_based": false
```
aHenryJard, line 97: question: what does `created_by_based` mean ?

aHenryJard (commented):
Looks good to my newbie eyes, but CI is failing, and I wonder if it's possible to have a test coverage for this feature or if the test already exists.

opencti-platform/opencti-graphql/config/default.json
richard-julien: created_by_based is used to add the createdBy information as part of the deduplication query.

JRI creates a relationship TARGETS, with FROM_DATE1 and FROM_DATE2 will not be dedupe if SAM creates relationship TARGETS, with FROM_DATE1 and FROM_DATE2, just creating a new one

Will take a look to the tests

richard-julien: Tests are now running fine

aHenryJard (approved):

PR #5139, opened 2023-12-07
[backend] Fix search in files query (#5129)

### LINKED ISSUE #5129: Double quote are not taken into account in the file text indexing search
## Description

When searching for a string with double quote, the search in texts from indexed files is done without the double quote

## Environment

2. OpenCTI version: 5.12.4 (testing)

## Reproducible Steps

- On testing platform, search "XL4 macros" (with double quotes)
- click one of the result in the file tab
- search the string "XL4 macros" directly in the pdf
- no result (the matching string is "macros")

## Expected Output

No results in the file tab, as no files contains the complete string "XL4 macros"

## Actual Output

2 results

SouadHadjiat:
### Proposed changes

* use query_string which handles natively double quote queries

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/5129

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/database/file-search.js
+++ b/opencti-platform/opencti-graphql/src/database/file-search.js
@@ -132,6 +132,15 @@ const buildFilesSearchResult = (data, first, searchAfter, connectionFormat = tru
   }
   return convertedHits;
 };
+const decodeSearch = (search) => {
+  let decodedSearch;
+  try {
+    decodedSearch = decodeURIComponent(search).trim();
+  } catch (e) {
+    decodedSearch = search.trim();
+  }
+  return decodedSearch;
+};
 const elBuildSearchFilesQueryBody = async (context, user, options = {}) => {
   const { search = null, fileIds = [], entityIds = [] } = options; // search options
   const { includeAuthorities = false } = options;
@@ -139,9 +148,10 @@ const elBuildSearchFilesQueryBody = async (context, user, options = {}) => {
   const must = [...dataRestrictions.must];
   const mustNot = [...dataRestrictions.must_not];
   if (search) {
+    const decodedSearch = decodeSearch(search);
     const fullTextSearch = {
-      multi_match: {
-        query: search,
+      query_string: {
+        query: decodedSearch,
         fields: ['attachment.content', 'attachment.title^2']
       }
     };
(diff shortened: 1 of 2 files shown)


opencti-platform/opencti-graphql/src/database/file-search.js
```diff
@@ -132,16 +132,26 @@ const buildFilesSearchResult = (data, first, searchAfter, connectionFormat = tru
...
+  return decodedSearch;
+};
 const elBuildSearchFilesQueryBody = async (context, user, options = {}) => {
   const { search = null, fileIds = [], entityIds = [] } = options; // search options
   const { includeAuthorities = false } = options;
   const dataRestrictions = await buildDataRestrictions(context, user, { includeAuthorities });
   const must = [...dataRestrictions.must];
   const mustNot = [...dataRestrictions.must_not];
   if (search) {
+    const decodedSearch = decodeSearch(search);
```
richard-julien, line 151: I think specialElasticCharsEscape should be used

labo-flg (approved):

labo-flg:
From an elastic newby perspective it seems ok to me.
Which is not much I reckon.

opencti-platform/opencti-graphql/src/database/file-search.js
SouadHadjiat: why do we need `specialElasticCharsEscape` ?

richard-julien: To prevent query to crash because user is not aware of elasticsearch query syntax

SouadHadjiat: Then we can use `simple_query_string` which will work with special characters even with syntax errors : 
https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-simple-query-string-query.html
https://opensearch.org/docs/latest/query-dsl/full-text/simple-query-string/

richard-julien: Should done the trick. Dont hesitate to try some weird search phrase to see the behavior

SouadHadjiat: yes I tried already and there is no error

Issue #4942, opened 2023-11-21
Interface bugs on data import page


gusandrews:
## Description

The UI on the data import page is looking messy. Text is crashing into other elements in the main and right columns. Also the import.csv function on the right (see attached image) has a little hyphen next to it that looks like it should be usable interface like a dropdown caret, but it isn't. 

## Environment

1. Mac 13.6.1; Chrome browser
2. OpenCTI version: { e.g. OpenCTI 1.0.2 }
3. OpenCTI client: Frontend
4. Other environment details:

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Run OpenCTI with the above specifications
2. Access the data import page

## Expected Output

Either there should be no marking next to import.csv, or if there is a dropdown there, there should be an actionable indicator (caret).

## Screenshots (optional)
<img width="1470" alt="Screenshot 2023-11-17 at 10 25 56 AM" src="https://github.com/OpenCTI-Platform/opencti/assets/1565981/5c0c3f08-b39d-47b9-adc8-46185a3b796b">

Jipegien:
@RomuDeuxfois & @labo-flg : seems lacking a date for ImportCSV. Is it also possible to limit the display of file_name  in uploaded files and the display of mime-types in the list of connectors, or do we need to create another issue for that?

RomuDeuxfois:
State management should be added at the importCsv connector level by setting the last csv import date. This will replace caret.
What do you think @Jipegien ?

Jipegien:
The comportment needs to be the same as for other connectors.

RomuDeuxfois:
We can set the state connector on polling time

lndrtrbn:
The issue on uploaded files is not specific to CSV page. There is the same issue on other pages listing long filenames.
CSS stuff, I can have a look if you want ? For the CSV date stuff, I leave that to you.

PR #5144, opened 2023-12-07
[frontend] Fix interface bugs on data import page

Goumies:
### Proposed changes

* truncate text where needed
* hide updated_at date when missing
* add navigation to the connector page on click on the list button with connector name

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/4942

### Checklist

### Further comments

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/common/files/FileLine.tsx
+++ b/opencti-platform/opencti-front/src/private/components/common/files/FileLine.tsx
@@ -44,8 +44,8 @@ import { FileLine_file$data } from './__generated__/FileLine_file.graphql';
 import { isNotEmptyField } from '../../../../utils/utils';
 import { truncate } from '../../../../utils/String';
 
-const Transition = React.forwardRef((props: SlideProps, ref) => (
-  <Slide direction="up" ref={ref} {...props} children={props.children} />
+const Transition = React.forwardRef(({ children, ...otherProps }: SlideProps, ref) => (
+  <Slide direction='up' ref={ref} {...otherProps}>{children}</Slide>
 ));
 Transition.displayName = 'TransitionSlide';
 
--- a/opencti-platform/opencti-front/src/private/components/common/files/FileLine.tsx
+++ b/opencti-platform/opencti-front/src/private/components/common/files/FileLine.tsx
@@ -31,7 +31,6 @@ import { RecordSourceSelectorProxy } from 'relay-runtime';
 import { PopoverProps } from '@mui/material/Popover';
 import MenuItem from '@mui/material/MenuItem';
 import Menu from '@mui/material/Menu';
-import ListItemSecondaryAction from '@mui/material/ListItemSecondaryAction';
 import FileWork from './FileWork';
 import { useFormatter } from '../../../../components/i18n';
 import {
@@ -285,7 +284,7 @@ const FileLineComponent: FunctionComponent<FileLineComponentProps> = ({
             }
           />
         </Tooltip>
-        <ListItemSecondaryAction>
+        <div onClick={(e) => e.stopPropagation()} style={{ flex: 'none' }}>
           {!disableImport && (
             <Tooltip title={t('Launch an import of this file')}>
               <span>
@@ -383,7 +382,7 @@ const FileLineComponent: FunctionComponent<FileLineComponentProps> = ({
               )}
             </>
           )}
-        </ListItemSecondaryAction>
+        </div>
       </ListItem>
       <FileWork file={file} nested={workNested} />
       <Dialog
--- a/opencti-platform/opencti-front/src/private/components/common/files/FileLine.tsx
+++ b/opencti-platform/opencti-front/src/private/components/common/files/FileLine.tsx
@@ -18,7 +18,6 @@ import {
 import Tooltip from '@mui/material/Tooltip';
 import ListItemIcon from '@mui/material/ListItemIcon';
 import ListItemText from '@mui/material/ListItemText';
-import ListItem from '@mui/material/ListItem';
 import CircularProgress from '@mui/material/CircularProgress';
 import Dialog from '@mui/material/Dialog';
 import DialogContent from '@mui/material/DialogContent';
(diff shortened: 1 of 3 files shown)


opencti-platform/opencti-front/src/private/components/import/ImportContent.jsx
```diff
@@ -336,6 +341,8 @@ class ImportContentComponent extends Component {
                 <FileUploader
                   onUploadSuccess={() => relay.refetch()}
                   size="medium"
+                 color={'primary'}
```
labo-flg:
```suggestion
                  color={'primary'}
```

opencti-platform/opencti-front/src/private/components/import/ImportContent.jsx
```diff
@@ -411,12 +418,12 @@ class ImportContentComponent extends Component {
                       </Tooltip>
                       <ListItemText
                         primary={connector.name}
-                        secondary={R.join(',', connector.connector_scope)}
+                        secondary={truncate(connector.connector_scope.join(','), 30)}
```
labo-flg: it might be useful to have a tooltip around this truncated content, to see the full content on mouse hover.

opencti-platform/opencti-front/src/private/components/import/ImportContent.jsx
```diff
@@ -411,12 +418,12 @@ class ImportContentComponent extends Component {
...
                       />
-                      <ListItemSecondaryAction>
-                        <ListItemText primary={nsdt(connector.updated_at)} />
-                      </ListItemSecondaryAction>
-                    </ListItem>
+                        {connector.updated_at && (<ListItemSecondaryAction>
+                            <ListItemText primary={nsdt(connector.updated_at)}/>
+                        </ListItemSecondaryAction>)}
+                    </ListItemButton>
```
labo-flg:
```suggestion
                      {connector.updated_at && (<ListItemSecondaryAction>
                        <ListItemText primary={nsdt(connector.updated_at)}/>
                      </ListItemSecondaryAction>)}
                    </ListItemButton>
```

Goumies: I don't understand those indentation issues. Especially since I ran `yarn lint --fix`

opencti-platform/opencti-front/src/private/components/common/files/FileLine.tsx
```diff
@@ -285,7 +284,7 @@ const FileLineComponent: FunctionComponent<FileLineComponentProps> = ({
             }
           />
         </Tooltip>
-        <ListItemSecondaryAction>
+        <div onClick={(e) => e.stopPropagation()} style={{ flex: 'none' }}>
```
Kedae: Not sure to understand this change ?

opencti-platform/opencti-front/src/private/components/import/Import.jsx
```diff
@@ -1,27 +1,32 @@
...
+      query={importContentQuery}
+      variables={{}}
+      render={({ props }) => {
+        if (props) {
+          return (
+            <ImportContent
+              connectorsImport={props.connectorsForImport}
+              importFiles={props.importFiles}
+              pendingFiles={props.pendingFiles}
+              navigate={navigate}
```
Kedae: Navigate shouldn't be a props of that component

opencti-platform/opencti-front/src/private/components/import/Import.jsx
```diff
@@ -1,27 +1,32 @@
...
-            importFiles={props.importFiles}
-            pendingFiles={props.pendingFiles}
-          />
-        );
-      }
-      return <Loader />;
-    }}
-  />
-);
+const Import = () => {
```
Kedae: This component should be untouched

Kedae:
image
Still a bit weird

opencti-platform/opencti-front/src/private/components/common/files/FileLine.tsx
Goumies: Neither am I.
The onClick handler might not be necessary. @lndrtrbn ?
I have noticed an ellipse in the style.

opencti-platform/opencti-front/src/private/components/import/Import.jsx
Goumies: Why, @Kedae?
Since ImportContent is a class component, I had to pass it down.

opencti-platform/opencti-front/src/private/components/import/Import.jsx
Goumies: Why?

opencti-platform/opencti-front/src/private/components/common/files/FileLine.tsx
lndrtrbn: `ListItemSecondaryAction` => MUI component to add one and only one icon button at the end of the row. So not our case, that's why text is overlapping the buttons.

Instead just use a div, the `stopPropagation` is needed to avoid calling the onClick callback when clicking on a line (which redirect to a new page)

Kedae: Seem with @Goumies we will rollback the change and update the text to truncate it before it reaches the buttons

@lndrtrbn I completely align with your vision, we are incorrectly using MUI components for list/lines & co, but this is a larger concern that we will need to tackle everywhere.
In fact it's mostly that we should not have been using ListItem components

lndrtrbn: That will be a mess to compute where you need to truncate the text as you don't really know how much buttons you have and what is the width of the row.

To achieve this you need to compute row width, ListItemSecondaryAction width, icon width, set the text width to (row - icon - ListItemSecondaryAction) and truncate using CSS

Goumies: My idea is to truncate text with a drastic limit which will be enough for the smallest available width.
A tooltip will reveal the entire text.

opencti-platform/opencti-front/src/private/components/import/Import.jsx
Goumies: As seen together, I will use `withRouter`

opencti-platform/opencti-front/src/private/components/import/Import.jsx
Goumies: This way, this component does not bear a useless concern

opencti-platform/opencti-front/src/private/components/import/Import.jsx
```diff
@@ -15,6 +15,7 @@ const Import = () => (
             connectorsImport={props.connectorsForImport}
             importFiles={props.importFiles}
             pendingFiles={props.pendingFiles}
+            navigate={props.navigate}
```
Kedae: To revert

labo-flg (approved):

Issue #4932, opened 2023-11-20
Add an optional default value for an attribute that is missing mapping value from a file


Jipegien:
## Use case

In CSV mapper, be able to define default value in representation to be used in case of missing value from the file.
Need to take into account possible default value set in Setteings/Customization
Need to take into account format.

lndrtrbn:
Not done yet, still need chunk 3 (manage markings)

PR #5152, opened 2023-12-08
[Frontend/Backend] Quick fixes for existing CSV Mapper (#4932)

CelineSebe:
### Proposed changes

* ~~Fix refresh validity status of required inputs when changing entity type~~ => Need to rework the code of the form
* Look into why there is two fields abstract in type Note
* Do not send fields authorized_members, casetemplate, authorized_authorities, grantable_groups

### Related issues

* #4932 

### Checklist

### Further comments


lndrtrbn:
wip

lndrtrbn:
About the following issue (first bullet in Proposed Changes): when changing the entity type inside a representation bloc, the errors were not reset and kept the state of the previous entity.

We needed to rework a bit how errors were managed inside the form. Initially they were managed using `useState` directly in components and went through components to components using props. But it wasn't working well, so we change this to use the error system given by Formik. There is still some parts of the code that are a bit strange but we had to go for a solution in a given amount of time.

A next step would be to rework the form to create custom components for Representation blocs using the Formik API to fully integrate Formik errors system easily. Out of scope for our case right now.

labo-flg (changes requested):
We noticed some performance issues with the changes on form validation.
Once you add a dozen representation, try to update the name or add a new representation : this can take several seconds to update.

labo-flg (approved):

PR #5154, opened 2023-12-08
[backend] authorize api filters keys for retention policies checking (#5153)

### LINKED ISSUE #5153: Can't add api filters to retention policies
If we create a retention policy with an api filter, nothing happens after clicking on the 'verify' button --> the 'create' button is not accesible and we can't create the retention policy.

image

Archidoit:
review by the Filters team

### DIFF
--- a/opencti-platform/opencti-graphql/src/utils/filtering/filtering-utils.ts
+++ b/opencti-platform/opencti-graphql/src/utils/filtering/filtering-utils.ts
@@ -222,16 +222,16 @@ const getConvertedRelationsNames = (relationNames: string[]) => {
  * - convert relation refs key if any
  * @param filterGroup
  */
-export const checkAndConvertFilters = (filterGroup?: FilterGroup | null) => {
+export const checkAndConvertFilters = (filterGroup?: FilterGroup | null, noFiltersChecking = false) => {
   if (!filterGroup) {
     return undefined;
   }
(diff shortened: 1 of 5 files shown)


opencti-platform/opencti-graphql/src/utils/filtering/filtering-utils.ts
```diff
@@ -222,16 +222,16 @@ const getConvertedRelationsNames = (relationNames: string[]) => {
  * - convert relation refs key if any
  * @param filterGroup
  */
-export const checkAndConvertFilters = (filterGroup?: FilterGroup | null) => {
+export const checkAndConvertFilters = (filterGroup?: FilterGroup | null, noFiltersChecking = false, authorizeNestedFiltersKeys = false) => {
```
labo-flg: Please prefer using an `opts` parameter with different options inside, so we don't expand the arguments lists infinitely :)

Archidoit: yes !!

Archidoit: fixed

labo-flg (approved):
Tested on dashboards, relationship filter by marking, triggers, and of course retention policies. It seems ok to me.

PR #5170, opened 2023-12-11
[frontend] handle widgets options in filters to enable all modes/operators

### LINKED ISSUE #5168: [Dashboards] Error when creating a Heat Map
## Environment

4. Other environment details:

## Reproducible Steps

1. Create a Heatmap > Entities
2. Select Entities perspective
3. Add the following filter (Entity type = Malware)
4. Create

## Actual Output

image

Archidoit:
only Filters team


opencti-platform/opencti-front/src/private/components/common/audits/AuditsMultiLineChart.jsx
```diff
@@ -99,31 +98,13 @@ const AuditsMultiLineChart = ({
       );
     }
     const timeSeriesParameters = dataSelection.map((selection) => {
-      let types = ['History', 'Activity'];
```
jpkha: Not sure of the need. Can we delete it?

PR #5180, opened 2023-12-12
[frontend] switch translation key when relationship (#5142)

### LINKED ISSUE #5142: Missing translations in investigation screen in expand dialog
## Description

Some translations are missing in investigation screen. 

They are prefixed `entity_` but should be prefixed `relationship_` I think (see Localization.js).

## Environment

OCTI 5.12.3

## Reproducible Steps
 start an investigation, add an entity, double click on the entity in the graph to open the expansion dialog.

## Screenshots (optional)

image

lndrtrbn:
### Proposed changes

* Condition to display entity or relationship translation

> The way of checking it (hard coded array of relation names) is not very elegant but I don't see how to do it otherwise. Making a query just to determine if the name is from an entity or a relationship seemed not a good solution neither, maybe I'm wrong?

### Related issues

* #5142 

### Checklist


opencti-platform/opencti-front/src/private/components/workspaces/investigations/InvestigationExpandForm.tsx
```diff
@@ -348,9 +348,12 @@ const InvestigationExpandFormContent = ({
       .filter(({ value }) => value > 0)
       .sort((a, b) => a.label.localeCompare(b.label));
 
+    // Uses to determine which key of translation to use below.
+    const relNames = ['Part-Of', 'Located-At', 'Indicates', 'Mitigates', 'Targets', 'Uses', 'Related-To', 'Attributed-To'];
```
lndrtrbn: Open to suggestions if you have better idea to check in mind

SouadHadjiat: I'm not sure to understand how we can see relationships in the targets list. 
The dialog displays targets on the left and relationships on the right, how can we have relationships on the left part ? @lndrtrbn

lndrtrbn: Because we can have relations targeting other relations

SouadHadjiat: is this list complete ? I can see we have a lot more in STIX_CORE_RELATIONSHIPS

labo-flg: could you get it from the schema ? 

```ts
const { schema } = useAuth();
// ...
const relationshipsNames = schema.sros.map(({id, label }) => label)
```
Something like that I believe

labo-flg: see how it's done is `useSearchEntities`, we use the schema to get generate all possible options in the filter autocomplete

lndrtrbn: Ok I'll have a look thanks @labo-flg

lndrtrbn: Indeed, seems way better now

labo-flg (approved):

PR #5183, opened 2023-12-12
[frontend] fix options filters in widgets (#5181)

### LINKED ISSUE #5181: Some filters don't support all modes and operators in widgets
## Description
In widgets, relationship_type, elementId and entity_type filters are always done with operator=eq, mode (inside the filter)=or, mode (with other filters)=and.

## Reproducible Steps
An example :

Create a widget with filters: relationship_type = uses OR related-to
The displayed graph is correct.
Same with filters: relationship_type = uses AND related-to
-> we have the same graph (the graph is done with operator 'OR')

Archidoit:
Only internal reviews by the Filters team

issue: https://github.com/OpenCTI-Platform/opencti/issues/5181

### DIFF
--- a/opencti-platform/opencti-front/src/utils/filters/filtersUtils.tsx
+++ b/opencti-platform/opencti-front/src/utils/filters/filtersUtils.tsx
@@ -264,7 +264,7 @@ export const findFilterIndexFromKey = (
 };
 
 // add entity types context in filters
-export const filtersWithEntityType = (
+export const injectEntityTypeFilterInFilterGroup = (
   filters: FilterGroup | undefined,
   type: string | string[],
 ): FilterGroup => {
@@ -286,11 +286,11 @@ export const filtersWithEntityType = (
 
 // remove filter with key=entity_type and values contains 'all'
 // because in this case we want everything, so no need for filters
-export const removeAllTypesFromFilter = (inputFilters?: FilterGroup) => {
+export const removeEntityTypeAllFromFilterGroup = (inputFilters?: FilterGroup) => {
   if (inputFilters && isFilterGroupNotEmpty(inputFilters)) {
     const { filters, filterGroups } = inputFilters;
     const newFilters = filters.filter((f) => !(f.key === 'entity_type' && f.values.includes('all')));
-    const newFilterGroups = filterGroups.map((group) => removeAllTypesFromFilter(group)) as FilterGroup[];
+    const newFilterGroups = filterGroups.map((group) => removeEntityTypeAllFromFilterGroup(group)) as FilterGroup[];
     return {
       mode: inputFilters.mode,
       filters: newFilters,
@@ -301,7 +301,7 @@ export const removeAllTypesFromFilter = (inputFilters?: FilterGroup) => {
 };
 
 // construct filters and options for widgets
-export const constructFiltersAndOptions = (
+export const buildFiltersAndOptionsForWidgets = (
   inputFilters: FilterGroup | undefined,
   opts: { removeTypeAll?: boolean, startDate?: string, endDate?: string, dateAttribute?: string } = {},
 ) => {
@@ -330,7 +330,7 @@ export const constructFiltersAndOptions = (
   );
   let filters = inputFilters ? { ...inputFilters, filters: filtersContent } : undefined;
   // 02. remove 'all' in filter with key=entity_type
-  if (removeTypeAll) filters = removeAllTypesFromFilter(filters);
+  if (removeTypeAll) filters = removeEntityTypeAllFromFilterGroup(filters);
   // 03. handle startDate and endDate options
   const dateFiltersContent = [];
   if (startDate) {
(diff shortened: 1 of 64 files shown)


opencti-platform/opencti-front/src/utils/filters/filtersUtils.tsx
```diff
@@ -282,6 +284,99 @@ export const filtersWithEntityType = (
...
 };
 
+// remove filter with key=entity_type and values contains 'all'
+// because in this case we want everything, so no need for filters
+export const removeAllTypesFromFilter = (inputFilters?: FilterGroup) => {
+  if (inputFilters && isFilterGroupNotEmpty(inputFilters)) {
+    const { filters, filterGroups } = inputFilters;
+    const newFilters: Filter[] = [];
+    const newFilterGroups: FilterGroup[] = [];
+    filters.forEach((filter) => {
```
jpkha, line 294: I think, you can use filter instead 
```
filters.filter((filter) =>!(filter.key === 'entity_type' && filter.values.includes('all')))
```

opencti-platform/opencti-front/src/utils/filters/filtersUtils.tsx
```diff
@@ -282,6 +284,99 @@ export const filtersWithEntityType = (
...
+      if (!(filter.key === 'entity_type' && filter.values.includes('all'))) {
+        newFilters.push(filter);
+      }
+    });
```
jpkha: I think, you can use filter instead 
```
filters.filter((filter) =>!(filter.key === 'entity_type' && filter.values.includes('all')))
```

opencti-platform/opencti-front/src/utils/filters/filtersUtils.tsx
```diff
@@ -282,6 +284,99 @@ export const filtersWithEntityType = (
...
+    filterGroups.forEach((group) => {
+      const newGroup = removeAllTypesFromFilter(group);
+      if (newGroup) filterGroups.push(newGroup);
+    });
```
jpkha: I think, you can use map instead and a filter to remove all `undefined ` if needed but I don't see a case where `removeAllTypesFromFilter ` can return undefined as we checked before ?
```
filterGroups.map((group) =>  removeAllTypesFromFilter(group))
```

jpkha (commented):
Wow a lot of deleted code 👍

opencti-platform/opencti-front/src/utils/filters/filtersUtils.tsx
Archidoit: fixed

opencti-platform/opencti-front/src/utils/filters/filtersUtils.tsx
Archidoit: fixed

jpkha (approved):

opencti-platform/opencti-front/src/utils/filters/filtersUtils.tsx
```diff
@@ -262,6 +263,7 @@ export const findFilterIndexFromKey = (
   return null;
 };
 
+// add entity types context in filters
 export const filtersWithEntityType = (
```
labo-flg: you should rename this function to a verb for consistency, for instance

```suggestion
export const injectEntityTypeFilterInFilterGroup = (
```

opencti-platform/opencti-front/src/utils/filters/filtersUtils.tsx
```diff
@@ -282,6 +284,90 @@ export const filtersWithEntityType = (
   };
 };
 
+// remove filter with key=entity_type and values contains 'all'
+// because in this case we want everything, so no need for filters
+export const removeAllTypesFromFilter = (inputFilters?: FilterGroup) => {
```
labo-flg: I think function name is a bit misleading, we naively understand that it removes all filters on entity_types, not the 'all' special case.

```suggestion
export const removeEntityTypesAllFromFilterGroup = (inputFilters?: FilterGroup) => {
```

opencti-platform/opencti-front/src/utils/filters/filtersUtils.tsx
```diff
@@ -282,6 +284,90 @@ export const filtersWithEntityType = (
...
+      mode: inputFilters.mode,
+      filters: newFilters,
+      filterGroups: newFilterGroups,
+    };
+  }
+  return inputFilters;
+};
+
+// construct filters and options for widgets
+export const constructFiltersAndOptions = (
```
labo-flg:
```suggestion
export const buildFiltersAndOptionsForWidgets = (
```

Archidoit: done

opencti-platform/opencti-front/src/utils/filters/filtersUtils.tsx
Archidoit: done

opencti-platform/opencti-front/src/utils/filters/filtersUtils.tsx
Archidoit: done

labo-flg (approved):

PR #5188, opened 2023-12-12
[backend/frontend] File indexing fixes and improvements (#5187)

### LINKED ISSUE #5187: File indexing fixes and improvements
## Description

- Fix number of files by file type when file type is included in supported file types (ex: application/vnd.ms-excel.addin.macroenabled.12, see screenshot)
- Remove `~` since number of files is not approximate anymore
- In global search, if there is no keyword, we should not count file search results (badge)
- Add missing translations

image

SouadHadjiat:
### Proposed changes

* Fix number of files by file type when file type is included in supported file types (ex: application/vnd.ms-excel.addin.macroenabled.12, see screenshot)
* Remove `~` since number of files is not approximate anymore
* In global search, if there is no keyword, we should not count file search results (badge)
* Add missing translations

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/5187

### Checklist


marieflorescontact (approved):
locally tested ok 👌

opencti-platform/opencti-graphql/src/domain/file.js
```diff
@@ -46,16 +47,16 @@ export const filesMetrics = async (context, user) => {
...
-      size: R.sum(filesByMimetype[mimeType].map((file) => file.size)),
-    });
-  }
+  supportedMimeTypes.forEach((mimeType) => {
+    const mimeTypeFiles = potentialIndexingFiles.filter((file) => file.metaData.mimetype?.startsWith(mimeType));
+    if (mimeTypeFiles.length > 0) {
+      metricsByMimeType.push({
+        mimeType,
+        count: mimeTypeFiles.length,
+        size: R.sum(mimeTypeFiles.map((file) => file.size)),
```
jpkha, line 56: If you want to remove Ramda.sum you can use reduce instead :
`
mimeTypeFiles.map((file) => file.size).reduce((accumulator, currentValue) => accumulator + currentValue, 0)
`

SouadHadjiat: Oh thanks for the suggestion, I find it however more readable to use R.sum than the reduce. Usually I avoid using Ramda, but there are a few cases where the functions are much easier to read : R.sum and R.groupBy

richard-julien (approved):

Issue #3585, opened 2023-06-28
In filters, be able to used all possible vocabularies / the current user


SamuelHassine:
## Use case

For instance, if there is no cases with a certain severity (like critical), it is not possible to create a trigger for severity critical. We need to change that.

![image](https://github.com/OpenCTI-Platform/opencti/assets/1334279/d0be5194-66bc-46ec-a02a-6a42e96273c0)

PR #5189, opened 2023-12-12
Search all possible values in identity filters (#3585)

labo-flg:
### Proposed changes

* Replace runtime attributes query with vocabulary query so we list all possible values 
    * this concerns the filters `priority`, `severity`, `pattern_type`, `malware_types`, `source_reliability`, `indicator_types`, `incident_type`, `report_types`, `channel_types`, `event_types`, `note_types`, `context`
* Add  `me` to the queries for `objectAssignee`, `objectParticipant`, `creatorId`, so you can always see yourself in filters 

~If user is admin, list all possible members when filtering on `objectAssignee`, `objectParticipant`, `creatorId`~ 
--> reverted. This will be rediscussed.

### Related issues

#3585 

### Checklist

### Further comments

I refactored `useSearchEntities` to reduce code size and redundancies.


opencti-platform/opencti-front/src/utils/hooks/useGranted.ts
```diff
@@ -58,4 +58,9 @@ const useGranted = (capabilities: string[], matchAll = false): boolean => {
   return numberOfAvailableCapabilities > 0;
 };
 
+export const useIsAdmin = () => {
+  // only admins have access to the platform settings, so it's a good shortcut for now
+  return useGranted([SETTINGS]);
+};
```
labo-flg: This is actually the recipe used in many places of the platform. I though it would be better to hide this under a dedicated hook.

marieflorescontact: Where is `useIsAdmin` used?

labo-flg: woops ! I used it for something that I removed during the development.

Right now it's not used anywhere.
We can keep it anyway... it's still a good idea to have a dedicated hook I think 🤔

I can do a quick search-replace to refactor this in the platform.

opencti-platform/opencti-front/src/private/components/common/entreprise_edition/EnterpriseEditionButton.tsx
```diff
@@ -33,7 +33,7 @@ const EnterpriseEditionButton = ({
   const {
     settings: { id: settingsId },
   } = useAuth();
-  const isAdmin = useGranted([SETTINGS]);
+  const isAdmin = useIsAdmin();
```
labo-flg: quick refactoring on the fly, 
I did not use this new hook in the final version of my task, but I though it was worth keeping anyway !

opencti-platform/opencti-front/src/utils/hooks/useGranted.ts
labo-flg: I pushed a commit for that; only in the cases where the code was explicitly looking for "isAdmin".

There are other places around using the grant on SETTINGS but semantically it might be better to leave it for code clarity.

opencti-platform/opencti-front/src/utils/hooks/useGranted.ts
```diff
@@ -58,4 +58,9 @@ const useGranted = (capabilities: string[], matchAll = false): boolean => {
   return numberOfAvailableCapabilities > 0;
 };
 
+export const useIsAdmin = () => {
+  // only admins have access to the platform settings, so it's a good shortcut for now
+  return useGranted([SETTINGS]);
```
SouadHadjiat: SETTINGS is not really an admin capability, it only gives access to the settings page, the name of the function "useIsAdmin" might give the wrong idea. Maybe it could be `useGrantedSettings` if we want a shortcut ?

labo-flg: Actually this is what we check to protect Admin-only features in the frontend, in various places, so I'm just adding an abstraction layer around it to make it clearer.

What should be the real way to check if a user is admin in your opinion ?

labo-flg: Maybe BYPASS ?

labo-flg: Well... it's out of scope for this PR I'll just revert it.

SouadHadjiat: sorry I didn't see your replies. "admin" doesn't mean the same thing everywhere. In some pages, "admin" is when you have "SETTINGS" capa, in others it is when you have "SETTINGS_SETACCCESS", and of course there is BYBASS.

SouadHadjiat (commented):
tested locally, looks good to me ✅
I tested most filters, but not all, and I tested also triggers with severity filter and it works well

SouadHadjiat (approved):

PR #5210, opened 2023-12-14
[frontend/backend] CSV Mapper default values (#4932)

lndrtrbn:
### Proposed changes

* [front] Rework CSV Mapper form to use Formik `Field` component,
* [front] Add default values input for fields that can have a custom default value,
* [back] Rework backend modules csvMapper and entitySettings to separate logic,
* [back] Use default values (from entity settings and csv mapper) to fill CSV file data.

In depth overview in the `Further comments` section.

### Related issues

* #4932 

### Checklist

### Further comments

I tried to make commits as understandable as possible to simplify the comprehension of this PR which is a bit huge.
You will find below descriptions of the content of each commit.

**1. [frontend] make defaultValues fields from entitySettings reusable**

> Objective: extract the logic of the component used in entity settings customization to display the correct input depending of the entity attribute. To be able to reuse this same logic inside the CSV mapper form.

The component `DefaultValueField` is simply a copy paste as component of the old function `field` from the component `EntitySettingAttributeEdition`

Other functions removed from `EntitySettingAttributeEdition` have been copy paste into `utils/defaultValues`, also to be able to reuse it inside CSV mapper.

For the hook `useDefaultValues`: I just moved all the logic it does inside a function and return this function. Because doing so I can retrieve this function in a component without calling it and then pass this function to an other one which is not a React component.

**2. [frontend] csv mapper form : rework + defaultValues field**

> Objective: add inputs to be able to set default values for attributes of the CSV mapper. Achieving this by reusing the component `DefaultValueField` extracted from first commit. But to be able to use this component we need to use the `Field` component from Formik. And to be able to do this I needed to rework the entire form.

A lot of stuff is going on here... But the main objective is to avoid as far as possible to manipulate Formik state directly and instead use `Field` component from Formik so it can manage its state itself.

Improvement of `RichTextField` component because before it wasn't possible to update the content programmatically after the first initialization.

In `CsvMappers` component I prepared the relay queries needed later on and use a React context `csvMappers.data` to pass the reference to be able to fetch those data wherever we want inside the provider of this context (@Kedae Relay stuff here). We are doing ultiple things in the file `csvMappers.data`: declare queries to fetch (there is 2 queries instead of one because one is refetchable and not the other one), create the react context containing the data, create a component Provider and a hook to fetch the context to facilitate the usage of this context in other components.

Update of `CsvMapper.d.ts`, `Representation.d.ts` and `Attribute.d.ts` to use generated types from Relay for the data coming from the API and create new types representing the format of the same data but in the csv mapper form. Before the refactoring, the form used the same format of data that the one returned by the API. But it was a bit complex with several levels of depth in properties that was not necessary for the form. So, now there is a separate format used only for the form to simplify the form state manipulation. Functions to move from one format to the other can be found inside `CsvMapperUtils`, `RepresentationUtils` and `AttributeUtils`.

`CsvMapperCreation` and `CsvMapperEdition` now use those functions to format data.

I won't go deep into details of every component of the form, same objective for each, be able to use Formik component API.

**3. [backend] rework on modules csvMapper and entitySettings to separate logic**

> Objective: move entitySettings and csvMapper logic into their own modules. Because before this logic was in Attribute domain and resolvers in a perspective to be generic but we faced specific issues to entitySettings and csvMapper.

So in `resolvers/attribute` and `domain/attribute` a bunch of code removed because now in entitySetting and csvMapper modules.

In `entitySetting-domain` now we have some functions that we removed from `domain/attribute` (copy paste). And so update of imports in `entitySetting-resolvers`.

In `entitySetting-utils` move the function `fillDefaultValues` that was in `middleware.js` before. And also fix an issue in this function as to determine if an attribute is multiple or not we were using only schema definition of attributes, so it returned always false in a case of a ref and not an attribute. Now use both definitions from attributes and refs (@richard-julien if you waant to check that I'm right or not).

`entitySetting-attributeUtils` contains the function that retrieves schemas attributes for an entity setting and the schema data with data from attributes configuration of the entity settings (to take in account overrides made by the admin of the platform). We found a similar function for csv mapper in `csvMapper-domain` but here we are fetching all attributes, not only for one type of entity setting.

Some additions also in `csvMapper-utils`. Add a new function `parseCsvMapperWithDefaultValues` to resolve defaultValues refs which are in the representations of the mapper. And in the `validate` function improve the test of verifying presence of required attributes by also testing refs.

**4. [backend] improve csv mapper + manage defaultValues**

> Objective: populate the data retrieved from the CSV file with default values (either default values from entity settings and ones defined directly in the mapper)

Improvement of the function `columnNameToIdx` by adding a test to verify the string to convert is valid. Add copy paste a function from the file `csv-mapper`.

Main changes are in the file `csv-mapper` with refactoring + populate default values.

**5. [backend] add tests for csv mapper**

> Objective: fix existing tests and add tests for mapper default values.

Removed tests from `csv-parser-test` have been copy paste and improve into `csv-helper-test`.

`csv-mapper-test` contains 2 tests to test default values from entity settings and from csv mapper.

**6. [frontend] fixes from rebase**

Some quick fixes


labo-flg:
After pair review:
* show a feedback on the settings cog icon to know there are some stuff in there that have been set (like an icon badge ?)
* default value for numerical fields: beware of the value 0 that is interpreted as false and is not applied (longitude e.g.)
* lack of tests on relationships

lndrtrbn:
@labo-flg 

> * show a feedback on the settings cog icon to know there are some stuff in there that have been set (like an icon badge ?)
> * default value for numerical fields: beware of the value 0 that is interpreted as false and is not applied (longitude e.g.)
> * lack of tests on relationships

1. Added a text label below the column input if default value is set, as it is done for date pattern and list separator
2. In fact only the case for latitude and longitude, due to converter stix, not sure we want to change this in this PR,
3. Manual tests done + auto tests written

labo-flg (approved):
Tested OK
Code changes are quite big, so I've reviewed what's highlighted in the PR descritpion.

PR #5211, opened 2023-12-14
[frontend/backend] handle entity_type filter in bookmark widgets (#5182)

### LINKED ISSUE #5182: In the widget bookmark, we propose many options to filter but only entity-types are taken into account.
## Description

In the widget bookmark, we propose many options to filter but only entity-types are taken into account.

## Reproducible Steps

1. Create a custom dashboard
2. Create a Bookmark
3. Select entities
4. We can select many filters
5. Only entities are taken into account

## Actual Output

image

Archidoit:
issue: https://github.com/OpenCTI-Platform/opencti/issues/5182


marieflorescontact (commented):
Seems that you fix the bug for `Bookmark edition` but not for `Bookmark creation` 💐

jpkha (approved):

marieflorescontact (approved):
locally tested 👌

PR #5213, opened 2023-12-14
[frontend] Fix error when creating Software from Vulnerability (#5198)

### LINKED ISSUE #5198: Error when creating Software from Vulnerability detail
## Description

When trying to create a new Software to add to the "Affected Sofware" list, the creation window doesn't get closed once the form submitted
Also, after that, the app crashes when we leave the creation screen

## Environment

Testing

## Reproducible Steps

1. Open Vulnerability Detail
2. Add Affected Software
3. Create a new one

## Expected Output

Software created and pop-up closed

## Actual Output

Creation window doesn't get closed and when clicking outside of the window, we get a persistent error in the whole app until we logout.

## Screenshots (optional)
issue_software_vulnerability

SouadHadjiat:
### Proposed changes

* Fix pagination options and pagination key used by updater to insert created node
* Add name for software creation (otherwise we'll see an empty line)
* Fix another issue that I found : there was an error crashing the front when trying to remove an observable from an indicator (based on)

### Related issues

* #5198 

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/arsenal/vulnerabilities/AddSoftwares.jsx
+++ b/opencti-platform/opencti-front/src/private/components/arsenal/vulnerabilities/AddSoftwares.jsx
@@ -68,6 +68,7 @@ class AddSoftwares extends Component {
   render() {
     const { t, classes, vulnerability, vulnerabilitySoftwares } = this.props;
     const paginationOptions = {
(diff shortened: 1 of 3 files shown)


aHenryJard (commented):
Tested locally, all good to me (yes I know it's merged, but since I tested better to say it aloud).

PR #5238, opened 2023-12-15
[frontend] New filter’s UI in Taxii/Feed/Live/Trigger/Retention Policies + Knowledge Relationship  (#5012)

### LINKED ISSUE #5012: New filter’s UI in for stored filters (Stream/Feeds/Triggers/Automation)
## Use case

New filters' UI has been deployed for dynamic filters. We need to deploy it for stored filters.
BUT, it is important to not based stored filters on the data model yet. Let's stick to the already implemented stored filters

jpkha:
### Proposed changes

* Create a generic hook for filters: useFiltersState
* Applied new filters in Taxii section
* Transform Taxii, TaxiiCreation, TaxiiEdition to tsx
* Fix : sortBy value wasn't keep between navigation

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/5012

### Checklist

### Further comments


opencti-platform/opencti-front/src/private/components/data/Feed.tsx
```diff
@@ -0,0 +1,123 @@
...
+        handleSort={handleSort}
+        handleSearch={handleSearch}
+        displayImport={false}
+        secondaryAction={true}
+        keyword={feedState.searchTerm}
+      >
+        <QueryRenderer
+          query={FeedLinesQuery}
+          variables={{ count: 25, ...paginationOptions }}
+          render={({ props }: { props: FeedLinesPaginationQuery$data }) => (
```
lndrtrbn, line 92: I think QueryRenderer is used in class component because no alternative but we should try to remove it as we refacto to functional component, using hooks instead. Do you have the time to do it in this PR?

opencti-platform/opencti-front/src/private/components/data/Taxii.tsx
```diff
@@ -0,0 +1,123 @@
...
+        handleSort={handleSort}
+        handleSearch={handleSearch}
+        displayImport={false}
+        secondaryAction={true}
+        keyword={searchTerm}
+      >
+        <QueryRenderer
+          query={TaxiiLinesQuery}
+          variables={{ count: 25, ...paginationOptions }}
+          render={({ props }: { props: TaxiiLinesPaginationQuery$data }) => (
```
lndrtrbn, line 93: Same question here about refactoring with hooks

opencti-platform/opencti-front/src/utils/filters/filtersManageStateUtil.ts
```diff
@@ -0,0 +1,90 @@
+import { Filter, FilterGroup } from './filtersUtils';
```
lndrtrbn, line 1: This file could be a good candidate to start making unit testing in front side

opencti-platform/opencti-front/src/utils/filters/useFiltersState.tsx
```diff
@@ -0,0 +1,118 @@
...
+interface useFiltersStateProps {
+  filters: FilterGroup,
+  latestAddFilterId?: string
+}
+const useFiltersState = (initFilters: FilterGroup = emptyFilterGroup): [FilterGroup, handleFilterHelpers] => {
+  const [filtersState, setFiltersState] = useState<useFiltersStateProps>({
+    filters: initFilters,
+    latestAddFilterId: undefined,
+  });
+  const helpers: handleFilterHelpers = {
```
lndrtrbn, line 25: Those helpers are re-renders every time the hook is rendered. I don't know if it's really an issue here but IMO it could be a case where `memo` has benefits.

An other point disturbing me is the fact to use `filterState` to call `setFiltersState` in helpers functions. Again maybe there is no issue right now but could lead to hard debugging ones later maybe. Have a look at https://react.dev/reference/react/useState#updating-state-based-on-the-previous-state

lndrtrbn (changes requested):
Nice improvements 💪 Just a few remarks, and I'll need a bit of help to know how to test locally

opencti-platform/opencti-front/src/utils/filters/filtersManageStateUtil.ts
lndrtrbn: (This remark is more a global thought than a request to you to make those tests 😅)

opencti-platform/opencti-front/src/private/components/data/Feed.tsx
jpkha: yes sure, I can try to use hooks instead. I think I have enough time to make it.

opencti-platform/opencti-front/src/utils/filters/useFiltersState.tsx
jpkha: Good point. I will look into it and let you know. Thanks 👍

jpkha:
> Nice improvements 💪 Just a few remarks, and I'll need a bit of help to know how to test locally

Yes sure, you can access at this adress : 
http://localhost:3000/dashboard/data/sharing/streams 
or data > data sharing (tabs).
You can test the three (Live streams, CSV Feeds, Taxii collections ) you should able to modify and create with the new filter

opencti-platform/opencti-front/src/utils/filters/filtersManageStateUtil.ts
jpkha: Every utils files as they are pure function are good candidate IMO :).
We can check together how we should handle the test in Front.

opencti-platform/opencti-front/src/utils/filters/useFiltersState.tsx
jpkha: Instead of calling filterState, I am calling prevState now. Very nice call.
For the first point, I am not sure how we should optimize with `memo` in this case.
Maybe `useCallback `on each helpers function ?

opencti-platform/opencti-front/src/utils/filters/filtersManageStateUtil.ts
jpkha: Let's talk about it in the next test session.

opencti-platform/opencti-front/src/private/components/data/Feed.tsx
jpkha: I am sorry, finally I did not have time to do it on this PR

Archidoit:
Missing new filter UI for live triggers 
image

Archidoit:
Same for retention policies
image

Archidoit:
Same for retention policies
image

Archidoit:
Can't update stream with a filter with nil operator:
- Update a live stream
- Add a filter on a field with the operator 'nil' or 'not_nil'
image

- The stream filters are not updated
image

Archidoit:
Can't create a stream with a filter with operator=nil:
- Create a live stream with filter: marking is empty
- The live stream is created with no filters

Archidoit:
In case of instance trigger creation, il suggest we display directly the only filter possible (that is 'related entity') instead of doing this : 
image

because the connectedToId filter is special: only a OR operator between the values

Archidoit:
Bug if we switch from instance trigger to classic trigger:
- Begin to fill the form to create an instance trigger, and then click elsewhere
- Try to create an other trigger, the 'instance trigger' field choice is reset to 'no', but not the filters: we can only chose a 'related entity' filter : 
image

--> present in master, I create an issue https://github.com/OpenCTI-Platform/opencti/issues/5449

jpkha:
Commit merged in this PR :
https://github.com/OpenCTI-Platform/opencti/pull/5407

PR #5252, opened 2023-12-18
Capabilities by entity types

Bonsai8863:
### Proposed changes

* Allow admins to define more granular access control over specific entity types
* Added a tab in role edit drawer for entity overrides
* Created new Security component to support overridden entity types
* Support granular RBAC on Reports; will add to other entity types upon approval of design

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/database/redis.ts
+++ b/opencti-platform/opencti-graphql/src/database/redis.ts
@@ -171,9 +171,15 @@ const delKeyWithList = async (keyId: string, listIds: string[]) => {
   await Promise.all([keyPromise, ...listsPromise]);
 };
 const setKeyWithList = async (keyId: string, listIds: string[], keyData: any, expirationTime: number) => {
-  const keyPromise = getClientBase().set(keyId, JSON.stringify(keyData), 'EX', expirationTime);
-  const listsPromise = listIds.map((listId) => setInList(listId, keyId, expirationTime));
-  await Promise.all([keyPromise, ...listsPromise]);
+  // Wrapped in try/catch fixed an issue with invalid expiration time in set
+  try {
+    const keyPromise = getClientBase().set(keyId, JSON.stringify(keyData), 'EX', expirationTime);
+    const listsPromise = listIds.map((listId) => setInList(listId, keyId, expirationTime));
+    await Promise.all([keyPromise, ...listsPromise]);
+    return keyData;
+  } catch (err) {
+    logApp.info(`[REDIS_DEBUG] expirationTime: ${expirationTime}`);
+  }
   return keyData;
 };
 const keysFromList = async (listId: string, expirationTime?: number) => {
--- a/opencti-platform/opencti-graphql/src/domain/subType.ts
+++ b/opencti-platform/opencti-graphql/src/domain/subType.ts
@@ -14,18 +14,21 @@ import { ENTITY_TYPE_EXTERNAL_REFERENCE } from '../schema/stixMetaObject';
 
 // -- ENTITY TYPES --
 
+// TODO: Temporary list of overridable sub types. Remove when support all sub types
+const isOverridable = ['Report'];
+
 export const queryDefaultSubTypes = async (context: AuthContext, user: AuthUser, search : string | null = null) => {
   const queryDefaultSubTypesFn = async () => {
     const sortByLabel = R.sortBy(R.toLower);
     const types = schemaTypesDefinition.get(ABSTRACT_STIX_DOMAIN_OBJECT).filter((n) => n.includes(search ?? ''));
     const finalResult = R.pipe(
       sortByLabel,
-      R.map((n) => ({ node: { id: n, label: n } })),
-      R.append({ node: { id: ABSTRACT_STIX_CORE_RELATIONSHIP, label: ABSTRACT_STIX_CORE_RELATIONSHIP } }),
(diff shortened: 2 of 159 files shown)


richard-julien:
Just rebased and upgraded to support new api results.

My testing results:

1. Configuration screen must be improved
image
Spacing/padding must be reworked.
Looks like there is some extra scrollbar in the list.

2. Configuration error on override
image
Some capabilities cannot be override (and must not be) but screen tell the user that is the case.

3. If base capabilities has no access knowledge, override is not accessible
I created a role with no capabilities + override to have access on Report and malware.
When I log with the user with this role the UI doesnt load any menu like when a user have no capabilities at all.
The menu on the left must be displayed according to the override possibilities.

4. I think security is missing on generic api. Like stixDomainObjectEdit(id: ID!) that will not check if the user have really the right to modify the element taking into account the overrides

On the technical side i will comment directly in the PR

opencti-platform/opencti-graphql/src/database/redis.ts
```diff
@@ -171,9 +171,15 @@ const delKeyWithList = async (keyId: string, listIds: string[]) => {
   await Promise.all([keyPromise, ...listsPromise]);
 };
 const setKeyWithList = async (keyId: string, listIds: string[], keyData: any, expirationTime: number) => {
-  const keyPromise = getClientBase().set(keyId, JSON.stringify(keyData), 'EX', expirationTime);
-  const listsPromise = listIds.map((listId) => setInList(listId, keyId, expirationTime));
-  await Promise.all([keyPromise, ...listsPromise]);
+  // Wrapped in try/catch fixed an issue with invalid expiration time in set
+  try {
```
richard-julien, line 175: Do you have repro case for that part that is not related to the feature?

opencti-platform/opencti-graphql/src/graphql/authDirective.js
```diff
@@ -6,6 +6,45 @@ import { defaultFieldResolver } from 'graphql/index.js';
...
+  }
+
+  try {
+    if (entityType) {
+      // Find the overrides of the user's role
+      const overrides = user?.roles?.flatMap((role) => role?.capabilities_overrides);
+
+      // Filter the overrides to be of the specified entity type
+      const filteredOverrides = overrides
+        ?.filter((capability) => capability?.entity.replace(/-/g, '') === entityType);
```
richard-julien, line 32: Im not really happy about this transform because of the graphql enum that doesnt support the dash.
Not sure but maybe removing the enum in the graphql to just a plain string to align the type definition could be a better approach

opencti-platform/opencti-graphql/src/graphql/authDirective.js
```diff
@@ -6,6 +6,45 @@ import { defaultFieldResolver } from 'graphql/index.js';
...
+
+      // Map the override capabilities to just their names
+      userCapabilities = map(
+        (capability) => capability.name,
+        filteredOverrides?.[0]?.capabilities
```
richard-julien, line 37: A bit confused about the [0] here

opencti-platform/opencti-graphql/src/domain/subType.ts
```diff
@@ -19,11 +19,11 @@ export const queryDefaultSubTypes = async (context: AuthContext, user: AuthUser,
     const types = schemaTypesDefinition.get(ABSTRACT_STIX_DOMAIN_OBJECT).filter((n) => n.includes(search ?? ''));
     const finalResult = R.pipe(
       sortByLabel,
-      R.map((n) => ({ node: { id: n, label: n } })),
-      R.append({ node: { id: ABSTRACT_STIX_CORE_RELATIONSHIP, label: ABSTRACT_STIX_CORE_RELATIONSHIP } }),
-      R.append({ node: { id: STIX_SIGHTING_RELATIONSHIP, label: STIX_SIGHTING_RELATIONSHIP } }),
-      R.append({ node: { id: ABSTRACT_STIX_CYBER_OBSERVABLE, label: ABSTRACT_STIX_CYBER_OBSERVABLE } }),
-      R.append({ node: { id: ENTITY_HASHED_OBSERVABLE_ARTIFACT, label: ENTITY_HASHED_OBSERVABLE_ARTIFACT } }),
+      R.map((n) => ({ node: { id: n, label: n, overridable: true } })),
```
richard-julien, line 23: Can you explain the new overridable value?

Bonsai8863: Overridable determines if the entity type appears on the 'Entities Override' tab on the role edit drawer. This line ensures the entity types can be selected for overriding capabilities.

opencti-platform/opencti-graphql/src/graphql/authDirective.js
Bonsai8863: filteredOverrides is the overrides property, just filtered down to the specified entity type. The overrides property could contain the overridden capabilities for several entities, but the filter should bring it down to just one. The [0] here is just accessing that filtered entity's overrides.

opencti-platform/opencti-graphql/src/database/redis.ts
Bonsai8863: No, I was sporadically getting redis issues and this was the only thing that staved it off for good. I wasn't sure what exactly what was causing it. Regardless, I believe it's best to wrap code that can throw errors in a try/catch block.
I last got this error maybe 5 months ago so I can't show you what exactly it was, but I can guarantee I haven't had it appear since.

Issue #5254, opened 2023-12-19
Filter menu is not showing on adding new filter and the menu is popping on sort. 


jpkha:
## Description

There are two issues with the filter menu:

- The menu fails to appear automatically when adding a new filter, and the filter list is empty initially.
- The popover appears when sorting the table.

PR #5259, opened 2023-12-19
Filter menu is not showing on adding new filter and the menu is popping on sort (#5254)

jpkha:
https://github.com/OpenCTI-Platform/opencti/issues/5254


SouadHadjiat (changes requested):
I just tested locally, and the issue is still here : when I go on reports for example and there is no filter at all, and I select a filter in the list, the filter popup doesn't open at first
image

If I remove this empty filter, and I add it again (or add another one), the popup opens this time.

SouadHadjiat (approved):
Tested locally, seems all good now ✅

PR #5262, opened 2023-12-19
[frontend/backend] Add filtering capabilities based on model definition (#4939)

### LINKED ISSUE #4939: Be able to filter on every properties with new filters
## Use case

Extend filtering capacity to every properties usable by users.
Give users helper to select easily relevant filters

### LINKED ISSUE #5032: Filter Refactoring Follow Up
- [x] #4939
- [x] #4940
- [x] #5012 
- [x] #5033
- [x] #3585

### LINKED ISSUE #5033: Ability to filter on Organization the data is shared with
### As a user I want to be able to filter on the organisation the item is shared with so that it’s convenient for me
- Acceptance criteria: Be able to filter Objects based on Organization they are shared with
GIVEN I’m on an item that can be shared
AND I’m EE 
WHEN I look at the filter
THEN I should see the possibility to filter on Shared with 

- Acceptance criteria: The filter is only available when EE is enabled
GIVEN I’m on an item that can be shared
AND I’m not EE
WHEN I look at the filter
THEN I should not see the filter

### LINKED ISSUE #5976: Entity type filter in relationship creation form can be removed
## Reproducible Steps

Go to an entity. Go to the Knowledge tab.
Select an entity type in the right panel (exemple: Campaign).
Click on the 'add' button to add a relationship.
A panel opens with the list of the available campaigns, and a filter on Campaigns at the top of the list : 
image

If you remove the filter, you have the list of all the entities.
Try to add the filter again by selecting 'entity type' in the filter list. You can only add the 'campaign' entity type (logic because the context is for Campaigns). 
image

## Expected Output

You shouldn't be able to remove the entity_type filter in this case.

## Actual Output

The entity_type filter can be removed.

Archidoit:
issue: https://github.com/OpenCTI-Platform/opencti/issues/4939 and https://github.com/OpenCTI-Platform/opencti/issues/5033 and https://github.com/OpenCTI-Platform/opencti/issues/5976

Notion page: https://www.notion.so/filigran/Add-filtering-capabilities-based-on-model-definition-efe70111ee7b43608002a62edaf9cc01


labo-flg:
I'm in Data/Entities, and some attributes appear twice in the list.
Probably 2 attributes with same name coming from different entities in schema.
image

We might want to either deduplicate (if key is same, and both type of entities are returned), or label it a bit differently (with entity type in parenthesis for instance)

Archidoit:
> I'm in Data/Entities, and some attributes appear twice in the list. Probably 2 attributes with same name coming from different entities in schema. image
> 
> We might want to either deduplicate (if key is same, and both type of entities are returned), or label it a bit differently (with entity type in parenthesis for instance)

Indeed I've done this check for attributes (different attributes should have different labels) but not for refs... (and sample is a ref). I'm gonna do that ! ---> FIXED

Archidoit:
Bug remaining: can't add several entities in 'related entity' filters (instance trigger)

labo-flg:
I found a potential bug when filtering relationships on "source type" (or "target type" too).
With a AND on 2 different types, I suppose I should not be given any result because the source type cannot be of 2 different types... but:

image


Filter sent is:

```json
{
  "count": 25,
  "orderMode": "desc",
  "orderBy": "created_at",
  "filters": {
    "mode": "and",
    "filters": [
      {
        "key": "entity_type",
        "values": [
          "stix-core-relationship"
        ],
        "operator": "eq",
        "mode": "or"
      }
    ],
    "filterGroups": [
      {
        "mode": "and",
        "filters": [
          {
            "key": "fromTypes",
            "values": [
              "Attack-Pattern",
              "Malware"
            ],
            "operator": "eq",
            "mode": "and"
          }
        ],
        "filterGroups": []
      }
    ]
  }
}
```
--> bug in master, cf issue https://github.com/OpenCTI-Platform/opencti/issues/6001

labo-flg:
I've tested a lot of places, but definitely not all of them.
All seems ok, including your last fix on instance triggers "Related entity" filter.

labo-flg (approved):

Archidoit:
DO NOT MERGE

Archidoit:
PR approved but still in draft because some bugs are remaining:
- error at background task creation

jpkha:
In the relation_type. You cannot delete from the chip. But you can still clear it by click on the "clear filters" button.
IMO we should in this case keep the initial state
image

jpkha (approved):

Issue #4304, opened 2023-09-11
Add Confidence level on Users


Jipegien:
## Use case

Users have a Max Confidence level that will be used when creating Knowledge. 
As connectors, stream and feeds used a user to upsert data, confidence level in their config is no longer needed. They will use the max confidence level of the users they are associated with.

In User's edition form, a dedicated panel allows to set the max confidence level

### No migration

We do not know what is the value used in the connectors configuration files of our users, and we cannot know which user is used for a connector, thus we cannot magically migrate the existing users with a smart max confidence value.
For instance, migrating users to 100 would be a very dangerous move as connectors would suddenly be able to overwrite all data in the platform. 

For this reason, no migration will be executed. Platform admins are expected to configure their confidence policy throughout their platform.

### Subtasks:
* Managing a “max confidence level” for a user
* Enforce specific rights to manage confidence levels of users (capability SETACCESS)
* [Product] Anticipate the breaking change using Filigran's communication tools

### Linked to issues
* #4901 
* #5692 
* #5694
* #5695
* #5697
* #5696

### Next steps
* Using max confidence level of the associated user when upserting data
* Taking into account the max confidence level of users when they try to modify the confidence level of an Object

### Further comments
Note this feature will then be leveraged for #2517 in the future.

SamuelHassine:
Also implement the option "validate_before_import".

PR #5283, opened 2023-12-21
[backend] add user_confidence_level attribute to Users (#4303)

labo-flg:
### Proposed changes

* new attribute `user_confidence_level `with default value and overrides per entity
* default confidence is set from platform settings (default to 100)
* elastic mapping as nested, to query overrides properly
* new migration script to add the mapping and update all users with default confidence

### Related issues

* #4304 

### Checklist

### Further comments

This Chunk 1 addresses the data model and align the database. 
New users are created with defaut (100) confidence to match current behavior ; old users are migrated with same value.

### DIFF
--- a/opencti-platform/opencti-graphql/src/database/engine.js
+++ b/opencti-platform/opencti-graphql/src/database/engine.js
@@ -607,7 +607,7 @@ const computePlatformMappings = () => {
           overrides: {
             type: 'nested',
             properties: {
-              entity_type: { type: 'text' },
+              entity_type: { type: 'keyword' },
               max_confidence: { type: 'integer' },
             },
           }
--- a/opencti-platform/opencti-graphql/src/database/engine.js
+++ b/opencti-platform/opencti-graphql/src/database/engine.js
@@ -601,6 +601,18 @@ const computePlatformMappings = () => {
           date_seen: { type: 'date' },
         },
       },
+      user_confidence_level: {
+        properties: {
+          max_confidence: { type: 'integer' },
+          overrides: {
+            type: 'nested',
+            properties: {
+              entity_type: { type: 'text' },
+              max_confidence: { type: 'integer' },
+            },
+          }
+        },
+      },
       timestamp: {
         type: 'date',
       },
--- a/opencti-platform/opencti-graphql/src/database/middleware.js
+++ b/opencti-platform/opencti-graphql/src/database/middleware.js
@@ -891,9 +891,6 @@ const rebuildAndMergeInputFromExistingData = (rawInput, instance) => {
   if (isDictionaryAttribute(key) && isNotEmptyField(R.head(value))) { // Only allow full cleanup
     throw UnsupportedError('Dictionary attribute cant be updated directly', { rawInput });
   }
-  if (!isMultiple && isObjectAttribute(key)) {
-    throw UnsupportedError('Object update only available for array attributes', { rawInput });
-  }
   // region rebuild input values consistency
   if (key.includes('.')) {
     // In case of dict attributes, patching the content is possible through first level path
--- a/opencti-platform/opencti-graphql/src/domain/user.js
+++ b/opencti-platform/opencti-graphql/src/domain/user.js
@@ -1,4 +1,5 @@
 import bcrypt from 'bcryptjs';
+import nconf from 'nconf';
 import { authenticator } from 'otplib';
 import * as R from 'ramda';
 import { uniq } from 'ramda';
@@ -76,6 +77,33 @@ const AUTH_BEARER = 'Bearer';
 const AUTH_BASIC = 'BasicAuth';
 export const TAXIIAPI = 'TAXIIAPI';
 const PLATFORM_ORGANIZATION = 'settings_platform_organization';
+
+export const userConfidenceSchema = {
+  type: 'object',
+  properties: {
+    max_confidence: { type: 'number' },
+    overrides: {
+      type: 'array',
+      items: {
+        type: 'object',
+        properties: {
+          entity_type: { type: 'string' },
+          max_confidence: { type: 'number' }
+        },
+        required: ['entity_type', 'max_confidence']
+      }
+    }
+  },
+  required: ['max_confidence', 'overrides']
+};
+
+/**
+ map: {
+ type: "object",
+ required: [],
+ additionalProperties: {type: "number"},
+ },
+ */
 
 const roleSessionRefresh = async (context, user, roleId) => {
   // Get all groups that have this role
@@ -498,6 +526,7 @@ export const addUser = async (context, user, newUser) => {
     R.assoc('account_status', newUser.account_status ? newUser.account_status : DEFAULT_ACCOUNT_STATUS),
     R.assoc('account_lock_after_date', newUser.account_lock_after_date),
     R.assoc('unit_system', newUser.unit_system),
+    R.assoc('user_confidence_level', { max_confidence: nconf.get('app:user_confidence_level_default') ?? 100, overrides: [] }),
     R.dissoc('roles')
   )(newUser);
   const { element, isCreation } = await createEntity(context, user, userToCreate, ENTITY_TYPE_USER, { complete: true });
--- a/opencti-platform/opencti-graphql/src/domain/user.js
+++ b/opencti-platform/opencti-graphql/src/domain/user.js
@@ -97,14 +97,6 @@ export const userConfidenceSchema = {
   required: ['max_confidence', 'overrides']
 };
 
-/**
- map: {
- type: "object",
- required: [],
- additionalProperties: {type: "number"},
- },
- */
-
 const roleSessionRefresh = async (context, user, roleId) => {
   // Get all groups that have this role
   const groupsRoles = await listAllRelations(context, user, RELATION_HAS_ROLE, { toId: roleId, fromTypes: [ENTITY_TYPE_GROUP] });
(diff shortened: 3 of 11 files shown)


opencti-platform/opencti-graphql/src/database/engine.js
```diff
@@ -602,6 +602,18 @@ const computePlatformMappings = () => {
...
       },
+      user_confidence_level: {
+        properties: {
+          max_confidence: { type: 'integer' },
+          overrides: {
+            type: 'nested',
+            properties: {
+              entity_type: { type: 'text' },
+              max_confidence: { type: 'integer' },
+            },
```
labo-flg: using `nested` top be able to query the overrides inside the array properly (like "give me users with overrides on Reports greater than 40").

Not needed now, might be in the future for administrative purposes.

opencti-platform/opencti-graphql/src/domain/user.js
```diff
@@ -498,6 +526,7 @@ export const addUser = async (context, user, newUser) => {
     R.assoc('account_status', newUser.account_status ? newUser.account_status : DEFAULT_ACCOUNT_STATUS),
     R.assoc('account_lock_after_date', newUser.account_lock_after_date),
     R.assoc('unit_system', newUser.unit_system),
+    R.assoc('user_confidence_level', { max_confidence: nconf.get('app:user_confidence_level_default') ?? 100, overrides: [] }),
```
labo-flg, line 521: user creation does not take the confidence as input; it's an internal attribute.
We will be able to update the confidence once created ofc.

opencti-platform/opencti-graphql/src/schema/schema-validator.ts
```diff
@@ -71,7 +71,18 @@ export const validateAndFormatSchemaAttribute = (
...
+        if (attributeDefinition.multiple) {
+          // type=object && multiple -> it's an array than can be validated as-is
+          valid = validate(validationValues);
+        } else {
+          // object not multiple, only one value to check
+          if (validationValues.length > 1) {
+            throw ValidationError(attributeName, { message: 'Object is not multiple but has more than 1 value', data: { attribute: attributeName } });
+          }
+          valid = validate(validationValues[0]);
+        }
```
labo-flg: the function `validateAndFormatSchemaAttribute` checks that the input provided is conform to the schema attribute definition. The input MUST be provided as an array, so when the value is unique, it's still turned into and array with `value: Array.isArray(v) ? v : [v]` (see `validateInputCreation`)

This case is here for when the type of attribute is `object`. There are very few cases of objects : TAI's `weight`, `height` ; and now the `user_confidence_level`.

Both `weight` and `height` attributes are actually **array** object, so the `multiple` flag is true.
`user_confidence_level` is a plain object, not an array so `multiple` is false.

➡️ We were not taking `multiple` into account to know if the values are actually to be validated as an array or not. It was only working with arrays (fortunately, our only case so far, before adding the confidence level).

opencti-platform/opencti-graphql/src/domain/user.js
SouadHadjiat: maybe `nconf.get('app:user_confidence_level_default')` should be parsed in integer before setting it here ? and fallback to 100 in any case ?

opencti-platform/opencti-graphql/src/database/engine.js
```diff
@@ -602,6 +602,18 @@ const computePlatformMappings = () => {
           date_seen: { type: 'date' },
         },
       },
+      user_confidence_level: {
+        properties: {
+          max_confidence: { type: 'integer' },
+          overrides: {
+            type: 'nested',
+            properties: {
+              entity_type: { type: 'text' },
```
SouadHadjiat: I'm not sure type: 'text' is enough, I think we also need to define "keyword" field in entity_type, if we want to be able to do a term query on entity_type.

opencti-platform/opencti-graphql/src/domain/user.js
```diff
@@ -77,6 +78,25 @@ const AUTH_BASIC = 'BasicAuth';
...
+        type: 'object',
+        properties: {
+          entity_type: { type: 'string' },
+          max_confidence: { type: 'number' }
+        },
+        required: ['entity_type', 'max_confidence']
+      }
+    }
+  },
+  required: ['max_confidence', 'overrides']
```
SouadHadjiat, line 97: how will we update this object with the api ? will we need to replace the whole object if we want to update only max_confidence ? I just tried to do a fieldPatch on user_confidence_level and it won't work if I want to only update max_confidence, I need to provide the whole object.

I tried this first (fails with a schema validation error) : 
```
fieldPatch(
      input: {
        key: "user_confidence_level"
        object_path: "max_confidence"
        value: [80]
      }
    )
``` 

Then I tried this, which fails because overrides is not present (and it's wrong because I don't want to replace the whole object) : 
``` 
fieldPatch(
      input: {
        key: "user_confidence_level"
        value: [{
          max_confidence: 80,
        }]
      }
    )
``` 

Then when I added to the input overrides with an empty array, I got this error : `Object update only available for array attributes`

opencti-platform/opencti-graphql/src/database/engine.js
labo-flg: indeed !

opencti-platform/opencti-graphql/src/domain/user.js
labo-flg: I just checked, and nconf returns it as a number, no need for parsing.

opencti-platform/opencti-graphql/src/domain/user.js
labo-flg: field patch is actually not implemented for non-multiple objects in schema, it was only done in the context of TAI's weight and height fields.

A large refactoring is ongoing in the schema validation/mapping area, so we will wait for this and sync afterwards.

Meanwhile, let's just update the full field at once, not partially.

I'll remove the UnsupportedOperation error for non-multiple object update in middleware.js, as it works fine now the that the schema-validator properly checks them.

opencti-platform/opencti-graphql/src/migrations/1703150526503-user-confidence-level.js
```diff
@@ -0,0 +1,54 @@
...
+
+export const up = async (next) => {
+  await elUpdateIndicesMappings({
+    user_confidence_level: {
+      properties: {
+        max_confidence: { type: 'integer' },
+        overrides: {
+          type: 'nested',
+          properties: {
+            entity_type: { type: 'keyword' },
```
SouadHadjiat, line 13: type "keyword" is the right type I think, but you should probably check with Julien what he is planning when enforcing the mapping with these properties. Usually with string fields it's mapped like this: 

```
type: 'text',
fields: {
  keyword: {
    type: 'keyword',
    normalizer: 'string_normalizer',
    ignore_above: 512,
  },
}
```

labo-flg: I'm not sure, I'll check with him indeed.

our `string_normalizer` is doing a lowercase for instance ; right for the ids but not for the entity type which are upper cased like 'Report'.

opencti-platform/opencti-graphql/src/database/middleware.js
```diff
@@ -891,9 +891,6 @@ const rebuildAndMergeInputFromExistingData = (rawInput, instance) => {
   if (isDictionaryAttribute(key) && isNotEmptyField(R.head(value))) { // Only allow full cleanup
     throw UnsupportedError('Dictionary attribute cant be updated directly', { rawInput });
   }
-  if (!isMultiple && isObjectAttribute(key)) {
-    throw UnsupportedError('Object update only available for array attributes', { rawInput });
-  }
```
labo-flg, line 896: now handled properly in schema-validator

opencti-platform/opencti-graphql/src/modules/attributes/internalObject-registrationAttributes.ts
```diff
@@ -93,6 +94,7 @@ const internalObjectsAttributes: { [k: string]: Array<AttributeDefinition> } = {
     { name: 'description', type: 'string', mandatoryType: 'no', editDefault: false, multiple: false, upsert: false },
     { name: 'firstname', type: 'string', mandatoryType: 'no', editDefault: false, multiple: false, upsert: false },
     { name: 'lastname', type: 'string', mandatoryType: 'no', editDefault: false, multiple: false, upsert: false },
+    { name: 'user_confidence_level', type: 'object', mandatoryType: 'internal', editDefault: false, multiple: false, upsert: false, schemaDef: userConfidenceSchema },
```
SarahBocognano, line 97: Just out of curiosity, why did you choose to put the user_confidence_level attribute in internalObjectsAttributes instead of the User attributes directly ?

labo-flg: to my knowledge, attribute registrations are all listed in `X-registrationAttributes.ts` files. User is an internal object.

Which file are you thinking of ?

labo-flg:
Status update: I need to rebase (redo) this work over the schema refactoring done in https://github.com/OpenCTI-Platform/opencti/pull/5340

Too much stuff have changed, I need to rewrite the whole stuff. It will also enable the patch field for my case.

opencti-platform/opencti-graphql/src/modules/attributes/internalObject-registrationAttributes.ts
SarahBocognano: My bad, in my memory User was register in a graphql module and not like the old entities. So you are definetly in the good file ! :D

labo-flg:
Closing this PR in favor of https://github.com/OpenCTI-Platform/opencti/pull/5436

PR #5320, opened 2023-12-29
[backend] Fix ignoring boolean value when false in csv mapper (#5300)

### LINKED ISSUE #5300: CSV mapper => bug while ingesting malware entity?
## Description

Hello, as reported on Slack https://filigran-community.slack.com/archives/CHZC2D38C/p1703196931785729 I could not manage to ingest a malware entity from a CSV via the CSV mapper module. 
I placed as a test two columns with headers separated with semi column (malware_name and is_family with 'test' and false as values, respectively). See screenshots below: 

I checked the oasis 2.1 STIX documentation and the only difference I see maybe is the required "malware_types" field that is actually optional in opencti but it did not solve my problem unfortunately. 

## Environment

OpenCTI version:  OpenCTI 5.12.9

## Expected Output

I expect to see as when I tried with Intrusion Set entity that results into json data while running a test from the CSV mapper.

## Actual Output

No data can be previewed while running a test via the CSV mapper. Data are empty [ ]

## Screenshots (optional)
image (1)
image

lndrtrbn:
### Proposed changes

* Change a condition when validating the csv file data that ignored boolean false values

### Related issues

* #5300 

### Checklist

### Further comments

### DIFF
--- a/opencti-platform/opencti-graphql/src/parser/csv-mapper.ts
+++ b/opencti-platform/opencti-graphql/src/parser/csv-mapper.ts
@@ -135,7 +135,7 @@ const handleDirectAttribute = (attributeKey: string, column: AttributeColumn, in
     throw UnsupportedError('Invalid attribute', { key: attributeKey, type: entity_type });
   }
   const computedValue = computeValue(value, column, attributeDef);
-  if (computedValue) {
+  if (computedValue !== null && computedValue !== undefined) {
     input[attributeKey] = computedValue;
   }
 };
--- a/opencti-platform/opencti-graphql/tests/02-integration/05-parser/csv-parser-test.js
+++ b/opencti-platform/opencti-graphql/tests/02-integration/05-parser/csv-parser-test.js
@@ -165,8 +165,10 @@ describe('CSV-PARSER', () => {
     const bundle = await bundleProcess(testContext, ADMIN_USER, filPath, csvMapperMalware);
     const { objects } = bundle;
 
-    expect(objects.length).toBe(2);
+    expect(objects.length).toBe(4);
     expect(objects[0].is_family).toBe(true);
     expect(objects[1].is_family).toBe(false);
+    expect(objects[2].is_family).toBe(true);
+    expect(objects[3].is_family).toBe(true);
   });
 });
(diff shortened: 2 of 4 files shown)


opencti-platform/opencti-graphql/src/parser/csv-mapper.ts
```diff
@@ -135,7 +135,7 @@ const handleDirectAttribute = (attributeKey: string, column: AttributeColumn, in
     throw UnsupportedError('Invalid attribute', { key: attributeKey, type: entity_type });
   }
   const computedValue = computeValue(value, column, attributeDef);
-  if (computedValue) {
+  if (computedValue !== null && computedValue !== undefined) {
```
aHenryJard, line 138: question: from the test below I see that when there is an invalid value (empty) the line is dropped. Don't we need to throw an exception to warn the end user that the line has been dropped when computerValue is null or undefined (line 'Gerard' in malware.csv) ?

opencti-platform/opencti-graphql/tests/02-integration/05-parser/csv-parser-test.js
```diff
@@ -158,4 +159,14 @@ describe('CSV-PARSER', () => {
     expect(threatActorWithTypes.threat_actor_types.length)
       .toBe(2);
   });
+
+  it('Parse CSV - manage boolean values', async () => {
```
aHenryJard, line 163: praise: thanks for this test, it's making the fix really explicit ❤️

opencti-platform/opencti-graphql/src/parser/csv-mapper.ts
lndrtrbn: To discuss with product team. For what I understood, giving files with missing data is very common so I don't know if it's relevant or not. In any case, if to be done, should be in a specific feature

aHenryJard: ok thanks

aHenryJard (approved):

opencti-platform/opencti-graphql/src/parser/csv-mapper.ts
SouadHadjiat: it's already the case I guess, since there is a test in computeValue which will return null if the value is empty, which might be an issue. If you have an empty value (empty string or empty object or empty array), it will ignore the line ?

lndrtrbn: The line is not always ignored.
If a cell is empty, no matter the type (string, number, whatever), then we set null for the associated property of the entity.
Then, if this property is required in the schema, the line is dropped, otherwise we keep it, just some properties are not filled but not an issue as they are not required

PR #5323, opened 2023-12-29
[frontend] add the ability to edit user_confidence_level (#4304)

labo-flg:
### Proposed changes

* Edit user's confidence level from the user edition panel, overview tab
* Only users with SET_ACCESS capability will see this field

### Related issues

* #4304 

### Checklist

### Further comments

Overrides are taken into account in the code (not changed when editing the global confidence level) but are not editable for now. This is out of scope.

### DIFF
--- a/opencti-platform/opencti-front/src/components/list_lines/ListLinesContent.jsx
+++ b/opencti-platform/opencti-front/src/components/list_lines/ListLinesContent.jsx
@@ -268,7 +268,7 @@ class ListLinesContent extends Component {
                           registerChild(ref);
                         }}
                         autoHeight={true}
-                        height={height}
+                        height={height || 0}
                         onRowsRendered={onRowsRendered}
                         isScrolling={isScrolling}
                         onScroll={onChildScroll}
--- a/opencti-platform/opencti-front/src/private/components/common/form/ConfidenceField.tsx
+++ b/opencti-platform/opencti-front/src/private/components/common/form/ConfidenceField.tsx
@@ -19,8 +19,10 @@ const useStyles = makeStyles(() => ({
 }));
 
 interface ConfidenceFieldProps {
+  name?: string;
+  label?: string;
   variant?: string;
-  onSubmit?: (name: string, value: string | number | number[]) => void;
+  onSubmit?: (name: string, value: string) => void;
   onFocus?: (name: string, value: string) => void;
   editContext?:
   | readonly ({
@@ -34,6 +36,8 @@ interface ConfidenceFieldProps {
 }
 
 const ConfidenceField: FunctionComponent<ConfidenceFieldProps> = ({
+  name = 'confidence',
+  label,
   variant,
   onFocus,
   onSubmit,
@@ -43,6 +47,8 @@ const ConfidenceField: FunctionComponent<ConfidenceFieldProps> = ({
   disabled,
 }) => {
   const { t } = useFormatter();
+  const finalLabel = label || t('Confidence level');
+
   const classes = useStyles();
   return (
     <Alert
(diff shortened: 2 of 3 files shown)


opencti-platform/opencti-front/src/components/list_lines/ListLinesContent.jsx
```diff
@@ -268,7 +268,7 @@ class ListLinesContent extends Component {
                           registerChild(ref);
                         }}
                         autoHeight={true}
-                        height={height}
+                        height={height || 0}
```
labo-flg, line 271: height is undefined when there is no data to list, showing a warning in the console.

opencti-platform/opencti-front/src/private/components/common/form/ConfidenceField.tsx
```diff
@@ -19,8 +19,10 @@ const useStyles = makeStyles(() => ({
 }));
 
 interface ConfidenceFieldProps {
+  name?: string;
+  label?: string;
   variant?: string;
-  onSubmit?: (name: string, value: string | number | number[]) => void;
+  onSubmit?: (name: string, value: string) => void;
```
labo-flg, line 25: This was misleading: the value returned is an unparsed string every time..

opencti-platform/opencti-front/src/private/components/common/form/ConfidenceField.tsx
```diff
@@ -19,8 +19,10 @@ const useStyles = makeStyles(() => ({
 }));
 
 interface ConfidenceFieldProps {
+  name?: string;
+  label?: string;
```
labo-flg, line 23: making the ConfidenceField customizable now that we can use it with users

opencti-platform/opencti-front/src/private/components/settings/users/UserEditionOverview.tsx
```diff
@@ -132,16 +144,31 @@ UserEditionOverviewComponentProps
...
       .validateAt(name, { [name]: value })
       .then(() => {
-        commitFieldPatch({
-          variables: {
-            id: user.id,
-            input: { key: name, value: value || '' },
-          },
-        });
+        // specific case for user confidence level that must be updated as a full object (we cannot field patch non-multiple objects for now)
+        // We pass the existing overrides for this user so they are unchanged
```
labo-flg, line 141: Significant changes are currently in development around schema validation and it will impact how we can patch attributes. I don't want our work to collide so I won't change current behavior.

This solution here is sufficient for the moment, but could be updated when a fieldPatch could be done properly.

opencti-platform/opencti-front/src/private/components/settings/users/UserEditionOverview.tsx
```diff
@@ -102,24 +109,29 @@ UserEditionOverviewComponentProps
...
+        'account_lock_after_date',
+      ],
+      {
+        ...user,
+        objectOrganization,
+      },
+    ),
+    // extract object attribute
+    max_confidence: user.user_confidence_level.max_confidence,
+  };
```
lndrtrbn: Should we not get rid of this R.pick call completely ?

```suggestion
  const initialValues = {
    name: user.name,
    user_email: user.user_email,
    firstname: user.firstname,
    lastname: user.lastname,
    language: user.language,
    api_token: user.api_token,
    description: user.description,
    account_status: user.account_status,
    account_lock_after_date: user.account_lock_after_date,
    max_confidence: user.user_confidence_level.max_confidence,
    objectOrganization
  };
```

labo-flg: you're right !

opencti-platform/opencti-front/src/private/components/settings/users/UserEditionOverview.tsx
labo-flg: update: I got confirmation, field patch will indeed be possible next, but not in the way we do it now. I'll update here when this is available, but not in the scope of this PR.

lndrtrbn (approved):

SarahBocognano (approved):

labo-flg:
Closing this PR in favor of https://github.com/OpenCTI-Platform/opencti/pull/5436

Issue #5315, opened 2023-12-28
Error incorrect filter keys not existing in any schema definition when running background task with specific filters


SouadHadjiat:
## Description

When creating a background task with some specific filters (reproduced with source entity on relationships), I have this error : 
`incorrect filter keys not existing in any schema definition`

![Capture d'écran 2023-12-27 175809](https://github.com/OpenCTI-Platform/opencti/assets/668192/df7937a9-15a5-40d4-8478-9dda1339931a)

## Environment

1. OS (where OpenCTI server runs): { e.g. Mac OS 10, Windows 10, Ubuntu 16.4, etc. }
2. OpenCTI version: 5.12.14
3. OpenCTI client: frontend
4. Other environment details:

## Reproducible Steps

I reproduced locally with "source entity" filter, but I suspect if can be reproduced with other filters (target entity, source and target types)
1. Go to Data -> Relationships
2. Filter by source entity
3. Try to create a background task (remove or add a label for example) => it fails with the above error

## Additional information

I suspect `buildQueryFiltersContent` function which doesn't test correctly "key" (it's an array, not a string)

PR #5324, opened 2023-12-29
[backend] fix background tasks on relationships with an api filter (#5315)

Archidoit:
issue: https://github.com/OpenCTI-Platform/opencti/issues/5315

This PR fixes the bug and enables to do background tasks on relationships with api filters.
BUT a bug is still present for complex filtering combinations (https://github.com/OpenCTI-Platform/opencti/issues/5325), it will be resolved in another PR to enable a quick merge of the 5315 issue (that is critical).

### DIFF
--- a/opencti-platform/opencti-graphql/src/domain/backgroundTask.js
+++ b/opencti-platform/opencti-graphql/src/domain/backgroundTask.js
@@ -107,7 +107,7 @@ const buildQueryFilters = async (context, user, rawFilters, search, taskPosition
   }
   const newFilters = buildQueryFiltersContent(adaptedFilterGroup);
   // Avoid empty type which will target internal objects and relationships as well
-  const types = newFilters?.filters.filter((f) => f.key === TYPE_FILTER)?.values ?? [ABSTRACT_STIX_DOMAIN_OBJECT];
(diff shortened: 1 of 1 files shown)


opencti-platform/opencti-graphql/src/domain/backgroundTask.js
```diff
@@ -107,7 +107,7 @@ const buildQueryFilters = async (context, user, rawFilters, search, taskPosition
   }
   const newFilters = buildQueryFiltersContent(adaptedFilterGroup);
   // Avoid empty type which will target internal objects and relationships as well
-  const types = newFilters.filters.filter((f) => f.key === TYPE_FILTER)?.values ?? [ABSTRACT_STIX_DOMAIN_OBJECT];
+  const types = newFilters?.filters.filter((f) => f.key === TYPE_FILTER)?.values ?? [ABSTRACT_STIX_DOMAIN_OBJECT];
```
SouadHadjiat: There is still a issue here : `newFilters?.filters.filter((f) => f.key === TYPE_FILTER)` is an array, we can't get the values with `.values` which only returns an iterator of the array.

SouadHadjiat (changes requested):

opencti-platform/opencti-graphql/src/domain/backgroundTask.js
Archidoit: fixed

SouadHadjiat (approved):
tested locally ✅
only need to rebase master to fix failing test before merging @Archidoit

PR #5326, opened 2023-12-29
[frontend] i18n, extract keyword and solution to auto-translate

jpkha:
Work on a proof of concept (POC) for a manual script to extract translation values.

- Create an **`en.json`** file in the **`lang`** directory.
- Always rewrite the JSON at the moment. We should only merge new keys, so if there are any modified translations for existing keys, we won't overwrite them.
- It has limitations; for the filter feature, most of the keys to translate come from the backend. Therefore, it's dynamic on the front end, and we cannot always expect to have static values.

Propose an implementation of this solution: https://github.com/while1618/i18n-auto-translation 

It's implemented and working with Deepl-free, but there are multiple providers available.

However, I also found this solution: https://github.com/leolabs/json-autotranslate, which serves a similar purpose. I read, but need to check again, that it translates all the keys all over again. Something we want to avoid for many reasons, such as manually correcting the translation and avoiding translation for a key that is already translated.


lndrtrbn (commented):
About this point :
> It has limitations; for the filter feature, most of the keys to translate come from the backend. Therefore, it's dynamic on the front end, and we cannot always expect to have static values.

We could use a workaround to always have static translations keys.
Instead of setting a dynamic translation key with backend data we could create a map between backend data and translations.

Something like:
```
const ourTranslations = {
  `filter_${whatever}`: t('A static translation that can be auto translate'),
  ...
}
```

Instead of:
```
t(`filter_${whatever}`)
```

lndrtrbn (approved):
Process works fine, well done !
Just, it seems the german version missed translations, still a lot of english words

PR #5334, opened 2024-01-02
[frontend] sort exports by export date in lists (#5316)

### LINKED ISSUE #5316: Export list displays exported files in disorder
## Description

When you export a list of entities, you should see the list of exported files in order of export date (newest first), but they are displayed in random order. Which can be confusing when you do an export, you will see the export in progress on top, then it disapears in the list, as if it didn't work, but it is just in the middle of the list.

image

## Environment

2. OpenCTI version: 5.12.14
3. OpenCTI client: frontend
4. Other environment details:

lndrtrbn:
### Proposed changes

* Sort exported files by lastModified attribute

### Related issues

* #5316 

### Checklist

### Further comments

### DIFF
--- a/opencti-platform/opencti-graphql/src/modules/internal/document/document-domain.ts
+++ b/opencti-platform/opencti-graphql/src/modules/internal/document/document-domain.ts
@@ -10,7 +10,7 @@ import type { AuthContext, AuthUser } from '../../../types/user';
 import { type DomainFindById } from '../../../domain/domainTypes';
 import type { BasicStoreEntityDocument } from './document-types';
 import type { BasicStoreCommon } from '../../../types/store';
-import { FilterMode, FilterOperator, OrderingMode, type File } from '../../../generated/graphql';
+import { type File, FilterMode, FilterOperator, OrderingMode } from '../../../generated/graphql';
 import { loadExportWorksAsProgressFiles } from '../../../domain/work';
 import { UnsupportedError } from '../../../config/errors';
 import { elSearchFiles } from '../../../database/file-search';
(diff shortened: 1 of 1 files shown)


opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectsExportsContent.jsx
```diff
@@ -33,7 +33,8 @@ const StixCoreObjectsExportsContentComponent = ({
       subscription.unsubscribe();
     };
   });
-  const stixCoreObjectsExportFiles = data?.stixCoreObjectsExportFiles?.edges ?? [];
+  const stixCoreObjectsExportFiles = [...(data?.stixCoreObjectsExportFiles?.edges ?? [])]
+    .sort((f1, f2) => f2?.node?.lastModified?.localeCompare(f1?.node?.lastModified));
```
SouadHadjiat: if we sort only in the frontend, we might not get the whole list of exported files (and not get the latest ones). the API should return the ordered list. 
What is strange in the current implementation is that we only get 25 files, without pagination. So if our last export is not in them, we won't see it.

lndrtrbn: Oh yes.. I'll have a look in the backend instead to change this

richard-julien:
Force pushed for a default backend approach. Please take a look @lndrtrbn

lndrtrbn:
Tested, seems good to me. Better solution than I suggested 😂
Can you have a look @SouadHadjiat ?

Kedae (approved):

Issue #5333, opened 2024-01-02
Error trying to export relationships in Data > Relationships page


lndrtrbn:
## Description

In the page: Data > Relationships, I try to export some relationships (using csv export) but after submitting:
- the export dialog does not close and buttons are disabled
- the export is not done
- in console I have an error about filters

## Environment

1. OS (where OpenCTI server runs): Edge
2. OpenCTI version: 5.12.15
3. OpenCTI client: frontend
4. Other environment details:

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Go to Data > Relationships page
2. Select some relationships
3. Click on top-right button "Open export panel"
4. Choose csv export and click "Create"

## Expected Output

The export is done and the dialog closed

## Screenshots (optional)

Testing:
![image](https://github.com/OpenCTI-Platform/opencti/assets/33682259/ad087b2a-d5ac-4bff-b832-2cac6d9e7fb5)
Local:
![image](https://github.com/OpenCTI-Platform/opencti/assets/33682259/6c022d40-909d-47c0-a438-3657d0bd1a71)

PR #5342, opened 2024-01-02
[frontend] Error trying to export relationships in Data > Relationships page (#5333)

jpkha:
https://github.com/OpenCTI-Platform/opencti/issues/5333


jpkha:
Fix the current issue. You can now export any filter or with an empty filter, there is one bug left on the special filters where it's not working.

labo-flg (approved):

PR #5345, opened 2024-01-03
[backend] Introduce /health endpoint

richard-julien:
Introduce /health endpoint
* Remove deprecated built in apollo healthcheck
* Introduce /health endpoint protected by a key (default to changeme)
Example http://localhost:3000/health?health_access_key=test

### DIFF
--- a/opencti-platform/opencti-front/builder/dev/dev.js
+++ b/opencti-platform/opencti-front/builder/dev/dev.js
@@ -99,6 +99,7 @@ esbuild.context({
   });
   app.set("trust proxy", 1);
   app.use(compression({}));
+  app.use(middleware("/health"));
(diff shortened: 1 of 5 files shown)


lndrtrbn (commented):
- When shutting down minio, the error message is empty,
- When shutting down redis, the request seems not terminate at all.

Otherwise seems good

richard-julien:
> * When shutting down minio, the error message is empty,
> * When shutting down redis, the request seems not terminate at all.
> 
> Otherwise seems good

For minio, added a timeout to 5 sec to prevent long waiting.
For Redis, its due to the platform that retry undefinedly for now. It will be detected as a timeout by the process that will check this uri.

lndrtrbn (approved):

PR #5353, opened 2024-01-03
[backend] improve coverage tooling

labo-flg:
This branch tests how we can rely on codecov to improve our test base.

### DIFF
--- a/opencti-platform/opencti-graphql/package.json
+++ b/opencti-platform/opencti-graphql/package.json
@@ -168,6 +168,7 @@
     "@types/xml2js": "0.4.14",
     "@typescript-eslint/eslint-plugin": "6.14.0",
     "@typescript-eslint/parser": "6.14.0",
+    "@vitest/coverage-istanbul": "^1.2.0",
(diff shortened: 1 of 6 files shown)


labo-flg:
Codecov before: https://app.codecov.io/gh/OpenCTI-Platform/opencti/blob/master/opencti-platform%2Fopencti-graphql%2Fsrc%2Fmodules%2Fcase%2Fcase-domain.ts

Codecov after (in this PR): https://app.codecov.io/gh/OpenCTI-Platform/opencti/blob/tests%2Fimprove-case-test/opencti-platform%2Fopencti-graphql%2Fsrc%2Fmodules%2Fcase%2Fcase-domain.ts

We can see that the reporting is inaccurate. `batchParticipants` shall have been covered by the new test.

Coverage should have been improved much more.

labo-flg:
When running the coverage locally, we get correct reporting.

I've just tested on another file (schema/identifier.js) in my last commit,  and we can see the coverage improved and the correct lines now covered (looking at the html local report.

Codecov seems to have trouble then.

PR #5340, opened 2024-01-02
[frontent/backend] Enforcing schema to generate mappings and prevent any type confusion

Archidoit:
Schema refacto
review by Julien, Cathia & Laurent

PR #5444, opened 2024-01-09
[backend] Adapt migrations and initialization of empty platforms (#5340)

richard-julien:
* Adapt migrations to rely on background elastic task
* Change init to always reset the templates


opencti-platform/opencti-graphql/src/database/engine.js
```diff
@@ -683,12 +683,13 @@ const updateIndexTemplate = async (name, mapping_properties) => {
 };
 
 const elCreateIndexTemplate = async (index, mappingProperties) => {
-  const isExistByName = await engine.indices.existsIndexTemplate({ name: index }).then((r) => oebp(r));
   // Compat with platform initiated prior 5.9.X
-  const isExist = await engine.indices.existsIndexTemplate({ name: `${ES_INDEX_PREFIX}-index-template` }).then((r) => oebp(r));
-  if (isExistByName || isExist) {
+  const isPriorVersionExist = await engine.indices.existsIndexTemplate({ name: `${ES_INDEX_PREFIX}-index-template` })
```
SouadHadjiat, line 687: I don't understand this condition, shouldn't it be `{ name: index }` ?

richard-julien: Its a condition i have kept for older platform versions that doesnt have the new templates.

SouadHadjiat (approved):

Issue #5440, opened 2024-01-09
Long names when merging is not displayed correctly


nino-filigran:
## Description

When merging 2 entities with a long name, the look and feel is not correct: the name/value displayed is not truncated.

## Environment
 Demo

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Go observables
2. select 2 observables of the same type to merge (both of them needs to have a long name/value)
3. Click on merge in the Massive Operation 
4. The panel opens and the look and feel is odd

## Expected Output

The long name should be truncated

## Actual Output

The long names are displayed in full
## Screenshots (optional)
![image](https://github.com/OpenCTI-Platform/opencti/assets/148041607/c31f9729-4f2c-4879-9a56-356131c7fff8)

PR #5455, opened 2024-01-11
[frontend] truncate merged observable name (#5440)

marieflorescontact:
### Proposed changes

* Truncate observables name in merged entites panel
*

### Related issues

* [issue/5440](https://github.com/OpenCTI-Platform/opencti/issues/5440)
*

### Checklist

### Further comments

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/data/ToolBar.jsx
+++ b/opencti-platform/opencti-front/src/private/components/data/ToolBar.jsx
@@ -1863,7 +1863,8 @@ class ToolBar extends Component {
                         <ItemIcon type={element.entity_type} />
                       </ListItemIcon>
                       <ListItemText
-                        primary={truncate(defaultValue(element), 50)}
+                        style={{ overflowX: 'hidden', textOverflow: 'ellipsis' }}
+                        primary={defaultValue(element)}
                         secondary={truncate(
                           element.description
                             || element.x_opencti_description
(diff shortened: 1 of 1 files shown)


lndrtrbn (commented):
That's also feels odd to truncate a text where there is plenty of space to display it.
IMO the solution should be CSS one to use all space the content can have

image

To help you, something like that:
image

opencti-platform/opencti-front/src/private/components/data/ToolBar.jsx
```diff
@@ -1863,6 +1863,7 @@ class ToolBar extends Component {
                         <ItemIcon type={element.entity_type} />
                       </ListItemIcon>
                       <ListItemText
+                        style={{ overflowX: 'hidden', textOverflow: 'ellipsis' }}
```
lndrtrbn: Not working because the style here is targeting the root element of `ListItemText` but the text is inside a `span` itself inside the root element. We need to target this `span` instead.
```suggestion
                        sx={{
                          '.MuiListItemText-primary': {
                            overflowX: 'hidden',
                            textOverflow: 'ellipsis',
                          },
                        }}
```

lndrtrbn: `sx` is a prop of MaterialUI and we can pass objects with css classes as keys to target elements inside the component

lndrtrbn (approved):

PR #5457, opened 2024-01-11
[frontend] Relationship creation graph

### LINKED ISSUE #1417: UI: Creating Relation in Knowledge Graph
## Problem to Solve

Somewhere between 4.2.3 and 4.5.4, the knowledge graph UI changed quite a bit.

In 4.2.3, creating relations was as simple as clicking and dragging from one entity to another. In 4.5.4, one has to select 2 entities (done by holding Ctrl) and click on the create a relation button at the bottom right. This is not intuitive (especially the holding Ctrl part) and more cumbersome.

## Proposed Solution

Reenable the old way of creating relations, or any other more intuitive, easy and fast way.

## Additional Information

I may be missing a more intuitive way of creating relationships?

Kedae:
Closes #1417

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/analyses/groupings/GroupingKnowledgeGraph.jsx
+++ b/opencti-platform/opencti-front/src/private/components/analyses/groupings/GroupingKnowledgeGraph.jsx
@@ -42,6 +42,7 @@ import { hexToRGB } from '../../../../utils/Colors';
 import { UserContext } from '../../../../utils/hooks/useAuth';
 import investigationAddFromContainer from '../../../../utils/InvestigationUtils';
 import { isNotEmptyField } from '../../../../utils/utils';
+import RelationSelection from '../../../../utils/graph/RelationSelection';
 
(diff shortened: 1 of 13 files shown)


lndrtrbn (commented):
Tested locally, working nicely.
One thing is confusing me in component RelationSelection. Functions use component variables like `freeHand` but if the component is re-rendered we loose the current value of those variables. Can't it lead to strange behavior ?

And extract `contextmenu` callback into a variable (cf deepscan)

Kedae:
> Tested locally, working nicely. One thing is confusing me in component RelationSelection. Functions use component variables like `freeHand` but if the component is re-rendered we loose the current value of those variables. Can't it lead to strange behavior ?
> 
> And extract `contextmenu` callback into a variable (cf deepscan)

That might actually be a good point, but to be honest I took **a lot** of the work I did on LassoSelection and we never heard of weird behavior. If we have something comming in the future I sure will think about that.

lndrtrbn (approved):

Issue #5466, opened 2024-01-11
Knowledge > Export leads to unknown page (null in URL)


labo-flg:
## Description

When starting an export from an entity > knowledge view, the export starts correctly but the redirection does not work.
This is just a redirection error, the export starts and runs correctly.

## Environment

Latest Testing (NOT REPRODUCIBLE on prod)

## Reproducible Steps

1. Go to a country (also tested with a Sector)
2. Go to knowledge tab
3. Export with first neighbours, to PDF

## Expected Output

We're redirected to the DATA tab of the entity, where our export is visible

## Actual Output

platform's 404 banner

## Additional information

URL redirection should be like `<root>/dashboard/entities/sectors/1918baa7-0255-44a2-a169-d9e1bb6f522a/files?currentFileId=work_c0n3ct0r-f1l1-gr4n-00io-000000000026_2024-01-11T15%3A10%3A20.068Z`

But is like `<root>/dashboard/null/91545bbe-3c3d-4789-b563-58e22d69a9db/files?currentFileId=work_b882dc08-4a9c-436d-9c07-a2a8a10092da_2024-01-11T15%3A13%3A32.940Z`

labo-flg:
This is present on latest Testing, so a55e545

First investigations: 
Take the country views, and look at the `RootCountryQuery`. It should give you a complete `country` data object.
On prod, it's complete indeed. On latest master it's just a name and an id. This causes the bug here but probably other bugs.

PR #5472, opened 2024-01-11
[frontend] fix link resolution for file export (#5466)

labo-flg:
### Proposed changes

* Make sure all use of `StixDomainObjectHeader` have the right entity type set
* resolveLink utility is now insensitive to capitalization of the entity type, to avoid future typo problems)

### Related issues

* #5466 

### Checklist

### Further comments
To test, go to a Country knowledge view > export (any format) and hit Create. You should properly redirected to the entity's data tab.

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/analyses/malware_analyses/Root.tsx
+++ b/opencti-platform/opencti-front/src/private/components/analyses/malware_analyses/Root.tsx
@@ -88,6 +88,7 @@ const RootMalwareAnalysis = () => {
               return (
                 <>
                   <StixDomainObjectHeader
+                    entityType={'Malware-Analysis'}
                     stixDomainObject={malwareAnalysis}
                     PopoverComponent={
                       <MalwareAnalysisPopover id={malwareAnalysisId} />
--- a/opencti-platform/opencti-front/src/private/components/common/stix_domain_objects/StixDomainObjectHeader.jsx
+++ b/opencti-platform/opencti-front/src/private/components/common/stix_domain_objects/StixDomainObjectHeader.jsx
@@ -485,7 +485,7 @@ const StixDomainObjectHeader = (props) => {
           )}
           <StixCoreObjectFileExport
             id={stixDomainObject.id}
-            type={stixDomainObject.entity_type}
+            type={entityType}
           />
           {isKnowledgeUpdater && (
             <StixCoreObjectContainer elementId={stixDomainObject.id} />
--- a/opencti-platform/opencti-front/src/utils/Entity.ts
+++ b/opencti-platform/opencti-front/src/utils/Entity.ts
@@ -1,10 +1,10 @@
-export const resolveLink = (type: string): string | null => {
+export const resolveLink = (type = 'unknown'): string | null => {
   switch (type) {
     case 'Dashboard':
-    case 'dashboard':
+    case 'dashboard': // for using resolveLink in workspaces
       return '/dashboard/workspaces/dashboards';
     case 'Investigation':
-    case 'investigation':
+    case 'investigation': // for using resolveLink in workspaces
       return '/dashboard/workspaces/investigations';
     case 'Attack-Pattern':
       return '/dashboard/techniques/attack_patterns';
--- a/opencti-platform/opencti-front/src/private/components/locations/administrative_areas/Root.tsx
+++ b/opencti-platform/opencti-front/src/private/components/locations/administrative_areas/Root.tsx
@@ -94,7 +94,7 @@ const RootAdministrativeAreaComponent = ({
           }}
         >
           <StixDomainObjectHeader
-            entityType="administrativeArea"
+            entityType="Administrative-Area"
             disableSharing={true}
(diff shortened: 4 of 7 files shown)


opencti-platform/opencti-front/src/private/components/analyses/malware_analyses/Root.tsx
```diff
@@ -88,6 +88,7 @@ const RootMalwareAnalysis = () => {
               return (
                 <>
                   <StixDomainObjectHeader
+                    entityType={'Malware-Analysis'}
```
labo-flg, line 91: this one was missing

opencti-platform/opencti-front/src/private/components/common/stix_domain_objects/StixDomainObjectHeader.jsx
```diff
@@ -485,7 +485,7 @@ const StixDomainObjectHeader = (props) => {
           )}
           <StixCoreObjectFileExport
             id={stixDomainObject.id}
-            type={stixDomainObject.entity_type}
+            type={entityType}
```
labo-flg, line 488: This could never work as `stixDomainObject` is a relay object and `entity_type` is inside a fragment

The entity type is given as prop here, so we have it already.

opencti-platform/opencti-front/src/utils/Entity.ts
```diff
@@ -1,10 +1,9 @@
 export const resolveLink = (type: string): string | null => {
-  switch (type) {
+  const upperCasedType = `${type[0].toUpperCase()}${type.slice(1)}`;
+  switch (upperCasedType) {
```
labo-flg: This won't cover all typos, but this one is quite common.

SouadHadjiat: I'm not sure to understand why we need to uppercase the first letter, for which cases do we need to ?

opencti-platform/opencti-front/src/private/components/analyses/malware_analyses/Root.tsx
SouadHadjiat: indeed the bug happens even in prod for malware analysis

SouadHadjiat:
There is still this one which should be `Administrative-Area` : 

``` 
<StixDomainObjectHeader
            entityType="administrativeArea"
            disableSharing={true}
            stixDomainObject={administrativeArea}
            PopoverComponent={
              <AdministrativeAreaPopover id={administrativeArea.id} />
            }
            enableQuickSubscription={true}
          />
```

SouadHadjiat:
Also, about the uppercase fix, I think it's better to put the right entity_type with the right case, or change the method to put everything in lowercase and lowercase the type. 
Anyway I found country and city which are in lowercase here : 

```
<StixDomainObjectHeader
            entityType="country"
            disableSharing={true}
            stixDomainObject={country}
            PopoverComponent={<CountryPopover id={country.id} />}
            enableQuickSubscription={true}
          />

<StixDomainObjectHeader
            entityType="city"
            disableSharing={true}
            stixDomainObject={city}
            PopoverComponent={<CityPopover id={city.id} />}
            enableQuickSubscription={true}
          />
```

labo-flg:
Yes you're right, let's use the right entity type, it makes much more sense.

opencti-platform/opencti-front/src/utils/Entity.ts
labo-flg: I've removed this tweak, this is irrelevant.

We shall use the right entity_type and that's it.
It's more a Typescript issue at this point (not setting prop entityType to `string` but the enum from GQL.

opencti-platform/opencti-front/src/utils/Entity.ts
```diff
@@ -1,10 +1,8 @@
-export const resolveLink = (type: string): string | null => {
+export const resolveLink = (type = 'unknown'): string | null => {
   switch (type) {
     case 'Dashboard':
-    case 'dashboard':
```
SouadHadjiat, line 4: I'm not sure we should remove this (and 'investigation'), I guess they have been added for a reason

SouadHadjiat: it was added here : https://github.com/OpenCTI-Platform/opencti/commit/dab14295b98dcc516d00e3295def4ee97c89676d#diff-93d809f17b979aa809952ecbb1617170d9b409015c6881c6fac76920340b83c3

labo-flg: Oh... yes I see. We use resolveLink in the context of the workspace too. Seems like a side-effect of using same enum words 'dashboard' and 'investigation' for the workspace type. We just use the same function for both cases (entity type / workspace type). It's a bit misleading.

I'll reintroduce the lower-cased version with a proper comment.

labo-flg: Good catch, thanks !

SouadHadjiat (approved):

PR #5487, opened 2024-01-15
[frontend] i18n improvement

jpkha:
### Proposed changes

- Rename the t property of the translate object to t_i18n to make it easier to distinguish between various static translation values that we need.
- Rework the script so that all new translations are appended to the end of the en.json file.
- Translate missing translations.


SarahBocognano (approved):
Tested ✅  Everything works fine, two translation were missing, I know it's not based on your feature but if you want to add them 
Capture d'écran 2024-01-15 152919
Capture d'écran 2024-01-15 152927

jpkha:
@SarahBocognano Thank you for testing. Sure I will add it and check if I can improve the script

Issue #5437, opened 2024-01-09
Automation: the "+" dispaears if you delete 2 times an item


nino-filigran:
## Description

When you add a first component (Listen knowledge event), then add another components after it, then delete it, add it again, delete it again, the + disapears: the user cannot add steps in its automation anymore. 

To add more steps, the users needs to delete the first component and re-add it again
## Environment

Testing + Local (not exsiting in Demo)

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. create a playbook
2. Add the frist component
3. Click on the + to add the next component (whatever it is)
4. Delete the newly added component
5. Add again a component using the +
6. Delete again this component
7. the + has disapeared

## Expected Output

The + CTA remains

## Actual Output

The + CTA disapears.

labo-flg:
I noticed a display problem while working on this issue: the output ports are on top of each others.
I see in the code it's a static style computation. I'll found a way to distribute evenly the ports at the bottom.

![image](https://github.com/OpenCTI-Platform/opencti/assets/146674147/f6024769-8564-4aed-ae32-28bb20fcac90)

labo-flg:
Better now
![image](https://github.com/OpenCTI-Platform/opencti/assets/146674147/e415968a-6d2d-4885-8e39-e6cd25eb0e4b)

PR #5524, opened 2024-01-16
[frontend] fix output ports handling in Playbooks (#5437)

labo-flg:
### Proposed changes

* when deleting a node in the playbook, also delete the edge that lead to the deleted node (otherwise this "dead" node is kept in the list and impair subsequent calls to the delete algorithm)

### Related issues

* #5437

### Checklist

### Further comments

I noticed a UI problem when working on this issue : multiple output ports are poorly displayed, and on top of each other if multiple.
This PR brings an automatic, even distribution of the output ports at the bottom of the node.

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/data/playbooks/hooks/useManipulateComponents.js
+++ b/opencti-platform/opencti-front/src/private/components/data/playbooks/hooks/useManipulateComponents.js
@@ -301,7 +301,6 @@ const useManipulateComponents = (playbook, playbookComponents) => {
     let newNodes = getNodes();
     let newEdges = getEdges();
     const targetNode = getNode(selectedEdge.target);
-    // Add the new node
     const newNode = {
       id: result.nodeId,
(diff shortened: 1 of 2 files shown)


opencti-platform/opencti-front/src/private/components/data/playbooks/hooks/useManipulateComponents.js
```diff
@@ -523,8 +522,10 @@ const useManipulateComponents = (playbook, playbookComponents) => {
     setEdges(newEdges);
   };
   const applyDeleteNode = () => {
+    // start by removing the node and the edges that leads to it
     let newNodes = getNodes().filter((n) => n.id !== selectedNode.id);
-    let newEdges = getEdges();
+    let newEdges = getEdges().filter((e) => e.target !== selectedNode.id);
```
labo-flg, line 527: This is the source of the bug #5437

lndrtrbn (approved):

PR #5555, opened 2024-01-17
[frontend] Add playwright

jpkha:
### Proposed changes

* Add Playwright


opencti-platform/opencti-front/tests_e2e/common/login.ts
lndrtrbn, line 1: It feels strange to have assertions outside of an `it(...)`.

In my opinion this login function should have only the action to login. And by so you can move this function directly inside the LoginPage class. What do you think?

opencti-platform/opencti-front/tests_e2e/common/logout.ts
lndrtrbn, line 1: Same as login. I don't think we need this common folder as reusable stuff can be set inside the page models

opencti-platform/opencti-front/tests_e2e/fixtures/baseFixtures.ts
lndrtrbn, line 1: What is that for? To have coverage on our tests?

opencti-platform/opencti-front/tests_e2e/report/deleteReport.spec.ts
```diff
@@ -0,0 +1,21 @@
...
+
+test.skip('Create a new report page and delete it', async ({ page }) => {
+  const reportPage = new ReportPage(page);
+  const reportDetailsPage = new ReportDetailsPage(page);
+  await page.goto('/dashboard/analyses/reports');
+  await reportPage.addNewReport();
+  await reportPage.getReportNameInput().click();
+  await reportPage.getReportNameInput().fill('Test delete report e2e');
+  await reportPage.getCreateReportButton().click();
+  await page.waitForTimeout(1000);
```
lndrtrbn: That's strange we need that as the standard behavior of playright is by default to wait until it can do the action, and in the config we wrote to wait 20 seconds so it should be ok.

My first impression is because we need to await both the click AND the getCreateButton. I'll do some tests on my side locally.

opencti-platform/opencti-front/tests_e2e/report/deleteReport.spec.ts
```diff
@@ -0,0 +1,21 @@
...
+  //Issue on checked line
+  await page.getByRole('link', { name: 'Test delete report Test e2e' }).first().getByRole('checkbox').check();
+  await page.getByLabel('delete', { exact: true }).click();
+  await page.getByRole('button', { name: 'Launch' }).click();
+  await page.waitForTimeout(10000);
```
lndrtrbn: Same as previous comment here

opencti-platform/opencti-front/tests_e2e/report/updateReport.spec.ts
```diff
@@ -0,0 +1,20 @@
...
+  const reportForm = new ReportFormPage(page);
+  await page.goto('/dashboard/analyses/reports');
+  await reportPage.addNewReport();
+  await reportForm.fillNameInput('Test Update e2e');
+  await reportPage.getCreateReportButton().click();
+  await reportPage.getItemFromList( 'Test Update e2e Unknown - admin No' ).click();
+  await reportDetailsPage.getEditButton().click();
+  await reportForm.fillNameInput('Modification Test Update e2e');
+  await reportForm.getCloseButton().click();
+  await page.waitForTimeout(2000);
```
lndrtrbn: Same

opencti-platform/opencti-front/tests_e2e/report/deleteReport.spec.ts
jpkha: Will be remove, we will use the delete function that it did not need to wait async task

opencti-platform/opencti-front/tests_e2e/report/updateReport.spec.ts
jpkha: I will try to find a better solution.

opencti-platform/opencti-front/tests_e2e/fixtures/baseFixtures.ts
jpkha: Yes it's for having coverage but you can also use it to customize the workers that will be used by Playwright.

lndrtrbn (approved):

PR #5587, opened 2024-01-18
[backend] Fix the file change right

### LINKED ISSUE #5435: Impossible to see a file uploaded if we are not part of the platform organization
## Description

When I upload a file in the platform (from the global import panel or from a "Data" tab of an entity), the file isn't appear if I am not part of the platform organization.

## Environment

OCTI 5.12.15

## Reproducible Steps

1. Set up user A admin and part of Org A
2. Set up user B normal user and part of Org B
3. Set up Org A as platform organization
4. Create a Report with User B and upload a file from the "Data" tab of the Report
5. Try to see this file from User B (you can't), then from User A (you can)

The screenshot below illustrates this

## Expected Output

I should see the file with User B

## Screenshots

- On the left, my view from User A part of the platform organization and admin
- On the right, my view from User B (not admin), who create the report and upload the file

Screenshot 2024-01-09 092509

Kedae:
Closes #5435

### DIFF
--- a/opencti-platform/opencti-graphql/src/modules/internal/document/document-domain.ts
+++ b/opencti-platform/opencti-graphql/src/modules/internal/document/document-domain.ts
@@ -4,18 +4,20 @@ import moment from 'moment';
 import { generateFileIndexId } from '../../../schema/identifier';
 import { ENTITY_TYPE_INTERNAL_FILE } from '../../../schema/internalObject';
 import { elCount, elDeleteInstances, elIndex } from '../../../database/engine';
-import { INDEX_INTERNAL_OBJECTS, isEmptyField, READ_INDEX_INTERNAL_OBJECTS } from '../../../database/utils';
+import { INDEX_INTERNAL_OBJECTS, isEmptyField, isNotEmptyField, READ_INDEX_INTERNAL_OBJECTS } from '../../../database/utils';
 import { type EntityOptions, type FilterGroupWithNested, internalLoadById, listAllEntities, listEntitiesPaginated, storeLoadById } from '../../../database/middleware-loader';
 import type { AuthContext, AuthUser } from '../../../types/user';
 import { type DomainFindById } from '../../../domain/domainTypes';
 import type { BasicStoreEntityDocument } from './document-types';
-import type { BasicStoreCommon } from '../../../types/store';
+import type { BasicStoreCommon, BasicStoreObject } from '../../../types/store';
 import { type File, FilterMode, FilterOperator, OrderingMode } from '../../../generated/graphql';
 import { loadExportWorksAsProgressFiles } from '../../../domain/work';
-import { UnsupportedError } from '../../../config/errors';
 import { elSearchFiles } from '../../../database/file-search';
 import { SYSTEM_USER } from '../../../utils/access';
 import { FROM_START_STR } from '../../../utils/format';
+import { buildContextDataForFile, publishUserAction } from '../../../listener/UserActionListener';
+import type { UserAction } from '../../../listener/UserActionListener';
+import { ForbiddenAccess } from '../../../config/errors';
 
 export const getIndexFromDate = async (context: AuthContext) => {
   const searchOptions = {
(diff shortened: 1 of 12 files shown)


opencti-platform/opencti-graphql/src/modules/internal/document/document-domain.ts
```diff
@@ -119,12 +121,36 @@ export const allRemainingFilesCount = async (context: AuthContext, user: AuthUse
...
+    throw ForbiddenAccess('Access to this file is restricted', { id: entity_id, file: id });
+  }
+  return true;
+};
+
 // Get Files paginated with auto enrichment
 // Images metadata for users
 // In progress virtual files from export
-export const paginatedForPathsWithEnrichment = async (context: AuthContext, user: AuthUser, paths: string[], opts?: FilesOptions<BasicStoreEntityDocument>) => {
+export const paginatedForPathsWithEnrichment = async (context: AuthContext, user: AuthUser, path: string, entity_id: string, opts?: FilesOptions<BasicStoreEntityDocument>) => {
```
SouadHadjiat: if it's now singular "path" instead of "paths", maybe we can rename the method too ? **paginatedForPathWithEnrichment** or **paginatedFilesWithEnrichment** so it's not dependent on the parameters

opencti-platform/opencti-graphql/src/modules/internal/document/document-domain.ts
```diff
@@ -133,15 +159,14 @@ export const paginatedForPathsWithEnrichment = async (context: AuthContext, user
     orderOptions.orderMode = OrderingMode.Desc;
   }
   const listOptions = { ...opts, ...findOpts, ...orderOptions };
-  const pagination = await listEntitiesPaginated<BasicStoreEntityDocument>(context, user, [ENTITY_TYPE_INTERNAL_FILE], listOptions);
+
+  await checkFileAccess(context, user, entity_id, { entity_id, id: path, filename: '' });
```
SouadHadjiat: the third parameter of `checkFileAccess` is `scope` and here it's set with `entity_id`, doesn't seem right, no ?

Kedae: Well, it's the scope for the "UserAction" log, so I think it's a good call to have the entity_id to tell "User A tried to access files of this specific entity and cannot". But I'm open to other thoughts

SouadHadjiat: scopes should be a limited list of operations I think, but I don't know well the activity logging part. I can see this graph in a user activity analytics which is a distribution based on scope : 

image

It might not help to have ids in scope

Kedae: You're completely right, my bad I'll change that :)

SouadHadjiat: it's even defined in UserActionListener, all `event_scope` are typed with specific values

SouadHadjiat:
There is an issue here : I can see all analyst workbenches in a report (I should see only the report analyst workbenches, none here)
I can't reproduce the issue on testing platform.

image

opencti-platform/opencti-graphql/src/resolvers/stixCoreObject.js
```diff
@@ -120,13 +120,13 @@ const stixCoreObjectResolvers = {
     connectors: (stixCoreObject, { onlyAlive = false }, context) => connectorsForEnrichment(context, context.user, stixCoreObject.entity_type, onlyAlive),
     importFiles: (stixCoreObject, { first, prefixMimeType }, context) => {
       const opts = { first, prefixMimeTypes: prefixMimeType ? [prefixMimeType] : null, entity_id: stixCoreObject.id };
-      return paginatedForPathsWithEnrichment(context, context.user, [`import/${stixCoreObject.entity_type}/${stixCoreObject.id}`], opts);
+      return paginatedForPathWithEnrichment(context, context.user, `import/${stixCoreObject.entity_type}/${stixCoreObject.id}`, stixCoreObject.id, opts);
     },
     pendingFiles: (stixCoreObject, { first }, context) => {
-      return paginatedForPathsWithEnrichment(context, context.user, ['import/pending'], { first, entity_id: stixCoreObject.id });
+      return paginatedForPathWithEnrichment(context, context.user, 'import/pending', stixCoreObject.id, { first });
```
SouadHadjiat, line 126: if the entity_id is not part of the options, it won't be used as a filter to list only files linked to this entity_id

SouadHadjiat (changes requested):
Fix analyst workbench list displayed in an entity data tab (should be filtered by entity, it displays all analyst workbenches)

opencti-platform/opencti-graphql/src/modules/internal/document/document-domain.ts
```diff
@@ -119,29 +121,52 @@ export const allRemainingFilesCount = async (context: AuthContext, user: AuthUse
...
+    throw ForbiddenAccess('Access to this file is restricted', { id: entity_id, file: id });
+  }
+  return true;
+};
+
 // Get Files paginated with auto enrichment
 // Images metadata for users
 // In progress virtual files from export
-export const paginatedForPathsWithEnrichment = async (context: AuthContext, user: AuthUser, paths: string[], opts?: FilesOptions<BasicStoreEntityDocument>) => {
+export const paginatedForPathWithEnrichment = async (context: AuthContext, user: AuthUser, path: string, entity_id: string, opts?: FilesOptions<BasicStoreEntityDocument>) => {
```
SouadHadjiat, line 151: I don't really understand why `entity_id` has been extracted from opts as a separate parameter knowing that `FilesOptions` type still has `entity_id`. Why not keep entity_id inside opts and extract in the method ?

Kedae: It's a bit of semantic actually. We want to enforce that entity_id is a *important* prop from this function to get files and not just some stuff you can add in opts.
I understand your concern, we've discussed that with Julien and thougth it would be more explicit that way

SouadHadjiat (approved):
Tested locally and it seem to work well ✅

PR #5590, opened 2024-01-18
[frontend] Add indicator pattern filter

### LINKED ISSUE #4940: Add a "contains" / "do not contains" operator in new filter
## Use case

In new filters, add an operator "contains" allowing to filter on properties containing a specific string
In new filters, add an operator "do not contains" allowing to filter on properties that do not contains a specific string

jpkha:
Add search on pattern indicator


Archidoit:
The issue doesn't close 4940 ?

Archidoit (approved):

PR #5631, opened 2024-01-22
[backend] Fix file indexing infinite retry on parsing error (#5629)

### LINKED ISSUE #5629: [file indexing] infinit retry on parsing
## Description

The indexing process retry on error when elastic respond an error on parsing the file

## Environment

1. docker / kube
3. OpenCTI version: 5.12.20
4. OpenCTI client: frontend

## Reproducible Steps

1 start idexing
2 get error with elastic (don't know the exact cause can be no available heap related)

## Expected Output

do not infinite retry

## Actual Output

infinite retry.
this also restart the manager and update startup date of the manager.
get big logs because of the base 64 of the file send to the logs.

SouadHadjiat:
### Proposed changes

* catch error on file indexing and index file without content
* don't log file_data in case of error

### Related issues

* #5629 

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/database/engine.js
+++ b/opencti-platform/opencti-graphql/src/database/engine.js
@@ -2614,7 +2614,9 @@ export const elIndex = async (indexName, documentBody, opts = {}) => {
     indexParams = { ...indexParams, pipeline };
   }
   await engine.index(indexParams).catch((err) => {
-    throw DatabaseError('Simple indexing fail', { cause: err, body: documentBody });
+    // remove content (used by file indexing) from log
+    const documentBodyToLog = R.dissoc('file_data', documentBody);
+    throw DatabaseError('Simple indexing fail', { cause: err, body: documentBodyToLog });
   });
   return documentBody;
 };
--- a/opencti-platform/opencti-graphql/src/database/engine.js
+++ b/opencti-platform/opencti-graphql/src/database/engine.js
@@ -2614,9 +2614,7 @@ export const elIndex = async (indexName, documentBody, opts = {}) => {
     indexParams = { ...indexParams, pipeline };
   }
   await engine.index(indexParams).catch((err) => {
-    // remove content (used by file indexing) from log
-    const documentBodyToLog = R.dissoc('file_data', documentBody);
-    throw DatabaseError('Simple indexing fail', { cause: err, body: documentBodyToLog });
+    throw DatabaseError('Simple indexing fail', { cause: err, body: documentBody });
   });
   return documentBody;
 };
--- a/opencti-platform/opencti-graphql/src/database/file-search.js
+++ b/opencti-platform/opencti-graphql/src/database/file-search.js
@@ -64,11 +64,11 @@ export const elIndexFiles = async (context, user, files) => {
       const documentBody = buildIndexFileBody(internal_id, fileObject, entity);
       await elIndex(INDEX_FILES, documentBody, { pipeline: 'attachment' }).catch((err) => {
         // catch & log error
-        logApp.error(err);
-        // try to index without content
         const documentWithoutFileData = R.dissoc('file_data', documentBody);
+        logApp.error('Error on file indexing', { message: err.message, causeStack: err.data?.cause?.stack, stack: err.stack, body: documentWithoutFileData });
+        // try to index without file content
         elIndex(INDEX_FILES, documentWithoutFileData).catch((e) => {
-          logApp.error(e);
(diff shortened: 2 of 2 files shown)


opencti-platform/opencti-graphql/src/database/engine.js
```diff
@@ -2614,7 +2614,9 @@ export const elIndex = async (indexName, documentBody, opts = {}) => {
     indexParams = { ...indexParams, pipeline };
   }
   await engine.index(indexParams).catch((err) => {
-    throw DatabaseError('Simple indexing fail', { cause: err, body: documentBody });
+    // remove content (used by file indexing) from log
+    const documentBodyToLog = R.dissoc('file_data', documentBody);
```
richard-julien: I would prefer to not change that as 99% of the time getting the body in the error log is really interesting.
The removal should appears on the caller in file-search.js

opencti-platform/opencti-graphql/src/database/file-search.js
```diff
@@ -62,7 +62,19 @@ export const elIndexFiles = async (context, user, files) => {
...
       };
       const documentBody = buildIndexFileBody(internal_id, fileObject, entity);
-      await elIndex(INDEX_FILES, documentBody, { pipeline: 'attachment' });
+      let success = true;
+      await elIndex(INDEX_FILES, documentBody, { pipeline: 'attachment' }).catch((err) => {
+        // catch & log error
+        logApp.error(err);
+        success = false;
+      });
+      if (!success) {
```
richard-julien: Why not directly in the catch?

opencti-platform/opencti-graphql/src/database/engine.js
SouadHadjiat: I'm just removing "file_data" which is the base64 content of the file, and we still log the document body

opencti-platform/opencti-graphql/src/database/file-search.js
SouadHadjiat: I tried but I had an error because it's async, I'm not sure it would work if I call just `elIndex(INDEX_FILES, documentWithoutFileData);` without await

richard-julien: Oh yes, i think in this case its more usual to use a catch block instead of the promise catch

opencti-platform/opencti-graphql/src/database/engine.js
richard-julien: For me its a bit strange to do a dissoc operation when 99.99% of index operations are not impacted by this field. Its more the responsibility of  the caller to log accordingly to his need

SouadHadjiat: What do you suggest exactly ? I don't understand how the caller would specify what we log in this method.

richard-julien: Something like

```
try {
   await elIndex(INDEX_FILES, documentBody, { pipeline: 'attachment' })
} catch (err) {
   logApp.error('Error on document indexing', { message: err.message } );
   const documentWithoutFileData = R.dissoc('file_data', documentBody);
   await elIndex(INDEX_FILES, documentWithoutFileData).catch((e) => {
     logApp.error('Error in fallback document indexing', { message: err.message } );
   });
}
```

SouadHadjiat: not so easy it seems, I don't want to loose the cause of the error (which I would if I use your suggestion as it is) and I want to log also the document body without the file_data.

If I write this : 

```        
const documentWithoutFileData = R.dissoc('file_data', documentBody);
logApp.error('Error on file indexing', { message: err.message, cause: err.data?.cause, body: documentWithoutFileData });
```
I still see the document body with file_data in the log

SouadHadjiat: ah this seem to work finally : 

`logApp.error('Error on file indexing', { message: err.message, stack: err.stack, body: documentWithoutFileData });`

SouadHadjiat: but I loose the cause stack which is not good :(

opencti-platform/opencti-graphql/src/database/file-search.js
SouadHadjiat: it works inside the promise catch

SouadHadjiat:
PR status : waiting for @richard-julien test and approval

richard-julien (approved):

PR #5633, opened 2024-01-22
[backend] manage long text filter like description and add search operator #4940

### LINKED ISSUE #4940: Add a "contains" / "do not contains" operator in new filter

jpkha:
#4940


jpkha:
`starts_with` and `ends_with` work like `contains `. The nil operator does not work for an empty string `""`, for example. We should consider it as empty, but Elasticsearch checks if the field exists only.

jpkha:
!missing test to do

opencti-platform/opencti-graphql/tests/02-integration/01-database/filterGroup-test.js
```diff
@@ -1226,6 +1226,115 @@ describe('Complex filters combinations for elastic queries', () => {
...
     expect(queryResult.errors.length).toEqual(1);
   });
+  it('should list entities according to search filters with trunc word', async () => {
+    const queryResult = await queryAsAdmin({ query: REPORT_LIST_QUERY,
+      variables: {
+        first: 25,
+        filters: {
+          mode: 'and',
+          filters: [{
+            key: ['description'],
```
Archidoit: array not required, in the front, we rather write: "key: 'description'"

opencti-platform/opencti-graphql/tests/02-integration/01-database/filterGroup-test.js
```diff
@@ -1226,6 +1226,115 @@ describe('Complex filters combinations for elastic queries', () => {
...
+      } });
+    expect(queryResult.data.reports.edges.length).toEqual(1);
+
+    queryResult = await queryAsAdmin({ query: REPORT_LIST_QUERY,
+      variables: {
+        first: 25,
+        filters: {
+          mode: 'and',
+          filters: [{
+            key: ['description'],
```
Archidoit: same here and below

opencti-platform/opencti-front/src/utils/filters/filtersUtils.tsx
```diff
@@ -686,6 +692,10 @@ export const getAvailableOperatorForFilterKey = (filterKey: string): string[] =>
     return ['eq', 'not_eq', 'nil', 'not_nil', 'contains', 'not_contains',
       'starts_with', 'not_starts_with', 'ends_with', 'not_ends_with'];
   }
+  if (longTextFilters.includes(filterKey)) {
+    return ['eq', 'not_eq', 'nil', 'not_nil', 'search'];
```
Archidoit: remove 'nil' and 'not_nil' (not working for the moment for long string)

Archidoit: same for eq and not_eq

opencti-platform/opencti-graphql/tests/02-integration/01-database/filterGroup-test.js
```diff
@@ -1226,6 +1226,81 @@ describe('Complex filters combinations for elastic queries', () => {
...
+            values: ['"for testing purposes"'],
+            operator: 'search',
+            mode: 'or',
+          }],
+          filterGroups: [],
+        },
+      } });
+    expect(queryResult.data.reports.edges.length).toEqual(1);
+  });
+
```
Archidoit, line 1303: maybe add a test for
- a query with a filter like this: operator=search, key=description, values=[a_word_in_no_description] and we should find no entity 
- same query with 2 words in values: values=[a wo_word_in_no_description   a_word_in_an_entity_description] and we should find an entity
- a query with operator=search but on a short string to check search also works for short strings

Archidoit:
eq and not_eq operators not working for long string

opencti-platform/opencti-graphql/tests/02-integration/01-database/filterGroup-test.js
```diff
@@ -1226,6 +1226,229 @@ describe('Complex filters combinations for elastic queries', () => {
...
+  it('should find entities according to search filters on short string unorder', async () => {
```
Goumies: Can you complete the test title with "unordered", if that was the intent ?

Goumies:
I have noticed that the `description` and `name` options in the `Add filter` dropdown menu are the only ones in lowercase.
Can you capitalize it?

image

opencti-platform/opencti-graphql/tests/02-integration/01-database/filterGroup-test.js
jpkha: Yes sure. I put unorder but I think I am not grammatically correct sorry.

jpkha:
> I have noticed that the `description` and `name` options in the `Add filter` dropdown menu are the only ones in lowercase. Can you capitalize it?
> 
> image

I fixed it from the translation. Looks like in the other language we already put an upper case by default.

Archidoit:
> > I have noticed that the `description` and `name` options in the `Add filter` dropdown menu are the only ones in lowercase. Can you capitalize it?
> > image
> 
> I fixed it from the translation. Looks like in the other language we already put an upper case by default.

It will be fixed in 4939 where we fetch labels from the back anyway. It's not so good to translate something without a first letter in upper case with something in upper case I think.

opencti-platform/opencti-front/lang/front/en.json
```diff
@@ -2163,7 +2163,7 @@
   "Level": "Level",
   "color": "color",
   "Force now": "Force now",
-  "name": "name",
+  "name": "Name",
```
Archidoit: I don't think name should be translated to Name because it changes the case. (Knowing that the 'name' and 'description' label will be ok in 4939

Archidoit (approved):

PR #5640, opened 2024-01-22
[backend] load creators from cache to improve perf

SouadHadjiat:
### Proposed changes

* load directly users from cache when querying for creators
* make sure we only return needed informations from users

### Related issues

* improve ES performance 
* Try to fix MaxListenersExceededWarning which traced to batchCreators

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/domain/user.js
+++ b/opencti-platform/opencti-graphql/src/domain/user.js
@@ -15,8 +15,8 @@ import {
   PLATFORM_VERSION
 } from '../config/conf';
 import { AuthenticationFailure, DatabaseError, ForbiddenAccess, FunctionalError, UnsupportedError } from '../config/errors';
-import { getEntitiesListFromCache, getEntityFromCache } from '../database/cache';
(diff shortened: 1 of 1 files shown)


opencti-platform/opencti-graphql/src/domain/user.js
```diff
@@ -198,19 +198,28 @@ export const batchGroups = async (context, user, userId, opts = {}) => {
   return batchListThroughGetTo(context, user, userId, RELATION_MEMBER_OF, ENTITY_TYPE_GROUP, opts);
 };
 
-const internalUserIds = Object.keys(INTERNAL_USERS);
+// build only a creator object with what we need to expose of users
+const buildCreatorUser = (user) => {
+  if (!user) {
+    return user;
+  }
```
marieflorescontact, line 205: is it ok to return user if `!user` ?

SouadHadjiat: yes, I understand the question, there are different possibilities here, I could have returned null, or false, but I thought that since it's a method that transforms the user to extract only the fields we need, if the user is not defined, I just return the user without change.

lndrtrbn (approved):

PR #5642, opened 2024-01-22
[backend] Improvement to limit engine memory usage

### LINKED ISSUE #5701: [backend] Improvement to limit engine memory usage
[backend] Improvement to limit engine memory usage

richard-julien:
The idea is to limit the memory footprint of the engine query.
First round before [oob/batch_list_through_refactor](https://github.com/OpenCTI-Platform/opencti/tree/oob/batch_list_through_refactor)

- Limit the pagination size to 500 elements
- Limit the window results of id resolution to minimum
- Infer the indices to read from the types if possible
- Remove deprecated ignore_throttled option


opencti-platform/opencti-graphql/src/database/engine.js
```diff
@@ -1006,7 +1012,7 @@ export const RUNTIME_ATTRIBUTES = {
       // eslint-disable-next-line no-use-before-define
       const users = await elPaginate(context, user, READ_INDEX_INTERNAL_OBJECTS, {
```
SouadHadjiat: could we load users from cache too here ?

opencti-platform/opencti-graphql/src/database/engine.js
```diff
@@ -1232,41 +1231,24 @@ export const elFindByIds = async (context, user, ids, opts = {}) => {
...
+        minimum_should_match: 1,
+      },
+    };
+    mustTerms.push(should);
     if (types && types.length > 0) {
       const typesShould = types.map((typeShould) => (
         [
-          { match_phrase: { 'entity_type.keyword': typeShould } },
-          { match_phrase: { 'parent_types.keyword': typeShould } }
+          { term: { 'entity_type.keyword': typeShould } },
```
SouadHadjiat: here we could also do a terms query with the list of types instead of a list of should with one term.

richard-julien: Updated

opencti-platform/opencti-graphql/src/database/engine.js
richard-julien: Updated

SouadHadjiat:
I tested locally and didn't find any bug related to this PR, but I'm not sure I tested all the cases. 
It seems already more performant even with my small db.

SouadHadjiat (approved):

PR #5659, opened 2024-01-23
[frontend/backend] optional user confidence level (create/edit)

SarahBocognano:
To be review internally


labo-flg:
Tested and review ok for merge in feature branch

Issue #5591, opened 2024-01-18
Authentication header provider does not support HTTP/1.1 style headers with capital letters


sbocahu:
## Description

#5107 followup: header provider does not support HTTP/1.1 style headers with capital letters. It seems to be related to the fact that HTTP 2.0 spec states that the headers shall be lowercase, and some library rewrite headers to this new spec (even on HTTP/1.1)

## Reproducible Steps

(q being the RootPrivateQuery graphql query)

with PROVIDERS__HEADER__CONFIG__HEADER_EMAIL: "auth-email-address":

```
# HTTP/2.0 negociation, curl rewrites heades to lowercase
$ curl -s  -X POST -H 'Content-Type: application/json'  -d @q 'https://myopenctiinstance/graphql' -H 'Auth-Email-Address: myname@mycompany.io' | jq -r '.data.me.id'
c41f983e-4ea0-48e7-8d37-c98507a4d7b5
```
```
# HTTP/1.1 negociation, curl DOES NOT rewrite headers to lowercase
$ curl -s  -X POST -H 'Content-Type: application/json'  -d @q 'https://myopenctiinstance/graphql' -H 'Auth-Email-Address: myname@mycompany.io' | jq -r '.data.me.id'
c41f983e-4ea0-48e7-8d37-c98507a4d7b5
```

with PROVIDERS__HEADER__CONFIG__HEADER_EMAIL: "Auth-Email-Address":

```
$ curl --http1.1 -s  -X POST -H 'Content-Type: application/json'  -d @q 'https://myopenctiinstance/graphql' -H 'Auth-Email-Address: myname@mycompany.io' | jq -r '.errors[0].message'
You must be logged in to do this.
```

## Expected Output

I expect to be able to use HTTP/1.1 style headers. At least, the documentation should states if this format is not supported.

richard-julien:
Will be covered by https://github.com/OpenCTI-Platform/opencti/issues/5434

PR #5664, opened 2024-01-23
Fix respecting case of groups and organizations in header provider

sbocahu:
### Proposed changes

* Respect case of headers for groups & organizations mapping of users automaticaly authenticated by proxy headers (authentication 'header' provider)

### Related issues

* #5591 

### Checklist

### Further comments

untested; however it seems to be straightforward


richard-julien:
Thanks be already fix by https://github.com/OpenCTI-Platform/opencti/commit/c8125ab28ffea70cb3d4d623bb6a698f23a0200f

PR #5667, opened 2024-01-24
Filters display in Report Timeline view (#5636)

### LINKED ISSUE #5636: Filters display in Report Timeline view
## Description

In the timeline view of a report, the filters are not correctly displayed (desing).

## Reproducible Steps
Go to a report
Go to the Knowledge tab
Select the 'timeline view'
Add filters via the filters button of the bottom bar
The filters are added next to the button.

Problems:
- the filters are not vertically aligned with the middle of the bottom bar (there are too low)
- if there are too many filters, the end is not displayed (in the capture below, there is a filter after the 'and' and we don't see it) -> either add a tooltip to display the missing filters, either display all the filters (several lines ...)

image

CelineSebe:
### Proposed changes

* I suggest adding the ability to expand the bar when the number of filters becomes significant. This would provide a better user experience.

### Related issues

* #5636

### Checklist

### Further comments

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/common/containers/ContainertKnowledgeTimeLineBar.jsx
+++ b/opencti-platform/opencti-front/src/private/components/common/containers/ContainertKnowledgeTimeLineBar.jsx
@@ -5,7 +5,6 @@ import Tooltip from '@mui/material/Tooltip';
 import Drawer from '@mui/material/Drawer';
 import Divider from '@mui/material/Divider';
 import { makeStyles } from '@mui/styles';
-import KeyboardArrowDownIcon from '@mui/icons-material/KeyboardArrowDown';
 import SearchInput from '../../../../components/SearchInput';
 import { useFormatter } from '../../../../components/i18n';
 import { MESSAGING$ } from '../../../../relay/environment';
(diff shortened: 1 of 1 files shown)


marieflorescontact (commented):
I find it a little odd not to display all the filters if you have enough space to do so.
IMO we should choose between tooltip and the ability to expand the bar.
Capture d'écran 2024-01-26 174946

SarahBocognano:
I agree with @marieflorescontact, also can you maybe just move the 'close' button because I think it's kind of overlap
Capture d'écran 2024-01-29 105217

CelineSebe:
It's a quick fix to display all the filters that were hidden before, only on this screen. The filter team is currently changing the display of each filter.

marieflorescontact (approved):

CelineSebe:
@marieflorescontact @SarahBocognano I reviewed this issue with JP. This problem exists on all screens where we have filters. However, he thinks that the chevron is not necessary. Thank you for your feedback.

CelineSebe:
@SarahBocognano @marieflorescontact I have implemented some changes, could you please provide a new review for these updates?

PR #5684, opened 2024-01-25
[frontend/backend] CSV Mapper default values for marking definitions (#4932)

### LINKED ISSUE #4932: Add an optional default value for an attribute that is missing mapping value from a file

lndrtrbn:
### Proposed changes

* Manage default values for marking definitions:
  * If markings in csv file use it,
  * If no markings in csv file:
    * If no default policy both in settings and mapper: no marking set,
    * If default policy in settings active: use the default markings of the user importing the file,
    * If mapper policy on "user defaults": use the default markings of the user importing the file,
    * If mapper  policy on "user choice": use the markings chosen by the user when importing the file.

### Related issues

* #4932 

### Checklist

### Further comments

In frontend code:

* Add an input in import file dialog to allow user to chose markings to set by default on imported data,
* Create a new component to manage default markings policy in csv mapper.

In backend code:

* Refacto how default markings are managed by the function `fillDefaultValues`,
* Manage markings in csv mapper (need to separate the logic from other attributes to fetch them correctly),
* Add automatic tests on csv mapper for default markings


opencti-platform/opencti-front/src/private/components/data/csvMapper/representations/attributes/CsvMapperDefaultMarking.tsx
```diff
@@ -0,0 +1,42 @@
...
+        displayEmpty
+      >
+        <MenuItem value="user-choice">
+          {t_i18n('Let the user choose marking definitions')}
+        </MenuItem>
+        <MenuItem value="user-default">
+          {t_i18n('Use default marking definitions of the user')}
+        </MenuItem>
+      </Field>
+      <Tooltip title='pouet'>
```
aHenryJard: Title looks like a leftover... 'pouet' ?

opencti-platform/opencti-front/src/private/components/common/files/FileManager.jsx
```diff
@@ -323,6 +336,19 @@ const FileManager = ({
...
                   )}
+                  {selectedConnector?.connector_scope?.includes('text/csv')
+                    && (
+                      <>
+                        <ObjectMarkingField
+                          name="objectMarking"
+                          style={fieldSpacingContainerStyle}
+                        />
+                        <DialogContentText>
+                          {t('Marking definitions to use by the csv mapper...')}
```
aHenryJard, line 347: question: shouldn't we use a longer name function for text now ?

opencti-platform/opencti-graphql/src/parser/csv-mapper.ts
```diff
@@ -227,21 +230,68 @@ const handleAttributes = (
...
+  chosenMarkings: string[],
+  user: AuthUser,
+) => {
+  if (input[INPUT_MARKINGS]) {
+    return;
+  }
+
+  // Find default markings policy in entity settings ("true" or undefined).
+  const settingAttributes = entitySetting ? getAttributesConfiguration(entitySetting) : undefined;
+  // console.log('pouet', representation.target.entity_type, settingAttributes);
```
aHenryJard, line 258: this looks like a left over

opencti-platform/opencti-graphql/src/parser/csv-mapper.ts
```diff
@@ -227,21 +230,68 @@ const handleAttributes = (
...
+    ?.find((attribute) => attribute.name === INPUT_MARKINGS)
+    ?.default_values?.[0];
+  // Find default markings policy in mapper representation ("user-default" or "user-choice"  or undefined).
+  const representationMarkingValue = representation.attributes
+    .find((attribute) => attribute.key === INPUT_MARKINGS)
+    ?.default_values?.[0];
+
+  // Retrieve default markings of the user.
+  const userDefaultMarkings = (user.default_marking ?? [])
+    .find((entry) => entry.entity_type === 'GLOBAL')
```
aHenryJard, line 268: I wonder if we don't already have a constant somewhere for this 'GLOBAL' string.

opencti-platform/opencti-graphql/src/parser/csv-mapper.ts
aHenryJard:
```suggestion
```
this looks like a left over

opencti-platform/opencti-front/lang/en.json
```diff
@@ -2550,5 +2550,8 @@
   "Settings default values": "Default value from entity settings: {value}",
   "Settings default values usage...": "If the data is not found in the CSV file then the default value set in Settings > Customization will be used instead. You can override this default value by filling the input above.",
   "No default value set in Settings...": "No default value set for this attribute in the entity settings. You can provide one with the input above. This value will be used if the data is missing in the CSV file.",
-  "Default values set": "Default values set"
+  "Default values set": "Default values set",
+  "Let the user choose marking definitions": "Let the user choose marking definitions",
+  "Use default marking definitions of the user": "Use default marking definitions of the user",
+  "Marking definitions to use by the csv mapper...": "Marking definitions to use by the csv mapper if some marking attributes have the policy set to: let the user choose the marking definitions."
```
aHenryJard: I found this text misleading, even more if you read too fast like I did
image

I can try to propose another one:
```suggestion
  "Marking definitions to use by the csv mapper...": "Marking definitions used by the csv mapper only when some marking attributes have policy set to 'let the user choose the marking definitions.' If not this Marking definitions is ignored."
```
Is it possible to use bold for 'only ' or 'if'  whatever the texts is at the end ?

Maybe mine is too long :)

opencti-platform/opencti-front/src/private/components/common/files/FileManager.jsx
lndrtrbn: In this component the translation function is taken as a prop instead of using the hook `useTranslate` (because it is a class component so it cannot use hooks), a prop called `t`. The way this component is called it is not trivial to change this name. And I don't think it should eb done in this PR but yes you are right that if possible we should use `t_i18n` as name of the translation function

opencti-platform/opencti-graphql/src/parser/csv-mapper.ts
lndrtrbn: I can't find one when looking into the backend code, but could be nice to create one

aHenryJard (approved):
Last version of this branch tested locally (using Indicator mapper) working all fine as expected with markings.

PR #5691, opened 2024-01-25
Change the way TAXII ingester paginates requests

### LINKED ISSUE #5681: Problem in the way the TAXII ingester paginates requests
There seems to be a problem in the way the TAXII ingester paginates requests.
Indeed, it seems to paginate on both the added_after date AND the next field.

For example, I have a collection that contains 500+ STIX elements.
The first requests OpenCTI makes is the following:

http://taxii/base/collections/719d8ee6-7359-11ee-a839-046c59cc782e/objects

The TAXII server responds with 50 objects, its default limit, and the following additional information:

The next field in the response envelope is set to 50, as we have additional objects to return
The x-taxii-date-added-first header, with value 2023-09-08T07:45:34.941571+00:00
The x-taxii-date-added-last header, with value 2024-01-17T16:11:20.183201+00:00
OpenCTI does, on the next ingestion, the following request:

http://taxii/base/collections/719d8ee6-7359-11ee-a839-046c59cc782e/objects?added_after=2024-01-17T16:11:20.183201+00:00&next=50

This request is incorrect, and should either be:
http://taxii/base/collections/719d8ee6-7359-11ee-a839-046c59cc782e/objects?added_after=2024-01-17T16:11:20.183201+00:00
if we are paginating by date (although this is not recommended, as there is the possibility that we have 50+ objects with the exact same date), or:

http://taxii/base/collections/719d8ee6-7359-11ee-a839-046c59cc782e/objects?limit=50
if we are paginating using the next cursor.

In fact, the TAXII documentation clearly presents the two options (here: https://docs.oasis-open.org/cti/taxii/v2.1/os/taxii-v2.1-os.html#_Toc31107519),  and states (emphase is mine):

If the more property is set to true and the next property is populated then the client can paginate through the remaining records using the next URL parameter along with the same original query options.

The current implementation will cause OpenCTI to skip STIX objects when importing from a TAXIIv2 server that sends a next parameter.

yassine-ouaamou:
### Proposed changes

Extracted from https://docs.oasis-open.org/cti/taxii/v2.1/os/taxii-v2.1-os.html#_Toc31107519
*  If the more property is set to true and the next property is populated then the client can paginate through the remaining records using the next URL parameter along with the same original query options.

* If the more property is set to true and the next property is empty then the client may paginate through the remaining records by using the added_after URL parameter with the date/time value from the X-TAXII-Date-Added-Last header along with the same original query options.

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/5681
*

### Checklist

### Further comments

### DIFF
--- a/opencti-platform/opencti-graphql/src/manager/ingestionManager.ts
+++ b/opencti-platform/opencti-graphql/src/manager/ingestionManager.ts
@@ -218,8 +218,13 @@ const taxiiHttpGet = async (ingestion: BasicStoreEntityIngestionTaxii): Promise<
   const httpClient = getHttpClient(httpClientOptions);
   const preparedUri = ingestion.uri.endsWith('/') ? ingestion.uri : `${ingestion.uri}/`;
   const url = `${preparedUri}collections/${ingestion.collection}/objects/`;
-  const next = ingestion.added_after_start ? ingestion.current_state_cursor : null;
-  const params = { next, added_after: ingestion.added_after_start };
+  const next = ingestion.current_state_cursor;
+  // https://docs.oasis-open.org/cti/taxii/v2.1/os/taxii-v2.1-os.html#_Toc31107519
+  // If the more property is set to true and the next property is populated then the client can paginate through the remaining records using the next URL parameter along with the
+  // same original query options.
+  // If the more property is set to true and the next property is empty then the client may paginate through the remaining records by using the added_after URL parameter with the
+  // date/time value from the X-TAXII-Date-Added-Last header along with the same original query options.
+  const params = next ? { next } : { added_after: ingestion.added_after_start };
   const { data, headers: resultHeaders } = await httpClient.get(url, { params });
   return { data, addedLast: resultHeaders['x-taxii-date-added-last'] };
 };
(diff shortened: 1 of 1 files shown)


opencti-platform/opencti-graphql/src/manager/ingestionManager.ts
```diff
@@ -218,8 +218,13 @@ const taxiiHttpGet = async (ingestion: BasicStoreEntityIngestionTaxii): Promise<
...
   const url = `${preparedUri}collections/${ingestion.collection}/objects/`;
-  const next = ingestion.added_after_start ? ingestion.current_state_cursor : null;
-  const params = { next, added_after: ingestion.added_after_start };
+  const next = ingestion.current_state_cursor;
+  // https://docs.oasis-open.org/cti/taxii/v2.1/os/taxii-v2.1-os.html#_Toc31107519
+  // If the more property is set to true and the next property is populated then the client can paginate through the remaining records using the next URL parameter along with the
+  // same original query options.
+  // If the more property is set to true and the next property is empty then the client may paginate through the remaining records by using the added_after URL parameter with the
+  // date/time value from the X-TAXII-Date-Added-Last header along with the same original query options.
+  const params = next ? { next } : { added_after: ingestion.added_after_start };
```
richard-julien: I dont think it will works.

the added_after must always be part of the params. The problem of the ticket is that date must remains the same.

yassine-ouaamou:
Changed the code as we said @richard-julien . We did some tests with @SouadHadjiat and we were able to retrieve all the pages

richard-julien (approved):

Issue #3154, opened 2023-04-13
Expand name in "progress works panel"


SYNchroACK:
## Use case

I would like to have in the name the following string: `Reports from 2020-01-01T00:00:00Z to 2022-12-31T23:59:59Z`
![image](https://user-images.githubusercontent.com/4184292/231793656-035f1444-9313-4e40-a9d4-009c50adf142.png)

#### If the feature request is approved, would you be willing to submit a PR?

No

SamuelHassine:
Increase the limit of the name characters and a tooltip with mouseover to display everything.

PR #5707, opened 2024-01-26
[frontend] improve the readability of connector names (#3154)

frapuks:
### Proposed changes

Improve the readability of connector names.
- Display fullname (before : only 40 characters).

### Related issues

issue #3154

### Checklist

- [x]  I consider the submitted work as finished
- [x]  I tested the code for its functionality

### Further comments

- I had to recreate sub-types that had not been individually generated by relay, with potentially null management.
If you'd like more information, or if you've already encountered this problem and have a better solution, let me know.


opencti-platform/opencti-front/src/private/components/data/connectors/ConnectorWorks.tsx
```diff
@@ -0,0 +1,388 @@
...
+}
+
+interface ConnectorWorksComponentProps {
+  data: ConnectorWorks_data$data
+  options: Options[]
+  relay: RelayRefetchProp
+}
+
+const ConnectorWorksComponent: FunctionComponent<ConnectorWorksComponentProps> = ({ data, options, relay }) => {
+  // const works: NonNullable<ConnectorWorks_data$data['works']>['edges'] = pathOr([], ['works', 'edges'], data);
```
lndrtrbn: To remove ?

opencti-platform/opencti-front/src/private/components/data/connectors/ConnectorWorks.tsx
```diff
@@ -0,0 +1,388 @@
...
+                  <Grid item={true} xs={12}>
+                    <Typography variant="h3" gutterBottom={true}>
+                      {t_i18n('Progress')}
+                    </Typography>
+                    <LinearProgress
+                      classes={{ root: classes.progress }}
+                      variant="determinate"
+                      value={
+                        // eslint-disable-next-line no-nested-ternary
+                        tracking && tracking.import_expected_number !== null && tracking.import_processed_number !== null
```
lndrtrbn: I know it's not from your code but there is typescript errors here. because we check only null value but it can be also undefined

opencti-platform/opencti-front/src/private/components/data/connectors/ConnectorWorks.tsx
```diff
@@ -0,0 +1,388 @@
...
+              <Grid item={true} xs={5}>
+                <Grid container={true} spacing={1}>
+                  <Grid item={true} xs={6}>
+                    <Typography variant="h3" gutterBottom={true}>
+                      {t_i18n('Name')}
+                    </Typography>
+                    <Tooltip title={work.name}>
+                      <Typography>
+                        {truncate(work.name, 80)}
+                      </Typography>
```
lndrtrbn:
```suggestion
                      <Typography sx={{ overflowX: 'hidden', textOverflow: 'ellipsis' }}>
                        {work.name}
                      </Typography>
```

We can use pure CSS to fix this issue without having to truncate the text arbitrarily

lndrtrbn (dismissed):

opencti-platform/opencti-front/src/private/components/data/connectors/ConnectorWorks.tsx
frapuks: my bad, i forgot this.

opencti-platform/opencti-front/src/private/components/data/connectors/ConnectorWorks.tsx
labo-flg: this issue here is to check for the existence of the field, but as it's a number we shall not fail the test on 0 value.

Maybe use ` tracking && Number.isFinite(tracking.import_expected_number) && Number.isFinite(tracking.import_processed_number) && `

Number.isFinite would return false on undefined and null, but not 0

opencti-platform/opencti-front/src/private/components/data/connectors/ConnectorWorks.tsx
frapuks: Nice, and adding `whiteSpace: 'nowrap'` to display name in only 1 line to prevent too long names

lndrtrbn: I'm not sure with this one. IMO there is no issue with the fact that the name is on multiple lines if it is long. I personally  prefer having the complete name on two lines rather than loosing information

opencti-platform/opencti-front/src/private/components/data/connectors/ConnectorWorks.tsx
frapuks: image
Apparently Number.isFinite doesn't prevent errors.

Here, variables are of type `number | null`, and cannot be undefined.

opencti-platform/opencti-front/src/private/components/data/connectors/ConnectorWorks.tsx
frapuks: it's relevant. In the Acceptance Criteria, they ask for a tooltip in the name hover. So I thought there was a logic in not allowing the possibility of displaying a whole name. I'll check this with Nino

frapuks: I've agreed with him to display the full name directly, removing the tooltip which is no longer useful. Thanks for the feedback

opencti-platform/opencti-front/src/private/components/data/connectors/ConnectorWorks.tsx
labo-flg: in fact using `!!tracking.import_expected_number && !!tracking.import_processed_number` might be sufficient in this case, because the fallback is `: 0` anyway, wichi is the result of the computation in any case.

frapuks: I tried with `!!`, and removed the 2nd check in the ternary operator since it was already solved by the first. It seems to work correctly with `null` and `undefined`, returning 0 in all cases.

opencti-platform/opencti-front/src/private/components/data/connectors/ConnectorWorks.tsx
```diff
@@ -0,0 +1,376 @@
...
+  return (
+    <div>
+      {works.length === 0 && (
+        <Paper
+          classes={{ root: classes.paper }}
+          variant="outlined"
+          style={{ marginBottom: 20 }}
+        >
+          <div
+            style={{
```
jpkha: It's better to use a `<table>` instead of a `<div>` with a `display: 'table'` 

what do you think ?

opencti-platform/opencti-front/src/private/components/data/connectors/ConnectorWorks.tsx
```diff
@@ -0,0 +1,376 @@
...
+    return () => subscription.unsubscribe();
+  }, []);
+
+  return (
+    <div>
+      {works.length === 0 && (
+        <Paper
+          classes={{ root: classes.paper }}
+          variant="outlined"
+          style={{ marginBottom: 20 }}
```
jpkha: Maybe we could add the `marginBottom` directly to the class `paper` instead add an inline style like this..

opencti-platform/opencti-front/src/private/components/data/connectors/ConnectorWorks.tsx
```diff
@@ -0,0 +1,376 @@
...
+                {work.errors?.length} {t_i18n('errors')}
+              </Button>
+              <Button
+                variant="outlined"
+                style={{ position: 'absolute', right: 10, bottom: 10 }}
+                onClick={() => handleDeleteWork(work.id)}
+                size="small"
+              >
+                <Delete fontSize="small" />
+                &nbsp;&nbsp;{t_i18n('Delete')}
```
jpkha: Can we delete it ? `&nbsp;&nbsp;`

frapuks: Obviously. it was used to put a space between the icon and the text in the button. I replace it with the `startIcon` property of the button

opencti-platform/opencti-front/src/private/components/data/connectors/ConnectorWorks.tsx
frapuks: That's true, especially as `margin` property is already in the class.

opencti-platform/opencti-front/src/private/components/data/connectors/ConnectorWorks.tsx
frapuks: I'd prefer to simply remove the table, because it's only used here to display the centred `no work` text, and replace it with a Typography component

opencti-platform/opencti-front/src/private/components/data/connectors/ConnectorWorks.tsx
```diff
@@ -0,0 +1,360 @@
...
+    setDisplayErrors(true);
+    setErrors(errorsList);
+  };
+
+  const handleCloseErrors = () => {
+    setDisplayErrors(false);
+    setErrors([]);
+  };
+
+  // eslint-disable-next-line class-methods-use-this
```
Kedae: Shouldn't be needed

opencti-platform/opencti-front/src/private/components/data/connectors/ConnectorWorks.tsx
```diff
@@ -0,0 +1,360 @@
...
+  mutation ConnectorWorksWorkDeletionMutation($id: ID!) {
+    workEdit(id: $id) {
+      delete
+    }
+  }
+`;
+
+type WorkMessages = NonNullable<NonNullable<NonNullable<ConnectorWorks_data$data['works']>['edges']>[0]>['node']['errors'];
+
+interface Options {
```
Kedae: Already defined in the code

opencti-platform/opencti-front/src/private/components/data/connectors/ConnectorWorks.tsx
```diff
@@ -0,0 +1,359 @@
...
+  };
+
+  const handleCloseErrors = () => {
+    setDisplayErrors(false);
+    setErrors([]);
+  };
+
+  // eslint-disable-next-line class-methods-use-this
+  const handleDeleteWork = (workId: string) => {
+    commitMutation({
```
Kedae: Should be a useMutation

opencti-platform/opencti-front/src/private/components/data/connectors/ConnectorWorks.tsx
```diff
@@ -0,0 +1,359 @@
...
+          variant="outlined"
+        >
+          <Typography align='center'>
+            {t_i18n('No work')}
+          </Typography>
+        </Paper>
+      )}
+      {works.map((workEdge) => {
+        const work = workEdge?.node;
+        if (!work) return null;
```
Kedae, line 130: I think you should check the graphQL, I don't see why we could have null workEdge here

opencti-platform/opencti-front/src/private/components/data/connectors/ConnectorWorks.tsx
```diff
@@ -0,0 +1,359 @@
...
+                        tracking && !!tracking.import_expected_number && !!tracking.import_processed_number
+                          ? Math.round((tracking.import_processed_number / tracking.import_expected_number) * 100)
+                          : 0
+                      }
+                    />
+                  </Grid>
+                </Grid>
+              </Grid>
+              <Button
+                style={{ position: 'absolute', right: 10, top: 10 }}
```
Kedae: Try avoiding inline styling

opencti-platform/opencti-front/src/private/components/data/connectors/ConnectorWorks.tsx
```diff
@@ -0,0 +1,359 @@
...
+                variant="contained"
+                color="secondary"
+                onClick={() => handleOpenErrors(work.errors ?? [])}
+                size="small"
+              >
+                {work.errors?.length} {t_i18n('errors')}
+              </Button>
+              <Button
+                variant="outlined"
+                style={{ position: 'absolute', right: 10, bottom: 10 }}
```
Kedae: Same

opencti-platform/opencti-front/src/private/components/data/connectors/ConnectorWorks.tsx
```diff
@@ -0,0 +1,359 @@
...
+                {t_i18n('Delete')}
+              </Button>
+            </Grid>
+          </Paper>
+        );
+      })}
+      <Dialog
+        PaperProps={{ elevation: 1 }}
+        open={displayErrors}
+        keepMounted={true}
```
Kedae: Useless ={true}

frapuks: tested and removed

opencti-platform/opencti-front/src/private/components/data/connectors/ConnectorWorks.tsx
frapuks: `works.edges` can be `null` or Array of (`object` or `null`), because the current relay generated type is :
```
export type ConnectorWorks_data$data = {
  readonly works: {
    readonly edges: ReadonlyArray<{...} | null> | null;
  } | null;
};
```

opencti-platform/opencti-front/src/private/components/data/connectors/ConnectorWorks.tsx
frapuks: refacto in a class

opencti-platform/opencti-front/src/private/components/data/connectors/ConnectorWorks.tsx
frapuks: refacto in a class

opencti-platform/opencti-front/src/private/components/data/connectors/ConnectorWorks.tsx
frapuks: replace by ConnectorWorksQuery$variables

opencti-platform/opencti-front/src/private/components/data/connectors/ConnectorWorks.tsx
frapuks: I've removed all unnecessary eslint comments

opencti-platform/opencti-front/src/private/components/data/connectors/ConnectorWorks.tsx
frapuks: update with useMutation

opencti-platform/opencti-front/src/private/components/data/connectors/ConnectorWorks.tsx
frapuks: GraphQL types :
```
type WorkConnection {
  pageInfo: PageInfo!
  edges: [WorkEdge]               <= edges can be null / WorkEdge can be null
}

type WorkEdge {
  cursor: String!
  node: Work!                          <= node can be null
}

type Work {
  id: ID!
  name: String
  user: Creator
  connector: Connector
  timestamp: DateTime!
  status: State!
  event_source_id: String
  received_time: DateTime
  processed_time: DateTime
  completed_time: DateTime
  completed_number: Int
  messages: [WorkMessage]
  errors: [WorkMessage]
  tracking: WorkTracking
}
```

frapuks:
ready for a 2nd round of reviews

opencti-platform/opencti-front/src/private/components/data/connectors/ConnectorWorks.tsx
labo-flg: @Kedae I had the same remark initially, but it appears that the gql types are defined with possible null at all levels, even inside arrays... It seems strange, but changing this requires checking the actual implementation backend side, see if it's _really_ not possible.

labo-flg (approved):
Beside the uncanny graphql typing with abusive null, I'm ok for these changes.
Now, if we want to revise the graphql API, this can be as easy as adding a bunch of `!`, but we need to check the implem backend side never sends null _for real_

PR #5763, opened 2024-01-31
[backend] Works tracking fail to complete at "update processed time" (#5762)

### LINKED ISSUE #5762: [backend] Works tracking fail to complete at "update processed time"
## Description

Auto complete works at processed time update fail because of type enforcing in the search engine

`illegal_argument_exception: Integer value passed as String`

Data in redis are stored in string so we need to convert this figures to int before using it.

Code can be refactored to only use one function (currently one is ok and the other not)

richard-julien:
See #5762

### DIFF
--- a/opencti-platform/opencti-graphql/src/domain/work.js
+++ b/opencti-platform/opencti-graphql/src/domain/work.js
@@ -3,7 +3,7 @@ import * as R from 'ramda';
 import { elDeleteInstances, elIndex, elLoadById, elPaginate, elRawDeleteByQuery, elUpdate, } from '../database/engine';
 import { generateWorkId } from '../schema/identifier';
 import { INDEX_HISTORY, isNotEmptyField, READ_INDEX_HISTORY } from '../database/utils';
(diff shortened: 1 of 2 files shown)


labo-flg (approved):

opencti-platform/opencti-graphql/src/domain/work.js
```diff
@@ -197,11 +197,6 @@ export const createWork = async (context, user, connector, friendlyName, sourceI
   return loadWorkById(context, user, workId);
 };
 
-const isWorkCompleted = async (workId) => {
-  const { import_processed_number: pn, import_expected_number: en } = await redisGetWork(workId);
-  return { isComplete: parseInt(pn, 10) === parseInt(en, 10), total: pn };
```
labo-flg, line 202: you would return also 'expected' now that was not returned before, no issue ?

lndrtrbn (approved):

opencti-platform/opencti-graphql/src/domain/work.js
richard-julien: Expected is used by the caller of redisUpdateWorkFigures and not directly by the caller of isWorkCompleted.
I dont think its a problem.

PR #5768, opened 2024-01-31
[backend] crop confidence values between 0 and 100

### LINKED ISSUE #5759: Make sure confidence level is always between 0 and 100
## issue
Following #4304, we need to make sure that the value saved in the platform for any `max_confidence` fields (user, groups) are always between 0 and 100.

We cannot enforce this in the schema mapping in elastic, so we need to check this in the backend code:

## test cases

This can only be tested through API calls, the UI already crop the value between 0 and 100.
This concerns operations:
* user add/edit
* group add/edit

Attention, edition can be plain object value, or field patch with only the `max_confidence` and corresponding `object_path`.

Example with edit group:

```json
{
  "id": "00b4e453-40e2-451d-8c32-09fd0532d80f",
  "input": {
    "key": "group_confidence_level",
    "object_path": "group_confidence_level/max_confidence",
    "value": 45
  }
}

{
  "id": "00b4e453-40e2-451d-8c32-09fd0532d80f",
  "input": {
    "key": "group_confidence_level",
    "value": {
      "max_confidence": 45,
      "overrides: []
    }
  }
}
```

labo-flg:
## Description
Closes #5759

This PR handles bad user input

I've split the code into utilities function with unit tests that should cover every possible cases.

### Test cases

For each of these cases, the values can be tested as < 0, between 0 and 100 and above 100, to verify cropping is correct.

* user creation, with `user_confidence level`
* user edition with complete `user_max_confidence_level` payload
* user edition with partial `user_max_confidence_level` payload > max_confidence value
* user edition with partial `user_max_confidence_level` payload > complete overrides value
* user edition with partial `user_max_confidence_level` payload > overrides.max_confidence value
* group creation, with `group_confidence level`
* group edition with complete `group_confidence ` payload
* group edition with partial `group_confidence ` payload > max_confidence value
* group edition with partial `group_confidence ` payload > complete overrides value
* group edition with partial `group_confidence ` payload > overrides.max_confidence value

Error cases
* group edition with a bad `object_path` > throws "Unhandled object_path for patching a confidence level"
* group edition with a correct `object_path` but value is mistyped > throws "Cannot crop non-finite input value"

### Checklist

### Further comments

Here are some test queries / payloads for the playground.

[confidence-queries.txt](https://github.com/OpenCTI-Platform/opencti/files/14114617/confidence-queries.txt)

### DIFF
--- a/opencti-platform/opencti-graphql/tests/01-unit/utils/confidence-level-test.ts
+++ b/opencti-platform/opencti-graphql/tests/01-unit/utils/confidence-level-test.ts
@@ -94,6 +94,11 @@ describe('Confidence level utilities', () => {
       }
     };
 
+    const groupNull = {
+      id: 'groupNull',
+      group_confidence_level: null
+    };
+
     // minimal subset of a real User
     const userA = {
       id: 'userA',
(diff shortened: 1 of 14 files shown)


opencti-platform/opencti-graphql/tests/01-unit/utils/confidence-level-test.ts
```diff
@@ -0,0 +1,150 @@
...
+
+    const userC = {
+      user_confidence_level: null,
+      groups: [groupNull, group70, groupNull],
+    };
+    expect(computeUserEffectiveConfidenceLevel(userC as unknown as AuthUser)).toEqual({
+      max_confidence: 70,
+      overrides: [],
+      source: group70,
+    });
```
labo-flg, line 136: This case was not handled properly before this PR. If a user had several groups, and the last one had null confidence, it would end up null.

labo-flg:
Wrong way. This PR is closed in favor of a simpler solution for now.
see #5787

PR #5785, opened 2024-02-01
[frontend] improve the readability of connector names (#3154)

frapuks:
UPDATE

Prevent long name whitout whitespace
re-Add Tooltip
Reorganize Grid to make space in leftSide to displaying maximum of chars.


frapuks:
Final result, which takes care of names that are too long and without spaces, and redistributes Grid values to give more space to the name.
We can't reduce the spaces any more, otherwise end-dates might be on 2 lines, and/or the progress bar might overlap the button.
image

SarahBocognano (approved):

Issue #5696, opened 2024-01-25
Enforce input validation against schema for patching object attributes


labo-flg:
## Problem
This issue appeared with the new schema definition introduced with #5392.
Patching attributes with `editField` was previously validated with json schema, which has been removed.

This is not an issue 99% of the time... but we introduced a new use case in #4304 : a non-multiple object attribute.

This kind of object can be updated with either a complete payload, such as
```json
{
  "id": "4b762ee9-b2da-49fe-99d7-dfd6b9e1fe33",
  "input": {
    "key": "user_confidence_level",
    "value": {
      "max_confidence": 50,
      "overrides": []
  }
}
```

... or a partial "patch" update like 

```json
{
  "id": "4b762ee9-b2da-49fe-99d7-dfd6b9e1fe33",
  "input": {
    "key": "user_confidence_level",
    "object_path": "/user_confidence_level/max_confidence",
    "value": 50
}
```

However, nothing prevents the user from sending a complete update with a partial object, such as:

```json
{
  "id": "4b762ee9-b2da-49fe-99d7-dfd6b9e1fe33",
  "input": {
    "key": "user_confidence_level",
    "value": {
      "max_confidence": 50,
    }
  }
}
```

This would still be valid for elastic, and there is no way of preventing this on elastic side (no concept of mandatory field).
The consequences are bad: the object is now partial in DB, but shall be complete for GQL (in my example, `overrides` cannot be `null`)

## Solution

Enforce a validation against the schema when editing an object field.

~~This can be done at the earliest time (resolver) or at the bottom of the chain (just before elastic query). TBD.~~
-> done just before indexing, when the payload is ready.

labo-flg:
Completed with #6046

PR #5787, opened 2024-02-01
[backend] crop effective confidence level in [0-100]

### LINKED ISSUE #5759: Make sure confidence level is always between 0 and 100

labo-flg:
Closes https://github.com/OpenCTI-Platform/opencti/issues/5759

### Proposed changes

* `user_effective_level` is cropped between [0-100] on computation, that's all
* no cropping when adding/editing the confidence values of users and groups. Thus we cannot trust them for now (this might be addressed in #5696 as part of a new validation against schema, including such numerical constraints)
* adding confidence information to the internal user in context ; for use internnaly
* built-in users have a confidence level now (all to 100 so they can do their job correctly).
* updated test base to improve coverage

### Checklist

### Further comments

This PR replaces #5768

### DIFF
--- a/opencti-platform/opencti-graphql/src/utils/confidence-level.ts
+++ b/opencti-platform/opencti-graphql/src/utils/confidence-level.ts
@@ -1,0 +1,44 @@
+import type { AuthUser } from '../types/user';
+import { cropNumber } from './math';
+
+export const computeUserEffectiveConfidenceLevel = (user: AuthUser) => {
+  // if user has a specific confidence level, it overrides everything and we return it
+  if (user.user_confidence_level) {
+    return {
+      // we make sure the levels are cropped in between 0-100
+      // other values were possibly injected in octi <6.0 though API calls
+      max_confidence: cropNumber(user.user_confidence_level.max_confidence, { min: 0, max: 100 }),
+      overrides: user.user_confidence_level.overrides.map((override) => ({
+        max_confidence: cropNumber(override.max_confidence, { min: 0, max: 100 }),
+        entity_type: override.entity_type,
+      })),
+      source: user,
+    };
+  }
+
+  // otherwise we get all groups for this user, and select the lowest max_confidence found
+  let minLevel = null;
+  let source = null;
+  if (user.groups) {
+    for (let i = 0; i < user.groups.length; i += 1) {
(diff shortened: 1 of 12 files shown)


opencti-platform/opencti-graphql/src/utils/confidence-level.ts
```diff
@@ -0,0 +1,44 @@
+import type { AuthUser } from '../types/user';
+import { cropNumber } from './math';
+
+export const computeUserEffectiveConfidenceLevel = (user: AuthUser) => {
```
labo-flg, line 4: I moved this function as an utility and added unit tests

opencti-platform/opencti-graphql/tests/02-integration/02-resolvers/group-test.js
```diff
@@ -167,7 +188,7 @@ describe('Group resolver standard behavior', () => {
   });
   it('should list groups', async () => {
     const queryResult = await queryAsAdmin({ query: LIST_QUERY, variables: { first: 10 } });
-    expect(queryResult.data.groups.edges.length).toEqual(6);
+    expect(queryResult.data.groups.edges.length).toEqual(8);
```
labo-flg:
```suggestion
    expect(queryResult.data.groups.edges.length).toEqual(7);
```

SarahBocognano:
Tested ✅ Works fine !

SarahBocognano (approved):

PR #5796, opened 2024-02-02
[backend] extend our vitest config for easier dev process

labo-flg:
Improved our tooling to be able to run dev tests locally quickly.

the `globalSetup` for vitest has been refactored and can be piloted by 2 env variables, `INIT_TEST_PLATFORM` and `SKIP_CLEANUP_PLATFORM`. We make use of these variables in new yarn scripts:
* run `yarn test:dev:init` to cleanup your test DB, start a platform, run the loader tests that populates the database. This command won't run any other test.
* once initialized this way,  run `yarn test:dev:resume <name-of-a-test-file.js>` to run the specified test file only, the DB won't be cleaned up after run
* ... or  run `yarn test:watch:resume <name-of-a-test-file.js>` to run in watch mode

This makes developing integration tests much quicker and without having to tweak some config files locally.

As for unit tests, I would still advise to use the vitest integration in the IDE. Nothing beats that as it just won't start a platform, useless for unit tests.

The previous commands still exists ; 
* `yarn test:dev` for 1 complete run (cleanup/setup/all tests suites)
* `yarn test:watch` for a complete execution in watch mode
* `yarn test` unchanged, used by CI to run once in silent with coverage

### DIFF
--- a/opencti-platform/opencti-graphql/package.json
+++ b/opencti-platform/opencti-graphql/package.json
@@ -19,9 +19,10 @@
     "start-cluster": "NODE_ENV=cluster node build/back.js",
     "serv": "node build/back.js",
     "insert:dev": "node build/script-insert-dataset.js",
-    "test:dev": "vitest watch --config vitest.config.dev.ts",
-    "test:dev-init-only": "ONLY_PLATFORM_INIT=1 vitest run --config vitest.config.dev.ts tests/02-integration/00-inject/loader-test.js",
-    "test:dev-no-cleanup": "SKIP_CLEANUP_PLATFORM_AT_START=1 vitest watch --config vitest.config.dev.ts",
+    "test:dev:init": "INIT_TEST_PLATFORM=true vitest run --config vitest.config.dev.ts --testNamePattern=\"Should import creation succeed\" tests/02-integration/00-inject/loader-test.js",
+    "test:dev:resume": "SKIP_CLEANUP_PLATFORM=true vitest run --config vitest.config.dev.ts",
+    "test:dev": "vitest run --config vitest.config.dev.ts",
+    "test:watch": "vitest watch --config vitest.config.dev.ts",
     "test": "vitest --silent run --config vitest.config.test.ts --coverage"
   },
   "pkg": {
--- a/opencti-platform/opencti-graphql/package.json
+++ b/opencti-platform/opencti-graphql/package.json
@@ -19,9 +19,10 @@
     "start-cluster": "NODE_ENV=cluster node build/back.js",
     "serv": "node build/back.js",
     "insert:dev": "node build/script-insert-dataset.js",
-    "test:dev:init": "INIT_TEST_PLATFORM=true vitest run --config vitest.config.dev.ts --testNamePattern=\"Should import creation succeed\" tests/02-integration/00-inject/loader-test.js",
+    "test:dev:init": "INIT_TEST_PLATFORM=true vitest run --config vitest.config.dev.ts --testNamePattern=\"Should import creation succeed\" tests/02-integration/00-inject/loader-test.ts",
     "test:dev:resume": "SKIP_CLEANUP_PLATFORM=true vitest run --config vitest.config.dev.ts",
     "test:dev": "vitest run --config vitest.config.dev.ts",
+    "test:watch:resume": "SKIP_CLEANUP_PLATFORM=true vitest watch --config vitest.config.dev.ts",
     "test:watch": "vitest watch --config vitest.config.dev.ts",
     "test": "vitest --silent run --config vitest.config.test.ts --coverage"
   },
--- a/opencti-platform/opencti-graphql/package.json
+++ b/opencti-platform/opencti-graphql/package.json
@@ -20,6 +20,8 @@
     "serv": "node build/back.js",
     "insert:dev": "node build/script-insert-dataset.js",
     "test:dev": "vitest watch --config vitest.config.dev.ts",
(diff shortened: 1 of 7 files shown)


opencti-platform/opencti-graphql/package.json
```diff
@@ -20,6 +20,8 @@
     "serv": "node build/back.js",
     "insert:dev": "node build/script-insert-dataset.js",
     "test:dev": "vitest watch --config vitest.config.dev.ts",
+    "test:dev-init-only": "ONLY_PLATFORM_INIT=1 vitest run --config vitest.config.dev.ts tests/02-integration/00-inject/loader-test.js",
```
aHenryJard: Can you use "true" instead of "1", when it's a number I expect to be able to put 3, 4, 10 etc...

opencti-platform/opencti-graphql/tests/utils/globalSetup.js
```diff
@@ -18,40 +18,67 @@ import { executionContext } from '../../src/utils/access';
...
+ * run yarn test:dev-init-only to cleanup and reinit a test platform (it also provision the data)
+ * run yarn test:dev-no-cleanup <file-pattern> to run directly some tests without cleanup and init of the test platform
+ */
+
+const { ONLY_PLATFORM_INIT, SKIP_CLEANUP_PLATFORM_AT_START } = process.env;
+
+const initializePlatform = async () => {
   const context = executionContext('platform_test_initialization');
-  logApp.info('[OPENCTI] Starting platform');
+  console.log('🚀 [vitest-global-setup] initializing platform');
```
aHenryJard: Is it intended to use console.log ?
And sadly 🚀 char does not show on my Idea console 😞

opencti-platform/opencti-graphql/tests/utils/globalSetup.js
```diff
@@ -63,15 +90,38 @@ const platformClean = async () => {
...
 
 export async function setup() {
-  // Platform cleanup before executing tests
-  await platformClean();
+  if (ONLY_PLATFORM_INIT) {
+    console.log('🚀 [vitest-global-setup] only running test platform initialization');
+    const stopTime = new Date().getTime();
+    await platformClean();
+    await testPlatformStart();
+    await wait(15000); // Wait 15 secs for complete platform start
```
aHenryJard: suggestion for the future: maybe we could use the platform healthcheck url here and a waiting loop instead of a magic number.

aHenryJard (commented):
Thanks for making tests easier :)

aHenryJard (commented):
Could you move globalSetup to ts or is it too complex ?
And maybe also tests/02-integration/00-inject/loader-test.js file.

If you have time, this comment is not a blocker at all.

opencti-platform/opencti-graphql/package.json
```diff
@@ -20,6 +20,8 @@
     "serv": "node build/back.js",
     "insert:dev": "node build/script-insert-dataset.js",
     "test:dev": "vitest watch --config vitest.config.dev.ts",
```
aHenryJard: Now that I'm playing a lot with tests, I hesitate to have both watch and not watch, something like:
```suggestion
     "test:watch": "vitest watch --config vitest.config.dev.ts",
    "test:dev": "vitest run --config vitest.config.dev.ts",
```

opencti-platform/opencti-graphql/tests/utils/globalSetup.js
labo-flg: I'll use logApp and remove emoji then. Will work for everyone.

opencti-platform/opencti-graphql/package.json
labo-flg: yes, makes sense. I'm rarely using watch mode personally.

opencti-platform/opencti-graphql/tests/utils/globalSetup.js
labo-flg: done with a probe query to `about {version }` for now. It speeds up the setup time.

opencti-platform/opencti-graphql/package.json
```diff
@@ -19,7 +19,11 @@
     "start-cluster": "NODE_ENV=cluster node build/back.js",
     "serv": "node build/back.js",
     "insert:dev": "node build/script-insert-dataset.js",
-    "test:dev": "vitest watch --config vitest.config.dev.ts",
+    "test:dev:init": "INIT_TEST_PLATFORM=true vitest run --config vitest.config.dev.ts --testNamePattern=\"Should import creation succeed\" tests/02-integration/00-inject/loader-test.ts",
```
aHenryJard, line 22: This step is not working on my computer, "test:dev:init" is not working on my computer, I have a error at "Should import creation succeed" step,

aHenryJard (commented):
As a side note, it might be a good time to improve the developper documentation as follow up PR https://docs.opencti.io/latest/development/platform/#for-testing

opencti-platform/opencti-graphql/package.json
labo-flg: I suppose you do not have more details ?
Are you sure it's not related to a python client version mismatch on your local setup?

SouadHadjiat: I just managed to run this step without any issue

SouadHadjiat (approved):

opencti-platform/opencti-graphql/package.json
aHenryJard: ok testing again now

aHenryJard (approved):

Issue #5697, opened 2024-01-25
Assert associated User's confidence when ingesting data from streams, feeds and connectors


labo-flg:
When a stream, a feed or a connector updates or creates data, the associated user's confidence level shall be used.

Feed, Stream and connector must have an associated user.

When inserting data, the confidence level of the data is thresholded by the user's max confidence level, before being send to ingestion/dedup process.

### linked to
* #4304
* #4901 
* #5692
* #5694 
* #5695
* #5696

PR #5800, opened 2024-02-02
[backend] enforce user max confidence on create/delete/update/upsert (#5697)

labo-flg:
Addresses backend part for #5697

### Proposed changes
The code changes are quite limited, but are located in critical places. They have a lot of impact.

If a user has no effective confidence level due to a bad configuration of the groups, a `LockTimeoutError` is thrown (a special kind of error that causes worker to retry infinitely - this way no message is lost).
We expect this error to disappear in the future (it's a manual migration for admins), and want to limit the _soon-to-be-dead_ code that handles it.

Legit errors are thrown on create, update or upsert. There are a lot of edge cases depending on the ingestion pipeline being used, but here are the main principles:
* a failed confidence control should throw an error on direct, manual operations (user action / api call)
* a failed confidence control during an automated operation (like auto-enrichment or background tasks) should not throw error (the operation is simply not done and the whole process continues)
* on create, confidence level of created entity cannot be higher than user's level. If not, the value is capped automatically (no error thrown, only a log warning).
* on create, if confidence is not provided we use the user's level ; which is 100 in all our tests. Before, it felt back to 0, so I had to mitigate the impact on the test base.
* on update, if user has insufficient confidence level compared to the element, an error is thrown
* on delete, if user has insufficient confidence level compared to the element, an error is thrown

On upsert (connectors, feeds, etc.) this is a lot trickier, I advise to read the code changes carefully.
One impactful change is that, if the input data has a higher confidence than the current element, fields are updated.
Before, it was not the case for refs (like createdBy).

The new confidence control enlighted several issues and tricky bugs that were silent until now, so we had to fix them too.

### Related issues

* #5697

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/common/files/workbench/WorkbenchFileContent.jsx
+++ b/opencti-platform/opencti-front/src/private/components/common/files/workbench/WorkbenchFileContent.jsx
@@ -807,6 +807,12 @@ const WorkbenchFileContentComponent = ({
                 history.push('/dashboard/data/import');
               }
             },
+            onError: ({ res }) => {
+              MESSAGING$.notifyError(res.errors?.[0]?.message || t_i18n('An unknown error has occurred! Please try again later.'));
+              setSubmitting(false);
+              resetForm();
+              setDisplayValidate(false);
+            },
           });
         }, 2000);
       },
--- a/opencti-platform/opencti-front/src/private/components/common/files/workbench/WorkbenchFileContent.jsx
+++ b/opencti-platform/opencti-front/src/private/components/common/files/workbench/WorkbenchFileContent.jsx
@@ -38,7 +38,7 @@ import StixItemLabels from '../../../../../components/StixItemLabels';
 import StixItemMarkings from '../../../../../components/StixItemMarkings';
 import SwitchField from '../../../../../components/SwitchField';
 import TextField from '../../../../../components/TextField';
-import { APP_BASE_PATH, commitMutation, MESSAGING$, QueryRenderer } from '../../../../../relay/environment';
+import { APP_BASE_PATH, commitMutation, handleError, MESSAGING$, QueryRenderer } from '../../../../../relay/environment';
 import { observableValue, resolveIdentityClass, resolveIdentityType, resolveLink, resolveLocationType, resolveThreatActorType } from '../../../../../utils/Entity';
 import { defaultKey, defaultValue } from '../../../../../utils/Graph';
 import useAttributes from '../../../../../utils/hooks/useAttributes';
@@ -807,8 +807,8 @@ const WorkbenchFileContentComponent = ({
                 history.push('/dashboard/data/import');
               }
             },
-            onError: ({ res }) => {
-              MESSAGING$.notifyError(res.errors?.[0]?.message || t_i18n('An unknown error has occurred! Please try again later.'));
+            onError: (error) => {
+              handleError(error);
               setSubmitting(false);
               resetForm();
               setDisplayValidate(false);
--- a/opencti-platform/opencti-front/src/private/components/data/tasks/TasksList.jsx
+++ b/opencti-platform/opencti-front/src/private/components/data/tasks/TasksList.jsx
@@ -392,7 +392,7 @@ const TasksList = ({ data }) => {
               <Button
                 style={{ position: 'absolute', right: 10, top: 10 }}
                 variant={task.errors.length > 0 ? 'contained' : 'outlined'}
-                color={'secondary'}
+                color={'error'}
                 disabled={task.errors.length === 0}
                 onClick={() => handleOpenErrors(task.errors)}
                 size="small"
--- a/opencti-platform/opencti-front/src/private/components/data/tasks/TasksList.jsx
+++ b/opencti-platform/opencti-front/src/private/components/data/tasks/TasksList.jsx
@@ -391,8 +391,9 @@ const TasksList = ({ data }) => {
               </Grid>
               <Button
                 style={{ position: 'absolute', right: 10, top: 10 }}
-                variant="contained"
-                color="secondary"
+                variant={task.errors.length > 0 ? 'contained' : 'outlined'}
+                color={'secondary'}
+                disabled={task.errors.length === 0}
                 onClick={() => handleOpenErrors(task.errors)}
                 size="small"
               >
--- a/opencti-platform/opencti-graphql/src/database/middleware.js
+++ b/opencti-platform/opencti-graphql/src/database/middleware.js
@@ -2462,16 +2462,19 @@ const upsertElement = async (context, user, element, type, basePatch, opts = {})
           // In full synchro, just replace everything
           if (isUpsertSynchro) {
             inputs.push({ key: inputField, value: patchInputData ?? [], operation: UPDATE_OPERATION_REPLACE });
-          } else if (isInputWithData && diffTargets.length > 0) {
-            // If data is provided and different from existing data, apply an add operation
+          } else if (isInputWithData && diffTargets.length > 0 && isConfidenceMatch) {
+            // If data is provided, different from existing data, and confidence is ok, apply an add operation
             inputs.push({ key: inputField, value: diffTargets, operation: UPDATE_OPERATION_ADD });
           }
         }
       } else {
         const currentData = element[relDef.databaseName];
-        const updatable = isUpsertSynchro || (isInputWithData && isEmptyField(currentData));
+        const updatable = isUpsertSynchro || (isInputWithData && isEmptyField(currentData)) || isConfidenceMatch;
         // If expected data is different from current data
-        // And data can be updated (complete a null value or forced through synchro upsert option
+        // And data can be updated:
+        //   forced synchro
+        //   OR the field was null -> better than nothing !
+        //   OR the confidence matches -> new value is "better" than existing value
         if (!R.equals(currentData, patchInputData) && updatable) {
           inputs.push({ key: inputField, value: [patchInputData] });
         }
--- a/opencti-platform/opencti-graphql/src/database/middleware.js
+++ b/opencti-platform/opencti-graphql/src/database/middleware.js
@@ -2667,10 +2667,7 @@ export const createRelationRaw = async (context, user, rawInput, opts = {}) => {
 
   // when creating stix ref, we must check confidence on from side (this count has modifying this element itself)
   if (isStixRefRelationship(relationshipType)) {
-    const fromSideHasConfidenceAttribute = schemaAttributesDefinition.getAttribute(from.entity_type, 'confidence');
-    if (fromSideHasConfidenceAttribute) {
-      controlUserConfidenceAgainstElement(user, from);
-    }
+    controlUserConfidenceAgainstElement(user, from);
   }
 
   // check if user has "edit" access on from and to
@@ -3204,16 +3201,10 @@ export const internalDeleteElementById = async (context, user, id, opts = {}) =>
     throw AlreadyDeletedError({ id });
   }
   // region confidence control
-  const hasConfidenceAttribute = schemaAttributesDefinition.getAttribute(element.entity_type, 'confidence');
-  if (hasConfidenceAttribute) {
-    controlUserConfidenceAgainstElement(user, element);
-  }
+  controlUserConfidenceAgainstElement(user, element);
   // when deleting stix ref, we must check confidence on from side (this count has modifying this element itself)
   if (isStixRefRelationship(element.entity_type)) {
-    const fromSideHasConfidenceAttribute = schemaAttributesDefinition.getAttribute(element.from.entity_type, 'confidence');
-    if (fromSideHasConfidenceAttribute) {
-      controlUserConfidenceAgainstElement(user, element.from);
-    }
+    controlUserConfidenceAgainstElement(user, element.from);
   }
   // endregion
   // Prevent individual deletion if linked to a user
--- a/opencti-platform/opencti-graphql/src/database/middleware.js
+++ b/opencti-platform/opencti-graphql/src/database/middleware.js
@@ -2474,14 +2474,15 @@ const upsertElement = async (context, user, element, type, basePatch, opts = {})
           }
         }
       } else { // not multiple
+        // If expected data is different from current data...
         const currentData = element[relDef.databaseName];
-        const updatable = isUpsertSynchro || (isInputWithData && isEmptyField(currentData)) || isConfidenceMatch;
-        // If expected data is different from current data
-        // And data can be updated:
+        const isInputDifferentFromCurrent = !R.equals(currentData, patchInputData?.internal_id);
+        // ... and data can be updated:
         //   forced synchro
         //   OR the field was null -> better than nothing !
         //   OR the confidence matches -> new value is "better" than existing value
-        if (!R.equals(currentData, patchInputData) && updatable) {
+        const updatable = isUpsertSynchro || (isInputWithData && isEmptyField(currentData)) || isConfidenceMatch;
+        if (isInputDifferentFromCurrent && updatable) {
           inputs.push({ key: inputField, value: [patchInputData] });
         }
       }
--- a/opencti-platform/opencti-graphql/src/database/middleware.js
+++ b/opencti-platform/opencti-graphql/src/database/middleware.js
@@ -198,6 +198,12 @@ import { confidence } from '../schema/attribute-definition';
 import { ENTITY_TYPE_INDICATOR } from '../modules/indicator/indicator-types';
 import { FilterMode, FilterOperator } from '../generated/graphql';
 import { getMandatoryAttributesForSetting } from '../modules/entitySetting/entitySetting-attributeUtils';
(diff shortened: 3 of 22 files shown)


opencti-platform/opencti-graphql/src/database/middleware.js
```diff
@@ -1987,7 +1987,30 @@ export const updateAttribute = async (context, user, id, type, inputs, opts = {}
...
+  if (!user.effective_confidence_level) {
+    throw FunctionalError('User has no effective max confidence level and cannot update the element', { id });
+  }
+  const elementConfidence = Math.max(Math.min(initial.confidence ?? 0, 100), 0);
+  const userMaxConfidenceLevel = user.effective_confidence_level.max_confidence;
+  if (userMaxConfidenceLevel < elementConfidence) {
+    throw FunctionalError('User effective max confidence level is insufficient to update this element', { id });
+  }
+  // if the attribute updated is "confidence", new value cannot exceed user's max confidence
+  const newInputs = inputs.map((input) => {
```
richard-julien: Maybe throw could be a better solution

opencti-platform/opencti-graphql/src/database/middleware.js
```diff
@@ -2375,8 +2398,22 @@ const upsertElement = async (context, user, element, type, basePatch, opts = {})
...
     inputs.push(fileImpact);
   }
-  // If confidence is passed at creation, just compare confidence
-  const isConfidenceMatch = (updatePatch.confidence ?? 0) >= (element.confidence ?? 0);
+
+  // region confidence control
+  if (!user.effective_confidence_level) {
+    // LockTimeoutError would make the worker retry (we do not want to lose messages in that case
+    // TODO: new error and update worker?
+    throw LockTimeoutError({ user }, 'User has no effective confidence level and cannot upsert data in the platform.');
```
richard-julien: The full user is maybe to much, user_email and id should be good

opencti-platform/opencti-graphql/src/database/middleware.js
```diff
@@ -2375,8 +2398,22 @@ const upsertElement = async (context, user, element, type, basePatch, opts = {})
...
+  }
+  const elementConfidence = Math.max(Math.min(element.confidence ?? 0, 100), 0);
```
richard-julien: Looks a bit difficult to read.

opencti-platform/opencti-graphql/src/database/middleware.js
```diff
@@ -2886,9 +2923,19 @@ const buildEntityData = async (context, user, input, type, opts = {}) => {
...
   // -- STIX-Domain-Object
   if (isStixDomainObject(type)) {
+    // region confidence control
+    if (!user.effective_confidence_level) {
+      throw FunctionalError('User has no effective max confidence level and cannot create the element');
+    }
+    // confidence of the object is thresholded by the user's effective confidence level
+    const inputConfidence = Math.max(Math.min(input.confidence ?? 0, 100), 0);
+    const userMaxConfidenceLevel = user.effective_confidence_level.max_confidence;
+    const effectiveConfidence = Math.min(userMaxConfidenceLevel, inputConfidence);
```
richard-julien: Could be interesting to have an util function for that with some tests

opencti-platform/opencti-graphql/src/database/middleware.js
labo-flg: Yes, that would be more consistent behavior.

opencti-platform/opencti-graphql/src/database/middleware.js
labo-flg: Once https://github.com/OpenCTI-Platform/opencti/pull/5787 is merged, I'll replace this line with a more readable `cropNumber(value, 0, 100)`

opencti-platform/opencti-graphql/src/database/middleware.js
labo-flg: yes ! Once we validate the general approach, I have some light refactoring to do + tests.

opencti-platform/opencti-graphql/src/database/middleware.js
labo-flg: after some more discussions, we want to be consistent with respect to create and upsert strategies, so we simply crop the input value.

opencti-platform/opencti-graphql/src/manager/taskManager.js
```diff
@@ -155,13 +156,25 @@ const computeListTaskElements = async (context, user, task) => {
...
+        lastError = error;
+      }
     }
   }
+  if (lastError) {
+    await appendTaskErrors(
+      task,
+      [{ id: 'error--confidence-control', message: lastError.message }]
+    );
+  }
```
labo-flg: @richard-julien I'm not sure about this. It will give just 1 error in the tasks error list, so the users knows something bad happened ; but not a zillion errors.

opencti-platform/opencti-graphql/src/domain/backgroundTask.js
```diff
@@ -44,20 +45,22 @@ export const findAll = (context, user, args) => {
   return listEntities(context, user, [ENTITY_TYPE_BACKGROUND_TASK], args);
 };
 
-const buildQueryFilters = async (rawFilters, search, taskPosition) => {
+const buildQueryFilters = async (user, filters, search, taskPosition) => {
+  const inputFilters = filters ? JSON.parse(filters) : undefined;
+  const finalFilters = adaptFiltersWithUserConfidence(user, inputFilters);
```
labo-flg, line 50: this produces a filter on user's confidence AND the initial task filters.
No error would be produced in that case...  the task would complete with success (unless we have other way to report it ?)

opencti-platform/opencti-front/src/private/components/common/files/workbench/WorkbenchFileContent.jsx
```diff
@@ -807,6 +807,12 @@ const WorkbenchFileContentComponent = ({
                 history.push('/dashboard/data/import');
               }
             },
+            onError: ({ res }) => {
+              MESSAGING$.notifyError(res.errors?.[0]?.message || t_i18n('An unknown error has occurred! Please try again later.'));
+              setSubmitting(false);
+              resetForm();
+              setDisplayValidate(false);
+            },
```
labo-flg: This is one of probably many places where we do not catch any error after a mutation.
As a minimal feedback, I've added the backend error displayed in the small messaging popup.

This is not ideal, but not to be fully solved here.

opencti-platform/opencti-front/src/private/components/data/tasks/TasksList.jsx
```diff
@@ -391,8 +391,9 @@ const TasksList = ({ data }) => {
               </Grid>
               <Button
                 style={{ position: 'absolute', right: 10, top: 10 }}
-                variant="contained"
-                color="secondary"
+                variant={task.errors.length > 0 ? 'contained' : 'outlined'}
+                color={'secondary'}
+                disabled={task.errors.length === 0}
```
labo-flg: unrelated to initial issue: simple UI improvement to highlight errors in bg tasks
image

opencti-platform/opencti-graphql/src/database/middleware.js
```diff
@@ -1851,7 +1857,7 @@ export const updateAttributeMetaResolved = async (context, user, initial, inputs
         const { value: targetsCreated } = meta[metaIndex];
         const targetCreated = R.head(targetsCreated);
         // If asking for a real change
-        if (currentValue?.standard_id !== targetCreated?.internal_id && currentValue?.id !== targetCreated?.internal_id) {
+        if (currentValue?.id !== targetCreated?.internal_id) {
```
labo-flg, line 1860: comparing std id and internal would never match, this check is useless.

opencti-platform/opencti-graphql/src/database/middleware.js
```diff
@@ -2441,24 +2455,34 @@ const upsertElement = async (context, user, element, type, basePatch, opts = {})
...
-        const updatable = isUpsertSynchro || (isInputWithData && isEmptyField(currentData));
-        // If expected data is different from current data
-        // And data can be updated (complete a null value or forced through synchro upsert option
-        if (!R.equals(currentData, patchInputData) && updatable) {
+        const isInputDifferentFromCurrent = !R.equals(currentData, patchInputData);
+        // ... and data can be updated:
+        //   forced synchro
+        //   OR the field was null -> better than nothing !
+        //   OR the confidence matches -> new value is "better" than existing value
+        const updatable = isUpsertSynchro || (isInputWithData && isEmptyField(currentData)) || isConfidenceMatch;
```
labo-flg, line 2484: this change has big impact on upsert mechanism.

before, we would update a field if this field was empty (_better than nothing_ strategy).
now, we would **also** update the field if the entity being updated is of lower confidence than the input data.

tests have been added to `middleware-test.js`  for that, and counters updated accordingly.

opencti-platform/opencti-graphql/src/database/middleware.js
```diff
@@ -2742,10 +2779,8 @@ export const createRelationRaw = async (context, user, input, opts = {}) => {
         });
       }
       // TODO Handling merging relation when updating to prevent multiple relations finding
-      existingRelationship = R.head(filteredRelations);
-      // We can use the resolved input from/to to complete the element
-      existingRelationship.from = from;
-      existingRelationship.to = to;
+      // resolve all refs so we can upsert properly
+      existingRelationship = await storeLoadByIdWithRefs(context, user, R.head(filteredRelations).internal_id);
```
labo-flg, line 2783: resolving the whole relation has a cost, but we have to... 
before, only `from` and `to` would be resolved, which would cause issues in the test at `updateAttributeMetaResolved:1870` when upserting the very same relation. 
Relation like createdBy would be duplicated.

This fixes all refs cases.

opencti-platform/opencti-graphql/src/manager/taskManager.js
labo-flg: I'll remove this to be consistent. Our strategy is: do not make operation not possible, user shall know better via the UI what he can do.

opencti-platform/opencti-graphql/tests/02-integration/01-database/file-storage-test.js
```diff
@@ -48,7 +48,7 @@ describe('File storage file listing', () => {
     expect(list.edges.length).toEqual(1);
     file = head(list.edges).node;
     expect(file.id).toEqual(importFileId);
-    expect(file.size).toEqual(10692);
+    expect(file.size).toEqual(10700);
```
labo-flg, line 51: file now ocntians text "confidence: 100" instead of "0" so the size is a bit changed

opencti-platform/opencti-graphql/tests/03-streams/00-Raw/raw-test.js
```diff
@@ -65,7 +65,7 @@ describe('Raw streams tests', () => {
       expect(updateEventsByTypes['threat-actor'].length).toBe(17);
       expect(updateEventsByTypes['vocabulary'].length).toBe(3);
       expect(updateEventsByTypes['vulnerability'].length).toBe(3);
-      expect(updateEvents.length).toBe(132);
+      expect(updateEvents.length).toBe(134);
```
labo-flg, line 68: +2 update events with the new tests on middleware-test.

(4 update attempts, only 2 OK due to confidence.)

opencti-worker/src/worker.py
```diff
@@ -334,13 +334,10 @@ def data_handler(  # pylint: disable=too-many-statements, too-many-locals
         except Exception as ex:  # pylint: disable=broad-except
             error = str(ex)
             error_msg = traceback.format_exc()
-            if (
-                "LOCK_ERROR" in error_msg
-                and self.processing_count < MAX_PROCESSING_COUNT
```
labo-flg, line 339: We remove this limit so the retry is indeed infinite. This is a valuable change even for regular lock error (happens when an entity is locked by another operation in progress).

opencti-platform/opencti-front/src/private/components/data/tasks/TasksList.jsx
lndrtrbn: good idea, displaying errors in green is a bit confusing though

opencti-platform/opencti-graphql/src/database/middleware.js
lndrtrbn: If I understand correctly it's not confidence related. But a bug you found during development of your feature?

lndrtrbn (approved):

opencti-platform/opencti-front/src/private/components/data/tasks/TasksList.jsx
labo-flg: hm you're right, I'll use the error color!

opencti-platform/opencti-graphql/src/database/middleware.js
labo-flg: Well, we can assume taht the bug would never happened before. The introduction of the confidence changed that.
The new confidence mechanism now allows to update refs that way, so we need a fully resolved object.

opencti-platform/opencti-front/src/private/components/common/files/workbench/WorkbenchFileContent.jsx
SouadHadjiat: I think you can use directly the method `handleError` defined in environment.jsx,
we use it in some components
```
onError: (error: Error) => {
  handleError(error);
},
```

opencti-platform/opencti-graphql/src/manager/taskManager.js
```diff
@@ -159,7 +160,11 @@ const computeListTaskElements = async (context, user, task) => {
     const elementToResolve = ids[indexId];
     const element = await internalLoadById(context, user, elementToResolve, { type: DEFAULT_ALLOWED_TASK_ENTITY_TYPES });
     if (element) {
-      processingElements.push({ element, next: element.id });
+      // only process elements allowed by confidence control (no throw)
+      const isConfidenceMatch = controlUserConfidenceAgainstElement(user, element, true);
```
SouadHadjiat, line 165: why not test also in computeQueryTaskElements ?

opencti-platform/opencti-graphql/src/domain/backgroundTask.js
SouadHadjiat, line 50: I think background tasks can be used on elements that don't necessarly have a confidence like notifications, will it work anyway with the filter on confidence ?

SouadHadjiat: well I just tested locally by trying to mark as read all my notifications and it didn't work

SouadHadjiat: It also doesn't work on stix cyber observables, which were created without a confidence (the field is not present in the db), so if you try to apply a background task query on observables, it won't work

SouadHadjiat (dismissed):
Query background tasks don't work on Notifications and observables (and maybe other entities that don't have confidence attribute)

labo-flg:
> Query background tasks don't work on Notifications and observables (and maybe other entities that don't have confidence attribute)

thanks ! I'm on it

opencti-platform/opencti-graphql/src/manager/taskManager.js
labo-flg: It is done in `computeQueryTaskElements` inside `executeTaskQuery`.
In this case, we directly adapt the filter used in `executeTaskQuery` so that we filter out elements that wont match the confidence level. This optimizes the code for large batches, no need to pass on each element in the batch.

In `computeListTaskElements` we have to do this filtering _after_ fetching all elements, no other way I think...

opencti-platform/opencti-graphql/src/domain/backgroundTask.js
labo-flg: indeed ! I'll adapt the filter so that it allows null confidence too.

labo-flg:
@SouadHadjiat [e5740f2](https://github.com/OpenCTI-Platform/opencti/pull/5800/commits/e5740f201ca8d131fb90981e50865749a22848c1) fixes this problem by crafting a better filter. Items without confidence (not SDO or Relationships) will not be filtered out.

Not that adding a filter "confidence is Nil" is not working if the element has no attribute "confidence" at all in schema, so I must use this filter on `entity_type` instead.

My final filter includes the 3 cases:
* the entity has no confidence (not a SDO or Relationship)
* or the confidence is null
* or the confidence is lower or equals to the user's level

This shall make background task work with observables and notifications also.

aHenryJard (commented):
I did a quick test with other PR on client python and connectors:
- recorded future notes: works fine, the confidence of the connector user is correctly applied to item created
- ipinfo, works fine, if confidence is lower, the enrichment is not applied. The only thing is that on UI it say ok, and the error is only in logs.

I also tested with an openid provider (keycloak), the user that is created that way inherit from Default group confidence, so it's all good for me.

aHenryJard (approved):

PR #5802, opened 2024-02-02
[backend/frontend] Introduce dev flags to deactivate feature in ongoing development

richard-julien:
This PR aims to introduce a dev configuration to disable features that are currently in development but not ready yet.
That's will ease the merging of large feature but providing simple tools to hide the feature.

In this PR you will find a example of Decay indicator feature deactivation.
- Deactivation of the manager registration
- Deactivation of the module registration
- Deactivation of a frontend specific menu

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/settings/CustomizationMenu.tsx
+++ b/opencti-platform/opencti-front/src/private/components/settings/CustomizationMenu.tsx
@@ -1,9 +1,7 @@
 import React, { FunctionComponent } from 'react';
 import NavToolbarMenu, { MenuEntry } from '../common/menus/NavToolbarMenu';
-import useHelper from '../../../utils/hooks/useHelper';
 
 const CustomizationMenu: FunctionComponent = () => {
-  const { isFeatureEnable } = useHelper();
   const entries: MenuEntry[] = [
     {
       path: '/dashboard/settings/customization/entity_types',
@@ -21,10 +19,6 @@ const CustomizationMenu: FunctionComponent = () => {
       path: '/dashboard/settings/customization/retention',
       label: 'Retention policies',
     },
-    ...(isFeatureEnable('INDICATOR_DECAY') ? [{
-      path: '/dashboard/settings/customization/decay',
-      label: 'Decay rules',
-    }] : []),
   ];
   return <NavToolbarMenu entries={entries} />;
 };
--- a/opencti-platform/opencti-front/src/private/components/settings/CustomizationMenu.tsx
+++ b/opencti-platform/opencti-front/src/private/components/settings/CustomizationMenu.tsx
@@ -1,7 +1,9 @@
 import React, { FunctionComponent } from 'react';
 import NavToolbarMenu, { MenuEntry } from '../common/menus/NavToolbarMenu';
+import useHelper from '../../../utils/hooks/useHelper';
 
 const CustomizationMenu: FunctionComponent = () => {
+  const { isFeatureEnable } = useHelper();
   const entries: MenuEntry[] = [
     {
       path: '/dashboard/settings/customization/entity_types',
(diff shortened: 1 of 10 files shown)


opencti-platform/opencti-front/src/private/components/settings/CustomizationMenu.tsx
```diff
@@ -1,7 +1,9 @@
 import React, { FunctionComponent } from 'react';
 import NavToolbarMenu, { MenuEntry } from '../common/menus/NavToolbarMenu';
+import useHelper from '../../../utils/hooks/useHelper';
 
 const CustomizationMenu: FunctionComponent = () => {
+  const { isFeatureEnable } = useHelper();
```
aHenryJard:
```suggestion
  const { isFeatureEnabled } = useHelper();
```

opencti-platform/opencti-front/src/private/components/settings/CustomizationMenu.tsx
```diff
@@ -19,6 +21,10 @@ const CustomizationMenu: FunctionComponent = () => {
       path: '/dashboard/settings/customization/retention',
       label: 'Retention policies',
     },
+    ...(isFeatureEnable('INDICATOR_DECAY') ? [{
```
aHenryJard:
```suggestion
    ...(isFeatureEnabled('INDICATOR_DECAY') ? [{
```

opencti-platform/opencti-front/src/utils/platformModulesHelper.ts
```diff
@@ -36,7 +36,7 @@ const isFeatureEnable = (
 ) => {
   const flags = settings.platform_feature_flags ?? [];
   const feature = flags.find((f) => f.id === id);
-  return feature !== undefined && feature.enable === true;
+  return feature === undefined || feature.enable === true;
```
aHenryJard, line 39:
```suggestion
  return feature === undefined || feature.enabled === true;
```

opencti-platform/opencti-front/src/private/components/settings/Settings.jsx
```diff
@@ -972,9 +971,7 @@ const Settings = () => {
                               <ListItemText primary={t_i18n(module.id)} />
                               <ItemBoolean
                                 variant="inList"
-                                label={
-                                  module.enable ? t_i18n('Enabled') : t_i18n('Disabled')
-                                }
+                                label={module.enable ? t_i18n('Enabled') : t_i18n('Disabled')}
```
aHenryJard, line 974:
```suggestion
                                label={module.enabled ? t_i18n('Enabled') : t_i18n('Disabled')}
```

aHenryJard (commented):
I just did a PR review, I did not tried it yet.

To avoid mispelling, I think it's better to have the 'ed' everywhere (enabled, disabled, activated etc..) so I did suggestions in that way.

labo-flg:
Let's enable all feature flags in the backend tests, to cover the code during development even if not enabled in production yet.

richard-julien:
> I just did a PR review, I did not tried it yet.
> 
> To avoid mispelling, I think it's better to have the 'ed' everywhere (enabled, disabled, activated etc..) so I did suggestions in that way.

That's a good point but there is more work and alignment to do than your proposal to really align everything around "enabled" like the modules that share the graphql object for the result. Modules "enable" is currently written in redis so the modification of the name must be changed carefully.

For now i will propose to only focus on the flag here and do the change of semantic in another PR

aHenryJard (approved):
Tested locally on Decay feature, it's working as expected.

I tested:
- without disabled_dev_features => decay hidden, ok
- with disabled_dev_features empty => decay enabled, ok
- with disabled_dev_features = ['INDICATOR_DECAY']  => decay hidden, ok

richard-julien:
> Let's enable all feature flags in the backend tests, to cover the code during development even if not enabled in production yet.

Done

labo-flg (approved):

Issue #5835, opened 2024-02-06
Default marking are not cleaned when marking is deleted, leading to full crashed platform


SamuelHassine:
## Description

1. Create a group
2. Add a some markings
3. Add a DEFAULT marking
4. Delete the marking (in marking definition list)

=> In Elastic, the rel_accesses-to is cleaned, but not the default_marking field, because it is not a rel_ and no logic has been implemented to do so.

Result as a user part of this group, platform is in full error state. Group is not accessible and cannot be updated. Need to clean default marking in Elastic

richard-julien:
Dont forget the associated backend test :)

PR #5839, opened 2024-02-06
[backend] Default marking are not cleaned when marking is deleted, leading to full crashed platform

jpkha:
#5835


lndrtrbn (changes requested):
Tested locally, the initial issue is fixed, I can use the platform with my user that had a deleted marking definition as default marking.

But it seems the fix broke something else on entity creation.
**Context:** taking the entity administrative area. If in settings we set default markings of area to "use user default markings" and the connected user is in a group with default markings.
**Action:** the user opens the form to create a new area.
**Expected:** the input markings is filled with those default markings.
**Actual:** the input is empty

lndrtrbn (approved):
Fix ok. Could be nice to have the review of a backend dev to check the code

PR #5845, opened 2024-02-07
[frontend] Fix multiple inconsistencies and errors on creation forms (#5741)

### LINKED ISSUE #5741: Multiple inconsistencies and errors on creation forms
## Description

On report, at opening creation, name is directly red in error state (should only happen after field is touched or on validation):

image

On threat actors, at opening everything seems correct but after the field is touched, only the red line is displayed, and the error does not replace the helper text:

image

Grouping:
detect duplicate doesn't work + context field appears to be touched but it is not
image

marieflorescontact:
### Proposed changes

* Refacto JS => TS pure function  `TextField` and `StixDomainObjectDetectDuplicates` components
* Fix AutoCompleteField ==> context fiel appears to be` touched` but it was not.
* we used `setTouched` instead of `setFieldTouched` wich does not set touched properly: we had `touched = {}`  wich was always truthy =>the display of the component in error (red) is conditioned to touched true/false, => it was always in error

### Related issues

* [issue/5741](https://github.com/OpenCTI-Platform/opencti/issues/5741)
*

### Checklist

### Further comments


opencti-platform/opencti-front/src/components/TextField.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+  const { value, ...otherProps }:MuiTextFieldProps = fieldToTextField(props);
+
+  return (
+    <MuiTextField
+      {...otherProps}
+      value={value ?? ''}
+      error={!isNil(meta.error) && meta.touched}
+      helperText={
+        (meta.value && meta.value.length > 2 && detectDuplicate) ? (
```
lndrtrbn, line 66: What is the difference between `meta.value` and `value` ?

opencti-platform/opencti-front/src/components/TextField.tsx
```diff
@@ -0,0 +1,76 @@
...
+  } = props;
+
+  const [, meta] = useField(name);
+
+  const internalOnChange = React.useCallback(
+    async (event: BaseSyntheticEvent) => {
+      const { value } = event.target;
+      await setFieldValue(name, value);
+      if (typeof onChange === 'function') {
+        onChange(name);
```
lndrtrbn, line 36: Why we do not return the value anymore?

opencti-platform/opencti-front/src/components/TextField.tsx
```diff
@@ -0,0 +1,76 @@
...
+      }
+    },
+    [onChange, setFieldValue, name],
+  );
+  const internalOnFocus = React.useCallback(async (event: BaseSyntheticEvent) => {
+    const { value } = event.target;
+    await setFieldValue(name, value);
+    if (typeof onFocus === 'function') {
+      onFocus(value);
```
lndrtrbn, line 45: That's strange passing the value instead of the name here

opencti-platform/opencti-front/src/components/TextField.tsx
lndrtrbn: And now that you are calling the `onChange` from Formik does it still work if we want our custom onChange callback event?

opencti-platform/opencti-front/src/components/TextField.tsx
```diff
@@ -0,0 +1,76 @@
...
+    }
+  }, [onFocus, name]);
+  const internalOnBlur = React.useCallback(
+    async () => {
+      await setFieldTouched(name, true);
+      if (typeof onBlur === 'function') {
+        onBlur(name);
```
lndrtrbn, line 52: Now that the `internalOnBlur ` callback does not call the custom `onSubmit` I'm not sure why do we need it at all?

opencti-platform/opencti-front/src/private/components/common/stix_domain_objects/StixDomainObjectDetectDuplicate.tsx
```diff
@@ -0,0 +1,183 @@
...
+interface StixDomainObjectDetectDuplicateProps {
+  types: string[];
+  value: string;
+}
+
+const StixDomainObjectDetectDuplicate: FunctionComponent<StixDomainObjectDetectDuplicateProps> = ({
+  types,
+  value,
+}) => {
+  const classes = useStyles();
```
lndrtrbn, line 52: Ideally do not use `makeStyle` which is deprecated

opencti-platform/opencti-front/src/private/components/common/stix_domain_objects/StixDomainObjectDetectDuplicate.tsx
```diff
@@ -0,0 +1,183 @@
...
+  useEffect(() => {
+    if (value.length > 2) {
+      fetchQuery(stixDomainObjectsLinesSearchQuery, {
+        types,
+        search: `"${value}"`,
+        count: 10,
+      })
+        .toPromise()
+        .then((data) => {
+          const duplicates = (data as StixDomainObjectsLinesSearchQuery$data)?.stixDomainObjects?.edges ?? [];
```
lndrtrbn, line 82: Use the `fetchQuery` generic argument to specify the type of data returned instead of casting manually here

lndrtrbn (commented):
For now I just made a quick review of the code. I let you resolve conflicts before I test on my side

PR #5866, opened 2024-02-08
[frontend] Added 'application/vnd.ms-excel' as an accepted MIME type in csv mapper

### LINKED ISSUE #5854: [CSV Mapper] TEST function always returns an error when trying to load a csv file on Firefox
## Description

When testing a CSV file in a CSV mapper on Firefox, the error "This file is not in the specified format" is always returned

## Environment

1. OS (where OpenCTI server runs):  Windows
2. OpenCTI version: 5.1.12.29
3. OpenCTI client: frontend
4. Other environment details: running on Firefox 122.0

## Reproducible Steps

1. Open any CSV mapper
2. Click on "TEST"
3. Select any csv file

## Expected Output

The data of the file, or an error if the data is in an incorrect format

## Actual Output

image

## Additional information

The problem doesn't occur on other browsers. 
The problem occurs because Firefox reads the csv file as a MIME type "application/vnd.ms-excel", whereas other browsers read the file as a "text/csv" file.
It seems to also be correlated with the fact that Windows reads the .csv files as an excel file, as seen in the other screenshots. It seems like Firefox trusts the OS with the type of the file, whereas other browsers have their own interpretation of the file type given the file extension.

## Screenshots (optional)

image

JeremyCloarec:
### Proposed changes

* This is the implementation of the fix proposed by [@lndrtrbn ] to the issue.

### Related issues

* #5854 

### Checklist

### Further comments

The drawback of thix fix is that it might confuse users by giving them the option to select .xls files to import.

An other possible fix would have been to drop the checking of the MIME type when importing the csv, and only check for the .csv file extension, but I'm not sure what all the impacts of not checking for the MIME type might be.

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/data/csvMapper/CsvMapperTestDialog.tsx
+++ b/opencti-platform/opencti-front/src/private/components/data/csvMapper/CsvMapperTestDialog.tsx
@@ -98,6 +98,7 @@ const CsvMapperTestDialog: FunctionComponent<CsvMapperTestDialogProps> = ({
           <CustomFileUploader
             setFieldValue={(field, v) => onChange(field, v)}
             label={'Your testing file limited to 100 lines (csv only, max 1MB)'}
+            // we also accept application/vnd.ms-excel type because that's how csv's seem to be seen as under WindowsOS + Firefox browser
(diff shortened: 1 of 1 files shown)


marieflorescontact (approved):
locally tested on firefox, works well 👌

Issue #3470, opened 2023-06-06
Organizations types to be in the vocabularies


RaulSokolova:
## Use case

We might need to add different types of organizations to identify them.

## Current Workaround

None

## Proposed Solution

Add a organization_type_ov for users to customize it

## If the feature request is approved, would you be willing to submit a PR?

Yes / No (Help can be provided if you need assistance submitting a PR)

Jipegien:
Can you specified your need plz? To what end do you want to add "types" to Organization and how the labels and "part of" relations do not cover it?

RaulSokolova:
We need to be able to add more. In the vocabularies is where we can add different types for each entity. 

![image](https://github.com/OpenCTI-Platform/opencti/assets/30608998/221db6c5-67ab-4b12-b4e5-36d9e9734886)

Jipegien:
Ok I was thinking about something else, my bad. We will add a organization_type_ov.

RaulSokolova:
Thank you for working on adding it, sorry for coming short explaining the feature request.

PR #5876, opened 2024-02-09
[backend/frontend] Organizations types to be in the vocabularies (#3470)

frapuks:
### Proposed changes

- Add Vocabulary for Organization type (Possibility to create, update or delete any type)
- Replace Field with OpenVocabField
- Remove type translate
- Display "unknown" in place of "other" for Organization overview without type
- Add neccessary for database initialisation and migration
- Allow define default value for organization type

### Related issues

- #3470 

### Checklist

### Further comments

Please pay particular attention to the fact that I've modified the counters to pass the tests


opencti-platform/opencti-graphql/src/migrations/1707207977998-vocabulary-add-organization-types.js
```diff
@@ -0,0 +1,29 @@
+import { executionContext, SYSTEM_USER } from '../utils/access';
```
SouadHadjiat: you will need to rename this migration to set a more recent timestamp than the last one merged on master (1707487253148-rename-data-sharing-capability-description.js).

opencti-platform/opencti-front/src/private/components/entities/organizations/OrganizationCreation.tsx
```diff
@@ -144,7 +145,7 @@ export const OrganizationCreationForm: FunctionComponent<OrganizationFormProps>
       name: '',
       description: '',
       x_opencti_reliability: undefined,
-      x_opencti_organization_type: 'other',
+      x_opencti_organization_type: undefined,
```
SouadHadjiat, line 148: why is it initialized at 'other' ? since vocabularies can be edited and deleted, 'other' might not even exist anymore.

frapuks: "Other" is the previous default value, when we still had hard choices.
Now the default value is "undefined", and works if you delete "other".

opencti-platform/opencti-graphql/src/migrations/1707207977998-vocabulary-add-organization-types.js
frapuks: OK, I'll check that everything is correct, and I'll modify it before merging to make sure I have the latest timestamp.

opencti-platform/opencti-front/src/private/components/entities/organizations/OrganizationCreation.tsx
SouadHadjiat: oh you're right sorry !

opencti-platform/opencti-graphql/src/migrations/1707207977998-vocabulary-add-organization-types.js
frapuks: Done

opencti-platform/opencti-front/src/private/components/entities/organizations/OrganizationEditionOverview.tsx
```diff
@@ -125,11 +154,11 @@ const OrganizationEditionOverviewComponent = (props) => {
     });
   };
 
-  const handleSubmitField = (name, value) => {
+  const handleSubmitField = (name: string, value: string | string[] | number | number[] | null) => {
```
SouadHadjiat: value might be an `Option` as I see in line 161, so in the list of possible types we should add `Option`, so you don't have then to cast with `as unknown`

SouadHadjiat:
Tested locally, all good ✅

opencti-platform/opencti-front/src/private/components/entities/organizations/OrganizationEditionOverview.tsx
frapuks: done, thank you

SouadHadjiat (approved):

PR #5881, opened 2024-02-09
[frontend] Confidence Level logic enforced in the platform (User) (#5697)

SarahBocognano:
### Proposed changes

* Add Confidence Field missing in several SDO in creation/edition form and overview
* Add a hook to check the confidence level of the user and compare it to each entity confidence level. If the user's confidence level is lower than the entity, then the user CAN'T edit the entity and an alert + error message are display in form.
* Refacto some component to use the check confidence hook with fieldPatch in useFormEditor 

### Checklist

### Further comments

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/analyses/malware_analyses/MalwareAnalysisEditionDetails.tsx
+++ b/opencti-platform/opencti-front/src/private/components/analyses/malware_analyses/MalwareAnalysisEditionDetails.tsx
@@ -4,6 +4,12 @@ import { Field, Form, Formik } from 'formik';
 import * as Yup from 'yup';
 import { FormikConfig } from 'formik/dist/types';
 import { GenericContext } from '@components/common/model/GenericContextModel';
+import Alert from '@mui/material/Alert';
+import {
+  malwareAnalysisEditionOverviewFocus,
+  malwareAnalysisMutationRelationAdd,
+  malwareAnalysisMutationRelationDelete,
+} from '@components/analyses/malware_analyses/MalwareAnalysisEditionOverview';
 import TextField from '../../../../components/TextField';
 import { SubscriptionFocus } from '../../../../components/Subscription';
 import CommitMessage from '../../common/form/CommitMessage';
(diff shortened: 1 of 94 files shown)


opencti-platform/opencti-front/src/private/components/analyses/malware_analyses/MalwareAnalysisEditionDetails.tsx
```diff
@@ -178,6 +205,15 @@ const MalwareAnalysisEditionDetails: FunctionComponent<MalwareAnalysisEditionDet
...
         <Form style={{ margin: '20px 0 20px 0' }}>
+          {(!checkConfidenceForEntity(malwareAnalysis) && (
+            <Alert severity="warning" variant="outlined"
+              style={{ marginTop: 20, marginBottom: 20 }}
+            >
+              {t_i18n(
+                'Your maximum confidence level is insufficient to edit this object.',
+              )}
+            </Alert>
+          ))}
```
labo-flg: I think you should wrap this into a dedicated component `<AlertConfidenceForEntity />`

That uses internally the hook useConfidenceLevel and simply returns null if check is ok.

labo-flg (approved):

labo-flg (commented):
I found a bug with the confidence field ; I'll invetsigate

opencti-platform/opencti-front/src/private/components/settings/users/UserConfidenceLevelField.tsx
```diff
@@ -128,6 +130,7 @@ const UserConfidenceLevelField: FunctionComponent<UserConfidenceLevelFieldProps>
         editContext={editContext}
         disabled={!switchValue || disabled}
         variant="edit"
+        maxLimit={effectiveConfidenceLevel?.max_confidence ?? 0}
```
labo-flg:
```suggestion
```

An admin shall not be limited to his own level. This is an administration feature...

lndrtrbn (approved):

labo-flg:
> I found a bug with the confidence field ; I'll invetsigate

the select options were not properly filtered according to the max limit, so a user could select a confidence above his own.
This is fixed now.

Issue #5808, opened 2024-02-05
Not possible to empty CVSS score in vulnerabilities


SamuelHassine:
## Description

1. Go to a vulnerability
2. Update the CVSS score
3. Try to empty the field

=> Nothing happens

PR #5931, opened 2024-02-12
[frontend] Not possible to empty CVSS score in vulnerabilities (#5808)

SarahBocognano:
### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/5808

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/arsenal/vulnerabilities/VulnerabilityEditionDetails.jsx
+++ b/opencti-platform/opencti-front/src/private/components/arsenal/vulnerabilities/VulnerabilityEditionDetails.jsx
@@ -69,8 +69,8 @@ const VulnerabilityEditionDetailsComponent = (props) => {
     const commitReferences = (references ?? []).map(({ value }) => value);
     const inputValues = Object.entries({
       ...otherValues,
-      x_opencti_cvss_base_score: parseFloat(String(values.x_opencti_cvss_base_score)),
       createdBy: values.createdBy?.value,
+      x_opencti_cvss_base_score: parseFloat(values.x_opencti_cvss_base_score),
       x_opencti_workflow_id: values.x_opencti_workflow_id?.value,
       objectMarking: (values.objectMarking ?? []).map(({ value }) => value),
     }).map(([key, value]) => ({ key, value: adaptFieldValue(value) }));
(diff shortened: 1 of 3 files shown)


Archidoit:
It's not working for me.. When I empty the field 'score' :
image
(the score is updated at refresh only)

Archidoit:
When there is no score, we should display it in the overview with just a '-' instead of a white square (like empty description) : 
image

@SarahBocognano : Not in this case ! When creating a vulnerability with no score we have the empty white bloc, it's hard coded in Vulnerability Details

CelineSebe:
Same remarks. 
Plus, when I clicked on another field after emptying the previous one, the CVSS score is correctly updated!

opencti-platform/opencti-front/src/private/components/arsenal/vulnerabilities/VulnerabilityEditionDetails.jsx
```diff
@@ -101,36 +92,33 @@ const VulnerabilityEditionDetailsComponent = (props) => {
...
+      vulnerabilityValidator
         .validateAt(name, { [name]: value })
         .then(() => {
-          commitMutation({
-            mutation: vulnerabilityMutationFieldPatch,
+          editor.fieldPatch({
             variables: {
               id: vulnerability.id,
-              input: { key: name, value: value || '' },
+              input: [{ key: name, value: value ?? '' }],
```
labo-flg: setting to an empty string won't work if you want to change the field to null.
Especially with numeric fields, the backend will try to parse this empty field and throw an error.

To actually set a nullable field to null (like the score here), you need to do this:
```suggestion
              input: [{ key: name, value: value ?? [null] }],
```

Note the use of an array for value. It's the type expected by GQL, but using a single value works because of type coercion (the little magic in gql.. from one single value it turns it into an array with this single value automatically).

labo-flg (changes requested):
Tested KO, but I added a comment to fix it I think.

opencti-platform/opencti-front/src/private/components/arsenal/vulnerabilities/VulnerabilityEditionDetails.jsx
frapuks: nice trick, it works

SarahBocognano: Done ! It works perfectly thank you !

CelineSebe (approved):

frapuks (approved):
Ok for principal feature.
Not in the case, I agree with Cathia about displaying a `-` if there are no values => VelnabilityDetails.jsx / l.123 `label = '-'

labo-flg (approved):

Issue #5841, opened 2024-02-06
Broken attack patterns matrix everywhere


SamuelHassine:
## Description

1. This issue leads multiple screens to not work correctly.
2. Do a query like:

```
query {
  attackPatterns(first: 5000) {
    edges {
      node {
        id
      }
    }
  }
}
```

Result is:

![image](https://github.com/OpenCTI-Platform/opencti/assets/1334279/faef69f1-a63b-44d1-824d-fb01d57bfe58)

Only 500 first are returned.

richard-julien:
This is expected behavior. If the attackPatterns ask for 5000 elements we need to rework the screen or create a specific API

Jipegien:
This current behaviour is not good enough. We need to fix this for the next major release. We are currently not consistent in our displaying of matrices and this is a major CTI use case.

PR #5937, opened 2024-02-13
[backend] Broken attack patterns matrix everywhere (#5841)

richard-julien:
See #5841

For now instead of changing the screen loading way, bring back the pagination window to 5000.
+ some adaptation to respect the other options
+ default pagination to 500  when nothing is specified.


lndrtrbn (commented):
Should we also modify the other usage of `first || ES_MAX_PAGINATION` in the app to use the new DEFAULT const instead?

- engine.js line 2546, 2601, 
- domain/container.js line 56
- elasticSearch-test.js line 397, 414, 474, 537, 554

PR #5941, opened 2024-02-13
[frontend] rework relative time with intl

jpkha:
Use intl instead


aHenryJard:
Will be integrated in another PR, thanks for the tip.

Issue #4903, opened 2023-11-14
Share Public Dashboard - MVP


nino-filigran:
## Use case

As mentioned in this [issue](https://github.com/OpenCTI-Platform/opencti/issues/2101), there is a need to share dashboard as a public link to be able to either embed them in a page or simply to provide access to some data to some users which do not have access to OpenCTI

PR #5951, opened 2024-02-13
[frontend/backend] Public dashboard chunk 2 (#4903)

lndrtrbn:
### Proposed changes

Context recap: possibility to publish a version of a dashboard that is visible by every having the url. No need to be connected to OpenCTI. In the previous chunk (merged) we defined the process to create public dashboard and implemented one widget to feature the scenario. The objective of this PR is to generalize to all the widgets we can have in a dashboard.

* [front] refactor all existing widgets (used in private dashboard) to avoid duplicating code and be able to generate easier their public versions used in public dashboard.
* [back] create the different APIs to be able to retrieve the data for widgets.
* [front] use those new APIs in public widgets called inside public dashboards.
* [back] some refactoring on private APIs where typescript errors were found.
* [back] deleting a private dashboard also delete its public dashboards.
* [back] tests CRUD public dashboard + basic usage of widgets.

### Related issues

* #4903 

### Checklist

### Further comments

80% of the lines are refactoring and new widgets on frontend side. And there is a lot, it's not possible to read all the changes so we invite to focus more on the backend code and manual testing for the widgets.

How to test?

- Create a dashboard and add it all sort of widgets for both entities and knowledge graph,
- Click on publish button (top right) to create some public dashboards,
- Play with markings and keep an eye on the difference of data displayed in the widgets,
- Delete some public dashboards, the URL should redirect to home or login page,
- Delete the private dashboard, all associated public URL should redirect to home or login page.


opencti-platform/opencti-graphql/src/utils/format.js
```diff
@@ -74,6 +74,11 @@ export const computeRangeIntersection = (a, b) => {
 
 export const minutesAgo = (minutes) => moment().utc().subtract(minutes, 'minutes');
 export const hoursAgo = (hours) => moment().utc().subtract(hours, 'hours');
+export const dayAgo = () => {
+  const currentDate = new Date();
+  currentDate.setDate(currentDate.getDate() - 1);
+  return currentDate.toISOString().split('T')[0];
+};
```
labo-flg:
```suggestion
export const dayAgo = (days = 1) => {
  const currentDate = new Date();
  currentDate.setDate(currentDate.getDate() - days);
  return currentDate.toISOString().split('T')[0];
};
```

labo-flg:
Could you move the widget definition in another file in the tests, please ?
It would make it easier to read, and focus on the test itself.

labo-flg (commented):
Pair reviewed, comments addressed.

Tested on several dashboards. 
Last remarks is about the date on the line graph: the date are technical in private dashboard, but seems to be the real creation date once public.

labo-flg (approved):

PR #5956, opened 2024-02-14
[frontend] disable workbench access

### LINKED ISSUE #5860: Workbench: user not authorised to access a workbench redirected to loggin page
## Requirement 
As per the last comment, this ticket which was previously a bug is now a feautre with the follwing scope: Analyst Workbench access and creation should not be linked to the Import Knowledge capability, but only Access Knowledge. The goal of this ticket is now to update the capability controling the Workbench

## Description

A user who does not have authorisation to access a workbench in an incident, for example, will still see the workbench. 
By clicking on it he will be redirected to the loggin page (while remaining logged in). It might be better to display a message when the user clicks: "_you do not have authorisation to access this workbench_", or not to display the workbench to an unauthorised user.

UPDATE: A user having access to an Entity should be able to create an Analyst Workbench to propose modification. Validation of this Analyst Workbench must be under Create/Update Knowledge.
 
## Environment

OpenCTI 5.12.29

## Reproducible Steps

1. Create a workbench in an incident with high right access
2. Go to the incident with a user who does not have the rights to consult the workbench in the incident
3. Click on the workbench

## Expected Output

A non-authorisation message is displayed to indicate that access rights are insufficient or the workbench is not displayed at all for the unauthorised user.

## Actual Output

The unauthorised user is redirected to the loggin page (without being logged out) after clicking on the workbench.

jpkha:
close #5860 

image

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/common/files/workbench/WorkbenchFileLine.jsx
+++ b/opencti-platform/opencti-front/src/private/components/common/files/workbench/WorkbenchFileLine.jsx
@@ -161,7 +161,7 @@ class WorkbenchFileLineComponent extends Component {
   }
 
   render() {
-    const { classes, t, file, dense, directDownload, nested, nsdt } = this.props;
+    const { classes, t, file, dense, directDownload, nested, nsdt, isAuthorizeUser } = this.props;
     const { displayDelete } = this.state;
     const { uploadStatus, metaData } = file;
     const { errors } = metaData;
@@ -176,7 +176,7 @@ class WorkbenchFileLineComponent extends Component {
           classes={{ root: nested ? classes.itemNested : classes.item }}
           button={true}
           component={isOutdated ? null : Link}
-          disabled={isProgress}
+          disabled={isProgress || !isAuthorizeUser}
           to={`/dashboard/data/import/pending/${toB64(file.id)}`}
         >
           <ListItemIcon>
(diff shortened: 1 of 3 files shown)


frapuks:
the buttons are disabled, but in the console I can click on the link and download the json.
image

labo-flg:
> the buttons are disabled, but in the console I can click on the link and download the json. image

This might be another issue 🤔
The access right is on the workbench not the file itself.
@frapuks Can you also download the file from the data / uploaded files section ?

labo-flg:
I confirm I can download the file anyway using the other panel
image

And my user has very limited access
image

labo-flg (dismissed):
Disabling the workbench works well.
If the ability to download the file is an issue, IMO it's another issue.

jpkha:
Maybe I should enable the download button? What I understand is the right between downloading resources and accessing the workbench could be different.

nino-filigran:
According to [this page](https://docs.opencti.io/latest/administration/users/), to be able to download ressources on data, you need the `Download knowledge export` (Download the exports generated in the entities (in the Data section).). Therefore, based on @labo-flg's case, the user should not be able to download the workbench. 
@Jipegien let me know if this is wrong.

Jipegien:
Mmh. Not sure about the solution. Wait my approval on this.

Jipegien:
That is not the right way to handle this problem. Don't merge this PR.

Details in comment of the issue.

PR #5958, opened 2024-02-14
[backend] Introduce fuzzy search and wildcard prefix options

richard-julien:
Introduce fuzzy search and wildcard prefix options.
As this kind of default search introduce high usage of elastic and a change in default search behavior, options are disabled by default.

```
  "elasticsearch": {
    ....
    "search_wildcard_prefix": false,
    "search_fuzzy": false,
    ....
  },
```

With search_fuzzy option activated, the search includes words with automatic fuzzy comparison depending on the word size.
With search_wildcard_prefix option activated the search will include words not starting with the word.

### DIFF
--- a/opencti-platform/opencti-graphql/config/default.json
+++ b/opencti-platform/opencti-graphql/config/default.json
@@ -212,6 +212,8 @@
     "engine_selector": "auto",
     "index_creation_pattern": "-000001",
     "search_ignore_throttled": false,
+    "search_wildcard_prefix": false,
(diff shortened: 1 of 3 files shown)


SamuelHassine:
@richard-julien I've updated the doc accordingly, it will be publish on the 6.0.X doc.

https://github.com/OpenCTI-Platform/docs/commit/34dbbde256c708baeae142b54f665b0e07aea429.

PR #5964, opened 2024-02-14
[backend/frontend] remove 'remove' operation from playbook (#5950)

### LINKED ISSUE #5950: [Playbooks] remove marking definition doesn't work
## Description

The action of removing marking definitions does not function in playbooks

## Environment

Testing

## Reproducible Steps

1. Have an entity with TLP:GREEN 
1. Create a playbook that does the following:
- Listen to edition
- Manipulate knowledge -> Remove - Marking definitions - TLP_GREEN
- Send to Ingestion
3. Start the playbook
4. Edit a field of the entity

## Expected Output

The marking TLP:GREEN is deleted

## Actual Output

The marking TLP:GREEN is not deleted

## Additional information

Because of this bug, the "Replace Marking Definition" doesn't work in some cases like when replacing TLP:RED by TLP:GREEN.
Since the remove doesn't work, the add is ignored as RED is higher that GREEN.
This case is interesting to check once the issue is fixed.

SouadHadjiat:
### Proposed changes

* operation "remove" is currently not supported by the platform, so we remove it from playbook operations
* display in the frontend action operations supported by the playbook component instead of hardcoded operations (I still kept a fallback on default operations in case the playbook component doesn't provide them)

### Related issues

* #5950 

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/data/playbooks/PlaybookAddComponents.jsx
+++ b/opencti-platform/opencti-front/src/private/components/data/playbooks/PlaybookAddComponents.jsx
@@ -231,8 +231,7 @@ const PlaybookAddComponentsContent = ({
                 value: n.value,
                 patch_value: n.value,
               })),
-            )
-                        }
+            )}
           />
         );
       case 'objectLabel':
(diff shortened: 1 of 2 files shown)


marieflorescontact (approved):
locally tested ok 👌

frapuks (approved):

PR #5975, opened 2024-02-15
[frontend] fix filters representation when filters with regardingOf (4974)

### LINKED ISSUE #5974: Display is broken when having regardingOf filter with label filter or Author
Go to entities.
Choose the "Label" filter and select a value (the label is correctly displayed).
Choose the "Regarding Of" filter and select a value.
=> The "Label" filter is displaying an ID instead.

Environment: Testing

image

Archidoit:
issue: https://github.com/OpenCTI-Platform/opencti/issues/5974

### DIFF
--- a/opencti-platform/opencti-graphql/src/utils/filtering/filtering-utils.ts
+++ b/opencti-platform/opencti-graphql/src/utils/filtering/filtering-utils.ts
@@ -101,19 +101,18 @@ export const extractFilterGroupValues = (inputFilters: FilterGroup, key: string 
   }
 
   const ids = [];
-  // regardingOf key is a composite filter id+type, values are [{ key: 'id', ...}, { key: 'type', ... }]
-  // we need to extract the ids that need representatives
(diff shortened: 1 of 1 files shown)


opencti-platform/opencti-graphql/src/utils/filtering/filtering-utils.ts
```diff
@@ -101,17 +101,19 @@ export const extractFilterGroupValues = (inputFilters: FilterGroup, key: string
   }
 
   const ids = [];
-  // regardingOf key is a composite filter id+type, values are [{ key: 'id', ...}, { key: 'type', ... }]
+  // regardingOf key is a composite filter id+type, values are [{ key: 'id', ...}, { key: 'relationship_type', ... }]
```
Archidoit:
```suggestion
  // regardingOf key is a composite filter id+type, values are [{ key: 'id', ...}, { key: 'type', ... }]
```

jpkha (approved):

PR #5988, opened 2024-02-15
[frontend] remove badge in the top bar trigger icon

Archidoit:
When there are unread notifications, the badge is present both on the notifcation icon and trigger icon of the topbar. It should only be on the notification icon.


aHenryJard (approved):
All good 👍

labo-flg (approved):

PR #5993, opened 2024-02-15
[backend] Fix orga sharing on Cases

### LINKED ISSUE #5960: [Data segregation] "Sharing with the organization" not applied to contained entities as previously
## Description

When I want to share a container (e.g. a report or a case) with an organization, the contained entities are not shared.

Previously, the entities (and relationships) contained in a container were automatically shared when the container was shared. 

## Environment

OCTI 5.12.31

## Reproducible Steps

1. Take a report with entities and relations inside
2. Share with an organization
3. Look at the entities contained in the report -> they are not shared

## Expected Output

The contained entities and relations are shared when I shared the container

Kedae:
Closes #5960


marieflorescontact (approved):
tested ok 👌

Issue #2859, opened 2023-02-10
Decay settings for Indicators scores


SamuelHassine:
## Use case

- Currently the Indicator’s score in OpenCTI can only change based on manual modification or upserting (or automation) but cannot change over time. Each indicators has a “valid_until” date. If the date is overdue, the Indicator is revoked (automatically with the revokeManager). It is a On/Off model that cannot represent the fact that an Indicator lose its “value” over time.
- Plus, it is not possible to react on particular moment of the score evolution of an Indicator, for example to trigger specific sharing or tagging.

## Proposed Solution

Admin can configure a decay formula for indicators matching a filter.
Decay formula are defined by acceleration factor and key reaction points. key reaction points are score values that need to be store in DB for possible reaction triggering (impossible to react on every possible score due to ressource consumption for millions of Indicators). 
A nice lifecycle curve allows user to display the lifecycle of an Indicator.

To implement:
- [MVP] Managing Decay Algorithms in the platform
- [MVP] View Details and Configuration of a Decay Algorithm
- [MVP] Enforce specific rights to manage Decay Algorithms
- [MVP] Providing Built-in Decay Algorithm
- [MVP] Applying Decay Algorithms to Indicators
- [MVP] Displaying impact of Decay algorithm on Indicator’s Overview
- [MVP] Creating a custom Decay Algorithm
- [MVP] Editing a custom Decay Algorithm
- [MVP] Deleting a custom Decay Algorithm

SamuelHassine:
@team: please also read #2774.

PR #6004, opened 2024-02-16
[backend/frontend] Indicator decay rules improvements (#2859)

aHenryJard:
### Proposed changes

* [frontend] Add a space in built-in rule Title (there was no space between the title and "active" chip)
* [frontend] add a tooltip to explain that if there is no indicator observable type filter, it will match all types. Add also "none = ALL" explanation in the input.
* [frontend] Add a maximum number to avoid sending backend invalid data (like 99999999999999999999999999999999999 for numeric data)
* [frontend] refactoring to extract IndicatorTypesField used in DecayRule creation & edition and to use it in indicator entity settings when trying to set a default value for indicator main observable type.
* [backend] select **first active and matching** rule by order (if the first one matches all indicators, we select this one) => inactive rules were not filtered before, and we used to select the first restrictive rule (ex: for url indicators, we selected the first rule matching specifically Urls, otherwise the first rule matching all. This shouldn't be the case anymore, we should select the first matching rule by order)
* [backend] Add test coverage on capacity required

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/2859

### What to test to review this PR
* Decay rule creation and edition
* Test that the right rule is selected when creating an indicator : should be the first **active** rule matching the indicator (highest order). Test with a rule matching all indicators at the top order, and test also with rules applying to specific indicator types. Test also that inactive rules are not used for newly created indicators.

### Checklist


opencti-platform/opencti-graphql/tests/utils/testQueryHelper.ts
```diff
@@ -18,3 +21,15 @@ export const queryAsAdminWithSuccess = async (request: { query: any, variables:
   expect(requestResult.errors, `This errors should not be there: ${requestResult.errors}`).toBeUndefined();
   return requestResult;
 };
+
+/**
+ * Execute the query as Participant (see testQuery.ts), and verify that access is forbidden.
```
aHenryJard:
```suggestion
 * Execute the query as some User (see testQuery.ts), and verify that access is forbidden.
```

opencti-platform/opencti-graphql/src/modules/decayRule/decayRule-domain.ts
```diff
@@ -336,12 +336,14 @@ export const initDecayRules = async (context: AuthContext, user: AuthUser) => {
...
+    .sort((a, b) => (b.order || 0) - (a.order || 0));
   const decayRule = orderedRules.find((rule) => {
-    if (!indicatorObservableType) {
-      return rule.decay_observable_types?.length === 0;
+    if (rule.decay_observable_types?.length > 0) {
+      // return first rule matching the indicator main observable type
+      return indicatorObservableType && rule.decay_observable_types.includes(indicatorObservableType);
     }
-    return rule.decay_observable_types?.includes(indicatorObservableType);
+    return true; // return first rule
```
JeremyCloarec, line 346: Now that rules applying to all types are always contained in decayRule, we probably don't need to check for it again on line 351, and we can always return the fallback decay if we found nothing with the first find

opencti-platform/opencti-front/src/private/components/settings/decay/DecayRuleValidator.ts
```diff
@@ -0,0 +1,14 @@
+import * as Yup from 'yup';
+
+const decayRuleValidator = (t: (value: string) => string) => Yup.object().shape({
+  name: Yup.string().min(2).required(t('This field is required')),
+  description: Yup.string().nullable(),
+  active: Yup.boolean(),
+  order: Yup.number().min(1).max(Number.MAX_SAFE_INTEGER),
+  decay_lifetime: Yup.number().min(1).max(Number.MAX_SAFE_INTEGER),
+  decay_pound: Yup.number().min(0).max(Number.MAX_SAFE_INTEGER),
```
JeremyCloarec: I tested the MAX_SAFE_INTEGER (9007199254740991) in all the fields, it works properly with the decay factor field, but it throws an "Update indexing fail" error in the other fields. I'm not sure what proper max number could be used though

opencti-platform/opencti-graphql/src/modules/decayRule/decayRule-domain.ts
SouadHadjiat: ah right

opencti-platform/opencti-front/src/private/components/settings/decay/DecayRuleValidator.ts
labo-flg: I'm unsure we should check such case actually (that's a path I'm not sure we want to follow... there is so much place to do so). IMO it is a platform for "educated" people, we can't check for all impossible values.

SouadHadjiat: yes I agree with you @labo-flg, I'm going to push a fix with a more reasonable check.

labo-flg: ... and (max for business) is different from (allowed max), which can also be different depending on where we llok at the stack (like @JeremyCloarec found with elastic).

If we have real business rules (for instance a confidence is always between 0 and 100) we might think about adding form validation (+ backend check, let's not forget API users).

opencti-platform/opencti-graphql/src/modules/decayRule/decayRule-domain.ts
SouadHadjiat: done

opencti-platform/opencti-front/src/private/components/settings/decay/DecayRuleValidator.ts
SouadHadjiat: I've changed the validation to have max values that make sense, but there aren't any specific check in the backend for min or max values.

labo-flg: Your changes look good enough to me 👍
For backend, it might also be excessive checking... 
Actually it's a broader issue. If there are business rules to enforce, we might as well enforce them through the schema - for instance a `min` and `max` options for numeric attributes. Something we might think of later, definitely not here :)

JeremyCloarec (approved):

labo-flg:
tested ok

labo-flg (approved):

Issue #6001, opened 2024-02-16
Source type filter on relationships with AND mode shouldn't give any result


Archidoit:
When filtering relationships on "source type" (or "target type" too) with a AND on 2 different types, we should not be given any result because the source type cannot be of 2 different types... but:

![image](https://github.com/OpenCTI-Platform/opencti/assets/146674147/7e2ed5e8-6547-4342-9df0-9260b0af04d4)

Filter sent is:

```json
{
  "count": 25,
  "orderMode": "desc",
  "orderBy": "created_at",
  "filters": {
    "mode": "and",
    "filters": [
      {
        "key": "entity_type",
        "values": [
          "stix-core-relationship"
        ],
        "operator": "eq",
        "mode": "or"
      }
    ],
    "filterGroups": [
      {
        "mode": "and",
        "filters": [
          {
            "key": "fromTypes",
            "values": [
              "Attack-Pattern",
              "Malware"
            ],
            "operator": "eq",
            "mode": "and"
          }
        ],
        "filterGroups": []
      }
    ]
  }
}
```

Archidoit:
same for filters: fromId, toId

PR #6008, opened 2024-02-16
[backend] enable local mode switching for nested relation filters (#6001)

Archidoit:
issue: https://github.com/OpenCTI-Platform/opencti/issues/6001


labo-flg (approved):
Tested OK

frapuks (approved):
it work

PR #6011, opened 2024-02-16
[frontend] Fix malware analyses unknown display

### LINKED ISSUE #6010: Malware analyses displayed as Unknown in filters list
## Description
Malware analyses are displayed as unknown in some locations: 
- filters in triggers

## Reproducible Steps

Create an instance trigger. Open the 'related entity' filter. Malware analyses are displayed as Unknown:

Create a trigger. Open the 'contains' filter. Malware analyses are displayed as Unknown:
image

Kedae:
Closes #6010

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/analyses/malware_analyses/Root.tsx
+++ b/opencti-platform/opencti-front/src/private/components/analyses/malware_analyses/Root.tsx
@@ -23,6 +23,8 @@ import FileManager from '../../common/files/FileManager';
 import StixCoreObjectHistory from '../../common/stix_core_objects/StixCoreObjectHistory';
 import StixCoreRelationship from '../../common/stix_core_relationships/StixCoreRelationship';
 import { useFormatter } from '../../../../components/i18n';
+import Breadcrumbs from '../../../../components/Breadcrumbs';
+import { defaultValue } from '../../../../utils/Graph';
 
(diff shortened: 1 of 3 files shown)


opencti-platform/opencti-front/src/utils/Graph.js
```diff
@@ -394,6 +394,7 @@ export const defaultValue = (n, fallback = 'Unknown') => {
       || n.attribute_abstract
       || n.opinion
       || n.value
+      || n.product
```
Archidoit: it's result_name that we want to display for malware analyses

Archidoit:
we should rather add ...on MalwareAnalyses {result_name} in the stix core object query of useSearchEntities.tsx

Archidoit (changes requested):
display the malware analyses result_name, not its product

opencti-platform/opencti-front/src/utils/Graph.js
Kedae: My bad, missed that. I assumed that the spread on the query was correct 😅
I'll fix that

Archidoit:
malware analyses backend representative is not correct (display the id instead of the result_name) : 
image

--> should add the malware analyses case in extractEntityRepresentativeName() of entity-representative.js

Archidoit (approved):

PR #6012, opened 2024-02-16
[backend] PoC: notify subscriptions after background tasks

### LINKED ISSUE #5479: fix(notifications): unread notification indicator stays on after all notifications are marked read
## Description

When you mark all notifications as read, it launched a background task that ultimately change the read flag to all your notifications.
When this task is done, the unread marker in the top right corner is still present.

Going to the notifs list and marking one of them unread, then read again, will make the marker disappear properly.

## Environment

OCTI 5.12.15 (Testing)

## Reproducible Steps

*  have some unread notifs ; the unread marker should be visible

image

* select all notifs, mark them as read, launch the background task
* wait for completion
* all notifs should be marked as read in the list

## Expected Output

Unread notification marker (red badge) should disappear.

## Actual Output

Marker still present

## Additional information

https://github.com/OpenCTI-Platform/opencti/assets/146674147/fd3ba404-41cd-42fd-9909-a43782351f3f

labo-flg:
DRAFT FOR DISCUSSIONS ONLY, DO NOT MERGE

### Proposed changes

* register callbacks during execution of background tasks, depending on the action and element being processed
* only 1 callback can be registered for a given `entity_type`
* execute these callbacks when job is done

### Related issues

* #5479


labo-flg:
After discussions with @richard-julien this is not a proper way to do the job.

* the tasks are batched by blocks of 500, so we'll call this callback several times potentially
* a deeper reflection is ongoing over the refactoring of background tasks (rely on the worker ? etc.)
* another solution would be to conceptualize a bus on which the frontend could subscribe and always listen to "do stuff" according to "platform messages". We could leverage this for many things including knowing when a background task ends, show some feedback, call a query, etc.

Closing this PR.

PR #6033, opened 2024-02-19
[backend/frontend] user's confidence level shall be the highest among user's group(s)

labo-flg:
This specific point of the specifications have been rediscussed, as we want to favor consistency in the platform.

To align on how we handle markings or capabilities, we will take the highest max confidence level among the user's group(s) as the user's effective max confidence level.

Note that if a user has a specific level defined, it will overwrite this level from their groups ; this mechanism does not change.

Finally, if the user has `BYPASS` capability, we consider they have a max confidence level to 100, allowing for all modification/creation in the platform.

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/settings/users/UserConfidenceLevel.tsx
+++ b/opencti-platform/opencti-front/src/private/components/settings/users/UserConfidenceLevel.tsx
@@ -6,6 +6,7 @@ import Tooltip from '@mui/material/Tooltip';
 import { InformationOutline } from 'mdi-material-ui';
 import Box from '@mui/material/Box';
 import { useFormatter } from '../../../../components/i18n';
+import useGranted, { BYPASS } from '../../../../utils/hooks/useGranted';
 
 type Data_UserConfidenceLevel = User_user$data['user_confidence_level'];
 type Data_EffectiveConfidenceLevel = User_user$data['effective_confidence_level'];
@@ -16,6 +17,8 @@ type UserConfidenceLevelProps = {
 
 const UserConfidenceLevel: React.FC<UserConfidenceLevelProps> = ({ confidenceLevel }) => {
   const { t_i18n } = useFormatter();
+
+  const isGrantedBypass = useGranted([BYPASS]);
 
   if (!confidenceLevel) {
     return (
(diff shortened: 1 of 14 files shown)


opencti-platform/opencti-graphql/src/utils/confidence-level.ts
```diff
@@ -14,6 +15,15 @@ type ObjectWithConfidence = {
 };
 
 export const computeUserEffectiveConfidenceLevel = (user: AuthUser) => {
+  // if a user has BYPASS capability, we consider a level 100
+  if (isUserHasCapability(user, BYPASS)) {
```
SouadHadjiat: I suggest to use `isBypassUser(user)` instead of `isUserHasCapability(user, BYPASS)`, since in `isUserHasCapability` we first test `isBypassUser`.

opencti-platform/opencti-front/src/private/components/settings/users/UserConfidenceLevel.tsx
```diff
@@ -32,6 +35,8 @@ const UserConfidenceLevel: React.FC<UserConfidenceLevelProps> = ({ confidenceLev
   const renderSource = () => {
     const source = (confidenceLevel as Data_EffectiveConfidenceLevel)?.source;
 
+    console.log(isGrantedBypass);
```
SouadHadjiat: console.log to remove

opencti-platform/opencti-front/src/private/components/settings/users/UserConfidenceLevel.tsx
```diff
@@ -55,11 +52,18 @@ const UserConfidenceLevel: React.FC<UserConfidenceLevelProps> = ({ confidenceLev
           </Tooltip>
         );
       }
-      // the user himself
+
+      // source is the user himself, most probably by setting at the user level
+      let title = t_i18n('The Max Confidence Level is currently defined at the user level. It overrides Max Confidence Level from user\'s groups.');
+      // ... or if user has BYPASS, it's fixed to 100; check below is a cheap proxy (as we do not have user's capabilities in UserEditionOverview
+      if (!user.user_confidence_level || user.user_confidence_level.max_confidence < 100) {
```
labo-flg, line 59: having a precise check on `user.capabilities` won't work immediately within `UserEditionOverview`. 

One thing leading to another, it requires returning capabilities from the field patch mutation, which is not made that way (it only returns unresolved object). I'm not inclined to modify this behavior and call `buildCompleteUser` after `updateAttribute`. Seems too costly and deviating from our initial strategy (specific mutations corresponding to resolved fields).

SouadHadjiat (approved):
tested locally, it works well ✅

Just a minor suggestion : there is a missing translation for this message "The user has BYPASS capability, their max confidence level is set to 100." 

image

labo-flg:
> tested locally, it works well ✅
> 
> Just a minor suggestion : there is a missing translation for this message "The user has BYPASS capability, their max confidence level is set to 100."

Fixed ! thanks

PR #6038, opened 2024-02-19
[frontend] Fix creation page + align CSV Mapper option panel

### LINKED ISSUE #5741: Multiple inconsistencies and errors on creation forms

Kedae:
Closes #5741


marieflorescontact (commented):
bug found for `switchField`:
In ` settings > security > Users > your User > update > Enable max confidence level`
it closes immediately the update panel

update: well...it appears to be my user...

marieflorescontact (approved):

PR #6041, opened 2024-02-19
[backend] CSV Mapper - fix how to retrieve entity settings (#6034)

### LINKED ISSUE #6034: CSV Mapper - Required fields are not highlighted in relationship representations
## Description

In the form to configure a CSV Mapper, specifically in a representation of a relationship: except for `from` and `to`, required fields are not highlighted.

## Environment

2. OpenCTI version: 5.12.32
3. OpenCTI client: frontend
4. Other environment details:

## Reproducible Steps

1. Go to Settings > Customization > Relationships,
2. Set some attributes mandatory,
3. Go to Data > Processing > CSV Mappers,
4. Create a mapper and add a relationship representation, open it.

## Expected Output

The fields you set mandatory should have a red asterix aside their name and highlighted in red if not set.

## Actual Output

No asterix or highlight.

lndrtrbn:
### Proposed changes

* Change how we retrieve the entity settings to be able to fetch parent type when necessary (for stix core relationships)

### Related issues

* #6034 

### Checklist

### Further comments

Use the function `getEntitySettingFromCache` instead of directly calling the cache with `getEntitiesListFromCache` because the first one manage the fact to fetch entity settings from parent type which is not the case with the second one. So need to readapt the code to work with this function.

### DIFF
--- a/opencti-platform/opencti-graphql/src/domain/attribute-utils.ts
+++ b/opencti-platform/opencti-graphql/src/domain/attribute-utils.ts
@@ -61,6 +61,7 @@ export const INTERNAL_ATTRIBUTES = [
   'authorized_authorities',
   'precision',
   'pattern_version',
+  'connections',
   // X - Mitre
   'x_mitre_permissions_required',
   'x_mitre_detection',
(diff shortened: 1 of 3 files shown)


JeremyCloarec (commented):
Tested OK on the front end side.
As for the test failing, I think the cache doesn't contain the update done in the beforeAll query? I'm not familiar enough with the cache behavior to be sure though

JeremyCloarec (approved):

PR #6046, opened 2024-02-20
[backend] validate elUpdateElement input value against schema before indexing (#5696)

labo-flg:
### Proposed changes

* insert a validation step at the start of engine `elUpdateElement` function
* validate the payload before indexing, against the schema

### Related issues

* #5696


opencti-platform/opencti-graphql/src/schema/attribute-definition.ts
```diff
@@ -148,10 +148,10 @@ export const files: AttributeDefinition = {
...
-    { name: 'order', label: 'Order in carousel', type: 'numeric', precision: 'integer', editDefault: false, multiple: false, mandatoryType: 'external', upsert: true, isFilterable: true },
     { name: 'version', label: 'Version', type: 'date', editDefault: false, mandatoryType: 'no', multiple: true, upsert: true, isFilterable: true },
     { name: 'mime_type', label: 'Mime type', type: 'string', format: 'short', editDefault: false, mandatoryType: 'no', multiple: true, upsert: true, isFilterable: true },
-    { name: 'inCarousel', label: 'Include in carousel', type: 'boolean', mandatoryType: 'internal', editDefault: false, multiple: false, upsert: true, isFilterable: true },
+    { name: 'inCarousel', label: 'Include in carousel', type: 'boolean', mandatoryType: 'no', editDefault: false, multiple: false, upsert: true, isFilterable: true },
+    { name: 'order', label: 'Order in carousel', type: 'numeric', precision: 'integer', mandatoryType: 'no', editDefault: false, multiple: false, upsert: true, isFilterable: true },
```
labo-flg, line 154: `inCarousel` and `order` were mandatory in schema but not in API. 
I guess it's a mistake on the schema side.

opencti-platform/opencti-graphql/src/domain/stixCoreObject.js
```diff
@@ -402,8 +402,7 @@ export const stixCoreObjectImportPush = async (context, user, id, file, noTrigge
     const eventFile = storeFileConverter(user, up);
     const files = [...(previous.x_opencti_files ?? []).filter((f) => f.id !== up.id), eventFile];
     await elUpdateElement({
-      _index: previous._index,
-      internal_id: internalId,
+      ...previous,
```
labo-flg: this is to prevent the error that is thrown due to new check on `element.entity_type` existence in my validate function.

This makes me wonder if this was properly working in the first place ! because then it would replace in elastic with missing pieces of the object.

@richard-julien wdyt?

opencti-platform/opencti-graphql/src/database/engine.js
```diff
@@ -3316,12 +3320,13 @@ const elUpdateConnectionsOfElement = async (documentId, documentBody) => {
...
 export const elUpdateElement = async (instance) => {
   const esData = prepareElementForIndexing(instance);
+  validateDataBeforeIndexing(esData);
   const dataToReplace = R.dissoc('representative', esData);
   const replacePromise = elReplace(instance._index, instance.internal_id, { doc: dataToReplace });
   // If entity with a name, must update connections
   let connectionPromise = Promise.resolve();
   if (esData.name && isStixObject(instance.entity_type)) {
-    connectionPromise = elUpdateConnectionsOfElement(instance.internal_id, { name: esData.representative.main });
+    connectionPromise = elUpdateConnectionsOfElement(instance.internal_id, { name: esData.representative?.main ?? esData.name });
```
labo-flg: unsure of that one.

Happens when using `stixCoreObjectImportPush`: representative is undefined

lndrtrbn: But the representative is not something computed? With the name especially, so I don't understand how we can have a name without a representative

opencti-platform/opencti-graphql/src/schema/attribute-definition.ts
lndrtrbn: To verify with someone who knows

opencti-platform/opencti-graphql/src/schema/schema-attributes.ts
```diff
@@ -218,3 +217,82 @@ export const isObjectFlatAttribute = (k: string): boolean => {
 export const isMultipleAttribute = (entityType: string, k: string): boolean => (
   k.startsWith(RULE_PREFIX) || schemaAttributesDefinition.isMultipleAttribute(entityType, k)
 );
+
+// -- utility functions independent of attribute registration --
+// (inner mappings are not registered like first-level attribute)
+
+export const isMandatoryAttributeMapping = (schemaDef: AttributeDefinition) => schemaDef.mandatoryType === 'external' || schemaDef.mandatoryType === 'internal';
+
+export const isNonFLatObjectAttributeMapping = (schemaDef: AttributeDefinition) : schemaDef is (ObjectAttribute | NestedObjectAttribute) => { // handy typeguard
```
lndrtrbn: Can't we use the type `ComplexAttributeWithMappings` that already declare this union of types? Maybe it's not relevant, you tell me

lndrtrbn: Also a typo in the name of the function

opencti-platform/opencti-graphql/src/schema/schema-attributes.ts
```diff
@@ -218,3 +217,82 @@ export const isObjectFlatAttribute = (k: string): boolean => {
...
+      return; // nothing to check (happens on 'remove' operation for instance
+    }
+
+    const inputValues = Array.isArray(input) ? input : [input];
+    inputValues.forEach((value) => {
+      // check the value adhere to its mapping
+      const valueKeys = Object.keys(value);
+      schemaDef.mappings.forEach((mapping) => {
+        // mandatory fields: the value must have a field with this name
+        if (isMandatoryAttributeMapping(mapping) && !valueKeys.includes(mapping.name)) {
```
lndrtrbn, line 260: what if the input is like: `{ myKey: null }`, `myKey` is include in the array of keys but the value is still invalid no?

lndrtrbn: ok maybe this check is done in the recursive step below

opencti-platform/opencti-graphql/tests/01-unit/database/validateDataBeforeIndexing-test.ts
```diff
@@ -0,0 +1,124 @@
...
+    updated_at: '2024-02-23T09:57:32.006Z'
+  };
+
+  it('validates correct payloads', () => {
+    expect(() => validateDataBeforeIndexing(malware)).not.toThrowError();
+    expect(() => validateDataBeforeIndexing(threatActorIndividual)).not.toThrowError();
+    expect(() => validateDataBeforeIndexing(user)).not.toThrowError();
+  });
+
+  it('throws error on invalid payloads', () => {
```
lndrtrbn, line 60: I'm more about one test for each case but ok, not sure this is really important if it works well

lndrtrbn (commented):
Using this (imo valid) input:
```
{
  "id": "38fe587f-a73b-46ad-856f-cc2710246d42",
  "input": {
    "key": "user_confidence_level",
    "object_path": "/user_confidence_level/overrides",
    "value": [
      {
        "entity_type": "Administrative-Area",
        "max_confidence": 60
      },
      {
        "entity_type": "Country",
        "max_confidence": 40
      }
    ]
  }
}
```

I have the following error:
```
{
      "message": "Validation error",
      "name": "VALIDATION_ERROR",
      "time_thrown": "2024-02-26T13:56:35.665Z",
      "data": {
        "http_status": 500,
        "genre": "BUSINESS",
        "field": "user_confidence_level",
        "message": "Attribute user_confidence_level cannot be multiple",
        "data": {
          "key": "user_confidence_level",
          "object_path": "/user_confidence_level/overrides",
          "value": [
            {
              "entity_type": "Administrative-Area",
              "max_confidence": 60
            },
            {
              "entity_type": "Country",
              "max_confidence": 40
            }
          ]
        }
      }
    }
```

Like if the `object_path` was not taking in account

labo-flg:
This seems like a valid input yes.
I'll investigate to see how it comes this payload is produced. If it's a valid payload, it's a bug on the validation process. 
Otherwise... Maybe we do not handle objectpath on multiple like that (only on indexed path).

opencti-platform/opencti-graphql/src/schema/schema-attributes.ts
labo-flg: you're right !

opencti-platform/opencti-graphql/src/schema/schema-attributes.ts
labo-flg: yes, in recursion it's caught

labo-flg:
I've pushed a commit to allow such case @lndrtrbn, and added some integration tests.

opencti-platform/opencti-graphql/src/schema/schema-attributes.ts
```diff
@@ -218,3 +217,82 @@ export const isObjectFlatAttribute = (k: string): boolean => {
...
+      if (schemaDef.multiple && !Array.isArray(input)) {
+        throw FunctionalError(`Validation against schema failed on attribute [${schemaDef.name}]: value must be an array`, { value: input });
+      }
+      if (!schemaDef.multiple && (Array.isArray(input) || !R.is(Object, input))) {
+        throw FunctionalError(`Validation against schema failed on attribute [${schemaDef.name}]: value must be an object`, { value: input });
+      }
+    }
+    if (!isMandatory && R.isNil(input)) {
+      return; // nothing to check (happens on 'remove' operation for instance
+    }
```
lndrtrbn: The last test you wrote failed because the function checks multiple constraint only if the attribute is mandatory. But, thanks to your test, I think it is a mistake because in the case the attribute is an object, not mandatory, and you give an input which is not an object then we don't throw.

The suggestion is to first check if not mandatory and no value then return, otherwise check the type of the input no matter what the mandatory type is.

```suggestion
    if (!isMandatory && R.isNil(input)) {
      return; // nothing to check (happens on 'remove' operation for instance
    }
    // check 'multiple' constraint
    if (schemaDef.multiple && !Array.isArray(input)) {
      throw FunctionalError(`Validation against schema failed on attribute [${schemaDef.name}]: value must be an array`, { value: input });
    }
    if (!schemaDef.multiple && (Array.isArray(input) || !R.is(Object, input))) {
      throw FunctionalError(`Validation against schema failed on attribute [${schemaDef.name}]: value must be an object`, { value: input });
    }
```

lndrtrbn (approved):

opencti-platform/opencti-graphql/tests/01-unit/database/validateDataBeforeIndexing-test.ts
```diff
@@ -0,0 +1,124 @@
...
+
+  const user = {
+    _index: 'opencti_internal_objects-000001',
+    internal_id: '421781aa-52cb-4019-abf1-3f8c3c8617bd',
+    entity_type: 'User',
+    representative: {
+      main: 'Plop'
+    },
+    user_confidence_level: {
+      max_confidence: 73, // missing mandatory field in a single object
```
labo-flg:
```suggestion
      max_confidence: 73,
```

opencti-platform/opencti-graphql/src/schema/schema-attributes.ts
```diff
@@ -218,3 +217,80 @@ export const isObjectFlatAttribute = (k: string): boolean => {
...
 );
+
+// -- utility functions independent of attribute registration --
+// (inner mappings are not registered like first-level attribute)
+
+export const isMandatoryAttributeMapping = (schemaDef: AttributeDefinition) => schemaDef.mandatoryType === 'external' || schemaDef.mandatoryType === 'internal';
+
+export const isNonFlatObjectAttributeMapping = (schemaDef: AttributeDefinition) : schemaDef is ComplexAttributeWithMappings => { // handy typeguard
+  return schemaDef.type === 'object' && schemaDef.format !== 'flat';
+};
```
labo-flg, line 228: @richard-julien following our discussions on these utility functions. Here I had the same issue as the existing utilities were only for first-level schema attribute defs.

I can refactor all of this in another PR I think.

richard-julien: Ok. Thats a good first round on this

PR #6055, opened 2024-02-20
[frontend] use of istanbul plugin dynamic

jpkha:
use of istanbul plugin dynamic.
By default do not use the plugin as we are just using for the coverage.


labo-flg:
Thanks ! Shall we set this env variable in the package.json scripts that run the tests then ?

labo-flg (approved):

PR #6064, opened 2024-02-21
[backend/frontend] entity_type filter available for abstract types

Archidoit:
Issue (on testing only) : entity_type filter not available in Data>Entities, Observations>Observables, Global Search


Kedae:
Can we have a small e2e test for that ? :)

Archidoit:
@Kedae I added a backend test since the problem is only from the back, but I can also add an e2e :)

JeremyCloarec (approved):
Tested OK

PR #6073, opened 2024-02-22
[backend] Fix observables query for network traffic with dst_port (#6070)

### LINKED ISSUE #6070: Having Network Traffic observable with a dst ref makes the observable listing crash
## Description

Discovered while working on #6056 
If a Network Traffic observable has a `dst` ref set, the pagination request in the observable list crashes.
=> Update : the bug happens is a network traffic has "dst_port" and we try to sort the observable list by "representative".

## Environment

latest 5.12

## Reproducible Steps

1. create a network traffic observable
2. from its knowledge tab, add a dst ref
3. go to observables list and sort by observable value (dashboard/observations/observables?sortBy=observable_value&orderAsc=true)

## Expected Output

no crash :)

## Actual Output

crash :(

SouadHadjiat:
### Proposed changes

* dst_port is an integer, we should not try to query it with ".keyword", it's also the case of other fields like "number", so we have to rewrite completely the script to handle all types of fields.
* Add an integration test create a network traffic with dst_port and to list observables ordered by "observable_value"

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/6070

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/utils/format.js
+++ b/opencti-platform/opencti-graphql/src/utils/format.js
@@ -231,7 +231,7 @@ export const runtimeFieldObservableValueScript = () => {
        } else {
          emit('Unknown')
        }
-     } else if (type == 'process') {
+    } else if (type == 'process') {
        if (have(doc, 'pid')) {
          emit(doc['pid.keyword'].value)
        } else if (have(doc, 'command_line')) {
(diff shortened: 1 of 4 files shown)


opencti-platform/opencti-graphql/src/utils/format.js
```diff
@@ -155,150 +155,64 @@ export const observableValue = (stixCyberObservable) => {
...
+    def getFieldValue(def doc, def key) {
+      if (!doc.containsKey(key)) {
+        return 'Unknown';
+      }
+      if (doc.containsKey(key + '.keyword')) {
+        if (doc[key + '.keyword'].size()!=0) {
+          return doc[key + '.keyword'].value;
+        }
+      } else if (doc[key].size()!=0) {
+        return String.valueOf(doc[key].value);
```
SouadHadjiat, line 167: we need to return a String value because the type of the runtime field is "keyword". Otherwise we have a cast exception

PR #6076, opened 2024-02-22
[frontend] disable field to edit default value in settings if needed (#6037)

### LINKED ISSUE #6037: Entity settings display edit default value input even if attribute has "editDefault" false
## Description

In Settings => customization => Entity types, when selecting an entity type that has an attribute with "mandatory: external" and "editDefault: false", you'll be able to edit the attribute and set a default value.

## Environment

2. OpenCTI version: latest master (5.12.32)
3. OpenCTI client: frontend
4. Other environment details:

## Reproducible Steps

Issue visible locally only

## Expected Output

We shouldn't edit the default value if it's editDefault is set to false.

lndrtrbn:
### Proposed changes

* disable field to edit default value in settings if not editable
* disable open form if can neither edit mandatory or default value

### Related issues

* #6037 

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/common/form/OpenVocabField.tsx
+++ b/opencti-platform/opencti-front/src/private/components/common/form/OpenVocabField.tsx
@@ -19,6 +19,7 @@ interface OpenVocabProps {
   onFocus?: (name: string, value: Option) => void;
   containerStyle?: Record<string, string | number>;
   editContext?: unknown;
+  disabled?: boolean;
   queryRef: PreloadedQuery<OpenVocabFieldQuery>;
   onChange?: (name: string, value: string | string[]) => void;
   onSubmit?: (name: string, value: string | string[]) => void;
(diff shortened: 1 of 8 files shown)


opencti-platform/opencti-front/src/private/components/common/form/OpenVocabField.tsx
```diff
@@ -60,6 +61,7 @@ Omit<OpenVocabProps, 'type'>
   containerStyle,
   editContext,
   queryRef,
+  disabled = false,
```
frapuks, line 64: Already initialized to false into `DefaultValueField.tsx`, not necessary here. Or if it's really necessary here, isn't it also necessary in the other files (KillChainPhasesField, ObservableTypesField) ?

lndrtrbn: Indeed it's not necessary as in JS the boolean value of undefined is false. Still I think it's a good practice to initialize the value to a boolean as it is optional. Could be done in other components but nothing mandatory

marieflorescontact (approved):
tested locally by playing with attributes definition. looks good. 👌

PR #6080, opened 2024-02-22
Prevent input name collision with db name for refs (#6056)

### LINKED ISSUE #6056: Worker error when importing Network-Traffic object with nested properties
We receive the error below when importing a Network-Traffic object via Worker from a stix bundle:

{'name': 'DATABASE_ERROR', 'message': 'Error in store update event'}

image

In the Worker logs we find the following error message:

`
ERROR:worker:{'name': 'DATABASE_ERROR', 'message': 'Error in store update event'}
Traceback (most recent call last):
  File "/opt/opencti/./worker/worker.py", line 263, in data_handler
    self.api.stix2.import_bundle_from_json(
  File "/usr/local/lib/python3.10/dist-packages/pycti/utils/opencti_stix2.py", line 215, in import_bundle_from_json
    return self.import_bundle(
  File "/usr/local/lib/python3.10/dist-packages/pycti/utils/opencti_stix2.py", line 2421, in import_bundle
    self.import_observable(item, update, types)
  File "/usr/local/lib/python3.10/dist-packages/pycti/utils/opencti_stix2.py", line 1066, in import_observable
    self.opencti.stix_nested_ref_relationship.create(
  File "/usr/local/lib/python3.10/dist-packages/pycti/entities/opencti_stix_nested_ref_relationship.py", line 266, in create
    result = self.opencti.query(
  File "/usr/local/lib/python3.10/dist-packages/pycti/api/opencti_api_client.py", line 348, in query
    raise ValueError(
ValueError: {'name': 'DATABASE_ERROR', 'message': 'Error in store update event'}
`

No associated error is found in the Platform logs

After some testing with the code that generates the bundles, we find:

The problem is related to the network-traffic object.
It only appears if we add src_ref, start or end properties. (we do not add dst_ref, haven’t checked the case).
If we create the object without those properties, it is created correctly.
If we only add start and end properties, the object is created correctly.
If we only add src_ref, most objects are created correctly, but the error still appears (let’s say 20% of the time).
If we add src_ref, start and end, ALL network-traffic object creation fails.

## Environment

Ubuntu 22.04 baremetal installation. All platform components in different VMs.
Current platform version 5.12.32

## Reproducible Steps

Difficult to reproduce as the same bundle that gives the error when importing via Worker, does not give an error when imported via WorkBench

## Expected Output

A nice network-traffic object with associated src_ref, start and end time. 

## Actual Output

The errors above

## Additional information

There are similar reports in Filigran's Slack, posted after ours:
https://filigran-community.slack.com/archives/C06CF1N302W/p1708432322249959

labo-flg:
Several refs attributes have the same input name as their db name, resulting in bugs during stream event publishing.

If the attribute is multiple, the event published in stream contains `null` in patch.
If it's a single value, the stream event won't be sent and an error will be thrown (`store update event error`)

### Proposed changes

* on schema registration, check the names collisions and throw an error 
* fix all collisions found by prefixing the name
* rename API keys that require changes (--> only `MalwareAnalysisAddInput.sample` to rename `MalwareAnalysisAddInput.analysisSample`

### Related issues

* #6056 

### Checklist

### Further comments

e2e tests would be added in #6073 to cover network traffic creation + ref, I won't add it here (or improve on it here once merge and rebase)

To test this PR, test the following:
* open the raw stream at http://localhost:3000/stream
* Create an observable > network traffic, with a random dst port
* Open it, go to knowledge, and add a nested object
* check single ref
  * select an ipv4 address, and create a "src" relationship (not dst, you will end up in #6070)
  * BEFORE --> store update event, NOW -> no error
  * check in stream the update event is correct
* check multiple ref
  * create 2 other network traffic using just some random DST port values
  * select the inital network traffic, and create 2 "encapsulates" relationship with the 2 other network traffic
  * BEFORE --> no store update event but NULL in the stream events, NOW -> no error, correct values in the stream events


opencti-platform/opencti-graphql/src/schema/stixRefRelationship.ts
labo-flg, line 1: None of the input names modified here are exposed in the graphql API

The refs can only be created through `stixRefRelationShipAdd`, using `relationship_type` key and it does not need to change there.

.drone.yml
```diff
@@ -32,16 +32,24 @@ steps:
     commands:
       - apk add build-base git libffi-dev cargo
       - pip3 install --upgrade setuptools
+      - cd opencti-platform
+      - OPENCTI_COMMIT=$(git rev-parse --short HEAD)
+      - echo [INFO] using opencti@$DRONE_COMMIT_BRANCH:$OPENCTI_COMMIT
```
labo-flg, line 37: just an improvement on the way, to finally have a clear feedback on what is the precise version used.

opencti-platform/opencti-front/src/schema/relay.schema.graphql
```diff
@@ -10163,7 +10163,7 @@ input MalwareAnalysisAddInput {
   result_name: String!
   result: String
   analysisSco: [String]
-  sample: String
+  analysisSample: String
```
labo-flg, line 10183: This is the only breaking change to the API.

All other renaming are on attribute never exposed in the API for now.

opencti-platform/opencti-graphql/src/database/middleware.js
```diff
@@ -601,7 +600,7 @@ const inputResolveRefs = async (context, user, input, type, entitySetting) => {
           fetchingIds.push(...elements);
           expectedIds.push(...elements.map((e) => e.id));
         } else { // Single
-          if (dst === INPUT_FROM && isStixRefRelationship(type)) {
+          if (dst === 'from' && isStixRefRelationship(type)) {
```
labo-flg, line 603: this one was tricky.

"from" is a key used in 2 different context: it's the "from" of a relationship, and the "from" in an email message.

`INPUT_FROM` was defined for use with email message only, but it was also imported and used in places related to relationship. 

Now that we renamed for email to `emailFrom` it's not working anymore for relationships. But it was never meant to work like that in the first place.

opencti-platform/opencti-front/src/schema/relay.schema.graphql
richard-julien: Why not objectSample?

opencti-platform/opencti-graphql/src/database/middleware.js
richard-julien: Now that email is now on INPUT_EMAIL_FROM, are you sure is still needed to not used the constant?

opencti-platform/opencti-graphql/src/schema/stixRefRelationship.ts
```diff
@@ -54,31 +54,31 @@ export const ABSTRACT_STIX_NESTED_REF_RELATIONSHIP = 'stix-nested-ref-relationsh
...
-export const INPUT_CONTAINS = 'contains';
+export const INPUT_CONTAINS = 'containsObservable';
 export const INPUT_RESOLVES_TO = 'resolvesTo';
 export const INPUT_BELONGS_TO = 'belongsTo';
-export const INPUT_FROM = 'from';
-export const INPUT_SENDER = 'sender';
-export const INPUT_TO = 'to';
-export const INPUT_CC = 'cc';
-export const INPUT_BCC = 'bcc';
+export const INPUT_SENDER = 'emailSender';
```
richard-julien, line 60: I thought we talk about pushing the objectXXX convention

opencti-platform/opencti-graphql/src/database/middleware.js
labo-flg: Most of the time the constant INPUT_NAME was not used, its' the string "from". 
It was an _input_ for email, but as for relationship it's not in any schema def, it's not an input.
We think it's an old confusion (email 'from' being used in place where we actually meant "from" of a relationship).

We removed the constant INPUT_NAME. It's not used in any schema def.

opencti-platform/opencti-graphql/src/schema/stixRefRelationship.ts
labo-flg: the convention objectXXX produces the change `contains` -> `objectContains`
but `objectContains` was an old filter key [still renamed on the fly](https://docs.opencti.io/latest/reference/filters-migration/) to `objects`, so we can't use objectContains

one thing leading to another, we thought it would be better to prefix with the entity they're used for as per the schema description. It's the case for a lot of other keys.

opencti-platform/opencti-graphql/src/database/middleware.js
SouadHadjiat: This part was causing the test "should all elements correctly deleted" in middleware-test to fail, because the "from" was not resolved for one of the entity (it was a stix ref relationship). and yes we think "INPUT_FROM" was used by mistake because it was "from" until we changed it to "emailFrom", since it was an input specific to email message. And it was the only place where it was used as a constant, in all other places in middleware we use the string 'from' directly.

lndrtrbn (approved):

Kedae (dismissed):
Need Julien eye on that

SouadHadjiat:
Related python client PR open for malwareAnalysis : https://github.com/OpenCTI-Platform/client-python/pull/563

PR #6086, opened 2024-02-22
[frontend/backend] add filters based on schema definition in the Add entities panels

Archidoit:
Add filters in some 'Add entities' panels :
- container > Entities > add entities
- container > observables > add entities
- investigation > add entities

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/common/containers/ContainerAddStixCoreObjects.jsx
+++ b/opencti-platform/opencti-front/src/private/components/common/containers/ContainerAddStixCoreObjects.jsx
@@ -13,11 +13,12 @@ import { useFormatter } from '../../../../components/i18n';
 import ContainerAddStixCoreObjectsLines, { containerAddStixCoreObjectsLinesQuery } from './ContainerAddStixCoreObjectsLines';
 import StixDomainObjectCreation from '../stix_domain_objects/StixDomainObjectCreation';
 import StixCyberObservableCreation from '../../observations/stix_cyber_observables/StixCyberObservableCreation';
-import { UserContext } from '../../../../utils/hooks/useAuth';
+import useAuth from '../../../../utils/hooks/useAuth';
 import ListLines from '../../../../components/list_lines/ListLines';
-import { constructHandleAddFilter, constructHandleRemoveFilter, emptyFilterGroup, filtersAfterSwitchLocalMode } from '../../../../utils/filters/filtersUtils';
+import { emptyFilterGroup } from '../../../../utils/filters/filtersUtils';
 import Drawer from '../drawer/Drawer';
 import useAttributes from '../../../../utils/hooks/useAttributes';
+import { usePaginationLocalStorage } from '../../../../utils/hooks/useLocalStorage';
 
 const useStyles = makeStyles((theme) => ({
   createButton: {
@@ -76,10 +77,11 @@ const ContainerAddStixCoreObjects = (props) => {
   const [openSpeedDial, setOpenSpeedDial] = useState(false);
   const [openCreateEntity, setOpenCreateEntity] = useState(false);
   const [openCreateObservable, setOpenCreateObservable] = useState(false);
-  const [sortBy, setSortBy] = useState('_score');
-  const [orderAsc, setOrderAsc] = useState(false);
 
   const { stixDomainObjectTypes, stixCyberObservableTypes } = useAttributes();
+  const {
+    platformModuleHelpers: { isRuntimeFieldEnable },
+  } = useAuth();
 
   const targetEntityTypesFilterGroup = {
     mode: 'and',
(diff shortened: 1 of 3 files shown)


opencti-platform/opencti-front/src/private/components/common/containers/ContainerAddStixCoreObjects.jsx
```diff
@@ -94,18 +96,71 @@ const ContainerAddStixCoreObjects = (props) => {
     ],
   };
 
-  const [filters, setFilters] = useState(
-    targetStixCoreObjectTypes
-    && !(targetStixCoreObjectTypes.includes('Stix-Domain-Object') || targetStixCoreObjectTypes.includes('Stix-Cyber-Observable'))
-      ? targetEntityTypesFilterGroup
-      : emptyFilterGroup,
+  const isTypeDomainObject = (types) => {
+    return !types
```
lndrtrbn, line 100: So it's `true` if we don't give any type? That feels strange

opencti-platform/opencti-front/src/private/components/common/containers/ContainerAddStixCoreObjects.jsx
```diff
@@ -94,18 +96,71 @@ const ContainerAddStixCoreObjects = (props) => {
...
+      || types.some((r) => stixDomainObjectTypes.indexOf(r) >= 0)
+      || (types.length === 1 && types[0] === 'Stix-Domain-Object');
+  };
+  const isTypeObservable = (types) => {
+    return !types
```
lndrtrbn, line 105: Same, `true` if no types given?

opencti-platform/opencti-front/src/private/components/common/containers/ContainerAddStixCoreObjects.jsx
```diff
@@ -94,18 +96,71 @@ const ContainerAddStixCoreObjects = (props) => {
...
+      || (isTypeObservable(targetStixCoreObjectTypes)
+        && isTypeDomainObject(targetStixCoreObjectTypes))
+    ) {
+      return 'Stix-Core-Object';
+    }
+    return null;
+  };
+
+  const LOCAL_STORAGE_KEY = `container-${containerId}-add-objects`;
+  const { viewStorage, helpers, paginationOptions: addObjectsPaginationOptions } = usePaginationLocalStorage(
```
lndrtrbn, line 136: What is the goal of the hook? The filters should be saved in local storage?

opencti-platform/opencti-front/src/private/components/workspaces/investigations/InvestigationAddStixCoreObjects.jsx
```diff
@@ -194,52 +173,48 @@ const InvestigationAddStixCoreObjects = (props) => {
...
+          sortBy={sortBy}
+          orderAsc={orderAsc}
+          dataColumns={buildColumns()}
+          handleSearch={helpers.handleSearch}
+          keyword={keyword}
+          handleSort={helpers.handleSort}
+          handleAddFilter={helpers.handleAddFilter}
+          handleRemoveFilter={helpers.handleRemoveFilter}
+          handleSwitchLocalMode={helpers.handleSwitchLocalMode}
+          handleSwitchGlobalMode={helpers.handleSwitchGlobalMode}
```
jpkha, line 181: I am not sure about the impact as we had many children components that passed these props but we should try to remove it later.
With `helpers ` we can use all the functions that is related to this..

opencti-platform/opencti-front/src/private/components/common/containers/ContainerAddStixCoreObjects.jsx
jpkha: Yes. But not only the filters but there are some other utilities.

opencti-platform/opencti-front/src/private/components/workspaces/investigations/InvestigationAddStixCoreObjects.jsx
Archidoit: yes it would be nice to refacto this later

opencti-platform/opencti-front/src/private/components/common/containers/ContainerAddStixCoreObjects.jsx
Archidoit: by default, if no types are specified, types=Stix-core-object

opencti-platform/opencti-front/src/private/components/common/containers/ContainerAddStixCoreObjects.jsx
Archidoit: same

opencti-platform/opencti-front/src/private/components/workspaces/investigations/InvestigationAddStixCoreObjects.jsx
Archidoit: (we pass those children + 'helpers' everywhere when listLines is called for the moment)

Archidoit: this refacto was not done for the moment because there are about 100 files to modify and the release is soon

opencti-platform/opencti-front/src/private/components/common/containers/ContainerAddStixCoreObjects.jsx
lndrtrbn: But we are not saving it in local storage here? Or at least we are not fetching what is in local storage when loading the page as if I refresh I don't see the previous filters I made

jpkha: Yes, you are correct. I'm unsure about the intention here. The saving in localStorage is functional, but upon closing the drawer, we reset the state and clear all filters.
https://github.com/OpenCTI-Platform/opencti/pull/6086/files/eb6c9427dcab8747dfda8053e566206e15ad277a#diff-c501c533b095ec5c7f5b484df27e33d102e40b4369f12d850a58a82b465ea31bR419-R421

I believe we're utilizing the useLocalStoragePagination for the infinite loader rather than for storing the filter in localStorage.

jpkha: In this history, we are using here a simple filter state. If we can confirm the behavior, I can change the code and use "useFilterState" instead.

jpkha: See with the product. We will keep the state. I remove the resetState function

lndrtrbn (approved):
Ok for me, need to resolve conflicts

PR #6088, opened 2024-02-22
[frontend] Improve playwright process

jpkha:
Improve playwright process


opencti-platform/opencti-front/builder/dev/dev-e2e.js
```diff
@@ -0,0 +1,91 @@
...
+const esbuild = require("esbuild");
+const compression = require("compression");
+const { RelayPlugin } = require("../plugin/esbuild-relay");
+const fsExtra = require("fs-extra");
+
+const basePath = "";
+const clients = [];
+const buildPath = "./builder/dev/build/";
+
+const middleware = (target, ws = false) => createProxyMiddleware(basePath + target, {
```
richard-julien: No need for a new specific file.
I think you can directly use the dev build.
Chokidar will not be a problem you can keep it.
And the BACK_END_URL env can be added to the dev config file without any impact.

opencti-platform/opencti-front/package.json
```diff
@@ -146,6 +146,7 @@
     "relay": "relay-compiler relay.config.json",
     "dev": "yarn relay && vite",
     "start": "yarn relay && node builder/dev/dev.js",
+    "start:e2e": "yarn relay && node builder/dev/dev-e2e.js",
```
richard-julien: Can be removed in favor of yarn start (due to previous command)

opencti-platform/opencti-front/tests_e2e/model/login.pageModel.ts
```diff
@@ -27,6 +27,9 @@ export class LoginPage {
     
     async login() {
         await this.page.goto('/');
+        await this.page.waitForTimeout(5000);
+        const buffer = await this.page.screenshot();
+        console.log(buffer.toString('base64'));
```
richard-julien: Must be removed

opencti-platform/opencti-front/tests_e2e/model/login.pageModel.ts
```diff
@@ -27,6 +27,9 @@ export class LoginPage {
     
     async login() {
         await this.page.goto('/');
+        await this.page.waitForTimeout(5000);
```
richard-julien: Can you explain that?

richard-julien (changes requested):

opencti-platform/opencti-front/tests_e2e/model/login.pageModel.ts
jpkha: it's for debugging. I got some issue on the login page. I am waiting 5sec to be sure to load the page then try a screenshot. Will be removed after.

opencti-platform/opencti-front/builder/dev/dev-e2e.js
jpkha: I tried using dev:build, but it took a very long time to start and all the tests timed out, I will give a try again. Therefore, I am trying a new esbuild configuration, similar to the production setup, to expedite the start and minimize issues during e2e test execution

jpkha: In the dev esbuild setup, we force the browser to not cache any content for the live reload by adding no-cache to the headers. 
It might be better to remove this part so the worker can avoid loading the same resources each time for the e2e. 
What do you think?

richard-julien (approved):

Issue #5454, opened 2024-01-11
Bad form validation in Create Observable form


labo-flg:
## Description

When you try to create a new observable, the form is not validated. There is no red marking or error to tell you which field is missing.

## Environment

OCTI 5.12.15 (seen on Testing and Demo platforms)

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. go to observables
2. click on create button
3. select File
4. scroll down and directly click CREATE without filling anything

## Expected Output

The form is validated, showing the fields that are required in red with a meaningful error message below the fields.

## Actual Output

![image](https://github.com/OpenCTI-Platform/opencti/assets/146674147/12e79a66-7fe2-4685-800c-cf6dedb0226b)

No red marking or error message.

Jipegien:
@labo-flg all observables or only File one?

labo-flg:
All observables I've tested are problematic.
They might show a different error banner than the one in my screenshot (probably the low-level graphql error), but never highlight in red the faulty fields.

richard-julien:
Not critical for me as the control is done even if the failure message is not clear sometimes.
Will be handle later

jpkha:
@labo-flg @CelineSebe we can close this issue ?

labo-flg:
The issue as a whole was not addressed, only errors related to "associated file" field that were not properly passed down to the component and displayed as form validation.

The form is still not properly validated, the user has to submit a faulty form to get an error from the backend (error might be unclear) and iteratively fix his form values. Not the best UX.

fixing this would require adapting the yup validator dynamically to the type of object being created, not a quick patch.

Jipegien:
@jborozco we need to define this as a validation system feature for every form. Can you create a Notion page for 6.3 and take the lead on this (business and tech refinement) ?

nino-filigran:
FYI @jborozco: 

**Exact same errors for ("missing required elements for XXXX creation (-)" ):** 
- Artifact
- Files
- X509

**Similar error for the following observables: (error is slightly different but not explicit either "observable is not correctly formatted")**
- Domain 
- Email adress 
- Hostname 
- IPV4
- IPV6
- Mac adress
- Software

**Different error, but still a bug for: there is no validation form for the following observables (can be created without entering any value in any field).** 
- Windows registry key value
- Email - MIME part

Jipegien:
Work have been done on this subject by a partner here : https://github.com/OpenCTI-Platform/opencti/pull/5877. 
[...]

PR #6090, opened 2024-02-23
[frontend] Fix error message when attempting to create an artifact without file (#6052)

### LINKED ISSUE #6052: No error message when attempting to create an artifact without file
## Description

In theory, when a object is attempted to be created, there is a validation on the mandatory fields. This validation occurs, but no error message is displayed. It would be great to have an error message indicating to the user that a file must be uploaded for sake of clarity (like other entities).

## Environment

4. Other environment details:

## Reproducible Steps

1. go to artifact
2. click on the + to create
3. click on create
4. nothing happens

## Expected Output

See an error message because I did not upload a file

## Actual Output

Nothing happen

CelineSebe:
### Proposed changes

* [Frontend] Add a useEffect to set the errorText in the CustomFileUploader component

### Related issues

* #6052 

### Checklist

### Further comments

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/common/files/CustomFileUploader.tsx
+++ b/opencti-platform/opencti-front/src/private/components/common/files/CustomFileUploader.tsx
@@ -1,4 +1,4 @@
-import React, { FormEvent, FunctionComponent, useState } from 'react';
+import React, { FormEvent, FunctionComponent, useEffect, useState } from 'react';
 import Box from '@mui/material/Box';
 import Button from '@mui/material/Button';
 import { Theme } from '@mui/material/styles/createTheme';
@@ -17,6 +17,7 @@ interface CustomFileUploadProps {
   ) => Promise<unknown>;
   isEmbeddedInExternalReferenceCreation?: boolean;
   label?: string;
+  formikErrors?: string;
   acceptMimeTypes?: string; // html input "accept" with MIME types only
   sizeLimit?: number; // in bytes
 }
@@ -60,6 +61,7 @@ const CustomFileUploader: FunctionComponent<CustomFileUploadProps> = ({
   label,
   acceptMimeTypes,
   sizeLimit = 0, // defaults to 0 = no limit
+  formikErrors,
 }) => {
   const { t_i18n } = useFormatter();
   const classes = useStyles();
@@ -72,6 +74,13 @@ const CustomFileUploader: FunctionComponent<CustomFileUploadProps> = ({
     const file = inputElement.files?.[0];
     const fileSize = file?.size || 0;
 
+    useEffect(() => {
+      if (formikErrors?.file) {
+        setErrorText(formikErrors?.file);
+      } else {
+        setErrorText('');
+      }
+    }, [formikErrors]);
     const newFileName = eventTargetValue.substring(
       eventTargetValue.lastIndexOf('\\') + 1,
     );
--- a/opencti-platform/opencti-front/src/private/components/common/files/CustomFileUploader.tsx
+++ b/opencti-platform/opencti-front/src/private/components/common/files/CustomFileUploader.tsx
@@ -17,7 +17,7 @@ interface CustomFileUploadProps {
   ) => Promise<unknown>;
   isEmbeddedInExternalReferenceCreation?: boolean;
   label?: string;
-  formikErrors: {
+  formikErrors?: {
     file?: string,
   }
   acceptMimeTypes?: string; // html input "accept" with MIME types only
--- a/opencti-platform/opencti-front/src/private/components/common/files/CustomFileUploader.tsx
+++ b/opencti-platform/opencti-front/src/private/components/common/files/CustomFileUploader.tsx
@@ -17,7 +17,9 @@ interface CustomFileUploadProps {
   ) => Promise<unknown>;
   isEmbeddedInExternalReferenceCreation?: boolean;
   label?: string;
-  formikErrors?: string;
+  formikErrors: {
+    file?: string,
+  }
   acceptMimeTypes?: string; // html input "accept" with MIME types only
   sizeLimit?: number; // in bytes
 }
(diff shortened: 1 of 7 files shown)


opencti-platform/opencti-graphql/src/domain/stixCyberObservable.js
```diff
@@ -389,6 +389,9 @@ export const artifactImport = async (context, user, args) => {
   const artifact = await addStixCyberObservable(context, user, artifactData);
   const meta = { version };
   await upload(context, user, `import/${artifact.entity_type}/${artifact.id}`, resolvedFile, { entity: artifact, meta });
+  if (!upload) {
+    throw FunctionalError('Something went wrong with the file upload');
+  }
```
labo-flg: the upload variable seems undefined

```suggestion
  const upload = await upload(context, user, `import/${artifact.entity_type}/${artifact.id}`, resolvedFile, { entity: artifact, meta });
  if (!upload) {
    throw FunctionalError('Something went wrong with the file upload');
  }
```

labo-flg:
Seen in pair review:

it's more like a frontend display issue. The yup validation is working fine as it is in `ArtifactCreation`. It's just that the error message is not displayed.

We need to pass down the formik errors to the `CustomFileUpload` so that it's internal error state is updated on formik error received, and not only on the custom error it checks on his own.

This way user gets a good feedback in the form.

As for the observable creation form, it's a lot more work as this is used for all types of observables, without yup validation... and backend errors are not always understandable. See #5454

SarahBocognano (approved):
Tested ✅

CelineSebe:
I just added a createArtifact test. I would appreciate a review of this part if possible.

opencti-platform/opencti-front/src/private/components/common/files/CustomFileUploader.tsx
```diff
@@ -60,12 +63,21 @@ const CustomFileUploader: FunctionComponent<CustomFileUploadProps> = ({
...
   sizeLimit = 0, // defaults to 0 = no limit
+  formikErrors,
 }) => {
   const { t_i18n } = useFormatter();
   const classes = useStyles();
   const [fileNameForDisplay, setFileNameForDisplay] = useState('');
   const [errorText, setErrorText] = useState('');
 
+  useEffect(() => {
+    if (formikErrors?.file) {
```
lndrtrbn, line 74: By passing the entire `formikErrors` object and hardcoding the property `.file` here we force to use this component with the property `file` instead of leaving the liberty to the caller to chose by itself how the property should be named. But I see it is already made this way for `setFieldValue` below so... Ok, don't bother change the entire component.

opencti-platform/opencti-front/tests_e2e/artifact/createArtifact.spec.ts
```diff
@@ -0,0 +1,15 @@
...
+import { expect, test } from "../fixtures/baseFixtures";
+
+
+test('Validate a new artifact file', async ({ page }) => {
+        const artifactPage = new ArtifactPage(page);
+        const artifactImport = new ArtifactImportPage(page);
+        await page.goto('/dashboard/observations/artifacts')
+        await artifactPage.addNewArtifactImport().click();
+        artifactImport.getFileInput();
+        // await artifactImport.fillFileInput('File');
```
lndrtrbn: To remove? 
And it seems you have indentation issues. Is the linter applied on test files here?

opencti-platform/opencti-front/tests_e2e/model/ArtifactImport.pageModel.tsx
lndrtrbn, line 1: Why have an other page model for import button instead of having all in the Artifact page model?

opencti-platform/opencti-front/tests_e2e/artifact/createArtifact.spec.ts
```diff
@@ -0,0 +1,15 @@
+import { ArtifactPage} from "../model/Artifact.pageModel";
+import { ArtifactImportPage} from "../model/ArtifactImport.pageModel";
+import { expect, test } from "../fixtures/baseFixtures";
+
+
+test('Validate a new artifact file', async ({ page }) => {
```
lndrtrbn: Be more explicit on what the test is testing: should display an error if trying to create an artifact without a file

opencti-platform/opencti-front/tests_e2e/model/ArtifactImport.pageModel.tsx
lndrtrbn: In my opinion you can regroup those two, especially as they are both very small

opencti-platform/opencti-front/tests_e2e/artifact/createArtifact.spec.ts
CelineSebe: Indeed, thanks :)
No issue with the linter

opencti-platform/opencti-front/tests_e2e/model/ArtifactImport.pageModel.tsx
CelineSebe: It allows placing the data-test-id on the Artifact page root. It seems to be a good practice to place this element fairly high in the file hierarchy. I believe they serve as anchor points that can be reused for other tests.

lndrtrbn: But regrouping does not mean you can't do that

opencti-platform/opencti-front/tests_e2e/artifact/createArtifact.spec.ts
lndrtrbn: After testing, there is an issue as the linter seems not linting e2e spec files but not related to your PR.

opencti-platform/opencti-front/tests_e2e/model/ArtifactImport.pageModel.tsx
CelineSebe: If I understand correctly, you want me to group the functions in the same pageModel?
If that's the case, I prefer to separate the actions by screen... I find that we understand better what's happening

opencti-platform/opencti-front/tests_e2e/artifact/createArtifact.spec.ts
CelineSebe: I see the indentation issue now, it's fixed! Thank you

opencti-platform/opencti-front/tests_e2e/model/ArtifactImport.pageModel.tsx
lndrtrbn: You still are on the same screen, the content inside `ArtifactImport.pageModel` is just the form of the page `Artifacts` to create new ones.  But anyway, as you prefer 🙂

lndrtrbn (approved):

PR #6100, opened 2024-02-23
[backend] Fix upsert indicator score since decay (#2859)

SouadHadjiat:
### Proposed changes

* Don't change x_opencti_score (and valid_from / valid_until) during upsert when indicator score is the same as the base score (score at creation)
* Add Decay manager user
* Fix display of current stable score in dialog in case the same score is present multiple times in history (happens when the base score is updated)
* Change messages & translations (seen with JP)

### Related issues

* #2859 

### Checklist


opencti-platform/opencti-graphql/src/database/middleware.js
```diff
@@ -2380,6 +2380,10 @@ const upsertElement = async (context, user, element, type, basePatch, opts = {})
   if (type === ENTITY_TYPE_INDICATOR) {
     if (updatePatch.decay_applied_rule && updatePatch.decay_base_score === element.decay_base_score) {
       logApp.debug('UPSERT INDICATOR -- no decay reset because no score change', { element, basePatch });
+      updatePatch.x_opencti_score = element.x_opencti_score; // don't change the score
+      // don't reset valid_from & valid_until
+      updatePatch.valid_from = element.valid_from;
```
SouadHadjiat, line 2388: I'm not really sure about this part, I wonder if we might have a case where valid_from / valid_until change with the same score

opencti-platform/opencti-front/src/private/components/observations/indicators/DecayDialog.tsx
```diff
@@ -87,13 +87,31 @@ const DecayDialogContent : FunctionComponent<DecayDialogContentProps> = ({ indic
...
+      style: { color: theme.palette.text.primary },
+      score: history.score,
+      updated_at: getDateAsTextFor(history),
+    };
+  };
+
   const labelledHistoryList: LabelledDecayHistory[] = [];
-  decayHistory.forEach((history) => (
-    labelledHistoryList.push(getDisplayFor(history))
+  const currentScoreIndex = decayHistory.findLastIndex((history) => history.score === indicator.x_opencti_score);
```
SouadHadjiat, line 101: fix display of current score when we have multiple times in history the same score (it was displayed "Current stable score" each time)

PR #6101, opened 2024-02-23
[backend] Organization sharing behavior change for upsert and enrichment

### LINKED ISSUE #6042: [backend] Organization sharing behavior change for upsert and enrichment
## Use case

The platform is currently handling refs differently in term of upsert.
For refs like markings, labels etc its an accumulation to prevent any data loss / data leak
For sharing (granted_refs) is sync/replace each time.
This situation can leads to some inconsistency. So we need to **Change the granted_refs behavior to act like marking defs when upserted**

Because of changing this behavior we also need to handle the impacts in enrichment process
* User ask enrichment needs to be done as the connector and not the user
* Granted_refs needs to be handle correctly by the enrichment connectors to follow this information and so create new enriched data with the correct sharing
* Playbook operation for remove/replace sharing must be removed as no longer supported.

Data cleaning/fix for this information will need to be done through massive operation in the UI.

This ticket cancel and replace.
https://github.com/OpenCTI-Platform/opencti/issues/5856
https://github.com/OpenCTI-Platform/opencti/issues/5857
https://github.com/OpenCTI-Platform/opencti/issues/5842

richard-julien:
Organization sharing behavior change for upsert and enrichment

Linked to PRs
https://github.com/OpenCTI-Platform/client-python/pull/568
https://github.com/OpenCTI-Platform/connectors/pull/1859

### DIFF
--- a/opencti-platform/opencti-graphql/src/domain/stixCoreObject.js
+++ b/opencti-platform/opencti-graphql/src/domain/stixCoreObject.js
@@ -190,7 +190,7 @@ export const askElementEnrichmentForConnector = async (context, user, enrichedId
   const message = {
     internal: {
       work_id: work.id, // Related action for history
-      applicant_id: user.id, // User asking for the import
+      applicant_id: null, // No specific user asking for the import
     },
     event: {
       entity_id: element.standard_id,
--- a/opencti-platform/opencti-graphql/src/domain/stixCoreObject.js
+++ b/opencti-platform/opencti-graphql/src/domain/stixCoreObject.js
@@ -12,9 +12,10 @@ import {
   ABSTRACT_STIX_CORE_OBJECT,
   ABSTRACT_STIX_DOMAIN_OBJECT,
   buildRefRelationKey,
+  CONNECTOR_INTERNAL_ENRICHMENT,
   ENTITY_TYPE_CONTAINER,
   INPUT_EXTERNAL_REFS,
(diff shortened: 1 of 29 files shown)


Archidoit:
Connecteurs testés : 
- Greynoise
- Abuseipdb

opencti-platform/opencti-graphql/src/domain/stixCoreObject.js
```diff
@@ -190,10 +191,12 @@ export const askElementEnrichmentForConnector = async (context, user, enrichedId
   const message = {
     internal: {
       work_id: work.id, // Related action for history
-      applicant_id: user.id, // User asking for the import
+      applicant_id: null, // No specific user asking for the import
```
JeremyCloarec, line 194: Shouldn't we set the applicant_id to the connector user id if we want the enrichment update to have the connector as the creator?

JeremyCloarec:
Tested OK when it comes to keeping the organization sharing during enrichment, and applying the organization sharing to objects created during enrichment.
I've encountered two problems during testing however:
- First one is related to the "creator" of the enrichment: in the entity history and list of creators, it is the worker user that is used as a creator and not the connector user
- When trying to export entities with the stix export connector, I either get an error during export (when trying to export an observable), or the granted_ref is not kept on export (I tried on a malware entity)

JeremyCloarec (approved):

PR #6114, opened 2024-02-26
[backend/frontend] add more details to effective confidence level source

labo-flg:
The effective confidence level is computed by the backend according top the user's level, their groups levels and their capabilities.

The model did not allow to know that the confidence source was the bypass capability, we only had user or group.

This PR improves the API about effective confidence level and allow the frontend to display precise information on why the user has such effective value, including the bypass capa case.


lndrtrbn (approved):

lndrtrbn (changes requested):
Missing generated file

lndrtrbn (approved):

PR #6124, opened 2024-02-26
[frontend] fix creator filter in user history redirection

Archidoit:
issue: get all relationships (or entities) created by a user, from the user view, seems to break the filters

Steps to reproduce: 
- go to a user overview
- in the history, click on the 'view all relationships created by the user' button
image
- we are redirected to the list of relationships with a filter on the creator
Bug: this filter can't be edited

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/settings/users/UserHistory.tsx
+++ b/opencti-platform/opencti-front/src/private/components/settings/users/UserHistory.tsx
@@ -12,6 +12,7 @@ import { useFormatter } from '../../../../components/i18n';
 import Loader, { LoaderVariant } from '../../../../components/Loader';
 import SearchInput from '../../../../components/SearchInput';
 import UserHistoryLines, { userHistoryLinesQuery } from './UserHistoryLines';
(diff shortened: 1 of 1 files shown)


labo-flg (approved):

opencti-platform/opencti-front/src/private/components/settings/users/UserHistory.tsx
```diff
@@ -65,6 +66,7 @@ const UserHistory: FunctionComponent<UserHistoryProps> = ({
         ],
         operator: 'eq',
         mode: 'or',
+        id: generateUUID(), // because filters in the URL
```
SouadHadjiat: why do we need an `id` in filters ? and it's using a method from `tests_e2e`, I don't think its the right import.

SouadHadjiat: in filtersUtils.tsx, I see that you use : 
`id: uuid(),`
I guess if we have now a logic with id in filters, we should use this same method `uuid()`, no?

jpkha: I change by uuid().

PR #6151, opened 2024-02-28
[backend] Improve csv feed

richard-julien:


opencti-platform/opencti-front/src/private/components/data/ingestionCsv/IngestionCsvLine.tsx
```diff
@@ -57,6 +57,7 @@ const ingestionCsvLineFragment = graphql`
     uri
     ingestion_running
     current_state_date
```
Kedae, line 59: I think this date is never valued. Am I missing something ?

SarahBocognano: It's working if you put a date in the "import from date" at the creation, but I just noticed that the date update doesn't work

richard-julien: No, need to be deleted in the future

PR #6155, opened 2024-02-28
[frontend] Fix fetch of containers in workspace

Kedae:
Investigation -> Add in container -> No item


SouadHadjiat (approved):
Tested locally, fixes the issue ✅
I don't really understand however why filterKey is 'id'

JeremyCloarec (approved):

PR #6160, opened 2024-02-28
[frontend/backend] Add PAP at init + fix leftbar

### LINKED ISSUE #6154: Adding PAP marking definition by default in fresh images
## Use case

Like TLP protocol, PAP protocol is widely used across security team.

We need to provide it as default marking definition (like TLP) in every new platform.

Spec:
- Create a new Marking type named "PAP"
- Create a Marking definition of type PAP named "PAP:CLEAR", order 1, same color as TLP:CLEAR
- Create a Marking definition of type PAP named "PAP:GREEN", order 2, same color as TLP:GREEN
- Create a Marking definition of type PAP named "PAP:AMBER", order 3, same color as TLP:AMBER
- Create a Marking definition of type PAP named "PAP:RED", order 4, same color as TLP:RED

Paper here : https://www.cert.ssi.gouv.fr/csirt/politique-partage/

PAP markings created in testing platform

Kedae:


opencti-platform/opencti-front/src/private/components/nav/LeftBar.jsx
```diff
@@ -860,7 +860,7 @@ const LeftBar = () => {
           </MenuList>
         </Security>
       </div>
-      <div style={height + (navOpen && selectedMenu ? 300 : 130) < dimension.height ? { position: 'fixed', bottom: 10, left: 0 } : {}}>
+      <div style={{ position: 'fixed', bottom: 10, left: 0 }}>
```
jpkha: I don't know what you are trying to fix but if the height is small we got  some element which is comming under :
image

Kedae: We support minimal requirement of height and width, so small screens breaking is not an issue.

I've fixed that sometimes if you open a menu, then open settings, the "Filigran" block gets up.

Kedae: image
On testing

frapuks (commented):
It works on initialisation
Shouldn't we have a migration?
According to the information I've read on the ANSII site, some TLPs and PAPs cannot be associated (TLP:CLEAR + PAP:RED). Will this be the subject of a new feature?

Kedae:
> It works on initialisation Shouldn't we have a migration? According to the information I've read on the ANSII site, some TLPs and PAPs cannot be associated (TLP:CLEAR + PAP:RED). Will this be the subject of a new feature?

No and no.
We just wanted to add some "classic" marking in fresh platform, that's it, we don't want to go any further on that matter.

frapuks (approved):

opencti-platform/opencti-front/src/private/components/nav/LeftBar.jsx
SamuelHassine: @Kedae this is a regression, I spent a lot of time to make this not overlapping. If you do this, on 90% of screens when menu is open Filigran logo will overlap. @nino-filigran and I have validated this together.

Issue #6074, opened 2024-02-22
The way to handle external ref is inconsistent 


nino-filigran:
## Description
We had a [bug](https://github.com/OpenCTI-Platform/opencti/issues/5884) to fix the behavior of external references in reports.
As mentioned in [this ](https://github.com/OpenCTI-Platform/opencti/issues/5884#issuecomment-1957280799) comment of the same ticket, the behavior of enforcing reference is bigger. 

## Environment

Demo

## Reproducible Steps

Edit an entity while having external reference 
1. Have external reference enforced for your entity 
2. Click on an entity (ex report) that have an external reference
3. Click on edit 
4. Click on the field external ref
5. You only see the external ref of the report, not the full list of external refs of your platform and cannot select existing ones already created that do not pertains to your entity.

Expected output: 
You should see the list of all the external refs to be able to add another one.

Sharing entity to an org
1. Have external reference enforced for your entity 
2. Click on an entity (ex report) that have an external reference enforced
3. Click on share to org
4. select an org
5. You got an error message because you have not added any external ref 

Expected output: 
It would help if you were prompted the pop up to add an external ref instead of having an error message. 

Remove all external references while enforce external reference is enabled
1. Have external reference enforced for your entity 
2. Click on an entity (ex report) that have an external reference enforced and have some external ref
3. Remove all external ref

Expected output: 
You should not be able to remove all external ref. At least always 1 external ref must remains in the entity.

Adding relations (core ) is sometimes possible without having to add an external ref
1. Have external reference enforced for your entity 
2. Go to your entity and click on add relation 

Expected output: 
You should not be able to add a relation without adding an external ref. The pop up must be pormpted if not in the form.

Kedae:
/!\ All this bug fix MUST be covered by e2e testing

PR #6175, opened 2024-02-29
[backend/frontend] Fix enable references behavior when adding entities to containers

### LINKED ISSUE #5884: Mechanisms inconsistency when "enforce reference" is activated
## Description

Starting point : "enforce reference" is activated on Report.

If I want to add an object to my report (from the Knowledge view, or entities, or observables), I get a "Validation error" because I haven't validated my change with a reference. I can't do this because the pop-up doesn't appear. To understand what pop-up I'm talking about, try modifying the report description and this time the pop-up will appear.

### Additional information

- If I try this time to create an entity or relationship to add directly to my report from knowledge graph, I also get this error but note that the object is created (it's just not added to my report because of the lack of reference).
- If I validate a workbench from the data tab of a report, its execution works (my entities are well created on my platform) but I still get the "Validation error" and my objects are not added to my report, still for the same problem.

## Environment

OCTI 5.12.29

## Reproducible Steps

1. Activate "enforce reference" on Report (settings > customization > report)
2. Navigate on the "Entities" tab of a Report
3. Try to add an entity

EDIT: **WARNING**

If your have the "Bypass all capabilities" or "Bypass mandatory references if any" right, you bypass the "enforce reference" policy and therefore don't reproduce the bug (because it's as if the policy wasn't activated).

## Expected Output

Obtain the pop up allowing me to add the reference needed

JeremyCloarec:
This PR is only to fix enable references behavior in the scope of the #5884 issue. Other PRs will aim to fully fix the enable references behavior for all entities, and the full list of problems can be seen in the #6074 

### Checklist


opencti-platform/opencti-graphql/config/schema/opencti.graphql
```diff
@@ -12126,8 +12126,8 @@ type ReportEditMutations {
   fieldPatch(input: [EditInput]!, commitMessage: String, references: [String]): Report @auth(for: [KNOWLEDGE_KNUPDATE])
   contextPatch(input: EditContext): Report @auth(for: [KNOWLEDGE_KNUPDATE])
   contextClean: Report @auth(for: [KNOWLEDGE_KNUPDATE])
-  relationAdd(input: StixRefRelationshipAddInput!): StixRefRelationship @auth(for: [KNOWLEDGE_KNUPDATE])
-  relationDelete(toId: StixRef!, relationship_type: String!): Report @auth(for: [KNOWLEDGE_KNUPDATE])
+  relationAdd(input: StixRefRelationshipAddInput!, commitMessage: String, references: [String]): StixRefRelationship @auth(for: [KNOWLEDGE_KNUPDATE])
```
SouadHadjiat, line 12137: I think all these API changes will need a client python update (here see `opencti_report.py` `add_stix_object_or_stix_relationship`)

JeremyCloarec: Good point thank you, I'll update the client as well

SouadHadjiat:
When trying to update a grouping with external reference enforced (as an admin), I don't see the list of external references in the dialog : 

image

I can see the query in the network inspector (which returns 0 results), maybe the filter is not right : 

image

JeremyCloarec:
> When trying to update a grouping with external reference enforced (as an admin), I don't see the list of external references in the dialog :

That's a bug that we also noticed, that is described here https://github.com/OpenCTI-Platform/opencti/issues/6074 with the description "Edit an entity while having external reference". It will be worked on in the next step.
Right now the external reference dialog only has the external references already added to the report, and you can not see other external references from the platform in it.

opencti-platform/opencti-front/tests_e2e/enableReferences/addReportObservable.spec.ts
```diff
@@ -0,0 +1,203 @@
...
+    await expect(loginPage.getPage()).toBeVisible();
+    await loginPage.fillLoginInput(noBypassUserLogin);
+    await loginPage.fillPasswordInput(noBypassUserPassword);
+    await loginPage.getSignInButton().click();
+    await expect(dashboardPage.getPage()).toBeVisible();
+    await page.context().storageState({ path: noBypassUserAuthFile });
+  });
+});
+
+test('Add and remove observable from Observables tab of a Report as Admin user', async ({ page }) => {
```
aHenryJard, line 108: question: is it intended that this "test" is not inside a test.describe block ?

opencti-platform/opencti-front/tests_e2e/enableReferences/addReportObservable.spec.ts
```diff
@@ -0,0 +1,203 @@
...
+    await rolesSettingsPage.getRoleInList(noBypassRoleName).click();
+    await rolePage.getEditButton().click();
+    await roleFormPage.getCapabilitiesTab().click();
+    await roleFormPage.getAccessKnowledgeCheckbox().click();
+    await expect(roleFormPage.getAccessKnowledgeCheckbox()).toBeChecked();
+    await roleFormPage.getCreateUpdateKnowledgeCheckbox().click();
+    await expect(roleFormPage.getCreateUpdateKnowledgeCheckbox()).toBeChecked();
+    await roleFormPage.getAccessAdministrationCheckbox().click();
+    await expect(roleFormPage.getAccessAdministrationCheckbox()).toBeChecked();
+  });
```
aHenryJard, line 49: Random thinking: I guess we could have some way to reuse that piece of test code that creates user/roles and groups I don't mean one groups for all test, but have a methode like createRole(roleName, listOfCapacities) ...etc

opencti-platform/opencti-graphql/src/resolvers/container.js
```diff
@@ -49,8 +49,9 @@ const containerResolvers = {
       fieldPatch: ({ input, commitMessage, references }) => stixDomainObjectEditField(context, context.user, id, input, { commitMessage, references }),
       contextPatch: ({ input }) => stixDomainObjectEditContext(context, context.user, id, input),
       contextClean: () => stixDomainObjectCleanContext(context, context.user, id),
-      relationAdd: ({ input }) => stixDomainObjectAddRelation(context, context.user, id, input),
-      relationDelete: ({ toId, relationship_type: relationshipType }) => stixDomainObjectDeleteRelation(context, context.user, id, toId, relationshipType),
+      relationAdd: ({ input, commitMessage, references }) => stixDomainObjectAddRelation(context, context.user, id, input, { commitMessage, references }),
+      // eslint-disable-next-line max-len
```
aHenryJard, line 53: can you split the line instead ? Or is it really not readable ?

aHenryJard (commented):
Only PR reading comment for now.

opencti-platform/opencti-graphql/config/schema/opencti.graphql
aHenryJard: reading python client code, I think it's better to add commitMessage and references data to StixRefRelationshipAddInput instead of adding new parameters here.

opencti-platform/opencti-front/tests_e2e/enableReferences/addReportObservable.spec.ts
JeremyCloarec: That's a good point, we already talked about that with Souad and Jpk actually. We are planning to discuss it more during the e2e workshop

opencti-platform/opencti-front/tests_e2e/enableReferences/addReportObservable.spec.ts
JeremyCloarec: Kind of! The describe blocks I used before were either to: group up tests that are dependant on each other (for the role/group/user creation), or to run the tests in the describe block with a specific context (the authentication describe block is used to run the authentication in a new browser context, to be able to log in the new user without having to log out the main user).
Since this test is independant from any others, I decided to not include it in a describe. 
But the structure of the test file and the usage of describe blocks in general could be up to debate for sure

opencti-platform/opencti-graphql/src/resolvers/container.js
JeremyCloarec: I didn't find a nice way to keep it readable and keep lint happy, so i went with this. Also, in other resolvers like stixCoreRelationship, the relationDelete line was also handled this way

opencti-platform/opencti-graphql/config/schema/opencti.graphql
SouadHadjiat: I think it's to be consistent with our existing APIs, which all declare (if I'm not mistaken) commitMessage and references as separate parameters. Ex: 
```
relationsAdd(input: StixRefRelationshipsAddInput!, commitMessage: String, references: [String]): StixCoreObject
relationDelete(toId: StixRef!, relationship_type: String!, commitMessage: String, references: [String]): StixCoreObject
```

SouadHadjiat (approved):

PR #6178, opened 2024-02-29
[frontend] Add Playwright teams webhook 

jpkha:
Add a team webhook for Playwright.
In case of failure, we receive some screenshots so that we can debug more easily.

image
image

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -90,6 +90,7 @@ steps:
     environment:
       BACK_END_URL: http://opencti-e2e-start:4500
       E2E_TEST: true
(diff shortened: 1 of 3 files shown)


labo-flg:
Let's move on with this PR, we can try and see how it looks and if it's useful

labo-flg (approved):

lndrtrbn:
@jpkha could you merge it if it's good for you?

jpkha:
I merged it, but we still don't have any way to hide the team's URL, so it won't work yet, but it also won't crash anything.

PR #6183, opened 2024-02-29
[backend/frontend] allow sorting groups by confidence level (#5756)

### LINKED ISSUE #5756: Implement sorting on max confidence level in groups
## Use case

Implement sorting on max confidence level in groups.

labo-flg:
### Proposed changes

* new enum value for `GroupsOrdering` 
* in the elastic query,handle this complex sorting simply with dot syntax `group_confidence_level.max_confidence: { order: 'asc' }`
* to generalize, add a new field in schema attribute definition for objects: `sortBy`; if defined, it tells how to build the sorting key

### Related issues

Closes #5756 

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/modules/attributes/internalObject-registrationAttributes.ts
+++ b/opencti-platform/opencti-graphql/src/modules/attributes/internalObject-registrationAttributes.ts
@@ -240,6 +240,7 @@ const internalObjectsAttributes: { [k: string]: Array<AttributeDefinition> } = {
       multiple: false,
       upsert: false,
       isFilterable: false,
+      sortBy: { path: 'group_confidence_level.max_confidence', type: 'numeric' },
       mappings: [
         { name: 'max_confidence', label: 'Max Confidence', type: 'numeric', precision: 'integer', editDefault: false, mandatoryType: 'internal', multiple: false, upsert: false, isFilterable: true },
         { name: 'overrides',
--- a/opencti-platform/opencti-graphql/src/schema/attribute-definition.ts
+++ b/opencti-platform/opencti-graphql/src/schema/attribute-definition.ts
@@ -38,11 +38,17 @@ type BasicDefinition = {
   update?: boolean // If attribute can be updated (null = true)
 };
 
+export type MappingDefinition = AttributeDefinition & {
+  associatedFilterKeys?: { key: string, label: string }[] // filter key and their label, to add if key is different from: 'parentAttributeName.nestedAttributeName'
+};
+
 type BasicObjectDefinition = BasicDefinition & {
-  mappings: (
-    { associatedFilterKeys?: { key: string, label: string }[] } // filter key and their label, to add if key is different from: 'parentAttributeName.nestedAttributeName'
-    & AttributeDefinition
-  )[],
+  mappings: MappingDefinition[],
+  // if the object attribute can be used for sorting, we need to know how
+  sortBy?: {
+    path: string // path leading to the value that serves for sorting
+    type: string // type of this value, copied for convenience from corresponding mapping (checked at registration)
+  }
 };
 export type DateAttribute = { type: 'date' } & BasicDefinition;
 export type BooleanAttribute = { type: 'boolean' } & BasicDefinition;
--- a/opencti-platform/opencti-graphql/src/schema/attribute-definition.ts
+++ b/opencti-platform/opencti-graphql/src/schema/attribute-definition.ts
@@ -154,7 +154,7 @@ export const files: AttributeDefinition = {
   mappings: [
     id,
     { name: 'name', label: 'Name', type: 'string', format: 'short', editDefault: false, mandatoryType: 'no', multiple: true, upsert: true, isFilterable: true },
-    { name: 'description', label: 'Name', type: 'string', format: 'text', editDefault: false, mandatoryType: 'no', multiple: true, upsert: true, isFilterable: true },
+    { name: 'description', label: 'Description', type: 'string', format: 'text', editDefault: false, mandatoryType: 'no', multiple: true, upsert: true, isFilterable: true },
     { name: 'version', label: 'Version', type: 'date', editDefault: false, mandatoryType: 'no', multiple: true, upsert: true, isFilterable: true },
     { name: 'mime_type', label: 'Mime type', type: 'string', format: 'short', editDefault: false, mandatoryType: 'no', multiple: true, upsert: true, isFilterable: true },
     { name: 'inCarousel', label: 'Include in carousel', type: 'boolean', mandatoryType: 'no', editDefault: false, multiple: false, upsert: true, isFilterable: true },
--- a/opencti-platform/opencti-graphql/src/schema/schema-attributes.ts
+++ b/opencti-platform/opencti-graphql/src/schema/schema-attributes.ts
@@ -1,7 +1,7 @@
 import * as R from 'ramda';
 import { RULE_PREFIX } from './general';
 import { FunctionalError, UnsupportedError } from '../config/errors';
-import type { AttributeDefinition, AttrType, ComplexAttributeWithMappings } from './attribute-definition';
+import type { AttributeDefinition, AttrType, ComplexAttributeWithMappings, MappingDefinition } from './attribute-definition';
 import { shortStringFormats } from './attribute-definition';
 import { getParentTypes } from './schemaUtils';
 
@@ -16,7 +16,45 @@ export const depsKeysRegister = {
   },
 };
 
+// -- Utilities to manipulate AttributeDefinitions --
+
+const isMandatoryAttributeDefinition = (schemaDef: AttributeDefinition) => schemaDef.mandatoryType === 'external' || schemaDef.mandatoryType === 'internal';
+
+const isNonFlatObjectAttributeDefinition = (schemaDef: AttributeDefinition) : schemaDef is ComplexAttributeWithMappings => { // handy typeguard
+  return schemaDef.type === 'object' && schemaDef.format !== 'flat';
+};
+
+/**
+ * Returns the attribute definition for a given dotted path inside the given AttributeDefinition,
+ * following the mappings recursively.
+ */
+const getAttributeMappingFromPath = (path: string, schemaDef: AttributeDefinition | MappingDefinition): MappingDefinition => {
+  const pathTokens = path.split('.');
+  if (pathTokens.length === 1) {
+    return schemaDef;
+  }
+  if (!isNonFlatObjectAttributeDefinition(schemaDef)) {
+    throw FunctionalError(`Cannot resolve path [${path}], [${schemaDef.name}] is not an object`);
+  }
+  const mapping = schemaDef.mappings.find((m) => m.name === pathTokens[1]);
+  if (!mapping) {
+    throw FunctionalError(`Schema definition named [${schemaDef.name}] is missing mapping for attribute [${pathTokens[1]}]`);
+  }
+
+  if (pathTokens.length > 2) {
+    // remove first and recursively check the rest of the path
+    pathTokens.shift();
+    return getAttributeMappingFromPath(pathTokens.join('.'), mapping);
+  }
+  return mapping;
+};
+
+// Flag to track if the schema was rea; when read for the first time the schema is then read-only and new registration is diallowed
 let usageProtection = false;
+
+/**
+ * Main utility object to write and read the schema in the platform
+ */
 export const schemaAttributesDefinition = {
   allAttributes: new Map<string, AttributeDefinition>(),
   attributes: {} as Record<string, Map<string, AttributeDefinition>>,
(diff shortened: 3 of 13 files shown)


opencti-platform/opencti-graphql/src/modules/attributes/internalObject-registrationAttributes.ts
```diff
@@ -240,6 +240,7 @@ const internalObjectsAttributes: { [k: string]: Array<AttributeDefinition> } = {
       multiple: false,
       upsert: false,
       isFilterable: false,
+      sortBy: 'max_confidence',
```
labo-flg: To enable sorting on complex object, I suggest we enhance the schema definition, to have this optional field `sortBy` that should point to one of the field in the mapping.

This implementation here allows only for a first-level field, but it might be improved with a dot-path for instance. In the elastic search, it ends up as a dot-path anyway.

opencti-platform/opencti-graphql/src/utils/sorting.ts
```diff
@@ -0,0 +1,35 @@
+import { isDateNumericOrBooleanAttribute, schemaAttributesDefinition } from '../schema/schema-attributes';
+import { UnsupportedError } from '../config/errors';
+
+export const buildElasticSortingForAttributeCriteria = (orderCriteria: string, orderMode: 'asc' | 'desc') => {
```
labo-flg, line 4: the algorithm here is a refactoring, it shall cover all previous cases properly and add the object case with sortBy.

opencti-platform/opencti-graphql/src/modules/attributes/internalObject-registrationAttributes.ts
labo-flg: cc @richard-julien WDYT?

richard-julien: Honestly i would prefer directly the support of user_confidence_level.max_confidence in sorting field.
Any problem or specific difficulty to allow use path notation directly in the sort field asked by the user?

labo-flg: If the approach is OK, I can go for the full path in the sortBy field, yes.

Note that one the client side, we manage sort-by using gql enums, and not a string variable. Thus we have fixed behaviors. It fits our mostly-flat model, but with complex object it's not a good fit.

A string variable with path notation compatible with all kind of attributes would be nice too (a bit like we have now with filters, but with sorting too). Not for this PR though 😇

labo-flg: To sum up our discussions: we'll go for a full path like we do with the patching API `object_path` param.

`sortBy` should be an array of objects like `{ key: 'confidence', path: 'group_confidence_level/max_confidence' }`
Having an array will allow for several possible sorting strategy for a single complex object.

This way, the API is future-proof (as far as we know today). The actual implementation will still be a first-level approach like we did first in this PR. When (if) we need to go further into that direction we'll know how to do it but such complexity is unnecessary for now.

labo-flg: We also could use some refactoring of the schema utilities: the isXXXAttribute functions are just checking the top-level attribute definition but the proper way is to accept an attribute def as parameter, so we can use these utility functions on inner mappings.

labo-flg:
@richard-julien I've updated this PR after our latest discussions. 

I did not use an array to define the possible sorting on an object, as it's impossible right now to do so without extensive modifications of the way sorting works API side + client side : we can only express a sorting key, part of an enum, that MUST correspond to an attribute key. We cannot use a new key name just for sorting.

I'm inclined to think adding an array would only add unnecessary complexity at this point.


EDIT: after some back and forth testing, we're settled on a simple approach: complex objects attribute, if meant to be used for sorting, shall define a `sortBy` property to tell in the schema how the sorting shall work. This field contains the `path` to a inner mapping, potentially deeply nested, and the `type` of the corresponding value.

This `type` is here for convenience, and we check at registration time that it corresponds to the given path. The goal is to avoid path/mapping resolving on every query (and this will never changed passed schema registration).

opencti-platform/opencti-graphql/src/utils/sorting.ts
```diff
@@ -0,0 +1,32 @@
...
+    // sorting on null dates results to an error, one way to fix it is to use missing: 0
+    // see https://github.com/elastic/elasticsearch/issues/81960
+    return { [orderCriteria]: { order: orderMode, missing: definition.type === 'date' ? 0 : '_last' } };
+  }
+
+  // for sorting by object attribute, we need the sortBy def to know which internal mapping to use
+  if (definition.type === 'object' && definition.format === 'standard') {
+    const { sortBy } = definition;
+    if (sortBy) {
+      const sortByMapping = getAttributeMappingFromPath(sortBy.path);
```
labo-flg: This function recursively checks the schema to find the correct mapping corresponding to the path.
sortBy path thus can be complex like `object_attr.some.inner.field`. `buildElasticSortingForAttributeCriteria` will build the right query sorting option for elastic.

opencti-platform/opencti-graphql/src/modules/attributes/internalObject-registrationAttributes.ts
```diff
@@ -240,6 +240,7 @@ const internalObjectsAttributes: { [k: string]: Array<AttributeDefinition> } = {
       multiple: false,
       upsert: false,
       isFilterable: false,
+      sortBy: { key: 'max_confidence', path: 'group_confidence_level.max_confidence' },
```
lndrtrbn: I do not understand why using an object instead of just a string `group_confidence_level.max_confidence`. I think you already explain to me something similar for filters but I still found the thing confusing.

opencti-platform/opencti-graphql/src/schema/attribute-definition.ts
```diff
@@ -53,7 +57,7 @@ export type EnumAttribute = { type: 'string', format: 'enum', values: string[] }
 export type VocabAttribute = { type: 'string', format: 'vocabulary', vocabularyCategory: string } & BasicDefinition;
 export type JsonAttribute = { type: 'string', format: 'json', multiple: false, schemaDef?: Record<string, any> } & BasicDefinition;
 export type FlatObjectAttribute = { type: 'object', format: 'flat' } & BasicDefinition;
-export type ObjectAttribute = { type: 'object', format: 'standard' } & BasicObjectDefinition;
+export type ObjectAttribute = { type: 'object', format: 'standard', sortBy?: ObjectSorting } & BasicObjectDefinition;
```
lndrtrbn: `NestedObjectAttribute` does not need this sortBy property also?

opencti-platform/opencti-graphql/tests/02-integration/01-database/elasticSearch-test.js
```diff
@@ -613,6 +613,46 @@ describe('Elasticsearch pagination', () => {
...
+      filters: [{ key: 'entity_type', operator: 'eq', values: ['Group'] }],
+      filterGroups: [],
+    };
+    const data = await elPaginate(testContext, ADMIN_USER, READ_ENTITIES_INDICES, {
+      filters,
+      orderBy: 'group_confidence_level',
+      orderMode: 'desc',
+    });
+    expect(data.edges.length).toEqual(5);
+    console.log(data.edges.map((e) => ({ confidence: e.node.group_confidence_level.max_confidence, name: e.node.name })));
```
lndrtrbn: oupsi

opencti-platform/opencti-graphql/tests/02-integration/01-database/elasticSearch-test.js
```diff
@@ -613,6 +613,46 @@ describe('Elasticsearch pagination', () => {
...
+      filters: [{ key: 'entity_type', operator: 'eq', values: ['Group'] }],
+      filterGroups: [],
+    };
+    const data = await elPaginate(testContext, ADMIN_USER, READ_ENTITIES_INDICES, {
+      filters,
+      orderBy: 'group_confidence_level',
+      orderMode: 'asc',
+    });
+    expect(data.edges.length).toEqual(5);
+    console.log(data.edges.map((e) => ({ confidence: e.node.group_confidence_level.max_confidence, name: e.node.name })));
```
lndrtrbn: oupsi

opencti-platform/opencti-graphql/src/utils/sorting.ts
```diff
@@ -0,0 +1,32 @@
...
+      if (sortByMapping?.type === 'numeric' || sortByMapping?.type === 'boolean' || sortByMapping?.type === 'date') {
```
lndrtrbn: This condition is not the same as the function `isDateNumericOrBooleanAttribute`?

lndrtrbn (commented):
Tested and works nicely. Still some remarks/questions about the implementation

opencti-platform/opencti-graphql/src/utils/sorting.ts
labo-flg: the functions isSomethingAttribute only check on the schema first level

Here I need to check on the inner mappings, in nested levels.

I want to do another refacto PR to change these utilities to take a SchemaAttribute as optional param, so that if not provided we take the root schema.

We could then use these utilities everywhere in all cases. But that's a potentially big refactor, these functions are everywhere. And I don't want to introduce chaos here :)

opencti-platform/opencti-graphql/src/modules/attributes/internalObject-registrationAttributes.ts
labo-flg: I agree with you, right now it does not seem useful.
The idea was to be a bit future proof, allowing to define multiple sorting paths, each one with a unique key that the API/client could provide. But we're far from here : the client uses fixed enum, enum values that must be attribute keys. 

I realize I need to either put this as an array, to embrace the future approach, or just get rid of the object and just give a path like you suggest.
Julien's approach was more toward the first solution so I'll go for that.

opencti-platform/opencti-graphql/src/schema/attribute-definition.ts
labo-flg: Yes good point

opencti-platform/opencti-graphql/src/utils/sorting.ts
```diff
@@ -0,0 +1,31 @@
...
+  // for sorting by object attribute, we need the sortBy def to know which internal mapping to use
+  if (definition.type === 'object' && (definition.format === 'standard' || definition.format === 'nested')) {
+    const { sortBy } = definition;
+    if (sortBy) {
+      if (sortBy.type === 'numeric' || sortBy.type === 'boolean' || sortBy.type === 'date') {
+        return { [sortBy.path]: { order: orderMode, missing: sortBy.type === 'date' ? 0 : '_last' } };
+      }
+      return { [`${sortBy.path}.keyword`]: { order: orderMode, missing: '_last' } };
+    }
+    throw UnsupportedError(`Sorting on [${orderCriteria}] is not supported: this criteria does not have a sortByPath definition in schema`);
```
lndrtrbn: `sortBy` is the name of the property no?

labo-flg: ah yes ! old version... I fix this.

opencti-platform/opencti-graphql/src/schema/schema-attributes.ts
```diff
@@ -105,6 +143,16 @@ export const schemaAttributesDefinition = {
           entityType
         });
       }
+      // Check sortBy on object
+      if (attribute.type === 'object' && attribute.format !== 'flat' && attribute.sortBy) {
```
SouadHadjiat, line 147: why `attribute.format !== 'flat'` ?

labo-flg: because it concerns non flat objects (nested / standard)

labo-flg: flat objects cannot define a sortBy

opencti-platform/opencti-graphql/src/modules/attributes/internalObject-registrationAttributes.ts
```diff
@@ -240,6 +240,7 @@ const internalObjectsAttributes: { [k: string]: Array<AttributeDefinition> } = {
       multiple: false,
       upsert: false,
       isFilterable: false,
+      sortBy: { path: 'group_confidence_level.max_confidence', type: 'numeric' },
```
SouadHadjiat, line 243: I'm not sure to understand how it would work here if we wanted to sort by another path then `group_confidence_level.max_confidence`. I can see in the unit tests that we could sort by `group_confidence_level.overrides.entity_type`, how would we be able to do this with this mapping definition ?

labo-flg: The theory has made us first think that way : being able to define different way of sorting a same object by different paths.

However this is not possible right now due to the way the API/client work (sorting criteria are fixed enum values that must correspond to a first level attribute name).
We decided to simplify and specify one possible path only for a given object.

>  I can see in the unit tests that we could sort by group_confidence_level.overrides.entity_type, how would we be able to do this with this mapping definition ?

This is only a unit test for the utility function `getAttributeMappingFromPath`. It's a generic purpose function, to recursively find the mapping at a given path in the schema. As said earlier, it would be great if we wanted more complex sorting possibilities, but we will not go into that direction for now.

lndrtrbn (approved):

opencti-platform/opencti-graphql/src/schema/schema-attributes.ts
SouadHadjiat: we could sort on flatten objects fields, but only as keywords (no support for numeric fields) https://www.elastic.co/guide/en/elasticsearch/reference/current/flattened.html#supported-operations
But it's not the subject here, we could say that we don't support sorting on flatten objects for now.

SouadHadjiat (approved):

PR #6199, opened 2024-03-01
Fix on instance trigger incorrect message when updating the standard Id

Kedae:


labo-flg (approved):
tested ok

labo-flg:
For a better solution, let's address https://github.com/OpenCTI-Platform/opencti/issues/6200

PR #6208, opened 2024-03-01
[frontend] Fix default pagination that could break screens 

Kedae:
Fixes the default count in pagination that could lead to screen failure with Relay.

This must be tested a lot, to ensure screens are not broken.
NO MERGING UNTIL STABLE 6.0.X


Kedae:
Stucked with Relay as DefaultValues are not working. Currently speaking with relay dev team.

PR #6269, opened 2024-03-04
[frontend] fix entity type filter right bar in containers (#6268)

### LINKED ISSUE #6268: Bulk deletion with the "select all" button deletes entities it shouldn't
## Description

When I delete entities in a report using the "select all" button, an unexpected behavior occurs: the selected entities are not deleted, but others are.

## Environment

OCTI 6.0.4

## Reproducible Steps

1. Create a report
2. Go to the "entities" tab
3. Add a country to it
4. Add incidents
5. Filter on the left to see only incidents
6. Select all 
7. Delete entities
-> Incidents are not deleted, but countries are.

Archidoit:
issue: https://github.com/OpenCTI-Platform/opencti/issues/6268


JeremyCloarec:
In the entities of a report, we now have an incorrect filter when selecting all elements and deleting as a task when having two or more items selected as entity types in the right bar. We have an EQ operator being used instead of an OR operator.
image

JeremyCloarec (approved):
tested OK

jpkha (approved):

PR #6272, opened 2024-03-04
[frontend] Fix cannot add TTP directly from matrix (#6112)

### LINKED ISSUE #6112: Cannot add TTP directly from matrix
## Description

In a matrix, when clicking on **Add** on a technique, the attack pattern is not added
## Environment

Testing

## Reproducible Steps

1. Go to an Intrusion Details
2. Knowledge -> Attack Pattern
3. Click on a technique in the matrix and press Add

## Expected Output

The technique should be added

## Actual Output

Nothing happens

## Screenshots (optional)
image

SarahBocognano:
### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/6112


lndrtrbn (commented):
The initial issue is fix.
Just, is it normal that we can add multiple times the same relation between an Intrusion and a Technique? That's not directly related to the fix but now that the Add button works we face this scenario

@SarahBocognano It actually doesn't add multiple times the same relation, but yes maybe we can "disable" the add button if the relation already exist

SarahBocognano:
Update: After a call with @nino-filigran , we will keep the "Add" button. Eveything is okay for product side, we can merge this issue like that

CelineSebe (approved):

Issue #6264, opened 2024-03-04
Labels are clickable in horizontal bars (home dashboard) and should not be


SamuelHassine:
## Description

Labels are clickable in horizontal bars (home dashboard) and should not be.

PR #6289, opened 2024-03-07
[frontend] horizontal bars representing labels should not be clickable

Archidoit:
issue: https://github.com/OpenCTI-Platform/opencti/issues/6264

### DIFF
--- a/opencti-platform/opencti-front/src/utils/Charts.js
+++ b/opencti-platform/opencti-front/src/utils/Charts.js
@@ -349,25 +349,33 @@ export const horizontalBarsChartOptions = (
       xAxisLabelClick: (event, chartContext, config) => {
         if (redirectionUtils) {
           const { labelIndex } = config;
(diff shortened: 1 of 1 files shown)


JeremyCloarec (commented):
Ok for labels not being clickable anymore. 
The change to the cursor isn't working properly though: we no longer have a pointer cursor when hovering clickable charts in dashboards (in the Most active threats widget for exemple).
I think we can keep the way cursors were displayed before, and only keep the changes to the xAsixLabelClick and click events

Archidoit:
> Ok for labels not being clickable anymore. The change to the cursor isn't working properly though: we no longer have a pointer cursor when hovering clickable charts in dashboards (in the Most active threats widget for exemple). I think we can keep the way cursors were displayed before, and only keep the changes to the xAsixLabelClick and click events

fixed :)

JeremyCloarec (approved):

Issue #6294, opened 2024-03-09
Max confidence level in user disabled field


SamuelHassine:
## Description

1. Go to a user
2. Update it
3. Put a max conf level to 400
4. Field is disabled and you cannot go back to 100 (error is normal)

PR #6299, opened 2024-03-11
[frontend] dont disable when value > limit (#6294)

frapuks:
### Proposed changes

- Update condition to manage diable : use initial value.

### Related issues

- #6294 

### Checklist


JeremyCloarec (commented):
This change breaks the behavior of the field when a user without enough confidence to edit an object opens the edit form: the field is not disabled when we want it to be

JeremyCloarec (approved):

PR #6309, opened 2024-03-11
[backend]Verify createdBy type on malware creation (#6291)

### LINKED ISSUE #6291: Creation of created-by relatiohship is not protected to target only identity
## Description
Creation of created-by relationship is not protected to target only identity.
At creation it will not fail but then the read will fail.

### Steps
1. Get a user id from the platform
2. Create a malware + read
```
mutation createMalware {
  malwareAdd(input: {
    name: "new_malware",
    createdBy: "dc5f1601-e4ff-4d17-9233-ec80ceef207d"
  }) {
    id
    createdBy {
      id
    }
  }
}
```

### Result
Create ok but return read fail.
`"message": "Runtime Object type \"User\" is not a possible type for \"Identity\".",`

### Expected
The creation must fail has USER is not a correct target for createdBy.

aHenryJard:
### Proposed changes

* Verify that the id given in createdBy relation is an Identity, if not throw exception and prevent from creating entity.
  * on creation
  * on update
  * on upsert
  * on relationAdd
* Test are written using malware use case, but the implementation is for all entity that have createdBy relationship
* Note that some changes in test (mainly '?' operator) are because of refactor of file from js to ts.

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/6291

### Checklist

### Further comments


opencti-platform/opencti-graphql/tests/02-integration/02-resolvers/malware-test.ts
```diff
@@ -0,0 +1,407 @@
...
+      mutation MalwareAdd($input: MalwareAddInput!) {
+        malwareAdd(input: $input) {
+          name
+          createdBy {
+            id
+          }
+        }
+      }
+    `;
+    const MALWARE_WITH_BAD_CREATED_BY = {
```
JeremyCloarec: small typo when copy/pasting from previous test, we can name it MALWARE_WITH_VALID_CREATED_BY

opencti-platform/opencti-graphql/src/database/middleware.js
```diff
@@ -144,7 +145,7 @@ import {
   isStixDomainObjectShareableContainer,
   isStixObjectAliased,
   resolveAliasesField,
-  STIX_ORGANIZATIONS_UNRESTRICTED,
+  STIX_ORGANIZATIONS_UNRESTRICTED
```
SouadHadjiat: why do we have this change ? the last ',' should be kept

opencti-platform/opencti-graphql/src/database/middleware.js
```diff
@@ -565,7 +566,33 @@ const idLabel = (label) => {
...
+ * @param context
+ * @param user
+ * @param createdById
+ * @returns {Bluebird.Promise<void>}
+ */
+export const validateCreatedBy = async (context, user, createdById) => {
+  if (createdById) {
+    const createdByEntity = await internalLoadById(context, user, createdById);
+    if (createdByEntity && createdByEntity.parent_types) {
+      if (!createdByEntity.parent_types.some((parent) => parent === ENTITY_TYPE_IDENTITY)) {
```
SouadHadjiat: I wonder if you could you use `isStixDomainObjectIdentity(createdByEntity.entity_type)` ?

opencti-platform/opencti-graphql/src/database/middleware.js
aHenryJard: I think it's ESLint that did it, I can revert that

opencti-platform/opencti-graphql/src/database/middleware.js
aHenryJard: you're right, it's working like a charm, thanks !

JeremyCloarec (approved):

opencti-platform/opencti-graphql/src/database/middleware.js
JeremyCloarec: I'm not sure I understand what's the convention for trailing commas in imports? There is already a mix of imports having a trailing comma on their last import line, and others not having it

PR #6323, opened 2024-03-12
[backend] use representatives in distrib graphs, including 'restricted' case (#6319)

### LINKED ISSUE #6319: Relationship Distribution gives inconsistent results when changing the limit parameter
## Description

On this example dashboard:
https://testing.octi.staging.filigran.io/dashboard/workspaces/dashboards/d4374a78-de09-4a00-a5c3-37b673bcea8b

The 2 widgets are the same (second one duplicated from the first), I only changed the "limit" parameter of the second one (from 10 to 20).

One widget gives 0 results, the other gives 2 results.

**This bug will not happen if the user has BYPASS capabilty**, so test with a user with only access to knowledge + dashboards.

I also tested with a user with ALL capabilities but not BYPASS, the bug happens. So probably a part of the code excluded when user has BYPASS.

## Environment

Testing environment 6.0

See user/role/dashboard with mention "6319"

## Screenshots (optional)

image

labo-flg:
### Proposed changes

* use `representative` in all distribution queries, removing a lot of lengthy duplicated query code.
* when distribution data gives restricted entities (as a side of a relationship), use 'restricted' representative
* refactor frontend code as much as possible in this context

### Related issues

Closes #6319 

### Further comments
A lot of changes but all the same.
In essence, I'm now using the `representative` field defined for every StixObject and StixRelationShip (and more, see `opencti.graphql`) in the frontend queries for Distribution widgets.
Backend side, distribution data will now send a dummy entity info, with "restricted" as representative, and no longer filter out these items (this is solving #6319).

Refactoring:
* `defaultValue` and other related utility functions, used everywhere in the app, have been moved to a specific file `defaultRepresentatives` out of `Graph.js`
* `defaultValue` and `defaultSecondaryValue` have been renamed to `getMainRepresentative` and `getSecondaryRepresentative` to avoid confusion with the settings default value concept.
*  `getMainRepresentative` and `getSecondaryRepresentative` first try to use the `representative` of the input object, then fallback to the previous possibilities
* new hook `useDistributionGraphData` to factorize common code in distribution graphs (utilities to turn query data into chart props)

A lot of other places in the code still use lengthy queries with repeated `... on XXX` instead of using `representative { main }`. I'm not addressing all of them here.

Ultimately, we shall be able to simplify `getMainRepresentative` and `getSecondaryRepresentative` to only use representative when possible, and other fields for the objects that do not define it (and a side question is... shall we define a representative for every object in the platform? that would make things lot easier).


opencti-platform/opencti-graphql/src/database/middleware.js
```diff
@@ -429,11 +429,16 @@ export const stixLoadByFilters = async (context, user, types, args) => {
...
+  const data = R.sortWith([orderingFunction(R.prop('value'))])(distribution);
+  // we try to resolve the complete batch of 100 results
   const resolveLabels = await elFindByIds(context, user, data.map((d) => d.label), { toMap: true });
-  return data // Depending of user access, info can be empty, must be filtered
+  // Depending of user access, info can be empty, must be filtered
+  const filteredEntities = data
     .filter((n) => isNotEmptyField(resolveLabels[n.label.toLowerCase()]))
     .map((n) => R.assoc('entity', resolveLabels[n.label.toLowerCase()], n));
+  // take only the 'limit' first match
+  // TODO: integrate data access restriction into aggregation query, and directly return 'limit' element instead of 100
```
labo-flg: we only have limited info in the connections, so it sounds actually impossible to get data restriction info (such as marking).

We could mitigate even more this issue by increasing the MAX_AGGREGATION_SIZE, but I'm unsure of the perf cost

labo-flg:
If you ask for the 10 first element matching a relationship distribution on target/source, you get the first 10. If you see only 3 it means 7 of them are not accessible to you (due to markings for instance). 
This issue is more about a confusing display, that might be perceived as buggy. 

Showing the first 10 that _YOU_ can see is not suitable : it means a dashboard is not presenting the same information to different users and that's not the point (dashboard are meant to be shared, etc.).

The solution we have in mind will require more work:
* Change the `relationshipDistribution` query API, to return only what's required for display and not the fully resolved entities (id, entity_type, representative)
* if user has no access over one of the entity, we put `<restricted>` as representative

This is a breaking change to the API but will allow us to simplify the query handling frontend side, where today we must make `... on` fragments for each `entity_type` possible, with a magic function to deduce the label to display.

labo-flg:
> If you ask for the 10 first element matching a relationship distribution on target/source, you get the first 10. If you see only 3 it means 7 of them are not accessible to you (due to markings for instance). This issue is more about a confusing display, that might be perceived as buggy.
> 
> Showing the first 10 that _YOU_ can see is not suitable : it means a dashboard is not presenting the same information to different users and that's not the point (dashboard are meant to be shared, etc.).
> 
> The solution we have in mind will require more work:
> 
> * Change the `relationshipDistribution` query API, to return only what's required for display and not the fully resolved entities (id, entity_type, representative)
> * if user has no access over one of the entity, we put `<restricted>` as representative
> 
> This is a breaking change to the API but will allow us to simplify the query handling frontend side, where today we must make `... on` fragments for each `entity_type` possible, with a magic function to deduce the label to display.

This initial idea implies we update the API at a high level, namely the `Distribution` type in our graphql spec. 
To limit the disruption, I suggest we actually use our current API without change: in fact, `Distribution.entity` can be null.
Let's assume that the item is not accessible when the entity is null, but we keep sending the computed stat for this object id.

labo-flg:
> This initial idea implies we update the API at a high level, namely the Distribution type in our graphql spec.
To limit the disruption, I suggest we actually use our current API without change: in fact, Distribution.entity can be null.
Let's assume that the item is not accessible when the entity is null, but we keep sending the computed stat for this object id.

In fact, no need to get around this problem... we need to use the representative even if it's a big refactoring. It's why it has been designed in the first place.

I'll use the representative in all queries and refactor to make it work. And use representative 'restricted' when needed.

labo-flg:
Note that Observed Data now use the representative, which is NOT the observed value but the time range for this observation.
image

As It's a container, there might be several entities in it... I'm unsure how we can improve this.

opencti-platform/opencti-front/src/private/components/analyses/reports/StixCoreObjectReportsHorizontalBars.jsx
```diff
@@ -68,7 +68,7 @@ class StixCoreObjectReportsHorizontalBars extends Component {
             && props.reportsDistribution.length > 0
           ) {
             const data = props.reportsDistribution.map((n) => ({
-              x: n.entity.name,
+              x: n.entity?.name ?? 'Restricted',
```
lndrtrbn: Do we want to translate this?

opencti-platform/opencti-front/src/private/components/observations/indicators/IndicatorsDonut.jsx
```diff
@@ -72,7 +72,7 @@ class IndicatorsDonut extends Component {
             let data = props.indicatorsDistribution;
             if (field && field.includes('internal_id')) {
               data = map(
-                (n) => assoc('label', n.entity.name, n),
+                (n) => assoc('label', n.entity?.name ?? 'Restricted', n),
```
lndrtrbn: Do we need a translation?

opencti-platform/opencti-front/src/private/components/analyses/reports/StixCoreObjectReportsHorizontalBars.jsx
labo-flg: This is supposedly coming from backend as a representative. I'll remove it from the code.

Utlimately, no, it won't be translated.

lndrtrbn (approved):

PR #6331, opened 2024-03-13
[frontend] Fix mass selection with the shift key in incidents container (#6283)

### LINKED ISSUE #6283: Mass selection with the shift key does not work on Incident
## Description

On the platform, it is possible to select a set of entities by selecting a first one -> scrolling -> selecting a second one by holding down the shift key -> all entities between the first and second are selected.

However, this mechanism doesn't work with Incidents entities

## Environment

OCTI 6.0.5

## Reproducible Steps

1. Navigate to Incident
2. In the list, select an Incident (i.e. check the box at the beginning of the line)
3. scroll a bit
4. Maintain shift and check the box of another report
-> only the second incident is selected, not all in between

## Expected Output

Do the repro case on Reports to understand what is expected

CelineSebe:
### Proposed changes

* add onToggleShiftEntity

### Related issues

* #6283 

### Checklist

### Further comments

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/events/incidents/IncidentLine.tsx
+++ b/opencti-platform/opencti-front/src/private/components/events/incidents/IncidentLine.tsx
@@ -61,6 +61,13 @@ interface IncidentLineComponentProps {
     event: React.SyntheticEvent
   ) => void;
   selectAll: boolean;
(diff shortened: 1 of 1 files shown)


SarahBocognano (approved):
Tested ✅

PR #6341, opened 2024-03-13
[frontend] Added auto collapse and submenu icon switches to settings

Bonsai8863:
### Proposed changes

* Add setting to toggle hiding of left nav submenu icons
* Add setting to toggle auto collapse left nav submenus

### Related issues

* https://filigran-community.slack.com/archives/C06CETVDJ5R/p1709241125254959

### Checklist

- [x] I wrote/updated test cases for the relevant uses case

### DIFF
--- a/opencti-platform/opencti-front/lang/front/fr.json
+++ b/opencti-platform/opencti-front/lang/front/fr.json
@@ -2673,7 +2673,7 @@
   "Strategic": "Stratégique",
   "Link created": "Lien créé",
   "Public dashboard": "Tableau de bord public",
-  "Show left navigation submenu icons": "Afficher les icônes du sous-menu de navigation de gauche",
-  "Auto collapse submenus in left navigation": "Sous-menus à réduction automatique dans la navigation de gauche",
+  "Show left navigation submenu icons": "Afficher les icônes des sous-menus de la navigation de gauche",
+  "Auto collapse submenus in left navigation": "Fermer automatiquement les sous-menus de la navigation de gauche",
   "Unless specific groups are selected, user will be created with default groups only.": "À moins que des groupes spécifiques ne soient sélectionnés, l'utilisateur sera créé avec les groupes par défaut uniquement."
 }
--- a/opencti-platform/opencti-front/lang/front/fr.json
+++ b/opencti-platform/opencti-front/lang/front/fr.json
@@ -2673,5 +2673,7 @@
   "Strategic": "Stratégique",
   "Link created": "Lien créé",
   "Public dashboard": "Tableau de bord public",
+  "Show left navigation submenu icons": "Afficher les icônes du sous-menu de navigation de gauche",
+  "Auto collapse submenus in left navigation": "Sous-menus à réduction automatique dans la navigation de gauche",
   "Unless specific groups are selected, user will be created with default groups only.": "À moins que des groupes spécifiques ne soient sélectionnés, l'utilisateur sera créé avec les groupes par défaut uniquement."
 }
--- a/opencti-platform/opencti-front/src/private/components/nav/LeftBar.jsx
+++ b/opencti-platform/opencti-front/src/private/components/nav/LeftBar.jsx
@@ -175,9 +175,6 @@ const useStyles = makeStyles((theme) => createStyles({
   menuItemSmallText: {
     padding: '1px 0 0 20px',
   },
-  menuToggleSubmenuIcons: {
-    marginRight: 15,
-  },
 }));
 
 const StyledTooltip = styled(({ className, ...props }) => (
@@ -226,7 +223,6 @@ const LeftBar = () => {
     data: useRef(null),
     settings: useRef(null),
   };
-  // console.log(localStorage.getItem('selectedMenu'));
   const [selectedMenu, setSelectedMenu] = useState(
     JSON.parse(localStorage.getItem('selectedMenu') ?? '[]'),
   );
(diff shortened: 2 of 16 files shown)


lndrtrbn:
Hello, thank you for your contribution !

We love the idea of saving auto collapse and submenu icons as preferences. But there is an issue with the implementation because the Settings page is not a page accessible by all the users (admin page).

Below, the implementation we suggest to do so every user can set their preferences:

- Move the toggle inputs in the Profile page instead of Settings (accessible from top right button > Profile),
- Save the 2 preferences in database below "user profile" and fetch them through the "me" object while loading the app instead of using local storage as the user will lose the preferences if switching computer by using local storage,
- Use the local storage to keep in browser memory the state of the left menu items so when refreshing the app the menu will keep the same state for each item (if it should be closed or opened).

Bonsai8863:
> Hello, thank you for your contribution !
> 
> We love the idea of saving auto collapse and submenu icons as preferences. But there is an issue with the implementation because the Settings page is not a page accessible by all the users (admin page).
> 
> Below, the implementation we suggest to do so every user can set their preferences:
> 
> * Move the toggle inputs in the Profile page instead of Settings (accessible from top right button > Profile),
> * Save the 2 preferences in database below "user profile" and fetch them through the "me" object while loading the app instead of using local storage as the user will lose the preferences if switching computer by using local storage,
> * Use the local storage to keep in browser memory the state of the left menu items so when refreshing the app the menu will keep the same state for each item (if it should be closed or opened).

@lndrtrbn Thanks for the feedback! All your points have been addressed in the latest push.

opencti-platform/opencti-front/lang/front/fr.json
```diff
@@ -2673,5 +2673,7 @@
   "Strategic": "Stratégique",
   "Link created": "Lien créé",
   "Public dashboard": "Tableau de bord public",
+  "Show left navigation submenu icons": "Afficher les icônes du sous-menu de navigation de gauche",
```
lndrtrbn:
```suggestion
  "Show left navigation submenu icons": "Afficher les icônes des sous-menus de la navigation de gauche",
```

opencti-platform/opencti-front/lang/front/fr.json
```diff
@@ -2673,5 +2673,7 @@
...
+  "Auto collapse submenus in left navigation": "Sous-menus à réduction automatique dans la navigation de gauche",
```
lndrtrbn:
```suggestion
  "Auto collapse submenus in left navigation": "Fermer automatiquement les sous-menus de la navigation de gauche",
```

opencti-platform/opencti-front/src/private/components/nav/LeftBar.jsx
```diff
@@ -220,25 +226,47 @@ const LeftBar = () => {
     data: useRef(null),
     settings: useRef(null),
   };
-  const [selectedMenu, setSelectedMenu] = useState(null);
+  // console.log(localStorage.getItem('selectedMenu'));
```
lndrtrbn: Leftover of debug I guess

opencti-platform/opencti-front/src/private/components/nav/LeftBar.jsx
```diff
@@ -175,6 +175,9 @@ const useStyles = makeStyles((theme) => createStyles({
   menuItemSmallText: {
     padding: '1px 0 0 20px',
   },
+  menuToggleSubmenuIcons: {
+    marginRight: 15,
+  },
```
lndrtrbn: It seems this style is never used?

opencti-platform/opencti-front/src/private/components/nav/LeftBar.jsx
```diff
@@ -320,12 +348,19 @@ const LeftBar = () => {
...
                     {entry.icon}
                   </ListItemIcon>}
                   <ListItemText
                     classes={{ primary: classes.menuSubItemText }}
                     primary={t_i18n(entry.label)}
+                    primaryTypographyProps={{
+                      style: {
+                        whiteSpace: 'wrap',
+                        lineHeight: 1.2,
+                        padding: '4px 0 4px 10px',
```
lndrtrbn: Is this padding property really useful? It doesn't seem to change much, the less custom css we have, the better

opencti-platform/opencti-graphql/src/domain/user.js
```diff
@@ -516,6 +516,8 @@ export const addUser = async (context, user, newUser) => {
     R.assoc('account_status', newUser.account_status ? newUser.account_status : DEFAULT_ACCOUNT_STATUS),
     R.assoc('account_lock_after_date', newUser.account_lock_after_date),
     R.assoc('unit_system', newUser.unit_system),
+    R.assoc('submenu_show_icons', newUser.submenu_show_icons),
+    R.assoc('submenu_auto_collapse', newUser.submenu_auto_collapse),
```
lndrtrbn: I'm not sure we should add those lines as those properties are not present in the input to create a user. So it will be always undefined.

lndrtrbn (commented):
I have tested locally and it works fine 👍
I just make some small remarks that should be easily fixed.

I was also questioned myself about this behavior: if several person use the same computer and browser then the state of the left navigation saved in local storage will collide between those users and they will have inconsistencies when login. It could be fixed by using the id of the user in the key for local storage. But I don't know if it is a real issue that need to be tackled, I'll ask info in the team.

EDIT : this local storage collision won't be tackle here

PR #6349, opened 2024-03-14
[frontend] fix filters adding for TimeLine views, Alert Live, and label icon (#6348)

### LINKED ISSUE #6347: Quick adding of label filter works only once
## Description

Quick adding of label filter works only once

## Reproducible Steps
Go to the reports list.
Click on a label. The associated label filter is added.
Click on an other label. No filter is added.

### LINKED ISSUE #6348: Can't add several values for a filter in Alert Live and Time Line views
Can't add more than one value for each filters in
- Alert Live creation,
- Alert live edition,
- TimeLine views (of Report, Incident, Case RFT, Case RFI)

And
implement new filters UI for Alert Live creation and edition

Archidoit:
PR content:
- issue https://github.com/OpenCTI-Platform/opencti/issues/6348 
- issue https://github.com/OpenCTI-Platform/opencti/issues/6347
- new filters UI in Alert Live creation and Alert live edition

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/analyses/reports/ReportKnowledge.jsx
+++ b/opencti-platform/opencti-front/src/private/components/analyses/reports/ReportKnowledge.jsx
@@ -14,7 +14,7 @@ import ReportPopover from './ReportPopover';
 import AttackPatternsMatrix from '../../techniques/attack_patterns/AttackPatternsMatrix';
 import { buildViewParamsFromUrlAndStorage, saveViewParameters } from '../../../../utils/ListParameters';
 import ReportKnowledgeTimeLine, { reportKnowledgeTimeLineQuery } from './ReportKnowledgeTimeLine';
-import { constructHandleAddFilter, constructHandleRemoveFilter, emptyFilterGroup, filtersAfterSwitchLocalMode } from '../../../../utils/filters/filtersUtils';
+import { useConstructHandleAddFilter, constructHandleRemoveFilter, emptyFilterGroup, filtersAfterSwitchLocalMode } from '../../../../utils/filters/filtersUtils';
 import ContentKnowledgeTimeLineBar from '../../common/containers/ContainertKnowledgeTimeLineBar';
 import ContainerContent, { containerContentQuery } from '../../common/containers/ContainerContent';
 import investigationAddFromContainer from '../../../../utils/InvestigationUtils';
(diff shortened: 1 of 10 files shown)


opencti-platform/opencti-front/src/private/components/analyses/reports/ReportKnowledge.jsx
```diff
@@ -251,181 +253,189 @@ class ReportKnowledgeComponent extends Component {
...
-              variables={{ id: report.id }}
-              render={({ props }) => {
-                if (props && props.container) {
-                  return <ContainerContent containerData={props.container} />;
-                }
-                return (
-                  <Loader
-                    variant={LoaderVariant.inElement}
-                    withTopMargin={true}
+      <UserContext.Consumer>
```
labo-flg: ouch. But probably "a bit" more work if we want to refactor this component and use hooks :)

opencti-platform/opencti-front/src/private/components/analyses/reports/ReportKnowledge.jsx
```diff
@@ -251,181 +253,189 @@ class ReportKnowledgeComponent extends Component {
...
                         this,
                       )}
-                      handleToggleColorsReversed={this.handleToggleColorsReversed.bind(
+                      timeLineFunctionalDate={timeLineFunctionalDate}
+                      handleToggleTimeLineFunctionalDate={this.handleToggleTimeLineFunctionalDate.bind(
                         this,
                       )}
-                      handleToggleModeOnlyActive={this.handleToggleModeOnlyActive.bind(
+                      timeLineFilters={timeLineFilters}
+                      handleAddTimeLineFilter={this.handleAddTimeLineFilter.bind(
```
SouadHadjiat: Just from reading the code, it looks like handleAddTimeLineFilter has a different signature now, but we pass it as a parameter to ContentKnowledgeTimeLineBar which then uses it directly to pass it to Filters component (handleAddFilter={handleAddTimeLineFilter}) but handleAddFilter type doesn't have the same signature anymore.

opencti-platform/opencti-front/src/private/components/analyses/reports/ReportKnowledge.jsx
Archidoit: yes it's far better to use hook... but need refacto ^^'

opencti-platform/opencti-front/src/private/components/analyses/reports/ReportKnowledge.jsx
Archidoit: fixed :)

PR #6366, opened 2024-03-15
[backend] Update errors that are not retryable

Kedae:
We need the actual error to be logged and not bypassed because they are not retryable


opencti-platform/opencti-graphql/src/graphql/loggerPlugin.js
```diff
@@ -88,7 +88,12 @@ export default {
         if (isCallError) {
           const currentError = head(context.errors);
           const callError = currentError.originalError ? currentError.originalError : currentError;
-          const isRetryableCall = isNotEmptyField(origin?.call_retry_number) && callError.name !== UNSUPPORTED_ERROR;
+          const isRetryableCall = isNotEmptyField(origin?.call_retry_number) && ![
+            UNSUPPORTED_ERROR,
+            ALREADY_DELETED_ERROR,
+            TYPE_DUPLICATE_ENTRY,
```
richard-julien: DuplicateEntryError seems never used

labo-flg (approved):

helene-nguyen (approved):

Issue #6327, opened 2024-03-13
Cannot hide entity type "External references"


rattat0r:
## Description

There is no option to hide the "External references" section in settings.

## Environment

1. OS: Ubuntu 20.04.6
2. OpenCTI version: OpenCTI 6.0.4
3. OpenCTI client: frontend

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. In [settings ](https://testing.octi.staging.filigran.io/dashboard/settings)> hidden entities type , Hide these types:
![image](https://github.com/OpenCTI-Platform/opencti/assets/122795003/ae50db05-2984-4651-9b79-e7a326964e28)
2. There is an External references type on the sidebar
![image](https://github.com/OpenCTI-Platform/opencti/assets/122795003/4135fc94-5efb-40ff-92b8-696368ceebe7)

## Expected Output

It is possible to hide the "External references" section in the settings

## Actual Output

There is no option to hide the "External references" section in settings

PR #6371, opened 2024-03-15
[frontend/backend] Cannot hide entity type "External references" (#6327)

SarahBocognano:
### Proposed changes

* Add missing external references in subtypes, available settings and hiddenTypesField

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/6327

### Checklist

### Further comments

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/settings/hidden_types/HiddenTypesField.tsx
+++ b/opencti-platform/opencti-front/src/private/components/settings/hidden_types/HiddenTypesField.tsx
@@ -15,7 +15,7 @@ import { fieldSpacingContainerStyle } from '../../../../utils/field';
 import HiddenTypesIndicator from './HiddenTypesIndicator';
 
 export const groups = new Map<string, string[]>([
-  ['Analysis', ['Report', 'Grouping', 'Note', 'Malware-Analysis']],
+  ['Analysis', ['Report', 'Grouping', 'Note', 'Malware-Analysis', 'External-Reference']],
   ['Cases', ['Case-Incident', 'Feedback', 'Case-Rfi', 'Case-Rft', 'Task']],
   ['Events', ['stix-sighting-relationship', 'Incident', 'Observed-Data']],
   ['Observations', ['Stix-Cyber-Observable', 'Artifact', 'Indicator', 'Infrastructure']],
--- a/opencti-platform/opencti-front/src/private/components/settings/hidden_types/HiddenTypesField.tsx
+++ b/opencti-platform/opencti-front/src/private/components/settings/hidden_types/HiddenTypesField.tsx
@@ -15,9 +15,9 @@ import { fieldSpacingContainerStyle } from '../../../../utils/field';
 import HiddenTypesIndicator from './HiddenTypesIndicator';
 
 export const groups = new Map<string, string[]>([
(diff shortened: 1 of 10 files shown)


Archidoit:
In hidden types list display, 'External reference' should be displayed after 'malware analysis' (not after 'Report') : 
image


Done ✅

opencti-platform/opencti-front/tests_e2e/settings/parameters.spec.ts
```diff
@@ -45,3 +47,13 @@ test('Check Logo replacement', async ({ page }) => {
...
+
+test('Check Hidden Entities', async ({ page }) => {
+  const leftBarPage = new LeftBarPage(page);
+  const settingPage = new SettingsPage(page);
+  await page.goto('/');
+  await page.getByTestId('ChevronRightIcon').click();
+  await leftBarPage.clickOnMenu('Settings', 'Parameters');
+  await settingPage.getHiddenEntityTypeSelect().click();
+  await expect(settingPage.getExternalReference()).toBeVisible();
+});
```
lndrtrbn: I'm not a fan of this test. E2E tests have a high cost (~30sec in CI for a small test) and so their purpose is to validate a user scenario that is important we know it works. 

Here we are just validating the presence of an item in a list, if we really want to check those kind of things it should be done with unit test using `vite` but not in E2E.

lndrtrbn: Imo just remove it

lndrtrbn (changes requested):

opencti-platform/opencti-front/tests_e2e/settings/parameters.spec.ts
SarahBocognano: Well that was a user scenario as it was a bug and noticed by a openCTI user himself. 
So I don't understand, do we or do we not want tests ? I think the team needs to be precise on when to do or not do tests.

lndrtrbn: I'll open a discussion between the team to discuss about it

opencti-platform/opencti-front/tests_e2e/settings/parameters.spec.ts
```diff
@@ -1,5 +1,6 @@
 import { test } from '@playwright/test';
-import { expect } from '../fixtures/baseFixtures';
+// eslint-disable-next-line import/no-extraneous-dependencies
```
lndrtrbn: This comment should not be necessary anymore with the new eslint config

opencti-platform/opencti-front/src/private/components/settings/hidden_types/HiddenTypesField.tsx
```diff
@@ -169,47 +169,52 @@ const HiddenTypesField: FunctionComponent<HiddenTypesFieldProps> = ({
...
-        <Box sx={{ display: 'flex', flexWrap: 'wrap', gap: 0.5 }}>
-          {selected.map((value) => (
-            <Chip key={value} label={t_i18n(`entity_${value}`)} />
-          ))}
-        </Box>
-      )}
-    >
-      {computeItems()}
-    </Field>
+    <div data-testid="hiddenEntityTypes">
```
lndrtrbn: Not needed anymore ?

lndrtrbn (approved):

Issue #6374, opened 2024-03-15
Live stream element dependency containers resolution fetching improvement


richard-julien:
## Description

In a specific case of magic resolution in live stream, volume of reports can sometimes be to large to be correctly fetch by elastic.
This situation leads to read failure and stream consumption stop.

## Hints
Let explain a bit why this "magic" is needed.
Assuming you want to share containers inside a live stream, you will create for example a live stream with filters configuration like this : Entity type = Report and Label = Sharing.
With this kind of configuration you expect to also share everything related the the container. But you also need to detect a change inside a element of the report to be able to share it into the stream to update the remove platform. However this element is not container and have no label. To handle this the platform needs to auto detect this behavior to let this element be pushed in the stream. For this we need to list reports connected to this element, resolve them and try to use the configured filters on it. This issue aims to improve this algorithm to fetch less data.

## Todo
Change the approach to forge the correct filters to directly count the potential result
If count > 0 publication must be done

PR #6375, opened 2024-03-15
[backend] Live stream element dependency containers resolution fetching improvement (#6374)

richard-julien:
See #6374

### DIFF
--- a/opencti-platform/opencti-graphql/src/graphql/sseMiddleware.js
+++ b/opencti-platform/opencti-graphql/src/graphql/sseMiddleware.js
@@ -43,7 +43,7 @@ import { findFiltersFromKey } from '../utils/filtering/filtering-utils';
 import { convertFiltersToQueryOptions } from '../utils/filtering/filtering-resolution';
 import { getParentTypes } from '../schema/schemaUtils';
 import { STIX_EXT_OCTI } from '../types/stix-extensions';
-import { listAllEntities, listAllRelations } from '../database/middleware-loader';
+import { listAllRelations } from '../database/middleware-loader';
 import { RELATION_OBJECT } from '../schema/stixRefRelationship';
 import { getEntitiesListFromCache } from '../database/cache';
 import { ENTITY_TYPE_STREAM_COLLECTION } from '../schema/internalObject';
--- a/opencti-platform/opencti-graphql/src/graphql/sseMiddleware.js
+++ b/opencti-platform/opencti-graphql/src/graphql/sseMiddleware.js
@@ -6,8 +6,8 @@ import conf, { basePath, logApp } from '../config/conf';
 import { authenticateUserFromRequest, TAXIIAPI } from '../domain/user';
 import { createStreamProcessor, EVENT_CURRENT_VERSION } from '../database/redis';
 import { generateInternalId } from '../schema/identifier';
-import { stixLoadById, stixLoadByIds, storeLoadByIdsWithRefs } from '../database/middleware';
-import { elList } from '../database/engine';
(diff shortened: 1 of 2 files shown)


aHenryJard (approved):
Code looks good to me, however I'm not sure on how to test. Do you know if it's already covered by automated tests ?

opencti-platform/opencti-graphql/src/graphql/sseMiddleware.js
```diff
@@ -598,23 +599,34 @@ const createSseMiddleware = () => {
...
-                    }
+                      return !isContainerMatching; // If match is found, stop the listing process
+                    };
+                    await listAllEntities(context, user, [ENTITY_TYPE_CONTAINER], {
+                      first: CONTAINER_FINDER_BATCH_RESOLUTION,
+                      baseData: true,
+                      orderBy: 'updated_at', // Get container by newly updated
+                      orderMode: 'desc',
+                      connectionFormat: false,
+                      filters: {
```
SouadHadjiat: couldn't we add `streamFilters` here to query only elements filtered by this stream ? we wouldn't even need to read the results, only count.

richard-julien: We could and we should. Just commit a new approach following this remark. Thanks

opencti-platform/opencti-graphql/src/utils/filtering/filtering-resolution.ts
```diff
@@ -217,11 +217,11 @@ export const resolveFiltersMapForUser = async (context: AuthContext, user: AuthU
...
 
 export const convertFiltersToQueryOptions = async (filters: FilterGroup | null, opts: any = {}) => {
-  const { after, before, defaultTypes = [], field = 'updated_at', orderMode = 'asc' } = opts;
+  const { after, before, defaultTypes = [], extraFilters = [], field = 'updated_at', orderMode = 'asc' } = opts;
   const types = [...defaultTypes];
   let finalFilters = filters;
-  if (after || before) {
-    const filtersContent = [];
+  if (after || before || extraFilters.length > 0) {
+    const filtersContent = [...extraFilters];
```
SouadHadjiat, line 224: I'm a bit confused about this, how would it work with after and before ? I know in this current case extraFilters is not used with after or before, but what if it's used in the future ? what filter would it create ? would it be a "And" or "Or" condition between them ?

richard-julien: For now its a AND condition that feat the needs i needs. Don't want to complexify the options before having a real use case.

SouadHadjiat: so it's an AND by default in filters ? I just find the code less easy to read when having an if condition between after before and extraFilters, since they are not related (after / before are related). I'm wondering if we really need this extraFilters param, couldn't we create the right FilterGroup object before calling this method ?

aHenryJard (approved):

opencti-platform/opencti-graphql/src/utils/filtering/filtering-resolution.ts
richard-julien: Before and after are not directly related, thats extra options that can accumulate a first set of filters. Its a AND because these filters are use with

```
mode: FilterMode.And,
filters: filtersContent,
```

ExtraFilters options is the simplest way to add some filters at the firt level to accumulate with other options.

SouadHadjiat (approved):

PR #6387, opened 2024-03-19
[backend] Improve bulk Stix loader to support 100K+ relations loading in one bulk (#6386)

### LINKED ISSUE #6386: Improve bulk Stix loader to support 100K+ relations loading in one bulk
## Description

Improve bulk Stix loader to support 100K+ relations loading in one bulk.
The error is due to a nodejs limitation where function doesnt support more than 100K parameters. This limitation throw the error "Maximum call size exceeded" when we use spread operators to add result in the global accumulation array.

## Reproducible Steps

1. Get a platform with lot of massive reports
2. Create a taxii sharing focus on reports
3. Consume the taxii collection

## Expected Output

List of reports correctly exported.

## Actual Output

Execution failure for "Maximum call size exceeded"

## Additional information
Improvement is done with 2 different approaches.
1. Accumulate the result in a callback instead of spread the complete list at the end.
2. Improve nested connections internal filter to be based on query terms to prevent field explosions

richard-julien:
See #6386

### DIFF
--- a/opencti-platform/opencti-graphql/src/database/engine.js
+++ b/opencti-platform/opencti-graphql/src/database/engine.js
@@ -1605,17 +1605,25 @@ const buildLocalMustFilter = async (validFilter) => {
       const parentKey = arrayKeys.at(0);
       const { key: nestedKey, values: nestedValues, operator: nestedOperator = 'eq' } = nestedElement;
       const nestedShould = [];
-      for (let i = 0; i < nestedValues.length; i += 1) {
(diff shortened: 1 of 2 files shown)


aHenryJard (approved):
Tested on report up to 501 relations, work fine. I also tested the for loop protection with MAX_EVENT_LOOP_PROCESSING_TIME it's working as expected too.

I haven't tested with 100k+ relations.

PR #6388, opened 2024-03-19
[frontend] add e2e tests for filters

Archidoit:
add e2e tests for filters

### DIFF
--- a/opencti-platform/opencti-front/tests_e2e/filters/addFilter.spec.ts
+++ b/opencti-platform/opencti-front/tests_e2e/filters/addFilter.spec.ts
@@ -5,11 +5,23 @@ import ReportPage from '../model/report.pageModel';
 test('Add a new filter in the observables list and check the filter is still present when we come back to the page', async ({ page }) => {
   await page.goto('/dashboard/observations/observables');
   const filterUtils = new FiltersPageModel(page);
-  await filterUtils.addFilter('Entity type', 'Artifact');
+  await filterUtils.addFilterFromList('Entity type', 'Artifact');
   await expect(page.getByRole('button', { name: 'Entity type = Artifact' })).toBeVisible();
   await page.goto('/dashboard/');
   await page.goto('/dashboard/observations/observables');
   await expect(page.getByRole('button', { name: 'Entity type = Artifact' })).toBeVisible();
+});
+
+test('Add different filters with different operators', async ({ page }) => {
+  await page.goto('/dashboard/observations/observables');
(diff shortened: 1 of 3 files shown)


opencti-platform/opencti-front/tests_e2e/model/report.pageModel.ts
```diff
@@ -31,4 +31,8 @@ export default class ReportPage {
   getItemFromList(name: string) {
     return this.page.getByRole('link', { name }).first();
   }
+
+  selectAllReports() {
+    return this.page.getByTestId('report-page').getByRole('listitem').getByRole('checkbox').check();
```
JeremyCloarec: just a comment, not sure if it will be relevant for your case, but for some of the e2e tests I did, the check() method wasn't working properly for checkboxes, I add to use a click() instead

opencti-platform/opencti-front/tests_e2e/filters/addFilter.spec.ts
```diff
@@ -1,12 +1,47 @@
...
-  await filterUtils.addFilter('Entity type', 'Artifact');
+  await filterUtils.addFilterFromList('Entity type', 'Artifact');
   await expect(page.getByRole('button', { name: 'Entity type = Artifact' })).toBeVisible();
   await page.goto('/dashboard/');
   await page.goto('/dashboard/observations/observables');
   await expect(page.getByRole('button', { name: 'Entity type = Artifact' })).toBeVisible();
 });
+
+test('Add different filters with different operators', async ({ page }) => {
+  await page.goto('/dashboard/observations/observables');
```
labo-flg: having multiple `test(...)` is good for readibility and all, but I think it makes the test suite very long to run.

Each call to `page.goto` takes forever to load, then everything is quite rapid.

If you could run all the tests that occur in the same page in the same `test()` that would make the suite run faster.
That's why the Reports CRUD tests done by @lndrtrbn & co. are all in one place.

lndrtrbn: Indeed 👍

Archidoit: fixed

opencti-platform/opencti-front/tests_e2e/model/report.pageModel.ts
Archidoit: fixed

Archidoit:
Closed because too old and deprecated in regards of e2e tests

PR #6397, opened 2024-03-20
[frontend] add entity_type filter in Victimology tab (#6391)

### LINKED ISSUE #6391: Several issues in filters of entities' knowledges tab
## Description

The filters available in the entity knowledge tabs are strange. For example, we don't have an "entity type" filter, but we do have "event type" filters...

NB: More generally, the filters on the knowledge tab timelines need to be reviewed. Some don't seem to work, some are missing, others we don't really understand why they're there, etc. cf another issue: https://github.com/OpenCTI-Platform/opencti/issues/6390

## Environment

OCTI 6.0.7

## Reproducible Steps

1. Go to the intrusion set APT29
2. Go to the knowledge tab
3. Go to victimology
4. Look into the filters available

## Screenshots

Screenshot 2024-03-19 162708

Archidoit:
PR content:
- add entity_type filter in Victimology tabs
- fix filters repartition between 'most used filters' and 'all other filters' in filters list in Victimology

issue: https://github.com/OpenCTI-Platform/opencti/issues/6391

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/common/lists/ListFilters.jsx
+++ b/opencti-platform/opencti-front/src/private/components/common/lists/ListFilters.jsx
@@ -63,7 +63,8 @@ const ListFilters = ({
     ? availableFilterKeys
       .map((key) => {
         const subEntityTypes = filterKeysMap.get(key)?.subEntityTypes ?? [];
-        const isFilterKeyForAllTypes = subEntityTypes.some((subType) => entityTypes.includes(subType));
+        const isFilterKeyForAllTypes = (entityTypes.length === 1 && subEntityTypes.some((subType) => entityTypes.includes(subType)))
+          || (entityTypes.length > 1 && entityTypes.every((subType) => subEntityTypes.includes(subType)));
         return {
           value: key,
           label: t_i18n(filterKeysMap.get(key)?.label ?? key),
--- a/opencti-platform/opencti-front/src/components/list_lines/ListLines.jsx
+++ b/opencti-platform/opencti-front/src/components/list_lines/ListLines.jsx
@@ -442,9 +442,7 @@ class ListLines extends Component {
           availableRelationFilterTypes={availableRelationFilterTypes}
           redirection
           entityTypes={entityTypes}
-          restrictedFiltersConfig={{
-            filterRemoving: additionalFilterKeys,
-          }}
+          restrictedFiltersConfig={additionalFilterKeys?.restrictedFiltersConfig ?? undefined}
         />
         <ErrorBoundary key={keyword}>
           {message && (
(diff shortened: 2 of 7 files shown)


opencti-platform/opencti-front/src/private/components/common/lists/ListFilters.jsx
```diff
@@ -63,7 +63,8 @@ const ListFilters = ({
     ? availableFilterKeys
       .map((key) => {
         const subEntityTypes = filterKeysMap.get(key)?.subEntityTypes ?? [];
-        const isFilterKeyForAllTypes = subEntityTypes.some((subType) => entityTypes.includes(subType));
+        const isFilterKeyForAllTypes = (entityTypes.length === 1 && subEntityTypes.some((subType) => entityTypes.includes(subType)))
```
SouadHadjiat, line 66: Not sure to understand this condition : isn't it enough to check that entity types are all included in subEntityTypes ? 
`entityTypes.length > 0 && entityTypes.every((subType) => subEntityTypes.includes(subType))`

But I don't really understand exactly what is `isFilterKeyForAllTypes`.

Archidoit: No it's not enough, because we can have : 
entityTypes = ['Stix-Core-Object']

and subEntityTypes = ['Stix-Core-Object', 'Stix-Domain-Object', 'City', etc]

SouadHadjiat: so in this case `isFilterKeyForAllTypes` is true ? what is `isFilterKeyForAllTypes` ?

Archidoit: if the filter key is for all the entity types of the screen, the filter key appears in the 'most used filters' section of the filters list

Archidoit: image

Archidoit: else, it appears in the 'all other filters' section: 
image

SouadHadjiat: There is still something I don't understand : `Author` is not listed in most used filters in victimology, but in all other filters : 
image

But in data => entities we can see `Author` in most used filters.
image
which is expected since Author is a common property for all stix objects.

Archidoit: fixed :)

SarahBocognano (approved):

PR #6413, opened 2024-03-22
[frontend] Fix filtering for toolbar

### LINKED ISSUE #6411: "Select all" selects the whole platform, not just what's on the screen
## Description

When I do a "select all" to delete relations, when I look at my background task, I realize that it contains 160K elements. That's not what I selected at the beginning.

## Environment

OCTI 6.0.7

## Reproducible Steps

1. Got to a threat actor group
2. Go to knowledge > incidents (if you don't have incident here, add one)
3. Go to the "relation view" 
4. Select all and click on the bin in the bottom right corner to remove the relation.
5. Launch the background task and check the amount of entity that is processed.

Kedae:
Closes #6411

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/common/stix_core_relationships/views/EntityStixCoreRelationshipsRelationshipsView.tsx
+++ b/opencti-platform/opencti-front/src/private/components/common/stix_core_relationships/views/EntityStixCoreRelationshipsRelationshipsView.tsx
@@ -143,18 +143,7 @@ const EntityStixCoreRelationshipsRelationshipsView: FunctionComponent<EntityStix
   }
   const contextFilters: FilterGroup = {
     mode: 'and',
-    filters: [
-      ...predefinedFilters,
-      {
-        key: 'regardingOf',
-        operator: 'eq',
-        mode: 'and',
-        values: [
-          { key: 'id', values: [entityId], operator: 'eq', mode: 'or' },
-          { key: 'relationship_type', values: relationshipTypes, operator: 'eq', mode: 'or' },
-        ] as unknown as string[], // Workaround for typescript waiting for better solution
-      },
-    ],
+    filters: predefinedFilters,
     filterGroups: userFilters && isFilterGroupNotEmpty(userFilters) ? [userFilters] : [],
   };
 
@@ -277,6 +266,7 @@ const EntityStixCoreRelationshipsRelationshipsView: FunctionComponent<EntityStix
         search={searchTerm}
         handleClearSelectedElements={handleClearSelectedElements}
         variant="medium"
+        type={'stix-core-relationship'}
       />
       <Security needs={[KNOWLEDGE_KNUPDATE]}>
         <StixCoreRelationshipCreationFromEntity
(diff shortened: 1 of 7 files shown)


opencti-platform/opencti-front/src/private/components/common/stix_core_relationships/views/EntityStixCoreRelationshipsRelationshipsView.tsx
```diff
@@ -143,7 +143,18 @@ const EntityStixCoreRelationshipsRelationshipsView: FunctionComponent<EntityStix
...
+    filters: [
+      ...predefinedFilters,
+      {
+        key: 'regardingOf',
+        operator: 'eq',
+        mode: 'and',
+        values: [
+          { key: 'id', values: [entityId], operator: 'eq', mode: 'or' },
+          { key: 'relationship_type', values: relationshipTypes, operator: 'eq', mode: 'or' },
+        ] as unknown as string[], // Workaround for typescript waiting for better solution
```
labo-flg: This means we have to define different cases for values depending on the key.
I can have a look if you need help.

Archidoit: 'in regards of' is a filter for entities, not relationships

Archidoit (approved):

opencti-platform/opencti-front/src/private/components/data/ToolBar.jsx
```diff
@@ -1645,9 +1646,7 @@ class ToolBar extends Component {
                                     )}
                                   </span>
                                 )}
-                                <TasksFilterValueContainer
-                                  filters={filters}
-                                ></TasksFilterValueContainer>
+                                <TasksFilterValueContainer filters={filters} entityTypes={entityTypes}></TasksFilterValueContainer>
```
labo-flg:
```suggestion
                                <TasksFilterValueContainer filters={filters} entityTypes={entityTypes} />
```

opencti-platform/opencti-front/src/private/components/data/tasks/TasksList.jsx
```diff
@@ -258,6 +258,7 @@ const TasksList = ({ data }) => {
                         && (isFilterGroupNotEmpty(filters)
                           ? <TasksFilterValueContainer
                               filters={filters}
+                              entityTypes={['Stix-Core-Object', 'stix-core-relationship', 'Notification']}
```
labo-flg, line 261: I see heterogeneous casing. We don' need capital letters ?

Archidoit: stix-core-relationship is not in capital in this case... yes that's heterogeneous :(

Archidoit: we need to refacto all the occurences if we want to homogenize...

labo-flg (approved):
Tested OK

PR #6418, opened 2024-03-22
[frontend/backend] Public dashboards (#4903)

lndrtrbn:
### Proposed changes

* Migration to rename some capability with more comprehensive names
* Migration to add a new capability to be able to share public dashboards
* URI Key of the public dashboard generated based on the name given by the user
* Possibility to disable a public dashboard (cannot be accessed with the public URL)
* Add backend tests

### Related issues

* #4903 

### Checklist

### Further comments


opencti-platform/opencti-front/lang/front/de.json
```diff
@@ -2691,9 +2686,16 @@
   "entity_Tracking-Number": "Tracking-Nummer",
   "entity_Credential": "Berechtigung",
   "Auto collapse submenus in left navigation": "Untermenüs in der linken Navigation automatisch einklappen",
-  "All add or remove actions done on the graph after the last expansion will be lost.": "Alle Hinzufügungs- oder Entfernungsaktionen, die nach der letzten Erweiterung am Diagramm vorgenommen wurden, gehen verloren.",
```
SouadHadjiat: doesn't seem right to loose this translation, maybe a rebase issue ?

opencti-platform/opencti-graphql/src/modules/publicDashboard/publicDashboard-domain.ts
```diff
@@ -193,7 +220,8 @@ export const publicDashboardEditField = async (
   id: string,
   input: EditInput[],
 ) => {
-  const invalidInput = input.some((item: EditInput) => item.key !== 'name' && item.key !== 'uri_key');
+  const validKeys = ['name', 'uri_key', 'enabled'];
+  const invalidInput = input.some((item: EditInput) => !validKeys.includes(item.key));
```
SouadHadjiat: Why do we need this test ? Can't we define attributes that can't be updated in the schema ?

opencti-platform/opencti-front/lang/front/de.json
lndrtrbn: Yes it seems a lot of translations have been removed... I'll check all of them manually

opencti-platform/opencti-graphql/src/modules/publicDashboard/publicDashboard-utils.ts
```diff
@@ -36,16 +40,20 @@ export const getWidgetArguments = async (
     throw UnsupportedError('User not found');
   }
 
-  // To replace User markings by publicDashboard allowed_markings
+  // Determine the marking definitions allowed.
   const allMarkings = await getEntitiesListFromCache<StoreMarkingDefinition>(context, SYSTEM_USER, ENTITY_TYPE_MARKING_DEFINITION);
+  const maxMarkings = await getMaxMarkings(context, SYSTEM_USER);
+  const availableMaxMarkings = computeAvailableMarkings(maxMarkings, allMarkings).map((m) => m.id);
+  const availableDashboardMarkings = computeAvailableMarkings(allowed_markings, allMarkings);
+  const availableMarkings = availableDashboardMarkings.filter((m) => availableMaxMarkings.includes(m.id));
```
SouadHadjiat: I think that all this block starting line 44 could be extracted in a dedicated method. I find it a bit difficult to understand. And part of it is duplicated in `publicDashboard-domain.ts`.

opencti-platform/opencti-graphql/src/modules/publicDashboard/publicDashboard-domain.ts
lndrtrbn: Hum, nothing I am aware of but I'm not an expert in schema. Is it the goal of the upsert boolean in schema?

opencti-platform/opencti-graphql/src/domain/settings.js
```diff
@@ -197,3 +197,11 @@ export const getCriticalAlerts = async (context, user) => {
   // no alert
   return [];
 };
+
+export const getMaxMarkings = async (context, user, settings) => {
```
SouadHadjiat: suggestion to rename to `getDataSharingMaxMarkings` ? to be more explicit that it's the max markings setting for data sharing.

opencti-platform/opencti-graphql/src/migrations/1704893567321-capability-publish-knowledge.js
```diff
@@ -1,16 +0,0 @@
-import { executionContext, SYSTEM_USER } from '../utils/access';
-import { addCapability } from '../domain/grant';
-
-export const up = async (next) => {
-  const context = executionContext('migration');
-  await addCapability(context, SYSTEM_USER, {
```
SouadHadjiat, line 6: I don't think we should remove existing migrations, we can only add newer ones, but we shouldn't delete the existing ones.

opencti-platform/opencti-graphql/src/migrations/1712129378000-investigations-and-dashboards-capabilities.js
```diff
@@ -0,0 +1,30 @@
+import { executionContext, SYSTEM_USER } from '../utils/access';
+import { addCapability } from '../domain/grant';
+import { elLoadById, elReplace } from '../database/engine';
+
+export const up = async (next) => {
+  const context = executionContext('migration');
+  // ------ Add Manage Public Dashboards capability
+  await addCapability(context, SYSTEM_USER, {
+    name: 'EXPLORE_EXUPDATE_PUBLISH',
```
SouadHadjiat: this capability has already been created in existing platforms, so I think we should rename it like the others.

opencti-platform/opencti-graphql/src/modules/publicDashboard/publicDashboard-domain.ts
```diff
@@ -42,16 +41,18 @@ import {
...
 import { isStixCoreObject } from '../../schema/stixCoreObject';
 import { ES_MAX_CONCURRENCY } from '../../database/engine';
+import { getMaxMarkings } from '../../domain/settings';
 
 export const findById = (
   context: AuthContext,
   user: AuthUser,
   id: string,
 ) => {
+  logApp.info('[OPENCTI] Public dashboard - trying to fetch public dashboard with id', { id });
```
SouadHadjiat: I guess these were logs for dev ? or do we want them in prod ?

opencti-platform/opencti-graphql/src/modules/publicDashboard/publicDashboard-domain.ts
SouadHadjiat: I think it should work with `update` attribute

opencti-platform/opencti-graphql/src/modules/publicDashboard/publicDashboard-domain.ts
lndrtrbn: Hum, that's from a suggestion made during the demo : put some logs on accessing public dashboards to facilitate debug if a user ask for support because when they go to a public dashboard URL they are redirected to home

SouadHadjiat: but this is not the public API that is called by external users, right ?

opencti-platform/opencti-graphql/src/modules/publicDashboard/publicDashboard-domain.ts
```diff
@@ -247,6 +275,8 @@ export const publicDashboardDelete = async (context: AuthContext, user: AuthUser
 // heatmap & vertical-bar & line & area
 export const publicStixCoreObjectsMultiTimeSeries = async (context: AuthContext, args: QueryPublicStixCoreObjectsMultiTimeSeriesArgs) => {
   const { user, parameters, dataSelection } = await getWidgetArguments(context, args.uriKey, args.widgetId);
+  context.user = user;
```
SouadHadjiat, line 271: why do we need to put the user in the context ?

lndrtrbn: As the request is made by a non connected user, there is no user in the context. That's why we construct a fake user to be able to call the private APIs of widgets. But passing directly this user in the API calls is not sufficient in the case the data retrieved has resolved fields because we use the context to get the current user at those places. For example the field country of administrative area.

SouadHadjiat: where exactly in the code please ?

lndrtrbn: `administrativeArea-resolver.ts` line 21

That's one example among others

opencti-platform/opencti-graphql/src/modules/publicDashboard/publicDashboard-domain.ts
marieflorescontact: done

opencti-platform/opencti-graphql/src/migrations/1704893567321-capability-publish-knowledge.js
marieflorescontact: this migration  had not been played and has not been removed, it has been merged with another (`investigations-and-dashboards-capabilities` ) in order to group together dashboard migrations.

opencti-platform/opencti-graphql/src/migrations/1712129378000-investigations-and-dashboards-capabilities.js
marieflorescontact: migration had not been played

opencti-platform/opencti-graphql/src/modules/publicDashboard/publicDashboard-utils.ts
lndrtrbn: Can you tell me if you find it easier to read now?

opencti-platform/opencti-graphql/src/modules/publicDashboard/publicDashboard-domain.ts
marieflorescontact: In public api, we call the private api, which need the `context` and, at some point, the `context.user`. As it's a public api, we don't have a user on the context, so we need to add the user retrieved from the `publicDashboard`.

opencti-platform/opencti-graphql/src/modules/publicDashboard/publicDashboard-domain.ts
lndrtrbn: Indeed, the api is findByUriKey. I added a log on both functions, I can remove this one if you think it is not relevant

opencti-platform/opencti-graphql/src/migrations/1704893567321-capability-publish-knowledge.js
SouadHadjiat: not sure to understand how the migration has not been played, since it has been merged in master two months ago and has been part of 6.0 release https://github.com/OpenCTI-Platform/opencti/commit/f266553f69206a4e26c9c3a2e8fdc32373412651#diff-7232600645a940725cb4961bf097b9b14fcc855af0e627b94a3a037e47a5785d

marieflorescontact: we forgot to change the timestamp before merging

SouadHadjiat: Ok I understand now that the migration was merged with an older timestamp than the last one (that has even been released) so there are many platforms that haven't run that migration

SouadHadjiat: If I initialize a plaform on master, then I checkout this branch, I still have the old name for "publish exploration" capability : 
image

opencti-platform/opencti-graphql/src/migrations/1712744229271-investigations-and-dashboards-capabilities.js
```diff
@@ -4,12 +4,19 @@ import { elLoadById, elReplace } from '../database/engine';
...
-    name: 'EXPLORE_EXUPDATE_PUBLISH',
-    description: 'Manage Public Dashboards',
-    attribute_order: 1300
-  });
+  // ------ Add or rename Manage Public Dashboards capability
+  const managePublicDashboardsCapability = await elLoadById(context, SYSTEM_USER, 'capability--40892bfe-13c2-5e3e-96a3-531760950451');
+  if (managePublicDashboardsCapability) {
+    const managePublicDashboardsCapabilityPatch = { description: 'Manage Public Dashboards' };
+    await elReplace(managePublicDashboardsCapability._index, managePublicDashboardsCapability.internal_id, { doc: managePublicDashboardsCapabilityPatch });
+  } else if (!managePublicDashboardsCapability) {
```
SouadHadjiat: this can be simplified by a simple else, without if (!managePublicDashboardsCapability).
Otherwise all good :)

marieflorescontact: done

opencti-platform/opencti-graphql/src/migrations/1704893567321-capability-publish-knowledge.js
marieflorescontact: done

SouadHadjiat:
I guess I need to configure the max markings in the administration settings before sharing a public dashboard ? otherwise I can't select any marking and it's mandatory. Screenshot when I try to share a public dashboard :

image

@lndrtrbn : indeed you have to, we add an info text to explain it to the user, maybe not enough clear

SouadHadjiat:
There is a "Link" message, shouldn't it be "public dashboard" ? 
image

@lndrtrbn : oh yes, have to change it

SouadHadjiat:
The error message is not very clear I think, I tried to create a dashboard with the same ID as an existing one, but the message is about "uri key". Suggestion : "Cannot publish the dashboard, a public dashboard with the same ID already exists" ?
image

@lndrtrbn : yes in fact it's really the URI KEY that already exists, but you are right as we use the term ID in the form we should kept it in the error message, or also use URI KEY in the form

SouadHadjiat:
When I create a dashboard with a list of entities (with filter "authors not empty") and I share this public dashboard with a max marking TLP:AMBER, I don't see the same list in the public dashboard.

Screenshot of original dashboard :
image

Screenshot of public dashboard : 
image

Looks like sorting is not the same.

opencti-platform/opencti-graphql/src/domain/settings.js
```diff
@@ -197,3 +199,19 @@ export const getCriticalAlerts = async (context, user) => {
...
   return [];
 };
+
+// Retrieves max level of markings that can be shared.
+export const getDataSharingMaxMarkings = async (context, user, settings) => {
+  const { platform_data_sharing_max_markings } = settings ?? await getEntityFromCache(context, user, ENTITY_TYPE_SETTINGS);
+  if (!platform_data_sharing_max_markings) {
+    return [];
+  }
+  return internalFindByIds(context, user, platform_data_sharing_max_markings);
```
SouadHadjiat: I suggest to use cache instead of querying markings in DB
```suggestion
  const dataSharingMaxMarkings = platform_data_sharing_max_markings ?? [];
  const allMarkingsMap = await getEntitiesMapFromCache(context, SYSTEM_USER, ENTITY_TYPE_MARKING_DEFINITION);
  return dataSharingMaxMarkings.map((m) => allMarkingsMap.get(m) ?? m);
```

opencti-platform/opencti-graphql/src/modules/publicDashboard/publicDashboard-domain.ts
```diff
@@ -125,6 +126,20 @@ export const addPublicDashboard = async (
     throw FunctionalError('Cannot published empty dashboard');
   }
 
+  const adminAuthorizedMembers = dashboard.authorized_members
+    .filter((member) => member.access_right === 'admin')
+    .map((member) => member.id);
+
+  if (!adminAuthorizedMembers.includes(user.id)) {
```
SouadHadjiat: what about when you are authorized as part of a group ? there are already dedicated functions in access.ts for authorized members. You can use getUserAccessRight : 

```suggestion
const access = getUserAccessRight(user, dashboard);
if (access !== MEMBER_ACCESS_RIGHT_ADMIN) {
     throw ForbiddenAccess('You are not allowed to do this.');
}
```

opencti-platform/opencti-graphql/src/modules/publicDashboard/publicDashboard-utils.ts
```diff
@@ -36,16 +40,19 @@ export const getWidgetArguments = async (
     throw UnsupportedError('User not found');
   }
 
-  // To replace User markings by publicDashboard allowed_markings
+  // Determine the marking definitions allowed.
   const allMarkings = await getEntitiesListFromCache<StoreMarkingDefinition>(context, SYSTEM_USER, ENTITY_TYPE_MARKING_DEFINITION);
+  const availableDataSharingMarkingIds = (await getAvailableDataSharingMarkings(context, SYSTEM_USER)).map((m) => m.id);
+  const availableDashboardMarkings = computeAvailableMarkings(allowed_markings, allMarkings);
```
SouadHadjiat: I don't understand this code, why do we call : 
```
computeAvailableMarkings(allowed_markings, allMarkings);
``` 
then filter with availableDataSharingMarkingIds instead of directly calling computeAvailableMarkings with availableDataSharingMarkings
``` 
const availableDataSharingMarkings = await getAvailableDataSharingMarkings(context, SYSTEM_USER));
computeAvailableMarkings(allowed_markings, availableDataSharingMarkings);

Comment (SouadHadjiat) [reply opencti-platform/opencti-graphql/src/modules/publicDashboard/publicDashboard-utils.ts]: also, when do we check allowed markings of the user ? since we call all these methods with SYSTEM_USER

Comment (lndrtrbn) [reply opencti-platform/opencti-graphql/src/modules/publicDashboard/publicDashboard-utils.ts]: I would love to be able to do your suggestion: `computeAvailableMarkings(allowed_markings, availableDataSharingMarkings);`, I have tried it when writing the algorithm but it does not work because `allowed_markings` can contains markings that are not present in `availableDataSharingMarkings` and this behavior throws an error, that's not what we want so we are force to compute based on all markings.

About the second point: we do not have a user on queries on public dashboard API as it is an API consumable without being connected. The markings we used are the ones already set in the public dashboard at its creation. But maybe we miss a check on the mutation to create a public dashboard that verifies given markings are accessible by the suer creating it indeed..

File: opencti-platform/opencti-graphql/src/modules/publicDashboard/publicDashboard-domain.ts [hunk 13]
```diff
@@ -151,15 +166,23 @@ export const addPublicDashboard = async (
     [{ id: user.id, access_right: 'admin' }, { id: 'ALL', access_right: 'view' }],
     user,
   );
+
+  // check platform data sharing max markings
+  const availableDataSharingMarkingIds = (await getAvailableDataSharingMarkings(context, SYSTEM_USER)).map((m) => m.id);
+  if (input.allowed_markings_ids?.some((id) => !availableDataSharingMarkingIds.includes(id))) {
```
SouadHadjiat, line 169: there is no check of user allowed markings. I could as a user share a public dashboard with a marking > that the ones I have access to (via the API)

marieflorescontact: done

opencti-platform/opencti-graphql/src/migrations/1708523502214-add-platform-data-sharing-max-markings.js
```diff
@@ -0,0 +1,18 @@
+import { executionContext, SYSTEM_USER } from '../utils/access';
+import { getSettings, settingsEditField } from '../domain/settings';
+import { logApp } from '../config/conf';
+
+export const up = async (next) => {
```
SouadHadjiat: this migration file name need to be updated, its timestamp is lower than 1711444304636-create-index-deleted-objects.js which is already present in master

SouadHadjiat (approved):

Issue #3167, opened 2023-04-15
Be able to rollback / cancel expanding an entity in investigation


SamuelHassine:
## Use case

Be able to rollback / cancel expanding an entity in investigation using CTRL+Z or a toolbar button.
First idea is to implement rollback capabilities directly by the front maintaining the previous states in the browser memory

PR #6468, opened 2024-03-25
[frontend] roll back investigation graph after expanding a node (#3167)

ValentinBouzinFiligran:
### Proposed changes

* sessionStorage history created on InvestigationGraph file
* new button added on InvestigationGraphBar
* translation files updated

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/3167

### Checklist

### Further comments


opencti-platform/opencti-front/src/private/components/workspaces/investigations/Dialog/InvestigationRollBackExpandDialog.tsx
```diff
@@ -0,0 +1,52 @@
+import React from 'react';
+import Dialog from '@mui/material/Dialog';
+import DialogTitle from '@mui/material/DialogTitle';
+import DialogContent from '@mui/material/DialogContent';
+import DialogActions from '@mui/material/DialogActions';
+import Button from '@mui/material/Button';
+import { useFormatter } from 'src/components/i18n';
```
labo-flg: this kind of import crashes when using vite build (yarn run dev)

labo-flg:
```suggestion
import { useFormatter } from '../../../../../components/i18n';
```

opencti-platform/opencti-front/src/private/components/workspaces/investigations/InvestigationGraph.jsx
```diff
@@ -1835,6 +1837,13 @@ class InvestigationGraphComponent extends Component {
       this.fetchObjectRelCounts(newElements);
     }
     if (newElementsIds.length > 0) {
+      setStackDataInSessionStorage({ valueToStore: {
+        dateTime: new Date().getTime(),
+        investigatedEntitiesIdsList: newElementsIds,
+      },
+      key: 'preExpansionStateList',
+      stackValue: 10 });
+
```
labo-flg:
```suggestion
      setStackDataInSessionStorage({ 
        valueToStore: {
          dateTime: new Date().getTime(),
          investigatedEntitiesIdsList: newElementsIds,
        },
        key: 'preExpansionStateList',
        stackValue: 10,
      });

```

Did you configure your linter "on save" ?

opencti-platform/opencti-front/src/private/components/workspaces/investigations/InvestigationGraph.jsx
```diff
@@ -2012,9 +2081,17 @@ class InvestigationGraphComponent extends Component {
                   onReset={this.onResetExpandElements.bind(this)}
                 />
               </Dialog>
+
+              <InvestigationRollBackExpandDialog
+                isRollBackPreExpandStateDialogOpen={isRollBackPreExpandStateDialogOpen}
```
labo-flg: I would remain this prop to simple "open" as in the context of the dialog, it's clear enough.
When I'm naming a prop, I just think "in the context of the component"  and not relative to how it will be used by the parent(s)
The state name though can be more precise of course.

opencti-platform/opencti-front/src/private/components/workspaces/investigations/InvestigationGraphBar.jsx
```diff
@@ -865,6 +870,13 @@ class InvestigationGraphBar extends Component {
                           />
                         </>
                       )}
+                      <Tooltip title={t('Restore the state of the graphic before the last expansion')}>
+                        <span>
+                          <IconButton color="primary" onClick={handleOpenRollBackToPreExpansionStateDialog} size="large" disabled={isRollBackToLastPreExpansionStateEnabled}>
```
labo-flg: I would print one prop per line for readability, maybe it's not covered by lint.

opencti-platform/opencti-front/src/utils/sessionStorage/setStackDataInSessionStorage/setStackDataInSessionStorage.test.ts
```diff
@@ -0,0 +1,71 @@
...
+      beforeEach(() => {
+        setStackDataInSessionStorage({
+          valueToStore: { test: 'value1' },
+          key: testKey,
+          stackValue,
+        });
+      });
+
+      describe('If I get datas with my key from session storage', () => {
+        it('contains the my value', () => {
```
labo-flg:
```suggestion
        it('contains my value', () => {
```

opencti-platform/opencti-front/src/utils/sessionStorage/setStackDataInSessionStorage/setStackDataInSessionStorage.ts
```diff
@@ -0,0 +1,17 @@
+type SetDataOnSessionStorageParams = {
+  valueToStore: unknown;
+  key: string;
+  stackValue: number;
+};
+const setStackDataInSessionStorage = ({ valueToStore, key, stackValue }: SetDataOnSessionStorageParams): void => {
```
labo-flg: I think we usually go for inlined parameters and not an object.
I suggest also this order and naming. It's more standard to use key-value wording I think.

```suggestion
const setStackDataInSessionStorage = (key: string, value: unknown, stackLimit: number): void => {
```

opencti-platform/opencti-front/src/utils/sessionStorage/setStackDataInSessionStorage/setStackDataInSessionStorage.ts
```diff
@@ -0,0 +1,17 @@
...
+  const storedStackData = sessionStorage.getItem(key);
+  const currentStoredStackData = storedStackData ? JSON.parse(storedStackData) : [];
+
+  currentStoredStackData.unshift(valueToStore);
+
+  if (currentStoredStackData.length > stackValue) currentStoredStackData.pop();
```
labo-flg: if we call this function with same key but different stackValue this would not work as intended.
We can slice the array instead of just pop I guess.

opencti-platform/opencti-front/lang/back/fr.json
```diff
@@ -626,5 +626,9 @@
   "X_Aliases": "X_Aliases",
   "Samples": "Échantillons",
   "Contains observable": "Contient des observables",
-  "Artifact content": "Contenu de l'artefact"
+  "Artifact content": "Contenu de l'artefact",
+  "Entity ID": "ID de l'entité",
+  "Source connector": "Connecteur source",
+  "Malware analysis operating System": "Système d'exploitation pour l'analyse des logiciels malveillants",
+  "Allowed markings": "Marques autorisées"
```
SouadHadjiat: Why do we need these backend translations ?
also "marques" is not the right translation for marking, it should be "marquages" (same translations present in frontend)

opencti-platform/opencti-front/src/private/components/workspaces/investigations/Dialog/InvestigationRollBackExpandDialog.tsx
```diff
@@ -0,0 +1,52 @@
...
+const InvestigationRollBackExpandDialog = ({ closeDialog, handleRollBackToPreExpansionState, isOpen }: InvestigationRollBackExpandDialogProps) => {
+  const { t_i18n, fldt } = useFormatter();
+
+  const handleSubmit = () => {
+    handleRollBackToPreExpansionState();
+    closeDialog();
+  };
+
+  const getLastRollBackExpandDate = () => {
+    const storedPreExpansion = sessionStorage.getItem('preExpansionStateList');
```
SouadHadjiat: maybe we could have a dedicated file with functions to store & retrieve investigation state in session storage ? it would keep the logic in one place. @labo-flg @ValentinBouzinFiligran

opencti-platform/opencti-front/src/private/components/workspaces/investigations/InvestigationGraphBar.jsx
```diff
@@ -865,6 +871,18 @@ class InvestigationGraphBar extends Component {
                           />
                         </>
                       )}
+                      <Tooltip title={t('Restore the state of the graphic before the last expansion')}>
+                        <span>
+                          <IconButton
+                            color="primary"
+                            disabled={isRollBackToLastPreExpansionStateEnabled}
```
SouadHadjiat: not sure to understand why we disable the restore button if  `isRollBackToLastPreExpansionStateEnabled` is true (or maybe it should be named isRollBackToLastPreExpansionStateDisabled ?)

SouadHadjiat:
I just tested locally : 
- Expand an entity -> I can see the button rollback expand -> rollback works very well
- Expand an entity -> then add another entity -> then I rollback : rollback of expand is all right, but I can still see the entity that I have added after. I guess it might be ok ?
- Expand an entity -> expand a second entity -> rollback : it rollbacks the last expand all right -> rollback again : it rollbacks the first expand which is all right too.

labo-flg:
> Expand an entity -> then add another entity -> then I rollback : rollback of expand is all right, but I can still see the entity that I have added after. I guess it might be ok ?

Rollback shall put your graph into the state before the last expansion, so any entity added after the expand shall be lost. @ValentinBouzinFiligran Am I right ?

Kedae:
> > Expand an entity -> then add another entity -> then I rollback : rollback of expand is all right, but I can still see the entity that I have added after. I guess it might be ok ?
> 
> Rollback shall put your graph into the state before the last expansion, so any entity added after the expand shall be lost. @ValentinBouzinFiligran Am I right ?

Nop, I think this is right. We want to rollback expand only. You need to check with @nino-filigran, but for me this is the expected behavior

nino-filigran:
In theory, yes @labo-flg you are right. What we said is that we come back to the state prior the expand, which means, anything that did happen **after** the rollback would be lost on rollback. 
Hence the documentation that we wrote @ValentinBouzinFiligran & @Kedae: 

"Expanding a graph can add a lot of entities and relations, making it not only difficult to read but sometimes counterproductive since it brings entities and relations that are not useful to your investigations.
To solve this problem, there is a button to undo the last expansion. 
 
**When clicking on this CTA, we will retrieve the state in which your graph was before your expansion. As a result, please note that all modifications made since the last expansion will be lost: in other words, if you have expanded your graph, and then have added some entities in your graph, when clicking on rollback button, the entities that you have added will not be in your graph.**
 
// insert pics here

Rollbacks are limited to 10."

opencti-platform/opencti-front/src/private/components/workspaces/investigations/InvestigationGraph.jsx
```diff
@@ -1803,6 +1806,16 @@ class InvestigationGraphComponent extends Component {
       return;
     }
 
+    const investigatedEntitiesIdsList = this.graphObjects.map(({ id }) => id);
+    setStackDataInSessionStorage(
```
Kedae, line 1808: This seems really deeply linked to Graph, I think it should be better in investigationStorage

Archidoit:
The state after a restore doesn't always correspond to the state before the expansion

Scénario 1:
Investigation with an entity A.
Expand an entity A, add an entity B.
Click on 'restore'.
The expansion of A and the added entity B are removed -> state before the expansion.

Scénario 2:
Investigation with an entity A and an entity B.
Expand an entity A, remove an other entity B.
Click on 'restore'.
The expansion is removed but the entity B is not added -> not really the state before the expansion.

Archidoit:
Everything works well :)

Just 2 missing translations in investigation graphs that can be added in this PR : 
image

image

opencti-platform/opencti-front/src/private/components/workspaces/investigations/utils/setStackDataInSessionStorage/setStackDataInSessionStorage.test.ts
```diff
@@ -0,0 +1,63 @@
...
+      });
+
+      describe('If I get datas with my key from session storage', () => {
+        it('contains my value', () => {
+          const storedData = sessionStorage.getItem('testKey');
+          expect(storedData).to.not.equal(null);
+
+          if (storedData) {
+            const parsedStoredData = JSON.parse(storedData);
+            expect(parsedStoredData[0].test).to.equal('value1');
```
Archidoit: what about using 'toEqual' to uniformize with the other tests?

opencti-platform/opencti-front/src/private/components/workspaces/investigations/utils/setStackDataInSessionStorage/setStackDataInSessionStorage.test.ts
```diff
@@ -0,0 +1,63 @@
...
+      });
+
+      describe('If I get datas with my key from session storage', () => {
+        it('does not contains my first value', () => {
+          const storedData = sessionStorage.getItem('testKey');
+          expect(storedData).to.not.equal(null);
+
+          if (storedData) {
+            const parsedStoredData = JSON.parse(storedData);
+            expect(parsedStoredData[2]).to.equal(undefined);
```
Archidoit: you can use 'toBeUndefined()'

opencti-platform/opencti-front/src/private/components/workspaces/investigations/utils/setStackDataInSessionStorage/setStackDataInSessionStorage.test.ts
```diff
@@ -0,0 +1,63 @@
...
+
+    describe('When I set a data in session storage', () => {
+      beforeEach(() => {
+        setStackDataInSessionStorage(testKey, { test: 'value1' }, stackValue);
+      });
+
+      describe('If I get datas with my key from session storage', () => {
+        it('contains my value', () => {
+          const storedData = sessionStorage.getItem('testKey');
+          expect(storedData).not.toEqual(null);
```
Archidoit:
```suggestion
          expect(storedData).not.toBeNull();
```

opencti-platform/opencti-front/src/private/components/workspaces/investigations/utils/setStackDataInSessionStorage/setStackDataInSessionStorage.test.ts
```diff
@@ -0,0 +1,63 @@
...
+
+          if (storedData) {
+            const parsedStoredData = JSON.parse(storedData);
+            expect(parsedStoredData[2]).toBeUndefined();
+          }
+        });
+
+        it('contains my two last values', () => {
+          const storedData = sessionStorage.getItem('testKey');
+          expect(storedData).to.not.equal(null);
```
Archidoit: same

opencti-platform/opencti-front/src/private/components/workspaces/investigations/utils/setStackDataInSessionStorage/setStackDataInSessionStorage.test.ts
```diff
@@ -0,0 +1,63 @@
...
+      beforeEach(() => {
+        Array.from(Array(3).keys()).forEach((_, i) => {
+          setStackDataInSessionStorage(testKey, { test: `value${i}` }, stackValue);
+        });
+      });
+
+      describe('If I get datas with my key from session storage', () => {
+        it('does not contains my first value', () => {
+          const storedData = sessionStorage.getItem('testKey');
+          expect(storedData).to.not.equal(null);
```
Archidoit: same

Archidoit (approved):

PR #6470, opened 2024-03-25
[frontend] add widget PolarArea for relationships

lndrtrbn:
### Proposed changes

* Add the widget PolarArea as possible widget in dashboard for relationships

### Related issues

* Wish from Jermain

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/utils/Charts.js
+++ b/opencti-platform/opencti-front/src/utils/Charts.js
@@ -626,6 +626,12 @@ export const radarChartOptions = (
   },
 });
 
+/**
+ * @param {Theme} theme
+ * @param {string[]} labels
+ * @param {function} formatter
+ * @param {string} legendPosition
+ */
 export const polarAreaChartOptions = (
   theme,
   labels,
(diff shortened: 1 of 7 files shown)


opencti-platform/opencti-front/src/utils/Charts.js
```diff
@@ -626,6 +626,12 @@ export const radarChartOptions = (
   },
 });
 
+/**
+ * @param {Theme} theme
+ * @param {string[]} labels
+ * @param {function} formatter
+ * @param {string} legendPosition
+ */
```
SarahBocognano, line 634: Just curious about that, can you explain to me what is the utility of it ? :)

SarahBocognano (approved):

opencti-platform/opencti-front/src/utils/Charts.js
lndrtrbn: It allows IDE like Intellij to be able to check the types of the parameters even if it is a JS file when you use this function in a TS file

PR #6473, opened 2024-03-26
[backend] Improve config restrictions / visibility for batch rel loading

richard-julien:
Different improvements out of the milestone.
- Better restrictions for logsWorkerConfig api
- Adapt restrictions for work management + wrap security in the UI
- Remove unneeded node sharing in rabbitMQ OverviewMetrics

### DIFF
--- a/opencti-platform/opencti-graphql/src/domain/stixCoreObject.js
+++ b/opencti-platform/opencti-graphql/src/domain/stixCoreObject.js
@@ -47,6 +47,7 @@ import { INSTANCE_REGARDING_OF, specialFilterKeysWhoseValueToResolve } from '../
 import { schemaRelationsRefDefinition } from '../schema/schema-relationsRef';
 import { ENTITY_TYPE_CONTAINER_GROUPING } from '../modules/grouping/grouping-types';
 import { getEntitiesMapFromCache } from '../database/cache';
+import { isUserCanAccessStoreElement, SYSTEM_USER } from '../utils/access';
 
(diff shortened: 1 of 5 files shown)


opencti-platform/opencti-graphql/src/domain/stixCoreObject.js
```diff
@@ -80,31 +81,46 @@ export const findById = async (context, user, stixCoreObjectId) => {
...
           throw UnsupportedError('Invalid loading of batched elements', { ids: relId });
         }
-        return resolve;
-      });
+        // If user have correct access right, return the element
+        if (await isUserCanAccessStoreElement(context, user, resolve)) {
+          return resolve;
+        }
+        // If access is not possible, return a restricted entity
+        return { id: resolve.internal_id, name: 'Restricted', entity_type: resolve.entity_type };
```
labo-flg, line 101: Would it be possible for a given entity_type to NOT have a `name` attribute, leading to an impossible object schema-wise ? an opinion for instance.

labo-flg: I have the same issue in my PR https://github.com/OpenCTI-Platform/opencti/pull/6323

labo-flg: As discussed, if you're confident it concerns only entities with names, let's go for it as it would not be disruptive for the frontend side.

labo-flg (approved):

PR #6491, opened 2024-03-26
[frontend] fix remove relation ref from Grouping Knowledge graph (#6482)

### LINKED ISSUE #6482: Nested refs are not properly removed in groupings knowledge graph
## Description

When trying to delete a ref from a groupings knowledge graph, the ref is deleted at first, but when switching to an other tab and switching back, the ref is still visible

## Reproducible Steps

1. Create a new grouping
2. Go to the knowledge graph, add a malware and file observable
3. Create a sample ref between the malware and the file
4. Delete the sample ref
5. Switch to an other tab and come back

## Expected Output

The ref being removed from the groupings graph

## Actual Output

The ref still visible in the graph

## Additional information

No graphql query is sent when removing the ref

### LINKED ISSUE #6484: Removing a relation in a grouping knowledge graph removes it in all containers
## Description

When removing a relation from a groupings graph, the relation is deleted even if it is contained in an other container.

## Reproducible Steps

1. Create a new report
2. Go to the knowledge graph, add two entities and create a relationship between them
3. Create a new grouping
4. Go to the knowledge graph, add the two same entities and add the same relationship between them
5. Remove this relationship

## Expected Output

The relationship still being there in the report

## Actual Output

The relationship is deleted

## Additional information

We are forced to do a cascade remove in groupings, but in reports we are prompted to choose between only removing the relation in this container or if we want to do a cascade remove. Maybe we also want to add this option in groupings and other containers?

### LINKED ISSUE #6486: Cascade remove in report knowledge graph is not working as expected
Description

When removing a relation from a report graph, when choosing a cascade remove, the relation is deleted even if it is contained in an other container.

Reproducible Steps

1. Create a new grouping
2. Go to the knowledge graph, add two entities and create a relationship between them
3. Create a new report
4. Go to the knowledge graph, add the two same entities and add the same relationship between them
5. Remove this relationship choosing a cascade remove

Expected Output

The relationship still being there in the grouping

Actual Output

The relationship is entirely deleted

Archidoit:
issues:
-  https://github.com/OpenCTI-Platform/opencti/issues/6482
- https://github.com/OpenCTI-Platform/opencti/issues/6486
- https://github.com/OpenCTI-Platform/opencti/issues/6484


opencti-platform/opencti-front/src/private/components/analyses/groupings/GroupingKnowledgeGraph.jsx
```diff
@@ -1033,14 +1092,33 @@ class GroupingKnowledgeGraphComponent extends Component {
...
-      });
+      fetchQuery(groupingKnowledgeGraphCheckObjectQuery, {
+        id: n.id,
+      })
+        .toPromise()
+        .then(async (data) => {
+          if (
+            deleteObject
+                && !data.stixObjectOrStixRelationship.is_inferred
+                && data.stixObjectOrStixRelationship.groupings.edges.length === 1
```
labo-flg: I see we still have some badly configured linter.. .4 spaces instead of 2 here.

Archidoit: fixed

opencti-platform/opencti-front/src/private/components/analyses/groupings/GroupingKnowledgeGraph.jsx
```diff
@@ -93,6 +94,63 @@ const groupingKnowledgeGraphCheckRelationQuery = graphql`
...
+const groupingKnowledgeGraphCheckObjectQuery = graphql`
+  query GroupingKnowledgeGraphCheckObjectQuery($id: String!) {
+    stixObjectOrStixRelationship(id: $id) {
+      ... on BasicObject {
+        id
+      }
+      ... on StixCoreObject {
+        is_inferred
+        parent_types
+        reports {
```
JeremyCloarec: We should get the groupings and not the reports

labo-flg (approved):

opencti-platform/opencti-front/src/private/components/analyses/groupings/GroupingKnowledgeGraph.jsx
```diff
@@ -1033,14 +1092,33 @@ class GroupingKnowledgeGraphComponent extends Component {
...
-      });
+      fetchQuery(groupingKnowledgeGraphCheckObjectQuery, {
+        id: n.id,
+      })
+        .toPromise()
+        .then(async (data) => {
+          if (
+            deleteObject
+            && !data.stixObjectOrStixRelationship.is_inferred
+            && data.stixObjectOrStixRelationship.groupings.edges.length === 1
```
JeremyCloarec: I saw that the behavior was the same on all containers, but is it actually the intended behavior? Shouldn't we check that there is only 1 container for the selected entity, and not only check for 1 grouping? Right now, if a relation is both in a report and in a grouping, when I cascade delete it in a grouping, it is fully deleted even though there was an other container for it (the report)

opencti-platform/opencti-front/src/private/components/analyses/groupings/GroupingKnowledgeGraph.jsx
labo-flg: good catch

labo-flg (commented):
Agreed with Jeremy, we shall check in all other types of containers.

opencti-platform/opencti-front/src/private/components/analyses/groupings/GroupingKnowledgeGraph.jsx
Archidoit: fixed

opencti-platform/opencti-front/src/private/components/analyses/groupings/GroupingKnowledgeGraph.jsx
Archidoit: done

JeremyCloarec (approved):

JeremyCloarec:
Tested OK, great job! Love the refactoring into KnowledgeGraphQuery!
Looks like the build is only failing because of some remaining console logs in GroupingKnowledgeGraph

Issue #3389, opened 2023-05-24
Global problem about displayed nested / refs relationships everywhere


SamuelHassine:
## Use case

Right now, only a few types of entities have a display widget to display nested entities:

* Observables

![image](https://github.com/OpenCTI-Platform/opencti/assets/1334279/2206a38c-3a5e-401a-bda2-5e54adfc938d)

* Malware

![image](https://github.com/OpenCTI-Platform/opencti/assets/1334279/e4464ef9-0e6d-4af5-9610-34ef2fe7c8ed)

This is because, in most cases, having nested ref objects is not supported by the type of entity.

However, in OpenCTI, there is a `linked-to_ref` attribute available for all entity types. Therefore, it is possible to create this in report, here an example for a relationship:

![image](https://github.com/OpenCTI-Platform/opencti/assets/1334279/71b532ea-929b-4299-baaf-8eabc0327217)

But this is possible basically on everything in the graph. Taking the example of this previous relationship, the "linked-to_ref" is not displayed anywhere in the relationship overview:

![image](https://github.com/OpenCTI-Platform/opencti/assets/1334279/1feb0b10-a8ed-4760-a6ba-c470b1976541)

This leads to an inconsistent behavior between the graph where it is possible to create at least a linked-to_ref attribute on anything, including relationships, but then the platform is not displaying them in overviews.

We need to think about displaying nested / ref attributes everywhere in the platform, as we are supporting this since v1.

SamuelHassine:
Nested relationships display in relationships.

SamuelHassine:
To be discussed.

nino-filigran:
After a discussion, decision to remove the capbility to created nested relationships "link to (ref)" (cc @Jipegien and @SamuelHassine ). This is what will be implemented.

PR #6513, opened 2024-03-29
[backend] Linked-to migration to related-to chunk1

JeremyCloarec:
### Proposed changes

* migrate all linked-to refs to related-to rels
* remove linked-to from the codebase

Client changes: https://github.com/OpenCTI-Platform/client-python/pull/610
Connector changes: https://github.com/OpenCTI-Platform/connectors/pull/2014

### Related issues

* #3389 

### Checklist

### Further comments


opencti-platform/opencti-graphql/src/migrations/1712225747215-convert_linked-to_related-to.js
```diff
@@ -0,0 +1,128 @@
...
+export const up = async (next) => {
+  const context = executionContext('migration');
+  logApp.info('[MIGRATION] Update all linked-to refs to related-to rels');
+
+  const linkedToType = 'x_opencti_linked-to';
+  const relLinkedTo = `rel_${linkedToType}.internal_id`;
+  const relRelatedTo = `rel_${RELATION_RELATED_TO}.internal_id`;
+
+  // We get all current related-to relations: will be used to check for duplicates in linked-to to related-to migration
+  const relatedToRelations = await elList(context, SYSTEM_USER, READ_RELATIONSHIPS_INDICES, { types: [RELATION_RELATED_TO] });
```
SouadHadjiat: This might list a lot of relationships, no ? compared to linked-to refs, I wonder if we couldn't list linked-to relationships first then search for duplicates in related_to, and exclude these duplicates ? And if we don't find any linked-to, then we don't even have to continue the migration.

opencti-platform/opencti-graphql/src/migrations/1712225747215-convert_linked-to_related-to.js
```diff
@@ -0,0 +1,128 @@
...
+    ctx._source.entity_type = params.relToType;
+    ctx._source.relationship_type = params.relToType;
+    ctx._source.parent_types = params.relToParentTypes;
+    ctx._source.standard_id = ctx._source.standard_id.replace("relationship-meta--", "relationship--");
+    for(connection in ctx._source.connections) {
+      if(connection.role == params.linkedToType + '_from' && connection.containsKey('internal_id')) {fromId = connection.internal_id}
+      if(connection.role == params.linkedToType + '_to' && connection.containsKey('internal_id')) {toId = connection.internal_id}
+      connection.role = connection.role.replace(params.linkedToType, params.relToType);
+    }
+    if(params.relatedToRelationsFromToArray.contains(fromId+toId)) {ctx.op = 'noop'}
```
SouadHadjiat: what does it mean `ct.op = 'noop'` ?

JeremyCloarec: it's used to skip the reindex operation for this given document. In the [documentation](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-reindex.html), in the "Modify documents during reindexing" section, there is also a ctx.op = 'delete' that is mentioned, but I'm not sure what the difference between the 2 are.

opencti-platform/opencti-graphql/src/migrations/1712225747215-convert_linked-to_related-to.js
```diff
@@ -0,0 +1,128 @@
...
+      params: { relLinkedTo, relRelatedTo }
+    },
+    query: {
+      exists: {
+        field: relLinkedTo
+      }
+    },
+  };
+
+  await elUpdateByQueryForMigration('[MIGRATION] Updating entities with rel_linked-to to rel_related-to', [READ_DATA_INDICES], updateDenormalizedLinkedToQuery);
```
SouadHadjiat: READ_DATA_INDICES includes all our indices with internal objects and relationships, but here I think we only need to read indices with stix objects right ? can we have linked-to refs between relationships ?

opencti-platform/opencti-graphql/src/migrations/1712225747215-convert_linked-to_related-to.js
JeremyCloarec: That's true, I think I'll add a listing of all linked-to to check if there are any in the first place, and skip the migration if that is not the case.
About listing the related-to, that's true that it could be a lot, but I'm not sure if listing them all is that time costly? When querying stixRelationships, even on large instances, it doesn't seem to take that much time

opencti-platform/opencti-graphql/src/migrations/1712225747215-convert_linked-to_related-to.js
JeremyCloarec: we can't right now, it's only defined for stix objects you're right. I wasn't sure if it was possible in the past though, since having it between relationships could make sense. Is it too costly performance-wise to keep READ_DATA_INDICES just in case?

opencti-platform/opencti-graphql/src/migrations/1712225747215-convert_linked-to_related-to.js
SouadHadjiat: here we list them all with elList (not elPaginate), we don't stop after the first pagination, so it can load a lot of data.

JeremyCloarec: you're right, I didn't keep in mind that it was only the first pagination. I'll change it to what you suggested

opencti-platform/opencti-graphql/src/migrations/1713253755013-convert_linked-to_related-to.js
```diff
@@ -0,0 +1,195 @@
...
+  READ_DATA_INDICES,
+  INDEX_STIX_META_RELATIONSHIPS,
+  READ_INDEX_STIX_META_RELATIONSHIPS,
+  READ_RELATIONSHIPS_INDICES,
+  READ_INDEX_STIX_CORE_RELATIONSHIPS
+} from '../database/utils';
+import { RELATION_RELATED_TO } from '../schema/stixCoreRelationship';
+import { executionContext, SYSTEM_USER } from '../utils/access';
+
+export const up = async (next) => {
```
SouadHadjiat: please rebase master and rename the migration to update the timestamp, there have been other migrations merged in the meantime.

SouadHadjiat (approved):
tested locally and with @JeremyCloarec

PR #6519, opened 2024-03-29
[frontend] possibility to select attributes in entity widgets (#6515)

### LINKED ISSUE #6515: Donut widget on Entities cannot be configured properly: field is free text
Seen on latest testing environment

When creating a widget Donut > Entities > no filter, the final configuration panel shall allow us to select a filed in a list and not input free text.

image

By default it's `entity_type` which works fine, but otherwise it's unusable. For instance if you'd like the marking you need field `object-marking.internal_id`

Archidoit:
issue: https://github.com/OpenCTI-Platform/opencti/issues/6515

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/workspaces/dashboards/WidgetConfig.jsx
+++ b/opencti-platform/opencti-front/src/private/components/workspaces/dashboards/WidgetConfig.jsx
@@ -1106,14 +1106,13 @@ const WidgetConfig = ({ workspace, widget, onComplete, closeMenu }) => {
                               if (
                                 attributesValues.filter((n) => n === 'hashes').length > 0
                               ) {
-                                attributesValues = R.sort([
+                                attributesValues = [
                                   ...attributesValues,
                                   'hashes.MD5',
                                   'hashes.SHA-1',
                                   'hashes.SHA-256',
                                   'hashes.SHA-512',
-                                ]
-                                  .filter((n) => n !== 'hashes'));
+                                ].filter((n) => n !== 'hashes').sort();
                               }
                               return (
                                 <Select
--- a/opencti-platform/opencti-front/src/private/components/workspaces/dashboards/WidgetConfig.jsx
+++ b/opencti-platform/opencti-front/src/private/components/workspaces/dashboards/WidgetConfig.jsx
@@ -897,21 +897,21 @@ const WidgetConfig = ({ workspace, widget, onComplete, closeMenu }) => {
               return (
                 <div key={i}>
                   {(getCurrentCategory() === 'distribution'
-                                        || getCurrentCategory() === 'list') && (
-                                        <TextField
-                                          label={t_i18n('Number of results')}
-                                          fullWidth={true}
-                                          type="number"
-                                          value={dataSelection[i].number ?? 10}
-                                          onChange={(event) => handleChangeDataValidationParameter(
-                                            i,
-                                            'number',
-                                            event.target.value,
-                                            true,
-                                          )
-                                            }
-                                          style={{ marginTop: 20 }}
-                                        />
+                    || getCurrentCategory() === 'list') && (
+                    <TextField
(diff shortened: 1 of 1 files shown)


opencti-platform/opencti-front/src/private/components/workspaces/dashboards/WidgetConfig.jsx
```diff
@@ -1080,116 +1078,128 @@ const WidgetConfig = ({ workspace, widget, onComplete, closeMenu }) => {
...
+                          }
+                        >
+                          {[
+                            { value: 'entity_type' },
+                            { value: 'created-by.internal_id' },
+                            { value: 'object-label.internal_id' },
+                            { value: 'object-assignee.internal_id' },
+                            { value: 'object-marking.internal_id' },
+                            { value: 'kill-chain-phase.internal_id' },
+                            { value: 'x_opencti_workflow_id' },
```
lndrtrbn: I suppose it is the list of common attributes no matter the entity type?

A detail, you could have used a simple array of string instead of using objects 😁

opencti-platform/opencti-front/src/private/components/workspaces/dashboards/WidgetConfig.jsx
```diff
@@ -1080,116 +1078,128 @@ const WidgetConfig = ({ workspace, widget, onComplete, closeMenu }) => {
...
-                                                  placeholder={t_i18n('Series attribute')}
-                                                  onChange={(event) => handleChangeDataValidationParameter(
-                                                    i,
-                                                    'attribute',
-                                                    event.target.value,
-                                                  )
-                                                        }
-                                                />
+                      && getCurrentSelectedEntityTypes(i).length === 0 && (
+                        <Select
```
lndrtrbn: We don't have a label on this field anymore

opencti-platform/opencti-front/src/private/components/workspaces/dashboards/WidgetConfig.jsx
Archidoit: Yes it is :)

Archidoit: Indeed, I just kept the same format as used for the other widget types

Archidoit: going to fix it :)

opencti-platform/opencti-front/src/private/components/workspaces/dashboards/WidgetConfig.jsx
Archidoit: fixed

opencti-platform/opencti-front/src/private/components/workspaces/dashboards/WidgetConfig.jsx
```diff
@@ -1076,116 +1076,123 @@ const WidgetConfig = ({ workspace, widget, onComplete, closeMenu }) => {
...
+                                attributesValues.filter((n) => n === 'hashes').length > 0
+                              ) {
+                                attributesValues = R.sort([
+                                  ...attributesValues,
+                                  'hashes.MD5',
+                                  'hashes.SHA-1',
+                                  'hashes.SHA-256',
+                                  'hashes.SHA-512',
+                                ]
+                                  .filter((n) => n !== 'hashes'));
```
lndrtrbn:
```suggestion
                                attributesValues = [
                                  ...attributesValues,
                                  'hashes.MD5',
                                  'hashes.SHA-1',
                                  'hashes.SHA-256',
                                  'hashes.SHA-512',
                                ].filter((n) => n !== 'hashes').sort();
```

It was throwing an error (try by selecting Artifact as entity type). Works well by completely remove ramda.

lndrtrbn (approved):

Issue #6407, opened 2024-03-21
"Select all" not available for relationship creation


Lhorus6:
## Description

When I try to create relationships with an entity from the knowledge tab, I can't use the "select all" button, which is grayed out (see screenshot).

## Environment

OCTI 6.0.7

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Got to a threat actor
2. Go to knowledge tab > incident
3. Click on "+" to create relation 
4. Try to select all -> you can't

## Screenshots

![Screenshot 2024-03-21 142822](https://github.com/OpenCTI-Platform/opencti/assets/70956706/571d31f6-8391-4e03-ab3d-646c1fcdf09b)

nino-filigran:
Not only for incidents, but for any kind of relations.

nino-filigran:
@Lhorus6 I did not know this, but this is done by default (we checked internally with some devs): the reason being that if you would link a large amount of entities with a single one might cause issues. 
As a result, we'll keep this CTA disabled for now, but allow MAJ+select to perform bulk selection. By selecting the correct filters, users will still be able to perform a mutli select more quickly and painlessly.

PR #6574, opened 2024-04-04
[frontend] Fix "Select all" not available for relationship creation (#6407)

SarahBocognano:
### Proposed changes

* As said in the issue the disabled "select all" was there on purpose. To avoid the pain of entities by entities selection, we just add the possiblity to MAJ+select to perform bulk selection for both StixCoreRelationships & StixNestedRef. By selecting the correct filters, users will still be able to perform a mutli select more quickly and painlessly.

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/6407


marieflorescontact (approved):
locally tested ok 👌

PR #6590, opened 2024-04-05
[frontend] CI Verify translation automatically

jpkha:
Add a verification step for translation in Drone


SouadHadjiat:
Do you think it would be possible to sort the translation files during keys generation and automatic translation ? so that the keys are added by alphabetical order, and not at the end of the file. This would avoid all our conflicts where all the keys are added automatically at the end of files. And also sorting will help organize the keys.

jpkha:
In case there is missing translation :
image

SouadHadjiat (approved):

PR #6628, opened 2024-04-08
Fix the redirection to "files" when exporting in observed data and notes

CelineSebe:
### Proposed changes

* Add the 'redirectToContent' props to the roots of notes and observed data

## Reproducible Steps

In Events/observed data and Events/Notes:
1. Click on the quick button export, at the top right
2. Verify that the page displays correctly with 'files?' in the URL and not 'content?'

## Actual Output
button export with content in path

### Checklist

### Further comments

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/analyses/notes/Root.jsx
+++ b/opencti-platform/opencti-front/src/private/components/analyses/notes/Root.jsx
@@ -106,6 +106,7 @@ class RootNote extends Component {
                       <ContainerHeader
                         container={props.note}
                         PopoverComponent={<NotePopover note={note} />}
+                        redirectToContent={false}
                       />
                     </CollaborativeSecurity>
                     <Box
--- a/opencti-platform/opencti-front/src/private/components/analyses/notes/Root.jsx
+++ b/opencti-platform/opencti-front/src/private/components/analyses/notes/Root.jsx
@@ -118,11 +118,7 @@ class RootNote extends Component {
                     >
                       <Tabs
                         value={
-                          location.pathname.includes(
-                            `/dashboard/analyses/notes/${note.id}/knowledge`,
-                          )
-                            ? `/dashboard/analyses/notes/${note.id}/knowledge`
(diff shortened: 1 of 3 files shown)


opencti-platform/opencti-front/src/private/components/analyses/notes/Root.jsx
```diff
@@ -117,11 +118,7 @@ class RootNote extends Component {
                     >
                       <Tabs
                         value={
-                          location.pathname.includes(
-                            `/dashboard/analyses/notes/${note.id}/knowledge`,
-                          )
-                            ? `/dashboard/analyses/notes/${note.id}/knowledge`
-                            : location.pathname
+                          location.pathname
```
frapuks, line 121: If you change this part, I think you should look for all the places where we have similar code, and modify it too.

opencti-platform/opencti-front/src/private/components/common/containers/ContainerHeader.jsx
```diff
@@ -898,7 +899,7 @@ const ContainerHeader = (props) => {
               <StixCoreObjectFileExport
                 id={container.id}
                 type={container.entity_type}
-                redirectToContent={true}
+                redirectToContent={!!redirectToContent}
```
frapuks, line 902: If it's a boolean before, and after, need to use `!!` ?

CelineSebe: There is uncertainty about the value we want to pass here, as it will not always be "true". For example, in the case of 'note' and 'observed_data', we want the value of 'redirectToContent' to remain "false".

opencti-platform/opencti-front/src/private/components/analyses/notes/Root.jsx
CelineSebe: Only those who lack knowledge?

frapuks (approved):
test ok

PR #6640, opened 2024-04-10
[backend] Improve objects resolutions event loop usage for large report (#6638)

### LINKED ISSUE #6638: Improve objects resolutions event loop usage for large report (60K+ object_refs)
## Use case

Improve objects resolutions event loop usage for large report (60K+ object_refs)

## Current Workaround

No

## Proposed Solution

- Adapt code to prevent event loop blocking in large data preparation resulting from elements resolution.
- Prevent sending object_refs in the pub/sub system as it can be too large and not use for now.

richard-julien:
See #6638

### DIFF
--- a/opencti-platform/opencti-graphql/src/database/middleware.js
+++ b/opencti-platform/opencti-graphql/src/database/middleware.js
@@ -101,6 +101,7 @@ import {
   ID_INTERNAL,
   ID_STANDARD,
   IDS_STIX,
+  INPUT_CREATED_BY,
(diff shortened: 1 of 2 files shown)


ckane:
I have deployed this fix to my instance as well. Initial testing looking good.

ckane:
This fix cleared the worker backlog. I am re-enabling the DigitalSide.it `misp-feed` connector instance now to see how that works.

ckane:
Looks like it's fixed, thanks @richard-julien!

PR #6672, opened 2024-04-11
Fix playbook Container Wrapper producing bundles that cannot be ingested  (#6302)

### LINKED ISSUE #6302: [Playbooks] Error when creating a container
## Description

When creating a playbook to automate the creation of containers (such as Case-Incident), the process is displayed as successful in the platform, but is in fact unsuccessful (error visible in Grafana).

## Environment

1. OCTI testing (6.0.5)

## Reproducible Steps

1. Creating a playbook,
2. Adding a step: "Container wrapper",
3. Choose any type of container (Case-Incident, Groupings, etc.),
4. Start the playbook.

## Expected Output

The playbook should create a container with the items listened to in it.

## Actual Output

The playbook displays that the steps have been successfully completed. In reality, the container has not been created and the error is visible in Grafana (see below).

## Additional information

Here's an [example of a playbook](https://testing.octi.staging.filigran.io/dashboard/data/processing/automation/6e26633d-99c5-48ae-9f59-b881271127e3) in the testing platform.

Error displayed in Grafana: 

Traceback (most recent call last):
  File "/opt/opencti-worker/worker.py", line 268, in data_handler
    self.api.stix2.import_bundle_from_json(
  File "/usr/local/lib/python3.12/site-packages/pycti/utils/opencti_stix2.py", line 215, in import_bundle_from_json
    return self.import_bundle(
           ^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/pycti/utils/opencti_stix2.py", line 2453, in import_bundle
    self.import_object(item, update, types)
  File "/usr/local/lib/python3.12/site-packages/pycti/utils/opencti_stix2.py", line 920, in import_object
    stix_object_results = do_import(
                          ^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/pycti/entities/opencti_case_incident.py", line 865, in import_from_stix2
    return self.create(
           ^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/pycti/entities/opencti_case_incident.py", line 706, in create
    result = self.opencti.query(
             ^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/pycti/api/opencti_api_client.py", line 351, in query
    raise ValueError(
ValueError: {'name': 'VALIDATION_ERROR', 'message': 'Validation error'}

labo-flg:
See [this comment](https://github.com/OpenCTI-Platform/opencti/issues/6302#issuecomment-2049843842).

### Proposed changes
We are limiting the possibilities offered by the Container Wrapper component in the playbooks, so that it actually works at ingestion time.

Some of the possible containers are requiring more fields to be created correctly, such as the `context` for a grouping. We are not inferring these fields automatically as it would make no sense.

A proper solution would be to allow configuring such fields in the playbooks, depending on the type of container used in the Container Wrapper. This is not an option we want top open right now.

### Related issues
* Closes #6302 

### Checklist


opencti-platform/opencti-graphql/src/modules/playbook/playbook-components.ts
```diff
@@ -340,11 +347,36 @@ const PLAYBOOK_CONTAINER_WRAPPER_COMPONENT: PlaybookComponent<ContainerWrapperCo
...
+      // if (container_type === ENTITY_TYPE_CONTAINER_REPORT) -- name is enough
+      if (container_type === ENTITY_TYPE_CONTAINER_GROUPING) {
+        containerData.context = 'unspecified';
+      }
+      // if (container_type === ENTITY_TYPE_CONTAINER_CASE_INCIDENT) -- name is enough
+      // if (container_type === ENTITY_TYPE_CONTAINER_CASE_RFI) -- name is enough
+      // if (container_type === ENTITY_TYPE_CONTAINER_CASE_RFT) -- name is enough
+      // if (container_type === ENTITY_TYPE_CONTAINER_FEEDBACK) -- name is enough
+      // if (container_type === ENTITY_TYPE_CONTAINER_TASK) -- name is enough
+      // if (container_type === ENTITY_TYPE_CONTAINER_CASE) -- stix object can be created, but cannot be ingested by opencti (abstract)
```
labo-flg: I think we shall prevent users from selecting this type of container in the Container Wrapper.
@richard-julien 'Case' are not part of stix v2, am I right ? Should it be made Abstract in our model ? This cannot be created in opencti actually.

opencti-platform/opencti-graphql/src/modules/playbook/playbook-components.ts
```diff
@@ -340,11 +347,36 @@ const PLAYBOOK_CONTAINER_WRAPPER_COMPONENT: PlaybookComponent<ContainerWrapperCo
...
+        containerData.number_observed = 1;
+      }
+      if (container_type === ENTITY_TYPE_CONTAINER_OPINION) {
+        containerData.explanation = `Generated Opinion explanation from playbook at ${created}`;
+        containerData.opinion = 'neutral';
+      }
+      // if (container_type === ENTITY_TYPE_CONTAINER_REPORT) -- name is enough
+      if (container_type === ENTITY_TYPE_CONTAINER_GROUPING) {
+        containerData.context = 'unspecified';
+      }
```
labo-flg: These are arbitrary values for mandatory fields, required if we want to ingest the container in a "Send to ingester step".
I'm not really sure it's a good idea. We could also limit to containers where it makes sense to automatize creation in a playbook.

richard-julien: Need to have a sync point about this. Thanks

opencti-platform/opencti-front/src/private/components/data/playbooks/Playbook.jsx
```diff
@@ -89,6 +92,15 @@ const PlaybookComponent = ({ playbook, playbookComponents }) => {
...
     <>
+      <Breadcrumbs
+        variant="list"
+        elements={[
+          { label: t_i18n('Data') },
+          { label: t_i18n('Processing') },
+          { label: t_i18n('Automation'), link: '/dashboard/data/processing/automation' },
+          { label: playbook.name, current: true },
+        ]}
+      />
```
labo-flg, line 103: Fix missing breadcrumbs in the playbook page, to go back to the list of playbooks

opencti-platform/opencti-front/src/private/components/data/playbooks/PlaybookAddComponents.jsx
```diff
@@ -604,20 +605,19 @@ const PlaybookAddComponentsContent = ({
                             placement="bottom-start"
                           >
                             <MenuItem value={value.const}>
-                              {value.title}
+                              {/* value might be an entity type, we try to translate it */}
+                              {translateEntityType(value.title)}
```
labo-flg, line 609: fix missing translations of the `oneOf` static list given by the playbook API. In the case of the container wrapper, it's entity types. Here I just make an attempt at translating the given value.

opencti-platform/opencti-graphql/src/database/redis.ts
```diff
@@ -787,7 +787,9 @@ export const getLastPlaybookExecutions = async (playbookId: string) => {
   return executions.map((e) => {
     const steps = Object.entries(e).filter(([k, _]) => k.startsWith('step_')).map(([k, v]) => {
       const bundle_or_patch = v.bundle ? JSON.stringify([v.bundle], null, 2) : JSON.stringify(v.patch, null, 2);
-      return ({ id: k.split('step_')[1], bundle_or_patch, ...v });
+      // beware, step key is the same for every execution, and we need to avoid id collision in Relay
+      const id = `${e.playbook_execution_id}-${k.split('step_')[1]}`;
+      return ({ id, bundle_or_patch, ...v });
```
labo-flg, line 792: This was throwing errors in the console frontend side, and would lead to data inconsistencies as displayed in the "last execution steps" of a playbook.

opencti-platform/opencti-graphql/src/modules/playbook/playbook-components.ts
```diff
@@ -113,7 +115,7 @@ const PLAYBOOK_LOGGER_COMPONENT: PlaybookComponent<LoggerConfiguration> = {
   schema: async () => PLAYBOOK_LOGGER_COMPONENT_SCHEMA,
   executor: async ({ bundle, playbookNode }) => {
     if (playbookNode.configuration.level) {
-      logApp._log(playbookNode.configuration.level, '[PLAYBOOK MANAGER] Logger component output', { bundle });
+      logApp._log(playbookNode.configuration.level, '[PLAYBOOK MANAGER] Logger component output', null, { bundle });
```
labo-flg, line 118: Logger step simply did not print the bundle data, so it was pretty useless.

opencti-platform/opencti-graphql/src/schema/general.js
```diff
@@ -88,6 +88,7 @@ export const ABSTRACT_TYPES = [
   ABSTRACT_STIX_CORE_OBJECT,
   ABSTRACT_STIX_DOMAIN_OBJECT,
   ENTITY_TYPE_CONTAINER,
+  ENTITY_TYPE_CONTAINER_CASE,
```
labo-flg, line 91: This is a small change that actually hides a lot of cleanup to do!
Case is seemingly abstract, but it defines stix-converter, register in the schema... like a regular type. For another PR.

opencti-platform/opencti-front/src/utils/hooks/useEntityTranslation.ts
```diff
@@ -0,0 +1,21 @@
+import { useFormatter } from '../../components/i18n';
+
+const useEntityTranslation = () => {
+  const { t_i18n } = useFormatter();
+
+  const translateEntityType = (type: string) => {
+    if (t_i18n(`entity_${type}`) !== `entity_${type}`) {
+      return t_i18n(`entity_${type}`);
+    }
+    if (t_i18n(`relationship_${type}`) !== `_${type}`) {
```
labo-flg:
```suggestion
    if (t_i18n(`relationship_${type}`) !== `relationship_${type}`) {
```

opencti-platform/opencti-graphql/src/modules/playbook/playbook-components.ts
```diff
@@ -320,6 +322,18 @@ const PLAYBOOK_CONTAINER_WRAPPER_COMPONENT_SCHEMA: JSONSchemaType<ContainerWrapp
   },
   required: ['container_type'],
 };
+
+// For now, only a fixed list of containers are compatible
+// these are the containers that can be created with a name and no specific mandatory fields
+const PLAYBOOK_CONTAINER_WRAPPER_COMPONENT_AVAILABLE_CONTAINERS = [
```
SouadHadjiat, line 328: this list is used only for displaying the supported components that can be selected ? for existing playbooks components with "grouping", it will still continue to be executed in error, right ? at least that what I have locally, my previous playbook with grouping still fails. I think we should ignore it right ? L353, we have this test that we could update : 
`if (container_type && isStixDomainObjectContainer(container_type)) {`

SouadHadjiat:
tested locally for all supported containers types, and they all have been created. However, for tasks containers, the task has been created (with a campaign), and I can see the campaign in the related entities, but when I select the campaign, I don't see the task in the containers : 
image
which is the case for all my other tests with other containers than tasks. Maybe not related to playbooks. 

Also, my previously created playbook with "groupings" which is not supported anymore is still running with error.

opencti-platform/opencti-graphql/src/modules/playbook/playbook-components.ts
labo-flg: Oh good catch ! I forgot playbook config was saved in DB.
I'll update the test, and fail with proper message so user can have feedback.

labo-flg:
> Also, my previously created playbook with "groupings" which is not supported anymore is still running with error.

This would now fail with a meaningful error at the Wrapper step, `this container type is incompatible with the Container Wrapper playbook component`. I hope this is clear enough for users to know they must update their playbook.
I'll see if I can update the user documentation as well.

> However, for tasks containers, the task has been created (with a campaign), and I can see the campaign in the related entities, but when I select the campaign, I don't see the task in the containers

I confirm for Campaigns, but it works fine with Incidents for instance. I suspect a bug unrelated to this issue. If it's simple enough I'll fix it here anayway.

labo-flg:
> I confirm for Campaigns, but it works fine with Incidents for instance. I suspect a bug unrelated to this issue. If it's simple enough I'll fix it here anyway.

@SouadHadjiat fixe din my last commit. It was a faulty filter in the "last containers" query.

SouadHadjiat (approved):

PR #6677, opened 2024-04-12
[backend] use reverse relation on countries located at regions (#6528)

### LINKED ISSUE #6528: Region countries are not highlited anymore
## Description

Sub-Saharan Africa (but same for all regions, nothing is highlighted) 

Expected output (in blue in standard theme, here in red), version 5.11.X:

image

Actual output (6.0.X):

image

frapuks:
### Proposed changes

- Update boolean `reverse_relation` to true, to find countries located at regions

### Related issues

- #6528 

### Checklist

### Further comments

- Before : the countries query for a region returns an empty table, even if countries have a located at relationship with this region.
- The highlithings was working before, but wasn't displaying anything because it wasn't receiving any data.
- Now : the countries query return countries list.
- The highlithings is dependent on country aliases ISO3. With local data, you can try displaying Europe, but first you need to add the FRA alias to France and the ESP alias to Spain
image

- Exemple of simple query after change :
image

### DIFF
--- a/opencti-platform/opencti-graphql/src/domain/region.js
+++ b/opencti-platform/opencti-graphql/src/domain/region.js
@@ -24,7 +24,7 @@ export const childRegionsPaginated = async (context, user, regionId, args) => {
 };
 
 export const countriesPaginated = async (context, user, elementId, args) => {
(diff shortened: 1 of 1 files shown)


SouadHadjiat (approved):
tested locally, I can know list countries for regions (0 results previously)

PR #6686, opened 2024-04-12
[frontend] fix status list limit in manipulate knowledge (#6581)

### LINKED ISSUE #6581: [Playbook] The drop-down menu doesn't show everything
## Description

The drop-down menu in the "manipulate knowledge" component does not display all available statuses, even when using the search function.

## Environment

OCTI 6.0.8

## Reproducible Steps

Before making the repro case, you need to have several status set up on your platform. For example, set 3 - 4 statuses on 5 different entities (otherwise, go directly to the testing platform).

1. Create a playbook and add a "manipulate knowledge" component
2. Try to replace a status -> you won't find all the statuses. Neither by scrolling nor by searching

## Screenshots

Screenshot 2024-04-04 184323

marieflorescontact:
### Proposed changes

* Set the query limit to 100 instead of 10

### Related issues

* [issue/6581](https://github.com/OpenCTI-Platform/opencti/issues/6581)

### Checklist

### Further comments

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/common/form/StatusField.jsx
+++ b/opencti-platform/opencti-front/src/private/components/common/form/StatusField.jsx
@@ -98,7 +98,7 @@ class StatusField extends Component {
 
   searchStatuses() {
     fetchQuery(statusFieldStatusesSearchQuery, {
-      first: 10,
+      first: 100,
       filters: this.props.type
         ? {
           mode: 'and',


opencti-platform/opencti-front/src/private/components/common/form/StatusField.jsx
```diff
@@ -98,7 +98,7 @@ class StatusField extends Component {
 
   searchStatuses() {
     fetchQuery(statusFieldStatusesSearchQuery, {
-      first: 10,
+      first: 100,
```
labo-flg, line 101: Reading the original issue, I'm not sure it's the source of the problem. It's not related to the status being _used_ to be queriable?

marieflorescontact: we want to list all entity workflow statuses, whether in use or not. which is the case. in `ToolBar.jsx` we have a similar use of the `statusFieldStatusesSearchQuery`

lndrtrbn: tested with the playbook given in the issue, that do not fix the problem. IMO it's more about what the query does in the backend side

lndrtrbn: after discussions: yes it fixes the issue but it seems there is an other issue more UI related, cf thread of the issue.

marieflorescontact: after discussion with @Jipegien , this fixes the bug but the UI needs to be improved by grouping status by entity type, which was done by @frapuks in this [pr](https://github.com/OpenCTI-Platform/opencti/pull/6707)

lndrtrbn (approved):

PR #6710, opened 2024-04-16
[frontend] Control on markings (#6699)

### LINKED ISSUE #6699: No control on marking when having multiple markings of the same type
## Description

In **a creation form**, when selecting 2 markings of the same type, the user is not prevented from doing it anymore. User can chose several markings of the same type. Utlimately though, it seems that the marking having the highest order is taken into account.

At the update, the popup is still present.

## Environment

Demo

## Reproducible Steps

1. Go to report
2. Create a report
3. Give it a name
4. Add a first value in marking "TLP:GREEN" 
5. Add a second value in marking "TLP:AMBER"

## Expected Output

The pop up displayed below should appear to let the user confirms its choice.

## Actual Output

Nothing happens, the user can add as many definitions of the same type. 

## Additional information

Not sure whether my understanding is correct, that we take the highest marking possible. I've done a few test and this was always verified.

## Screenshots (optional)
image

CelineSebe:
### Related issues

* #6699 

### Checklist

### Further comments

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/analyses/reports/ReportCreation.tsx
+++ b/opencti-platform/opencti-front/src/private/components/analyses/reports/ReportCreation.tsx
@@ -270,6 +270,8 @@ export const ReportCreationForm: FunctionComponent<ReportFormProps> = ({
           <ObjectMarkingField
             name="objectMarking"
             style={fieldSpacingContainerStyle}
+            setFieldValue={setFieldValue}
+
           />
           <ExternalReferencesField
             name="externalReferences"
--- a/opencti-platform/opencti-front/src/private/components/analyses/groupings/GroupingCreation.tsx
+++ b/opencti-platform/opencti-front/src/private/components/analyses/groupings/GroupingCreation.tsx
@@ -229,6 +229,7 @@ export const GroupingCreationForm: FunctionComponent<GroupingFormProps> = ({
           <ObjectMarkingField
             name="objectMarking"
             style={fieldSpacingContainerStyle}
+            setFieldValue={setFieldValue}
           />
           <ExternalReferencesField
             name="externalReferences"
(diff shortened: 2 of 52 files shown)


opencti-platform/opencti-front/src/private/components/analyses/reports/ReportCreation.tsx
```diff
@@ -269,6 +269,9 @@ export const ReportCreationForm: FunctionComponent<ReportFormProps> = ({
           <ObjectMarkingField
             name="objectMarking"
             style={fieldSpacingContainerStyle}
+            onChange={() => {
+            }}
```
labo-flg: this prop can be removed I suppose, as the callback does nothing.

labo-flg:
@CelineSebe I see you updated `ReportCreation` but is this problem also happening on creation forms other than Report ?

Also, If I remember @lndrtrbn worked on adding e2e tests for CRUD operations on reports. This could be added to the tests I believe.

opencti-platform/opencti-front/src/private/components/analyses/reports/ReportCreation.tsx
lndrtrbn: In fact, adding this empty callback "fix" the issue.
And it is the `setFieldValue` that is not needed.

Nonetheless it should not be fixed this way. The real issue is in the `ObjectMarkingField` component where the check on markings is not done if the props `onChange` is undefined.

lndrtrbn (changes requested):
Fix the component `ObjectMarkingField` instead of forcing an empty callback on every creation form

opencti-platform/opencti-front/src/private/components/analyses/reports/ReportCreation.tsx
CelineSebe: It's a quick fix found to enable the pop-up display when the max-markings are merged. We did what was most urgent.

lndrtrbn: In any case this solution is working only on report creation so not sufficient because it is the same on every creation form. It will be less tedious to implement the real fix than apply this work around on every creation form

ValentinBouzinFiligran:
Instead of  forcing an empty callback, I think we should remove the `typeof onChange === 'function'` and enable the `handleOnChange` everytime

but, in this case, to set the new value after the replace action, we will have to pass in props the `setFieldValue` from `Formik` everywhere

It cans be a solution, what do you think @labo-flg @lndrtrbn ?

lndrtrbn (commented):
Missing `setFieldValue` props in some forms

lndrtrbn (approved):

Issue #6702, opened 2024-04-16
Very slow session refresh and possible infinite loop in the cache management


SamuelHassine:
## Description

Very slow session refresh and possible infinite loop in the cache management

PR #6712, opened 2024-04-16
[backend] improve users loading using cache (#6702)

SouadHadjiat:
### Proposed changes

* use cache to load users (for connectors, task manager)
* use cache to load user to get effective confidence level
* add debug log for cache reset

### Related issues

* #6702 

### Checklist

### Further comments

### DIFF
--- a/opencti-platform/opencti-graphql/src/database/cache.ts
+++ b/opencti-platform/opencti-graphql/src/database/cache.ts
@@ -1,5 +1,6 @@
 import { SEMATTRS_DB_NAME, SEMATTRS_DB_OPERATION } from '@opentelemetry/semantic-conventions';
 import type { BasicStoreIdentifier, StoreEntity, StoreRelation } from '../types/store';
+import { logApp } from '../config/conf';
 import { UnsupportedError } from '../config/errors';
 import { telemetry } from '../config/tracing';
 import type { AuthContext, AuthUser } from '../types/user';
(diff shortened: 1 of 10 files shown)


opencti-platform/opencti-graphql/src/database/cache.ts
```diff
@@ -57,6 +58,7 @@ export const resetCacheForEntity = (entityType: string) => {
   const types = [entityType, ...(STORE_ENTITIES_LINKS[entityType] ?? [])];
   types.forEach((type) => {
     if (cache[type]) {
+      logApp.debug('Reset cache for entity', { type, entityType });
```
labo-flg, line 61:
```suggestion
      logApp.debug('Reset cache for entity', { type });
```
`types` array already contains `entity_type` as first value.

SouadHadjiat: in the log we will have "type": the entity type of the cached data, and "entityType" which is the entity type of the instance received in the event. They can be different because for example when we update a group, we will invalidate the cache of users (so entityType will be "Group" and type will be "User"). 
What do you suggest we should log ?

labo-flg: Understood ! I did not catch this was in the `if (cache[type]) {` clause, so yeah you're right !

PR #6714, opened 2024-04-16
[backend] handle string fields trimming in create/update operations (#6708)

### LINKED ISSUE #6708: Creating an entity with ending/starting whitespaces can lead to stream sync infinite loop
## Description
Knowing that:
* (1) All string values are trimmed just before indexing, so if you create an entity with a name having trailing whitespaces, creation will succeed and in the database it will end up without these spaces.
* (2) When a platform receives an update event to ingest, it checks if it would result in changes, and will not patch the object if there is nothing to change.

Now if you receive an update event with just the same name with the whitespaces, the patch check (2) will detect this is a valid patch as it's indeed different from what's in database (the trimmed version IS different from the raw version).
Update will occur, and an event will be emitted. But the object name is not changed in database because (1).

Now the fun part: if you have bidirectional synchronization of 2 OCTI platform on these events, you enter an infinite loop of update events. The initial emitter sees their own creation event coming back, ingested, and thrown back again, etc.

## Environment
6.0.9

## Reproducible Steps

A simple repo case can be achieved with Playbook to simulate this bidirectional sync loop

1. Create a playbook with a Listener on create/update reports + ingester
2. Start the playbook (in local, do not forget to start a Worker)
3. Open the live stream
4. Create a report with name ending with several whitespaces, like `aa         `
5. checkout the live stream 

## Expected Output

1 Creation event and that's it

## Actual Output

Infinite loop of update events

## Additional information

image
image

labo-flg:
When receiving an update payload, we check if such patch would result in real change compared to database, and prevent these blank change to happen.

With a string value having heading or trailing whitespaces, the patched value is different from the database value, as the database value is always trimmed before indexing. Thus the change is authorized whereas there is no change.

In context described in #6708, this leads to an infinite syncing loop.

### Proposed changes

* compare database value with trimmed input value during said check.

### Related issues
* #6708 

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/schema/identifier.js
+++ b/opencti-platform/opencti-graphql/src/schema/identifier.js
@@ -328,11 +328,11 @@ const filteredIdContributions = (contrib, way, data) => {
     }
     const destKey = dest || src;
     const resolver = contrib.resolvers[src];
-    const resolvedValue = (resolver && value) ? resolver(value) : value;
-    if (isEmptyField(resolvedValue)) {
-      return {}; // missing contribution
+    if (resolver) {
+      objectData[destKey] = value ? resolver(value) : value;
+    } else {
+      objectData[destKey] = value;
     }
-    objectData[destKey] = resolvedValue;
   }
   return R.filter((keyValue) => !R.isEmpty(keyValue) && !R.isNil(keyValue), objectData);
 };
--- a/opencti-platform/opencti-graphql/src/schema/identifier.js
+++ b/opencti-platform/opencti-graphql/src/schema/identifier.js
@@ -328,11 +328,11 @@ const filteredIdContributions = (contrib, way, data) => {
     }
     const destKey = dest || src;
     const resolver = contrib.resolvers[src];
-    if (resolver) {
-      objectData[destKey] = value ? resolver(value) : value;
-    } else {
-      objectData[destKey] = value;
+    const resolvedValue = (resolver && value) ? resolver(value) : value;
+    if (isEmptyField(resolvedValue)) {
+      return {}; // missing contribution
     }
+    objectData[destKey] = resolvedValue;
   }
   return R.filter((keyValue) => !R.isEmpty(keyValue) && !R.isNil(keyValue), objectData);
 };
--- a/opencti-platform/opencti-graphql/src/database/middleware.js
+++ b/opencti-platform/opencti-graphql/src/database/middleware.js
@@ -933,7 +933,12 @@ const rebuildAndMergeInputFromExistingData = (rawInput, instance) => {
     }
     finalVal = [patchedInstance[key]];
   } else {
-    const evaluateValue = value ? R.head(value) : null;
+    // now we  check if the new value would actually result in no change in database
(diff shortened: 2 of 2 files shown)


labo-flg:
@Kedae @richard-julien I'd like you're thoughts on this one.

We silently trim all string values before indexing [here](https://github.com/OpenCTI-Platform/opencti/blob/9f8d905345924e8f2eaa21df597f8b57d911df86/opencti-platform/opencti-graphql/src/database/engine.js#L3313). 

Here I trim the input value just before the comparison algo, but I hesitated with a more direct approach : could we trim all string values at the very top, in the graphql layer ?

lndrtrbn:
What about the case where there is only spaces in the input ? After trim it will result into an empty string.

For example you can create a Report with a name looking like "   " and it will be ok for the API. On the contrary you cannot do it for a Malware or an Administrative area for example. Maybe a missing check in the Report API ?

labo-flg:
> For example you can create a Report with a name looking like " " and it will be ok for the API. On the contrary you cannot do it for a Malware or an Administrative area for example. Maybe a missing check in the Report API ?

In fact, the malware creation fails not because of a schema validation but because it cannot generate a standard id from such payload (only a name).

With reports, the GQL validation will force you to give both a `name` and a `published` string date. Thus when generating a standard id, the name gets ditched properly but the published is kept, and id generation is a success: see `identifier.js:generateDataUUID` calling identifier resolvers defined in schema.

```
// for instance in case-incident.ts
identifier: {
    definition: {
      [ENTITY_TYPE_CONTAINER_CASE_INCIDENT]: [{ src: NAME_FIELD }, { src: 'created' }]
    },
    resolvers: {
      name(data: object) {
        return normalizeName(data); // this trims the name
      },
    },
  }
```

opencti-platform/opencti-graphql/src/schema/identifier.js
```diff
@@ -328,11 +328,11 @@ const filteredIdContributions = (contrib, way, data) => {
...
     const destKey = dest || src;
     const resolver = contrib.resolvers[src];
-    if (resolver) {
-      objectData[destKey] = value ? resolver(value) : value;
-    } else {
-      objectData[destKey] = value;
+    const resolvedValue = (resolver && value) ? resolver(value) : value;
+    if (isEmptyField(resolvedValue)) {
+      return {}; // missing contribution
     }
```
labo-flg: For instance, name resolvers always lower-case and trim the given value.
We could thus end up with an empty string where the code expected a real contribution to the uuid generation. Later in the code, this contribution would be discarded.

With malwares that require only name, this would lead to error when generating the identifier.
With Reports, the published field would still be kept allowing to generate the id.

With this fix, creating a report with only space in the name should fail with an error here at identifier generation time.

```
mutation {
  reportAdd(input: { name: "      ", published: "2024-04-17T08:49:10.677Z"}) {id}
  # malwareAdd(input: { name: "     " }) {id}
}
```

```
"errors": [
    {
      "message": "Missing required elements for Report creation (name - published)",
      "name": "UNSUPPORTED_ERROR",
    ...
```

Kedae (approved):

PR #6721, opened 2024-04-17
[frontend] update frontend when adding a user to a group (#6703)

### LINKED ISSUE #6703: Frontend not updated properly when adding a user to a group
## Description

1. Go to a group
2. Open edition
3. Go in members
4. Check a box

=> The box is never checked, you have to full refresh

Archidoit:
issue: https://github.com/OpenCTI-Platform/opencti/issues/6703


labo-flg:
Tested OK, but I'm still having the warns about duplicated keys `(console.js:213 Warning: Encountered two children with the same key,`)


---> fixed

opencti-platform/opencti-front/src/private/components/settings/groups/GroupEditionUsers.tsx
```diff
@@ -0,0 +1,145 @@
...
+      });
+    }
+  };
+
+  return (
+    <List dense={true} className={classes.root}>
+      {users.map((user) => {
+        const groupUser = groupUsers.find((g) => g.id === user.id);
+        return (
+          <ListItem key={groupId} divider={true}>
```
labo-flg:
```suggestion
          <ListItem key={user.id} divider={true}>
```

maybe :)

Archidoit: thanks :) fixed

labo-flg (approved):

PR #6727, opened 2024-04-17
[frontend] fix handling of references enforcment in containers

### LINKED ISSUE #6074: The way to handle external ref is inconsistent

JeremyCloarec:
### Proposed changes

* [x] fix enable references in containers entity lines popover 
* [x] fix enable references in containers knowledge graph (ask for references when removing from graph)
* [x] fix malware analysis creation and edition when external references are enforced
* [x] fix enable references behavior for organization sharing (skip reference in case of orga?)

### Related issues

* #6074 

### Checklist

### Further comments

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/analyses/groupings/GroupingKnowledgeGraphBar.jsx
+++ b/opencti-platform/opencti-front/src/private/components/analyses/groupings/GroupingKnowledgeGraphBar.jsx
@@ -313,10 +313,9 @@ class GroupingKnowledgeGraphBar extends Component {
   submitReference(values, { setSubmitting, resetForm }) {
     const { handleDeleteSelected } = this.props;
     const { deleteObject } = this.state;
-    const commitMessage = values.message;
-    const references = R.pluck('value', values.references || []);
+    const references = (values.references || []).map((ref) => ref.value);
     this.handleCloseRemove();
-    handleDeleteSelected(deleteObject, commitMessage, references, setSubmitting, resetForm);
+    handleDeleteSelected(deleteObject, values.message, references, setSubmitting, resetForm);
   }
 
   handleSubmitRemoveElements() {
@@ -1128,7 +1127,7 @@ class GroupingKnowledgeGraphBar extends Component {
                           {t('Cancel')}
                         </Button>
                         <Button
-                          onClick={ this.handleSubmitRemoveElements.bind(this)}
+                          onClick={this.handleSubmitRemoveElements.bind(this)}
                           color="secondary"
                         >
                           {t('Remove')}
--- a/opencti-platform/opencti-front/src/private/components/analyses/groupings/GroupingKnowledgeGraphBar.jsx
+++ b/opencti-platform/opencti-front/src/private/components/analyses/groupings/GroupingKnowledgeGraphBar.jsx
@@ -36,11 +36,13 @@ import DialogContent from '@mui/material/DialogContent';
 import DialogActions from '@mui/material/DialogActions';
 import Button from '@mui/material/Button';
 import Slide from '@mui/material/Slide';
+import { Form, Formik } from 'formik';
 import Typography from '@mui/material/Typography';
 import Alert from '@mui/material/Alert';
 import AlertTitle from '@mui/material/AlertTitle';
 import FormGroup from '@mui/material/FormGroup';
 import FormControlLabel from '@mui/material/FormControlLabel';
+import CommitMessage from '../../common/form/CommitMessage';
 import StixNestedRefRelationshipCreationFromKnowledgeGraph from '../../common/stix_nested_ref_relationships/StixNestedRefRelationshipCreationFromKnowledgeGraph';
 import inject18n from '../../../../components/i18n';
 import ContainerAddStixCoreObjects from '../../common/containers/ContainerAddStixCoreObjects';
@@ -105,6 +107,7 @@ class GroupingKnowledgeGraphBar extends Component {
       openEditObservable: false,
       displayRemove: false,
       deleteObject: false,
+      referenceDialogOpened: false,
     };
   }
 
@@ -303,6 +306,30 @@ class GroupingKnowledgeGraphBar extends Component {
     this.handleCloseSelectByType();
   }
 
+  closeReferencesPopup() {
+    this.setState({ referenceDialogOpened: false });
+  }
+
+  submitReference(values, { setSubmitting, resetForm }) {
+    const { handleDeleteSelected } = this.props;
+    const { deleteObject } = this.state;
+    const commitMessage = values.message;
+    const references = R.pluck('value', values.references || []);
+    this.handleCloseRemove();
+    handleDeleteSelected(deleteObject, commitMessage, references, setSubmitting, resetForm);
+  }
+
+  handleSubmitRemoveElements() {
+    const { enableReferences, handleDeleteSelected } = this.props;
+    const { deleteObject } = this.state;
+    if (enableReferences) {
+      this.setState({ referenceDialogOpened: true });
+    } else {
+      this.handleCloseRemove();
+      handleDeleteSelected(deleteObject);
+    }
+  }
+
   render() {
     const {
       t,
@@ -373,6 +400,7 @@ class GroupingKnowledgeGraphBar extends Component {
       openEditObservable,
       openEditNested,
       deleteObject,
+      referenceDialogOpened,
     } = this.state;
     const isInferred = selectedNodes.filter((n) => n.inferred).length > 0
       || selectedLinks.filter((n) => n.inferred).length > 0;
@@ -1100,16 +1128,39 @@ class GroupingKnowledgeGraphBar extends Component {
                           {t('Cancel')}
                         </Button>
                         <Button
-                          onClick={() => {
-                            this.handleCloseRemove();
-                            handleDeleteSelected(deleteObject);
-                          }}
+                          onClick={ this.handleSubmitRemoveElements.bind(this)}
                           color="secondary"
                         >
                           {t('Remove')}
                         </Button>
                       </DialogActions>
                     </Dialog>
+                    {enableReferences && (
+                    <Formik
+                      initialValues={{ message: '', references: [] }}
+                      onSubmit={this.submitReference.bind(this)}
+                    >
+                      {({
+                        submitForm,
+                        isSubmitting,
+                        setFieldValue,
+                        values,
+                      }) => (
+                        <Form>
+                          <CommitMessage
+                            handleClose={this.closeReferencesPopup.bind(this)}
+                            open={referenceDialogOpened}
+                            submitForm={submitForm}
+                            disabled={isSubmitting}
+                            setFieldValue={setFieldValue}
+                            values={values.references}
+                            id={grouping.id}
+                            noStoreUpdate={true}
+                          />
+                        </Form>
+                      )}
+                    </Formik>
+                    )}
                   </div>
                 )}
                 <div className="clearfix" />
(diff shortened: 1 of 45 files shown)


nino-filigran:
after a discussion ith @JeremyCloarec we decided to split the fix in 2: the following bug will tackle the rest, since this bug has been opened since a while. Also, the missing cases are a bit bigger to tackle and impact is less important. https://github.com/OpenCTI-Platform/opencti/issues/7023

SarahBocognano:
What is happening with this fix ? Can we review it or it will be split in 2 ?

JeremyCloarec:
> What is happening with this fix ? Can we review it or it will be split in 2 ?

it can be reviewed, the remaining things to fix will be done on a separate PR and will link to this issue https://github.com/OpenCTI-Platform/opencti/issues/7023

opencti-platform/opencti-front/src/private/components/analyses/groupings/GroupingKnowledgeGraphBar.jsx
```diff
@@ -303,6 +306,30 @@ class GroupingKnowledgeGraphBar extends Component {
...
   }
 
+  closeReferencesPopup() {
+    this.setState({ referenceDialogOpened: false });
+  }
+
+  submitReference(values, { setSubmitting, resetForm }) {
+    const { handleDeleteSelected } = this.props;
+    const { deleteObject } = this.state;
+    const commitMessage = values.message;
```
ValentinBouzinFiligran: this constant can be removed to use value.message directly as params on the function ?

opencti-platform/opencti-front/src/private/components/analyses/groupings/GroupingKnowledgeGraphBar.jsx
```diff
@@ -1100,16 +1128,39 @@ class GroupingKnowledgeGraphBar extends Component {
                           {t('Cancel')}
                         </Button>
                         <Button
-                          onClick={() => {
-                            this.handleCloseRemove();
-                            handleDeleteSelected(deleteObject);
-                          }}
+                          onClick={ this.handleSubmitRemoveElements.bind(this)}
```
ValentinBouzinFiligran: blank space

opencti-platform/opencti-front/src/private/components/analyses/malware_analyses/MalwareAnalysisEditionDetails.tsx
```diff
@@ -77,25 +79,28 @@ interface MalwareAnalysisEditionDetailsProps {
 
 interface MalwareAnalysisEditionFormValues {
   message?: string
-  references?: Option[]
+  references: ExternalReferencesValues | undefined
```
ValentinBouzinFiligran, line 82: references: ExternalReferencesValues | undefined better than references?: ExternalReferencesValues

opencti-platform/opencti-front/src/private/components/analyses/groupings/GroupingKnowledgeGraphBar.jsx
```diff
@@ -303,6 +306,30 @@ class GroupingKnowledgeGraphBar extends Component {
...
 
+  closeReferencesPopup() {
+    this.setState({ referenceDialogOpened: false });
+  }
+
+  submitReference(values, { setSubmitting, resetForm }) {
+    const { handleDeleteSelected } = this.props;
+    const { deleteObject } = this.state;
+    const commitMessage = values.message;
+    const references = R.pluck('value', values.references || []);
```
lndrtrbn:
```suggestion
    const references = (values.references || []).map(ref => ref.value);
```
Trying to not use ramda

opencti-platform/opencti-front/src/private/components/analyses/malware_analyses/MalwareAnalysisEditionOverview.tsx
```diff
@@ -116,10 +118,11 @@ interface MalwareAnalysisEditionOverviewProps {
 
 interface MalwareAnalysisEditionFormValues {
   message?: string
-  references?: Option[]
+  references: ExternalReferencesValues | undefined
```
lndrtrbn: Could we fix this interface ? It does not represent the values of the form

opencti-platform/opencti-front/src/private/components/analyses/reports/ReportKnowledgeGraphBar.jsx
```diff
@@ -303,6 +306,30 @@ class ReportKnowledgeGraphBar extends Component {
...
 
+  closeReferencesPopup() {
+    this.setState({ referenceDialogOpened: false });
+  }
+
+  submitReference(values, { setSubmitting, resetForm }) {
+    const { handleDeleteSelected } = this.props;
+    const { deleteObject } = this.state;
+    const commitMessage = values.message;
+    const references = R.pluck('value', values.references || []);
```
lndrtrbn:
```suggestion
    const references = (values.references || []).map(ref => ref.value);
```
Trying to not use ramda

opencti-platform/opencti-front/src/private/components/cases/case_incidents/IncidentKnowledgeGraphBar.jsx
```diff
@@ -303,6 +306,30 @@ class IncidentKnowledgeGraphBar extends Component {
...
 
+  closeReferencesPopup() {
+    this.setState({ referenceDialogOpened: false });
+  }
+
+  submitReference(values, { setSubmitting, resetForm }) {
+    const { handleDeleteSelected } = this.props;
+    const { deleteObject } = this.state;
+    const commitMessage = values.message;
+    const references = R.pluck('value', values.references || []);
```
lndrtrbn:
```suggestion
    const references = (values.references || []).map(ref => ref.value);
```
Trying to not use ramda

opencti-platform/opencti-front/src/private/components/cases/case_rfis/CaseRfiKnowledgeGraphBar.jsx
```diff
@@ -306,6 +309,30 @@ class CaseRfiKnowledgeGraphBar extends Component {
...
 
+  closeReferencesPopup() {
+    this.setState({ referenceDialogOpened: false });
+  }
+
+  submitReference(values, { setSubmitting, resetForm }) {
+    const { handleDeleteSelected } = this.props;
+    const { deleteObject } = this.state;
+    const commitMessage = values.message;
+    const references = R.pluck('value', values.references || []);
```
lndrtrbn:
```suggestion
    const references = (values.references || []).map(ref => ref.value);
```
Trying to not use ramda

opencti-platform/opencti-front/src/private/components/cases/case_rfts/CaseRftKnowledgeGraphBar.jsx
```diff
@@ -306,6 +309,30 @@ class CaseRftKnowledgeGraphBar extends Component {
...
 
+  closeReferencesPopup() {
+    this.setState({ referenceDialogOpened: false });
+  }
+
+  submitReference(values, { setSubmitting, resetForm }) {
+    const { handleDeleteSelected } = this.props;
+    const { deleteObject } = this.state;
+    const commitMessage = values.message;
+    const references = R.pluck('value', values.references || []);
```
lndrtrbn:
```suggestion
    const references = (values.references || []).map(ref => ref.value);
```
Trying to not use ramda

opencti-platform/opencti-front/src/private/components/common/containers/ContainerStixCoreObjectPopover.jsx
```diff
@@ -231,8 +263,21 @@ class ContainerStixCoreObjectPopover extends Component {
...
   }
 
+  closeReferencesPopup() {
+    this.setState({ referenceDialogOpened: false });
+  }
+
+  submitReference(values, { setSubmitting, resetForm }) {
+    const { displayRemove, displayDeleteMapping } = this.state;
+    const commitMessage = values.message;
+    const references = R.pluck('value', values.references || []);
```
lndrtrbn:
```suggestion
    const references = (values.references || []).map(ref => ref.value);
```
Trying to not use ramda

lndrtrbn (commented):
You duplicate the references form a lot of times, could you extract this part in its own component?

opencti-platform/opencti-front/src/private/components/cases/tasks/Root.tsx
```diff
@@ -70,6 +72,7 @@ const RootTaskComponent = ({ queryRef, taskId }) => {
     [taskId],
   );
   const location = useLocation();
+  const enableReferences = useIsEnforceReference('Task') && !useGranted([BYPASSREFERENCE]);
```
SarahBocognano: Just out of curiosity : Why did you use useIsEnforceReference only on Task and Feedbacks and not on other Root components .

JeremyCloarec: I already added useIsEnforceReference in other Root components in a previous PR, I had to add it to Task and Feedbacks also in this PR following the changes to the ContainerStixObjectsOrStixRelationships component

SarahBocognano: Thank you ! :) 
Just a question again: I cannot set an enableReference in EntitySettings for Note/Opinion/Task. Can you explain to me why we did the change in frontend then ?

JeremyCloarec: Good point, I didn't have that in mind when calling useInEnforceReference in Note/Opinion/Task ! In theory we could just not use it, and always set `enabledReferences` to `false` for these components. But I think that it doesn't hurt to leave it like this, this way if we decide that Note/Opinion/Task can be configured with enableReference it will work without having to change it back to use useIsEnforceReference

SarahBocognano:
Capture d'écran 2024-06-14 120843

When I try to edit a report, I submit an external reference as required, the mutation executes but the form doesn't disappear. It stays blocked like on the screenshot

SarahBocognano:
image

When trying to add a new entity in knowledge graph I have a validation error but the message is empty in snackbar. (Although I have it in the backend "errors":[{"attributes":{"field":"references","genre":"BUSINESS","http_status":500,"message":"You must provide at least one external reference to update"}"

JeremyCloarec:
> image
> 
> When trying to add a new entity in knowledge graph I have a validation error but the message is empty in snackbar. (Although I have it in the backend "errors":[{"attributes":{"field":"references","genre":"BUSINESS","http_status":500,"message":"You must provide at least one external reference to update"}"

I couldn't reproduce, were your backend/frontend completely rebuilt when switching to the branch? This looks like the behavior without the modifications of the branch

SarahBocognano:
Update: Behavior is working well for me, the only thing remaining is that the form doesn't disappear when update is selected. Need to check the updater

SarahBocognano (approved):

PR #6747, opened 2024-04-18
[backend] update GQL API: names cannot be blank (#6742)

### LINKED ISSUE #6742: Reports can be created with blank names
## Description

Following this PR: https://github.com/OpenCTI-Platform/opencti/pull/6714#issuecomment-2060646879

Using the frontend or the API, today we can create some entities (including reports) with names made only of whitespaces characters.
These attributes would be trimmed (like any string) before indexing, ending up empty.

For some mandatory field, this means we pass all check/validation to finally still be empty.

Note that depending on the schema and frontend validation, sometimes you are not allowed to do so.
For instance, a malware only needs a name for generating the stix id. The generation process would fail as a name made of blanks would not be valid for creating the stix id, and the creation would fail at this step.
For reports, the contributions to the stixid generation are `name`+`published`, thus it would still have `published` to make the id.... and the mandatory name would end up empty in DB because no other check is made!

## Environment

6.0.10

## Reproducible Steps

Playground query that sould not work but is successful:
```
mutation {
  reportAdd(input: { name: "     ", published: "2024-04-17T08:49:10.677Z"}) {id}
}
```

Playground query that fails for identifier reason (wer're lucky)
```
mutation {
  malwareAdd(input: { name: "     " }) {id} 
}
```

## Additional information

Solutions to consider: 
* use new scalar type in graphql `StringNotEmpty` for all required string fields such as the report name.
* use new constraint format

labo-flg:
Some schema constraints are in place in our GQL schema, but this is not consistent across the API.

With this PR we targets the mandatory `name` string fields, that should all given 2 constraints: minimum length 2 (this is already the case for some but not all object types), and the impossibility to be only made of whitespaces

### Proposed changes

* introduce new `format` constraint named `not-blank` registered in `graphql-constraints-directive` plugin to check if a value is only made of whitespaces
* pass on all the Input types in the API to consistently add constraints `name: String! @constraint(minLength: 2, format: "not-blank")`

### Related issues
* closes #6742 

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/modules/vocabulary/vocabulary.graphql
+++ b/opencti-platform/opencti-graphql/src/modules/vocabulary/vocabulary.graphql
@@ -123,7 +123,7 @@ type Query {
 input VocabularyAddInput {
   stix_id: StixId
   x_opencti_stix_ids: [StixId]
-  name: String! @constraint(minLength: 1)
+  name: String! @constraint(minLength: 2, format: "not-blank")
   description: String
   category: VocabularyCategory!
   order: Int
--- a/opencti-platform/opencti-graphql/src/modules/vocabulary/vocabulary.graphql
+++ b/opencti-platform/opencti-graphql/src/modules/vocabulary/vocabulary.graphql
@@ -123,7 +123,7 @@ type Query {
 input VocabularyAddInput {
   stix_id: StixId
   x_opencti_stix_ids: [StixId]
-  name: String! @constraint(minLength: 2, format: "not-blank")
+  name: String! @constraint(format: "not-blank")
   description: String
   category: VocabularyCategory!
   order: Int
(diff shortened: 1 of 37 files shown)


opencti-platform/opencti-graphql/src/modules/vocabulary/vocabulary.graphql
```diff
@@ -123,7 +123,7 @@ type Query {
 input VocabularyAddInput {
   stix_id: StixId
   x_opencti_stix_ids: [StixId]
-  name: String! @constraint(minLength: 1)
+  name: String! @constraint(format: "not-blank")
```
labo-flg:
```suggestion
  name: String! @constraint(minLength: 1, format: "not-blank")
```

marieflorescontact (commented):
Did you miss `Location`?

labo-flg:
> Did you miss `Location`?

indeed, need constraints  in `LocationAddInput.name` too

lndrtrbn (approved):

PR #6750, opened 2024-04-18
[frontend] error message display when commit mutation (#6725)

### LINKED ISSUE #6725: No error message when max confidence level not enough to update an entity
## Reproducible Steps
- Log as a user with max confidence level  = 50
- Delete an entity with confidence level = 100
- The screen freezes and nothing happens
image

## Expected behavior
An error message, as when the user update an entity
image

## Comment
The bug is not the same for all the entities:
- bug for regions, case RFI
- no bug for attack patterns
- empty error message for malware analysis : 
image

Archidoit:
issue: https://github.com/OpenCTI-Platform/opencti/issues/6725
(POC done for Report and case RFT)

Create a hook handling both useMutation and the eventual errors thrown from backend to display

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/locations/regions/RegionPopover.tsx
+++ b/opencti-platform/opencti-front/src/private/components/locations/regions/RegionPopover.tsx
@@ -80,10 +80,6 @@ const RegionPopover = ({ id }: { id: string }) => {
         handleClose();
         navigate('/dashboard/locations/regions');
       },
-      onError: (error) => {
-        const { errors } = (error as unknown as RelayError).res;
-        MESSAGING$.notifyError(errors[0].message);
-      },
     });
   };
   return (
--- a/opencti-platform/opencti-front/src/private/components/locations/regions/RegionPopover.tsx
+++ b/opencti-platform/opencti-front/src/private/components/locations/regions/RegionPopover.tsx
@@ -8,7 +8,7 @@ import DialogActions from '@mui/material/DialogActions';
 import DialogContent from '@mui/material/DialogContent';
 import DialogContentText from '@mui/material/DialogContentText';
 import MoreVert from '@mui/icons-material/MoreVert';
-import { graphql, useMutation } from 'react-relay';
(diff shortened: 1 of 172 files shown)


opencti-platform/opencti-front/src/utils/hooks/useApiMutation.ts
```diff
@@ -0,0 +1,17 @@
+import { Disposable, GraphQLTaggedNode, IEnvironment, MutationConfig, MutationParameters } from 'relay-runtime';
+import { useMutation, UseMutationConfig } from 'react-relay';
+import { useCallback } from 'react';
+import { relayErrorHandling } from '../../relay/environment';
+
+const useApiMutation = <T extends MutationParameters>(
```
lndrtrbn, line 9:
```suggestion
/**
 * Hook wrapping Relay useMutation to automatically display an error popup if the mutation fails
 */ 
const useApiMutation = <T extends MutationParameters>(
```

opencti-platform/opencti-front/src/private/components/locations/regions/RegionPopover.tsx
```diff
@@ -77,6 +80,10 @@ const RegionPopover = ({ id }: { id: string }) => {
         handleClose();
         navigate('/dashboard/locations/regions');
       },
+      onError: (error) => {
+        const { errors } = (error as unknown as RelayError).res;
+        MESSAGING$.notifyError(errors[0].message);
+      },
```
lndrtrbn: Is it on purpose we override the default onError from useApiMutation ?

Archidoit: Sorry, I forgot to remove that.
Fixed

opencti-platform/opencti-front/src/utils/hooks/useApiMutation.ts
Archidoit: done :)

lndrtrbn (approved):

Issue #4900, opened 2023-11-14
Add overrides to the max confidence level of Users for specific Entities


Jipegien:
## Use case

Users (and connectors/feeds/stream) might be more or less trustable in accordance with the nature of data they provide. We will add the possibility to configure overrides to the max confidence level of a user for specific Entities.

Subtask:
* Managing overrides of the max confidence level of a user for specified Objects
* Take into account override when upserting data from a connector or a feed
* Be able to view the max confidence level in the list of users when setting a stream or a feed

Blocked by #4304

labo-flg:
Only chunk 1 merged.

PR #6778, opened 2024-04-22
[frontend] Confidence level by users: Overrides by Entity types (#4900)

SarahBocognano:
### Proposed changes

* Add the front part of the previous backend chunk concerning 'Confidence level by users: Overrides by Entity Types' 
* The ability to change a user's confidence level has now been moved to its own section, and it is possible to set a confidence level for a user for a specific entity. It is also possible to have a confidence level for an entity even if the user does not have a defined confidence level.

### Related issues
* https://github.com/OpenCTI-Platform/opencti/issues/4900

### Checklist

### Further comments


opencti-platform/opencti-front/src/utils/hooks/useSchema.ts
```diff
@@ -9,7 +16,35 @@ const useSchema = () => {
...
+      )
+      .concat(
+        smos.map((smo) => ({
+          ...smo,
+          value: smo.id,
+          type: 'entity_Stix-Meta-Objects',
+        })),
+      );
+    setAvailableEntityTypes(entityTypes);
+  }, [schema]);
```
lndrtrbn: FYI: instead of using a state + useEffect you could use a simple useMemo

opencti-platform/opencti-front/src/private/components/settings/users/UserConfidenceLevel.tsx
```diff
@@ -24,7 +24,17 @@ const UserConfidenceLevel: React.FC<UserConfidenceLevelProps> = ({ user }) => {
...
+    return (
+      <>
+        {user.effective_confidence_level?.overrides?.map((override, index) => (
+          <div key={index} style={{ marginTop: '5px' }}>
+            {`${override.entity_type}: ${override.max_confidence}`}
+          </div>
+        ))}
+      </>
+    );
+  };
```
lndrtrbn: It's more React oriented to created a component instead of creating some "render" functions.
Here something like `ConfidenceOverrides`

lndrtrbn: Same for `renderSource` below

opencti-platform/opencti-front/src/private/components/settings/users/edition/UserEditionConfidence.tsx
```diff
@@ -0,0 +1,213 @@
...
+              variables: {
+                id: user.id,
+                input: {
+                  key: 'user_confidence_level',
+                  object_path: '/user_confidence_level/max_confidence',
+                  value: [null],
+                },
+              },
+            });
+          }
```
lndrtrbn: Only the value changes between those cases. Compute them in the ifs and make only one call to `commitFieldPatch` after

opencti-platform/opencti-front/src/utils/hooks/useConfidenceLevel.ts
```diff
@@ -5,19 +5,39 @@ import { useFormatter } from '../../components/i18n';
...
       }
       return false;
-    } if (!effectiveConfidenceLevel && entity.confidence) {
+    }
+
+    const entityConfidenceLevel = effectiveConfidenceLevel(entity.entity_type);
+
+    if (entity.confidence && entityConfidenceLevel >= entity.confidence) {
+      return true;
+    } if (entity.confidence) {
```
lndrtrbn: Missing a new line

lndrtrbn (commented):
Do we need a not-yet-merged branch for the backend side? I cannot make it works locally

lndrtrbn (approved):

PR #6831, opened 2024-04-29
[backend] restrict knowledge widgets to core relationships, sightings and 'contains' (#6777)

### LINKED ISSUE #6777: Widgets handle relation between Role and Capability
## Reproductible steps
- Create an horizontal bar widgets with attribute = entity_type
image

In the graph, there is a line for 'capability' (i.e. the relationship between a role and a capability is taken into account in the graph) : 
image

- Update the widget with attribute = entity and enable 'display the source'.
image

An error occurs because of the 'role' entity type.
image

## Comment
If you have lots of relationships, you should 
- add filters (exemple: create a new relation between a role and a capability and filter on the creation date to keep the most recent relationships) 
- or upgrade the number of results

to see the relations between Role and Capability

Archidoit:
issue: https://github.com/OpenCTI-Platform/opencti/issues/6777#issuecomment-2082086679

### DIFF
--- a/opencti-platform/opencti-graphql/src/domain/stixRelationship.js
+++ b/opencti-platform/opencti-graphql/src/domain/stixRelationship.js
@@ -65,16 +65,17 @@ export const stixRelationshipDelete = async (context, user, stixRelationshipId) 
   return stixRelationshipId;
 };
 
-const buildValidWidgetRelationshipType = (relationship_type) => {
-  const isValidRelationshipType = isStixCoreRelationship(relationship_type)
-    || isStixSightingRelationship(relationship_type)
-    || relationship_type === RELATION_CONTAINS;
-  return isValidRelationshipType ? relationship_type : widgetsRelationshipTypes;
+const buildValidWidgetRelationshipTypes = (relationshipTypes) => {
+  const isValidRelationshipTypes = relationshipTypes && relationshipTypes.length > 0 && (relationshipTypes.every((type) => (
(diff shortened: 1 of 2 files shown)


opencti-platform/opencti-graphql/src/domain/stixRelationship.js
```diff
@@ -62,16 +65,24 @@ export const stixRelationshipDelete = async (context, user, stixRelationshipId)
   return stixRelationshipId;
 };
 
+const buildValidWidgetRelationshipType = (relationship_type) => {
+  const isValidRelationshipType = isStixCoreRelationship(relationship_type)
+    || isStixSightingRelationship(relationship_type)
+    || relationship_type === RELATION_CONTAINS;
+  return isValidRelationshipType ? relationship_type : widgetsRelationshipTypes;
```
labo-flg: it can be both a string and an array of strings, is it ok ?

Archidoit: it's not ok indeed...

marieflorescontact (approved):
locally tested ok 👌

opencti-platform/opencti-graphql/src/domain/stixRelationship.js
Archidoit: fixed :)

labo-flg (approved):

PR #6841, opened 2024-04-29
[backend] Add missing auth directives

SouadHadjiat:
### Proposed changes

* add missing auth directives with the right capabilities

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/graphql/authDirective.js
+++ b/opencti-platform/opencti-graphql/src/graphql/authDirective.js
@@ -7,6 +7,23 @@ import { AuthRequired, ForbiddenAccess, OtpRequired, OtpRequiredActivation } fro
 import { OPENCTI_ADMIN_UUID } from '../schema/general';
 import { BYPASS, VIRTUAL_ORGANIZATION_ADMIN, SETTINGS_SET_ACCESSES } from '../utils/access';
 
+const PUBLIC_ENDPOINTS = [
+  'about',
+  'aiEndpoints',
+  'feeds',
+  'filterKeysSchema',
+  'schemaRelationsRefTypesMapping',
+  'schemaRelationsTypesMapping',
+  'settings',
+  'streamCollections',
+  'taxiiCollections',
+  'token',
+];
+
+const isPublicEndpointInSchema = (fieldName) => {
+  return fieldName.startsWith('public') || PUBLIC_ENDPOINTS.includes(fieldName);
+};
+
 // eslint-disable-next-line
 export const authDirectiveBuilder = (directiveName) => {
   const typeDirectiveArgumentMaps = {};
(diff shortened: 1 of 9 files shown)


aHenryJard (approved):

opencti-platform/opencti-graphql/src/graphql/authDirective.js
```diff
@@ -7,6 +7,23 @@ import { AuthRequired, ForbiddenAccess, OtpRequired, OtpRequiredActivation } fro
 import { OPENCTI_ADMIN_UUID } from '../schema/general';
 import { BYPASS, VIRTUAL_ORGANIZATION_ADMIN, SETTINGS_SET_ACCESSES } from '../utils/access';
 
+const PUBLIC_ENDPOINTS = [
```
richard-julien: Im not  a big fan of this part has it duplicate the definition of the graphql schema security.
So i think this code needs to be removed and some annotation fixed.

**Partially public**

----- ok
* about is not fully public, only the version (schema is well defined)
* settings is not fully public, only some fields to correctly display the colors, etc (schema is well defined)

**Full public**

----- ok
* feeds / streamCollections / taxiiCollections are public because visibility is managed through the code (all good)
* token is for form login (all good)

----- not ok
* aiEndpoints seems not implemented, need to be removed from the schema
* filterKeysSchema must be protected with @auth
* schemaRelationsRefTypesMapping must be protected with @auth
* schemaRelationsTypesMapping must be protected with @auth

richard-julien (changes requested):

opencti-platform/opencti-graphql/src/graphql/authDirective.js
SouadHadjiat: @richard-julien when you say "this code needs to be removed", do you mean I should remove the whole protection I added in authDirective ? this was meant to throw an error if there are unprotected endpoints.

richard-julien: Yes. As it is it doest really protect complex object like about and duplicate what already defined in the schema.

SouadHadjiat: Ok I will remove it. 
But how will we be able to know then if we introduce unprotected queries or mutations ? we have so many endpoints, it's impossible to check, at least here we had a test for queries and mutations

richard-julien: Thats a good question :) Maybe we can force in another PR a @public annotation and check if all element have at least a @auth or @public to force the declaration. It will never prevent some errors of capabilities in auth but maybe help to identify any public api. wdyt?

SouadHadjiat: yes we could, I also thought about it, but we would still test only queries and mutations ?

SouadHadjiat: I just tested adding a public directive and it works, should I wait to create another PR or push a commit in this one ?
The test in authDirective (line 26) would be :
``` 
if (!authDirective && (typeName === 'Query' || typeName === 'Mutation')) {
  const publicDirective = getDirective(schema, fieldConfig, 'public')?.[0];
  if (!publicDirective) {
    throw new Error(`Unsecure schema: missing auth or public directive for ${_fieldName}`);
  }
}
```

richard-julien: If you want to push it, i think i have the time to do a review.

SouadHadjiat: I just pushed it

richard-julien: Looks good.
Dont know why the e2e test fail, but if not related, go merge 👍🏽

richard-julien (approved):

PR #6735, opened 2024-04-17
[frontend/backend]  Marking definitions for uploaded files (#5823)

ValentinBouzinFiligran:
### Proposed changes

* marking step added before file upload 
* marking form added on free text upload form
* fileMarkings key passed on global upload query
* fileMarkings key passed on entity upload query

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/5823
*

### Checklist

- [x] I consider the submitted work as finished
- [x] I tested the code for its functionality
- [ ] I wrote test cases for the relevant uses case (coverage and e2e)
- [ ] I added/update the relevant documentation (either on github or on notion)
- [x] Where necessary I refactored code to improve the overall quality

### Further comments

If this is a relatively large or complex change, kick off the discussion by explaining why you chose the solution you did and what alternatives you considered, etc...

* https://github.com/OpenCTI-Platform/opencti/pull/6030
* https://github.com/OpenCTI-Platform/client-python/pull/613
* https://github.com/OpenCTI-Platform/connectors/pull/2011

PR #6510, opened 2024-03-28
[backend/frontend] Adding support package + copy/paste of UI errors (#5548)

aHenryJard:
#### Support package generation and download
* Add support package feature backend + UI
* User can request to create a support package
* User can request to zip all file even if it's not all uploaded
* User can see the list of support package that can be downloaded
* Users can delete a support package by id (both elastic data and storage data)

On backend side it means:
- one new module `support`
- one new listener `supportPackageListener`

On frontend side:
- A new screen on settings menu
Capture d'écran 2024-04-29 152253

Refactors:
- Adding file-storage-helper to slowly move from js to ts. And moved some funtions there.
- Moved NODE_INSTANCE_ID from `clusterManager.ts` to `conf.js` to avoid dependencies order issues
- Add a way to run 2 backend + frontend locally with intelliJ configuration
- use only one `wait` method everywhere, instead of code duplication (wait duplicated in managers)
- type file-storage#upload (in file-storage-helper) and use it everywhere (diff `upload` -> `uploadToStorage` )

#### Copy/Paste of stacktrace of UI errors
* Add copy/paste button on error popins, that allows to copy the underlying stacktrace.
Capture d'écran 2024-04-29 152340

#### Side fix:
- Test cleanup was not deleting the test bucket, because bucket need to be empty first. Cleanup of storage is working on test cleanup run now.

#### Development addition:
- A way to run a cluster locally
Capture d'écran 2024-04-29 151928

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/5548

### Checklist

- [x] I consider the submitted work as finished
- [x] I tested the code for its functionality
- [x] I wrote test cases for the relevant uses case
- [x] I added/update the relevant documentation (either on github or on notion)
- [x] Where necessary I refactored code to improve the overall quality

### Further comments

If this is a relatively large or complex change, kick off the discussion by explaining why you chose the solution you did and what alternatives you considered, etc...

Issue #5548, opened 2024-01-17
Be able to generate a "support package" by clicking on a button


SamuelHassine:
## Use case

Generate a zip file with

- Environment variables with redacted secrets => environment.txt
- Statistics => stats.txt
- Error logs files => in a folder "logs"

nino-filigran:
stats - like in connector. Generate suppot package CTA in the platform.

SamuelHassine:
It should be available only to people with "SETACCESS" capability (aka admins), in the settings, somewhere....

PR #6868, opened 2024-05-02
[backend] missing opts definition file_marking for file upload(#5548)

aHenryJard:
### Proposed changes

* Missing a opts definition in file-storage-helper.

### Related issues

Small left over between 2 PRs:
* https://github.com/OpenCTI-Platform/opencti/pull/6735
* https://github.com/OpenCTI-Platform/opencti/pull/6510

### Checklist

### Further comments

### DIFF
--- a/opencti-platform/opencti-graphql/src/database/file-storage-helper.ts
+++ b/opencti-platform/opencti-graphql/src/database/file-storage-helper.ts
@@ -14,6 +14,7 @@ interface FileUploadOpts {
   meta? : any,
   noTriggerImport?: boolean,
   errorOnExisting?: boolean,
+  file_markings?: [],
 }
 
 interface FileUploadData {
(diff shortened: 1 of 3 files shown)


opencti-platform/opencti-graphql/src/database/file-storage-helper.ts
```diff
@@ -14,6 +14,7 @@ interface FileUploadOpts {
   meta? : any,
   noTriggerImport?: boolean,
   errorOnExisting?: boolean,
+  file_markings?: [],
```
lndrtrbn: By typing it using `[]` you are telling that file_markings is an empty array, need to replace with an array of what, if we don't know the type we can use `unkown[]` or `any[]`

lndrtrbn (dismissed):

opencti-platform/opencti-graphql/src/database/file-storage-helper.ts
aHenryJard: ah ok I see

Kedae: Should be string[] I think

labo-flg: will be addressed in https://github.com/OpenCTI-Platform/opencti/pull/6868

Kedae (approved):

PR #6901, opened 2024-05-07
Fix flaky e2e test

marieflorescontact:
### Proposed changes

* Flaky tests when trying to edit multiple fields,
* Solution suggested: click outside an input before going to the next

### Checklist

### Further comments

In CI it looks good but the better test we can do is to merge and see how it goes on other dev branches


aHenryJard (approved):
Let's try it.

PR #6910, opened 2024-05-07
[frontend/backend] Fix incorrect use of stixCoreRelationshipDistribution queries (#6873)

### LINKED ISSUE #6873: [technical] incorrect use of stixCoreRelationshipDistribution queries
In Report details, the entities distribution uses widget `EntityStixCoreRelationshipsHorizontalBars` which uses the query `stixCoreRelationshipsDistribution`.

In this case we send `relationship_type: "object"` to this query, but this is a ref relationship not a stix-core-relationship.

With #6872 we now properly check if the provided `relationship_type` is consistent with the query expectations, so sending `object` would throw an error.

We currently disabled the error throwing in the check in stixCoreRelationshipsDistribution to give time to fix the issue frontend side.

We need to properly use other widget such as `StixRelationshipsHorizontalBars` instead of `EntityStixCoreRelationshipsHorizontalBars`, but this means a refactoring (StixRelationshipsHorizontalBars currently won't accept props such as `relationship_type`).

We also need to check throughout the frontend code if there are other places with failing widgets for same reasons.

marieflorescontact:
### Proposed changes

* use `StixRelationshipsHorizontalBars` instead of ` EntityStixCoreRelationshipsHorizontalBars` in `GroupingDetails` and `reportDetails` components => these were the only 2 components where  `EntityStixCoreRelationshipsHorizontalBars` used an `object` relationshipType
*  error  throws for stixCoreRelationshipDistribution query is not commented any more.

### Related issues

* [issue/6873](https://github.com/OpenCTI-Platform/opencti/issues/6873)

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/analyses/groupings/GroupingDetails.jsx
+++ b/opencti-platform/opencti-front/src/private/components/analyses/groupings/GroupingDetails.jsx
@@ -170,7 +170,6 @@ const GroupingDetailsComponent = (props) => {
               dataSelection={entitiesDistributionDataSelection}
               parameters={{ title: 'Entities distribution' }}
               variant="inEntity"
-              withExportPopover={true}
               isReadOnly={true}
             />
           </Grid>
--- a/opencti-platform/opencti-front/src/private/components/analyses/groupings/GroupingDetails.jsx
+++ b/opencti-platform/opencti-front/src/private/components/analyses/groupings/GroupingDetails.jsx
@@ -19,6 +19,7 @@ import inject18n from '../../../../components/i18n';
 import ExpandableMarkdown from '../../../../components/ExpandableMarkdown';
 import ItemIcon from '../../../../components/ItemIcon';
 import ItemMarkings from '../../../../components/ItemMarkings';
+import { emptyFilterGroup } from '../../../../utils/filters/filtersUtils';
 
 const styles = (theme) => ({
   paper: {
@@ -112,7 +113,7 @@ const GroupingDetailsComponent = (props) => {
   });
   const expandable = grouping.relatedContainers.edges.length > 5;
 
-  const entitiesDidstributionDataSelection = [
+  const entitiesDistributionDataSelection = [
     {
       label: '',
       attribute: 'entity_type',
@@ -133,16 +134,8 @@ const GroupingDetailsComponent = (props) => {
         ],
         filterGroups: [],
       },
-      dynamicFrom: {
-        mode: 'and',
-        filters: [],
-        filterGroups: [],
-      },
-      dynamicTo: {
-        mode: 'and',
-        filters: [],
-        filterGroups: [],
-      },
+      dynamicFrom: emptyFilterGroup,
+      dynamicTo: emptyFilterGroup,
     },
   ];
 
@@ -174,8 +167,8 @@ const GroupingDetailsComponent = (props) => {
               startDate={null}
               endDate={null}
               relationshipType="object"
-              dataSelection={entitiesDidstributionDataSelection}
-              parameters={'Entities distribution'}
+              dataSelection={entitiesDistributionDataSelection}
+              parameters={{ title: 'Entities distribution' }}
               variant="inEntity"
               withExportPopover={true}
               isReadOnly={true}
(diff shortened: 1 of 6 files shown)


opencti-platform/opencti-front/src/private/components/analyses/reports/ReportDetails.jsx
```diff
@@ -164,15 +199,17 @@ const ReportDetailsComponent = (props) => {
...
               fromId={report.id}
-              toTypes={['Stix-Core-Object']}
+              startDate={null}
+              endDate={null}
               relationshipType="object"
-              field="entity_type"
-              seriesName={t('Number of entities')}
-              isTo={true}
+              dataSelection={entitiesDidstributionDataSelection}
+              parameters={'Entities distribution'}
```
labo-flg: parameters is not an object?

labo-flg: judging from the use of `parameters` [here](https://github.com/OpenCTI-Platform/opencti/blob/c394cc812d99db764aba933376cd0089990eba1a/opencti-platform/opencti-front/src/private/components/common/stix_relationships/StixRelationshipsHorizontalBars.jsx#L166) and [there](https://github.com/OpenCTI-Platform/opencti/blob/c394cc812d99db764aba933376cd0089990eba1a/opencti-platform/opencti-front/src/private/components/common/stix_relationships/StixRelationshipsHorizontalBars.jsx#L184) you probably
want to set `parameters={{ title: 'Entities distribution' }}`

opencti-platform/opencti-front/src/private/components/analyses/reports/ReportDetails.jsx
```diff
@@ -122,6 +122,41 @@ const ReportDetailsComponent = (props) => {
...
+      dynamicFrom: {
+        mode: 'and',
+        filters: [],
+        filterGroups: [],
+      },
+      dynamicTo: {
+        mode: 'and',
+        filters: [],
+        filterGroups: [],
+      },
```
labo-flg: you can `import { emptyFilterGroup } from '../../../utils/filters/filtersUtils';` to simplify this part

```suggestion
      dynamicFrom: emptyFilterGroup,
      dynamicTo: emptyFilterGroup,
```

marieflorescontact:
PR status: all comments have been taken into account

opencti-platform/opencti-front/src/private/components/analyses/reports/ReportDetails.jsx
```diff
@@ -164,15 +192,17 @@ const ReportDetailsComponent = (props) => {
...
+              startDate={null}
+              endDate={null}
               relationshipType="object"
-              field="entity_type"
-              seriesName={t('Number of entities')}
-              isTo={true}
+              dataSelection={entitiesDistributionDataSelection}
+              parameters={{ title: 'Entities distribution' }}
+              variant="inEntity"
+              withExportPopover={true}
```
labo-flg:
```suggestion
```

opencti-platform/opencti-front/src/private/components/analyses/groupings/GroupingDetails.jsx
```diff
@@ -133,15 +161,17 @@ const GroupingDetailsComponent = (props) => {
...
+              startDate={null}
+              endDate={null}
               relationshipType="object"
-              field="entity_type"
-              seriesName={t('Number of entities')}
-              isTo={true}
+              dataSelection={entitiesDistributionDataSelection}
+              parameters={{ title: 'Entities distribution' }}
+              variant="inEntity"
+              withExportPopover={true}
```
labo-flg:
```suggestion
```

labo-flg (changes requested):
We do not want to be able to export as image the result of this widget, if we want same behavior as before

opencti-platform/opencti-front/src/private/components/analyses/reports/ReportDetails.jsx
marieflorescontact: oups I forgot this

labo-flg (approved):

Issue #6799, opened 2024-04-24
Ability to filter on representatives starting with X in playbook and streams


Jipegien:
## Use case

I want to filter my playbook on entity with name/title/value (representative) starting with a defined text value.

Should also be made possible in data sharing streams/feeds.

Lhorus6:
"starting" and "contains if possible

Archidoit:
The scope has been restricted to adding the representative filter for: streams, playbooks, triggers

Archidoit:
Scope reduced to playbooks, streams and triggers.

Archidoit:
Scope reduced to playbooks, streams and triggers.

PR #6925, opened 2024-05-10
[frontend/backend] Ability to filter on representatives in streams/playbooks/triggers (#6799)

Archidoit:
Chunk 1 of issue: https://github.com/OpenCTI-Platform/opencti/issues/6799

Chunk 1: ability to filter on representatives in streams/playbooks/triggers

(Chunk 2 later: ability to filter on representatives in feeds, taxii and entity lists)

doc PR : https://github.com/OpenCTI-Platform/docs/pull/165


opencti-platform/opencti-front/src/private/components/profile/triggers/TriggerEditionOverview.tsx
```diff
@@ -469,9 +447,9 @@ const TriggerEditionOverview: FunctionComponent<TriggerEditionOverviewProps> = (
                     filters={filters}
                     helpers={helpers}
                     redirection
-                    searchContext={{ entityTypes: ['Stix-Core-Object', 'stix-core-relationship'] }}
-                  />
-                )}
+                    searchContext={{ entityTypes: ['Stix-Core-Object', 'stix-core-relationship'] }}entityTypes={['Stix-Core-Object', 'stix-core-relationship', 'Stix-Filtering']}
```
SouadHadjiat: looks like an issue here

opencti-platform/opencti-front/src/utils/filters/filtersUtils.tsx
```diff
@@ -79,6 +79,32 @@ export const contextFilters = [
   'contextObjectMarking',
 ];
 
+// filters available in stix filtering (streams, playbooks, triggers)
+export const stixFilters = [
```
SouadHadjiat, line 83: nice refactoring 😌

opencti-platform/opencti-front/src/private/components/data/stream/StreamCollectionCreation.tsx
```diff
@@ -225,6 +203,7 @@ const StreamCollectionCreation: FunctionComponent<StreamCollectionCreationProps>
                 styleNumber={2}
                 redirection
                 searchContext={{ entityTypes: ['Stix-Core-Object', 'stix-core-relationship'] }}
+                entityTypes={['Stix-Core-Object', 'stix-core-relationship', 'Stix-Filtering']}
```
SouadHadjiat, line 206: What is the difference between `searchContext `and `entityTypes `for `FilterIconButton` ? Sometimes we have one of them only, sometimes they are the same, sometimes not.

opencti-platform/opencti-front/src/utils/filters/filtersUtils.tsx
Archidoit: thanks :)

opencti-platform/opencti-front/src/private/components/profile/triggers/TriggerEditionOverview.tsx
Archidoit: fixed

opencti-platform/opencti-front/src/private/components/data/stream/StreamCollectionCreation.tsx
Archidoit: searchContext is used to look for the available entities to display in some filter lists that display entities (exemple: creator filter) and it is used in useSearchEntities // entityTypes is the entity types whose filter keys we want to fetch from the schema

SouadHadjiat:
Not really related to this PR, but do you know why we have two methods for representative extraction :  extractRepresentative (in entity-representative) and extractStixRepresentative (in stix-representative) ? what is the difference between the two ?

Archidoit:
@SouadHadjiat : extractStixRepresentative extracts the representative of an entity that is in the stix format, extractRepresentative extracts the representative of an entity in the BasicStoreCommon format

SouadHadjiat:
Can they return different values ? about the main representative.

Archidoit:
@SouadHadjiat there are some difference. For instance, a status is not defined in stix-representative. But if things have been correctly done, they should be no difference for a given entity type.

Archidoit:
@SouadHadjiat  there are some differences (exemple for Attack Patterns : name in stix representative // composition of x_mitre_id and name in entity representative)

SouadHadjiat (approved):
tested locally ✅

PR #6938, opened 2024-05-13
[frontend] new filters in nested relationships from entity creation list (#6918)

### LINKED ISSUE #6918: Missing Filters option when selecting entities for a nested relationship
## Description

In the drawer for selecting an entity to link to as a nested object, new filters are not implemented.

image

## Reproducible Steps

1. Got to an Observable' knowledge tab
2. Click on the + button to add a nested object
3. deploy drop down "add filter"

## Expected Output

New filters options
image

## Actual Output

Old filters options
image

Archidoit:
PR content:
- issue https://github.com/OpenCTI-Platform/opencti/issues/6918
- improve the 'not removable value for a certain filter key' filter restriction option


CelineSebe:
Tested 👌 
There is an issue with the clear filter button: when clicked, all filters disappear, and we need to refresh the page to restore the default filters.

CelineSebe:
The bug is fixed. It's okay now. 👌

CelineSebe (dismissed):

CelineSebe (approved):

PR #6958, opened 2024-05-14
[frontend] improve large HTML content editing (#6383)

### LINKED ISSUE #6383: System non-responsive (UI nearly unusable) when editing/downloading/saving large HTML files in content editor
## Description

When editing a large HTML file for an Event, the UI updates super slowly as you type. To the level that the UI is non-responsive and unusable. You type something and it takes several seconds for what you type to populate on the page. Additionally, if you try to download the large HTML file as a PDF, the download never happens or takes a long time to start. Saving also has a very long response time.  This seems to happen with HTML files that are larger than 75k-100k. Plain text files are better, but not great

Navigate to the content tab using **_http://your-opencti-url/dashboard/events/incidents/event-UUID/content_**

*** Our groups workflow depends on the ability to edit/save/download/etc. an HTML file.  So a fix for this is greatly appreciated versus an alternate non-HTML notes option.

## Reproducible Steps

1. Download one of the attached TXT files from this issue and rename it to be a HTML file
2. Open the Events dashboard
3. Create a new Event by pressing the + sign
4. Fill in the Name, Incident Type, Severity, etc..
5. Using the **Select Your File** button, attach the large HTML file (more than 100k that was attached here)
6. Press **Create**

Now to recreate.. 
1. Navigate to the newly created Event
2. Navigate to the content tab by appending **/content** to the URL (as shown above)
3. Try the following:
   -  Try typing in the editor window
   -  Try saving the content (lower left icon)
   -  Try downloading the file as PDF or as is (lower right icon)
4. All seem to take a long time to have any UI response.

## Expected Output

The system should be responsive and allow the user to edit/save/download HTML files

## Actual Output

The UI becomes sluggish and essentially non-functional.  Especially when the files are much larger than 100k

## Additional information

None

## Screenshots (optional)

[100013.txt](https://github.com/OpenCTI-Platform/opencti/files/14640136/100013.txt)

frapuks:
### Proposed changes

- Add a content tab for Event/Incident (route already exist)
- Add a paddingRight method to render good padding page
- Dissociate OnChange method (activate saving button) and OnBlur method (update "currentContent" state)

### Related issues

- close #6383 

### Checklist

### Further comments

Steps follow to reproduce :
- Download the attached file in issue page.
- Update extension file to have a html file.
- Create an "Event/Incident" with this html file selected in the "Associated File" field

Before :
- We dont have content tabs to access the page
- It's really slowly to update content
- Contrary to issue reporting, i don't have speed problem during downloading the file or downloading in PDF

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/events/incidents/Root.tsx
+++ b/opencti-platform/opencti-front/src/private/components/events/incidents/Root.tsx
@@ -82,7 +82,7 @@ const RootIncidentComponent = ({ queryRef }) => {
   useSubscription(subConfig);
   const data = usePreloadedQuery(incidentQuery, queryRef);
   const { incident, connectorsForImport, connectorsForExport } = data;
-  const isOverview = location.pathname === `/dashboard/events/incidents/${incident.id}`;
+  const isOverview = location.pathname === `/dashboard/events/incidents/${incident?.id}`;
   const paddingRightValue = () => {
     if (location.pathname.includes(`/dashboard/events/incidents/${incident.id}/knowledge`)) return 200;
     if (location.pathname.includes(`/dashboard/events/incidents/${incident.id}/content`)) return 350;
--- a/opencti-platform/opencti-front/src/private/components/events/incidents/Root.tsx
+++ b/opencti-platform/opencti-front/src/private/components/events/incidents/Root.tsx
@@ -84,8 +84,8 @@ const RootIncidentComponent = ({ queryRef }) => {
   const { incident, connectorsForImport, connectorsForExport } = data;
   const isOverview = location.pathname === `/dashboard/events/incidents/${incident.id}`;
   const paddingRightValue = () => {
-    if (location.pathname.includes(`/knowledge`)) return 200;
-    if (location.pathname.includes(`/content`)) return 350;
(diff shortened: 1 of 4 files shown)


SamuelHassine:
@frapuks Can we also upgrade ckeditor to the last version please?

opencti-platform/opencti-front/src/private/components/events/incidents/Root.tsx
```diff
@@ -83,17 +83,16 @@ const RootIncidentComponent = ({ queryRef }) => {
   const data = usePreloadedQuery(incidentQuery, queryRef);
   const { incident, connectorsForImport, connectorsForExport } = data;
   const isOverview = location.pathname === `/dashboard/events/incidents/${incident?.id}`;
+  const paddingRightValue = () => {
+    if (location.pathname.includes(`/dashboard/events/incidents/${incident.id}/knowledge`)) return 200;
+    if (location.pathname.includes(`/dashboard/events/incidents/${incident.id}/content`)) return 350;
+    return 0;
+  };
```
lndrtrbn, line 90: What I am reading right now 💀😂
Why do we need those excessive paddings ?

lndrtrbn: Because of base app layout

lndrtrbn (approved):

PR #6970, opened 2024-05-14
[frontend] Fix relationships widgets not using the date attribute (#6732)

### LINKED ISSUE #6732: Dashboard time filter not properly taken into account depending on view
## Description

Some views do not take into account the date set in the widget when filtering on time. Let me explain:

When you set up a widget, the "Date attribute" used by default is "created_at". This parameter defines the date taken into account when displaying the data. However, for certain views (like donuts, for example), the date is not taken into account by the time filter. It seems that the time filter takes the "created_at" into account, whatever the configuration.

## Environment

OCTI 6.0.9

## Reproducible Steps

1. Create two identical widget but one in list view and the other in donuts view. Here are the parameters:
    - Perspective Knowledge graph
    - Filters : "Source type: Intrusion set" AND " target type: tool"
    - Parameter : Date attribute = "start_time" and Attribute = "entity"
2. Create a relation between an intrusion set and a tool with "start time = [two days ago]"
3. Set your dashboard to "relative time = Last 7 days" (at the top of the dashboard) -> you should see your newly created relationship on both widgets.
4. Now change the relative time to "Last 24 hours" -> you should no longer see your relationships because the "start_time" is more than 24 hours old. On the list widget, the relationship has disappeared, but not on the donuts.

You can find the dashboard "Test issue 6732" in the testing platform with the two widgets already configured.

labo-flg:
`dateAttribute` prop in relationships widgets was never used, so the corresponding query variable would always fallback to default `created_at`.

### Proposed changes

* remove `dateAttribute` prop, use the dateAttribute in `dataSelection` instead
* many React warnings and errors fixed around dashboard code

### Related issues

* closes #6732

### Checklist


opencti-platform/opencti-front/src/components/ExportButtons.jsx
```diff
@@ -179,28 +178,21 @@ class ExportButtons extends Component {
...
-                >
-                  <ContentCopyOutlined fontSize="small" />
-                </IconButton>
+                <ContentCopyOutlined fontSize="small" color="primary" />
               </ToggleButton>
             </Tooltip>
           )}
           <Tooltip title={t('Export to image')}>
-            <ToggleButton onClick={this.handleOpenImage.bind(this)}>
+            <ToggleButton onClick={this.handleOpenImage.bind(this)} value={'Export-to-image'} >
```
marieflorescontact, line 190: just curious: why is the value this string of characters, why do we add the "-" between the words?

opencti-platform/opencti-front/src/private/components/common/form/AuthorizedMembersField.tsx
```diff
@@ -318,7 +318,7 @@ const AuthorizedMembersField = ({
                   </Field>
                   <IconButton
                     color="secondary"
-                    aria-tag="More"
+                    aria-label="More"
```
marieflorescontact, line 321: is this change to avoid a console error?

opencti-platform/opencti-front/src/private/components/workspaces/WorkspacePopover.jsx
```diff
@@ -122,12 +122,10 @@ const WorkspacePopover = ({ workspace, paginationOptions }) => {
...
-        {workspace.type === 'dashboard' && (
-        <>
-          <MenuItem onClick={handleDashboardDuplication}>{t_i18n('Duplicate')}</MenuItem>
-          <MenuItem onClick={() => handleExportJson(workspace)}>{t_i18n('Export')}</MenuItem>
-        </>
-        )}
+        {workspace.type === 'dashboard' && [
+          <MenuItem key="menu_duplicate" onClick={handleDashboardDuplication}>{t_i18n('Duplicate')}</MenuItem>,
+          <MenuItem key="menu_export" onClick={() => handleExportJson(workspace)}>{t_i18n('Export')}</MenuItem>,
+        ]}
```
marieflorescontact, line 128: is that ok to have component in a list?

opencti-platform/opencti-front/src/private/components/common/form/AuthorizedMembersField.tsx
labo-flg: yes ! `aria-tag` does not exist actually.

opencti-platform/opencti-front/src/private/components/workspaces/WorkspacePopover.jsx
labo-flg: MUI's Menu seem to work this way, judging from the error it threw before:

```
MUI: The Menu component doesn't accept a Fragment as a child.
Consider providing an array instead. 
```

opencti-platform/opencti-front/src/components/ExportButtons.jsx
labo-flg: honestly it could be anything at this point. As far as I understood the doc of mui's `Toggle`, all values of `ToggleButtons` inside a `ToggleButtonGroup` should have a different value, but this can be any string.

I used this pattern because it's how it's done in other values in the component, so just for consistency.

marieflorescontact (approved):
locally tested ok 👌

PR #6988, opened 2024-05-16
[frontend] fix incorrect query used in RuleList (#6873)

### LINKED ISSUE #6873: [technical] incorrect use of stixCoreRelationshipDistribution queries

labo-flg:
### Proposed changes

stixCoreRelationshipsNumber was also wrongly used with relationship_type `stix-relationship` ; we simply use the right query stixRelationshipsNumber instead.

We found no other mention of this query (or used with default relationship_type, so `stix-core-relationship`, which is ok).

Also, stixCoreRelationshipsTimeSeries is never used, so no worry of using it with the wrong relationship_type.

### Related issues

* closes #6873 

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/settings/RulesList.jsx
+++ b/opencti-platform/opencti-front/src/private/components/settings/RulesList.jsx
@@ -288,7 +288,7 @@ const RulesListComponent = ({ relay, data, keyword }) => {
       y: entry.value,
     };
   });
-  const chartDataRelations = data.stixCoreRelationshipsTimeSeries.map(
+  const chartDataRelations = data.stixRelationshipsTimeSeries.map(
     (entry) => {
       const date = new Date(entry.date);
       date.setDate(date.getDate() + 15);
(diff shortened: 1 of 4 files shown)


labo-flg:
@Kedae this PR might replace yours https://github.com/OpenCTI-Platform/opencti/pull/6988

labo-flg:
I'll update this PR to reintroduce the error thrown and according tests in backend, but this will wait next release.
It will then give us enough time to validate properly.

labo-flg:
@Kedae I think we're good to go now 6.1.2 is out ?

Kedae (approved):

PR #6995, opened 2024-05-16
[backend] Test CSV bundler with Indicators with external references  …

### LINKED ISSUE #6287: [CSV Mapper] External reference creation
## Description

When importing a CSV containing external references, they are not created

## Environment

Testing

## Reproducible Steps

1. Create the following mapper:
- Check "My CSV contains headers"
- Indicator (name: A, pattern_type: Default to **stix**, pattern: A main_observable_type: Default IP address)
- External reference: (source name: C)
- Link the external reference to the indicator
image
3. Import the attached CSV using the created mapper

## Expected Output

Correctly imported indicators with embedded external references

## Actual Output

No external references inside the created indicators

## Additional information

[feed_1_mini.csv](https://github.com/OpenCTI-Platform/opencti/files/14525497/feed_1_mini.csv)

The reason behind this is that the URL is mandatory (at least in the creation form) while it's shown as not mandatory in the CSV mapper. When I map a column containing the url to the the attribute, it works fine.
image

Goumies:
Closes #6287

### Proposed changes

* Add test to make sure to spot unexpected import with external references

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/6287

### Checklist

### Further comments
Self-assigned this bug a few months ago.
It is not reproducible anymore, since the 16th of may.

### DIFF
--- a/opencti-platform/opencti-graphql/tests/01-unit/domain/csv-bundler-test.ts
+++ b/opencti-platform/opencti-graphql/tests/01-unit/domain/csv-bundler-test.ts
@@ -1,0 +1,88 @@
+import { describe, it, expect } from 'vitest';
+import { bundleProcess } from '../../../src/parser/csv-bundler';
+import { ADMIN_USER, testContext } from '../../utils/testQuery';
+import type { StixBundle, StixDomainObject } from '../../../src/types/stix-common';
+import {
+  indicatorsWithExternalReferencesCsvContent,
+  indicatorsWithExternalReferencesCsvMapper
+} from '../../data/csv-bundler/external-references-constants';
+import type { CsvMapperParsed } from '../../../src/modules/internal/csvMapper/csvMapper-types';
+
+describe('CSV bundler', () => {
+  describe('Embedded properties', () => {
+    it('Should list external references', async() => {
+      const indicatorsWithExternalReferencesExpectedBundle: StixBundle = {
+        id: "bundle--bfb3d6f4-6961-4fd0-8fb6-afbcbf2e0d59",
+        spec_version: "2.1",
+        type: "bundle",
+        objects: [
+        {
+          id: "indicator--7be2cb5d-ec2b-5bdd-89eb-5802b71faabd",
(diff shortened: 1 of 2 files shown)


opencti-platform/opencti-graphql/tests/01-unit/domain/csv-bundler-test.ts
```diff
@@ -0,0 +1,84 @@
...
+import {
+  indicatorsWithExternalReferencesCsvContent,
+  indicatorsWithExternalReferencesCsvMapper
+} from "../../data/csv-bundler/external-references-constants";
+import type {CsvMapperParsed} from "../../../src/modules/internal/csvMapper/csvMapper-types";
+
+describe('CSV bundler', () => {
+  describe('Embedded properties', () => {
+    it('Should list external references', async() => {
+      const indicatorsWithExternalReferencesExpectedBundle: StixBundle = {
```
SouadHadjiat, line 14: I'm wondering why do we need this whole bundle if we only use it to compare the external references with the result of the csv mapper ? 
It took me a bit of time to understand how the test works, the bundle being an expected result, and the inputs we use (csvmapper & csv input) are defined in external-references-constants.ts

Goumies: This assertion is based on the bug we had before.
This way, we will know the root cause if it occurs again.
Since it was fixed by another evolution, I wanted to document it.

labo-flg: I'm not sure it's the same remark as @SouadHadjiat but in the `expect` clause, you only check the `external_references`. So why this complete bundle object ?

Now, it might make sense to expect a deep equal on the full bundle, if this was not yet covered in the tests before.

Goumies: Sure, let me do that

labo-flg (approved):

Issue #7018, opened 2024-05-19
Improve search to prevent score ordering when specific ordering is requested


richard-julien:
## Use case

Currently the platform always as a _score criteria when a searchterm is specify.
The concept of this improvement is to change this behavior to let the user search and specify its own order criteria.

## Expected behavior.

1. User change the search criteria and validate
2. Order Desc + _score is specified automatically
3. User change the order to creation date => sorting is changed and search term still used.
4. User go back and validate a new term => Order Desc + _score is specified automatically

## Hints
_score ordering must be added in graphql enums of all entities
_score still need to be enforced when search term is used without any explicit ordering

PR #7019, opened 2024-05-19
[backend/frontend] Improve search to prevent score ordering when specific ordering is requested (#7018)

richard-julien:
See #7018


labo-flg (approved):
All tested OK, user orderBy is properly preserved with search terms.

PR #7030, opened 2024-05-20
[frontend/backend] Default dashboard available in popover

### LINKED ISSUE #6843: Be able to select the default dashboard for home
## Use case

Be able to select the default dashboard for home.

Kedae:
Closes #6843


SarahBocognano (approved):

opencti-platform/opencti-front/src/private/components/DashboardSettings.jsx
```diff
@@ -42,6 +42,8 @@ const useStyles = makeStyles(() => createStyles({
   }),
 }));
 
+export const DEFAULT_DASHBOARD = 'cf093b57-713f-404b-a210-a1c5c8cb3791';
```
SouadHadjiat: what is this id ?

opencti-platform/opencti-front/lang/front/fr.json
```diff
@@ -528,6 +528,7 @@
   "dashboard": "Tableau de bord",
   "Dashboard settings": "Paramètres du tableau de bord",
   "Dashboards": "Tableaux de bord",
+  "Dashboards from your groups & organizations": "Tableaux de bord de vos groupes & ; organisations",
```
SouadHadjiat: why "& ;" in this translation?

Kedae: Ho, didn't see that, deepl doesn't like & it seems :D I will update the translations

opencti-platform/opencti-front/src/private/components/DashboardSettings.jsx
Kedae: In order to be able to select the platform default dashboard within UI I added a hardcoded id representing this dashboard.

SouadHadjiat:
What is the difference between "Platform dashboard" and "Default dashboard" ? I see both locally, and they display the same dashboard (the default one)

image

Kedae:
> What is the difference between "Platform dashboard" and "Default dashboard" ? I see both locally, and they display the same dashboard (the default one)
> 
> image

As described in the Notion page : Default dashboard is the new name of "Automatic" that chooses the dashboard for you (either or platform or your first recommended dashboard).
Platform dashboard is the default dashboard provided by Filigran within OpenCTI at first launch.

SouadHadjiat:
> > What is the difference between "Platform dashboard" and "Default dashboard" ? I see both locally, and they display the same dashboard (the default one)

> As described in the Notion page : Default dashboard is the new name of "Automatic" that chooses the dashboard for you (either or platform or your first recommended dashboard). Platform dashboard is the default dashboard provided by Filigran within OpenCTI at first launch.

Oh I thought it was the opposite : "default dashboard" being the default opencti dashboard, and "platform dashboard" the one selected by the admin on the group. It's a bit confusing in my opinion.

opencti-platform/opencti-front/src/private/components/DashboardSettings.jsx
SouadHadjiat: so this "DEFAULT_DASHBOARD" is the default of opencti ? but in the UI we named "default dashboard" the custom one selected for the group, and "platform dashboard" is the default. I think it's completely confusing and will be hard to remember that "default dashboard" in the ui doesn't mean "default_dashboard" in the code.

Kedae: I renamed the constant

PR #7033, opened 2024-05-20
[frontend] Fix workbench SDO deduplication (#6797)

### LINKED ISSUE #6797: Entities disappearing when editing container on workbench
## Description

When, inside an analyst workbench, I edit a report, some entities get deleted from the report.

## Environment

Testing

## Reproducible Steps

1. Import the following STIX bundle in the platform: [sample_issue.json](https://github.com/OpenCTI-Platform/opencti/files/15082601/sample_issue.json)
2. Edit the generated workbench
3. Go to the Containers section
4. Click on the report to edit it and then click on **Update and Complete** to edit its contained entities.

## Expected Output

See the two included indicators

## Actual Output

One of the two indicators got deleted from the workbench. Before editing the report, there was two indicators in Entities section

## Screenshots (optional)
issue_import_contained_indicators

SouadHadjiat:
### Proposed changes

* fix stix domain objects update deduplication : adding missing fields (pattern for indicators).

### Related issues
Issue was happening with indicators because we were deduplicating them based only on name (and we didn't have a name, so the two entities were considered the same).

* #6797 

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/common/files/workbench/WorkbenchFileContent.jsx
+++ b/opencti-platform/opencti-front/src/private/components/common/files/workbench/WorkbenchFileContent.jsx
@@ -243,6 +243,8 @@ const importValidation = (t) => Yup.object().shape({
   connector_id: Yup.string().trim().required(t('This field is required')),
 });
 
+const uniqStixDomainObjectsFields = ['name', 'type', 'pattern', 'identity_class', 'x_opencti_location_type'];
+
 const WorkbenchFileContentComponent = ({
   connectorsImport,
   file,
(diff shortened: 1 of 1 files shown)


opencti-platform/opencti-front/src/private/components/common/files/workbench/WorkbenchFileContent.jsx
```diff
@@ -243,6 +243,8 @@ const importValidation = (t) => Yup.object().shape({
   connector_id: Yup.string().trim().required(t('This field is required')),
 });
 
+const uniqStixDomainObjectsFields = ['name', 'type', 'pattern', 'identity_class', 'x_opencti_location_type'];
```
SarahBocognano, line 246: If I understand well, you added the "pattern" field for indicator. Does "identity_class" & "x_opencti_location_type" are the only other field types possible in the platform ? (with name & type of course)

SarahBocognano (approved):

opencti-platform/opencti-front/src/private/components/common/files/workbench/WorkbenchFileContent.jsx
SouadHadjiat: I'm not sure if there are other types, I only gathered the ones I found in the code, they were already used, and I added pattern for indicators.

PR #7059, opened 2024-05-21
[backend/frontend] Add an option in user to force stateless session (#7058)

### LINKED ISSUE #7058: [backend/frontend] Add an option in user to force stateless session
## Use case

Some external client using the api or the taxii collections are sometimes not able to manage correctly http sessions.
In this situation the client is constantly creating a new session at each query, leading to sessions explosion.

## Solution
Add an extra boolean in user configuration to specify a user as "stateless".
It will be an advanced option to use when its really needed.

## How to test
- Create a new user USER01
- Edit the user to select the advanced option "Use stateless mode"
- Use USER01 to do graphQL queries with token authentication with a browser in private mode
- Verify with an admin User that no session have been created for USER01

# Impact
Using this option will reload the user at every call. This will increase the query pressure.
Another ticket to improve the user loading could be interesting to plan

richard-julien:
close #7058


opencti-platform/opencti-front/src/private/components/settings/users/edition/UserEditionOverview.tsx
```diff
@@ -317,6 +325,27 @@ UserEditionOverviewComponentProps
             onFocus={handleChangeFocus}
             onChange={handleSubmitField}
           />
+          <div style={{ marginTop: 20 }}>
+            <Accordion expanded={openOptions} onChange={() => setOpenOptions(!openOptions)}>
```
labo-flg, line 329: you do not need to control the expanded state of the Accordion if there is no additional thing to do on change.

```suggestion
            <Accordion>
```

opencti-platform/opencti-front/src/private/components/settings/users/edition/UserEditionOverview.tsx
```diff
@@ -100,6 +106,7 @@ UserEditionOverviewComponentProps
   const [commitFieldPatch] = useApiMutation(userMutationFieldPatch);
   const [commitOrganizationAdd] = useApiMutation(userMutationOrganizationAdd);
   const [commitOrganizationDelete] = useApiMutation(userMutationOrganizationDelete);
+  const [openOptions, setOpenOptions] = useState(user.stateless_session);
```
labo-flg, line 109: Not needed
```suggestion
```

labo-flg (approved):
Tested OK, looks good to me.

Issue #6704, opened 2024-04-16
Test E2E of dashboards


nino-filigran:
## Use case

Implement test E2E on dashboards.

PR #7075, opened 2024-05-22
[frontend] E2E Dashboards - add data for testing purpose (#6704)

lndrtrbn:
### Proposed changes

* Add data needed for e2e dashboard tests in the file used to populate platform
* Add possibility to insert non stix data (users, groups, etc)

### Related issues

* #6704 

### Checklist


opencti-platform/opencti-front/playwright.config.ts
```diff
@@ -51,17 +51,30 @@ export default defineConfig({
...
+    },
+    {
+      name: 'init data',
+      testMatch: "init.data.ts",
+      use: {
+        ...devices['Desktop Chrome'],
+        storageState: 'tests_e2e/.setup/.auth/user.json',
+      },
+      dependencies: ['setup'],
+    },
```
lndrtrbn, line 66: For info: add a new project to be able to run the test "Init data" after "Authenticate" but before all other tests

opencti-platform/opencti-graphql/script/script-insert-dataset.js
```diff
@@ -24,6 +25,11 @@ const scriptInsertDataset = async () => {
     // Insert dataset
     await execChildPython(executeContext, ADMIN_USER, PYTHON_PATH, 'local_importer.py', sample1);
     await execChildPython(executeContext, ADMIN_USER, PYTHON_PATH, 'local_importer.py', sample2);
+    const isE2e = process.argv.some((val) => val === '--context=e2e');
```
labo-flg: why not use `if(process.env.ELASTICSEARCH__INDEX_PREFIX === 'e2e')` ?

labo-flg: not ideal too I reckon... but I'm not fond of this solution.
Could we hide all required config under one environment variable or option ?

Like specifiy as arguments of the script the name of all datasets you want to insert ; with the right list in the package.json scripts depending on the case. This way it's explicit.

```
"insert:dev": "node build/script-insert-dataset.js --datasets=default,poisonivy",
"insert:dev:e2e": "ELASTICSEARCH__INDEX_PREFIX=e2e node build/script-insert-dataset.js --datasets=default,poisonivy,e2e",
```

lndrtrbn: Yep I was not a fan of the fact to use the name of elastic prefix to determine if a dataset should be inserted or not. But your solution with a dataset option looks good to me!

Do you think it is better to always set the option or we can set the fact that not passing the option is equivalent to `--datasets=default,poisonivy`?

labo-flg: This is a good question. I prefer having explicit list, so it's clear at a quick look.

Now, if the JS script is called elsewhere directly and not through yarn script, we will break how it's supposed to work.
We cannot be sure actually... So better fallback to the default+poisonivy so it won't break anyone's custom stuff.

I'll still type `--datasets=default,poisonivy` for script `insert:dev` in package.json, even if redundant, so when you look at the script command you know what's inside.

opencti-platform/opencti-graphql/package.json
```diff
@@ -19,8 +19,8 @@
     "start:light": "NODE_ENV=dev node build/back.js",
     "start:cluster": "NODE_ENV=cluster node build/back.js",
     "serv": "node build/back.js",
-    "insert:dev": "node build/script-insert-dataset.js",
-    "insert:dev:e2e": "ELASTICSEARCH__INDEX_PREFIX=e2e node build/script-insert-dataset.js",
+    "insert:dev": "node build/script-insert-dataset.js --datasets=DATA-TEST-STIX2_v2,poisonivy",
+    "insert:dev:e2e": "ELASTICSEARCH__INDEX_PREFIX=e2e node build/script-insert-dataset.js --datasets=DATA-TEST-STIX2_v2,data-test-stix-e2e",
```
labo-flg, line 23: no poisonivy?

labo-flg (approved):

PR #7107, opened 2024-05-23
[frontend/backend] Max shareable markings at the group level for exports and public dashboard (#7006)

### LINKED ISSUE #7006: Max shareable marking definitions in groups (for export and public dashboard)
## Use case

Current situation: 

1. We have the ability to define a max marking definition when exporting an entity or a list (one shot)
2. We have the ability to restrict marking definition at the platform level for public dashboards (settings)

What we need:

1. We need to add a settings in the group named "Max shareable marking".
2. We need to remove the marking definition restriction at the platform level

=> Use the new Max shareable in marking in for public dashboards and also to restrict the list of available max marking when exporting.

**By default, when the max shareable marking is not set, the default behavior applies, which means all group markings are available for export and public dashboard**.

**Some screens**

In the parameters of a group:

image

=> Add a new parameter here, "Max shareable marking"

In export:

image

=> This list should be restricted until the max shareable marking

In public dashboards:

image

=> This list should be restricted until the max shareable marking

In Security Policies:

image

=> This settings should be completely removed and not used anymore.

Archidoit:
issue: https://github.com/OpenCTI-Platform/opencti/issues/7006

PR in docs : https://github.com/OpenCTI-Platform/docs/pull/175


opencti-platform/opencti-graphql/src/domain/user.js
```diff
@@ -297,21 +299,49 @@ export const computeAvailableMarkings = (userMarkings, allMarkings) => {
...
 
+// Return all the available markings a user can share
+export const getAvailableDataSharingMarkings = async (context, user) => {
+  const maxMarkings = user.max_shareable_marking;
+  const allMarkings = await getEntitiesListFromCache(context, SYSTEM_USER, ENTITY_TYPE_MARKING_DEFINITION);
+  return computeAvailableMarkings(maxMarkings, allMarkings);
+};
+
+export const checkUserCanShareMarkings = async (context, user, markingsToShare) => {
+  const shareableMarkings = await getAvailableDataSharingMarkings(context, user);
```
lndrtrbn, line 310: I'm not sure we need to get the complete list of available markings here, I have the impression that it would work with only max markings?

Archidoit: indeed you're right ! fixed

lndrtrbn (approved):

PR #7115, opened 2024-05-24
[backend] do not check confidence of indicator when creating observable from it (#7112)

### LINKED ISSUE #7112: Throwing User effective max confidence level is insufficient to update this element at upsert / ingestion
## Description

image

I'm very worried about this error. Can someone checked that this error does not prevent upserting the "other STIX IDs" of this element and that we actually continue to cumulate IDs despite of this error?

labo-flg:
### Proposed changes

* Remove confidence check when creating observable from indicator

### Related issues
* closes #7112 

### Checklist

### Further comments
 
 Creating an observable from indicator should not be under confidence check over the indicator. We are basically creating an observable + relationship between the 2 entities. This is not a ref, this is a stix-core-relationship. This is simply knowledge creation.


SouadHadjiat (approved):
tested locally : observables / indicators created even if confidence level is lower than the entity.

PR #7130, opened 2024-05-27
[backend] fix user capabilities check for  notification (#7109)

### LINKED ISSUE #7109: [Notification] Bulk notification manipulation impossible
## Description

When I try to manipulate (e.g. mark as read, or delete) lots of notifications I received in my interface in the same time, I get an error: "You are not allowed to do this"

## Environment

OCTI 6.1.3

## Reproducible Steps

1. Have a full access user except the "bypass all capabilities" capability, i.e.
    - All capabilities (except "bypass all capabilities"),
    - Access to all marking
    - Maw confidence level = 100
    - In the platform organization if any
-> The goal is that it is not supposed to have any restriction in terms of data manipulation (although a user's notifications have no reason to be restricted anyway).
2. Generated a lot (1000+) notifications in the user interface (tips: make a trigger on the indicator modifications then launch a background task which updates 1000+ indicators).
3. When you have all your notifications, do "select all" then try to mark all of them as "read" (bottom right button)
-> error

Screenshot 2024-05-23 234936

#### Web browser network

image

#### Platform logs

image

## Additional information

- I know that in the past we had bugs that came from the "select all" button, maybe it's still that
- With the capability "bypass all capabilities", I don't have the error

marieflorescontact:
### Proposed changes

* We were checking for SETTINGS_SETACCESSES value in `userCapabilities`, which is an array of split capabilities (=>[ 'SETTINGS, 'SETACCESSES]

### Related issues

* [issue/7109](https://github.com/OpenCTI-Platform/opencti/issues/7109)

### Checklist


lndrtrbn (commented):
I do not understand the explanation of the issue. Even with the array [SETTINGS, SETACCESSES] it should work as SETACCESSES is a child of SETTINGS so if you have it you also have the parent?

marieflorescontact:
> I do not understand the explanation of the issue. Even with the array [SETTINGS, SETACCESSES] it should work as SETACCESSES is a child of SETTINGS so if you have it you also have the parent?

as userCapabilities is an array of capabilities split on the "_" (line 37), the condition `userCapabilities.includes(SETTINGS_SET_ACCESSES)` was never true.

lndrtrbn:
> as userCapabilities is an array of capabilities split on the "_" (line 37), the condition `userCapabilities.includes(SETTINGS_SET_ACCESSES)` was never true.

Ah, yes ok. Still feels strange we need to add a new constant to resolve the issue. Do we not have a function somewhere that already checks capabilities instead of calling directly `userCapabilities.includes`?

SarahBocognano (approved):

lndrtrbn (changes requested):
Use isUserHasCapability instead?

marieflorescontact:
> Use isUserHasCapability instead?

the use of `isUserHasCapability` requires a more global rework on the capa check. i created a technical issue #7184 because while investigating with @labo-flg we found  2 other places where the check was badly done (in notes and opinions). So for now the bug is fixed, the check is done on a string hardcoded, I've put a TODO for the next step.

lndrtrbn:
> > Use isUserHasCapability instead?
> 
> the use of `isUserHasCapability` requires a more global rework on the capa check. i created a technical issue #7184 because while investigating with @labo-flg we found 2 other places where the check was badly done (in notes and opinions). So for now the bug is fixed, the check is done on a string hardcoded, I've put a TODO for the next step.

Even if we are not taking care of the global rework in this PR we can still at least use the function correctly for the 2 lines triggering the bug instead of adding hard coded strings no?

lndrtrbn (approved):

PR #7134, opened 2024-05-27
[frontend/backend] Ability to add indicator generated from the observables of a container in the container (#6171)

### LINKED ISSUE #6171: Ability to add indicator generated from the observables of a container in the container
## Use case

Currently, in the context of a container, when creating an indicator from an observable, the indicator is not included in the report. We should be able to automatically add it to the report at its creation?

ValentinBouzinFiligran:
### Proposed changes

* checkbox added to add the created entity to the container
* containerId passed to the background task
* if the containerId is present, the created entity will be add to the container
* promoteObservableToIndicator now return the created indicator
* on observable page, when we generate or delete an inficator, we update the relay store directly

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/6171
*

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/lang/back/de.json
+++ b/opencti-platform/opencti-front/lang/back/de.json
@@ -19,6 +19,8 @@
   "Actions": "Aktionen",
   "Active": "Aktiv",
   "Activity listeners IDs": "IDs von Aktivitätshörern",
+  "Add in generated & existing indicators in the container": "Hinzufügen von generierten & vorhandenen Indikatoren im Container",
+  "Add in generated & existing observables in the container": "Hinzufügen von generierten & vorhandenen Observablen im Container",
   "Added after": "Hinzugefügt nach",
   "Added_after_start": "Hinzugefügt_nach_Start",
   "Additional names": "Zusätzliche Namen",
--- a/opencti-platform/opencti-front/lang/back/de.json
+++ b/opencti-platform/opencti-front/lang/back/de.json
@@ -19,8 +19,6 @@
   "Actions": "Aktionen",
   "Active": "Aktiv",
   "Activity listeners IDs": "IDs von Aktivitätshörern",
-  "Add in generated & existing indicators in the container": "Hinzufügen von generierten & vorhandenen Indikatoren im Container",
-  "Add in generated & existing observables in the container": "Hinzufügen von generierten & vorhandenen Observablen im Container",
   "Added after": "Hinzugefügt nach",
   "Added_after_start": "Hinzugefügt_nach_Start",
   "Additional names": "Zusätzliche Namen",
--- a/opencti-platform/opencti-front/lang/back/fr.json
+++ b/opencti-platform/opencti-front/lang/back/fr.json
@@ -19,6 +19,8 @@
   "Actions": "Actions",
   "Active": "Actif",
   "Activity listeners IDs": "ID des auditeurs d'activité",
+  "Add in generated & existing indicators in the container": "Ajouter les indicateurs générés & ; existants dans le conteneur",
+  "Add in generated & existing observables in the container": "Ajouter des observables générés & ; existants dans le conteneur",
   "Added after": "Ajouté après",
   "Added_after_start": "Ajouté_après_début",
   "Additional names": "Noms supplémentaires",
--- a/opencti-platform/opencti-front/lang/back/fr.json
+++ b/opencti-platform/opencti-front/lang/back/fr.json
@@ -19,8 +19,6 @@
   "Actions": "Actions",
   "Active": "Actif",
   "Activity listeners IDs": "ID des auditeurs d'activité",
-  "Add in generated & existing indicators in the container": "Ajouter les indicateurs générés & ; existants dans le conteneur",
-  "Add in generated & existing observables in the container": "Ajouter des observables générés & ; existants dans le conteneur",
   "Added after": "Ajouté après",
   "Added_after_start": "Ajouté_après_début",
   "Additional names": "Noms supplémentaires",
--- a/opencti-platform/opencti-front/src/private/components/observations/stix_cyber_observables/StixCyberObservableIndicatorPopover.jsx
+++ b/opencti-platform/opencti-front/src/private/components/observations/stix_cyber_observables/StixCyberObservableIndicatorPopover.jsx
@@ -15,6 +15,7 @@ import Slide from '@mui/material/Slide';
 import MoreVert from '@mui/icons-material/MoreVert';
 import inject18n from '../../../../components/i18n';
 import { commitMutation } from '../../../../relay/environment';
+import { deleteNodeFromId } from '../../../../utils/store';
 
 const styles = (theme) => ({
   container: {
@@ -89,11 +90,17 @@ class StixCyberObservableIndicatorPopover extends Component {
         toId: this.props.observableId,
         relationship_type: 'based-on',
       },
+      updater: (store) => {
+        deleteNodeFromId(
+          store,
+          this.props.observableId,
+          'Pagination_stixCyberObservables_indicators',
+          {},
+          this.props.indicatorId,
+        );
+      },
       onCompleted: () => {
         this.handleCloseDelete();
-        if (this.props.onDelete) {
-          this.props.onDelete();
-        }
       },
     });
   }
@@ -103,6 +110,7 @@ class StixCyberObservableIndicatorPopover extends Component {
     return (
       <div className={classes.container}>
         <IconButton
+          data-testid="stixCyberObserbleIndicatorItemPopoverButton"
           onClick={this.handleOpen.bind(this)}
           aria-haspopup="true"
           style={{ marginTop: 3 }}
@@ -161,7 +169,6 @@ StixCyberObservableIndicatorPopover.propTypes = {
   paginationOptions: PropTypes.object,
   classes: PropTypes.object,
   t: PropTypes.func,
-  onDelete: PropTypes.func,
 };
 
 export default compose(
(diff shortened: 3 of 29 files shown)


opencti-platform/opencti-front/lang/back/de.json
```diff
@@ -19,6 +19,8 @@
   "Actions": "Aktionen",
   "Active": "Aktiv",
   "Activity listeners IDs": "IDs von Aktivitätshörern",
+  "Add in generated & existing indicators in the container": "Hinzufügen von generierten & vorhandenen Indikatoren im Container",
+  "Add in generated & existing observables in the container": "Hinzufügen von generierten & vorhandenen Observablen im Container",
```
SouadHadjiat: "add in ... in container" ? is it the right message ? not "add ... in the container" ? or "add ... to container" ? to check with product

opencti-platform/opencti-front/lang/back/fr.json
```diff
@@ -19,6 +19,8 @@
   "Actions": "Actions",
   "Active": "Actif",
   "Activity listeners IDs": "ID des auditeurs d'activité",
+  "Add in generated & existing indicators in the container": "Ajouter les indicateurs générés & ; existants dans le conteneur",
```
SouadHadjiat: automatic translation is bad with &

opencti-platform/opencti-graphql/config/schema/opencti.graphql
```diff
@@ -12594,7 +12595,7 @@ type StixCyberObservableEditMutations {
   relationAdd(input: StixRefRelationshipAddInput!): StixRefRelationship @auth(for: [KNOWLEDGE_KNUPDATE])
   relationsAdd(input: StixRefRelationshipsAddInput!): StixCyberObservable @auth(for: [KNOWLEDGE_KNUPDATE])
   relationDelete(toId: StixRef!, relationship_type: String!): StixCyberObservable @auth(for: [KNOWLEDGE_KNUPDATE])
-  promote: StixCyberObservable @auth(for: [KNOWLEDGE_KNUPDATE])
+  promote: Indicator! @auth(for: [KNOWLEDGE_KNUPDATE])
```
SouadHadjiat: is Indicator always returned ? can't it be null or undefined ?

opencti-platform/opencti-graphql/src/modules/indicator/indicator-domain.ts
```diff
@@ -197,6 +197,7 @@ export const createObservablesFromIndicator = async (
       return createRelation(context, user, relationInput);
     })
   );
+  return observablesToLink;
```
SouadHadjiat, line 199: so we return a list with observables ids ?

SouadHadjiat: I ask the question since `promoteObservableToIndicator` returns an indicator objects, and here it seems that `promoteIndicatorToObservable` which calls `createObservablesFromIndicator` returns observables ids, not the objects.

opencti-platform/opencti-graphql/tests/03-streams/00-Raw/raw-test.js
```diff
@@ -25,9 +25,9 @@ describe('Raw streams tests', () => {
       expect(createEventsByTypes.label.length).toBe(15);
       expect(createEventsByTypes.identity.length).toBe(31);
       expect(createEventsByTypes.location.length).toBe(15);
-      expect(createEventsByTypes.relationship.length).toBe(132);
+      expect(createEventsByTypes.relationship.length).toBe(136);
```
SarahBocognano, line 28: Are you sure we need to change de raw numbers here ? Where this number is coming from ?
I thought we said to delete everything after your test will be run to avoid changing numbers. Is this because it's the manager and his behaviour is different ?

opencti-platform/opencti-front/src/private/components/observations/stix_cyber_observables/StixCyberObservableIndicators.jsx
```diff
@@ -193,12 +205,10 @@ class StixCyberObservableIndicatorsComponent extends Component {
...
         <div className="clearfix" />
-        <List style={{ marginTop: -15 }}>
-          {filter(
-            (n) => !includes(n.node.id, this.state.deleted),
-            stixCyberObservable.indicators.edges,
-          ).map((indicatorEdge) => (
+        <List style={{ marginTop: -15 }} data-testid='stixCyberObservableIndicatorsList'>
+          {stixCyberObservable.indicators.edges.map((indicatorEdge) => (
             <ListItem
+              data-testid={`stixCyberObservableIndicatorsItem-${indicatorEdge.node.id}`}
```
SarahBocognano: same

opencti-platform/opencti-front/src/private/components/observations/stix_cyber_observables/StixCyberObservableIndicators.jsx
```diff
@@ -171,6 +182,7 @@ class StixCyberObservableIndicatorsComponent extends Component {
         </Typography>
         <Security needs={[KNOWLEDGE_KNUPDATE]}>
           <IconButton
+            data-testid='AddOrCreateIndicatorsButton'
```
SarahBocognano: I don't see the related e2e test in this PR.

opencti-platform/opencti-front/src/private/components/observations/stix_cyber_observables/StixCyberObservableIndicators.jsx
```diff
@@ -193,12 +205,10 @@ class StixCyberObservableIndicatorsComponent extends Component {
           </Menu>
         </Security>
         <div className="clearfix" />
-        <List style={{ marginTop: -15 }}>
-          {filter(
-            (n) => !includes(n.node.id, this.state.deleted),
-            stixCyberObservable.indicators.edges,
-          ).map((indicatorEdge) => (
+        <List style={{ marginTop: -15 }} data-testid='stixCyberObservableIndicatorsList'>
```
SarahBocognano: same

opencti-platform/opencti-front/src/private/components/observations/stix_cyber_observables/StixCyberObservableIndicators.jsx
```diff
@@ -112,15 +116,10 @@ class StixCyberObservableIndicatorsComponent extends Component {
       displayPromoteStix: false,
       displayCreate: false,
       promotingStix: false,
-      deleted: [],
     };
     this.indicatorParams = { first: 200 };
   }
 
-  onDelete(id) {
```
SarahBocognano, line 120: Possible to remove the class component and refactor it to a pure function/typescript one ?

opencti-platform/opencti-front/src/private/components/observations/stix_cyber_observables/StixCyberObservableIndicatorPopover.jsx
```diff
@@ -103,6 +110,7 @@ class StixCyberObservableIndicatorPopover extends Component {
     return (
       <div className={classes.container}>
         <IconButton
+          data-testid="stixCyberObserbleIndicatorItemPopoverButton"
```
SarahBocognano: I don't see the related e2e test in this PR.

opencti-platform/opencti-graphql/tests/03-streams/00-Raw/raw-test.js
ValentinBouzinFiligran: after the test, all created elements are deleted, but creation events are taken into account

ValentinBouzinFiligran:
I've added data-testid to make it easier to find components in the code.
This could be useful if we need to work on these components again.

lndrtrbn:
> I've added data-testid to make it easier to find components in the code. This could be useful if we need to work on these components again.

Why do you need data-testid? As far as possible we want to not use them to fetch elements during testing

lndrtrbn (approved):

SarahBocognano (approved):
Tested ✅

opencti-platform/opencti-graphql/src/domain/stixCyberObservable.js
```diff
@@ -143,13 +143,13 @@ const createIndicatorFromObservable = async (context, user, input, observable) =
         externalReferences: input.externalReferences,
         update: true,
       };
-      await addIndicator(context, user, indicatorToCreate);
-    } else {
-      logApp.warn('Cannot create indicator - cant generate pattern', { key, value });
+      return addIndicator(context, user, indicatorToCreate);
```
SouadHadjiat: what happens if addIndicator throws an error here ? I guess it won't be catched here, and it will throw an error where it is called, but then since it's called at the end of addStixCyberObservable, it might fail the observable creation even though the observable has been created.

opencti-platform/opencti-graphql/tests/03-streams/00-Raw/raw-test.js
```diff
@@ -42,11 +42,11 @@ describe('Raw streams tests', () => {
       expect(createEventsByTypes.file.length).toBe(4);
       expect(createEventsByTypes.campaign.length).toBe(5);
       expect(createEventsByTypes.incident.length).toBe(2);
-      expect(createEventsByTypes.report.length).toBe(31);
+      expect(createEventsByTypes.report.length).toBe(33);
```
SouadHadjiat, line 45: why +2 reports? I only see 1 new report created in taskManager-test, where is the other one?

SouadHadjiat (commented):
tested locally, works well ✅

Need to update the client python calling promote API, so that it returns an indicator instead of an observable.

SouadHadjiat (changes requested):
Before merge of this PR, we need to update the client python calling promote API, so that it returns an indicator instead of an observable (stix_cyber_observable.promote_to_indicator)

SouadHadjiat (approved):

Issue #5371, opened 2024-01-04
Have a workbench created automatically from RSS Feed


Jipegien:
## Use case

When importing data from RSS feed, match existing entities présent in its content and add it to the created container.

## Planned Solution
Automatically launch import external reference connector.

richard-julien:
Be careful also about https://github.com/OpenCTI-Platform/opencti/issues/7080

PR #7137, opened 2024-05-27
[backend] Trigger enrichment connectors with filters (#5371)

SouadHadjiat:
### Proposed changes

* Define connector_trigger_filters
* Filter enrichment connectors during entity creation
* Display and update connector trigger filters for internal enrichment connectors

### Related issues

* #5371 

### Checklist

### Further comments

Different ways to test this feature:
- Create an external reference => should trigger the import external reference connector if it matches the filter
- OR try with rss feeds

To test this PR, you need to run import-document and import-external-reference connectors locally. For import-external-reference, you will need to install wkhtmltopdf https://wkhtmltopdf.org/downloads.html and in config.yml set
wkhtmltopdf_path: 'C:\\Program Files\\wkhtmltopdf\\bin\\wkhtmltopdf.exe'

When you test import-document, set your config with auto: true and validate_before_import: true, then make sure the workbench is created on the container of the external-ref.

Capture d'écran 2024-06-07 102859

### DIFF
--- a/opencti-platform/opencti-graphql/src/domain/connector.js
+++ b/opencti-platform/opencti-graphql/src/domain/connector.js
@@ -5,6 +5,8 @@ import { completeConnector, connector, connectors, connectorsFor } from '../data
 import { getConnectorQueueDetails, purgeConnectorQueues, registerConnectorQueues, unregisterConnector, unregisterExchanges } from '../database/rabbitmq';
 import { ENTITY_TYPE_CONNECTOR, ENTITY_TYPE_SYNC, ENTITY_TYPE_WORK } from '../schema/internalObject';
 import { FunctionalError, UnsupportedError, ValidationError } from '../config/errors';
+import { validateFilterGroupForStixMatch } from '../utils/filtering/filtering-stix/stix-filtering';
+import { isFilterGroupNotEmpty } from '../utils/filtering/filtering-utils';
 import { now } from '../utils/format';
 import { elLoadById } from '../database/engine';
 import { INTERNAL_SYNC_QUEUE, isEmptyField, READ_INDEX_HISTORY } from '../database/utils';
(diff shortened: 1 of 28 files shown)


JeremyCloarec:
Just a detail, but maybe the filter box could be moved just a bit down? The 'Add filter' text goes a bit onto the 'Trigger filters' text when we click on the filter box
image

JeremyCloarec:
What's the expected behavior when auto is set to true in import-external-reference? Does it still take into account the trigger filters?

SouadHadjiat:
> What's the expected behavior when auto is set to true in import-external-reference? Does it still take into account the trigger filters?

When trigger filters are set, it overrides the "auto" configuration, it will always trigger automatically the connector based on the filters, whether it was set to manual or automatic.

opencti-platform/opencti-graphql/src/domain/enrichment.js
```diff
@@ -40,3 +46,35 @@ export const createEntityAutoEnrichment = async (context, user, element, scope)
...
+};
+
+export const filterConnectorsForElementEnrichment = async (context, connectors, element, scope) => {
+  // first filter active & enrichment connectors only
+  const activeConnectors = connectors.filter((conn) => conn.active === true && conn.connector_type === CONNECTOR_INTERNAL_ENRICHMENT);
+  const targetConnectors = [];
+  for (let i = 0; i < activeConnectors.length; i += 1) {
+    const conn = activeConnectors[i];
+    const scopeMatch = scope ? (conn.connector_scope ?? []).some((s) => s.toLowerCase() === scope.toLowerCase()) : true;
+    const autoTrigger = conn.auto === true || (conn.connector_trigger_filters && await isStixMatchConnectorFilter(context, element, conn.connector_trigger_filters));
```
JeremyCloarec: shouldn't it be something like 
`conn.connector_trigger_filters ?? await isStixMatchConnectorFilter(context, element, conn.connector_trigger_filters) : conn.auto === true` ?
if we want to override connector auto config?

SouadHadjiat: yes you're right, it's fixed now

SouadHadjiat:
all comments have been fixed, waiting for review & approval

opencti-platform/opencti-graphql/src/domain/enrichment.js
```diff
@@ -40,3 +46,35 @@ export const createEntityAutoEnrichment = async (context, user, element, scope)
...
+};
+
+const isStixMatchConnectorFilter = async (context, element, stringFilters) => {
+  if (!stringFilters) {
+    return true; // no filters, match all
+  }
+  const jsonFilters = JSON.parse(stringFilters);
+  if (!isFilterGroupNotEmpty(jsonFilters)) {
+    return false; // filters empty -> we don't try to match
+  }
```
JeremyCloarec: Is is intended that no filters means matching all, but filters empty means matching nothing?

opencti-platform/opencti-graphql/src/domain/connector.js
```diff
@@ -130,6 +132,41 @@ export const connectorDelete = async (context, user, connectorId) => {
...
+
+export const connectorTriggerUpdate = async (context, user, connectorId, input) => {
+  const conn = await storeLoadById(context, user, connectorId, ENTITY_TYPE_CONNECTOR);
+  if (!conn) {
+    throw FunctionalError('Cant find element to update', { id: connectorId, type: ENTITY_TYPE_CONNECTOR });
+  }
+  if (!['INTERNAL_ENRICHMENT', 'INTERNAL_IMPORT_FILE'].includes(conn.connector_type)) {
+    throw FunctionalError('Update is only possible on internal enrichment or import file connectors types');
+  }
+  const supportedInputKeys = ['connector_trigger_filters', 'auto'];
```
JeremyCloarec: Why is 'auto' a supported input key, since we currently only change the filters on the frontendside? Is it going to be added in the future?

opencti-platform/opencti-graphql/src/domain/connector.js
```diff
@@ -130,6 +132,41 @@ export const connectorDelete = async (context, user, connectorId) => {
...
+  if (filtersItem?.value) {
+    const jsonFilters = JSON.parse(filtersItem.value);
+    if (isFilterGroupNotEmpty(jsonFilters)) {
+      // our stix matching is currently limited, we need to validate the input filters
+      validateFilterGroupForStixMatch(jsonFilters);
+    } else {
+      filtersItem.value = ''; // empty filter
+    }
+  }
+  const { element } = await updateAttribute(context, user, connectorId, ENTITY_TYPE_CONNECTOR, input);
```
JeremyCloarec, line 160: I'm not sure how the cache currently works, but shouldn't resetCacheForEntity be called for connector entity type? During my tests, it looks like the filter isn't taken into account without having to restart the backend

opencti-platform/opencti-graphql/src/domain/enrichment.js
SouadHadjiat: oh right, I changed the logic, and I forgot about this part

opencti-platform/opencti-graphql/src/domain/connector.js
SouadHadjiat: yes I added this in prevision of being able to change the auto param directly in the front, but maybe I could remove it

opencti-platform/opencti-graphql/src/domain/connector.js
SouadHadjiat: the cache works by listening to update events, which are sent here `notify(BUS_TOPICS[ENTITY_TYPE_CONNECTOR].EDIT_TOPIC, element, user);`

opencti-platform/opencti-graphql/src/domain/connector.js
JeremyCloarec: I think it's OK to keep, but I was mostly wondering how it would work with the auto coming from the connector config file? Won't the auto configured with this method be overriden everytime the connector is restarted and registers to the API with its config from config file?

SouadHadjiat: done

opencti-platform/opencti-graphql/src/domain/enrichment.js
SouadHadjiat: done

JeremyCloarec (approved):

Issue #7118, opened 2024-05-24
Add height / weight do not work


SamuelHassine:
## Description

![image](https://github.com/OpenCTI-Platform/opencti/assets/1334279/954afffe-15d0-40ce-94e2-29bda2a7d2bc)

PR #7165, opened 2024-05-29
[frontend] Fix Add height / weight do not work (#7118)

SarahBocognano:
### Proposed changes

* Add a validator check to help the user knowing the field is mandatory
* On edit form, when pushes a new height or weight line, put a value instead of null to avoid errors 

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/7118

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/common/form/HeightField.tsx
+++ b/opencti-platform/opencti-front/src/private/components/common/form/HeightField.tsx
@@ -15,6 +15,7 @@ import TextField from '../../../../components/TextField';
 import { commitMutation, defaultCommitMutation } from '../../../../relay/environment';
 import useUserMetric from '../../../../utils/hooks/useUserMetric';
 import DateTimePickerField from '../../../../components/DateTimePickerField';
+import { isNotEmptyField } from '../../../../utils/utils';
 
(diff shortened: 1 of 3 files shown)


Archidoit:
If we enter something else than a date for the height/weight, the error is not well formatted:
image


You can include this part of code in your dateSeenValidator: 
image

Archidoit:
What is the meaning for the (') unit (simple curiosity)? (I know 'lb' for weight but I don't know (') )

image


@SarahBocognano : this is how we define height in America everytime you'll see a height, I will be like:  5'7" (172cm)

Archidoit:
Usually when a required field is not specified, there is only a red message in the form : 
image


But for height/weight date or measure, a second message appears at the top of the screen "Validation against the schema... " :

image

image

CelineSebe:
When adding a height, after delete the previous date, this date is reinitialized automatically.

@SarahBocognano : Yes I did that on purpose to avoid any error for the user and to show that this field is indeed mandatory and has no sense of use without the date

lndrtrbn (approved):
I can create TAI with null measure or date for height and weight, I suppose not intended but at least when setting real values it works well

PR #7173, opened 2024-05-29
[worker] fix opentelemetry worker dependencies

Archidoit:
fix the running of 'pip install -r requirements.txt' in worker/src
--> update opentelemetry worker dependencies to be consistant with client-python and opencti (version 1.24)

### DIFF
--- a/opencti-worker/src/requirements.txt
+++ b/opencti-worker/src/requirements.txt
@@ -1,4 +1,4 @@
 pycti==6.1.5
-opentelemetry-api==1.22.0
-opentelemetry-sdk==1.22.0
+opentelemetry-api<=1.24.0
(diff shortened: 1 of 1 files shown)


SamuelHassine:
Sorry Cathia, just pushed the update.

Archidoit:
@SamuelHassine your update doesn't seem to work : 
image

helene-nguyen:
I tried to add this to take into account the both versions

image


But still have a conflict

image


I think this solution can be good, as it takes into account even the 1.22.0 until 1.24.0

```
opentelemetry-api<=1.24.0
opentelemetry-sdk<=1.24.0
```

Archidoit:
I think it's better to use the following code because it's more restrictive : 
image

Issue #6878, opened 2024-05-03
Override of confidence level per Entity per Group


nino-filigran:
## Use case

In 6.1 we introduced the override of confidence per entity at user level, but not at group level. This feature aims to implement it at Group Level.

PR #7186, opened 2024-05-30
[frontend/backend] show effective max confidence with creators (#6878)

labo-flg:
### Proposed changes

* update `creators` API to be able to query `effective_confidence_level` in case of user creator, or `group_confidence_level` in case of group creator
* update CreatorField to show the confidence level of the user in the Autocomplete box

### Related issues

#6878

### Checklist

image

### DIFF
--- a/opencti-platform/opencti-front/lang/front/fr.json
+++ b/opencti-platform/opencti-front/lang/front/fr.json
@@ -1,7 +1,7 @@
 {
   " (reversed)": "(inversé)",
   " & ": " & ",
-  "+ N override(s)": "+ {count}} surcharge(s)",
+  "+ N override(s)": "+ {count} surcharge(s)",
   "0 equals no maximum": "0 pour aucun maximum",
   "24 hours": "24 heures",
   "24h": "24h",
--- a/opencti-platform/opencti-front/lang/front/fr.json
+++ b/opencti-platform/opencti-front/lang/front/fr.json
@@ -1,6 +1,7 @@
 {
   " (reversed)": "(inversé)",
   " & ": " & ",
+  "+ N override(s)": "+ {count}} surcharge(s)",
   "0 equals no maximum": "0 pour aucun maximum",
   "24 hours": "24 heures",
   "24h": "24h",
--- a/opencti-platform/opencti-front/src/private/components/common/form/CreatorField.tsx
+++ b/opencti-platform/opencti-front/src/private/components/common/form/CreatorField.tsx
@@ -1,7 +1,7 @@
 import React, { FunctionComponent, useState } from 'react';
 import { Field } from 'formik';
-import makeStyles from '@mui/styles/makeStyles';
 import { graphql } from 'react-relay';
+import Box from '@mui/material/Box';
 import { fetchQuery } from '../../../../relay/environment';
 import AutocompleteField from '../../../../components/AutocompleteField';
 import { useFormatter } from '../../../../components/i18n';
@@ -9,30 +9,12 @@ import { CreatorFieldSearchQuery$data } from './__generated__/CreatorFieldSearch
 import ItemIcon from '../../../../components/ItemIcon';
 import { Option } from './ReferenceField';
 
-// Deprecated - https://mui.com/system/styles/basics/
-// Do not use it for new code.
-const useStyles = makeStyles(() => ({
-  icon: {
-    paddingTop: 4,
-    display: 'inline-block',
-  },
-  text: {
-    display: 'inline-block',
-    flexGrow: 1,
-    marginLeft: 10,
-  },
-  autoCompleteIndicator: {
-    display: 'none',
-  },
-}));
-
 interface CreatorFieldProps {
   name: string;
   label: string;
-  isOptionEqualToValue?: (option: Option, value: string) => boolean;
   onChange?: (name: string, value: Option) => void;
   containerStyle?: Record<string, string | number>;
-  helpertext?: string;
+  showConfidence?: boolean;
 }
 
 const CreatorFieldQuery = graphql`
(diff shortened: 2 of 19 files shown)


opencti-platform/opencti-front/src/private/components/common/form/CreatorField.tsx
```diff
@@ -74,49 +89,73 @@ const CreatorField: FunctionComponent<CreatorFieldProps> = ({
...
           onFocus: searchCreators,
         }}
+        disableClearable
         onChange={onChange}
         style={containerStyle}
         noOptionsText={t_i18n('No available options')}
-        options={creators}
-        isOptionEqualToValue={isOptionEqualToValue}
+        options={creatorOptions}
+        isOptionEqualToValue={(option: CreatorOption, selected: CreatorOption) => option.value === selected.value}
```
labo-flg, line 131: Having `isOptionEqualToValue` as an external props is unnecessary, as it's always the same in all cases.
It was also badly typed and used, and resulted in react  warnings all around.

Now it's the right signature.

opencti-platform/opencti-front/src/private/components/common/form/CreatorField.tsx
```diff
@@ -74,49 +89,73 @@ const CreatorField: FunctionComponent<CreatorFieldProps> = ({
...
+                  textAlign: 'right',
+                  color: 'text.disabled',
+                }}
+              >
+                {option.extra}
+              </Box>
+            )}
           </li>
         )}
-        classes={{ clearIndicator: classes.autoCompleteIndicator }}
```
labo-flg, line 119: this can be replaced by MUI's Autocomplete prop `disableClearable` (a global refactor could be done on this, I'll do it in another PR as it would add too much noise here.)

opencti-platform/opencti-front/src/private/components/data/ingestionCsv/IngestionCsvCreation.tsx
```diff
@@ -286,11 +286,9 @@ const IngestionCsvCreation: FunctionComponent<IngestionCsvCreationProps> = ({ pa
               )}
               <CreatorField
                 name="user_id"
-                label={t_i18n(
-                  'User responsible for data creation (empty = System)',
-                )}
-                isOptionEqualToValue={(option: Option, value: string) => option.value === value}
+                label={t_i18n('User responsible for data creation (empty = System)')}
                 containerStyle={fieldSpacingContainerStyle}
+                showConfidence
```
labo-flg, line 291: I went for an opt-in optional display, even if in most cases we want it displayed.

I'm generally in favor of explicit opt-in options for clarity ; but happy to discuss if you think otherwise.

opencti-platform/opencti-graphql/src/resolvers/user.js
```diff
@@ -94,6 +94,12 @@ const userResolvers = {
       }
       return (ENABLED_DEMO_MODE && context.user.id !== current.id) ? REDACTED_USER.name : current.name;
     },
+    effective_confidence_level: (current, _, context) => {
+      if (current.entity_type === ENTITY_TYPE_USER) {
+        return getUserEffectiveConfidenceLevel(current, context);
```
labo-flg, line 99: We need to manually resolve this field as it's computed and not in DB
note that `getUserEffectiveConfidenceLevel` will fetch from cache if possible

opencti-platform/opencti-graphql/config/schema/opencti.graphql
```diff
@@ -1331,7 +1331,12 @@ type Member {
   id: ID! # internal_id
   name: String!
   entity_type: String!
+  # entity_type == User
+  effective_confidence_level: EffectiveConfidenceLevel
+  # entity_type == Group
+  group_confidence_level: ConfidenceLevel
```
labo-flg, line 1337: RFC here

labo-flg:
back to draft : need improvements to link to user's details

opencti-platform/opencti-front/lang/front/fr.json
```diff
@@ -1,6 +1,7 @@
 {
   " (reversed)": "(inversé)",
   " & ": " & ",
+  "+ N override(s)": "+ {count}} surcharge(s)",
```
lndrtrbn: A `}` that should be removed?

opencti-platform/opencti-front/src/private/components/common/form/CreatorField.tsx
```diff
@@ -42,28 +26,66 @@ const CreatorFieldQuery = graphql`
...
+    readonly max_confidence: number;
+    readonly overrides: ReadonlyArray<{
+      readonly entity_type: string;
+      readonly max_confidence: number;
+    }>;
+  } | null | undefined;
+  readonly entity_type: string;
+  readonly id: string;
+  readonly name: string;
+};
```
lndrtrbn: You cannot use the type generated by the query `CreatorFieldQuery`?

Archidoit (approved):

opencti-platform/opencti-front/src/private/components/common/form/CreatorField.tsx
```diff
@@ -74,49 +96,72 @@ const CreatorField: FunctionComponent<CreatorFieldProps> = ({
...
-          option: { color: string; label: string },
+          option: CreatorOption,
         ) => (
-          <li {...props}>
-            <div className={classes.icon} style={{ color: option.color }}>
-              <ItemIcon type="user" />
-            </div>
-            <div className={classes.text}>{option.label}</div>
+          <li {...props} >
+            <Box
```
lndrtrbn: You could use `div` instead of MUI component `Box`. In our case Box just render a div but with performance cost. From the MUI documentation:
image
https://mui.com/system/getting-started/usage/#performance-tradeoffs

opencti-platform/opencti-front/lang/front/fr.json
labo-flg: ooh.. yes

opencti-platform/opencti-front/src/private/components/common/form/CreatorField.tsx
labo-flg: I tried. I failed.

opencti-platform/opencti-front/src/private/components/common/form/CreatorField.tsx
labo-flg: ok sure. When shall we use Box instead then ? only if we need theme-related styles ?

opencti-platform/opencti-front/src/private/components/common/form/CreatorField.tsx
labo-flg: Type selector is a mess in that case, if you have some tips I'm all ears

opencti-platform/opencti-front/src/private/components/common/form/CreatorField.tsx
lndrtrbn: I have no idea, never been in a situation where I need this component...

lndrtrbn (approved):

opencti-platform/opencti-front/src/private/components/common/form/CreatorField.tsx
```diff
@@ -74,49 +99,72 @@ const CreatorField: FunctionComponent<CreatorFieldProps> = ({
...
+            >
+              {option.label}
+            </Box>
+            {option.extra && (
+              <Box
+                sx={{
+                  flexGrow: 1,
+                  marginLeft: 1,
+                  textAlign: 'right',
+                  color: 'text.disabled',
```
labo-flg, line 160: @lndrtrbn like here, with sx you have access to theme properties directly

opencti-platform/opencti-graphql/config/schema/opencti.graphql
SouadHadjiat: where is the resolver for this field ? I see only effective_confidence_level resolver for Member.

labo-flg: this field does not need specific resolution. We pull the entities from elastic and `group_confidence_level` is one of the field returned in case it's a group.

on the contrary `effective_confidence_level` is a computed field, not stored in elastic.

SouadHadjiat: do we need it somewhere `group_confidence_level` ? I can't find where we use it in the frontend calling members API.

labo-flg: indeed, `group_confidence_level` is unused for now but I added in anticipation.

Issue #7064, opened 2024-05-22
Mutliple nested relations does not work 


nino-filigran:
## Description

When I attempt to create mutliple nested relationships from an observable, it only select the first entity I've selected. This beahvior is not aligned with relationship creations, where I can create directly mutliple relations at once.
This should happen

## Environment

Demo

## Reproducible Steps

Go into a File Observable
Go to its knowledge tab
Click on + on nested objects
Select at least 2 other files (or any entities **of the same type**)
Click on continue
Only the first file appears 

## Expected Output

If I've selected mutliple entities of the same type, I should see in the relation screen under the entity type "mutliple entities selected"

## Actual Output

I only see the first entity 
## Screenshots (optional)
Example of the relations screen when creating mutliple relation between an intrusion set and mutliple malwares:

![image](https://github.com/OpenCTI-Platform/opencti/assets/148041607/be96b2a4-e6e9-4198-bda0-3faba8d89a28)

PR #7193, opened 2024-05-30
[frontend] Mutliple nested relations does not work (#7064)

ValentinBouzinFiligran:
### Proposed changes

* component update to created multiple nested objects from entities with same type 

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/7064

### Checklist


Archidoit:
Behavior and code Ok :)

Archidoit:
What do we want if we select entities of different entity types ? for the moment, no 'relationship type' is available in this case, even if there are common available relationship types for the selected entities

Archidoit (approved):

PR #7195, opened 2024-05-30
[frontend] Allow Users to Add an Entity as Context When Uploading File

VerboseCat:
### Proposed changes

* Allow user to select a related entity when uploading a file from global context
* Add noMultiSelect prop to filter to allow only one value per filter

### Checklist

### Further comments
 No frontend tests exist for the file upload process.


Archidoit:
Thanks for you PR on this interesting functionnality ! I recommend a modification on the way we select the related entity :

When you upoload a file from Data/import, a pop-up opens. Instead of having a boolean switch button ('add related entity') and a filter icon button to select an entity (screenshot 1), we should rather have just a dropdown that enables to select entities (like in screenshot 2). This dropdown would not be mandatory to fill. This way the selection is easier, requires less clicks and is more consistant with what is done elsewhere in the platform.
We can even imagine that the user can select several entities (The file will be uploaded in the context of each of the selected entities and the user will be redirected to the first selected entity).

- What has been done (screenshot 1):
image

- Recommended solution (screenshot 2):
image

image


- Entities dropdown precisions:
The dropdown contains all the entities, grouped by entity type.
You can also add a palette to select the entity type (and thus prefilter the entities list), like it is done in the 'target type' filter :
image

Jipegien:
@Archidoit after discussion with Craig and others, here what we will do:
image
No palette, no sorting by entity type.

MVP vibe.

VerboseCat:
@Archidoit Where was this screenshot of the AutoCompleteField taken in the platform? I am familiar with the component, but this screenshot is the first time I have seen it with multiple entity types in the dropdown. Thank you!

Archidoit:
@VerboseCat Here is how to find the screenshot:

- Go to the observables list
- Select some observables
- Click on 'add in container' in the bottom toolbar
- A pop-up opens with the dropdown we're talking about

image

image

VerboseCat:
I believe I have updated the component according to the requested changes. Please let me know when you have had a chance to re-review. Thank you!

Archidoit:
@VerboseCat
Behavior tested and everything is OK
I'm just gonna push some little changes in the code before approving (remove the unused 'noMultiSelect' option for FilterIconButton)

VerboseCat:
Great, thank you!

Archidoit (approved):

PR #7218, opened 2024-06-04
[frontend/backend] Share the result of inferrence rules in TAXII collections (#7069)

### LINKED ISSUE #7069: Share the result of inferrence rules in TAXII collections
## Use case
We have created https://github.com/OpenCTI-Platform/opencti/issues/6171 to be able to automatically add indicators/observables promoted from indicators/observables in a container.
Once this is done, we would 3 ways to achieve this: 
- use the inference rules 
- use the automation (only EE users)
- use the above-mentioned feature

The problem with the inference rule is that its results (the entities/relations created from an inference rule) are not shared within the TAXII feeds. It creates a gap in the whole use case which would be to automatically add generated indicators/observables in a container & share them in the different feeds you have set up. 

## Proposed Solution

Add an option, enabled by default, to share by default in a feed the result of inference rules.
This will not affect already existing rules.

ValentinBouzinFiligran:
### Proposed changes

* checkbox added to taxii creation drawer
* value created on elastic
* value usable on collectionQuery to change the index

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/7069

### Checklist


opencti-platform/opencti-graphql/src/modules/attributes/internalObject-registrationAttributes.ts
```diff
@@ -360,6 +360,7 @@ const internalObjectsAttributes: { [k: string]: Array<AttributeDefinition> } = {
     { name: 'description', label: 'Description', type: 'string', format: 'text', mandatoryType: 'no', editDefault: false, multiple: false, upsert: true, isFilterable: true },
     { name: 'filters', label: 'Filters', type: 'string', format: 'text', mandatoryType: 'no', editDefault: false, multiple: false, upsert: false, isFilterable: false },
     { name: 'taxii_public', label: 'Public taxii', type: 'boolean', mandatoryType: 'external', editDefault: true, multiple: false, upsert: false, isFilterable: true },
+    { name: 'withInferences', label: 'With inferences', type: 'boolean', mandatoryType: 'no', editDefault: true, multiple: false, upsert: false, isFilterable: false },
```
SouadHadjiat: attributes are usually named with underscore like "with_inferences".
Maybe "include_inferences" could be more specific ?

opencti-platform/opencti-front/src/private/components/data/taxii/http-client.private.env.json
```diff
@@ -0,0 +1,5 @@
+{
```
SouadHadjiat: what is this file?

opencti-platform/opencti-front/src/private/components/data/taxii/TaxiiCollectionCreation.tsx
```diff
@@ -88,6 +89,12 @@ const TaxiiCollectionCreation: FunctionComponent<TaxiiCollectionCreationProps> =
   const { t_i18n } = useFormatter();
   const classes = useStyles();
   const [filters, helpers] = useFiltersState(emptyFilterGroup);
+  const [areInferenceRulesResultAdded, setAreInferenceRulesResultAdded] = useState(true);
```
SouadHadjiat: why is `include_inferrences` in a separate state and not directly part of the form ?

opencti-platform/opencti-front/src/private/components/data/taxii/http-client.private.env.json
ValentinBouzinFiligran: I don't know where it comes from

opencti-platform/opencti-front/src/private/components/data/taxii/TaxiiCollectionCreation.tsx
ValentinBouzinFiligran: updated ! :) 
thanks !

SouadHadjiat:
When creating a taxii collection including inferrences, I can see them in the collection. Then when I create a feed ingesting this collection, I get a duplicate of my inferred relationships :

Capture d'écran 2024-06-17 171114

=> seen with product: not an issue

SouadHadjiat (approved):

Issue #6836, opened 2024-04-29
Ensure the valid_until date on Indicators is set to a greater value than valid_from when empty (compliance with STIX 2.1


Jipegien:
## Use case

From community : 
I observe that when a “revoked” indicator is published with an empty “valid_until” date, it is modified by OpenCTI to define a “valid_until” date equal to the “valid_from” date according to this code: https://github.com/OpenCTI-Platform/opencti/blob/b1257ec8e4a4e7f185943fbf541ab297a34b4022/opencti-platform/opencti-graphql/src/modules/indicator/indicator-utils.ts#L82
However, according to the STIX 2.1 specification, it seems that the 'valid_until' date MUST be greater than the 'valid_from' property:
https://docs.oasis-open.org/cti/stix/v2.1/os/stix-v2.1-os.html#_wfiae74706sw
Setting 'valid_until' to the date of 'valid_from' causes compatibility problems with the STIX2 python library, which generates an error such as “Exception: indicator--7a7b3cb1-8e39-57ee-b80d-698fcea8c9f0 ‘valid_until’ must be greater than ‘valid_from’” and prevent to consume TAXII feed with the usage of standard TAXII/STIX python library.
https://filigran-community.slack.com/archives/C06CF1N302W/p1714386968202249

Jipegien:
just ensure that the valid_until date is one second greater than valid_from attribute when empty

SarahBocognano:
Update: After discussing with the product team, we noticed that there was another issue on this topic. Currently, in the platform, we can set a valid_from equal to valid_until, which is not standard in the STIX format. Therefore, we have decided to add a validation check and an error message to enforce that valid_until must be greater than valid_from to comply with the STIX standard.

PR #7227, opened 2024-06-04
[backend] Ensure the valid_until date on Indicators is set to a greater value than valid_from when empty (#6836)

SarahBocognano:
### Proposed changes

* According to STIX 2.1 compliance, the until_date must be greater than the valid_from. However right now, if we do not put an until_date manually, we fill the field with the exact same date as valid_from.
* So I add one second to the valid_from if the field until_date is empty

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/6836

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/modules/indicator/indicator-converter.ts
+++ b/opencti-platform/opencti-graphql/src/modules/indicator/indicator-converter.ts
@@ -2,15 +2,15 @@ import moment from 'moment';
 import { buildKillChainPhases, buildMITREExtensions, buildStixDomain, cleanObject, convertToStixDate } from '../../database/stix-converter';
 import { STIX_EXT_MITRE, STIX_EXT_OCTI } from '../../types/stix-extensions';
 import type { StixIndicator, StoreEntityIndicator } from './indicator-types';
+import { isNotEmptyField } from '../../database/utils';
 
 const convertIndicatorToStix = (instance: StoreEntityIndicator): StixIndicator => {
   const indicator = buildStixDomain(instance);
-  const computedValidUntil = () => {
-    if (instance.valid_from === instance.valid_until) {
-      return moment(instance.valid_from).add(1, 'seconds').toDate();
-    }
-    return instance.valid_until;
-  };
+  // Adding one second to the valid_until if valid_from and valid_until are equals,
+  // because according to STIX 2.1 specification the valid_until must be greater than the valid_from.
+  const computedValidUntil = (
+    isNotEmptyField(instance.valid_from) === isNotEmptyField(instance.valid_until)
+  ) ? moment(instance.valid_from).add(1, 'seconds').toDate() : instance.valid_until;
   return {
     ...indicator,
     name: instance.name,
@@ -20,7 +20,7 @@ const convertIndicatorToStix = (instance: StoreEntityIndicator): StixIndicator =
     pattern_type: instance.pattern_type,
     pattern_version: instance.pattern_version,
     valid_from: convertToStixDate(instance.valid_from),
-    valid_until: convertToStixDate(computedValidUntil()),
+    valid_until: convertToStixDate(computedValidUntil),
     kill_chain_phases: buildKillChainPhases(instance),
     extensions: {
       [STIX_EXT_OCTI]: cleanObject({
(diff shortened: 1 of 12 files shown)


Archidoit:
I can have an empty valid_until and a not empty valid_from... 
image

@SarahBocognano : Hmm in which case those this happening ? Because I didn't touch that part, and on my platforme I can have an indicator with no valid_from and no valid_until
image

SarahBocognano:
Update : @Archidoit your point seems valid, need to confirm with product but will probably add the change in the frontend as well to add a automatic fill of valid_until (+1seconde) if valid_from is set

SarahBocognano:
Update: After discussing with the product team, we noticed that there was another issue on this topic. Currently, in the platform, we can set a valid_from equal to valid_until, which is not standard in the STIX format. Therefore, we have decided to add a validation check and an error message to enforce that valid_until must be greater than valid_from to comply with the STIX standard.

To resume : 
- We can have an empty valid_until
- We cannot have a valid_until equal or inferior to a valid_from
- We add one seconde to valid_until if the field was empty AND it's revoked for stix compliance

Archidoit:
I can edit an indicator and have valid_from = valid_until: 
image

Archidoit:
When valid_until = valid_from at indicator creation, the error is badly formatted :
image


--> use Yup to have the same kind of error when valid_from > valid_until 
image

Archidoit:
Indicators with valid_from = valid_until can exist in the platform... Do we have to migrate them and add 1s to valid_until for instance ? @romain-filigran

romain-filigran:
> Indicators with valid_from = valid_until can exist in the platform... Do we have to migrate them and add 1s to valid_until for instance ? @romain-filigran

I don't have a very good view of the difficulty this represents in regard of the potential number of indicators in this situation (particularly due to decay). Do you think it's mandatory ? What is the risk if we don't migrate them, knowing that this should mainly correspond to indicators considered as "Revoked"?

opencti-platform/opencti-graphql/src/modules/indicator/indicator-domain.ts
```diff
@@ -299,12 +299,21 @@ export const addIndicator = async (context: AuthContext, user: AuthUser, indicat
...
+    if (e.key === 'valid_until') [valid_until] = e.value;
+  });
+  if (new Date(valid_until) <= new Date(valid_from)) {
+    throw ValidationError('valid_from', { message: 'The valid until date must be greater than the valid from date' });
+  }
   const scoreEditInput = input.find((e) => e.key === 'x_opencti_score');
   if (scoreEditInput) {
-    const indicator = await findById(context, user, id);
-    if (!indicator) {
-      throw FunctionalError('Cannot edit the field, Indicator cannot be found.');
```
Archidoit, line 306: why remove this part of code?

Archidoit:
Behavior ok 👍

opencti-platform/opencti-graphql/src/modules/indicator/indicator-domain.ts
SarahBocognano: I didn't really remove it, I just put it two lines above to be able to use the indicator const not only in the scorEditInput case

Archidoit (approved):

Archidoit:
Approved, waiting for the PR in client-python to align the behavior

opencti-platform/opencti-graphql/src/modules/indicator/indicator-converter.ts
```diff
@@ -1,9 +1,16 @@
...
 import { STIX_EXT_MITRE, STIX_EXT_OCTI } from '../../types/stix-extensions';
 import type { StixIndicator, StoreEntityIndicator } from './indicator-types';
+import { isNotEmptyField } from '../../database/utils';
 
 const convertIndicatorToStix = (instance: StoreEntityIndicator): StixIndicator => {
   const indicator = buildStixDomain(instance);
+  // Adding one second to the valid_until if valid_from and valid_until are equals,
+  // because according to STIX 2.1 specification the valid_until must be greater than the valid_from.
+  const computedValidUntil = (
+    isNotEmptyField(instance.valid_from) === isNotEmptyField(instance.valid_until)
```
Archidoit:
```suggestion
    isNotEmptyField(instance.valid_from) && isNotEmptyField(instance.valid_until) && instance.valid_until === instance.valid_from
```

Archidoit: isNotEmptyField retourne un boolean, le test suivant retournera toujours vrai si les deux instances sont toutes deux null ou toutes deux non définies:  isNotEmptyField(instance.valid_from) === isNotEmptyField(instance.valid_until)

PR #7233, opened 2024-06-05
[backend] Change external reference resolution for OpenBAS integration

### LINKED ISSUE #7223: Security Posture on OpenCTI
## Description

When I create a scenario from OpenCTI, I want to have my security posture of the latest simulation based on scenarios created from my entity.

## Solution

OpenCTI should get the latest ended simulation based on scenarios instead of the latest simulation based on the first scenario.

## Testing

- Generate a scenario from OpenCTI to OpenBAS
- Launch a simulation with some expectations
- Wait for the end of this simulation
- Go to opencti and checked that the security posture is the same as that on OpenBAS
- Re launch the simulation, change the expectation result and verify the new posture in both platform

RomuDeuxfois:
### Proposed changes

OpenCTI should get the latest ended simulation based on scenarios instead of the latest simulation based on the first scenario.

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/7223

### Testing

- Generate a scenario from OpenCTI to OpenBAS
- Launch a simulation with some expectations
- Wait for the end of this simulation
- Go to opencti and checked that the security posture is the same as that on OpenBAS
- Re launch the simulation, change the expectation result and verify the new posture in both platform

OpenBAS PR -> https://github.com/OpenBAS-Platform/openbas/pull/1048

### DIFF
--- a/opencti-platform/opencti-graphql/src/database/xtm-obas.ts
+++ b/opencti-platform/opencti-graphql/src/database/xtm-obas.ts
@@ -175,7 +175,7 @@ export const getScenarioResult = async (id: string) => {
   }
   const httpClient = buildXTmOpenBasHttpClient();
   try {
-    const { data: exercise } = await httpClient.get(`/scenarios/external_reference/${id}`);
+    const { data: exercise } = await httpClient.get(`/opencti/v1/exercises/latest/${id}`);
     if (!exercise || !exercise.exercise_id) {
       return noResult;
     }
--- a/opencti-platform/opencti-graphql/src/database/xtm-obas.ts
+++ b/opencti-platform/opencti-graphql/src/database/xtm-obas.ts
@@ -2,7 +2,6 @@ import conf, { getBaseUrl, logApp } from '../config/conf';
 import { type GetHttpClient, getHttpClient } from '../utils/http-client';
 import type { Label } from '../generated/graphql';
 import { DatabaseError } from '../config/errors';
-import { utcDate } from '../utils/format';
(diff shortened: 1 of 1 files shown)


opencti-platform/opencti-graphql/src/database/xtm-obas.ts
```diff
@@ -176,19 +175,11 @@ export const getScenarioResult = async (id: string) => {
   }
   const httpClient = buildXTmOpenBasHttpClient();
   try {
-    const { data: scenario } = await httpClient.get(`/scenarios/external_reference/${id}`);
-    if (!scenario || !scenario.scenario_id) {
+    const { data: exercise } = await httpClient.get(`/scenarios/external_reference/${id}`);
```
SouadHadjiat: just to understand: this openBAS API has changed its signature and returns now an exercise instead of a scenario, right ? 
So if openBAS and openCTI are not both updated to the latest version, this code will fail I guess ?

RomuDeuxfois: Exactly, we need to have a good synchronisation between the two products

lndrtrbn: Do we write somewhere which version of octi is compatible with which version of obas?

RomuDeuxfois: We are thinking about a robuste plan to handle this.
Good point :)

SouadHadjiat (approved):
tested locally

PR #7235, opened 2024-06-05
[backend] prevent wrong type for empty pid (#6968)

### LINKED ISSUE #6968: Can't remove pid of a process
Create a process (it is a type of observable) with PID = 22 and command_line = 'command'.
Edit the process to remove the PID. An error occurs : 
image

Note: We should be able to remove the PID of a process since it is possible to create a process with no PID.

frapuks:
### Proposed changes

- Manage value of PID (type number) when removing the value (empty string is sending)

### Related issues

- close #6968 

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/domain/stixCyberObservable.js
+++ b/opencti-platform/opencti-graphql/src/domain/stixCyberObservable.js
@@ -42,6 +42,7 @@ import { addFilter } from '../utils/filtering/filtering-utils';
 import { ENTITY_TYPE_INDICATOR } from '../modules/indicator/indicator-types';
 import { controlUserConfidenceAgainstElement } from '../utils/confidence-level';
 import { uploadToStorage } from '../database/file-storage-helper';
+import { isNumericAttribute } from '../schema/schema-attributes';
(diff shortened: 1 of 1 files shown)


opencti-platform/opencti-graphql/src/domain/stixCyberObservable.js
```diff
@@ -273,6 +273,7 @@ export const stixCyberObservableEditField = async (context, user, stixCyberObser
     input,
     opts
   );
+  if (isEmptyField(stixCyberObservable.pid)) stixCyberObservable.pid = undefined;
```
Archidoit: are you sure there is no other solution than implemented this specific line for pid ? For instance 'dst_packets' (for Network traffic) is also of type 'number' and we can remove it without problem and without implemented a specific line like that...

Archidoit:
Working but the code solution may not be the best...

Archidoit:
Working well, code ok for me 👍

SarahBocognano (approved):

Archidoit (approved):

PR #7238, opened 2024-06-05
[frontend]CSV Mapper default value for boolean field (#7116)

### LINKED ISSUE #7116: CSV Mapper default value for boolean fields are not working
## Environment
Testing 6.1.4

## Reproducible Steps

* Create a new mapper
* add representation for entity type malware
* _name_ in column A
* add a default value for _is family_

## Expected Output

Valus is set to false or true depending on the selection in the default value popup window

## Actual Output

* if you select "false" with the toggle, it does not take false as the default value
* if you select "true"  with the toggle, it does take something as default value apparently, but display is incorrect

## Screenshots (optional)

image
image
image

CelineSebe:
### Proposed changes
* replace default value toggle by true, false or nothing for is_family attribute configuration (Malware)
* replace "has a default value" message by a badge

### Related issues

* #7116
 
### Checklist


opencti-platform/opencti-front/src/private/components/data/csvMapper/representations/attributes/CsvMapperRepresentationDialogOption.tsx
```diff
@@ -1,23 +1,32 @@
...
   const { t_i18n } = useFormatter();
   const handleClickOpen = () => {
     setOpen(true);
   };
+  const [invisible, setInvisible] = useState(true);
+
+  useEffect(() => {
+    const hasDefaultValues = value?.default_values !== null && value?.default_values !== undefined && JSON.stringify(value.default_values) !== '[]';
+    setInvisible(!hasDefaultValues);
+  }, [value]);
```
lndrtrbn: You can replace this by a useMemo to avoid using a useEffect

lndrtrbn (commented):
image

In Settings > Customization, missing a label "Default value" above the buttons to be able to understand what we are setting.
Same for CSV mapper

lndrtrbn:
image
image

On malware representation, the bullet is displayed on the name even if nothing is configure

lndrtrbn:
When the bullet is displayed, there is an extra padding at the bottom

opencti-platform/opencti-front/lang/front/fr.json
```diff
@@ -1036,7 +1036,7 @@
   "EXTERNAL_IMPORT": "Import de données",
   "Eye Color": "Couleur des yeux",
   "Failure": "Échec",
-  "false": "Non",
+  "false": "Faux",
```
lndrtrbn, line 1043: The translation is also incorrect in english

opencti-platform/opencti-front/lang/front/en.json
```diff
@@ -1040,7 +1040,7 @@
   "Eye Color": "Eye Color",
   "Failed": "Failed",
   "Failure": "Failure",
-  "false": "No",
+  "false": "false",
```
labo-flg, line 1043: This change might have more impact. the intl key "false" was maybe used elsewhere with intent on displaying "No".
If so, we should change there to intl key "No" (and same with yes/true)

lndrtrbn: It is not used anywhere else in the app for what I can see

labo-flg (approved):

PR #7248, opened 2024-06-06
[Frontend] Fix toolbar selection in knowledge -> indicators (#7245)

### LINKED ISSUE #7245: "Select all" selects the whole platform, not just what's on the screen
## Description

There looks to be a bug related to issue https://github.com/OpenCTI-Platform/opencti/issues/6411 affecting Intrusion-sets and Malware. Unfortunately, I don't have enough data in my test environment to test threat actor/tools etc etc.

## Environment

OpenCTI 6.1.8

## Reproducible Steps

Using data from the Alienvault Connector...
Arsenal -> Malware -> Knowledge -> Indicators
Select all indicators using the select all check box
Update entities -> Add -> Label -> test (or whatever label you want to add) -> Update
Confirm in Launch a background task window that only a small number of elements are selected -> Launch
Go to background task page to see that target entities has changed to all indicators

These steps are also repeatable with Intrusion-Sets

## Expected Output

Only the selected entities should be having labels added.

## Actual Output

All entities are having labels added.

## Additional information

image
image

This bug ruined my night.

Kedae:
Closes #7245


SouadHadjiat (approved):
Issue fixed ✅
image

PR #7275, opened 2024-06-08
[frontend/backend] In CSV Feed Ingester, take into account Default Marking definition options from CSV Mapper

### LINKED ISSUE #6149: In CSV Feed Ingester, take into account Default Marking definition options from CSV Mapper
## Use case

In CSV Mapper, it is possible to choose between 2 options to add Marking definition to imported data when marking are not directly provided by the file's data.
image

When importing a file directly (for example via Data/Import), user is able to provide a marking definition to be used, if the right option in CSV mapper has been chosen.
image

But, when importing from a CSV feed ingester, user cannot define such a marking in the configuration of the csv feed.

We need to give user the ability to provide necessary Default marking definition in CSV feed configuration if the associated CSV mapper ask to do so.

Goumies:
### Proposed changes

* handle CSV mapper configuration on CSV feed ingester creation
* handle CSV mapper configuration on CSV feed ingester edition
* update Launch import dialogs

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/6149

### Checklist

### Further comments

To test this feature, create CSV mappers with both options: 
- Let the user choose
- Use default marking definitions of the user

image

image

image

image

To set default markings:
- you have to go to Customization > Entity types
image

- select an entity type
- select the Marking attribute
image

- activate default values
image

- navigate to Security > Groups, select the current group (=> default)
image

- select allowed markings, so they can be added to the list of options for default markings
- select default markings
image

Single column CSV file to test your CVS mapper: 
[6149-ipv4-addr.csv](https://github.com/user-attachments/files/15805659/6149-ipv4-addr.csv)

Single column CSV feed to ingest IPv4 addresses: https://lists.blocklist.de/lists/all.txt

To check for the successful ingestion: 
- make sure your worker is running
- go to Observations > Observables
- see the IPv4 addresses with corresponding markings

To check different use cases :
- purge messages in push_sync queue in RabbitMQ
image
- delete ingested entities in elastic console
```
POST /opencti_stix_cyber_observables-000001/_delete_by_query
{
  "query": {
    "bool": {
      "must": [
          { "term":{"entity_type.keyword": "IPv4-Addr" }}
        ]
    } // match_all
  }
}
```
image


opencti-platform/opencti-graphql/src/modules/ingestion/ingestion-csv.graphql
```diff
@@ -23,6 +23,7 @@ type IngestionCsv implements InternalObject & BasicObject {
     ingestion_running: Boolean
     current_state_hash: String
     current_state_date: DateTime
+    markings: [String]
```
SouadHadjiat: Can we have null values in this list ? Otherwise I suggest to set the type to `[String!]` which will make sure that the list returns non null values (which is easier to handle afterwards).

opencti-platform/opencti-graphql/src/modules/ingestion/ingestion-csv.graphql
```diff
@@ -69,6 +70,7 @@ input IngestionCsvAddInput {
     csv_mapper_id: String!
     ingestion_running: Boolean
     user_id: String!
+    markings: [String]
```
SouadHadjiat: same comment here => `[String!]`

opencti-platform/opencti-graphql/src/manager/ingestionManager.ts
```diff
@@ -299,7 +299,10 @@ const csvDataToObjects = async (csvBuffer: Buffer | string, ingestion: BasicStor
 const csvDataHandler = async (context: AuthContext, ingestion: BasicStoreEntityIngestionCsv) => {
   const user = context.user ?? SYSTEM_USER;
   const csvMapper = await findById(context, user, ingestion.csv_mapper_id);
-  const csvMapperParsed = parseCsvMapper(csvMapper);
+  const csvMapperParsed = {
+    ...parseCsvMapper(csvMapper),
+    markings: ingestion.markings ?? []
```
SouadHadjiat: `markings` is not declared in `CsvMapperParsed` interface (nor in BasicStoreEntityCsvMapper), why do we add it here ? Is it used afterwards ?

Goumies: Since the CSV feed ingestion entry point to the CSV Import logic is `bundleProcess`, we have opted to pass it like this.
In the CSV import process, the markings are sent in the message. The only thing stored in the database is the configuration (user choice or user default markings).

Markings are retrieved at the beginning of the mapping process.

opencti-platform/opencti-front/src/private/components/analyses/external_references/ExternalReferenceFileImportViewer.tsx
```diff
@@ -227,7 +248,7 @@ ExternalReferenceFileImportViewerBaseProps
                 onClose={() => handleReset()}
                 fullWidth={true}
               >
-                <DialogTitle>{t_i18n('Launch an import')}</DialogTitle>
+                <DialogTitle>{`${t_i18n('Launch an import')}`}</DialogTitle>
```
SouadHadjiat: no need to change this line if the title didn't change, right ?

CelineSebe:
In the other fields, a green badge is displayed on the toothed wheel when setting a default value. 
This does not seem to be the case when setting the default marking in the config.

image

opencti-platform/opencti-front/src/private/components/analyses/external_references/ExternalReferenceFileImportViewer.tsx
Goumies: It was just an improvement attempt.
I will remove it.

Goumies:
> In the other fields, a green badge is displayed on the toothed wheel when setting a default value. This does not seem to be the case when setting the default marking in the config.
> 
> image

I had no idea about this badge. I have noticed other places without the green badge on attributes set.
@Jipegien, do we take care of this in this PR?

Jipegien:
> > In the other fields, a green badge is displayed on the toothed wheel when setting a default value. This does not seem to be the case when setting the default marking in the config.
> > image
> 
> I had no idea about this badge. I have noticed other places without the green badge on attributes set. @Jipegien, do we take care of this in this PR?

No.

SouadHadjiat:
In data -> Import 
Markings selector doesn't appear for CSV mapper with option "let the user choose"

image

opencti-platform/opencti-front/src/private/components/analyses/external_references/ExternalReferenceFileImportViewer.tsx
```diff
@@ -103,16 +106,23 @@ ExternalReferenceFileImportViewerBaseProps
   const importConnsPerFormat = scopesConn(connectorsImport);
   const handleOpenImport = (file: FileLine_file$data | null | undefined) => setFileToImport(file);
   const handleCloseImport = () => setFileToImport(null);
-  const onSubmitImport: FormikConfig<{ connector_id: string, configuration: string }>['onSubmit'] = (
+  const onSubmitImport: FormikConfig<{ connector_id: string, configuration: string, objectMarking: Option[] }>['onSubmit'] = (
```
SouadHadjiat, line 110: configuration & objectMarkings are not always defined (for other connectors than CSV) :
image

image

Goumies: Fix on every launch an import dialogs

opencti-platform/opencti-front/src/private/components/analyses/external_references/StixCoreObjectExternalReferencesLines.tsx
```diff
@@ -565,16 +573,14 @@ StixCoreObjectExternalReferencesLinesContainerProps
                   </Field> : <ManageImportConnectorMessage name={selectedConnector?.name }/>
                 }
                 {selectedConnector?.name === 'ImportCsv'
+                    && hasUserChoiceCsvMapper
```
Goumies, line 582: Marking field hidden with user choice config
External lines of any entity, in overview

opencti-platform/opencti-front/src/private/components/common/files/FileManager.jsx
```diff
@@ -161,7 +161,7 @@ const FileManager = ({
     if (isCsvConnector && configuration && objectMarking) {
       const parsedConfig = JSON.parse(configuration);
       if (typeof parsedConfig === 'object') {
-        parsedConfig.user_chosen_markings = objectMarking.map((marking) => marking.value);
+        parsedConfig.markings = objectMarking.map((marking) => marking.value);
```
SouadHadjiat: why change this property name here ? I think it should stay user_chosen_markings

opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectFilesAndHistory.jsx
```diff
@@ -164,7 +164,7 @@ const StixCoreObjectFilesAndHistory = ({
     if (isCsvConnector && configuration && objectMarking) {
       const parsedConfig = JSON.parse(configuration);
       if (typeof parsedConfig === 'object') {
-        parsedConfig.user_chosen_markings = objectMarking.map((marking) => marking.value);
+        parsedConfig.markings = objectMarking.map((marking) => marking.value);
```
SouadHadjiat: why change this property name here ? I think it should stay user_chosen_markings

opencti-platform/opencti-front/src/private/components/data/import/ImportContent.jsx
```diff
@@ -229,7 +240,7 @@ class ImportContentComponent extends Component {
     if (isCsvConnector && configuration && objectMarking) {
       const parsedConfig = JSON.parse(configuration);
       if (typeof parsedConfig === 'object') {
-        parsedConfig.user_chosen_markings = objectMarking.map((marking) => marking.value);
+        parsedConfig.markings = objectMarking.map((marking) => marking.value);
```
SouadHadjiat, line 243: why change this property name here ? I think it should stay user_chosen_markings

SouadHadjiat:
Need to optimize perf in `mappingProcess` (extract all refEntities part to pass it as parameter of the method, so we don't load entities for each record)

opencti-platform/opencti-front/lang/front/fr.json
```diff
@@ -1226,6 +1227,7 @@
   "IF": "SI",
   "if needed": "si nécessaire",
   "If needed, you can set a": "Si nécessaire, vous pouvez définir un",
+  "If the CSV mapper is configured with \"Use default markings definitions of the user\", the default markings of the user responsible for data creation are applied to the ingested entities. Otherwise, you can choose markings to apply.": "Si le mappeur CSV est configuré avec \"Use default markings definitions of the user\" ;, les marquages par défaut de l'utilisateur responsable de la création des données sont appliqués aux entités ingérées. Sinon, vous pouvez choisir les marquages à appliquer.",
```
aHenryJard: There is a part of text that is in English here.

opencti-platform/opencti-front/src/private/components/common/form/ObjectMarkingField.tsx
```diff
@@ -76,16 +115,68 @@ const ObjectMarkingField: FunctionComponent<ObjectMarkingFieldProps> = ({
...
   const [newMarking, setNewMarking] = useState<
   Option[] | OptionValues | undefined
   >(undefined);
   const [operation, setOperation] = useState<string | undefined>(undefined);
+  const [otherUserAllowedMarkingsData, setOtherUserAllowedMarkingsData] = useState(
+    [] as { definition: string | null | undefined, id: string, x_opencti_color: string | null | undefined }[] | undefined,
+  );
+
+  const fetchCreatorAllowedMarking = async (creatorId: string) => {
+    return fetchQuery(objectMarkingFieldAllowedMarkingQuery, {
```
aHenryJard, line 135: Usage of fetchQuery qhould be avoid in new code, we should use more usePreloadedQuery and loadQuery

opencti-platform/opencti-front/src/private/components/common/form/ObjectMarkingField.tsx
```diff
@@ -76,16 +115,68 @@ const ObjectMarkingField: FunctionComponent<ObjectMarkingFieldProps> = ({
...
+      id: creatorId,
+    })
+      .toPromise();
+  };
+
+  const fetchCreatorAllowedMarkings = async (creatorAllowedMarkingIds: string[]) => {
+    if (creatorAllowedMarkingIds.length) {
+      return fetchQuery(objectMarkingFieldOtherUserAllowedMarkingsQuery, {
```
aHenryJard, line 143: Usage of fetchQuery qhould be avoid in new code, we should use more usePreloadedQuery and loadQuery

opencti-platform/opencti-front/src/utils/csvMapperUtils.ts
```diff
@@ -0,0 +1,16 @@
+import { CsvMapperFieldOption } from '@components/common/form/CsvMapperField';
+
+const resolveHasUserChoiceParsedCsvMapper = (option: CsvMapperFieldOption) => {
+  return option.representations.some(
+    (representation) => representation.attributes.some(
+      (attribute) => attribute.key === 'objectMarking' && attribute.default_values.some((value) => {
+        if (typeof value === 'string') {
+          return value === 'user-choice';
```
aHenryJard: I see this string 'user-choice' a lot in code, it may be handy to use a constant instead.

aHenryJard (commented):
Works as expected locally, I think we should try to change the fetchQuery on frontend before merge.

opencti-platform/opencti-front/src/private/components/common/form/ObjectMarkingField.tsx
```diff
@@ -76,16 +115,67 @@ const ObjectMarkingField: FunctionComponent<ObjectMarkingFieldProps> = ({
...
+        const creatorAllowedMarkingIds = ((await fetchCreatorAllowedMarking(allowedMarkingOwnerId) as unknown as ObjectMarkingFieldAllowedMarkingQuery$data)
+          .user?.groups?.edges ?? [])
+          .flatMap((group) => (group?.node?.allowed_marking ?? [])
+            .map((marking) => marking.id));
+        const markingsData = ((await fetchCreatorAllowedMarkings(creatorAllowedMarkingIds) as unknown as ObjectMarkingFieldOtherUserAllowedMarkingsQuery$data)
+          ?.markingDefinitions?.edges
+          .map((marking) => ({ ...marking.node })));
+        setOtherUserAllowedMarkingsData(markingsData);
+      };
+      fetchData().then();
```
Kedae: No need for then()

Goumies: The promise must be resolved.
Otherwise it is ignored.
image

Adding async/await break the useEffect signature.

Kedae: Is this an error of IntelliJ or check-ts or eslint ?

Goumies: IntelliJ.
Let me remove it then.

Kedae (approved):

Issue #6801, opened 2024-04-24
Infrastructure Breaking Investigations


explorecti:
## Description

When selecting infrastructure(add only) from a created or existing investigation all targets are shown/expanded.

## Environment

1. OS: Ubuntu
2. OpenCTI version: 6.0.10
3. OpenCTI client: Frontend

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Create a new investigation
2. Add entities and select TYPE: INFRASTRUCTURE
3. Select and click to expand
4. Under "All types of target" select Infrastructure 
5. click EXPAND
6. Results show ALL

## Expected Output

Should only return Infrastructure target

## Actual Output

Returns all targets

## Additional information

None

## Screenshots (optional)

None

Jipegien:
Can't reproduce it.
MalwareA targeting multiple entities. MalwareA targeting Infra1. Creating investigation that contains MalwareA. Expand -> Infrastructure. Only Infra1 appears.
![image](https://github.com/OpenCTI-Platform/opencti/assets/104078945/ac4803c7-f7e3-4594-a941-e99b23f0999d)

Do you have any additional information about your situation?

explorecti:
@Jipegien Please select Type "Infrastructure" when adding entities then expand just the "All types of targets" and choose "Infrastructure", then it displays all targets. Don't use Malware as the example because that doesn't expose the issue.

nino-filigran:
@explorecti I confirm, I've been able to reproduce, by first adding the entity type = infra in the graph, then choosing to expand only infra, resulting in having not only infrastructures but all other linked entities added to my graph.

explorecti:
Is there an update when this will be fixed?

nino-filigran:
Hi @explorecti as you can see @SarahBocognano assigned herself on the ticket. Before, nobody did work on this. Therefore, this will be fixed soon. The ticket will be updated once done.

SarahBocognano:
Update: This issue is applicable to all entities, example :

- I add a threat Actor in investigation
- I select this threat actor 
- I select "Threat Actor" to expand only
Result : Every other relationships of this entity if expanded too

PR #7343, opened 2024-06-11
[frontend] Fix Infrastructure breaking Investigation (#6801)

SarahBocognano:
In this issue, the user revealed that the expansion of an infrastructure with another infrastructure was breaking investigation. 
After research, the problem was global and every time we wanted to expand on the same entity type (example: Threat Actor A + Threat Actor B), the expansion displayed every relation related to the entity Threat Actor A and not just the one between A and B. 

### Proposed changes

* Adapt filters to be able to expand on the same entity type

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/6801


SarahBocognano:
I'm open to rework this PR if needed in order to have an optimized code base !

lndrtrbn (approved):
Works fine, but the name of the commit may be changed to reflect more what the fix is more about, infrastructure was not the issue

PR #7345, opened 2024-06-11
[frontend] paragraphs value check in Ask AI pop-up (#7240)

### LINKED ISSUE #7240: When using Generate Report Document using AI, the first digit in "number of paragraphs" is fixed
## Environment

2. OpenCTI version: 6.1.8
3. OpenCTI client: Frontend MS Edge on MacOS
4. Other environment details: 

## Reproducible Steps

1. Open a Grouping containing 1-3 reports
2. Click "Generate Report Document" 
3. On the dialogue, select HTML and Tactical
4. On the number of paragraphs, try deleting the default of 10

The number should be deleted

## Actual Output

The first digit (1) is not deleted.
if using the up/down buttons to change it, this works, but this soon goes crazy and number auto-increments continuously without input and cycles.

## Additional information

Video available

Archidoit:
issue: https://github.com/OpenCTI-Platform/opencti/issues/7240


opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectAskAI.tsx
```diff
@@ -210,7 +211,21 @@ const StixCoreObjectAskAI: FunctionComponent<StixCoreObjectAskAiProps> = ({ inst
...
       default:
-        // do nothing
+      // do nothing
+    }
+  };
+  const handleAskAi = () => {
+    // check paragraphs value is correct
+    if (action === 'container-report' || action === 'summarize-files') {
+      if (Number.isNaN(paragraphs)) {
+        MESSAGING$.notifyError('Number of paragraphs should be a number');
```
SarahBocognano, line 221: Was it possible to add an error with Yup in the field directly instead of launching a MESSAGING error ?

SarahBocognano (commented):
Tested ✅ works fine

opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectAskAI.tsx
Archidoit: I wanted to but it requires a formik with a submission... don't know if it's really possible, gonna check

ValentinBouzinFiligran (approved):

ValentinBouzinFiligran:
Tested

PR #7355, opened 2024-06-12
[frontend] Fix enrichment of multiple observables when shift-selecting (#7269)

### LINKED ISSUE #7269: Cannot enrich multiple observables when shift-selecting
## Environment

Testing (6.1.10)

## Reproducible Steps

1. Create a report and add to it at least 4 domain names
2. Access the report and then go the Observables tab
3. Select the first domain name, press shift and select the last one

## Expected Output

Being able to enrich

## Actual Output

The enrich button is absent. Unable to enrich

CelineSebe:
### Proposed changes
* remove Ramda
* filter 'undefined' type
* fix shift selecting in entities et observables list

### Related issues

* #7269 

## Reproducible Steps

1. Create a report and add to it at least 4 domain names
2. Access the report and then go the Observables tab
3. Select the first domain name, press shift and select the last one

## Expected Output

Being able to enrich
### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/common/containers/ContainerStixCyberObservablesLines.tsx
+++ b/opencti-platform/opencti-front/src/private/components/common/containers/ContainerStixCyberObservablesLines.tsx
@@ -87,6 +87,7 @@ const ContainerStixCyberObservablesLinesFragment = graphql`
             ... on StixCyberObservable {
               id
               observable_value
+              entity_type
             }
             ...ContainerStixCyberObservableLine_node
           }
--- a/opencti-platform/opencti-front/src/private/components/common/containers/ContainerStixDomainObjectsLines.tsx
+++ b/opencti-platform/opencti-front/src/private/components/common/containers/ContainerStixDomainObjectsLines.tsx
@@ -106,6 +106,7 @@ export const containerStixDomainObjectsLinesFragment = graphql`
           node {
             ... on BasicObject {
               id
+              entity_type
             }
             ...ContainerStixDomainObjectLine_node
           }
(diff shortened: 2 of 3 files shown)


Archidoit:
The enrichment button is not available if I do a 'select all' in the observable list or entities list of a report.
image

CelineSebe:
@Archidoit  I think the enrichment is possible in case: only one entity type is selectionned.
@nino-filigran Could you confirm this functionality?

Archidoit:
Select all entities in the observables list of a report, then deselect all the observables that are not Domain-Name. There is no tool bar : 
image

Archidoit:
And if you select entities of different types one by one, no enrichement button. But if you select them via the Shift button, the enrichemnt button is present.

- Select entities with the shift key : 
image

- Select entities one by one : 
image

CelineSebe:
> Select all entities in the observables list of a report, then deselect all the observables that are not Domain-Name. There is no tool bar : image

This bug already exists in testing. I'm looking for a solution for this new bug as well.

nino-filigran:
After a discussion with @CelineSebe we'll mirror the behavior that exists in other screens as in: 
- "select all" should not present the enrichment CTA, even if all entities are the same type.
- enabling one by one (or shift click) of entities **of same type** should display the enrichment CTA
- selecting through shift click/one by one mutliple entities **of different type** should not display the enrichment CTA. But as soon as the selection is only entities of the same type, the CTA should be displayed.

Archidoit:
@nino-filigran You can do a 'select all' after having applied a filter 'entity_type = X' to facilitate the selection. Maybe this scenario can display CTA...

nino-filigran:
@Archidoit I agree it's better this way. It's not a behavior currently available in the Observable section though of the app (not in container).
I guess it's safe to say that  in a container, the quantity of entities of different type would be limited. Given we can shift+select, this is a nice to have for me. 
@CelineSebe can you maybe assess how difficult it would be to it? And if we're talking about more than 1d of work, then feel free to drop this.

Archidoit:
Select entities of different types with Shift. The Enrichment button appear : 
image

Archidoit (approved):

PR #7365, opened 2024-06-13
[frontend] disable redirection for delete notification lines (#7308)

### LINKED ISSUE #7308: Disable link of the line for delete notification
## Description

image

When clicking on the line => 404.

We should remove the link in the line of delete event, and having a default pointer.

Archidoit:
issue: https://github.com/OpenCTI-Platform/opencti/issues/7308


CelineSebe:
Tested ok ✅
Suggestion: When the trash button is clicked, there is no dialog box to confirm the deletion.

CelineSebe (approved):

PR #7368, opened 2024-06-13
[frontend] E2E tests on Dashboard - chunk 2 (#6704)

### LINKED ISSUE #6704: Test E2E of dashboards

lndrtrbn:
### Proposed changes

* Add some tests on manipulating widgets
* Add tests to change access restriction of dashboards

### Related issues

* #6704 

### Checklist


opencti-platform/opencti-front/tests_e2e/model/DashboardWidgets.pageModel.ts
```diff
@@ -65,20 +67,12 @@ export default class DashboardWidgetsPageModel {
...
-    await this.filters.addFilter('Entity type', 'Malware', false);
-    await this.filters.addFilter('Label', 'e2e');
-    await this.validateFilters();
-    await this.titleField.fill('Number of malwares');
-    return this.createWidget();
+  getWidgetNumberValue(name: string, value: string) {
+    return this.page.getByRole('heading', { name }).locator('..').getByText(value);
   }
 
+  // region Premade widgets
```
aHenryJard, line 74: What is premade ?

opencti-platform/opencti-front/tests_e2e/dashboard/dashboard.spec.ts
```diff
@@ -4,9 +4,17 @@ import DashboardPage from '../model/dashboard.pageModel';
...
 import DashboardWidgetsPageModel from '../model/DashboardWidgets.pageModel';
+import TopMenuProfilePage from '../model/menu/topMenuProfile.pageModel';
+import LoginFormPageModel from '../model/form/loginForm.pageModel';
 import LeftBarPage from '../model/menu/leftBar.pageModel';
+import AccessRestrictionPageModel from '../model/AccessRestriction.pageModel';
 import MalwareDetailsPage from '../model/malwareDetails.pageModel';
 
+// Because of login/logout stuff in access restriction test below, running
+// both in parallel make conflicts.
+// Need to be resolved in an other way but for now we do it like this.
```
aHenryJard, line 15: I wonder if we could tag tests starting from now, for example '@dashboard' in the current PR, to ease a  parrallel run of test in the future. It seems to be a way to have independent suite in the way we are implementing end to end today. Also maybe it will be usefull to run locally only a subset of them.

https://playwright.dev/docs/test-annotations#tag-tests

And maybe one tag for the init part if it make sense, so it can be run in addition of any test tag.

opencti-platform/opencti-front/tests_e2e/model/DashboardWidgets.pageModel.ts
lndrtrbn: It's just to tell that below are functions allowing you to add a preconfigured widget in the dashboard in one line without the need to configure it yourself. Typically for common widgets to be able to add them quicky

opencti-platform/opencti-front/tests_e2e/dashboard/dashboard.spec.ts
lndrtrbn: I didn't know this tag feature, could be useful indeed. Maybe it can help us run some in parallel but in that case I think tags should be more like @serie or @parallel more than the name of the feature?

For the init part, even without tag it runs before any test

aHenryJard (approved):

Issue #6609, opened 2024-04-06
feat(customization): implement monochrome (single color) mode for labels and entity types


SamuelHassine:
## Use case

Implement mono-chrome (single color) mode (in settings) for labels and entity types

Bonsai8863:
I can work on this issue. I'll add a toggle option to user's profile settings.

Jipegien:
following meeting of June 11th, scope is only labels.

labo-flg:
#7373 has been merged. We have the monochrome option ready backend side (saved in user's profile), and the toggle button is commented out until the feature is actually implemented in the UI.

PR #7373, opened 2024-06-13
[backend/frontend] Marking label and entity type chip color adjustment (#6609)

Bonsai8863:
### Proposed changes

* Adjusted marking labels to be more readable
* Adjusted entity/relationship type chips to be more readable
* Moved most instances of entity/relationship type chips to improved ItemEntityType component

### Related issues

* #6609

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/components/ItemEntityType.tsx
+++ b/opencti-platform/opencti-front/src/components/ItemEntityType.tsx
@@ -34,23 +34,25 @@ const useStyles = makeStyles(() => ({
 interface ItemEntityTypeProps {
   entityType: string;
   maxLength?: number;
-  variant?: string;
+  inList?: boolean;
   showIcon?: boolean;
   isRestricted?: boolean;
-  styles?: React.CSSProperties;
+  style?: React.CSSProperties;
+  size?: 'small' | 'medium' | 'large'
 }
 
 const ItemEntityType: FunctionComponent<ItemEntityTypeProps> = ({
-  variant = 'inList',
+  inList = true,
   maxLength,
   entityType,
   showIcon = false,
   isRestricted = false,
-  styles = {},
+  style = {},
+  size,
 }) => {
   const classes = useStyles();
   const { t_i18n } = useFormatter();
-  const style = variant === 'inList' ? classes.chipInList : classes.chip;
+  const rootStyle = inList ? classes.chipInList : classes.chip;
 
   const { isRelationship: checkIsRelationship } = useSchema();
   const isRelationship = checkIsRelationship(entityType);
@@ -60,18 +62,32 @@ const ItemEntityType: FunctionComponent<ItemEntityTypeProps> = ({
     ? ThemeDark()
     : ThemeLight();
   const getStyle = () => {
+    let width;
+    switch (size) {
+      case 'small':
+        width = 100;
+        break;
+      case 'large':
+        width = 140;
+        break;
+      case 'medium':
+      default:
+        width = 120;
+    }
     if (isRestricted) {
       const restrictedColor = itemColor('Restricted');
       return {
         backgroundColor: theme.palette.background.default,
         color: theme.palette.chip.main,
         border: `1px solid ${restrictedColor}`,
+        width,
       };
     }
     return {
       backgroundColor: theme.palette.background.default,
       color: isRelationship ? theme.palette.primary.main : theme.palette.chip.main,
       border: `1px solid ${isRelationship ? theme.palette.primary.main : itemColor(entityType)}`,
+      width,
     };
   };
   const getIcon = () => {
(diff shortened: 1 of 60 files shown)


labo-flg:
I've done some tests, it looks great!

When an item in a list has multiple markings, there is a tooltip that shows the full list. I think we could have the same style in the tooltip as in the other cases:
image

opencti-platform/opencti-front/src/components/ItemEntityType.tsx
```diff
@@ -28,28 +33,76 @@ const useStyles = makeStyles(() => ({
...
+  const getIcon = () => {
+    if (showIcon && !isRelationship) {
+      return (
+        <ItemIcon
+          variant="inline"
+          type={isRestricted ? 'restricted' : entityType}
+        />
+      );
+    }
+    return <></>;
```
labo-flg:
```suggestion
    return null;
```

labo-flg:
TLP;CLEAR is set as `#ffffff` but appears like this:

image

How could we improve this?

labo-flg:
The style is off in the knowledge graph view, when selecting an item:

image

opencti-platform/opencti-front/src/components/ItemEntityType.tsx
```diff
@@ -28,28 +33,76 @@ const useStyles = makeStyles(() => ({
...
+  styles?: React.CSSProperties;
 }
 
 const ItemEntityType: FunctionComponent<ItemEntityTypeProps> = ({
   variant = 'inList',
+  maxLength,
   entityType,
+  showIcon = false,
+  isRestricted = false,
+  styles = {},
```
labo-flg: Is there a reason why `styles` and not `style`?

opencti-platform/opencti-front/src/components/ItemEntityType.tsx
```diff
@@ -28,28 +33,76 @@ const useStyles = makeStyles(() => ({
...
 
   return (
     <Chip
       classes={{ root: style }}
       style={{
-        backgroundColor: hexToRGB(itemColor(entityType), 0.08),
-        color: itemColor(entityType),
-        border: `1px solid ${itemColor(entityType)}`,
+        ...styles,
+        ...getStyle(),
```
labo-flg: I would rather make the prop style capable of overwriting any default style (this is standard behavior), so:
```suggestion
        ...getStyle(),
        ...styles,
```

opencti-platform/opencti-front/src/private/components/common/stix_core_relationships/EntityStixCoreRelationshipLineAll.jsx
```diff
@@ -113,48 +105,20 @@ class EntityStixCoreRelationshipLineAllComponent extends Component {
...
               >
-                <Chip
-                  variant="outlined"
-                  classes={{ root: classes.chipInList }}
-                  style={{ width: 120 }}
-                  color="primary"
-                  label={t(`relationship_${node.relationship_type}`)}
+                <ItemEntityType
+                  entityType={node.relationship_type}
+                  styles={{ width: 120 }}
```
labo-flg: I think 120 is the default width anyway.

I think you are using either 120 or 140.

If this is really necessary, I would suggest to add a prop `size` to `ItemEntityType` with 'medium' giving 120 width and 'large' giving 140. if necessary, 'small' for 100 width.

My point is to avoid hardcoded values copy-pasted throughout the app, to keep UI consistent.

opencti-platform/opencti-front/src/components/ItemEntityType.tsx
```diff
@@ -28,28 +33,92 @@ const useStyles = makeStyles(() => ({
...
 
 const ItemEntityType: FunctionComponent<ItemEntityTypeProps> = ({
-  variant = 'inList',
+  inList = true,
+  maxLength,
   entityType,
+  showIcon = false,
+  isRestricted = false,
+  style = {},
+  size,
```
labo-flg:
```suggestion
  size = 'medium',
```
to facilitate code comprehension (otherwise you have to dig into the code, see the switch)

labo-flg:
All tests are good now, good job 🙌

labo-flg:
I've tested with or without monochrome_labels option enabled. This is not used for now in the code, am I right?

If so, if we need to release a new version with this PR being merged, people would get the option visible in profile and probably wonder what it does... until another PR actually implements the option.

I suggest we put the field in the profile view under a new feature flag until the feature is complete.

labo-flg:
> I suggest we put the field in the profile view under a new feature flag until the feature is complete.

for the record ; no feature flag. We keep the code change backend side even if the flag `monochrome_labels` is unused for now. The toggle button is commented out though, it has no effect and that could be confusing.

labo-flg (approved):

PR #7399, opened 2024-06-17
[backend] fix max shareable markings migration (#7392)

### LINKED ISSUE #7392: Error at max shareable marking migration if version < 6.0
## Description

If upgrading from OpenCTI 6.0 to OpenCTI 6.1.11, the migration does not work.
You need to upgrade to 6.1.10 and then to 6.1.11.

## Error message

image

Archidoit:
PR content:
- fix issue: https://github.com/OpenCTI-Platform/opencti/issues/7392
- delete the platform_max_shareable_markings field

### DIFF
--- a/opencti-platform/opencti-graphql/src/migrations/1717060854132-max-shareable-marking-in-groups.js
+++ b/opencti-platform/opencti-graphql/src/migrations/1717060854132-max-shareable-marking-in-groups.js
@@ -48,11 +48,11 @@ export const up = async (next) => {
 
   await Promise.all(groupMaxMarkingRelationCreationsPromises);
 
-  // remove platform_data_sharing_max_markings from settings
+  // set platform_data_sharing_max_markings to null in settings
   const updateQuery = {
     script: {
-      params: { null: null },
-      source: 'ctx._source.platform_data_sharing_max_markings = params.null',
+      params: { fieldToRemove: 'platform_data_sharing_max_markings' },
+      source: 'ctx._source.remove(params.fieldToRemove)',
     },
     query: {
       bool: {
@@ -64,11 +64,10 @@ export const up = async (next) => {
   };
   await elUpdateByQueryForMigration(
     message,
-    [READ_INDEX_INTERNAL_OBJECTS],
+    READ_INDEX_INTERNAL_OBJECTS,
     updateQuery
   );
 
-  // do your migration
   logApp.info(`${message} > done`);
   next();
 };
--- a/opencti-platform/opencti-graphql/src/migrations/1713449762001-add-platform-data-sharing-max-markings.js
+++ b/opencti-platform/opencti-graphql/src/migrations/1713449762001-add-platform-data-sharing-max-markings.js
@@ -1,15 +1,21 @@
-import { executionContext, SYSTEM_USER } from '../utils/access';
-import { getSettings, settingsEditField } from '../domain/settings';
-import { logApp } from '../config/conf';
+// import { executionContext, SYSTEM_USER } from '../utils/access';
+// import { getSettings, settingsEditField } from '../domain/settings';
+// import { logApp } from '../config/conf';
 
 export const up = async (next) => {
-  logApp.info('[MIGRATION] Add platform data sharing max markings');
-  const context = executionContext('migration');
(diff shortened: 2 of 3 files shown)


marieflorescontact (approved):

opencti-platform/opencti-graphql/src/migrations/1717060854132-max-shareable-marking-in-groups.js
```diff
@@ -48,27 +48,30 @@ export const up = async (next) => {
...
+      params: { fieldToRemove: 'platform_data_sharing_max_markings' },
+      source: 'ctx._source.remove(params.fieldToRemove)',
     },
     query: {
       bool: {
         must: [
           { term: { 'entity_type.keyword': { value: 'Settings' } } },
         ],
+        should: [
+          { exists: { field: 'platform_data_sharing_max_markings' } },
```
SouadHadjiat: I guess this query works already, but I wonder why not add 
{ exists: { field: 'platform_data_sharing_max_markings' } } directly in "must" list just above like this:

```
query: {
      bool: {
        must: [
          { term: { 'entity_type.keyword': { value: 'Settings' } } },
          { exists: { field: 'platform_data_sharing_max_markings' } },
        ],
       },
    },
```

Archidoit: yes you're right ! fixed :)

SouadHadjiat (changes requested):
The migration doesn't work for me locally, the field is still present in my Settings with null, exists condition doesn't work in this case : https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-exists-query.html exists will not match null or [] values.

=> I think that you should just remove the field, without the exists condition, just the term query on entity_type "Settings"

opencti-platform/opencti-graphql/src/migrations/1717060854132-max-shareable-marking-in-groups.js
```diff
@@ -48,27 +48,27 @@ export const up = async (next) => {
...
-      params: { null: null },
-      source: 'ctx._source.platform_data_sharing_max_markings = params.null',
+      params: { fieldToRemove: 'platform_data_sharing_max_markings' },
+      source: 'ctx._source.remove(params.fieldToRemove)',
     },
     query: {
       bool: {
         must: [
           { term: { 'entity_type.keyword': { value: 'Settings' } } },
+          { exists: { field: 'platform_data_sharing_max_markings' } },
```
SouadHadjiat: exists condition doesn't match null or [] values, so I think it should be removed

opencti-platform/opencti-graphql/src/migrations/1718631740774-remove-platform-max-shareable-markings.js
```diff
@@ -0,0 +1,34 @@
...
+  const updateQuery = {
+    script: {
+      params: { fieldToRemove: 'platform_data_sharing_max_markings' },
+      source: 'ctx._source.remove(params.fieldToRemove)',
+    },
+    query: {
+      bool: {
+        must: [
+          { term: { 'entity_type.keyword': { value: 'Settings' } } },
+          { exists: { field: 'platform_data_sharing_max_markings' } },
```
SouadHadjiat: same here: remove exists condition

opencti-platform/opencti-graphql/src/migrations/1717060854132-max-shareable-marking-in-groups.js
Archidoit: I'm gonna modify the condition with: field exist OR is field is null

opencti-platform/opencti-graphql/src/migrations/1718631740774-remove-platform-max-shareable-markings.js
Archidoit: done

SouadHadjiat (approved):

SouadHadjiat:
make sure before merging that the new migration has an up to date timestamp compared to latest migrations on master

Issue #7406, opened 2024-06-18
ExternalReferences with automatic creation is broken


Kedae:
We have 2 main issues:

1. The Automatic references at file upload is broken because of the missing fileId attributes in external-references
2. When creatin an entity with that setting and associating a file at creation tries to create a "externalReferences" attributes in elastic.
-> In "recent" platforms : this is broken
-> In "old" platforms : the mapping allows the creation with this attribute and it leads to errors

/!\ Also check the stixCoreObject@stixCoreObjectImportPush method that also uses the settings and the attribute

PR #7417, opened 2024-06-18
[backend] Fix handling of platform_entity_files_ref config (#7406)

JeremyCloarec:
### Proposed changes

* Added fileId to external reference schema attributes
* Fixed creation of external ref in createEntityRaw when platform_entity_files_ref is configured
* Fixed updater for external ref deletion

### Related issues

* #7406 

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/analyses/external_references/StixCoreObjectExternalReferencesLines.tsx
+++ b/opencti-platform/opencti-front/src/private/components/analyses/external_references/StixCoreObjectExternalReferencesLines.tsx
@@ -395,6 +395,7 @@ StixCoreObjectExternalReferencesLinesContainerProps
                             isExternalReferenceAttachment={isFileAttached}
                             handleRemove={() => handleOpenDialog(externalReferenceEdge)
                             }
+                            objectId={stixCoreObjectId}
                           />
                         </Security>
                       </ListItemSecondaryAction>
--- a/opencti-platform/opencti-graphql/src/database/middleware.js
+++ b/opencti-platform/opencti-graphql/src/database/middleware.js
@@ -99,7 +99,6 @@ import {
   ID_STANDARD,
   IDS_STIX,
   INPUT_CREATED_BY,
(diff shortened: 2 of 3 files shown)


Archidoit:
Creation works but we can't delete or remove the external ref from the entity.

- Enforce external ref creation at file upload for Malwares
- Create a malware with a file
- An external reference associated with the file is well created and well associated to the malware
- But you can't remove / delete the external ref from the malware

image

If you click on 'delete', the file is deleted but the external ref is still there : 
image

If you try to delete again, error message : 
image

Kedae:
Have you checked stixCoreObject@stixCoreObjectsImportPush ?

JeremyCloarec:
Yes, tested both entity creation and import push behavior. stixCoreObjectsImportPush was working properly once fileId was added back as an external ref attribute

Archidoit:
We have only the choices 'update' or 'delete'. 
image

We should also have 'remove', like it is proposed when we add the external ref after creation via the 'add' button : 
image

What's more, when we click on 'delete', the external ref is from the object. But if you reload the external ref still apears (but without the file). And the external ref is still present (it has not been deleted from the platform) : 
image

Archidoit:
Create an object with a file, an external reference with the file is created.

Go to an entity overview.
Click on 'add external reference' : 
image

You can't select the external references created via the file upload (see test.txt in the list of the screenshot below) : 
image

Archidoit:
Create an entity with a file. The external ref is well created.
Click on Update the external ref : 
image

Change the source name. An error occurs for the URL field : 
image

But the external ref is well updated if we leave the edition form : 
image

JeremyCloarec:
URL validation problem has been moved to a separate issue: https://github.com/OpenCTI-Platform/opencti/issues/7431

Archidoit (approved):

JeremyCloarec:
The questions about the expected behavior of these external references have been moved to this issue https://github.com/OpenCTI-Platform/opencti/issues/7432

Issue #7291, opened 2024-06-09
Bad FR translation encrypted archives


SamuelHassine:
## Description

![image](https://github.com/OpenCTI-Platform/opencti/assets/1334279/305f1d45-26f2-46fd-b4a6-a020ab29cf21)

Should be "Archive chiffrées". Also the next in English is not translated.

PR #7423, opened 2024-06-19
[frontend] Fix Bad FR translation encrypted archives (#7291)

SarahBocognano:
### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/7291


Archidoit (approved):

SamuelHassine:
In the text, it should also be "chiffré" and not "crypté". 

https://chiffrer.info/

Issue #7477, opened 2024-06-25
OpenCTI fails to detect successfully authenticated OpenID Connect SSO via ADFS


animedbz16:
## Description

After successfully configuring OpenCTI to leverage ADFS via OpenID Connect for SSO, a user can successfully log in but then OpenCTI will fail to create a new user because it is not able to detect the email / name / etc from the userinfo.

The relevant code block is here:
https://github.com/OpenCTI-Platform/opencti/blob/6.1.12/opencti-platform/opencti-graphql/src/config/providers.js#L350-L367

When I inspected the data obtained inside of the `userinfo` object, it only contains an object like the following `{ sub: <BASE64 encoded data> }`, so attempting to access a key of `email`, `name`, etc will yield an undefined value.

When looking at ADFS documentation, it seems that the userinfo endpoint will not return additional claims see link and screenshot below:

> https://learn.microsoft.com/en-us/windows-server/identity/ad-fs/overview/ad-fs-faq#i-m-trying-to-get-more-claims-on-the-userinfo-endpoint--but-it-s-only-returning-subject--how-can-i-get-more-claims-
>
> ![image](https://github.com/OpenCTI-Platform/opencti/assets/2822632/5f5a4aa4-4b90-4e32-9569-1cdbc3110a17)

Microsoft recommends that if you would like to access claims, that you should do so by including them into the `id_token`:

> https://learn.microsoft.com/en-us/entra/identity-platform/userinfo#consider-using-an-id-token-instead
>
> ![image](https://github.com/OpenCTI-Platform/opencti/assets/2822632/8662791e-a2fd-4d6c-875c-ca3866786f9b)

With that in mind, I inspected the decoded JWT for the `id_token` and it does include all the claims for `email`, `name`, etc. 

With that in mind, I propose that a change should be made to support looking for these values within the id_token as well as the current userinfo so that authentication can be supported for ADFS

## Environment

1. OS (where OpenCTI server runs): Docker
2. OpenCTI version: 6.1.12
3. OpenCTI client: Docker
4. Other environment details:

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Configure OpenCTI for OpenID Connect SSO provided by Microsoft ADFS
2. Successfully authenticate via ADFS
3. OpenCTI will fail to create user

## Expected Output

OpenCTI should be able to detect successfully authenticated users

## Actual Output

OpenCTI is not detecting successfully authenticated users

CountyWorker:
[...]

PR #7478, opened 2024-06-25
[Backend] Add OpenID Connect SSO support for Microsoft ADFS to get user claims from the id_token

animedbz16:
https://github.com/OpenCTI-Platform/opencti/issues/7477

### Proposed changes
* If the user claims are not in userinfo, then attempt to get them from the id_token

### Related issues
* https://github.com/OpenCTI-Platform/opencti/issues/7477

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/config/providers.js
+++ b/opencti-platform/opencti-graphql/src/config/providers.js
@@ -368,10 +368,14 @@ for (let i = 0; i < providerKeys.length; i += 1) {
               const emailAttribute = mappedConfig.email_attribute ?? 'email';
               const firstnameAttribute = mappedConfig.firstname_attribute ?? 'given_name';
               const lastnameAttribute = mappedConfig.lastname_attribute ?? 'family_name';
-              const name = userinfo[nameAttribute];
-              const email = userinfo[emailAttribute];
-              const firstname = userinfo[firstnameAttribute];
-              const lastname = userinfo[lastnameAttribute];
+              const get_user_attributes_from_id_token = mappedConfig.get_user_attributes_from_id_token ?? false;
+
+              const user_attribute_obj = get_user_attributes_from_id_token ? jwtDecode(tokenset.id_token) : userinfo;
+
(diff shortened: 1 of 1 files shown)


animedbz16:
Spoke with @SarahBocognano and @marieflorescontact in Slack on whether or not test need to be added since the CodeCov report was complaining that the code I have added is not being tested, but it appears that most of the provider.js code does not have any related tests and was mentioned that it is likely acceptable to not add tests within this PR since there the provider.js tests likely would be added holistically at some future point in time.

I have refactored the code slightly to ensure that the id_token is not decoded unless any of the userinfo attributes do not exist and then attempt to obtain them from the decoded id_token, so the code I have added should be bypassed entirely unless the userinfo data does not exist.

CountyWorker:
We need this fixed merged. This is a blocker for us. As explained in the connected issue: OpenCTI just picks the "sub" field which is base64 encrypted data... if it just picked another field it would be great. Thanks.

richard-julien:
Hi @animedbz16,
Can you change your PR to check a boolean in the mapped config instead of testing if the attributes are empty?
Something like 
```
if (mappedConfig.get_basic_info_from_token === true) {
   const decodedIdToken = jwtDecode(tokenset.id_token);
   name = decodedIdToken[nameAttribute];
   email = decodedIdToken[emailAttribute];
   firstname = decodedIdToken[firstnameAttribute];
   lastname = ldecodedIdToken[lastnameAttribute];
} 
```
Thanks

animedbz16:
Yes, I will look to make those updates to support a new config / environment variable and default this to false since this would require the user to explicitly configure this true to enable this behavior and if not continue to function as it does today.

animedbz16:
@richard-julien I refactored this to add a `get_user_attributes_from_id_token` configuration value and utilize the decrypted id_token if this is set otherwise to use the userinfo object.

Let me know if there is anything else that would need to be done here. I assume another PR would likely need to be created to document this new configuration option once it is approved  / merged.

PR #7490, opened 2024-06-26
[backend] on Artifact creation pass mime_type to upload (#7486)(#7434)

### LINKED ISSUE #7434: Bad .docx file upload at artifact creation
## Description
If you create an artifact with a .docx file, the file name and extensions are changed, leading to errors.

## Reproducible Steps
- Create an artifact by selecting a .docx file : 
image

- The artifact is created but the file name is incorrect and always 'Content_Types.xml'
image

- In the Data tab of the created artifact, the file name and extension are not correct and the import failed
image

### LINKED ISSUE #7486: MIME type in uploaded artifacts is not correct and lead to multiple errors
## Description

1. Take an exe file, rename it without extension:
[calc-exe.zip](https://github.com/user-attachments/files/15986379/calc-exe.zip)

2. Upload it in the artifact section

3. Result:

MIME Type of the artifact is correct:

image

MIME Type of the associated file is always "text/plain":

image

## Other example with the connector

Correct MIME Type here:
image

Incorrect in the associated file:
image

labo-flg:
### Proposed changes

when creating artifact, we have the MIME type. We simply pass it down to the upload function as option ; if defined we use otherwise we try to guess it.

### Related issues

closes #7486
closes [7434](https://github.com/OpenCTI-Platform/opencti/issues/7434)

### DIFF
--- a/opencti-platform/opencti-graphql/src/database/file-storage.js
+++ b/opencti-platform/opencti-graphql/src/database/file-storage.js
@@ -385,7 +385,7 @@ export const uploadJobImport = async (context, user, fileId, fileMime, entityId,
 
 // Please consider using file-storage-helper#uploadToStorage() instead.
 export const upload = async (context, user, filePath, fileUpload, opts) => {
-  const { entity, meta = {}, noTriggerImport = false, errorOnExisting = false, file_markings = [], importContextEntities = [], mime_type } = opts;
+  const { entity, meta = {}, noTriggerImport = false, errorOnExisting = false, file_markings = [], importContextEntities = [] } = opts;
   const metadata = { ...meta };
   if (!metadata.version) {
     metadata.version = now();
@@ -407,7 +407,7 @@ export const upload = async (context, user, filePath, fileUpload, opts) => {
 
   // Upload the data
   const readStream = createReadStream();
-  const fileMime = mime_type ?? guessMimeType(key);
+  const fileMime = metadata.mime_type ?? guessMimeType(key);
   const fullMetadata = {
     ...metadata,
     filename: encodeURIComponent(truncatedFileName),
--- a/opencti-platform/opencti-graphql/src/database/file-storage.js
+++ b/opencti-platform/opencti-graphql/src/database/file-storage.js
@@ -385,7 +385,7 @@ export const uploadJobImport = async (context, user, fileId, fileMime, entityId,
 
 // Please consider using file-storage-helper#uploadToStorage() instead.
 export const upload = async (context, user, filePath, fileUpload, opts) => {
-  const { entity, meta = {}, noTriggerImport = false, errorOnExisting = false, file_markings = [], importContextEntities = [] } = opts;
+  const { entity, meta = {}, noTriggerImport = false, errorOnExisting = false, file_markings = [], importContextEntities = [], mime_type } = opts;
   const metadata = { ...meta };
   if (!metadata.version) {
     metadata.version = now();
@@ -407,7 +407,7 @@ export const upload = async (context, user, filePath, fileUpload, opts) => {
 
   // Upload the data
   const readStream = createReadStream();
-  const fileMime = guessMimeType(key);
+  const fileMime = mime_type ?? guessMimeType(key);
   const fullMetadata = {
     ...metadata,
     filename: encodeURIComponent(truncatedFileName),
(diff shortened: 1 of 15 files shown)


opencti-platform/opencti-graphql/src/database/file-storage.js
```diff
@@ -407,7 +407,7 @@ export const upload = async (context, user, filePath, fileUpload, opts) => {
 
   // Upload the data
   const readStream = createReadStream();
-  const fileMime = guessMimeType(key);
+  const fileMime = mime_type ?? guessMimeType(key);
```
labo-flg: I did not change the behavior of guessMimeType ; it still returns plain/text by defaut.
Do we want to get rid of this too ?
What happens if there is no MIME type given and guess is impossible (maybe that's a dumb too theoretical question)

SamuelHassine: I think we can still try to guess and fallback but the fallback should be application/octet-stream rather than text/plain, what do you think?

opencti-platform/opencti-graphql/src/database/file-storage.js
```diff
@@ -385,7 +385,7 @@ export const uploadJobImport = async (context, user, fileId, fileMime, entityId,
 
 // Please consider using file-storage-helper#uploadToStorage() instead.
 export const upload = async (context, user, filePath, fileUpload, opts) => {
-  const { entity, meta = {}, noTriggerImport = false, errorOnExisting = false, file_markings = [], importContextEntities = [] } = opts;
+  const { entity, meta = {}, noTriggerImport = false, errorOnExisting = false, file_markings = [], importContextEntities = [], mime_type } = opts;
```
Kedae: I would have put the mime_type in the metaData personnaly, but that's a detail

opencti-platform/opencti-graphql/src/database/file-storage.js
Kedae: That would make more sens as a plain text generally have a file extension

opencti-platform/opencti-graphql/src/database/file-storage.js
labo-flg: metas are then saved in db, that's a change in model

opencti-platform/opencti-graphql/src/database/file-storage.js
labo-flg: ok !

labo-flg:
@aHenryJard found another issue ; I'm putting back to draft

aHenryJard:
All good for me now, tested locally with connector malwarebaazar

opencti-platform/opencti-graphql/src/database/file-storage.js
```diff
@@ -247,15 +247,16 @@ export const loadFile = async (user, fileS3Path, opts = {}) => {
...
       id: fileS3Path,
-      name: getFileName(fileS3Path),
+      name: onlyFilename,
       size: object.ContentLength,
       information: '',
       lastModified: object.LastModified,
       lastModifiedSinceMin: sinceNowInMinutes(object.LastModified),
       uploadStatus: 'complete',
-      metaData: { mimetype: mimeTypeResolved },
+      metaData: { mimetype: mimeTypeResolved, filename: onlyFilename },
```
aHenryJard: this `filename` is required at least by the download as encrypted archive feature.

opencti-platform/opencti-graphql/src/database/file-storage.js
```diff
@@ -247,15 +247,16 @@ export const loadFile = async (user, fileS3Path, opts = {}) => {
       Key: fileS3Path
     }));
     const mimeTypeResolved = guessMimeType(fileS3Path);
```
SouadHadjiat: we still only guess the mimeType here

opencti-platform/opencti-graphql/src/database/file-storage.js
```diff
@@ -237,25 +237,91 @@ export const storeFileConverter = (user, file) => {
...
+      lastModified = fileInformationFromDB.lastModified;
+      lastModifiedSinceMin = lastModified; // ??
+    } else {
+      // Else try to get them from S3 instead
+      const object = await s3Client.send(new s3.HeadObjectCommand({
+        Bucket: bucketName,
+        Key: fileS3Path
+      }));
+      metaData = {
+        version: object.Metadata.version,
```
SouadHadjiat: not sure this would work, since we don't store Metadata anymore in S3, or is it just a fallback for previously stored files ? I wonder if object.Metadata would be defined for new files.

Kedae (approved):

PR #7493, opened 2024-06-26
[backend] Migration to remove representative in DB (#7484)

### LINKED ISSUE #7484: In some old instances, we have representative (main / secondary) indexed which leads to issues
## Description

In some old instances, we have representative (main / secondary) indexed which leads to issues.

We need to create a migration to cleanup those attributes:

image

SouadHadjiat:
### Proposed changes

* add migration to remove representative
* remove representative from partialInstanceWithInputs, wasn't used anywhere.

### Related issues

* #7484 

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/database/engine.js
+++ b/opencti-platform/opencti-graphql/src/database/engine.js
@@ -3605,7 +3605,7 @@ export const elUpdateElement = async (instance) => {
   // If entity with a name, must update connections
   let connectionPromise = Promise.resolve();
   if (esData.name && isStixObject(instance.entity_type)) {
-    connectionPromise = elUpdateConnectionsOfElement(instance.internal_id, { name: esData.representative.main });
+    connectionPromise = elUpdateConnectionsOfElement(instance.internal_id, { name: extractEntityRepresentativeName(esData) });
   }
   return Promise.all([replacePromise, connectionPromise]);
 };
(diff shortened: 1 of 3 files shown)


richard-julien:
Also need to remove a specific code introduce in #6319

```
export const extractRepresentative = (entityData) => {
  // the representative can already be given (happens when the internals set it to "Restricted")
  if (entityData.representative) {
    return entityData.representative;
  }
```

SouadHadjiat:
> Also need to remove a specific code introduce in #6319
> 
> ```
> export const extractRepresentative = (entityData) => {
>   // the representative can already be given (happens when the internals set it to "Restricted")
>   if (entityData.representative) {
>     return entityData.representative;
>   }
> ```

seems to be linked to this part in middleware (convertAggregateDistributions) : 

```
return {
        ...n,
        entity: {
          id: element.id,
          entity_type: element.entity_type,
          parent_types: element.parent_types,
          representative: { main: 'Restricted', secondary: 'Restricted' }
        }
      };
```

richard-julien:
I missed that :) Thanks for the pointer. 
For this not sure we have an alternative. If you can quickly check with Laurent if he have an idea.
If not we can keep it like this.

SouadHadjiat:
I just saw this method where I wonder why we added a representative (this is what caused updates to index representative)

```
const partialInstanceWithInputs = (instance, inputs) => {
  const inputData = updatedInputsToData(instance, inputs);
  return {
    _index: instance._index,
    internal_id: instance.internal_id,
    entity_type: instance.entity_type,
    representative: extractRepresentative(instance),
    ...inputData,
  };
};
```
Can we remove it here ? @richard-julien

Kedae (approved):

PR #7548, opened 2024-07-01
[frontend] add tooltips to marking chips (#7519)

### LINKED ISSUE #7519: Marking chips are missing a tooltip for when text is truncated
## Description

It happens that, depending on the view and the marking label, the Chip content is truncated.

Example:
image

In that case, we do not know for sure which marking it is.

Standard UI/UX would be to have a tooltip displayed when hovering truncated label, displaying the whole text, like this:

image

## Environment

ALL

## Reproducible Steps

Seen in Data > Entities for a marking with long text label.

labo-flg:
Long texts in Chips end up truncated and without tooltips we cannot know the real text. This can be misleading when several markings start with the same texts for instance.

Adding mui basic tooltip around the chip will do the trick.

image

## details

I enable this html title when the marking chip is alone, or inside the themed tooltip when several markings are displayed.

Closes #7519

### DIFF
--- a/opencti-platform/opencti-front/src/components/ItemMarkings.jsx
+++ b/opencti-platform/opencti-front/src/components/ItemMarkings.jsx
@@ -6,6 +6,7 @@ import Badge from '@mui/material/Badge';
 import Chip from '@mui/material/Chip';
 import Grid from '@mui/material/Grid';
 import makeStyles from '@mui/styles/makeStyles';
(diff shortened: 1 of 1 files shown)


ValentinBouzinFiligran (approved):

labo-flg:
@ValentinBouzinFiligran I've changed the implementation after discussion with @Jipegien

PR #7557, opened 2024-07-01
[frontend] Root file updated from JS to TS (#7500)

### LINKED ISSUE #7500: Apply the Bulk Relationship POC on all required entities
From #7443 we need to apply the same changes to all needed entities defined by Product in notion page.
Chunks will be the different menus

ValentinBouzinFiligran:
### Proposed changes

* https://github.com/OpenCTI-Platform/opencti/issues/7500
*

### Related issues

* Update jsx root file to tsx file
* Bulk relation component expansion on stand by

### Checklist


lndrtrbn (commented):
In the case of `useParams() as ...` it would be better to define a type and use it for the generic of useParams `useParams<MyType>()` instead of hard casting

lndrtrbn (approved):

Issue #7431, opened 2024-06-20
External reference URL are not properly validated during creation


JeremyCloarec:
## Description

The validation doesn't recognize some valid URLs. 
URLs in the form of http://localhost:4000/storage/get/* are not accepted for exemple, as well as ftp addresses.

## Environment

1. OS (where OpenCTI server runs): { e.g. Mac OS 10, Windows 10, Ubuntu 16.4, etc. }
2. OpenCTI version: { e.g. OpenCTI 1.0.2 }
3. OpenCTI client: { e.g. frontend or python }
4. Other environment details:

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Go to External references
2. Click on create an external ref 
3.  Set URL to http://localhost:4000/storage/get/blabla.txt

## Expected Output

URL is validated

## Actual Output

We get a message saying that 'The value must be an URL'

Archidoit:
Other scenario : 

Create an entity with a file. The external ref is well created.
Click on Update the external ref : 
![image](https://github.com/OpenCTI-Platform/opencti/assets/75783086/6b5bee91-abff-4400-b326-80716bb2116e)

Change the source name. An error occurs for the URL field : 
![image](https://github.com/OpenCTI-Platform/opencti/assets/75783086/13045933-08ff-4a5a-8a4c-9770dd5af8d9)

But the external ref is well updated if we leave the edition form : 
![image](https://github.com/OpenCTI-Platform/opencti/assets/75783086/526948c2-27b0-429b-ade1-96f81047dbf8)

PR #7593, opened 2024-07-04
[frontend] External reference URL are not properly validated during creation (#7431)

SarahBocognano:
### Proposed changes

* The "url" property of Yup was too strict, so I made a custom regex to accept all urls starting with "http" and "https" allowing us to accept local urls as "localhost:3000"

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/7431


Goumies (approved):

marieflorescontact (approved):
locally tested with lots of differents urls types => ok 👌

Issue #7400, opened 2024-06-17
CSV mapper Improvement Phase 1


nino-filigran:
## Use case

All tickets below represents the requests that we have received. It does not yet depict what will be done. Additionnally, some ideas are being investigated internally. Once the scope of the rework of CSV mappers will be clear, more details will be added regarding functionalities tackled/not tackled, solution taken. 

## Tackled in phase 1: 
- duplicate CSV mappers & CSV feeds: https://github.com/OpenCTI-Platform/opencti/issues/6279- 
- Ability to ingest CSV files where each lines can correspond to different objects https://github.com/OpenCTI-Platform/opencti/issues/4505

nino-filigran:
Ability to import entities that are present in multiple rows with various data for the same attribute https://github.com/OpenCTI-Platform/opencti/issues/8403 will be done in a minor release.

PR #7594, opened 2024-07-04
Improve CSV parser to stream file to handle very large volume of data

### LINKED ISSUE #7589: Improve CSV parser to stream file to handle very large volume of data
## Use case

Improve CSV parser to stream file to handle very large volume of data
- Adapt file download stream to use node:readline to handle the file line per line.
- Generate bundle every CSV_MAX_BUNDLE_SIZE_GENERATION streamed lines
- csvMapperTest query with content deprecated for csvMapperTest mutation with file
- global refactor of api pour testing and other api
- tested with a 1,4GB CSV file (generated with https://gist.github.com/richard-julien/9cac61bab212da1f3f086d5d31080525)

This feature will also allow to have mutliple row containing same values within the same CSV and these value will not be skipped anymore but will be upserted.

richard-julien:
See https://github.com/OpenCTI-Platform/opencti/issues/7589

### DIFF
--- a/opencti-platform/opencti-graphql/src/connector/importCsv/importCsv-connector.ts
+++ b/opencti-platform/opencti-graphql/src/connector/importCsv/importCsv-connector.ts
@@ -1,4 +1,5 @@
 import { Readable } from 'stream';
+import * as readline from 'node:readline';
 import type { SdkStream } from '@smithy/types/dist-types/serde';
 import conf, { logApp } from '../../config/conf';
 import { executionContext } from '../../utils/access';
@@ -13,10 +14,13 @@ import { parseCsvMapper } from '../../modules/internal/csvMapper/csvMapper-utils
 import type { ConnectorConfig } from '../connector';
 import { IMPORT_CSV_CONNECTOR } from './importCsv';
 import { internalLoadById } from '../../database/middleware-loader';
-import { DatabaseError, FunctionalError } from '../../config/errors';
+import { FunctionalError } from '../../config/errors';
 import { uploadToStorage } from '../../database/file-storage-helper';
+import type { CsvMapperParsed } from '../../modules/internal/csvMapper/csvMapper-types';
+import type { BasicStoreBase } from '../../types/store';
 
 const RETRY_CONNECTION_PERIOD = 10000;
+const BULK_LINE_PARSING_NUMBER = conf.get('import_csv_built_in_connector:bulk_creation_size') || 5000;
 
 const connectorConfig: ConnectorConfig = {
   id: 'IMPORT_CSV_BUILT_IN_CONNECTOR',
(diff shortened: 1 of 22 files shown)


opencti-platform/opencti-graphql/src/connector/importCsv/importCsv-connector.ts
```diff
@@ -36,6 +40,35 @@ const initImportCsvConnector = () => {
...
+    const validateBeforeImport = connectorConfig.config.validate_before_import;
+    if (validateBeforeImport) {
+      await updateExpectationsNumber(context, applicantUser, workId, 1);
+      const contentStream = Readable.from([JSON.stringify(bundle, null, '  ')]);
+      const file = {
+        createReadStream: () => contentStream,
+        filename: `${workId}.json`,
+        mimetype: 'application/json',
+      };
+      await uploadToStorage(context, applicantUser, 'import/pending', file, { entity });
```
SouadHadjiat, line 57: so it will upload an analyst workbench file with a bundle for the current lines that are being parsed ? does it append the file or replace it only ? From what I read it looks like it will replace it.

aHenryJard (commented):
I have an issue with count and number, so I have some doubt on the split data part. I tested will all french cities from https://www.data.gouv.fr/fr/datasets/villes-de-france/.

My csv file has an header and 39146 lines
```
wc -l cities.csv
39146 cities.csv
```
But in the import screen I see 32759 expected lines=>
image

And at the end in OpenCTI UI in data > entities I have 32759 (when I'm expecting 39145). I only got 1 error in one line of the csv.

I'm adding the csv mapper that I used:
image

So I think that the split in chunck of CSV_MAX_BUNDLE_SIZE_GENERATION is not correct somehow

opencti-platform/opencti-graphql/src/connector/importCsv/importCsv-connector.ts
```diff
@@ -56,42 +89,26 @@ const initImportCsvConnector = () => {
...
-              const content = Buffer.from(JSON.stringify(bundle), 'utf-8').toString('base64');
-              await pushToSync({
-                type: 'bundle',
-                update: true,
-                applicant_id: applicantId ?? OPENCTI_SYSTEM_UUID,
-                work_id: workId,
-                content
-              });
+        let lines: string[] = [];
+        const rl = readline.createInterface({ input: stream, crlfDelay: 5000 });
```
aHenryJard, line 93: Is this 5000, a delay between 2 ingestions of chuncks ?

richard-julien: No, its the delay for the streaming waiting for the end of line.

aHenryJard:
> I have an issue with count and number, so I have some doubt on the split data part. I tested will all french cities from https://www.data.gouv.fr/fr/datasets/villes-de-france/.
> 
> My csv file has an header and 39146 lines
> 
> ```
> wc -l cities.csv
> 39146 cities.csv
> ```
> 
> But in the import screen I see 32759 expected lines=> image
> 
> And at the end in OpenCTI UI in data > entities I have 32759 (when I'm expecting 39145). I only got 1 error in one line of the csv.
> 
> I'm adding the csv mapper that I used: image
> 
> So I think that the split in chunck of CSV_MAX_BUNDLE_SIZE_GENERATION is not correct somehow

Actually there is some duplicate, but I'm still missing 6 cities, I will look

opencti-platform/opencti-graphql/src/connector/importCsv/importCsv-connector.ts
richard-julien: Really good point, i need to dig in into this

aHenryJard:
@richard-julien FYI if it's fine with you, I'm going to update this branch and review in the following days/ weeks as part of https://github.com/OpenCTI-Platform/opencti/issues/7400 feature.

richard-julien:
> @richard-julien FYI if it's fine with you, I'm going to update this branch and review in the following days/ weeks as part of #7400 feature.

Ok great! Thanks.
We need to discuss about the generation of workbench that currently is not handled by my approach.
Maybe we need to wait for the draft for this.

aHenryJard:
Superseded by https://github.com/OpenCTI-Platform/opencti/pull/8638

PR #7599, opened 2024-07-04
Add option in deployment to disable the user interface (#6600)

### LINKED ISSUE #6600: Add option in deployment to disable the user interface
## Use case
By the docs and by experience the ingestion endpoint and the frontend endpoint should be separate containers / services. Currently the way to go is to deploy the platform container once for the frontend and once for ingestion. It would make the deployment more undestandable, easier to configure and reduce size and potential risks if there would be a separate container image that could be used for ingestion that only contains the graphql api, without the actual frontend.

## Current Workaround
Build the images myself (But not done)

## Proposed Solution
Build an additional container image with only the api.

## Additional Information
Currently I don't know if the frontend is ssr or if it only uses the api. If it only uses the api it would make even more sense to build 2 images, 1 with the api and 1 with a webserver and the frontend.

Yes, but would probably need help.

richard-julien:
See #6600

Option prevent the backend to send html base page and statics.

### DIFF
--- a/opencti-platform/opencti-graphql/src/config/conf.js
+++ b/opencti-platform/opencti-graphql/src/config/conf.js
@@ -405,22 +405,23 @@ export const getPlatformHttpProxyAgent = (uri) => {
   return undefined;
 };
 
+// General
+export const ENABLED_API = booleanConf('app:enabled', true);
+export const ENABLED_UI = booleanConf('app:enabled_ui', true);
+
 // Playground
 export const ENABLED_DEMO_MODE = booleanConf('demo_mode', false);
-export const PLAYGROUND_INTROSPECTION_DISABLED = DEV_MODE ? false : booleanConf('app:graphql:playground:force_disabled_introspection', true);
-export const PLAYGROUND_ENABLED = booleanConf('app:graphql:playground:enabled', true);
+export const PLAYGROUND_INTROSPECTION_DISABLED = DEV_MODE ? false : (!ENABLED_UI || booleanConf('app:graphql:playground:force_disabled_introspection', true));
+export const PLAYGROUND_ENABLED = ENABLED_UI && booleanConf('app:graphql:playground:enabled', true);
 export const GRAPHQL_ARMOR_ENABLED = booleanConf('app:graphql:armor_enabled', true);
 
 // Default activated managers
-export const ENABLED_API = booleanConf('app:enabled', true);
 export const ENABLED_TRACING = booleanConf('app:telemetry:tracing:enabled', false);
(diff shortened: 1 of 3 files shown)


opencti-platform/opencti-graphql/src/config/conf.js
```diff
@@ -405,22 +405,23 @@ export const getPlatformHttpProxyAgent = (uri) => {
...
-export const PLAYGROUND_ENABLED = booleanConf('app:graphql:playground:enabled', true);
+export const PLAYGROUND_INTROSPECTION_DISABLED = DEV_MODE ? false : (!ENABLED_UI || booleanConf('app:graphql:playground:force_disabled_introspection', true));
+export const PLAYGROUND_ENABLED = ENABLED_UI && booleanConf('app:graphql:playground:enabled', true);
 export const GRAPHQL_ARMOR_ENABLED = booleanConf('app:graphql:armor_enabled', true);
 
 // Default activated managers
-export const ENABLED_API = booleanConf('app:enabled', true);
 export const ENABLED_TRACING = booleanConf('app:telemetry:tracing:enabled', false);
 export const ENABLED_METRICS = booleanConf('app:telemetry:metrics:enabled', false);
-export const ENABLED_TELEMETRY = booleanConf('app:telemetry:filigran:enabled', false);
```
labo-flg, line 418: These variables were unused ? (`ENABLED_TELEMETRY` and `ENABLED_TELEMETRY_MANAGER`)

richard-julien: Looks like. As telemetry cannot be disable, should have been forgotten

labo-flg:
I've tested and I'm not sure to understand what it's supposed to do.
The static resources are not served (no access to playground or flags images for instance), but the app itself is still accessible.

opencti-platform/opencti-graphql/src/http/httpPlatform.js
```diff
@@ -134,14 +134,16 @@ const createApp = async (app) => {
...
-  // -- Serv flags resources
-  app.use(`${basePath}/static/flags`, express.static('static/flags'));
+    // -- Serv flags resources
+    app.use(`${basePath}/static/flags`, express.static('static/flags'));
 
-  // -- Serv frontend static resources
-  app.use(`${basePath}/static`, express.static(path.join(__dirname, '../public/static')));
+    // -- Serv frontend static resources
+    app.use(`${basePath}/static`, express.static(path.join(__dirname, '../public/static')));
+  }
```
labo-flg, line 146:
```suggestion
  } else {
    logApp.info('[HTTP] UI not enabled, not serving static resources');
  }
```

labo-flg (approved):

Issue #7525, opened 2024-06-29
Incorrect placeholder in Security / Organizations list


SamuelHassine:
## Description

![image](https://github.com/OpenCTI-Platform/opencti/assets/1334279/6b131d4b-51b6-4701-a418-68f5571175e5)

Should be:

![image](https://github.com/OpenCTI-Platform/opencti/assets/1334279/133338b9-11cf-450c-bbf6-e3c590ebf74c)

nino-filigran:
Similar to: https://github.com/OpenCTI-Platform/opencti/issues/7523, but for the current bug, not in knowledge.

PR #7601, opened 2024-07-04
[frontend] Incorrect placeholder in Security / Organizations list (#7525)

SarahBocognano:
### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/7525

To be tested in local because even in slow 2G my connection is too fast to really see the placeholders haha


marieflorescontact (approved):
locally tested ok👌

PR #7636, opened 2024-07-08
[backend] Add support for AWS Signature Version 4 for OpenSearch

Markus98:
### Proposed changes

* Add support for AWS Signature Version 4 for the OpenSearch client

### Related issues

None

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/database/engine.js
+++ b/opencti-platform/opencti-graphql/src/database/engine.js
@@ -1,6 +1,9 @@
 /* eslint-disable no-underscore-dangle */
+import { defaultProvider } from '@aws-sdk/credential-provider-node';
 import { Client as ElkClient } from '@elastic/elasticsearch';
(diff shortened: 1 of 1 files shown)


Markus98:
Hi!
We have had this code working in production for a while and our setup requires it.

I did not add tests for this change. I considered adding some in [elasticSearch-test.js](https://github.com/OpenCTI-Platform/opencti/blob/master/opencti-platform/opencti-graphql/tests/02-integration/01-database/elasticSearch-test.js), but I would have to modify the configuration to hit the code I added. Modifying the configuration doesn't seem to be done in this file and I'm not sure what is your preferred way of doing it. If you have some advice on how I could add a test for this change, I can implement it.

If there are other considerations that need to be looked at before this change can be merged, please let me know. It would be important for us to have this change in the open-source version so that we do not need to fork and rebase our platform every time we need to upgrade the platform version.

Issue #4352, opened 2023-09-18
Bulk entity creation of SDO/SCO with Copy/Pasting


Jipegien:
## Use case

Possibility to manually create multiple entities (and their attributes) from a copy/pasted list.

labo-flg:
Still under FF, reopening until FF removed.

PR #7655, opened 2024-07-09
[frontend] Bulk creation for Channel entity (#4352)

lndrtrbn:
### Proposed changes

* UI components to manage bulk creation
* Possibility to add Channels in bulk

### Related issues

* #4352 

### Checklist

### Further comments

Chunk 1 of the feature https://github.com/OpenCTI-Platform/opencti/issues/4352


opencti-platform/opencti-front/src/components/fields/BulkTextField/BulkTextField.tsx
```diff
@@ -0,0 +1,40 @@
+import React from 'react';
+import { FieldProps } from 'formik';
+import { Alert, Typography } from '@mui/material';
+import TextField from '../../TextField';
+import { useFormatter } from '../../i18n';
+
+// TODO make TextField component as typescript and reuse its props
```
lndrtrbn, line 8: FYI this todo was not expected to be resolved in this PR

opencti-platform/opencti-front/src/private/components/arsenal/channels/ChannelCreation.tsx
```diff
@@ -156,90 +158,114 @@ export const ChannelCreationForm: FunctionComponent<ChannelFormProps> = ({
...
+              formValue={values.name}
+            />
+          )}
+          <Form style={{ margin: '20px 0 20px 0' }}>
+            <Field
+              component={isFeatureEnable('BULK_ENTITIES') ? BulkTextField : TextField}
+              name="name"
+              label={t_i18n('Name')}
+              fullWidth={true}
+              detectDuplicate={['Channel', 'Malware']}
```
aHenryJard, line 181: why do we check duplicates on malware for channel ?

aHenryJard (approved):
Tested locally it's all fine with and without feature flag.

opencti-platform/opencti-front/src/private/components/arsenal/channels/ChannelCreation.tsx
lndrtrbn: I don't know, it was here before my work, and it is also the case for other threats entities

Issue #7312, opened 2024-06-09
In the connector overview, display the user associated to the connector and additional infos


SamuelHassine:
## Use case

In the connector overview, display the user associated to the connector and additional infos:

* Max confidence level of the user
* Role(s) of the user
* Group(s) of the user
* Organization(s) of the user

SamuelHassine:
Remove connector queues names.

PR #7658, opened 2024-07-10
[frontend/backend] Connector queuing auto backpressure (#6325)

### LINKED ISSUE #6325: Some connectors will pause based on queues threshold
## Use case

As a platform admin, I sometimes need to stop connectors to relieve the worker queue. Currently, I am only able to do it by shutting down the entire docker of the connector, that is very cumbersome and cause multiple occasions to make mistakes.

Megafredo:
### Proposed changes

Implementation of feature
**Connector overview**
* Adding show the date of the last run and next run
* Added a specific use case, when the connector is in run_and_terminate the next_run will be in `External schedule`
* Added a specific use case: when the connector has a state with "last_run", and it doesn't have a scheduler defined, then the last_run will take over and convert the last_run of the state.
* Added server capacity, and when this exceeds the capacity indicated by the config, the connector goes into buffer memory, adding an alert with an explanation.
* Remove `listen_queue` and `push_queue`
* Adding more informations in basic information (`Associated user`, `Max confidence level`, `User's roles`, `User's groups`, `User's organization`) 

**Connectors overview**
* Addition of a new "status" column indicating whether the connector is Active or Inactive
* Possibility of sorting connectors by status
* Removal of the colour from the icons in the list of connectors, as it was not accessible.

Refactor and improvements:
* migration of domain/connector from js to typescript, a small part was harder to migrate and is now in connector-utils.js

### Related issues

* closes #6325
* #7312

### Checklist


opencti-platform/opencti-graphql/src/connector/connector.d.ts
```diff
@@ -21,6 +21,7 @@ export interface Connector {
   name: string;
   built_in: boolean;
   connector_schema_runtime_fn: <T extends BasicStoreEntityCsvMapper> (context: AuthContext, user: AuthUser) => Promise<T[]>;
+  connector_state_reset: boolean;
```
aHenryJard: is this really needed here ?

opencti-platform/opencti-graphql/src/connector/importCsv/importCsv.ts
```diff
@@ -16,5 +16,6 @@ export const IMPORT_CSV_CONNECTOR: Connector = {
   connector_type: CONNECTOR_INTERNAL_IMPORT_FILE,
   name: 'ImportCsv',
   built_in: true,
+  connector_state_reset: true, // FIXME what value here ?
```
aHenryJard: I think it's not needed

opencti-platform/opencti-front/src/private/components/data/connectors/Connector.jsx
```diff
@@ -241,6 +251,30 @@ const ConnectorComponent = ({ connector, relay }) => {
       filterGroups: [],
     },
   };
+  const { me } = useAuth();
+  const theme = useTheme();
+  const userCapabilities = (me.capabilities ?? []).map((c) => c.name);
+  const userHasSettingsCapability = userCapabilities.includes(SETTINGS_SETACCESSES) || userCapabilities.includes(BYPASS);
```
SouadHadjiat:
```suggestion
  const userHasSettingsCapability = useGranted([SETTINGS_SETACCESSES]);
```

opencti-platform/opencti-graphql/src/domain/connector.ts
```diff
@@ -316,6 +308,13 @@ export const deleteQueues = async (context, user) => {
...
 // endregion
 
-export const queueDetails = async (connectorId) => {
-  return await getConnectorQueueDetails(connectorId);
+export const queueDetails = async (connectorId: string) => {
+  return getConnectorQueueDetails(connectorId);
+};
+
+export const connectorUser = async (context: AuthContext, user: AuthUser, userId: string) => {
+  if (isUserHasCapability(user, SETTINGS_SET_ACCESSES)) {
```
SouadHadjiat, line 315: this should be protected in the API side directly. See comment on opencti.graphql

opencti-platform/opencti-graphql/config/schema/opencti.graphql
```diff
@@ -1735,7 +1754,9 @@ type Connector implements BasicObject & InternalObject {
   connector_schema_ui: String
   connector_state_reset: Boolean
   connector_user_id: ID
+  connector_user: User
```
SouadHadjiat, line 1767: This should be protected directly here (instead of in domain) : 
`@auth(for: [SETTINGS_SETACCESSES])`

opencti-platform/opencti-front/src/private/components/data/connectors/Connector.jsx
```diff
@@ -382,7 +414,167 @@ const ConnectorComponent = ({ connector, relay }) => {
...
               )}
+              <Security needs={[SETTINGS_SETACCESSES]}>
+                <Grid item xs={6}>
+                  <Typography variant="h3" gutterBottom={true}>
+                    {t_i18n('Associated user')}
+                  </Typography>
+                  {connector.connector_user ? (
+                    <FieldOrEmpty source={connector.connector_user}>
+                      <List>
+                        {[connector.connector_user].map((user) => (userHasSettingsCapability ? (
```
SouadHadjiat: since `connector_user` is protected anyway with this capability on the API side, why do we add this test here in the frontend ? connector.connector_user would be null if the capability is not present, right ? 

Also, why put connector.connector_user in a list ?

opencti-platform/opencti-front/src/private/components/data/connectors/Connector.jsx
```diff
@@ -382,7 +414,167 @@ const ConnectorComponent = ({ connector, relay }) => {
...
+                  )}
+                </Grid>
+                <Grid item xs={6}>
+                  <Typography variant="h3" gutterBottom={true}>
+                    {t_i18n("User's roles")}
+                  </Typography>
+                  {connector.connector_user ? (
+                    <FieldOrEmpty source={connector.connector_user.roles ?? []}>
+                      <List>
+                        {(connector.connector_user.roles ?? []).map((role) => (userHasSettingsCapability ? (
```
SouadHadjiat, line 445: product question : why do we want to display the roles and groups in the connector screen, since the user could directly go into the user details view to see all these information ?

opencti-platform/opencti-front/src/private/components/data/connectors/Connector.jsx
```diff
@@ -567,6 +859,42 @@ const Connector = createRefetchContainer(
         connector_type
         connector_scope
         connector_state
+        connector_user_id
+        connector_user {
```
SouadHadjiat, line 832: I'm a little worried that we might query connector_user with roles & groups & organization too many times since this component refetches the connector every 10 seconds

opencti-platform/opencti-front/src/private/components/data/connectors/Connector.jsx
```diff
@@ -404,17 +604,109 @@ const ConnectorComponent = ({ connector, relay }) => {
...
+                    <FieldOrEmpty source={connector.connector_info?.queue_messages_size}>
+                      <span style={isBuffering() ? { color: theme.palette.warning.main } : {}}>{connector.connector_info?.queue_messages_size.toFixed(2)}</span>
+                      <span> / {connector.connector_info?.queue_threshold} Mo</span>
+                    </FieldOrEmpty>
+                  ) : (
+                    connector.connector_info && connector.connector_info.queue_messages_size === 0 && connector.connector_info.last_run_datetime
+                      ? (<FieldOrEmpty source={connector.connector_info?.queue_messages_size}>
+                        <span style={isBuffering() ? { color: theme.palette.warning.main } : {}}>{connector.connector_info?.queue_messages_size.toFixed(2)}</span>
+                        <span> / {connector.connector_info?.queue_threshold} Mo</span>
+                      </FieldOrEmpty>)
```
SouadHadjiat: I don't understand this whole condition : we display exactly the same thing if connector.connector_info.queue_messages_size === 0 or not, so why the condition ?

opencti-platform/opencti-front/src/private/components/data/connectors/Connector.jsx
```diff
@@ -241,6 +251,30 @@ const ConnectorComponent = ({ connector, relay }) => {
...
+    let lastRun = false;
+    const connectorStateConverted = JSON.parse(connector.connector_state);
+    if (Object.hasOwn(connectorStateConverted, 'last_run')) {
+      lastRun = true;
+      return lastRun;
+    }
+    return lastRun;
+  };
+  const getLastRunConverted = () => {
+    const connectorStateConverted = JSON.parse(connector.connector_state);
```
SouadHadjiat: I don't think it's a good idea to JSON.parse `connector.connector_state` in methods, it should be done once only, not at every call of these methods.

opencti-platform/opencti-front/src/private/components/data/connectors/Connector.jsx
Megafredo: Decision taken in this feature : https://github.com/OpenCTI-Platform/opencti/issues/7312

opencti-platform/opencti-front/src/private/components/data/connectors/Connector.jsx
Megafredo: This raises a number of questions !

opencti-platform/opencti-front/src/private/components/data/connectors/Connector.jsx
Megafredo: yes, you're right, it's a mistake

aHenryJard (approved):
All good for me, let's go !

PR #6972, opened 2024-05-14
Initial Changes to support dynamic configurable required fields via settings --> customization

ParamConstructor:
### Proposed changes
Modify system to support dynamic required fields using the existing capability per object type off the Admin console (setting--> customization). These fields that are required will be marked with a "*" star shown next to labels and result in a read error call out when not filled in/provided.

This change is implemented off the Analyses objects in this intial PR. The Related Issues section will cover where the change was not implemented and why on the Analyses --> External References section. On that area, the platform behavior for required field remains unchanged.

### Related issues

* External-Reference is not directly supported for this required fields change. Based on communication with the Filigran team - this field will require many complex relationships between mandatory attributes to fully implement

* There are two translations unrelated to this PR but were corrected as part of the PR:
   The word "Let" was replaced with "Leave" in these two translations:
        - "Leave the field empty to grant all authenticated users"
        - "Leave the field empty to grant all users"

### Checklist

- [X] I consider the submitted work as finished
- Note - there will be a second PR to apply these changes to the remainder of the platform forms/objects. Filigran requested that PR (https://github.com/OpenCTI-Platform/opencti/pull/5877) be decomposed into 2 PRs for ease of merging/review.
- [X] I tested the code for its functionality
- [X]  (N/A - Existing test cases should work properly) I wrote test cases for the relevant uses case (coverage and e2e)
- [ ] I added/update the relevant documentation (either on github or on notion)
- [X] Where necessary I refactored code to improve the overall quality

### Further comments

There will be an additional PR to adopt this change to the reset of the platform, after the PR is accepted. Some general notes on required fields.

* Files requiring change have been determined by a search of the codebase for 'This field is required' and by limiting found files to ./*/.tsx. The exceptions are listed in these comments.

* There are roughly 46 *.jsx files that need to be rewritten as *.tsx. That is considered outside tech debt and can be performed, as need, adopting this required fields implementation at that time.

* ScaleConfiguration, StixCoreObjectFileExport - Do not currently store attribute configuration in GraphQL, so unable to manage required fields via useSchemaAttributes or useEntitySettings for these fields. [Tech Debt]

* NotifierCreation and NotifierEdition - FORMIK errors were not reported back to user on these objects, unclear how to fix this issue, so not implemented on these objects.

* CsvMapperForm appears to be a work in progress - this change has skipped that form for now.

[...]

PR #7665, opened 2024-07-10
[backend] add new capability 'Bypass mandatory fields' (#3304)

### LINKED ISSUE #3304: Create a new capa to byPass mandatory fields
## Problem to Solve

So I found something interesting. If I try to enforce mandatory attributes for the Report entity, it breaks the MITRE Datasets connector. If it breaks MITRE, it likely breaks other connectors handling it.
image

The reason a user may want to enforce mandatory attributes on that, is to prevent users from improperly filling out the details for the Report entity.

The error I started getting from the worker logs was:
`2023-05-09T22:04:58.235430037Z ERROR:pycti.api:Invalid field createdBy`
`2023-05-09T22:04:58.235457637Z ERROR:root:{'name': 'ValidationError', 'message': 'Invalid field createdBy'}`
`2023-05-09T22:04:58.263740311Z ERROR:pycti.api:[SEARCH] Error updating elastic (update)`
`2023-05-09T22:04:58.263763511Z ERROR:pycti.api:Cannot report expectation`

The error pointed to something going wrong with the Author entity or createdBy field.
image

The end result is that reports might still get output but certain entities or relationships getting excluded from the Report container.

I believe this error might be caused when certain fields in the STIX bundles have a null createdBy field, leading to that data getting excluded upon matching the mandatory attributes rules. 

## Current Workaround

To resolve this issue, undoing the enforcement of mandatory attributes on the Report entity causes the problem to go away.

But this means the user becomes unable to make use of the enforce mandatory attributes feature, because it breaks the connectors that handle the Report entity.

## Proposed Solution

I believe a sound, interim solution would be to add a new permission that can be added to the Connector Role, that would allow the connector to bypass mandatory field attributes.
image

A longer term solution I think should be considered is adopting the concept of internal service users (these handle all backend OpenCTI operations) and regular users. This is because often times, what we want to happen for a front end user is different from the service user (connector).

## Final approach
Create a new capa called "Bypass mandatory fields" that would allow admins to enable some custom mandatory attributes so that their connectors for instance will be able to byPass any fields that are marked as required, while analyst will be have to comply with the filling of the needed fields when creating or updating an entity.

frapuks:
### Proposed changes

- add new capa (initialisation + migration)
- bypass schema validation in create and update entity
- migration of schema-validation-test to typescript + add test coverage on bypass.
- Doc : https://github.com/OpenCTI-Platform/docs/pull/189

### Related issues

- close #3304 

### Checklist


opencti-platform/opencti-graphql/src/modules/entitySetting/entitySetting-attributeUtils.ts
```diff
@@ -123,5 +124,8 @@ export const getMandatoryAttributesForSetting = async (
   entitySetting: BasicStoreEntityEntitySetting
 ) => {
   const attributes = await getEntitySettingSchemaAttributes(context, user, entitySetting);
+  if (isUserHasCapability(user, KNOWLEDGE_KNUPDATE_KNBYPASSFIELDS)) {
```
Kedae, line 127: Might be missing something but I don't think this is a good way of handling this as you will "lose" the fact that the attribute is mandatory.

I mean, having the capability the bypass them is interesting, having no clue they indeed should be mandatory is another.

frapuks: Currently, we simply return a list of attribute names: `[‘name’, ‘description’ ]`. And in the Front end, we try to match these names with the fields in the form.
You can return all the mandatory attributes and filter in the front-end. We didn't opt for this solution when we were brainstorming, because in front-end, you can always do what you want and work around it manually. And if you don't use the front end to create the entity, the validation has to be in the back end.

lndrtrbn: I have same point of view as @Kedae here. 

We have an API `EntitySetting.mandatoryAttributes` in graphql that is returning which are the required attributes for an entity type. No matter you have the bypass capability or not, those fields stay required. This solution makes the result of this API inconsistent depending of your capabilities and could lead to misunderstanding or issues depending on how the community is using this API.

Moreover, this `getMandatoryAttributesForSetting` function is also used in the middleware to prevent deleting a relationship attached to required stuff. I have the feeling that could lead to issues also here?

---

> in front-end, you can always do what you want and work around it manually

Even with this solution you can still do what you want and inspect code to remove front validation.

> And if you don't use the front end to create the entity, the validation has to be in the back end.

Don't you already done it in `schema-validator.ts`? Because this `getMandatoryAttributesForSetting` function is not about validating inputs so I don't get it. For what I understand you have already done the check on configured required attributes in `schema-validator.ts` so you cannot fake the API no matter what, we good on this point. 

It's just the way to handle those required attributes in UI that, in my opinion, need to be done in frontend based on capabilities and not on the result of this `EntitySetting.mandatoryAttributes` API.

ParamConstructor:
May be some value in looking at the new opencti-platform/opencti-front/src/utils/hooks/useSchemaAttributes.tsx for compatability in https://github.com/OpenCTI-Platform/opencti/pull/6972 that we are proposing for validation of required fields handling and "*"'ing the fields on the forms UI.

opencti-platform/opencti-graphql/src/modules/entitySetting/entitySetting-attributeUtils.ts
frapuks: I can't find any cases where the ‘deleteRelationsByFromAndTo’ method is triggered with customisable mandatory attributes.

lndrtrbn: Acceptable solution. Will be changed later if necessary

lndrtrbn (approved):

Issue #7311, opened 2024-06-09
Domain name values should be case insensitive


SamuelHassine:
## Description

Create domain 1: Google.com
Create domain 2: google.com

=> Duplicate

SarahBocognano:
What about url, email adress, hostname ? And is there others that are case sensitive ?

richard-julien:
In my memories i have implemented the STIX specification.
They explicitly describe that the data must be canonicalize but nothing about removing the case.

```
The value of the name portion SHOULD be the list of "ID Contributing Properties" (property-name and property value pairs) as defined on each SCO object and SHOULD be represented as a JSON object that is then serialized / stringified according to [
[RFC8785](https://docs.oasis-open.org/cti/stix/v2.1/os/stix-v2.1-os.html#kvy09cwtcw71)
] to ensure a canonical representation of the JSON data.
```

@SamuelHassine, I think the solution will be to lowercase the value before creating the standard_id instead of changing the way the standard_id is generated.

buddycat:
Not sure if this is the place to comment, but urls can contain case sensitive data (hashes, etc.) that forcing lower casing would break.

PR #7677, opened 2024-07-11
[backend] Domain name values should be case insensitive (#7311)

SarahBocognano:
### Proposed changes

* Force lowercase in creation for domain name, url and email adress.

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/7311

### DIFF
--- a/opencti-platform/opencti-graphql/src/domain/stixCyberObservable.js
+++ b/opencti-platform/opencti-graphql/src/domain/stixCyberObservable.js
@@ -194,7 +194,7 @@ export const addStixCyberObservable = async (context, user, input) => {
   if (!input[graphQLType]) {
     throw FunctionalError(`Expecting variable ${graphQLType} in the input, got nothing.`);
   }
-  const lowerCaseTypes = ['Domain-Name', 'Url', 'Email-Addr'];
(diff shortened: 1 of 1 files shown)


opencti-platform/opencti-graphql/src/domain/stixCyberObservable.js
```diff
@@ -194,6 +194,11 @@ export const addStixCyberObservable = async (context, user, input) => {
   if (!input[graphQLType]) {
     throw FunctionalError(`Expecting variable ${graphQLType} in the input, got nothing.`);
   }
+  const lowerCaseTypes = ['Domain-Name', 'Url', 'Email-Addr'];
```
lndrtrbn: Do we really want URL here ? See https://github.com/OpenCTI-Platform/opencti/issues/7311#issuecomment-2223003003

SarahBocognano: Kinda validated it as "good candidat" with Sam yesterday, but I will remove it for now until we're sure. Especially after this comment. thanks for sharing it

lndrtrbn (approved):

Issue #7479, opened 2024-06-25
Malware Infrastructure Display Issue


explorecti:
## Description

When a relationship of Malware USES Infrastructure exists, Infrastructure is included under Malware/Knowledge Infrastructure, but Malware is not shown under Infrastructure/Knowledge/Malware. In addition Infrastructure/Knowledge/Malware relationships is displayed for relationships Malware CONTROLS Infrastructure.

## Environment

1. OS: RHEL8
2. OpenCTI version: 6.1.11
3. OpenCTI client: Frontend

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Navigate to http://opencti:8080/dashboard/arsenal/malware/UUID_from_Malware/knowledge/relations/UUID_from_Infrastructure
2. Now navigate in reverse http://opencti:8080/dashboard/arsenal/malware/UUID_from_Infrastructure/knowledge/relations/UUID_from_Malware

## Expected Output

The UI should show Infrastructure USES Malware

## Actual Output

The Malware is not shown

nino-filigran:
On the top of this, because uses is not displayed, it also means we cannot create create infra USES malware from this same view.

lndrtrbn:
Seems to be more related to the relation `uses` itself as I can also reproduce with "Malware uses Tool"

PR #7681, opened 2024-07-12
Fix relation uses in knowledge (#7479)

lndrtrbn:
### Proposed changes

* Add uses relation where it was missing (based on STIX documentation)

### Related issues

* #7479 

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/observations/infrastructures/InfrastructureKnowledge.tsx
+++ b/opencti-platform/opencti-front/src/private/components/observations/infrastructures/InfrastructureKnowledge.tsx
@@ -240,6 +240,7 @@ const InfrastructureKnowledge = ({ infrastructure }: { infrastructure: Infrastru
               entityLink={link}
               defaultStartTime={infrastructureData.first_seen}
               defaultStopTime={infrastructureData.last_seen}
+              allDirections={true}
             />
         }
         />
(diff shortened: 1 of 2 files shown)


opencti-platform/opencti-front/src/private/components/observations/infrastructures/InfrastructureKnowledge.tsx
```diff
@@ -243,7 +248,7 @@ const InfrastructureKnowledge = ({ infrastructure }: { infrastructure: Infrastru
           element={
             <EntityStixCoreRelationships
               entityId={infrastructureData.id}
-              stixCoreObjectTypes={['Tool']}
+              stixCoreObjectTypes={['Tool', 'uses']}
               relationshipTypes={['hosts']}
```
JeremyCloarec: uses should be put in the relationshipTypes

lndrtrbn:
```suggestion
              stixCoreObjectTypes={['Tool']}
              relationshipTypes={['hosts', 'uses']}
```

JeremyCloarec:
Switching to allDirections for malwares in infrastructure fixes the display issue. But it creates an inconsistency with the relation creation behavior: when creating relations, we can only create relation from the infrastructure to the malware (so the controls, delivers and hosts relations).
What's the expected functional behavior on theses views?

SamuelHassine:
> Switching to allDirections for malwares in infrastructure fixes the display issue. But it creates an inconsistency with the relation creation behavior: when creating relations, we can only create relation from the infrastructure to the malware (so the controls, delivers and hosts relations). What's the expected functional behavior on theses views?

Yes it is, it will be reworked later.

Issue #7540, opened 2024-06-29
When using a list in dashboard widgets that is not clickable, pointer / behavior not correct


SamuelHassine:
## Description

https://filigran.octi.filigran.io/dashboard/workspaces/dashboards/905236b3-33c3-4b6c-8a61-0eeab425d692

Indicator types: pointer + button but not clickable:

![image](https://github.com/OpenCTI-Platform/opencti/assets/1334279/f373eff8-6ba3-4018-a5ed-37513eab3a81)

PR #7686, opened 2024-07-12
[frontend] When using a list in dashboard widgets that is not clickable, pointer / behavior not correct (#7540)

SarahBocognano:
### Proposed changes

* cursor set to default in sx props

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/7540


marieflorescontact (commented):
cursor fixed.
don't you think we should remove hoover effect on the line if not cliquable? (hoover effect makes me want to click 🤣)

lndrtrbn:
> cursor fixed. don't you think we should remove hoover effect on the line if not cliquable? (hoover effect makes me want to click 🤣)

Maybe the component `ListItemButton` is not relevant here and we should use a classic list

marieflorescontact (approved):

Issue #7597, opened 2024-07-04
Blank space on the right for some relationships overviews


Archidoit:
## Description

Blank space on the right for some relationships overviews

## Reproducible Steps
- Create a relationship between a relationship and a Data Source
- Go to this relationship overview
- There is a blank space on the right

![image](https://github.com/OpenCTI-Platform/opencti/assets/75783086/0d3e11ce-4a0a-4912-b0cf-ebd4d5540857)

## Comment
Same for Data Components and Courses of Action

SamuelHassine:
Can you please use `<>` instead of `<div>`?

PR #7690, opened 2024-07-12
[frontend] Blank space on the right for some relationships overviews (#7597)

SarahBocognano:
### Proposed changes

* Remove the method getPaddingRight in entities without knowledge

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/7597


frapuks:
In the header of the tab content, we need a paddingRight because we have a right pannel for files.
I think we can keeping the method `getPaddingRight` and using the third argument `applyKnowledgePadding = false`

frapuks (approved):

Issue #4353, opened 2023-09-18
Unifying creation and edition form


Jipegien:
## Use case

- A single creation and edition form, more complex but common to every SDO and SRO to facilitate evolution.
- This form allows users to choose which type of entity they will create. It contains at least 2 panels (Mandatory and Details) but can have other optional panels (biodata for example). Panels (which ones and their content) depends of the entity’s type chosen.

PR #7711, opened 2024-07-16
[frontend] Apply bulk creation on other entities than Channels (#4352)

lndrtrbn:
### Proposed changes

* Apply same code used in Channels for bulk creation to other entities

### Related issues

* #4352 

### Checklist

### Further comments

Chunk 3 of #4353


Archidoit:
Tested for all the entity types: bulk creation is working

Archidoit:
But some inconsistances in the error message if we write a name that has only one character.
Exemple

- for threat actor individual
image


- for tools
image

Archidoit:
The popup of successful creation only closes if we click on the 'Close' button.
For consistancy within the platform (like for entity deletion warning popup for instance), maybe also close the popup if we click elsewhere outside the popup

image

Archidoit:
It would be good to have no error display after indicating several names (if possible :) )
image

Archidoit (approved):

Issue #7712, opened 2024-07-16
Inferred relationships are no longer considered “inferred” after the restoration of a deleted entity


romain-filigran:
## Description

When deleting an entity that has "inferred relations", previously inferred relationships are no longer considered inferred after restoration but are considered as created by user behind the restoration.

## Environment

OpenCTI version: 6.2.5

## Reproducible Steps

1. Delete an entity with "inferred relations" like this one (contains Inferred relations created by Rule Manager in history)
https://testing.octi.staging.filigran.io/dashboard/locations/countries/1c99d8f8-bd1f-48c5-ba94-dc3116f6ba37

2. Navigate to "Trash" panel and restore the entity

3. Check the "author" of the previous "inferred relations" 

Example of history of my tests

Before delete: 

![before-delete-inferred-4](https://github.com/user-attachments/assets/e264b8ab-18d2-45cc-9cc4-25af2589f455)

After restoration 

![after-restore-inferred-2](https://github.com/user-attachments/assets/f4c8af50-26d0-4008-9b5c-c898658be284)

PR #7733, opened 2024-07-17
[backend] fixed inferred relations restoration when an entity is restored from trash (#7712)

JeremyCloarec:
### Proposed changes

* inferred relations are now restored using createInferredRelation instead of createRelation
* we now check for upserted entities/relations during restore to be able to link remaining relations to restore to the correct id

### Related issues

* #7712 

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/modules/deleteOperation/deleteOperation-domain.ts
+++ b/opencti-platform/opencti-graphql/src/modules/deleteOperation/deleteOperation-domain.ts
@@ -325,7 +325,7 @@ export const restoreDelete = async (context: AuthContext, user: AuthUser, id: st
       else {
         const rule = Object.keys(relationToRestore).find((k) => k.startsWith(RULE_PREFIX));
         if (rule === undefined) {
-          logApp.warn('Inferred rule could not be found');
+          logApp.warn('Inferred rule could not be found', { relation: relationToRestore });
         } else {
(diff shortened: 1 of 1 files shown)


opencti-platform/opencti-graphql/src/modules/deleteOperation/deleteOperation-domain.ts
```diff
@@ -300,16 +309,36 @@ export const restoreDelete = async (context: AuthContext, user: AuthUser, id: st
...
+        } else {
+          const ruleID = rule.substring(RULE_PREFIX.length);
+          const ruleValues = (relationToRestore as Record<string, any>)[rule][0];
+          const ruleContent = createRuleContent(ruleID, ruleValues.dependencies, ruleValues.explanation, ruleValues.data);
+          result = await createInferredRelation(context, relationshipInput, ruleContent);
+        }
+      }
+      if (result.id && result.id !== relationshipInput.id) {
+        upsertedElements[relationshipInput.id] = result.id;
+        logApp.warn('Relationship has been restored with with different id (upsert)');
```
aHenryJard:
```suggestion
        logApp.warn('Relationship has been restored with different id (upsert)');
```

aHenryJard: could you add some ids in logs ?

aHenryJard (approved):
tested locally, all good for me.

Issue #7717, opened 2024-07-17
When displaying an organization or an individual as "author", right padding is not correct in knowledge


SamuelHassine:
## Description

When displaying an organization or an individual as "author", right padding is not correct in knowledge

![image](https://github.com/user-attachments/assets/f8e82852-a84d-46bc-addb-68900fea488a)

nino-filigran:
Repro case for clarity: 
- go to an organization
- switch to author view
- go to knowledge view

If the result is unclear, switch back to the knowledge entity.

The expected result is that the knowledge view should be on full screen and not have the placeholder of the knowledge menu.

lndrtrbn:
We had multiple issues related to this "paddingRight" stuff. Maybe it could be nice to take time to investigate if we can refacto the base layout of the UI to be able to not use this artificial padding depending to the context of the page.

PR #7736, opened 2024-07-17
[frontend] When displaying an organization or an individual as "author", right padding is not correct in knowledge (#7717)

SarahBocognano:
### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/7717

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/entities/organizations/Root.jsx
+++ b/opencti-platform/opencti-front/src/private/components/entities/organizations/Root.jsx
@@ -131,7 +131,7 @@ class RootOrganization extends Component {
             if (props) {
               if (props.organization) {
                 const { organization } = props;
-                const paddingRight = getPaddingRight(location.pathname, organization.id, '/dashboard/entities/organizations');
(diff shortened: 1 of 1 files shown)


opencti-platform/opencti-front/src/private/components/entities/organizations/Root.jsx
```diff
@@ -131,7 +131,7 @@ class RootOrganization extends Component {
             if (props) {
               if (props.organization) {
                 const { organization } = props;
-                const paddingRight = getPaddingRight(location.pathname, organization.id, '/dashboard/entities/organizations');
+                const paddingRight = viewAs === 'knowledge' ? getPaddingRight(location.pathname, organization.id, '/dashboard/entities/organizations') : '';
```
JeremyCloarec: We still need the padding right when in content tab, both in knowledge and author view. It needs to be changed to something like :
`const paddingRight = getPaddingRight(location.pathname, organization.id, '/dashboard/entities/organizations', viewAs === 'knowledge');`

JeremyCloarec (changes requested):

JeremyCloarec (approved):

PR #7032, opened 2024-05-20
Replaced Edit Floating Action Buttons for Analyses

Bonsai8863:
### Proposed changes

* Replace the floating action button for editing analyses with a dedicated button in the top right the page.

### Related issues

* #6738
* #6756

### Checklist

- [x] I consider the submitted work as finished
- [x] I tested the code for its functionality
- [x] I wrote test cases for the relevant uses case (coverage and e2e)
- [ ] I added/update the relevant documentation (either on github or on notion)
- [x] Where necessary I refactored code to improve the overall quality

PR #6756, opened 2024-04-18
Replaced Creation Floating Action Buttons for Analyses

Bonsai8863:
### Proposed changes

* Replaced FAB for creation of Reports, Groupings, Malware Analyses, Notes, and External References.
* Replaced contextual FAB for creation of Notes and External References.

### Related issues
* #6738

### Checklist

- [x] I consider the submitted work as finished
- [x] I tested the code for its functionality
- [x] I wrote test cases for the relevant uses case (coverage and e2e)
- [ ] I added/update the relevant documentation (either on github or on notion)
- [x] Where necessary I refactored code to improve the overall quality

PR #6738, opened 2024-04-17
[backend] Added feature flag for Floating Action Button replacement

Bonsai8863:
### Proposed changes

* Added "FAB_REPLACEMENT" to default.json to ignore the ongoing FAB replacement tasking in non-development environments.

### Checklist

- [x] I consider the submitted work as finished
- [x] I tested the code for its functionality
- [ ] I wrote test cases for the relevant uses case (coverage and e2e)
- [ ] I added/update the relevant documentation (either on github or on notion)
- [ ] Where necessary I refactored code to improve the overall quality

### Further comments

I am working to decompose #5269 into smaller PRs. This feature flag enables non-development environments to not be affected by efforts to replace the floating action button throughout the platform in the meantime.

PR #7741, opened 2024-07-17
Replaced Edit Floating Action Buttons for Observations

Bonsai8863:
### Proposed changes

* Replace the floating action button for editing observations with a dedicated button in the top right the page.

### Related issues

* https://github.com/OpenCTI-Platform/opencti/pull/6738
* https://github.com/OpenCTI-Platform/opencti/pull/6756
* https://github.com/OpenCTI-Platform/opencti/pull/7032

### Checklist


opencti-platform/opencti-front/src/private/components/observations/stix_cyber_observables/StixCyberObservableEdition.tsx
```diff
@@ -0,0 +1,159 @@
...
+        anchor="right"
+        sx={{
+          zIndex: 1202,
+          minHeight: '100vh',
+          width: '50%',
+          position: 'fixed',
+          overflow: 'auto',
+          transition,
+          padding: 0,
+        }}
```
labo-flg: the style is not good on the drawer, for instance it does not take 50% of the screen width.

It's because the style should actually be applied to the Paper inside the Drawer (see the original JS code, the class was applied onto `paper` field in `classes`

You can still use `sx`, but applied to the `PaperProps`, like this:


```suggestion
        paperProps={{
          sx: {
            zIndex: 1202,
            minHeight: '100vh',
            width: '50%',
            position: 'fixed',
            overflow: 'auto',
            transition,
            padding: 0,
          }
        }}
```

opencti-platform/opencti-front/src/private/components/observations/stix_cyber_observables/StixCyberObservableEdition.tsx
```diff
@@ -0,0 +1,159 @@
...
+      elevation={1}
+      sx={{
+        zIndex: 1202,
+        minHeight: '100vh',
+        width: '30%',
+        position: 'fixed',
+        overflow: 'auto',
+        transition,
+        padding: 0,
+      }}
```
labo-flg: same here

opencti-platform/opencti-front/src/private/components/observations/stix_cyber_observables/StixCyberObservableEdition.tsx
```diff
@@ -0,0 +1,163 @@
...
+          sx: {
+            zIndex: 1202,
+            minHeight: '100vh',
+            width: '50%',
+            position: 'fixed',
+            overflow: 'auto',
+            transition,
+            padding: 0,
+          },
+        }}
```
labo-flg: I've tested again and it appears that doing so, the drawer gets a wrong z-index and appear below the main header bar.
If we want the drawer to appear on top of the whole screen, we need to apply zIndex on the Drawer and not the paper it seems.

```suggestion
        sx={{ 
          zIndex: 1202,
        }}
        PaperProps={{
          sx: {
            minHeight: '100vh',
            width: '50%',
            position: 'fixed',
            overflow: 'auto',
            transition,
            padding: 0,
          },
        }}
```

opencti-platform/opencti-front/src/private/components/observations/stix_cyber_observables/StixCyberObservableEdition.tsx
```diff
@@ -0,0 +1,163 @@
...
+      open={graphOpen}
+      anchor="right"
+      elevation={1}
+      PaperProps={{
+        sx: {
+          zIndex: 1202,
+          minHeight: '100vh',
+          width: '30%',
+          position: 'fixed',
+          overflow: 'auto',
```
labo-flg: same here

labo-flg (approved):
all tested ok, with and without FF

PR #7752, opened 2024-07-18
Proposed Color Changes to the Left Navigation Menu

dotdev8:
### Proposed changes

* The left navigation top level and submenu items turn blue when user is on one of those submenu pages. 

### Related issues

* No Issues

### Checklist

### Further comments
Tests for this PR were done in browser by observing the behavior when a submenu page is selected. Test cases are not relevant for this PR.

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/nav/LeftBar.jsx
+++ b/opencti-platform/opencti-front/src/private/components/nav/LeftBar.jsx
@@ -83,6 +83,7 @@ import useGranted, {
   KNOWLEDGE_KNUPDATE,
   KNOWLEDGE_KNUPDATE_KNDELETE,
   MODULES,
+  SETTINGS,
   SETTINGS_SETPARAMETERS,
   SETTINGS_SECURITYACTIVITY,
   SETTINGS_SETACCESSES,
@@ -91,6 +92,7 @@ import useGranted, {
   SETTINGS_SETMARKINGS,
   SETTINGS_FILEINDEXING,
   SETTINGS_SUPPORT,
+  TAXIIAPI_SETCOLLECTIONS,
   CSVMAPPERS,
   VIRTUAL_ORGANIZATION_ADMIN,
   INGESTION,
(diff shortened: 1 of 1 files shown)


Kedae:
Hello @dotdev8,

Currently your code is not aligned with what we want the leftbar to look like.
You have lost the "Dashboard" button in top of the left bar + we don't want the whole subMenu to be colored when selected, the "section" part can be usefull, but that's all.

For example there : 
image
I would like Data to be colored, but not Entities, Relationships & co, and I'm missing the Home button

Kedae:
Additionnaly, please do not commit files that have nothing to do with the matter. In your case the yarn upgrade shoud not be in your review

opencti-platform/opencti-front/src/private/components/nav/LeftBar.jsx
```diff
@@ -108,22 +110,21 @@ import logoFiligranTextLight from '../../../static/images/logo_filigran_text_lig
...
 import useDimensions from '../../../utils/hooks/useDimensions';
 
-const SMALL_BAR_WIDTH = 55;
-const OPEN_BAR_WIDTH = 180;
-
 // Deprecated - https://mui.com/system/styles/basics/
 // Do not use it for new code.
 const useStyles = makeStyles((theme) => createStyles({
   drawerPaper: {
-    width: SMALL_BAR_WIDTH,
```
Kedae: I don't think this is a good idea to replace our constant with the exact same values.

opencti-platform/opencti-front/src/private/components/nav/LeftBar.jsx
```diff
@@ -214,16 +215,7 @@ const LeftBar = () => {
...
-    SETTINGS_SETPARAMETERS,
-    SETTINGS_SETACCESSES,
-    SETTINGS_SETMARKINGS,
-    SETTINGS_SETCUSTOMIZATION,
-    SETTINGS_SETLABELS,
-    SETTINGS_SECURITYACTIVITY,
-    SETTINGS_FILEINDEXING,
-    SETTINGS_SUPPORT,
-  ]);
+  const isGrantedToSettings = useGranted([SETTINGS]);
```
Kedae: Nop, SETTINGS is no longer an "available" capa, you can not use useGranted on that

opencti-platform/opencti-front/src/private/components/nav/LeftBar.jsx
```diff
@@ -265,7 +257,6 @@ const LeftBar = () => {
   const handleToggle = () => {
     setSelectedMenu([]);
     localStorage.setItem('navOpen', String(!navOpen));
-    window.dispatchEvent(new StorageEvent('storage', { key: 'navOpen' }));
```
Kedae: Why this removal ?

PR #7760, opened 2024-07-19
[frontend] reset order after search in lists (#7726)

### LINKED ISSUE #7726: In Threats => Intrusion sets, when emptying the search it should come back to default sorting
## Description

1. Go in intrusion sets list
2. Searching
3. Sorting is automatically set to "Score" (expected)
4. Remove your search keyword and press enter

=> Staying on score, which has no sense and the list is never complete, should come back to the previous sorting of the user

Archidoit:
### Proposed changes
Reset order mode and order by after a search in lines and cards lists

### Related issues
https://github.com/OpenCTI-Platform/opencti/issues/7726


SarahBocognano (approved):
Tested ✅

Issue #4538, opened 2023-10-10
Extend authorized_member system to Cases, Groupings and Reports


Jipegien:
## Use case

Same as #3181 but for Cases

richard-julien:
@marieflorescontact , @SouadHadjiat , think also about 

The 2 methods userFilterStoreElements and isUserCanAccessStixElement
They are used to control the access after a loading and also need to support the conditions of the authorized members

A possible capability to bypass the authorized members restrictions for the user that represent connectors (to discuss with product)

PR #7773, opened 2024-07-22
[frontend/backend] Add authorized members on caseIncident (#4538)

marieflorescontact:
### Proposed changes

* Chunk 1 : Add authorized members on caseIncident
* Create a backend utils method to `editAuthorizedMembers `
* Create a frontend utils method to `getCurrentUserAccessRight`

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/4538

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/common/form/AuthorizedMembersField.tsx
+++ b/opencti-platform/opencti-front/src/private/components/common/form/AuthorizedMembersField.tsx
@@ -17,6 +17,7 @@ import SelectField from '../../../../components/fields/SelectField';
 import { useFormatter } from '../../../../components/i18n';
 import { AccessRight, ALL_MEMBERS_AUTHORIZED_CONFIG, AuthorizedMemberOption, Creator, CREATOR_AUTHORIZED_CONFIG } from '../../../../utils/authorizedMembers';
 import SwitchField from '../../../../components/fields/SwitchField';
+import useAuth from '../../../../utils/hooks/useAuth';
 
 /**
  * Returns true if the authorized member option is generic.
@@ -40,6 +41,7 @@ interface AuthorizedMembersFieldProps
   owner?: Creator;
   showAllMembersLine?: boolean;
   showCreatorLine?: boolean;
+  showCurrentUserLine?: boolean;
   canDeactivate?: boolean;
 }
 
@@ -50,6 +52,7 @@ interface AuthorizedMembersFieldInternalValue {
   newAccessRight: AccessRight;
   allAccessRight: AccessRight;
   creatorAccessRight: AccessRight;
+  currentUserAccessRight: AccessRight;
 }
 
 // Validation for the internal formik form.
@@ -71,9 +74,11 @@ const AuthorizedMembersField = ({
   owner,
   showAllMembersLine = false,
   showCreatorLine = false,
+  showCurrentUserLine = false,
   canDeactivate = false,
 }: AuthorizedMembersFieldProps) => {
   const { t_i18n } = useFormatter();
+  const { me } = useAuth();
   const { setFieldValue } = form;
   const { name, value } = field;
   // Value in sync with internal Formik field 'applyAccesses'.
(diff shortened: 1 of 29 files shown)


JeremyCloarec:
Activating authorized members on a case where you are not the creator doesn't work properly: the creator is added as an admin to the authorized members automatically, but you are not added automatically to the authorized members.

JeremyCloarec:
When activating authorized members on en entity created by a user that doesn't have the 'Manage authorized members' capability, we can validate the authorized members without any admins
image

JeremyCloarec:
image
I'm getting an error when adding authorized members on a feedback created by an other user: the request doesn't seem to add the user Joe properly to the authorized members

JeremyCloarec:
When activating authorized members on a newly created incident case, my admin user is not added as an authorized member 
image

SouadHadjiat (commented):
Where is the feature flag ?

marieflorescontact:
> Where is the feature flag ?

omg thanks @SouadHadjiat , I totally forgot i was focused on getting the CI green

marieflorescontact:
> Activating authorized members on a case where you are not the creator doesn't work properly: the creator is added as an admin to the authorized members automatically, but you are not added automatically to the authorized members.

✅ fixed 

> When activating authorized members on en entity created by a user that doesn't have the 'Manage authorized members' capability, we can validate the authorized members without any admins

✅ fixed 

> I'm getting an error when adding authorized members on a feedback created by an other user: the request doesn't seem to add the user Joe properly to the authorized members

=>I can't reproduce neither on Feedbacks noe on Cas incident

JeremyCloarec:
Current bugs:
- When saving authorized members on en entity where I'm not the creator, the query only contains the creator and not my current user. In addition, if the creator doesn't have the manage authorized members capability, the mutation fails
image
- When first opening authorized members on an entity created y an other user, the adming access doesn't show any message
image
- When delescting the authorized members dialog and reopening it, the creator is removed from the authorized members list
image

opencti-platform/opencti-front/src/private/components/common/form/AuthorizedMembersField.tsx
```diff
@@ -98,6 +106,11 @@ const AuthorizedMembersField = ({
     type: CREATOR_AUTHORIZED_CONFIG.type,
     value: CREATOR_AUTHORIZED_CONFIG.id,
   };
+  const currentUserOption: Option = {
+    label: t_i18n(me.name),
```
SouadHadjiat: no translation here

opencti-platform/opencti-front/src/private/components/common/form/FormAuthorizedMembers.tsx
```diff
@@ -64,7 +67,9 @@ const FormAuthorizedMembers = ({
                 name="authorizedMembers"
                 component={AuthorizedMembersField}
                 owner={owner}
+                currentUser={me}
```
SouadHadjiat: why do we add currentUser here since it's not used afterwards in AuthorizedMembersField ? we only use 'me'. Maybe we don't need at all this param

opencti-platform/opencti-front/src/private/components/common/form/FormAuthorizedMembers.tsx
```diff
@@ -64,7 +67,9 @@ const FormAuthorizedMembers = ({
...
                 showAllMembersLine
+                showCurrentUserLine={showCurrentUserLine}
```
SouadHadjiat: I don't think we need this param either, since AuthorizedMembersField has all information to choose how to display current user.

opencti-platform/opencti-front/src/private/components/common/form/AuthorizedMembersField.tsx
```diff
@@ -169,6 +183,16 @@ const AuthorizedMembersField = ({
           accessRight: 'admin',
         },
       ]);
+    } else if (showCurrentUserLine) {
```
SouadHadjiat: something is wrong with this condition, because it means we can't display creator & current user at the same time. Not sure about the condition on owner.

lndrtrbn:
@JeremyCloarec most of the issues should be fixed if you wanna try. The only one still present I think is _if the creator doesn't have the manage authorized members capability, the mutation fails_.

Goumies:
> When activating authorized members on en entity created by a user that doesn't have the 'Manage authorized members' capability, we can validate the authorized members without any admins image

@JeremyCloarec, how can the user click the button to open this dialog? As a user without the capability, I could not see the button when I tried to replicate.

Goumies:
> @JeremyCloarec most of the issues should be fixed if you wanna try. The only one still present I think is _if the creator doesn't have the manage authorized members capability, the mutation fails_.

I will write a test to make sure we have the expected behavior.

JeremyCloarec (approved):

PR #7774, opened 2024-07-23
[frontend] Fix the display of name in widgets when entities are named 'Unknown' (#7692)

### LINKED ISSUE #7692: When threat actor is named "unknown", distribution is displaying the standard ID
## Description

image

image

CelineSebe:
### Proposed changes

* stixCoreObjectsTimelineQuery  with name entities
*  entity-representation don't return standard standard_id

### Related issues

* #7692 

### Checklist


opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectsDistributionList.jsx
```diff
@@ -44,6 +44,149 @@ const stixCoreObjectsDistributionListDistributionQuery = graphql`
       label
       value
       entity {
+        ... on StixDomainObject {
+          created
+          modified
+        }
+        ... on AttackPattern {
+          name
```
Kedae: You should use representative to avoid huge list I think

opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectsTimeline.jsx
```diff
@@ -29,13 +29,157 @@ const stixCoreObjectsTimelineQuery = graphql`
           id
           entity_type
           created_at
+          created_at
+          ... on StixDomainObject {
+            created
+            modified
```
Kedae: Same

opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectsTimeline.jsx
```diff
@@ -29,13 +29,157 @@ const stixCoreObjectsTimelineQuery = graphql`
           id
           entity_type
           created_at
+          created_at
```
Kedae: Doublon

opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectsTimeline.jsx
```diff
@@ -67,6 +211,7 @@ const stixCoreObjectsTimelineQuery = graphql`
 `;
 
 const StixCoreObjectsTimeline = ({
+  title,
```
Kedae: This is unused ?

opencti-platform/opencti-front/src/utils/defaultRepresentatives.ts
```diff
@@ -92,6 +92,9 @@ export const defaultKey = (n: any) => {
 // equivalent to querying representative.main
 export const getMainRepresentative = (n: any, fallback = 'Unknown') => {
   if (!n) return '';
+  if (n.name === 'Unknown') {
```
Kedae, line 95: I don't get this change

CelineSebe: In this function, "Unknown" is indirectly handled as a fallback value. It will only be returned: if none of the properties or function calls preceding it in the mainValue assignment are truthy. 
To ensure that the name "Unknown" is returned if it is set as n.name, we could modify the function to handle the "Unknown" case explicitly before.

opencti-platform/opencti-front/src/utils/defaultRepresentatives.ts
```diff
@@ -95,43 +95,49 @@ export const getMainRepresentative = (n: any, fallback = 'Unknown') => {
   if (typeof n.definition === 'object') {
     return defaultValueMarking(n);
   }
+
+  // Check for n.name explicitly and give it precedence.
+  if (n.name !== undefined) {
```
Kedae: Still don't get it. Name is the second choice of mainValue, so if is defined it will be applied...

CelineSebe: Because 'OR' operator returns the first truthy value it encounters.
 n.representative?.main = threat-actor--e984d89c-48e8-5c5e-8201-95517aff63bd, so n.name is not assigned.

opencti-platform/opencti-front/src/utils/defaultRepresentatives.ts
```diff
@@ -96,42 +96,42 @@ export const getMainRepresentative = (n: any, fallback = 'Unknown') => {
...
-    || getMainRepresentative((R.head(R.pathOr([], ['objects', 'edges'], n)) as any)?.node)
-    || (n.from
-      && n.to
-      && `${truncate(getMainRepresentative(n.from), 20)} ➡️ ${truncate(
-        getMainRepresentative(n.to),
-        20,
-      )}`)
-    || n.main_entity_name
-    || fallback;
+      || n.name
```
SouadHadjiat: did something change in this file? I only see reformatting

opencti-platform/opencti-graphql/src/database/entity-representative.js
```diff
@@ -83,8 +83,8 @@ export const extractEntityRepresentativeName = (entityData) => {
     mainValue = entityData.description;
   }
   // If no representative value found, return the standard id
-  if (isEmptyField(mainValue) || mainValue === 'Unknown') {
-    return entityData.standard_id;
+  if (isEmptyField(mainValue)) {
+    return 'Unknown';
```
SouadHadjiat, line 87: so we want to return "Unknown" instead of standard_id ?

opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectsTimeline.jsx
```diff
@@ -28,38 +28,10 @@ const stixCoreObjectsTimelineQuery = graphql`
...
-            entity_type
-          }
-          ... on StixObject {
             representative {
               main
               secondary
             }
-          }
-          ... on StixDomainObject {
-            created
```
SouadHadjiat, line 57: created seem to be used in WidgetTimeline component

opencti-platform/opencti-graphql/src/database/entity-representative.js
Kedae: Not really "instead of", but we couldn't find any logic behind that choice of setting the id rather than "Unknown" when we don't know the representative.
Neither Romuald (you wrote the code) nor Julien could remember a reason, so for now I suggest we do what seems to be coherent with the front : hence "Unknown"

opencti-platform/opencti-front/src/utils/defaultRepresentatives.ts
lndrtrbn: And I think it is not the right formatting as we are using a 2 spaces formats in other files, not 4

opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectsTimeline.jsx
```diff
@@ -43,22 +43,9 @@ const stixCoreObjectsTimelineQuery = graphql`
...
-              main
-              secondary
-            }
-          }
-          ... on StixDomainObject {
-            created
-            modified
-          }
-          ... on StixCyberObservable {
-            observable_value
```
lndrtrbn, line 61: `observable_value` is also used in the function to get representatives

SouadHadjiat: we don't need it if we get representatives directly from the API

lndrtrbn: Or we do not need anymore as we retrieve representatives directly from backend ?

Kedae (approved):

Issue #6929, opened 2024-05-11
[RBAC - User] Information not updated after modification of Groups


Lhorus6:
## Description

When a User is configured and its Groups are changed, the Roles associated with the Group are not updated on the page. You need to refresh the page to see the correct Roles.

NB: The bug is just a display problem that requires to refresh the page to get the latest information. There are no right/security issues to report.

## Environment

OCTI 6.0.10

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Create Group A with Role A
2. Create Group B with Role B
3. Create User Test and add Group A and B
-> Refresh the page and you will see Role A and B on the page of the user 

![Screenshot 2024-05-11 110946](https://github.com/OpenCTI-Platform/opencti/assets/70956706/a73e522f-fd95-41e7-8efe-ee9d5866cac4)

4. Remove User A from the Group B
-> Don't refresh, you will still se Role B in the list of Role

![Screenshot 2024-05-11 111002](https://github.com/OpenCTI-Platform/opencti/assets/70956706/635e616f-037d-43a4-b1a7-7a546901ced0)

5. Refresh the page
-> The right Role are displayed

![Screenshot 2024-05-11 111014](https://github.com/OpenCTI-Platform/opencti/assets/70956706/a8693194-4c32-49ad-ad13-eb6dc7115c95)

PR #7782, opened 2024-07-23
[frontend] [RBAC - User] Information not updated after modification of Groups (#6929)

SarahBocognano:
### Proposed changes

* Add roles to query

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/6929

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/settings/users/edition/UserEditionGroups.jsx
+++ b/opencti-platform/opencti-front/src/private/components/settings/users/edition/UserEditionGroups.jsx
@@ -167,6 +167,10 @@ const UserEditionGroups = createFragmentContainer(UserEditionGroupsComponent, {
           }
         }
       }
+      roles {
+        id
+        name
+      }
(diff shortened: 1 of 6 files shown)


Archidoit (approved):

Archidoit:
In user overview, groups are ordered alphabetically, but not roles: 
image

opencti-platform/opencti-front/src/private/components/settings/users/edition/UserEditionGroups.jsx
```diff
@@ -167,6 +167,10 @@ const UserEditionGroups = createFragmentContainer(UserEditionGroupsComponent, {
           }
         }
       }
+      roles {
```
Archidoit, line 170: order roles alphabetically

Archidoit: order roles alphabetically

Archidoit (changes requested):

Archidoit (approved):

PR #7783, opened 2024-07-23
[backend] Allow iframe usage for public dashboard (#7756)

### LINKED ISSUE #7756: Unable to embed a "public dashboard" in a web page through an IFrame
## Description

It is not currently possible to integrate a public dashboard into an external web page (e.g. WordPress), which is one of the objectives of public dashboards.
This is not possible due to the presence or absence of certain HTTP Response headers required to allow content to be embedded in Iframe.

I managed to integrate a dashboard in a WordPress page by injecting this HTTP header (through a reverse proxy) in OpenCTI response: 
Content-Security-Policy "frame-ancestors https: http:";

But other HTTP headers returned by OpenCTI should also be reviewed as “X-Frame-Options” for older browsers.

A possible approach would be to be able to adapt these "HTTP headers" only on the dedicated endpoint for public dashboards (to evaluate)

## Environment

OpenCTI version: 6.2.5

aHenryJard:
### Proposed changes

* Allow iframe for "/public" resources if domain name is allowed in configuration.

see also https://github.com/OpenCTI-Platform/docs/pull/201

### Related issues

* closes #7756

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/config/default.json
+++ b/opencti-platform/opencti-graphql/config/default.json
@@ -3,6 +3,7 @@
     "port": 4000,
     "base_path": "",
     "base_url": "http://localhost:4000/",
+    "public_dashboard_authorized_domains": "",
     "enabled": true,
     "enabled_ui": true,
     "disabled_dev_features": [
--- a/opencti-platform/opencti-graphql/src/config/conf.js
+++ b/opencti-platform/opencti-graphql/src/config/conf.js
@@ -278,7 +278,7 @@ const AppBasePath = BasePathConfig.endsWith('/') ? BasePathConfig.slice(0, -1) :
 export const basePath = isEmpty(AppBasePath) || AppBasePath.startsWith('/') ? AppBasePath : `/${AppBasePath}`;
 
 const BasePathUrl = nconf.get('app:base_url')?.trim() ?? '';
-const baseUrl = BasePathUrl.endsWith('/') ? BasePathUrl.slice(0, -1) : BasePathUrl;
+export const baseUrl = BasePathUrl.endsWith('/') ? BasePathUrl.slice(0, -1) : BasePathUrl;
 
 export const getBaseUrl = (req) => {
   // If base url is defined, take it in priority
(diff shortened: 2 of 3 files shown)


richard-julien:
As frameAncestors is super seeding x-from-options, can you please rewrite this to use this option dynamically?
Idea for me is to have in file configuration for now, a list of allowed domains that can be used to have an iframe. (the idea is to put that in a UI configuration but conf could do the trick to start)
Like in default.conf > app { public_dashboard_authorized_domains: ['mydomain.com'] }
Then you use tge app.use trick to only add a frameAncestors to these domains whe uri match /public/dashboard/*
So the securityMiddleware will be always the same expect you put 'none' or the list of domains depending of the uri match a public dasboard.

frameAncestors: ['mydomain.com'], // Super seeding frameguard

Something like

```
 const buildSecurity= (frameAncestorsDomains) => helmet({.....)}
 
 app.use((req, res, next) => {
    const urlString = req.url;
    const frameAncestorsDomains = (urlString && (urlString.startsWith(`${basePath}/public/dashboard`)) ? conf.get{'app:.....) : ['none'];
    buildSecurity(frameAncestorsDomains);
  });
```

aHenryJard:
As I understand :
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/frame-src 
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/frame-ancestors

- frame-ancestors list the domain name that will use frames from OpenCTI, sor for example if "https://mywordpress.com" want to display a public dashboard then we need to add "mywordpress.com" in ancestors accepted list.

- frame-src: are URL from OpenCTI that can be exposed in an iframe, so I tried to add only public/* there but it's not working....

On current PR status, totally deny iframe, or allowed them works. But not allowed with restrictions to "/public" and "/static".

So I'm keeping as draft for now.

FYI I'm testing with an html deployed locally by an nginx (so on http://localhost) such as:
```
<head>

</head>
<body>
  <div style='display: flex'>
    <div style='margin: 12px'>
    <h3>Should work</h3>
    <iframe
    id="inlineFrameExample"
    title="Should works fine"
    width="800"
    height="600"
    src="http://localhost:4000/public/dashboard/coucou">
    </iframe>
  </div>

  <div style='margin: 12px'>
    <h3>Should be denied</h3>
    <iframe
      id="inlineFrameExample"
      title="Should be refused"
      width="800"
      height="600"
      src="http://localhost:4000/dashboard">
    </iframe>
  </div>
</div>

</body>
```

SouadHadjiat (approved):
pair reviewed and tested

Issue #7635, opened 2024-07-08
[Trash] Impossible to delete certain entities


Lhorus6:
## Description

Some entities cannot be deleted (an error appears).

In view of the error, it seems that we're trying to put inferred relationships in the trash can, which poses a problem.

## Environment

OCTI 6.2.2

## Reproducible Steps

Try to remove this entity : https://testing.octi.staging.filigran.io/dashboard/locations/countries/0e1865e1-feed-4d1f-8643-e920e45bfe9f

I tried to reproduce the bug with the following repro case, but was unsuccessful.

Steps to create the smallest reproducible scenario: (basically : Create an entity and then make sure you have an inferred relationship with it -> depending on the inference rules you've enabled)
1. Create Intrusion set A
2. Create an Country C
3. Create Region R 
4. Create a relationship Country C -> located at -> Region R
5. Create a relationship Intrusion set A -> targets -> Country A
6. Activate "Targeting propagation via location" inference rule
7. Wait for inference relation Intrusion set A -> targets -> Region R to be generated
8. Try to delete Intrusion set A

## Screenshots

![Screenshot 2024-07-08 120759](https://github.com/OpenCTI-Platform/opencti/assets/70956706/eadaa7e2-d86a-4c79-91d4-40d016237655)

![Screenshot 2024-07-08 120717](https://github.com/OpenCTI-Platform/opencti/assets/70956706/b263d097-3ca5-4f6f-a1b5-e20840557886)

romain-filigran:
Unable to reproduce error with your scenario

Lhorus6:
I didn't make it either. Based on the observed error, I deduced the cause and tried to reproduce accordingly, but without success.

- Either my hypothesis about the cause is wrong, or
- My hypothesis is correct, but the error only appears under certain conditions (conditions I haven't identified).

But on the link I provided you have the same error as me, right?

SamuelHassine:
Fully reproductible on the provided link. It seems it is when you have inferred relationships on an entity no?

romain-filigran:
Not sure that occurred on entity with inferred relationships. For example, it's possible to delete: https://testing.octi.staging.filigran.io/dashboard/locations/countries/a582cbd3-1d56-4678-bb32-c71bfd89f476

But I'm seeing new problems with restoration: 

- previously inferred relationships are no longer considered inferred after restoration)
[...]

PR #7808, opened 2024-07-25
[backend] fix impossible deletion of some entities

JeremyCloarec:
Issue was caused by the old has-reference relation that wasn't properly removed from some inferred relations.
Error message in testing:
"reason\":\"[1:74] mapping set to strict, dynamic introduction of [rel_has-reference] within [_doc] is not allowed\"}

### Proposed changes

* remove rel_has-reference field during reindex

### Related issues

* #7635 

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/database/engine.js
+++ b/opencti-platform/opencti-graphql/src/database/engine.js
@@ -3279,7 +3279,7 @@ export const elReindexElements = async (context, user, ids, sourceIndex, destInd
         index: destIndex
       },
       script: { // remove old fields that are not mapped anymore but can be present in DB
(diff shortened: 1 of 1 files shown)


SouadHadjiat:
do you have more information about the entity having this rel_has-reference field stored ? rel_has-reference should have been deleted in a past migration.

JeremyCloarec:
The migration '1692961699032-investigation-relations-from-has-reference-to-list-of-entities-and-relations-ids' didn't properly clean the has-reference relations. The removeInvestigationsReferencesFromInvestigatedEntities method didn't include the inferred indexes

labo-flg (approved):

Issue #6724, opened 2024-04-17
Ability to custom Entity overviews' widgets position at platform level


Jipegien:
## Use case

Ability to change "widget" positions in the Overview tab of Entities and save it for user, group or platform.

Ideally, ability to change content of "widgets" too.

labo-flg:
Still under FF, reopening until removed

PR #7825, opened 2024-07-26
[backend] Edit customization through the API (#6724)

Goumies:
### Proposed changes

* add dedicated API

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/6724

### Checklist


opencti-platform/opencti-graphql/src/modules/entitySetting/entitySetting.ts
```diff
@@ -93,6 +93,20 @@ export const ENTITY_SETTING_DEFINITION: ModuleDefinition<StoreEntityEntitySettin
...
+      type: 'object',
+      format: 'standard',
+      mandatoryType: 'no',
+      editDefault: false,
+      multiple: true,
+      upsert: true,
+      isFilterable: false,
+      mappings: [
+        { name: 'key', label: 'Key', type: 'string', format: 'short', mandatoryType: 'internal', upsert: true, editDefault: false, multiple: false, isFilterable: false },
+        { name: 'order', label: 'Order', type: 'numeric', precision: 'integer', mandatoryType: 'internal', upsert: true, editDefault: false, multiple: false, isFilterable: false },
```
labo-flg: I think we can completely get rid of the "order" everywhere.
We store an array ; the order is the index in the array.
When sending the new array, we send the array in the right order, no need for sorting it manually before storing in DB. This way we have a consistent API.

It's up to the frontend to send the right payload, the right array in the right order.
Today it can use a form with a numeric input for the order ; tomorrow it will be a drag'n'drop.... it's really a frontend problematic.

CelineSebe (approved):

labo-flg (approved):
Tested OK, code looks good

Issue #7838, opened 2024-07-29
Work cleanup is not working anymore


aHenryJard:
## Environment

1. OS (where OpenCTI server runs): { e.g. Mac OS 10, Windows 10, Ubuntu 16.4, etc. }
2. OpenCTI version: { e.g. OpenCTI 1.0.2 } 6.2.7
3. OpenCTI client: { e.g. frontend or python }
4. Other environment details:

## Reproducible Steps

Look for work that are older than connector_manager.works_day_range days (default 7) in elastic, in an existing platform there are many.

## Expected Output
Complete work cleanup after works_day_range  days

## Additional information

This is over here, the filter may not work anymore or something else if preventing the cleanup to be done.

https://github.com/OpenCTI-Platform/opencti/blob/b0c6d12d9747ebe2c83b0d0645c4e48ecb89cd2f/opencti-platform/opencti-graphql/src/manager/connectorManager.js#L79-L106

PR #7850, opened 2024-07-29
[backend] follow up fix for the work cleanup bug and tests coverage (#7838)

aHenryJard:
### Proposed changes

* Fix the index used for work cleanup (for platform with rolling, need to use *_history* and not just history)
* Fix pagination size parameter for elPaginate
* Add test coverage on work cleanup.

### Related issues

* #7838 
### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/types/work.d.ts
+++ b/opencti-platform/opencti-graphql/src/types/work.d.ts
@@ -16,7 +16,7 @@ interface Work {
   received_time: string | null
   processed_time: string | null
   completed_time: string | null
-  completed_number: string
+  completed_number: number
   messages: string[]
   errors: string[]
 }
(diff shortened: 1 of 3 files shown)


opencti-platform/opencti-graphql/src/types/work.d.ts
```diff
@@ -16,7 +16,7 @@ interface Work {
   received_time: string | null
   processed_time: string | null
   completed_time: string | null
-  completed_number: string
```
aHenryJard, line 19: I don't know why it was set as string, it's used as number in work domain
https://github.com/OpenCTI-Platform/opencti/blob/839fca43230f54fd4c71d08e03410ba151be2ea8/opencti-platform/opencti-graphql/src/domain/work.js#L193

SouadHadjiat (approved):

Issue #7786, opened 2024-07-23
Bad alignment in Data section of an overview


SarahBocognano:
![image](https://github.com/user-attachments/assets/596810d1-2529-49be-ae4a-241f827c233d)

PR #7854, opened 2024-07-30
[frontend] Bad alignment in Data section of an overview (#7786)

SarahBocognano:
### Proposed changes

* Remove style

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/7786


CelineSebe:
Tested ✅  I see a little height difference between the two blocks
image

Archidoit:
The 'analyst workbench' square size doesn't adapt when file are uploaded : 

image

(whereas the 'external reference files do adapt)

SamuelHassine:
Yes, we need to have a height at 100% to ensure squares are aligned.

Kedae (approved):

Issue #7846, opened 2024-07-29
When creating a relationship within a shared report, Cannot read properties of undefined (reading 'internal_id') 


yassine-ouaamou:
## Description

As a user from an organization with which a report has been shared, I get an error when creating a relationship between two entities within the report

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Share a report with an organization
2. Log in with a user from this organization (user is not super admin in my case)
3. Add two entities to the report
4. Create a relationship between both

## Expected Output

No error

## Actual Output

Cannot read properties of undefined (reading 'internal_id')

## Additional information

If we try to create the relationship again, then we have this error: **Restricted relation already exists**

SamuelHassine:
It seems the relationship is not correctly shared at creation with all the organizations of the current user.

romain-filigran:
@yassine-ouaamou @SamuelHassine Not able to reproduce with the current “rolling” version. Is there any chance that this has been fixed in the meantime?

yassine-ouaamou:
Reproduced in testing. I will send you more data to reproduce

romain-filigran:
Ok reproduced on testing

PR #7859, opened 2024-07-30
[backend] fix granted_refs during rel creation for users with org sharing capa but not in platform org (#7846)

JeremyCloarec:
### Proposed changes

* Fixed buildRelationData algorithm. A bug was occuring when a user had the org sharing capa but wasn't in the platform org: the created relation was invisible to the user who just created it

### Related issues

* #7846 

### Checklist


SouadHadjiat (approved):
tested locally

PR #7911, opened 2024-08-01
[frontend/backend] Fix default value of variable "first" that cannot be null (#7905)

### LINKED ISSUE #7905: Can no longer access the "Data" tab of an Incident
## Description

Can no longer access the "Data" tab of an Incident (Event > Incident) -> An error is raised

## Environment

OCTI 6.2.10

## Reproducible Steps

1. Go to an Incident ("Event > Incident")
2. Open the "Data" tab

Screenshot 2024-08-01 085431

Screenshot 2024-08-01 085439

lndrtrbn:
### Proposed changes

* In front: change count variable of markings query to first to sync with backend name
* In back: avoid passing null to Elastic for pagination because not acceptable value

### Related issues

* #7905 

### Checklist

### Further comments

Elastic does not accept `null` value for its "size" parameter. Which correspond to the "first" option in our backend. We manage this option with a default value in case it was `undefined` but not when it was `null`. Now both cases fall back in default value.

### DIFF
--- a/opencti-platform/opencti-graphql/src/database/engine.js
+++ b/opencti-platform/opencti-graphql/src/database/engine.js
@@ -2550,7 +2550,8 @@ const completeSpecialFilterKeys = async (context, user, inputFilters) => {
 
 const elQueryBodyBuilder = async (context, user, options) => {
   // eslint-disable-next-line no-use-before-define
-  const { ids = [], first = ES_DEFAULT_PAGINATION, after, orderBy = null, orderMode = 'asc', noSize = false, noSort = false, intervalInclude = false } = options;
+  const { ids = [], after, orderBy = null, orderMode = 'asc', noSize = false, noSort = false, intervalInclude = false } = options;
+  const first = options.first ?? ES_DEFAULT_PAGINATION;
   const { types = null, search = null } = options;
   const filters = checkAndConvertFilters(options.filters, { noFiltersChecking: options.noFiltersChecking });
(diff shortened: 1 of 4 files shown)


opencti-platform/opencti-graphql/src/database/engine.js
```diff
@@ -2974,7 +2976,8 @@ export const elPaginate = async (context, user, indexName, options = {}) => {
     );
 };
 export const elList = async (context, user, indexName, opts = {}) => {
-  const { first = ES_DEFAULT_PAGINATION, maxSize = undefined } = opts;
+  const { maxSize = undefined } = opts;
+  const first = opts.first ?? ES_DEFAULT_PAGINATION;
```
SouadHadjiat, line 2980: what is the difference between `const { first = ES_DEFAULT_PAGINATION } = opts;` and `const first = opts.first ?? ES_DEFAULT_PAGINATION;` ?

Kedae: The null handling. { first = ES... } does not apply if first is null.

aHenryJard (commented):
Can you add the reproduction of issue in this test ? https://github.com/OpenCTI-Platform/opencti/blob/11faded7a432cb1dc1c84b1c58f6d90ac66b0d13/opencti-platform/opencti-graphql/tests/02-integration/02-resolvers/markingDefinition-test.js#L88-L90

(I mean same query as we have with null => should not crash)

Archidoit (approved):

PR #7937, opened 2024-08-02
[backend/frontend] Remove last SETTINGS capabilities (#3304)

frapuks:
### Proposed changes

(each scope is in one commit)

Front :
- throw error if you used the SETTINGS capabilitie
- remove parent Security component in Root, because children root are already protected

Back
- replace authorized autorities of notification (+migration)
- replace authorized autorities of backgroundTask
- deprecate internalObject Subscription API
- replace the 'split on _' by the right method to check user capabilities in backgroundTask

### Related issues

- #3304 

### Checklist


opencti-platform/opencti-graphql/src/migrations/1722428479372-trigger-authorized-authorities.js
```diff
@@ -0,0 +1,30 @@
...
+
+export const up = async (next) => {
+  logApp.info(`${message} > started`);
+  const context = executionContext('migration');
+
+  // ------ Update triggers
+  const callback = async (notifiers) => {
+    for (let i = 0; i < notifiers.length; i += 1) {
+      const notifier = notifiers[i];
+      const patch = { authorized_authorities: ['SETTINGS_SECURITYACTIVITY'] };
```
SouadHadjiat: this change should only apply on Activity Triggers, not all triggers, right ?

frapuks: I have updated this migration to only apply on activity triggers.

opencti-platform/opencti-graphql/src/migrations/1723468670664-trigger-authorized-authorities.js
```diff
@@ -0,0 +1,32 @@
...
+
+export const up = async (next) => {
+  logApp.info(`${message} > started`);
+  const context = executionContext('migration');
+
+  // ------ Update triggers
+  const callback = async (notifiers) => {
+    for (let i = 0; i < notifiers.length; i += 1) {
+      const notifier = notifiers[i];
+      if (notifier.trigger_scope === 'activity') {
```
SouadHadjiat: I suggest to add this condition as a filter in `opts` (line 23), so that we directly list triggers that have this 'activity' scope, instead of listing all triggers.

SouadHadjiat: I actually think that it would be better to use `elUpdateByQueryForMigration` with a query targetting specifically triggers with scope 'activity', and updating with a script authorized_authorities. So that we update with only one query, instead of listing all triggers and updating them one by one.

frapuks: I have updated this migration with the new function `elUpdateByQueryForMigration`, and filter triggers by the trigger_scope 'activity'

opencti-platform/opencti-graphql/src/migrations/1724143733795-trigger-authorized-authorities.js
```diff
@@ -0,0 +1,34 @@
+import { logApp } from '../config/conf';
```
SouadHadjiat: Could you rebase master and update the timestamp of this migration please ?

frapuks: Done

SouadHadjiat (approved):
tested locally

PR #7942, opened 2024-08-02
[frontend] filters: lte and gt operators with dates (#7697)

### LINKED ISSUE #7697: [filters] 'lower than / equals' operator for date filters doesn't take 'equals' into account
## Description

[filters] 'lower than / equals' operator for date filters doesn't take 'equals' into account

## Reproducible Steps
- Go to the reports list
- Change the filter with : lower than / equals 01/07/2024
- You have only reports published before this date, but no report published at this date

## Expected output

With the lower than / equals operator, the equals should be taken into account

Archidoit:
### Proposed changes
- fix lte and gt operator filters with dates

### Related issues
https://github.com/OpenCTI-Platform/opencti/issues/7697


aHenryJard (approved):
Tested locally, works as expected.

PR #7949, opened 2024-08-03
[backend] Fix Bypass external ref issue (#7945)

### LINKED ISSUE #7945: Bypass External Ref not working as expecting
When setting the Bypass External Ref we expect to be able to not trigger any error without setting external ref.

But in the code we do not bypass the validation part, therefore we have an error.

Kedae:
Closes #7945 
Needs to be discussed with @SamuelHassine or @richard-julien regarding bypass enforce reference


SamuelHassine:
You're good to go @Kedae, we just forgot it I think.

SarahBocognano (approved):

Issue #7482, opened 2024-06-26
retention for FILES MVP


nino-filigran:
## Use case

Ensure that we have a short term solution for this https://github.com/OpenCTI-Platform/opencti/issues/3016

Archidoit:
Chunk 2 (built-in retention rules for files) to be done

labo-flg:
reopening until feature flag is removed

PR #7961, opened 2024-08-05
[backend] Add two built-in retention rules on files and workbenches (#7482)

Archidoit:
Chunk 2 of 7482

### Proposed changes
- Add two built-in retention rules (on for files and one for worbenches) at platform initialization for new platforms
- Add these two buit-in retention rules for existing platforms via a migration

### Related issues
https://github.com/OpenCTI-Platform/opencti/issues/7482

### Warning
PR to be merged just before the 6.3 release, when feature flag FILE_RETENTION_RULES is removed


Archidoit:
NOT PLANNED

PR #7966, opened 2024-08-06
[backend] display created field for csv mapper (#7226)

### LINKED ISSUE #7226: Created field not present but required for CSV Mapper
## Description

Any of the Cases items that have the MANDATORY field 'created' fail because the field does not show up in the CSV Mapper and does not automatically pull the field, even if it is present in the file.

## Environment

1. OS (where OpenCTI server runs): container
2. OpenCTI version: <6.1.8
3. OpenCTI client: frontend
4. Other environment details:

## Reproducible Steps

1. generate a CSV from outside source of RFIs, Incident Responses, etc
2. Create a basic CSV Map for any of those items and attempt to test, receive message about 'created' attribute missing.

## Expected Output

Ingest those entity types

## Actual Output

image

## Additional information

`{
   "category":"APP",
   "errors":[
      {
         "attributes":{
            "attribute":"created",
            "genre":"BUSINESS",
            "http_status":400,
            "representation":"#1 Case-Rfi"
         },
         "message":"Missing values for required attribute",
         "name":"FUNCTIONAL_ERROR",
         "stack":"FUNCTIONAL_ERROR: Missing values for required attribute\n    at error (/opt/opencti/build/src/config/errors.js:8:10)\n    at FunctionalError (/opt/opencti/build/src/config/errors.js:95:50)\n    at /opt/opencti/build/src/modules/internal/csvMapper/csvMapper-utils.ts:165:17\n    at Array.forEach (<anonymous>)\n    at /opt/opencti/build/src/modules/internal/csvMapper/csvMapper-utils.ts:159:8\n    at processTicksAndRejections (node:internal/process/task_queues:95:5)\n    at async Promise.all (index 0)\n    at validateCsvMapper (/opt/opencti/build/src/modules/internal/csvMapper/csvMapper-utils.ts:137:3)\n    at bundleProcess (/opt/opencti/build/src/parser/csv-bundler.ts:32:3)\n    at csvMapperTest (/opt/opencti/build/src/modules/internal/csvMapper/csvMapper-domain.ts:31:18)"
      }
   ],
   "inner_relation_creation":0,
   "level":"error",
   "message":"Missing values for required attribute",
   "operation":"Unspecified",
   "query_attributes":[
      [
         {
            "arguments":[
               [
                  {
                     "is_empty":true,
                     "name":"configuration",
                     "type":"Variable"
                  }
               ],
               [
                  {
                     "is_empty":true,
                     "name":"content",
                     "type":"Variable"
                  }
               ]
            ],
            "name":"csvMapperTest"
         }
      ]
   ],
   "size":484,
   "source":"backend",
   "time":3,
   "timestamp":"2024-06-04T14:19:25.888Z",
   "type":"READ_ERROR",
   "user":{
      "group_ids":[
         "668ad6ec-ff03-450f-889f-656371d42deb"
      ],
      "ip":"::ffff:172.19.0.1",
      "organization_ids":[
         
      ],
      "referer":"http://localhost:8080/dashboard/data/processing/csv_mapper",
      "socket":"query",
      "user_id":"88ec0c6a-13ce-5e39-b486-354fe4a7084f",
      "user_metadata":{
         
      }
   },
   "version":"6.1.8"
}`

frapuks:
### Proposed changes

- Remove `created` from INTERNAL_ATTRIBUTES list
- Display `created` field into csv mapper creation form

### Related issues

- close #7226 

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/modules/internal/csvMapper/csvMapper-domain.ts
+++ b/opencti-platform/opencti-graphql/src/modules/internal/csvMapper/csvMapper-domain.ts
@@ -100,7 +100,7 @@ export const csvMapperSchemaAttributes = async (context: AuthContext, user: Auth
   types.forEach((type) => {
     const attributesDef = schemaAttributesDefinition.getAttributes(type);
     const attributes: CsvMapperSchemaAttribute[] = Array.from(attributesDef.values()).flatMap((attribute) => {
-      if (INTERNAL_ATTRIBUTES.includes(attribute.name)) return [];
+      if (INTERNAL_ATTRIBUTES.includes(attribute.name) || attribute.label === 'Original creation date') return [];
       return [{
         name: attribute.name,
         label: attribute.label,
--- a/opencti-platform/opencti-graphql/src/modules/internal/csvMapper/csvMapper-domain.ts
+++ b/opencti-platform/opencti-graphql/src/modules/internal/csvMapper/csvMapper-domain.ts
@@ -17,6 +17,7 @@ import { schemaTypesDefinition } from '../../../schema/schema-types';
 import { ABSTRACT_STIX_CORE_RELATIONSHIP, ABSTRACT_STIX_CYBER_OBSERVABLE, ABSTRACT_STIX_DOMAIN_OBJECT, ABSTRACT_STIX_META_OBJECT } from '../../../schema/general';
 import { type BasicStoreEntityIngestionCsv, ENTITY_TYPE_INGESTION_CSV } from '../../ingestion/ingestion-types';
 import { FunctionalError } from '../../../config/errors';
+import { created } from '../../../schema/attribute-definition';
 
 // -- UTILS --
 
@@ -100,7 +101,7 @@ export const csvMapperSchemaAttributes = async (context: AuthContext, user: Auth
   types.forEach((type) => {
     const attributesDef = schemaAttributesDefinition.getAttributes(type);
     const attributes: CsvMapperSchemaAttribute[] = Array.from(attributesDef.values()).flatMap((attribute) => {
-      if (INTERNAL_ATTRIBUTES.includes(attribute.name) || attribute.label === 'Original creation date') return [];
+      if (INTERNAL_ATTRIBUTES.includes(attribute.name) || attribute === created) return [];
       return [{
         name: attribute.name,
         label: attribute.label,
--- a/opencti-platform/opencti-graphql/src/modules/internal/csvMapper/csvMapper-domain.ts
+++ b/opencti-platform/opencti-graphql/src/modules/internal/csvMapper/csvMapper-domain.ts
@@ -17,7 +17,6 @@ import { schemaTypesDefinition } from '../../../schema/schema-types';
 import { ABSTRACT_STIX_CORE_RELATIONSHIP, ABSTRACT_STIX_CYBER_OBSERVABLE, ABSTRACT_STIX_DOMAIN_OBJECT, ABSTRACT_STIX_META_OBJECT } from '../../../schema/general';
 import { type BasicStoreEntityIngestionCsv, ENTITY_TYPE_INGESTION_CSV } from '../../ingestion/ingestion-types';
 import { FunctionalError } from '../../../config/errors';
-import { created } from '../../../schema/attribute-definition';
 
 // -- UTILS --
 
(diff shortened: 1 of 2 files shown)


opencti-platform/opencti-graphql/src/modules/internal/csvMapper/csvMapper-domain.ts
```diff
@@ -100,7 +100,7 @@ export const csvMapperSchemaAttributes = async (context: AuthContext, user: Auth
   types.forEach((type) => {
     const attributesDef = schemaAttributesDefinition.getAttributes(type);
     const attributes: CsvMapperSchemaAttribute[] = Array.from(attributesDef.values()).flatMap((attribute) => {
-      if (INTERNAL_ATTRIBUTES.includes(attribute.name)) return [];
+      if (INTERNAL_ATTRIBUTES.includes(attribute.name) || attribute.label === 'Original creation date') return [];
```
SouadHadjiat: I don't understand this fix, could you explain a bit more how this works ?

frapuks: the name `created` is used for 2 types of attributes :
- Created (only on Incident response, Request for information, Request for takedown)
- Original creation date (for many of other entities)

We want to display only the first one `created` for these 3 entities, not the second one `Original creation date`

SouadHadjiat: I think it's a coincidence that the label of created attribute is not the same for Cases and other entities, and I don't think we should have a test on labels like "Original creation date" that could change.

I don't know yet what we could do. Maybe add a condition on mandatory attributes ? INTERNAL_ATTRIBUTES list is not enough.

frapuks: I have improved the condition. We can discuss to improve it if necessary

opencti-platform/opencti-graphql/src/modules/internal/csvMapper/csvMapper-domain.ts
```diff
@@ -100,7 +101,7 @@ export const csvMapperSchemaAttributes = async (context: AuthContext, user: Auth
   types.forEach((type) => {
     const attributesDef = schemaAttributesDefinition.getAttributes(type);
     const attributes: CsvMapperSchemaAttribute[] = Array.from(attributesDef.values()).flatMap((attribute) => {
-      if (INTERNAL_ATTRIBUTES.includes(attribute.name)) return [];
+      if (INTERNAL_ATTRIBUTES.includes(attribute.name) || attribute === created) return [];
```
SouadHadjiat: I still don't like this solution. Why do we ignore this field created ? it's an information that could be useful for all entities and could be present in a CSV file or feed, no ? @nino-filigran

frapuks:
CURENTLY : We have 2 different attributes with the same `name : 'created'` :
- a general attributes referenced in `attribute-definition.ts` (line 366)
- a specific attribute referenced for example in `case-incident.ts` (line 39)
- The only condition to filter results is based on the `INTERNAL_ATTRIBUTES` list, created with only the property `name`

TARGET : The fix should be able to separate these two attributes and only return the second.

PROPOSALS :
- [ ] remove `created.name` from `INTERNAL_ATTRIBUTES` and filter manually the `created` (general) attribute ? => this PR
- [ ] update `INTERNAL_ATTRIBUTES` to list all attribute properties (not only the `name`) ?
- [ ] update the `name` of geneal attribute ?
- [ ] update the `name` of specific attribute ?
- [ ] keep `INTERNAL_ATTRIBUTES` and filter on the `attribute.mandatoryType === 'external'`  ?

nino-filigran:
@frapuks & @SouadHadjiat I'm not sure to understand this issue between create name internal and general. Could you elaborate?
- I see that a date (incident date/RFI date...) is needed when creating manually a case. Does this "created" attribute correspond to this date field?
- Can we simply ensure that we have this field displayed in the CSV mapper for these 3 entities? If any of the solutions you offered @frapuks are working then it's good for me.

SouadHadjiat:
`created` field is not internal, it's the "original creation date" of the entity and it's part of stix, we have another field `createdAt` that is the date the entity has been created in the platform (and this is internal). So my question is why `created` is considered internal? Why don't we display this field in the CSV representation for all entities? @nino-filigran @frapuks 

image

image

nino-filigran:
I think we can indeed add it in the CSV, but not as a mandatory field, especially useful if the entity has been created in the past before being imported in OpenCTI.

frapuks:
Decision with @SouadHadjiat and @nino-filigran :
Display the field `created` for all entities.

SouadHadjiat (approved):

PR #7969, opened 2024-08-06
[backend/frontend] Configure TAI overview layout from settings (#6724)

labo-flg:
Chunk 3 of #6724 : being able to change the overview of an entity from the settings > customization screens

This is under Feature Flag `OVERVIEW_LAYOUT_CUSTOMIZATION` for now, and limited to Threat Actor Individuals only.

It includes also the preview panel to show the result of the configuration in the setting page.

We also included changing order of the widgets with _drag and drop_. 
Changing order with input numbers was actually impractical.

### Checklist


opencti-platform/opencti-front/package.json
```diff
@@ -12,6 +12,7 @@
     "@emotion/styled": "11.13.0",
     "@fontsource/geologica": "5.0.5",
     "@fontsource/ibm-plex-sans": "5.0.20",
+    "@hello-pangea/dnd": "^16.6.0",
```
labo-flg, line 15: `react-beautiful-dnd` is no longer maintained. `@hello-pangea/dnd` is its direct replacement (a well maintained fork).

opencti-platform/opencti-front/src/private/components/settings/sub_types/Root.tsx
```diff
@@ -21,27 +16,15 @@ export const subTypeQuery = graphql`
 
 const RootSubType = () => {
   const { subTypeId } = useParams() as { subTypeId: string };
+
+  const data = useLazyLoadQuery<RootSubTypeQuery>(subTypeQuery, { id: subTypeId });
```
labo-flg, line 20: using QueryRenderer caused a full page refresh every time one setting is changed  (not only the settings related to layout)

CelineSebe:
Is it not possible to use createRefetchContainer to refresh only the component EntitySettingsCustomOverview? I think this could avoid navigating to the top of the page after a switch.

CelineSebe:
Tested ✅ Everything is ok 👌

labo-flg:
> Is it not possible to use createRefetchContainer to refresh only the component EntitySettingsCustomOverview? I think this could avoid navigating to the top of the page after a switch.

Thanks for the hint ! I can investigate in that direction, but maybe after this PR. I'll need a relay expert to help :)
Note that createRefetchContainer is legacy API, we should use[ relay hooks](https://relay.dev/docs/api-reference/use-refetchable-fragment/) now

opencti-platform/opencti-front/src/private/Root.tsx
```diff
@@ -311,7 +311,7 @@ const RootComponent: FunctionComponent<RootComponentProps> = ({ queryRef }) => {
   const overviewLayoutCustomizationEntries = entitySettingsData
     ?.map(({ target_type, overview_layout_customization }) => ({ key: target_type, values: overview_layout_customization }))
     .filter((entry) => !!entry.values)
-    .map(({ key: entityTypeKey, values: widgetsValues }) => [entityTypeKey, new Map(widgetsValues?.map(({ key, width }) => [key, width]))]);
+    .map(({ key: entityTypeKey, values: widgetsValues }) => [entityTypeKey, widgetsValues]);
```
JeremyCloarec: Instead of adding the overviewLayourCustomization in the UserContext, we can probably move this logic to the useOverviewLayourCustomization hook, and retrieve the overview layour from the entitySettings in the UserContext

opencti-platform/opencti-graphql/src/modules/entitySetting/entitySetting-domain.ts
```diff
@@ -63,6 +63,19 @@ export const findAll = (context: AuthContext, user: AuthUser, opts: QueryEntityS
...
+      return editInput.key === 'overview_layout_customization'
+        ? ({
+          key: editInput.key,
+          value: editInput.value
+            .sort((previous, current) => previous.order - current.order),
+          operation: editInput.operation
+        })
+        : (editInput);
+    });
+};
```
JeremyCloarec: overview_layout_customization inputs aren't correct: we keep the order sent by the frontend in the attributes, and the indexing fails because there is no 'order' attribute in the mapping.
'order' attribute needs to be removed from the editInput.value object after the sort

opencti-platform/opencti-graphql/src/modules/entitySetting/entitySetting-domain.ts
```diff
@@ -80,7 +93,7 @@ export const entitySettingEditField = async (context: AuthContext, user: AuthUse
     }
   }
 
-  const { element } = await updateAttribute(context, user, entitySettingId, ENTITY_TYPE_ENTITY_SETTING, input);
+  const { element } = await updateAttribute(context, user, entitySettingId, ENTITY_TYPE_ENTITY_SETTING, entitySettingOverviewLayoutCustomizationEditField(input));
```
JeremyCloarec: I think it could be cleaner and easier to understand to split this line into two different ones,
Have a first line modyfing the overview_layout_customization from the input, and then call the updateAttribute with the updated inputs.
```
const cleanedInputs = entitySettingOverviewLayoutCustomizationEditField(input);
const { element } = await updateAttribute(context, user, entitySettingId, ENTITY_TYPE_ENTITY_SETTING, cleanedInputs );
```

opencti-platform/opencti-graphql/src/modules/entitySetting/entitySetting.ts
```diff
@@ -103,8 +103,9 @@ export const ENTITY_SETTING_DEFINITION: ModuleDefinition<StoreEntityEntitySettin
       upsert: true,
       isFilterable: false,
       mappings: [
```
JeremyCloarec, line 105: The overview_layout_customization attribute is set to multiple: true, I don't think that's correct?

opencti-platform/opencti-front/src/utils/hooks/useOverviewLayoutCustomization.ts
```diff
@@ -1,9 +1,30 @@
...
+      .filter((entry) => !!entry.values)
+      .map(({ key: entityTypeKey, values: widgetsValues }) => [entityTypeKey, widgetsValues]);
+    const overviewLayoutCustomizations = overviewLayoutCustomizationEntries
+      ? new Map(overviewLayoutCustomizationEntries.map(([key, values]) => [key, values]))
+      : new Map();
+    setOverviewLayoutCustomization(overviewLayoutCustomizations.get(entityType) ?? []);
+  }, [entitySettingsData, entityType]);
+
+  return overviewLayoutCustomization;
 };
```
labo-flg: @JeremyCloarec I moved the code from useAuth/Root here after discussions.

I also removed the related unit tests ; this was too complex to setup due to fragments. 
I did not spend too much time on fixing them, because I believe we should have e2e tests instead.

Kedae: Shouldn't it be a useMemo ?

Kedae: Or even a useCallback on entityType ?

labo-flg: definitely _not_ a useEffect indeed

opencti-platform/opencti-graphql/src/database/middleware.js
```diff
@@ -909,7 +909,8 @@ const rebuildAndMergeInputFromExistingData = (rawInput, instance) => {
         finalVal = value;
       }
     }
-    if (compareUnsorted(finalVal ?? [], currentValues)) {
+    // TODO: solve case where ordering is important and we should use regular 'compare'
+    if (key !== 'overview_layout_customization' && compareUnsorted(finalVal ?? [], currentValues)) {
```
labo-flg, line 913: I'm facing an issue with this check : I'm sending an update on an array field that is a simple reordering of the array.
The check detects no change since we use `compareUnsorted`.

This is a temporar fixy, we will find a better solution and implement it rapidly.

We can:
* remove this optimization
* remove this optimization for a static set of keys, like `overview_layout_customization`
* add an attribute in the schema for "multiple" attribute, like "preserveOrder" to use `compare` instead of `compareUnsorted` here
* ... any better idea?

labo-flg: ping @richard-julien WDYT ?

JeremyCloarec (approved):

opencti-platform/opencti-graphql/src/modules/entitySetting/entitySetting-domain.ts
labo-flg: code removed as we do not need to sort backend side anymore

opencti-platform/opencti-graphql/src/modules/entitySetting/entitySetting-domain.ts
labo-flg: fixed, we do need any order and use indexing as natural order. It's a frontend concern to send the right array.

opencti-platform/opencti-graphql/src/modules/entitySetting/entitySetting.ts
labo-flg: it is multiple: it's an array of widgets config

PR #7980, opened 2024-08-07
[backend] Kill session on synchro when an HTTP error is send to client. (#7842)

### LINKED ISSUE #7842: Platform is flooded with sessions when stopping a stream consumed by other remote platforms
## Description

1. Create a stream on platform A
2. Consume the stream with 2 other platforms for instance (B and C)
3. Stop the stream on platform A

=> Platform A starts to be flooded with one session every 10 seconds (each time platforms B and C are trying to reconnect).

The behavior is quiet dangerous if, for instance, you are serving multiple customers in a context of a POC, and then you disable the stream, it will create millions of sessions without control. Obviously, the best practice should be to disable also the users but I think we can try to maintain session / cookie opened for the retries instead of creating a new HTTP client for each retry query.

aHenryJard:
### Proposed changes

* Kill session when an HTTP error is send back because EventSource that is used on syncManager closes the stream automatically on HTTP code that are different from 200 or 300 (redirects).

From my test locally, session flood can happen:
- When remote OCTI stream sever is "running = false"
- When user has not enough rights (marking, capacity or other checks on data)

### Related issues

* closes #7842 

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/graphql/sseMiddleware.js
+++ b/opencti-platform/opencti-graphql/src/graphql/sseMiddleware.js
@@ -60,6 +60,16 @@ const MAX_CACHE_TIME = (conf.get('app:live_stream:cache_max_time') ?? 1) * ONE_H
 const MAX_CACHE_SIZE = conf.get('app:live_stream:cache_max_size') ?? 5000;
 const HEARTBEAT_PERIOD = conf.get('app:live_stream:heartbeat_period') ?? 5000;
 
+const sendErrorStatusAndKillSession = (req, res, httpStatus) => {
+  try {
+    res.status(httpStatus).end();
+    req.session.destroy();
+  } catch (error) {
+    // We don't care but can be interesting for debug.
+    logApp.info('Error when trying to kill a session', { error });
+  }
+};
+
 const createBroadcastClient = (channel) => {
   return {
     id: channel.id,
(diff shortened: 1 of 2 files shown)


opencti-platform/opencti-graphql/src/graphql/sseMiddleware.js
```diff
@@ -326,7 +338,8 @@ const createSseMiddleware = () => {
       await processor.start(startStreamId);
     } catch (err) {
       res.statusMessage = `Error in stream: ${err.message}`;
-      res.status(500).end();
+      sendErrorStatusAndKillSession(req, res, 500);
+      logApp.info('[ANGIE] genericStreamHandler 500 - Error, we need to kill session');
```
aHenryJard:
```suggestion
```

opencti-platform/opencti-graphql/src/graphql/sseMiddleware.js
```diff
@@ -306,7 +317,8 @@ const createSseMiddleware = () => {
       // Generic stream only available for bypass users
       if (!isUserHasCapability(sessionUser, BYPASS)) {
         res.statusMessage = 'Consume generic stream is only authorized for bypass user';
-        res.status(401).end();
+        sendErrorStatusAndKillSession(req, res, 401);
+        logApp.info('[ANGIE] genericStreamHandler 401 - Error, we need to kill session');
```
aHenryJard:
```suggestion
```

SouadHadjiat (approved):

PR #7991, opened 2024-08-08
[backend] fix entity organization sharing for users outside platform organization (#7974)

### LINKED ISSUE #7974: [Org segregation] Entities created with workbenches are not shared
## Description

When a user who is not a member of the main organization creates an object in the platform, this object is by default shared with his Organization (otherwise he would lose access to it directly).

However, when he creates entities using workbench, it is not shared with him.

## Environment

OCTI 6.2.11

## Reproducible Steps

1. Set up the platform like this
    - User A with all capabilities except "bypass all"
    - User A part of Org A
    - Org A is not platform Organization
2. With User A, create a workbench with entities inside (the entities must not already exist in the platform)
3. With User A, validate the workbench

With User A, try to find the entities you've created -> you won't find them.
With User admin, try to find the created entities -> they have been created, but are not shared with Org A.

## Expected Output

Entities created in a workbench by User A must be shared with Org A

## Actual Output

Entities are not shared with Org A and therefore User A no longer has access to them and cannot work on them.

JeremyCloarec:
### Proposed changes

* we now check that the INPUT_GRANTED_REFS array is not empty during creation to prevent users outside platform org but with organization restrict capability to lose access to created entity

### Related issues

* #7974 

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/database/data-builder.js
+++ b/opencti-platform/opencti-graphql/src/database/data-builder.js
@@ -117,7 +117,8 @@ export const buildEntityData = async (context, user, input, type, opts = {}) => 
     if (input[inputField] || relType === RELATION_GRANTED_TO) {
       // For organizations management
       if (relType === RELATION_GRANTED_TO && isSegregationEntity) {
-        if (isUserHasCapability(user, KNOWLEDGE_ORGANIZATION_RESTRICT) && input[inputField] && [...input[inputField]].length > 0) {
+        if (isUserHasCapability(user, KNOWLEDGE_ORGANIZATION_RESTRICT) && input[inputField]
+            && (!Array.isArray(input[inputField]) || input[inputField].length > 0)) {
           relToCreate.push(...buildInnerRelation(data, input[inputField], RELATION_GRANTED_TO));
         } else if (!user.inside_platform_organization) {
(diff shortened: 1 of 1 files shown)


opencti-platform/opencti-graphql/src/database/data-builder.js
```diff
@@ -117,7 +117,7 @@ export const buildEntityData = async (context, user, input, type, opts = {}) =>
     if (input[inputField] || relType === RELATION_GRANTED_TO) {
       // For organizations management
       if (relType === RELATION_GRANTED_TO && isSegregationEntity) {
-        if (isUserHasCapability(user, KNOWLEDGE_ORGANIZATION_RESTRICT) && input[inputField]) {
+        if (isUserHasCapability(user, KNOWLEDGE_ORGANIZATION_RESTRICT) && input[inputField] && [...input[inputField]].length > 0) {
```
Kedae: Hm... `input[inputField].length > 0 ?`

labo-flg:
could we add some backend tests to cover this ?

lndrtrbn (approved):

labo-flg:
> could we add some backend tests to cover this ?

I understand this is hard to put up from scratch right now.
We'll add this to the backlog as a technical task : coverage of orga sharing in tests

Issue #6509, opened 2024-03-28
Dashboard Refacto: Admin to View all, New list for public dash & Massive operations


nino-filigran:
## Use case

In the context of sharing public dashboards, admins must be able to list all dashboards created by all users of the platform in order to be able to disable or delete links for some specific dashboard if they feel like they pause a security risk in terms of leakage of data.

nino-filigran:
Link to https://github.com/OpenCTI-Platform/opencti/issues/4903

PR #8007, opened 2024-08-09
[frontend] Create public dashboard from list page (#6509)

lndrtrbn:
### Proposed changes

* Refacto form to create a public dashboard
* Can create a new public dashboard from the list of public dashboards

### Related issues

* #6509 

### Checklist


opencti-platform/opencti-front/src/private/components/workspaces/WorkspaceShareForm.tsx
lndrtrbn, line 1: Form has been moved in publicDashboards folder

PR #8031, opened 2024-08-14
[backend] Trial for improving import page

Kedae:


opencti-platform/opencti-front/src/private/components/data/import/ImportWorkbenchesContent.tsx
```diff
@@ -54,14 +54,6 @@ export const workbenchLineFragment = graphql`
...
-      entity_id
-      entity {
-        ... on StixObject{
-          representative {
-            main
-          }
-        }
-      }
     }
     works {
```
SouadHadjiat: do we really need works here ? since it's in the fragment FileWork_file ?

Kedae: Fixed + multiple improvement around relay

PR #8032, opened 2024-08-14
[front] fix date format (timestamp) in horizontal bars widget (#7921)

### LINKED ISSUE #7921: [Dashboard] Date displayed as non-human readable format (timestamp)
## Description

When I make a widget breakdown by date, the dates are timestamps, which is unreadable/incomprehensible by a human. The dates should be displayed in another format (e.g. DD-MM-YYYY)

Screenshot 2024-08-01 155258

## Environment

OCTI 6.2.10

## Reproducible Steps

1. Create a widget horizontal bar
2. Entities perspective
3. Filter on "Entity type: Intrusion set"
4. Parameters like this: 

Screenshot 2024-08-01 155307

 5. Validated and looked at the dates

CelineSebe:
### Proposed changes

* change horizontalBarsChartOptions

### Related issues

* #7921 

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/utils/Charts.js
+++ b/opencti-platform/opencti-front/src/utils/Charts.js
@@ -1,6 +1,7 @@
 import * as C from '@mui/material/colors';
 import { resolveLink } from './Entity';
 import { truncate } from './String';
+import { dateFormat, timestamp } from './Time';
 
 const colors = (temp) => [
   C.red[temp],
@@ -31,6 +32,7 @@ const toolbarOptions = {
       columnDelimiter: ',',
       headerCategory: 'category',
       headerValue: 'value',
+      // eslint-disable-next-line @typescript-eslint/no-shadow
       dateFormatter(timestamp) {
         return new Date(timestamp).toDateString();
       },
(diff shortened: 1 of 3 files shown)


SarahBocognano (approved):
Tested ✅

opencti-platform/opencti-front/src/utils/Charts.js
```diff
@@ -31,6 +32,7 @@ const toolbarOptions = {
       columnDelimiter: ',',
       headerCategory: 'category',
       headerValue: 'value',
+      // eslint-disable-next-line @typescript-eslint/no-shadow
```
lndrtrbn, line 34: Please do not use those disabling rules. Change the name of the argument instead

lndrtrbn (dismissed):

opencti-platform/opencti-front/src/utils/Charts.js
```diff
@@ -524,7 +526,17 @@ export const horizontalBarsChartOptions = (
     show: stackType !== '100%',
     labels: {
       show: stackType !== '100%',
-      formatter: (value) => (yFormatter && typeof value === 'number' ? yFormatter(value) : value),
+      formatter: (value) => {
+        if (yFormatter && typeof value === 'number') {
+          return yFormatter(value);
+        }
+        if (typeof value === 'string' && value.length === 15) {
```
lndrtrbn: What means this `value.length === 15` ?

opencti-platform/opencti-front/src/utils/Charts.js
lndrtrbn, line 1: In my opinion the fix should be done in the file calling the function `horizontalBarsChartOptions`, not in this file. We have the possibility already to pass a function `yFormatter` to format the value. So use this to transform the timestamp in date string when necessary directly in the widget component when the attribute is a date.

opencti-platform/opencti-front/src/utils/Charts.js
CelineSebe: Timestamp has 15 characters...

Issue #7623, opened 2024-07-07
[Bulk search] The column ordering doesn't work


Lhorus6:
## Description

During bulk search, ordering/sorting on marking column doesn't work.

![Screenshot 2024-07-07 121004](https://github.com/OpenCTI-Platform/opencti/assets/70956706/6e2a6080-3e65-4ca0-8d59-10dc62863bc8)

NB: It's possible that other columns don't work either, please check.

## Environment

OCTI 6.2.1

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Make a bulk search with lots of results
2. Play with the different columns to see if the sorting is done correctly

romain-filigran:
Reproduced with markings column

PR #8034, opened 2024-08-14
[frontend] [Bulk search] The column ordering of markings doesn't work (#7623)

SarahBocognano:
### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/7623


opencti-platform/opencti-front/src/private/components/SearchBulk.jsx
lndrtrbn, line 1: The column Creators is also not working (same reason: it's an array)

lndrtrbn (approved):

PR #8052, opened 2024-08-16
[frontend/backend] authorized members bypass organization sharing for case IR (#4538)

marieflorescontact:
### Proposed changes

*  Add filter in elastic query to let user access entity if they is in authorized_members (even if user is member of the organization)
* backend tests
* Add refresh session if plateform organization is updated
* Ensure session is updated in the cache
*  add 1 identity => +1 live event
*  5 create / 4 updates / 2 delete => +9 raw events

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/4538
*

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/cases/case_incidents/Root.tsx
+++ b/opencti-platform/opencti-front/src/private/components/cases/case_incidents/Root.tsx
@@ -143,7 +143,7 @@ const RootCaseIncidentComponent = ({ queryRef, caseId }) => {
         container={caseData}
         PopoverComponent={<CaseIncidentPopover id={caseData.id} />}
         EditComponent={
-          <Security needs={[KNOWLEDGE_KNUPDATE]}>
+          <Security needs={[KNOWLEDGE_KNUPDATE]} hasAccess={currentAccessRight.canEdit}>
             <CaseIncidentEdition caseId={caseData.id} />
           </Security>
         }
--- a/opencti-platform/opencti-front/src/private/components/cases/case_incidents/Root.tsx
+++ b/opencti-platform/opencti-front/src/private/components/cases/case_incidents/Root.tsx
@@ -32,6 +32,7 @@ import useGranted, { KNOWLEDGE_KNUPDATE_KNBYPASSREFERENCE, KNOWLEDGE_KNUPDATE } 
 import { getCurrentTab, getPaddingRight } from '../../../../utils/utils';
 import CaseIncidentEdition from './CaseIncidentEdition';
 import useHelper from '../../../../utils/hooks/useHelper';
+import { useGetCurrentUserAccessRight } from '../../../../utils/authorizedMembers';
 
 const subscription = graphql`
   subscription RootIncidentCaseSubscription($id: ID!) {
@@ -104,9 +105,7 @@ const caseIncidentAuthorizedMembersMutation = graphql`
 `;
 
 const RootCaseIncidentComponent = ({ queryRef, caseId }) => {
-  const subConfig = useMemo<
-  GraphQLSubscriptionConfig<RootIncidentSubscription>
-  >(
+  const subConfig = useMemo<GraphQLSubscriptionConfig<RootIncidentSubscription>>(
     () => ({
       subscription,
       variables: { id: caseId },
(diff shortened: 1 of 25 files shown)


opencti-platform/opencti-graphql/src/database/session.js
```diff
@@ -112,14 +112,23 @@ export const killUserSessions = async (userId) => {
 };
 
 export const markSessionForRefresh = async (id) => {
+  const { store } = applicationSession;
   const currentSession = await getSession(id);
   if (currentSession) {
     const newSession = { ...currentSession, session_refresh: true };
-    await setSession(id, newSession, currentSession.expiration);
+    const sessId = id.includes(store.prefix) ? id.split(store.prefix)[1] : id;
```
Kedae: I don't understand the change

SouadHadjiat: we now use store.set which calls setSession and also sets the session in the cache, instead of setSession. And in order to call store.set we need to remove the prefix. This is what we also do when we kill a session.

opencti-platform/opencti-front/src/utils/authorizedMembers.ts
```diff
@@ -54,8 +55,9 @@ export const authorizedMembersToOptions = (
     });
 };
 
-export const getCurrentUserAccessRight = (userAccessRight: string | null | undefined) => {
-  const canManage = userAccessRight === 'admin';
+export const useGetCurrentUserAccessRight = (userAccessRight: string | null | undefined) => {
+  const isBypass = useGranted([BYPASS]);
```
SouadHadjiat: why ?

opencti-platform/opencti-front/src/private/components/cases/case_incidents/Root.tsx
```diff
@@ -126,148 +125,148 @@ const RootCaseIncidentComponent = ({ queryRef, caseId }) => {
   const isOverview = location.pathname === `/dashboard/cases/incidents/${caseData?.id}`;
   const paddingRight = getPaddingRight(location.pathname, caseData?.id, '/dashboard/cases/incidents', false);
   const { isFeatureEnable } = useHelper();
-  const canManageAuthorizedMembers = caseData?.currentUserAccessRight === 'admin' && isFeatureEnable('CONTAINERS_AUTHORIZED_MEMBERS');
+
+  if (!caseData) {
+    return <ErrorNotFound />;
+  }
+
+  const currentAccessRight = useGetCurrentUserAccessRight(caseData);
```
SouadHadjiat: useGetCurrentUserAccessRight should be called on caseData.currentUserAccessRight. I don't understand how it can work with caseData

opencti-platform/opencti-front/src/private/components/common/form/AuthorizedMembersField.tsx
```diff
@@ -70,15 +70,16 @@ const AuthorizedMembersField = ({
...
   showCreatorLine = false,
   canDeactivate = false,
 }: AuthorizedMembersFieldProps) => {
   const { t_i18n } = useFormatter();
-  const { me } = useAuth();
+  const { me, settings } = useAuth();
   const { setFieldValue } = form;
   const { name, value } = field;
 
+  const showAllMembersLine = !settings.platform_organization?.id;
```
SouadHadjiat: Just to be sure: we don't want to show "All" when a platform orga is set, right ?

opencti-platform/opencti-front/src/utils/authorizedMembers.ts
Kedae: A bypass user is considered as canManage of everything, why does it seem to be a problem ?

opencti-platform/opencti-front/src/private/components/cases/case_incidents/Root.tsx
Kedae: Yep, don't think this is working

opencti-platform/opencti-front/src/private/components/common/form/AuthorizedMembersField.tsx
Kedae: Exactly :) BUT I think I missed a part, where this is only valid for Knowledge and not Dashboards for example

opencti-platform/opencti-graphql/src/database/engine.js
```diff
@@ -500,10 +502,29 @@ export const buildDataRestrictions = async (context, user, opts = {}) => {
         }
         // For tasks
         should.push({ match: { 'initiator_id.keyword': user.internal_id } });
+
         // Finally build the bool should search
-        must.push({ bool: { should, minimum_should_match: 1 } });
+        if (userAccess.length === 0) {
```
SouadHadjiat: I don't understand this condition, neither the whole change in the query

aHenryJard:
@marieflorescontact can you update the PR title please ?

opencti-platform/opencti-graphql/src/utils/authorizedMembers.ts
```diff
@@ -106,6 +109,11 @@ export const editAuthorizedMembers = async (
       return isValidMemberAccessRight(value.access_right) && array.findIndex((e) => e.id === value.id) === index;
     });
 
+    const settings = await getEntityFromCache<BasicStoreSettings>(context, user, ENTITY_TYPE_SETTINGS);
+    if (filteredInput.some(({ id }) => id === MEMBER_ACCESS_ALL) && settings.platform_organization && isInternalObject(entityType)) {
```
SouadHadjiat: should be `!isInternalObject`

opencti-platform/opencti-front/src/private/components/common/form/AuthorizedMembersField.tsx
SouadHadjiat: Yes, ALL is not visible anymore on custom dashboards when a platform orga is set.

marieflorescontact:
bug known:
accessing default dahsboard with a user from different organization is not possible
image

SouadHadjiat (approved):

opencti-platform/opencti-front/src/private/components/common/containers/ContainerHeader.jsx
```diff
@@ -753,6 +753,7 @@ const ContainerHeader = (props) => {
       ],
     },
   };
+  const { canEdit } = useGetCurrentUserAccessRight(container.currentUserAccessRight);
```
SouadHadjiat: this causes an issue for all containers that doesn't have authorized members. We need to check that container.currentUserAccessRight is well defined before calling the method

PR #8054, opened 2024-08-16
[backend] Fix UI issue when changing organization in a user  (#7682)

### LINKED ISSUE #7682: Changing "Organizations" in a user make the page full re-render and close the drawer
## Description

1. Go to Security => Users
2. Go to a user
3. Edit the user (without any organization)
4. Add an organization

=> Drawer is closed, page is full re-rendered

CelineSebe:
### Proposed changes

* replace return user in assignOrganizationToUser

### Related issues

* #7682 

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/domain/user.js
+++ b/opencti-platform/opencti-graphql/src/domain/user.js
@@ -419,6 +419,11 @@ export const assignOrganizationToUser = async (context, user, userId, organizati
   if (!isUserHasCapability(user, SETTINGS_SET_ACCESSES) && isUserHasCapability(user, VIRTUAL_ORGANIZATION_ADMIN)) {
     throw ForbiddenAccess();
   }
+  const targetUser = await storeLoadById(context, user, userId, ENTITY_TYPE_USER);
+  if (!targetUser) {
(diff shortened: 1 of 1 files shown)


opencti-platform/opencti-graphql/src/domain/user.js
```diff
@@ -419,6 +419,11 @@ export const assignOrganizationToUser = async (context, user, userId, organizati
   if (!isUserHasCapability(user, SETTINGS_SET_ACCESSES) && isUserHasCapability(user, VIRTUAL_ORGANIZATION_ADMIN)) {
     throw ForbiddenAccess();
   }
+  const targetUser = await storeLoadById(context, user, userId, ENTITY_TYPE_USER);
```
SouadHadjiat: I have a concern about this assignOrganizationToUser, I wonder if we shouldn't use another method, or add more checks when the user is an organization admin (without SETTINGS_SET_ACCESSES).

SouadHadjiat: I think we should at least use findById method instead of storeLoadById which will do the right security checks.
But I also wonder if this method is usefull for an organization admin. To be checked with product team.

SouadHadjiat (approved):

Issue #8050, opened 2024-08-16
Group members are inconsistent and can lead to mistakes on managing RBAC


SamuelHassine:
## Description

Checking / unchecking boxes is not doing anything:

![image](https://github.com/user-attachments/assets/a2a8cc0f-959a-4b7c-bfca-baa35e83d5d9)

So you don't really know happen, you always have a doubt about: "is the user in this group or not?".

https://testing.octi.staging.filigran.io/dashboard/settings/accesses/groups/ebc96a69-b81e-431a-98ca-6b961ae5415f

NB: Seems to happen when you have more than 20 members in a group.

Note: I've checked, everything is ok in Elastic + API answers.

PR #8098, opened 2024-08-19
[frontend] Group members are inconsistent and can lead to mistakes on managing RBAC (#8050)

SarahBocognano:
### Proposed changes

* Change default pagination of group's member query from 20 (by default) to 500

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/8050

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/settings/groups/GroupEditionContainer.tsx
+++ b/opencti-platform/opencti-front/src/private/components/settings/groups/GroupEditionContainer.tsx
@@ -39,7 +39,7 @@ const GroupEditionContainerFragment = graphql`
     rolesOrderMode: { type: "OrderingMode", defaultValue: asc }
   ) {
     id
-    members {
+    members(first: 500) {
       edges {
         node {
           id


opencti-platform/opencti-front/src/private/components/settings/groups/GroupEditionContainer.tsx
```diff
@@ -39,7 +39,7 @@ const GroupEditionContainerFragment = graphql`
     rolesOrderMode: { type: "OrderingMode", defaultValue: asc }
   ) {
     id
-    members {
+    members(first: 500) {
```
Kedae, line 42: This will be handled by DataTable, do not merge yet :)

SouadHadjiat:
What is the status of this PR @Kedae now that data tables have been merged?

labo-flg:
Not fixed by latest changes to DataTables.
`GroupEditionContainer` now uses a `DataTableWithoutFragment`, which does not include pagination. passed 20 members selected, the list in the drawer does not work right.

labo-flg:
@SarahBocognano please check against latest master if the issue is still here. 
You can apply your quick fix if it solves the problem.


@labo-flg After checking with lastest master branch, the problem still occurs. The fix made by this PR is fixing it though

SarahBocognano:
Update: This PR is still valid because de data tables didn't resolved the pagination problem

Gwendoline-FAVRE-FELIX (approved):

Issue #7943, opened 2024-08-02
Active session count is not performant on platform having lots of users


aHenryJard:
## Description

One telemetry data is the number of active users.

User sessions are stored in redis: to count active users it's necessary to call redis but in platform having many users and some technical (machine to machine) users that does not manage http session it lead to performances issues.

## Environment

1. OS (where OpenCTI server runs): { e.g. Mac OS 10, Windows 10, Ubuntu 16.4, etc. }
2. OpenCTI version: 6.2.11
3. OpenCTI client: { e.g. frontend or python }
4. Other environment details:

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. { e.g. Run ... }
2. { e.g. Click ... }
3. { e.g. Error ... }

SouadHadjiat:
closing this issue as we removed active users count from telemetry

PR #8119, opened 2024-08-20
[backend] Improve redis sessions list perf (#7943)

SouadHadjiat:
### Proposed changes

* prevent locking the event loop when parsing users sessions

### Related issues

* #7943 

### Checklist


opencti-platform/opencti-graphql/src/database/redis.ts
```diff
@@ -189,66 +197,109 @@ export const redisTx = async (client: Cluster | Redis, chain: (tx: ChainableComm
...
+        }
+      });
+    }
+  }
+  await getClientBase().del(PLATFORM_KEY_SESSIONS);
+  return true;
 };
-export const getSession = async (key: string) => {
+export const getSession = async (baseKeyId: string) => {
+  const key = `{${PLATFORM_KEY_SESSIONS}}:${baseKeyId}`; // Key of a list must be attached to the same redis slot
```
SouadHadjiat: I don't understand, does it mean the session key has now PLATFORM_KEY_SESSIONS as a prefix ?

opencti-platform/opencti-front/src/private/components/settings/SessionsList.jsx
```diff
@@ -163,14 +151,11 @@ class SessionsListComponent extends Component {
                   </ListItemIcon>
                   <ListItemText
                     primary={
-                      <div>
-                        <div className={classes.name}>{user.name}</div>
-                        <div className={classes.email}>{user.email}</div>
-                      </div>
+                      <div className={classes.name}>{user.name} (last 10 of {session.total} sessions)</div>
```
SouadHadjiat: TODO: need to change this part to display "last 10 of X" only when there are more then 10 sessions

opencti-platform/opencti-graphql/src/database/redis.ts
SouadHadjiat: something is wrong with this `{${PLATFORM_KEY_SESSIONS}}:${baseKeyId}`, when I debug this method, I find myself with a duplicate {${PLATFORM_KEY_SESSIONS}}: prefix, I'm not sure we should add it here, since the key is already named in redis with this prefix

opencti-platform/opencti-graphql/src/database/redis.ts
```diff
@@ -190,66 +198,109 @@ export const redisTx = async (client: Cluster | Redis, chain: (tx: ChainableComm
...
+      }
+    }
+    return keysElements;
   }
   return [];
 };
 // endregion
 
 // region session
+const PLATFORM_KEY_SESSIONS = 'platform_session';
```
SouadHadjiat, line 285: just to be sure: was there a reason to rename the key `platform_sessions` to `platform_session` (without the `s` at the end) ? maybe to make sure we don't use the existing key ?

opencti-platform/opencti-graphql/src/database/session.js
```diff
@@ -44,46 +44,66 @@ const createSessionMiddleware = () => {
...
+      if (maxSessionsPerUser === undefined || preparedSessions[currentUserId].sessions.length < maxSessionsPerUser) {
+        preparedSessions[currentUserId].sessions.push(data);
+      }
+    } else {
+      preparedSessions[currentUserId] = { sessions: [data], total: 1, isActiveUser: isActiveSession };
+    }
+    if (new Date().getTime() - startProcessingTime > MAX_EVENT_LOOP_PROCESSING_TIME) {
+      startProcessingTime = new Date().getTime();
+      await new Promise((resolve) => {
+        setImmediate(resolve);
```
SouadHadjiat, line 83: why would we need this setImmediate for this loop ? There is no parsing, what was blocking the event loop ?

opencti-platform/opencti-graphql/src/http/httpPlatform.js
```diff
@@ -351,7 +351,6 @@ const createApp = async (app) => {
           event_scope: 'logout',
           context_data: undefined
         });
-        await delUserContext(user);
```
SouadHadjiat, line 351: why remove this line ? we don't need to clear user context anymore ?

SouadHadjiat:
is it still a PR to be worked on or should we close it @richard-julien ?

richard-julien:
I think you can close it.

PR #8164, opened 2024-08-26
[backend] fix elements to delete filtering in retentionManager

### LINKED ISSUE #8150: [Retention policy] Entities are not deleted
## Description

I have configured a retention rule, since two days, that matches many entities (160 reports) but the entities are never deleted.

## Environment

OCTI 6.2.14

## Reproducible Steps

1. Create a retention rule
2. Apply some filters that will match a small set of reports, e.g. the test I did:

image

3. Create the rule and wait for it to execute (you can wait 24 hours if you want). You will see that nothing is ever deleted

NB: I ran the tests on "testing". I'm leaving my retention rule active if you want to go see it (named "Test deletion issue"). It was a customer who brought this problem to my attention.

#### My rule, active for 2 days

The value "remaining" does not decrease

Screenshot 2024-08-23 100225

#### The reports are there and not deleted.

Screenshot 2024-08-23 100447

JeremyCloarec:
### Proposed changes

* knowledge elements retrieved by the retentionManager were incorrectly going through the filter on their uploading status and works
* added non regression test

### Related issues

* #8150

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/tests/02-integration/04-manager/retentionManager-test.ts
+++ b/opencti-platform/opencti-graphql/tests/02-integration/04-manager/retentionManager-test.ts
@@ -6,10 +6,12 @@ import { utcDate } from '../../../src/utils/format';
 import { deleteElement, getElementsToDelete } from '../../../src/manager/retentionManager';
 import { allFilesForPaths } from '../../../src/modules/internal/document/document-domain';
 import { uploadToStorage } from '../../../src/database/file-storage-helper';
-import { elRawUpdateByQuery } from '../../../src/database/engine';
-import { READ_INDEX_INTERNAL_OBJECTS } from '../../../src/database/utils';
+import { elLoadById, elRawUpdateByQuery } from '../../../src/database/engine';
+import { READ_INDEX_INTERNAL_OBJECTS, READ_INDEX_STIX_DOMAIN_OBJECTS } from '../../../src/database/utils';
 import { DatabaseError } from '../../../src/config/errors';
 import { deleteFile, loadFile } from '../../../src/database/file-storage';
+import { deleteElementById } from '../../../src/database/middleware';
+import { ENTITY_TYPE_CONTAINER_REPORT } from '../../../src/schema/stixDomainObject';
 
 describe('Retention Manager tests ', () => {
   const context = testContext;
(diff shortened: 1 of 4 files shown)


labo-flg (approved):

richard-julien:
@labo-flg , @JeremyCloarec can we add a non regression test for this?

opencti-platform/opencti-graphql/tests/02-integration/04-manager/retentionManager-test.ts
```diff
@@ -224,14 +299,25 @@ describe('Retention Manager tests ', () => {
...
+    const reportsToDelete = await getElementsToDelete(context, 'knowledge', before);
+    expect(reportsToDelete.edges.length).toEqual(1); // workbench1 is the only workbench that has not been modified since 'before'
+    expect(reportsToDelete.edges[0].node.id).toEqual(report1Id);
+  });
   it('should delete the fetched files and workbenches', async () => {
     // delete file
     await deleteElement(context, 'file', fileId); // should delete fileToTestRetentionRule
     const files = await allFilesForPaths(testContext, ADMIN_USER, [globalPath]);
-    expect(files.length).toEqual(8); // 7 files from index-file-test + the 2 created files - fileToTestRetentionRule that should have been deleted
+    expect(files.length).toEqual(1); // 7 files from index-file-test + the 2 created files - fileToTestRetentionRule that should have been deleted
```
labo-flg: from 8 to 1 ? 
what about the comment

JeremyCloarec: oops, i changed it to test locally but forgot to revert it back, thanks!

richard-julien:
@JeremyCloarec @labo-flg be careful about where you put the test.
If we can its really better to activate rules at the end of all the tests to prevent any counting issue.
We also need to test the 3 types of rule

JeremyCloarec:
> @JeremyCloarec @labo-flg be careful about where you put the test. If we can its really better to activate rules at the end of all the tests to prevent any counting issue. We also need to test the 3 types of rule

We built onto the test file that already existed for retention manager testing. The file and workbench retention rules were already tested, we added the testing on the knowledge rules.
The test is run during integration tests though, so not at the end of all the tests, but counters have been updated accordingly

Issue #7378, opened 2024-06-14
Testing E2E: verify all URLs 


nino-filigran:
## Use case

Goal of the testing E2E of this iteration is to have tests than runs on all pages to ensure they do break.

PR #8167, opened 2024-08-26
[ci] Add a sandbox optional step for flaky e2e tests (#7378)

aHenryJard:
### Proposed changes

Because flaky end to end test run fine locally most of time, we need a way to run them without annoying everyone.

* Add a drone step to run flacky e2e test without failing build.
* Run after the "normal" end to end step to avoid having flake on tests because of the sandbox one.

### Related issues
* Relates to #7378 work, does not close it.

### Checklist


.drone.yml
```diff
@@ -378,4 +404,7 @@ volumes:
       path: /tmp/cache-node-frontend
   - name: cache-node-frontend-e2e
     host:
-      path: /tmp/cache-node-frontend-e2e
\ No newline at end of file
+      path: /tmp/cache-node-frontend-e2e
+  - name: cache-node-frontend-e2e-flake
+    host:
+      path: /tmp/cache-node-frontend-e2e-flake
```
aHenryJard:
```suggestion
```

aHenryJard:
I'm not sure that it's really needed, and I don't want to complexify the drone build if not necessary, keeping that work as draft for now.

PR #8169, opened 2024-08-26
[backend] Use of dedicated RabbitMQ queues for integrated feeders/ingestors (#7704)

### LINKED ISSUE #7704: Mirror Integrated feeds in built in connector to allow works monitoring (#7704)
The ingestors/feeders integrated into OpenCTI, such as TAXII, RSS and CSV ingestors push content into a common "push_sync" RabbitMQ queue used for other purposes (sync between OpenCTI platforms).
In order to have better visibility over the ingestion process of these ingestors, it would be preferable to have dedicated RabbitMQ queues and dedicated "built-in" connectors/users.

# Solution draft
Each type must introduce a built in connector (like csv)
Each connector must have their dedicated queue.
When ingester received information, a work must be created and bundle injected in the connector dedicated queue.
This way each ingester type will have a dedicated ingestion pipe and user will be able to follow the process

richard-julien:
See #7704
Valid for RSS, CSV and TAXII.
Streams is way more complex and so will be handled in another feature.


opencti-platform/opencti-front/lang/back/en.json
```diff
@@ -77,6 +77,7 @@
   "Body raw reference": "Body raw reference",
   "Bookmarks": "Bookmarks",
   "Born In": "Born In",
+  "Built in": "Built in",
```
SouadHadjiat: I suggest to use directly our existing "Built-in" translation

opencti-platform/opencti-graphql/src/domain/connector.ts
```diff
@@ -201,7 +222,27 @@ export const connectorTriggerUpdate = async (context: AuthContext, user: AuthUse
 // endregion
 
 // region syncs
-export const patchSync = async (context: AuthContext, user: AuthUser, id: string, patch: { runnning: boolean }) => {
+export const registerConnectorForIngestion = async (context: AuthContext, type: string, id: string, name: string, is_running: boolean) => {
```
SouadHadjiat: since we have 4 parameters here (type, id, name, is_running), I think it would be better to create a type and regroup them in one parameter (which will make sure we call the method with the right input)

opencti-platform/opencti-graphql/src/manager/ingestionManager.ts
```diff
@@ -50,6 +52,43 @@ const asArray = (data: unknown) => {
   }
   return [];
 };
+
+const updateBuiltInConnectorInfo = async (context: AuthContext, id: string, state: object) => {
+  // Patch the related connector
+  const connectorId = uuidv5(id, OPENCTI_NAMESPACE);
```
SouadHadjiat: I suggest to create a method `generateConnectorId` in identifier.js

opencti-platform/opencti-graphql/src/manager/ingestionManager.ts
```diff
@@ -50,6 +52,43 @@ const asArray = (data: unknown) => {
...
+  };
+  await patchAttribute(context, SYSTEM_USER, connectorId, ENTITY_TYPE_CONNECTOR, connectorPatch);
+};
+
+const pushBundleToConnectorQueue = async (context: AuthContext, ingestion: BasicStoreEntityIngestionTaxii
+| BasicStoreEntityIngestionRss | BasicStoreEntityIngestionCsv, bundle: StixBundle) => {
+  // Push the bundle to absorption queue
+  const connector = { internal_id: uuidv5(ingestion.id, OPENCTI_NAMESPACE), connector_type: ConnectorType.ExternalImport };
+  const workName = `run @ ${now()}`;
+  const work: any = await createWork(context, SYSTEM_USER, connector, workName, connector.internal_id, { receivedTime: now() });
```
SouadHadjiat, line 82: Maybe we could create and use an INGESTION_MANAGER_USER ?

opencti-platform/opencti-graphql/src/manager/ingestionManager.ts
```diff
@@ -170,15 +210,15 @@ const rssDataHandler = async (context: AuthContext, httpRssGet: Getter, turndown
...
-    const content = Buffer.from(stixBundle, 'utf-8').toString('base64');
-    await pushToWorkerForSync({ type: 'bundle', applicant_id: ingestion.user_id ?? OPENCTI_SYSTEM_UUID, content, update: true });
+    await pushBundleToConnectorQueue(context, ingestion, bundle);
     // Update the state
-    const lastPubDate = R.last(items)?.pubDate;
+    lastPubDate = R.last(items)?.pubDate;
     await patchRssIngestion(context, SYSTEM_USER, ingestion.internal_id, { current_state_date: lastPubDate });
   }
+  // Patch the related connector
+  await updateBuiltInConnectorInfo(context, ingestion.id, { current_state_date: lastPubDate });
```
SouadHadjiat: what can we store exactly in this `connector_state` ? I see here `current_state_date`, and in other calls I see `current_state_cursor`, `taxii_more`, `added_after_start`. It's a json type, so it's store as a string text. What do we do with this state ?

richard-julien: We do nothing directlty with it. Its a way to see the current state of the ingester in the built in connector

opencti-platform/opencti-graphql/src/manager/ingestionManager.ts
richard-julien: I dont think its needed as we dont track who create the work.

aHenryJard (approved):
Tested locally on Taxii feed and RSS feed: looks good for me.
I had already existing taxii feed => migration is ok
I deleted a taxii feed => it's gone from UI and rabbitMQ push and listen queue are cleared
I added an RSS Feed all good. 
And OCTI sync still works as usual.

The only non bloking remark that I have is on the generated name: 
I find `[FEED] Local taxii on http://localhost:4010 (TAXII)` is less readable then if it was all at the same place like `[FEED][TAXII] Local taxii on http://localhost:4010`

image

opencti-platform/opencti-front/src/private/components/data/Sync.jsx
```diff
@@ -92,7 +92,7 @@ const Sync = () => {
         message={
           <>
             {t_i18n(
-              'You can configure your platform to consume remote OCTI streams. A list of public and commercial native feeds is available in the',
+              'You can configure your platform to consume OpenCTI Streams. A list of public and commercial native feeds is available in the',
```
aHenryJard: I think there is a missing translation for this text.

Issue #3016, opened 2023-03-14
Handle massive operation / retention for files


SamuelHassine:
## Use case

1. Add checkboxes with a bottom toolbar in global / import for workbenches and uploaded files to allow mass delete.
2. Add checkboxes with a bottom toolbar in any data / all files panel to allow mass delete of uploaded / exported / workbenches files.
3. Add a global settings to handle retention policy for all "pending" files (aka workbenches).

NB: Retention policy is only relevant for "pending files". All other types of file should not be deleted automatically.

richard-julien:
@Jipegien this one i think will start to be problematic short term. Need to discuss again about the scheduling.

romain-filigran:
related to https://github.com/OpenCTI-Platform/opencti/issues/7297

Archidoit:
done with the new import screen in 6.6

PR #8184, opened 2024-08-28
[frontend/backend] Global files filtering and ordering (#3016)

Archidoit:
### Proposed changes
- Fix ordering by markings in Data/Import/Uploaded files
- Add the possibility to filter global files and global workbenches list (Data/Import)
- Improve code
- Background tasks on global files and workbenches from the list
- Fix filters display in 'Launch a background task' pop-up

### Related issues
part of https://github.com/OpenCTI-Platform/opencti/issues/3016


opencti-platform/opencti-front/src/utils/filters/filtersUtils.tsx
```diff
@@ -455,12 +455,27 @@ export const addFilter = (
...
-    filters: (filters?.filters ?? []).concat([filterFromParameters]),
-    filterGroups: filters?.filterGroups ?? [],
+    mode: 'and',
+    filters: [filterFromParameters],
+    filterGroups: [filters],
   };
 };
 
+// create a new filter: filterGroup1 AND filterGroup2
+export const addFilterGroup = (
```
lndrtrbn: As general rule, an optional parameter should be the last. So in this case I think we should have `filter1` as first and mandatory and `filter2` as second and possible undefined. And when reading `addFilterGroup` I understand more "add a filter group inside an other one", here it seems we are joining 2 filter groups. And your function cannot return undefined so I don't understand why specifying undefined in signature.

What do you think of this:
```ts
export const joinFilterGroups = (
  filters1: FilterGroup,
  filters2: FilterGroup | undefined,
  mode = 'and',
): FilterGroup => {
  if (!filter2) return filter1;
  return {
    mode,
    filters: [],
    filterGroups: [filters1, filters2],
  };
}
```

Archidoit: indeed you're right! But the function is unused now so I'm gonna delete it :)

lndrtrbn (approved):

PR #8200, opened 2024-08-29
[frontend/backend] make unit mandatory for Retention Rules

### LINKED ISSUE #8190: Retention rules created before 6.2.14 have no unit
## Reproducible Steps
The retention rules created before 6.2.14 have no unit :

image

It's not critical since the retention manager put the unit to 'day' by default if no unit.

## Proposed solution
Do a migration that add the unit 'days' to all the retention rules with no unit

Archidoit:
### Proposed changes
- make retention_unit mandatory in RetentionRule
- add a migration to set retention_unit to 'days' for retention rules with no unit

### Related issues
https://github.com/OpenCTI-Platform/opencti/issues/8190

### DIFF
--- a/opencti-platform/opencti-front/src/schema/relay.schema.graphql
+++ b/opencti-platform/opencti-front/src/schema/relay.schema.graphql
@@ -1104,7 +1104,7 @@ input RetentionRuleAddInput {
   name: String!
   filters: String!
   max_retention: Int!
-  retention_unit: RetentionUnit
+  retention_unit: RetentionUnit!
   scope: RetentionRuleScope!
 }
 
@@ -1134,7 +1134,7 @@ type RetentionRule {
   name: String!
   filters: String!
   max_retention: Int!
-  retention_unit: RetentionUnit
+  retention_unit: RetentionUnit!
   last_execution_date: DateTime
   last_deleted_count: Int
   remaining_count: Int
--- a/opencti-platform/opencti-front/src/schema/relay.schema.graphql
+++ b/opencti-platform/opencti-front/src/schema/relay.schema.graphql
@@ -1102,9 +1102,9 @@ input QueryTaskAddInput {
 
 input RetentionRuleAddInput {
   name: String!
-  filters: String!
+  filters: String
   max_retention: Int!
-  retention_unit: RetentionUnit!
+  retention_unit: RetentionUnit
   scope: RetentionRuleScope!
 }
 
--- a/opencti-platform/opencti-graphql/config/schema/opencti.graphql
+++ b/opencti-platform/opencti-graphql/config/schema/opencti.graphql
@@ -1055,9 +1055,9 @@ input QueryTaskAddInput {
 
 input RetentionRuleAddInput {
   name: String! @constraint(minLength:2, format: "not-blank")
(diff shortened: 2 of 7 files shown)


opencti-platform/opencti-graphql/config/schema/opencti.graphql
```diff
@@ -1054,7 +1054,7 @@ input RetentionRuleAddInput {
   name: String! @constraint(minLength:2, format: "not-blank")
   filters: String!
   max_retention: Int! @constraint(min: 1)
-  retention_unit: RetentionUnit
+  retention_unit: RetentionUnit!
```
labo-flg: having the value mandatory in type `RetentionRule` I agree, if this is indeed a value always present when querying a rule.

But having the value mandatory in the input, is a **breaking change**.
If the API accepts nothing (and defaults to 'days', it's fine)

labo-flg (changes requested):

opencti-platform/opencti-graphql/config/schema/opencti.graphql
Archidoit: Indeed ! Fixed :)
I've also made filters not mandatory in RetentionRuleAddInput to facilitate creation via the API (user can write no filters instead of an empty filter group)

opencti-platform/opencti-front/src/schema/relay.schema.graphql
```diff
@@ -1099,7 +1099,7 @@ input QueryTaskAddInput {
 
 input RetentionRuleAddInput {
   name: String!
-  filters: String!
```
labo-flg, line 1105: 100% sure it defaults to an empty filter group ? :)

opencti-platform/opencti-graphql/src/domain/retentionRule.js
```diff
@@ -45,7 +45,10 @@ export const checkRetentionRule = async (context, input) => {
 // input { name, filters }
 export const createRetentionRule = async (_, user, input) => {
   // filters must be a valid json
-  const { filters } = input;
+  let { filters } = input;
+  if (!filters) { // filters is undefined or an empty string
+    filters = JSON.stringify({ mode: 'and', filters: [], filterGroups: [] });
```
labo-flg, line 50: ah yes. nevermind.

opencti-platform/opencti-front/src/schema/relay.schema.graphql
Archidoit: yes if you create a retention rule in the UI, no if you create it via the API...

opencti-platform/opencti-graphql/config/schema/opencti.graphql
lndrtrbn: I don't understand, it's not fixed? The fix should be to remove this "!" no? Or I am misunderstanding?

Archidoit: to remove the '!' for RetentionRuleAddInput indeed, but not for RetentionRule

lndrtrbn (approved):

labo-flg (approved):

PR #8218, opened 2024-08-30
[backend] filtering-utils fixes and tests

Archidoit:
### Proposed changes
- When using the addFilter function of backend, we want the result filter to be the concatenation of the old filter and a new filter, separated with a AND.
- add backend tests for functions in filtering-utils

### DIFF
--- a/opencti-platform/opencti-graphql/src/utils/filtering/filtering-utils.ts
+++ b/opencti-platform/opencti-graphql/src/utils/filtering/filtering-utils.ts
@@ -120,35 +120,34 @@ export const extractFilterGroupValues = (inputFilters: FilterGroup, key: string 
 };
 
 /**
- * Insert a Filter inside a FilterGroup
+ * Construct a filter: filterGroup AND (new filter constructed from key, values, operator and mode)
  * If the input filterGroup is not defined, it will return a new filterGroup with only the added filter (and / or).
  * Note that this function does input coercion, accepting string[] and string alike
  */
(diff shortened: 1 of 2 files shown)


SouadHadjiat:
what use case does this change fixes exactly? is it possible to write tests?

opencti-platform/opencti-graphql/src/utils/filtering/filtering-utils.ts
```diff
@@ -120,35 +120,34 @@ export const extractFilterGroupValues = (inputFilters: FilterGroup, key: string
...
-export const addFilter = (filterGroup: FilterGroup | undefined | null, newKey: string | string[], newValues: string | string[] | undefined | null, operator = 'eq'): FilterGroup => {
+export const addFilter = (filterGroup: FilterGroup | undefined | null, newKey: string | string[], newValues: string | string[] | undefined | null, operator = 'eq', localMode = 'or'): FilterGroup => {
   const keyArray = Array.isArray(newKey) ? newKey : [newKey];
   let valuesArray: string[] = [];
   if (newValues) {
     valuesArray = Array.isArray(newValues) ? newValues : [newValues];
   }
   return {
-    mode: filterGroup?.mode ?? 'and',
+    mode: 'and',
```
labo-flg, line 134: This looks like an impactful change... it was only used with 'and' and we were lucky until now ?

Archidoit: It was used with input filters enter by the user, and usually users keep the 'and' as global mode...

Archidoit: And indeed it's an important fix

labo-flg (approved):

Issue #7962, opened 2024-08-05
The derived-from relationship is not supported for all STIX domain objects and cyber observables


sweet-mentat:
## Description

I specifically want to have a "derived-from" relationship between two identity objects, but it would be great to get this relationship type supported for every SDO and SCO. According to the STIX 2.1 documentation (https://docs.oasis-open.org/cti/stix/v2.1/csprd01/stix-v2.1-csprd01.html#_Toc16070612), this is supported. I saw that there was a prior issue submitted (https://github.com/OpenCTI-Platform/opencti/issues/1059) where the "derived-from" relationship was requested between two Indicator Objects, which was implemented. Can this please be expanded to all objects?

## Environment

1. OS (where OpenCTI server runs): Docker image
2. OpenCTI version: 6.2.3
3. OpenCTI client: frontend and python

nino-filigran:
Fix: allow the creation of relation between 2 objects of the same type.

PR #8220, opened 2024-08-30
[backend/frontend] Add derived-from relationship in SDO & SCO of same type (#7962)

SarahBocognano:
### Proposed changes

* Add derived-from relationship to SDO & SCO of the same type (same behavior as related-to in front except when relationship between entity screen already exist, then it's on their respective screens)

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/7962


marieflorescontact (commented):
I don't know what is the scope so maybe it is ok but we cannot create relation `derived-from`  in UI in `Entity > knowledge` for entities:
Channel
Vulnerability
Narrative
Area
Event
Organization

marieflorescontact:
Did you forget translation ? 

==> No it was already existing, thank you for the review for the missing entities though !!

marieflorescontact (approved):
tested ok 👌

PR #8222, opened 2024-09-01
Add auto_create_users option

initstring:
### Proposed changes

* Implement `AUTO_CREATE_USERS` option for Google, GitHub, and Facebook auth

### Related issues

(N/A)

### Checklist

### Further comments

Hello! This is currently an untested draft pull request, as I'm still exploring and understanding the codebase.

This PR adds a configurable option to control the auto-creation of users from providers (Google, GitHub, Facebook). It sets this option to "true" by default, which will retain backwards compatibility with the existing behavior.

I see in the code and documentation that "providers are trusted" and users are automatically created. This means that if you enable Google Authentication, anyone with a Google account can log in. There is an option to restrict to specific email domains, which is great, but more granularity would be nice.

For example, I may want to use Google Authentication but only authorize users I've explicitly added to the platform. Google is the strategy I am personally testing, but this would probably apply to GitHub and Facebook auth as well.

With this new option, setting `PROVIDERS__GOOGLE__CONFIG__AUTO_CREATE_USERS` to false would mean that users could log in with Google Auth, but only if that user has an account already created on the server. This would allow an admin to have strict control over who logs in, even from their own Google domain.

A further iteration would be to allow only auto user creation from specific existing Google groups. I may look at that in the future.

Thank you for this great open source product.

### DIFF
--- a/opencti-platform/opencti-graphql/src/config/providers.js
+++ b/opencti-platform/opencti-graphql/src/config/providers.js
@@ -389,6 +389,7 @@ for (let i = 0; i < providerKeys.length; i += 1) {
     }
     if (strategy === STRATEGY_FACEBOOK) {
       const providerRef = identifier || 'facebook';
+      const autoCreateUsers = mappedConfig.auto_create_users || true;
       const specificConfig = { profileFields: ['id', 'emails', 'name'], scope: 'email' };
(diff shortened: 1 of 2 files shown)


initstring:
Closing out this PR, as it seems the better option is to simply use SAML. Thanks!

PR #8228, opened 2024-09-02
[backend/frontend] implement authorized members for containers (#4538)

marieflorescontact:
### Proposed changes

* declare authorized members for containers graphQL/Schema: Cases, Report, Grouping
* handle display of `Update` button / popover according to authorized members
* update tests

### Related issues

* #4538

### Checklist


opencti-platform/opencti-graphql/src/domain/container.js
```diff
@@ -213,3 +223,18 @@ export const knowledgeAddFromInvestigation = async (context, user, { containerId
   // Reload on this is mandatory to get the rel_ from the element for accurate counting
   return internalLoadById(context, user, patched.internal_id, patched.entity_type);
 };
+
+export const containerEditAuthorizedMembers = async (context, user, entityId, input) => {
+  if (!isFeatureEnabled('CONTAINERS_AUTHORIZED_MEMBERS')) {
```
SouadHadjiat: we should test if it's a feedback too, since feedback are not part of feature flag, or remove this feature flag test here

opencti-platform/opencti-graphql/src/modules/case/case-incident/case-incident.graphql
```diff
@@ -234,5 +234,4 @@ input CaseIncidentAddInput {
 type Mutation {
   caseIncidentAdd(input: CaseIncidentAddInput!): CaseIncident @auth
   caseIncidentDelete(id: ID!): ID @auth(for: [KNOWLEDGE_KNUPDATE_KNDELETE])
-  caseIncidentEditAuthorizedMembers(id: ID!, input:[MemberAccessInput!]): CaseIncident @auth(for: [KNOWLEDGE_KNUPDATE_KNMANAGEAUTHMEMBERS])
```
SouadHadjiat, line 237: this API was removed, it was feature flagged, so no breaking change

opencti-platform/opencti-graphql/src/domain/container.js
SouadHadjiat: To remove

Archidoit:
Remark about creating workbench inside an entity : 
Set the access to a report r1 to : everyone can read, admin can manage
Log in a user1 with bypass. The user can create a workbench inside the report (but can't edit the report).

Note that for the moment, a user only needs the capa 'import knowledge' to create a workbench inside a report (don't need to have the capa Knowledge update).

Archidoit:
A user with only read access on a container can't see the 'update' button, but can see the '+' button near 'labels' in the report overview : 
image

same for '+' icon near 'external references, '+' icon to add relationships in Knowledge tab, etc
Don't know if it's wanted

Archidoit:
Bad error format if you try to save accesses with no valid admin member : 
image

marieflorescontact:
> A user with only read access on a container can't see the 'update' button, but can see the '+' button near 'labels' in the report overview : image
> 
> same for '+' icon near 'external references, '+' icon to add relationships in Knowledge tab, etc Don't know if it's wanted

Thank you for your review @Archidoit . You made a good point. All  `+` buttons shouldn't be visible on entity if User has only `view` access right.

marieflorescontact:
> Bad error format if you try to save accesses with no valid admin member : image

Well, this part needs to be rework apparently. `Everyone on the platform` is visble only if you have no platform organization.  But  shoul we have `everyone` on `knowledge` ? @SouadHadjiat

SouadHadjiat:
PR status:
- Feature flag added on API side on authorized members update
- previous issues found by @Archidoit will be handled in separate PR

SouadHadjiat (approved):

PR #8238, opened 2024-09-03
[frontend/backend] Pagination for all workbenches (#3016)

Archidoit:
### Proposed changes
- In Data > Import > Analyst Workbenches
    - display all the workbenches (global workbenches and worbenches of entities)
    - possibility to easily select global workenches only (checkbox near the Search bar)
    - add a filter and a column for the workbench related entity
- in Data > Import : remove the 'Enabled import connectors' part
- All these new screens under IMPORT_NEW_SCREENS feature flag

### Related issues


Archidoit:
Deprogrammed, Will be done with new import files

PR #8239, opened 2024-09-03
Improved accessibility for static labels and chips

Bonsai8863:
### Proposed changes

* Changed text color for most labels/chips to pure black/white to improve readability
* Added monochromatic setting in user profile to make labels/chips completely grayscale

### Related issues

* #6609 

### Checklist

### Further comments

I didn't write auto tests since they are mostly style changes


opencti-platform/opencti-front/src/components/ItemColor.ts
```diff
@@ -0,0 +1,544 @@
...
+import { stringToColour } from '../utils/Colors';
+
+/**
+ * itemColor function that uses hooks to determine monochrome label and theme mode
+ * Can only be used in functional components
+ * @param type type of the entity
+ * @param reversed is the relation reversed
+ * @returns color hexcode
+ */
+const itemColor = (type: string, reversed = false) => {
```
Kedae, line 12: Isn't that the same as in utils/Colors ?

opencti-platform/opencti-front/src/components/ItemAccountStatus.tsx
```diff
@@ -72,6 +76,11 @@ const ItemAccountStatus: FunctionComponent<ItemAccountStatusProps> = ({
   account_status,
   variant,
 }) => {
+  const { me: { monochrome_labels } } = useAuth();
+  const { palette: { mode } } = useTheme();
+  const theme = mode === 'dark'
```
Kedae: I don't understand your code. 
```const theme = useTheme<Theme>();```
should be enough

File: opencti-platform/opencti-front/src/components/ItemConfidence.tsx [hunk 3]
```diff
@@ -35,20 +39,38 @@ interface ItemConfidenceProps {
 }
 
 const ItemConfidence: FunctionComponent<ItemConfidenceProps> = ({ confidence, variant, entityType }) => {
+  const { me: { monochrome_labels } } = useAuth();
+  const { palette: { mode } } = useTheme();
+  const theme = mode === 'dark'
```
Kedae: I don't understand your code.
```const theme = useTheme<Theme>();```
should be enough

File: opencti-platform/opencti-front/src/components/ItemStatus.jsx [hunk 4]
```diff
@@ -36,6 +40,11 @@ const styles = () => ({
 });
 
 const ItemStatus = (props) => {
+  const { me: { monochrome_labels } } = useAuth();
+  const { palette: { mode } } = useTheme();
+  const theme = mode === 'dark'
```
Kedae: I don't understand your code.
```const theme = useTheme<Theme>();```
should be enough

File: opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectLabels.tsx [hunk 5]
```diff
@@ -0,0 +1,126 @@
...
+const StixCoreObjectLabels: React.FC<StixCoreObjectLabelsProps> = ({
+  labels,
+  onClick,
+  variant,
+  revoked,
+}) => {
+  const { t_i18n } = useFormatter();
+  const { me: { monochrome_labels } } = useAuth();
+  const { palette: { mode } } = useTheme();
+  const theme = mode === 'dark'
```
Kedae: I don't understand your code.
```const theme = useTheme<Theme>();```
should be enough

File: opencti-platform/opencti-front/src/private/components/common/stix_core_objects_or_stix_relationships/StixCoreObjectOrCoreRelationshipLabelsView.js [hunk 6]
```diff
@@ -49,6 +53,11 @@ const useStyles = makeStyles(() => ({
 }));
 
 const StixCoreObjectOrCoreRelationshipLabelsView = (props) => {
+  const { me: { monochrome_labels } } = useAuth();
+  const { palette: { mode } } = useTheme();
+  const theme = mode === 'dark'
```
Kedae: I don't understand your code.
```const theme = useTheme<Theme>();```
should be enough

File: opencti-platform/opencti-front/src/private/components/entities/organizations/OrganizationDetails.tsx [hunk 7]
```diff
@@ -4,55 +4,59 @@ import Paper from '@mui/material/Paper';
...
 
 interface OrganizationDetailsComponentProps {
   organization: OrganizationDetails_organization$data;
 }
 
 const OrganizationDetailsComponent: FunctionComponent<OrganizationDetailsComponentProps> = ({ organization }) => {
-  const classes = useStyles();
+  const { me: { monochrome_labels } } = useAuth();
+  const { palette: { mode } } = useTheme();
+  const theme = mode === 'dark'
```
Kedae: I don't understand your code.
```const theme = useTheme<Theme>();```
should be enough

File: opencti-platform/opencti-front/src/private/components/entities/organizations/OrganizationDetails.tsx [hunk 8]
```diff
@@ -4,55 +4,59 @@ import Paper from '@mui/material/Paper';
...
         {t_i18n('Details')}
       </Typography>
-      <Paper classes={{ root: classes.paper }} className={'paper-for-grid'} variant="outlined">
+      <Paper
+        className={'paper-for-grid'}
+        variant="outlined"
+        style={{
+          margin: '10px 0 0 0',
+          padding: '15px',
+          borderRadius: 6,
```
Kedae: Border radius is only 4

opencti-platform/opencti-front/src/private/components/entities/organizations/OrganizationDetails.tsx
```diff
@@ -4,55 +4,59 @@ import Paper from '@mui/material/Paper';
...
       <Typography variant="h4" gutterBottom={true}>
         {t_i18n('Details')}
       </Typography>
-      <Paper classes={{ root: classes.paper }} className={'paper-for-grid'} variant="outlined">
+      <Paper
+        className={'paper-for-grid'}
+        variant="outlined"
+        style={{
+          margin: '10px 0 0 0',
+          padding: '15px',
```
Kedae: Should be theme.spacing(2)

opencti-platform/opencti-front/src/private/components/entities/organizations/OrganizationDetails.tsx
```diff
@@ -4,55 +4,59 @@ import Paper from '@mui/material/Paper';
...
     <div style={{ height: '100%' }}>
       <Typography variant="h4" gutterBottom={true}>
         {t_i18n('Details')}
       </Typography>
-      <Paper classes={{ root: classes.paper }} className={'paper-for-grid'} variant="outlined">
+      <Paper
+        className={'paper-for-grid'}
+        variant="outlined"
+        style={{
+          margin: '10px 0 0 0',
```
Kedae: should be : ```marginTop: theme.spacing(1)```

opencti-platform/opencti-front/src/components/ItemConfidence.tsx
```diff
@@ -35,20 +39,38 @@ interface ItemConfidenceProps {
...
-      <Chip
-        classes={{ root: style, label: classes.label }}
-        style={{
-          color: confidenceLevel.color,
-          borderColor: confidenceLevel.color,
-          backgroundColor: hexToRGB(confidenceLevel.color),
-        }}
-        label={confidenceLevel.label}
-      />
+      {monochrome_labels
```
Kedae: Only the style change, might be better to add the ternary inside the style

opencti-platform/opencti-front/src/components/ItemLikelihood.jsx
```diff
@@ -64,46 +66,21 @@ const ItemLikelihood = ({ likelihood, classes, variant, t, theme }) => {
...
-  if (likelihood <= 100) {
-    return (
-      <Chip
-        classes={{ root: style }}
-        style={inlineStyles.green}
-        label={`${likelihood} / 100`}
-      />
-    );
-  }
+  let chipStyle;
```
Kedae: Better way of doing this is : 
```let chipStyle = inlineStyles.white;```

To avoid a useless else.

File: opencti-platform/opencti-front/src/components/ItemColor.ts (hunk 1 above)
Comment (Bonsai8863) [reply opencti-platform/opencti-front/src/components/ItemColor.ts:12]: This function determines if the platform is in dark mode and using monochromatic labels using hooks, whereas the one in utils/Colors needs those passed in as parameters. I noticed there were few to no places where the utils/Colors' itemColors call was passed the `dark` parameter.

In my opinion, it would be better to leave the determination of whether the platform is in dark mode and monochromatic labels in one place, rather than relying on devs to properly pass these as props to the itemColors function, hence why I made a new one that grabs that info using hooks. Unfortunately this means it can only be called in functional components and not the ubiquitous remaining class components.

Comment (Kedae) [reply opencti-platform/opencti-front/src/components/ItemColor.ts:12]: Ok, I missed your comment on utils/Colors. Could you add a bigger "deprecated" warning on utils/Colors ? Thx :)

File: opencti-platform/opencti-front/src/components/ItemAccountStatus.tsx (hunk 2 above)
Comment (Bonsai8863) [reply opencti-platform/opencti-front/src/components/ItemAccountStatus.tsx]: I had issues in other places where `useTheme` gave me TypeScript errors, but `ThemeDark` and `ThemeLight` didn't. I'll double check these calls to see if it still throws those errors for me and change it to your simplified version if not.

Comment (Kedae) [reply opencti-platform/opencti-front/src/components/ItemAccountStatus.tsx]: In that case I think you just missed the < Theme > from components/Theme.d.ts in the useTheme that handles typescript correctly :)

File: opencti-platform/opencti-front/src/private/components/entities/organizations/OrganizationDetails.tsx (hunk 10 above)
Comment (Bonsai8863) [reply opencti-platform/opencti-front/src/private/components/entities/organizations/OrganizationDetails.tsx]: Should it be 1 or 2? https://github.com/OpenCTI-Platform/opencti/pull/8239#discussion_r1773373142

Comment (Kedae) [reply opencti-platform/opencti-front/src/private/components/entities/organizations/OrganizationDetails.tsx]: Starting now we want to rationalize the spacing with multiple of 8
So theme.spacing(1) = 8
theme.spacing(2) = 16
....


In your case 10 should become 8 => theme.spacing(1)
15 should become 16 => theme.spacing(2)

Mostly just apply the one that seems the more "logical"

File: opencti-platform/opencti-front/src/components/ItemAccountStatus.tsx [hunk 13]
```diff
@@ -80,7 +85,12 @@ const ItemAccountStatus: FunctionComponent<ItemAccountStatusProps> = ({
       classes={{ root: style }}
       variant={variant}
       label={label}
-      style={classStyle}
+      style={{
+        ...classStyle,
+        color: theme.palette.chip.main,
+        borderColor: monochrome_labels ? theme.palette.background.accent : classStyle.borderColor,
```
Kedae, line 91: Just having a thought right now : wouldn't it be better to have a attribute in the chip if we want the monochrome or the normal behavior and have the theme.palette.... inside all chips directly ?

Bonsai8863: Looks like we're using the MUI chip, so we could create a wrapper for it to determine monochromatic label status and apply the appropriate colors/styling.

Ideally, at least in my opinion, we'd extend the Theme interface and then be able to create a custom theme like light mode or dark mode then be able to switch to a monochromatic theme instead of all these ternary checks. At the moment, I think the platform is currently set to only handle dark or light mode themes.

Kedae: That actually is a good idea, why don't we chose that path ?

Bonsai8863: I talked to @Jipegien about it over email/conversation and he said to wait for feedback before implementing this expanded theme approach.

Bonsai8863:
Opening new PR for customizable themes concept instead of this one.

PR #8244, opened 2024-09-04
[frontend] Fix: query background task in Data>Import also delete files under entities (#3016)

Archidoit:
### Proposed changes
Fix an issue on background task in Import > Data > Uploaded files

Reproductible steps:
- Do a select all the files of this screen via the top checkbox.
- Delete them via the toolbar.
- All the files (including those contained in entities) and the workbenches are deleted.
- We should only delete global files in this screen.

Explanations: the filters given to the tool bar are cleaned according to the entity type (InternalFile in this case) filterable attributes...

### Related issues


opencti-platform/opencti-front/src/private/components/data/import/ImportFilesContent.tsx
```diff
@@ -179,7 +179,7 @@ const ImportFilesContent = () => {
         mode: 'or',
       },
       {
-        key: 'entity_id',
+        key: 'metaData.entity_id',
```
SouadHadjiat, line 182: I don't understand how we could delete all uploaded files (not only global) since we have also this filter in toolbarFilters : 
`{
        key: 'file_id',
        values: ['import/global'],
        operator: 'starts_with',
      }`
Does this filter work ?

Archidoit:
Deprogrammed, will be fixed with new import screens

PR #8254, opened 2024-09-04
[backend] Fix meEdit & authorized authorities update

Archidoit:


opencti-platform/opencti-graphql/src/database/middleware.js
```diff
@@ -1839,7 +1839,7 @@ export const updateAttributeMetaResolved = async (context, user, initial, inputs
     return { element: initial };
   }
   // Check user access update
-  const manageAccessUpdate = updates.some((e) => e.key === 'authorized_members');
+  const manageAccessUpdate = updates.some((e) => e.key === 'authorized_members' || e.key === 'authorized_authorities');
```
SouadHadjiat: is it enough to add this test ? validateUserAccessOperation will only test if you have the right capa for changing authorized members and check current authorized members, but if there are no authorized members on the entity, it would grant you access. How would it keep from changing authorized_authorities when there are no authorized members ?

opencti-platform/opencti-graphql/src/domain/user.js
```diff
@@ -792,6 +809,14 @@ export const meEditField = async (context, user, userId, inputs, password = null
   if (user.external && PROTECTED_EXTERNAL_ATTRIBUTES.includes(key)) {
     throw ForbiddenAccess();
   }
+  // Check the user has the capability to edit some fields
+  const capabilities = user.capabilities.map((c) => c.name);
+  if (
+    !capabilities.some((c) => USER_EDITION_CAPABILITIES.includes(c))
```
SouadHadjiat: I don't think we should test capabilities, because this API meEdit should not be used to change other fields than the one listed, for all users.

Archidoit: Okays, so it means even admin with BYPASS won't be able to edit some user fields via a mutation on meUser?

Archidoit: Maybe it's easier sometimes for an admin that wants to edit itself to do it via meUser than via a query on a user with the user id

opencti-platform/opencti-graphql/src/database/middleware.js
Archidoit: it is done the line just below with validateUserAccessOperation (if the user can manage the access, he will be able to edit the field as he wants)

Archidoit: to modify authorized_authorities --> have the capa SET_ACCESS ??
Is it the case only for organizations? or all the entities (ex : notif etc) ?

opencti-platform/opencti-graphql/src/utils/access.ts
```diff
@@ -545,11 +546,20 @@ export const getUserAccessRight = (user: AuthUser, element: any) => {
...
     return true;
   }
-  if (isStixObject(element.entity_type) && operation === 'manage-access' && !isUserHasCapability(user, KNOWLEDGE_KNUPDATE_KNMANAGEAUTHMEMBERS)) {
+  if (isStixObject(element.entity_type)
+    && (operation === 'manage-access' || operation === 'manage-authorities-access')
+    && !isUserHasCapability(user, KNOWLEDGE_KNUPDATE_KNMANAGEAUTHMEMBERS)
+  ) {
+    return false;
+  }
+  if (isStixObject(element.entity_type)
```
SouadHadjiat: I wonder if we need this condition isStixObject

opencti-platform/opencti-graphql/src/utils/access.ts
```diff
@@ -545,11 +546,20 @@ export const getUserAccessRight = (user: AuthUser, element: any) => {
...
 
 // ensure that user can access the element (operation: edit / delete / manage-access)
-export const validateUserAccessOperation = (user: AuthUser, element: any, operation: 'edit' | 'delete' | 'manage-access') => {
+export const validateUserAccessOperation = (user: AuthUser, element: any, operation: 'edit' | 'delete' | 'manage-access' | 'manage-authorities-access') => {
   if (isInternalObject(element.entity_type) && isUserHasCapability(user, SETTINGS_SET_ACCESSES)) {
     return true;
   }
-  if (isStixObject(element.entity_type) && operation === 'manage-access' && !isUserHasCapability(user, KNOWLEDGE_KNUPDATE_KNMANAGEAUTHMEMBERS)) {
+  if (isStixObject(element.entity_type)
+    && (operation === 'manage-access' || operation === 'manage-authorities-access')
```
SouadHadjiat: why add 'manage-authorities-access' ?

opencti-platform/opencti-graphql/src/domain/user.js
```diff
@@ -781,6 +781,25 @@ export const addBookmark = async (context, user, id, type) => {
 
 const PROTECTED_USER_ATTRIBUTES = ['api_token', 'external'];
 const PROTECTED_EXTERNAL_ATTRIBUTES = ['user_email', 'user_name'];
+const ME_USER_MODIFIABLE_ATTRIBUTES = [
+  'api_token',
+  'external',
```
SouadHadjiat: api_token & external are protected and can't be modified by this query, so we shouldn't have them here

SouadHadjiat (approved):

PR #8266, opened 2024-09-05
[frontend] New import screens under FF (#3016)

Archidoit:
### Proposed changes
Put new Data>Import screens under feature glag IMPORT_NEW_SCREENS

### Related issues


aHenryJard (approved):
Tested locally with and without FF, all good to me.

Issue #8249, opened 2024-09-04
Update Observable header has wrong color


labo-flg:
## Description

The header colors of the edition drawer are not consistent (bg color, font color) with other drawers (see in light mode).

Good one (indicator)
![image](https://github.com/user-attachments/assets/ce2a23ea-5bc3-4da2-bdbe-1c7e6bc901f8)

bad one (artifact/observables)
![image](https://github.com/user-attachments/assets/820acd05-4ebb-4766-823e-87601d2bf594)

## Environment

6.2.18 (testing)

PR #8277, opened 2024-09-07
[frontend] fixed header colors (#8249)

stefan1anuby:
After fix:

image
image

### Related issues

part of #8249 

### Checklist

### Further comments

From what I can see, the problem is not limited to Light mode. Here are a few examples of drawers that have the same issue (though there are more):

- Observations -> Observables -> Create
- Observations -> Artifacts -> Create
- Events -> Sightings -> Sighting -> Edit
- Data -> Relationships -> Edit

Should I apply the same quick fix to each, or would it be better to refactor them to use the Drawer component? It looks like the 'correct' drawers use the Drawer component internally, but that would require a major refactor and take more time.

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/observations/stix_cyber_observables/StixCyberObservableEditionContainer.jsx
+++ b/opencti-platform/opencti-front/src/private/components/observations/stix_cyber_observables/StixCyberObservableEditionContainer.jsx
@@ -13,7 +13,7 @@ import { useIsEnforceReference } from '../../../../utils/hooks/useEntitySettings
 // Do not use it for new code.
 const useStyles = makeStyles((theme) => ({
   header: {
-    backgroundColor: theme.palette.background.nav,
+    backgroundColor: theme.palette.background.default,
     padding: '20px 20px 20px 60px',
   },
   closeButton: {
@@ -49,7 +49,7 @@ const StixCyberObservableEditionContainer = (props) => {
         >
           <Close fontSize="small" color="primary" />
         </IconButton>
-        <Typography variant="h6" classes={{ root: classes.title }}>
+        <Typography variant="subtitle2" classes={{ root: classes.title }}>
           {t_i18n('Update an observable')}
         </Typography>
         <SubscriptionAvatars context={editContext} />


opencti-platform/opencti-front/src/private/components/observations/stix_cyber_observables/StixCyberObservableEditionContainer.jsx
```diff
@@ -49,7 +49,7 @@ const StixCyberObservableEditionContainer = (props) => {
         >
           <Close fontSize="small" color="primary" />
         </IconButton>
-        <Typography variant="h6" classes={{ root: classes.title }}>
+        <Typography variant="subtitle2" classes={{ root: classes.title }}>
```
Kedae, line 52: Ok, your fix might work but I think there is a better way : 

We do have an internal Drawer component that is using MaterialUI inside. 
I think the correct fix is to handle the usages of StixCyberObservableEditionContainer by replacing the old Drawer by the new one. (For example you can look at ReportCreation.tsx to see how we use the new Drawer)

Kedae (changes requested):

opencti-platform/opencti-front/src/private/components/observations/stix_cyber_observables/StixCyberObservableEditionContainer.jsx
```diff
@@ -1,69 +1,19 @@
...
-      <div className={classes.container}>
-        <StixCyberObservableEditionOverview
-          stixCyberObservable={stixCyberObservable}
-          enableReferences={useIsEnforceReference('Stix-Cyber-Observable')}
-          context={editContext}
-          handleClose={handleClose}
-        />
-      </div>
-    </div>
+    <StixCyberObservableEditionOverview
```
stefan1anuby: Should I add the custom Drawer here instead of in all three places?

Kedae: Well if you think this is relevant you can :)

Kedae:
Ok, I have some bad news.

While testing it I found out I gave you wrong direction => the replacement is not straight-forward for StixCyberObs*
Can you go back to your first commit so I can merge your fix ?

Sorry about that.

You can still try to improve the Drawer in a second review but if you take a look at SCO Detail page or even some KnowledgeGraphs (Report for example) you will see that it's not displaying the way we want currently

stefan1anuby:
I recently reached the same conclusion during some unrelated testing. I could have done more testing earlier to catch the issue sooner, or at least addressed it right away when I discovered the problem.

Kedae (approved):

Issue #8265, opened 2024-09-05
Worbench creation pop-up freezes if workbench of same name already exists


Archidoit:
## Description
Can't create workbench if the user has already try to create a workbench with an already-used name.

## Environment

1. OS (where OpenCTI server runs): { e.g. Mac OS 10, Windows 10, Ubuntu 16.4, etc. }
2. OpenCTI version: { e.g. OpenCTI 1.0.2 }
3. OpenCTI client: { e.g. frontend or python }
4. Other environment details:

## Reproducible Steps
- Go to an entity overview
- Create a workbench
- Create a second workbench with the same name
- An error occurs, which is normal  : 

![image](https://github.com/user-attachments/assets/5ef6dab5-ca8e-4e00-90bc-3550135f8239)

But then if you try to create another workbench with a different name, it's not possible. The workbench creation pop-up is still grey and disabled : 
![image](https://github.com/user-attachments/assets/f8749ca7-3a4e-4d76-91df-9f6cb90326e6)

PR #8285, opened 2024-09-08
[frontend] fixed workbench creation pop-up in case of error (#8265)

stefan1anuby:
### Proposed changes

* When trying to create a workbench with the same name, it currently shows a generic error:
image
If I hadn't set that generic error, the message would have been:
"RRNLRequestError: Relay request for WorkbenchFileCreatorMutation failed for the following reason: 1. A file already exists with this name. uploadPending(file: $file, labels: $labels, errorOnExistin ^^^"
Would this be too confusing for users?
* The pop-up no longer freezes after that workflow: 
image

### Related issues

part of #8265 

### Checklist

### Further comments

Should I implement form validation to check for duplicate workbench names when the "Create" button is pressed? Doing so would allow us to display a more specific error message, such as "A file with this name already exists."

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/common/files/workbench/WorkbenchFileCreator.tsx
+++ b/opencti-platform/opencti-front/src/private/components/common/files/workbench/WorkbenchFileCreator.tsx
@@ -105,6 +105,20 @@ const WorkbenchFileCreator: FunctionComponent<WorkbenchFileCreatorProps> = ({
     if (!name.endsWith('.json')) {
       name += '.json';
     }
+
+    const handleCompleted = () => {
+      setSubmitting(false);
+      resetForm();
+      handleCloseCreate();
+      onCompleted?.();
+    };
+
+    const handleError = () => {
+      setSubmitting(false);
+      resetForm();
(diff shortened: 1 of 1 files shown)


opencti-platform/opencti-front/src/private/components/common/files/workbench/WorkbenchFileCreator.tsx
```diff
@@ -130,6 +134,11 @@ const WorkbenchFileCreator: FunctionComponent<WorkbenchFileCreatorProps> = ({
               handleCloseCreate();
               onCompleted?.();
             },
+            onError: () => {
+              setSubmitting(false);
```
Kedae: As this code is shared between multiple places (onCompleted, onError, twice). I suggest to extract it into a "handleCompleted" or something like that method.

Kedae (changes requested):

Kedae (approved):

Kedae:
@stefan1anuby You need to sign your commits, can you check GCP to sign your commits ?

Issue #8268, opened 2024-09-05
[Dashboard] Exporting a dashboard with a “.” in the name doesn't add the “.json”.


Lhorus6:
## Description

The file obtained when exporting a dashboard containing a dot in the name generates a file that cannot be imported as it does not have the “.json” extension.

## Environment

OCTI 6.2.18

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Create a dashboard named “Beautiful dashboard filigran.io”.
2. Export it
3. Check the resulting file

You can also simply export this dashboard : https://filigran.octi.filigran.io/dashboard/workspaces/dashboards/428866e8-36a8-480f-8c26-b0eda406f341

ValentinBouzinFiligran:
![Image](https://github.com/user-attachments/assets/5a7c57e1-2f73-41a1-a9a2-1b6c915006cc)

https://github.com/OpenCTI-Platform/opencti/pull/8746

PR #8287, opened 2024-09-08
[frontend] replaced . with [dot] in filename when exporting dashboard…

stefan1anuby:
The fix that I came up with is to replace "." with "[dot]"
image

### Related issues

* part of #8268


labo-flg:
Thanks for your contribution

It looks like the issue #8268 has been fixed in latest version (tested on 6.4.4).
I'm closing this PR but feel free to reopen it if needed.

PR #7753, opened 2024-07-18
Updated Titles to Functional Component Pages Only

dotdev8:
### Proposed changes

* Updated Titles to Functional Component Pages

### Related issues

* Adding Titles to the class component pages causes the page to throw an error.
* On pages where class components were converted to functional components, the title disappears on refresh. Assuming this may have something to do with the saveView() function. 

### Checklist

- [ x] I consider the submitted work as finished
- [ x] I tested the code for its functionality
- [ ] I wrote test cases for the relevant uses case (coverage and e2e)
- [ ] I added/update the relevant documentation (either on github or on notion)
- [ ] Where necessary I refactored code to improve the overall quality

### Further comments
The pages described in the Related Issues above have not been altered. Only functional component pages where the title persists have been updated. The other pages will be updated in a later PR once the issues have been identified and resolved.

PR #7374, opened 2024-06-13
[frontend] Initial approach to dynamic update of Title and Language for pages

ParamConstructor:
### Proposed changes
This PR is to address Accessibility Issues of  Issues (#6606 and #6607)
* [Change page title when navigating across the platform #6606](https://github.com/OpenCTI-Platform/opencti/issues/6606)
* [Change <html lang="XX"> depending on the selected language #6607](https://github.com/OpenCTI-Platform/opencti/issues/6607)

It has been applied only to the Root (for Language), Dashboard page, and the pages under Analysis (i.e. Reports, Groupings, Notes, Malware Analysis, and External References) to demonstrate the approach. If accepted, a future PR can be used to update the other pages using the useHeader component.

If the user's profile has selected "automatic", the page language will be determined by the navigator/browser language setting (i.e. navigator.language ). 

### Update: June 20, 2024

Uses a hook, versus the React Helmet module in initial commits to address @richard-julien comment/request to simplify the approach.

### Related issues

None

### Checklist

- [X] I consider the submitted work as finished
- [X] I tested the code for its functionality
- [X] (N/A) I wrote test cases for the relevant uses case (coverage and e2e)
- [ ] I added/update the relevant documentation (either on github or on notion)
- [X] Where necessary I refactored code to improve the overall quality

PR #8300, opened 2024-09-10
[frontend] Add page titles to Data pages

VerboseCat:
### Proposed changes
This PR is a follow-up to the the PR which introduced custom page titles (https://github.com/OpenCTI-Platform/opencti/pull/7374). This PR converts several 'Data' components from classes to functional and adds page titles. These components were not included in the original PR because they were class components and could not use the page title hook.

**Impacted Pages**
* 

### Related issues

* https://github.com/OpenCTI-Platform/opencti/pull/7374

### Checklist

### Further comments
This PR has a known issue, which has also been reported in a comment on (https://github.com/OpenCTI-Platform/opencti/pull/7753). Every time a page renders, both the the custom page title setting function `useConnectedDocumentModifier.setTitle` and the default page title setting function `useDocumentLangModifier` run and set the title. In some instances, the default page title is set after the custom page title is set. This results in the default page title being displayed. I was unable to find a cause or pattern for which components exhibit this bug. 

**The following components from this PR are impacted:**
* Entities / Sectors (entities/Sectors.jsx)
* Data / Ingestion / Connectors (data/Connectors.jsx)
* Data / Data sharing / Live streams (data/Streams.tsx)


SarahBocognano (approved):
✅

PR #8309, opened 2024-09-11
[frontend] Observables exports list of containers should be independant (#8241)

### LINKED ISSUE #8241: There are 'Exports lists' in another report (OBSERVABLES in report)
OpenCTI version : 6.2.12

There are 'Export lists' generated by other reports, so it's hard to tell which report generated the list.
image

This does not happen on the ENTITIES tab, but only on the OBSERVABLES tab.

Can you please confirm?

Archidoit:
### Proposed changes
In a container > observable list > export panel : the list of export files should only shows the exports generated from the  container (and not other containers).

### Related issues
https://github.com/OpenCTI-Platform/opencti/issues/8241


SouadHadjiat (approved):
tested locally

PR #8338, opened 2024-09-13
[frontend] Authorized members default value should not contain me user (#8311)

### LINKED ISSUE #8311: Authorized Members - Default value
## Description

Currently, when editing an entity in customization to apply a default value on Authorized members, my user is automatically added to authorized members with the right can manage. This is wrong, user should not be added by default as authorized member.

## Environment

Demo

## Reproducible Steps

1. Go in Setting/customization
2. Edit Feedback
3. Click on Authorized members
4. 

## Expected Output

The current user should not be automatically added as an authorized members (and even more, should not be as can manage)

## Actual Output

Current user is added as an authorized member with the capa "can manage".

Archidoit:
### Proposed changes
Authorized members default values should not add current user with manage access by default.
Authorized members default values can be found in : 
- Security > Customization > Feedback > Authorized members > Activate access restriction
- or in : a feedback I can manage > button manage access > Activate access restriction

### Related issues
https://github.com/OpenCTI-Platform/opencti/issues/8311

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/common/form/AuthorizedMembersField.tsx
+++ b/opencti-platform/opencti-front/src/private/components/common/form/AuthorizedMembersField.tsx
@@ -17,6 +17,7 @@ import SelectField from '../../../../components/fields/SelectField';
 import { useFormatter } from '../../../../components/i18n';
 import { AccessRight, ALL_MEMBERS_AUTHORIZED_CONFIG, AuthorizedMemberOption, Creator, CREATOR_AUTHORIZED_CONFIG } from '../../../../utils/authorizedMembers';
 import SwitchField from '../../../../components/fields/SwitchField';
+import useAuth from '../../../../utils/hooks/useAuth';
 
 /**
  * Returns true if the authorized member option is generic.
@@ -41,6 +42,7 @@ interface AuthorizedMembersFieldProps
   showAllMembersLine?: boolean;
   showCreatorLine?: boolean;
   canDeactivate?: boolean;
+  addMeUserWithAdminRights?: boolean;
 }
 
 // Type of data for internal form, not exposed to others.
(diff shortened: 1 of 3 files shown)


labo-flg:
When in the settings it makes 100% sense to not have the current user as mandatory.
What about in the feedback ? (or any other entities on which we will enable restrictions)
If you enable from there, and do not enable yourself as manager, you lose access to the object. 
If it's a human error you cannot recover from that, feedback is locked forever.
@nino-filigran WDYT ?

marieflorescontact (commented):
This change removes the take over we wanted on case incident:
On restriction activation, we wanted the current user to be the first authorized member with can manage access right without any extra action. The second if the creator is someone else. [Takeover]
I don't know if this is ok?

Archidoit:
> When in the settings it makes 100% sense to not have the current user as mandatory. What about in the feedback ? (or any other entities on which we will enable restrictions) If you enable from there, and do not enable yourself as manager, you lose access to the object. If it's a human error you cannot recover from that, feedback is locked forever. @nino-filigran WDYT ?

@labo-flg Nino told me to remove me user for entity authorized members edition too... And it's not possible to lock an entity forever since if there are no manager, the edition raises an error and is not done.

Archidoit:
> This change removes the take over we wanted on case incident: On restriction activation, we wanted the current user to be the first authorized member with can manage access right without any extra action. The second if the creator is someone else. [Takeover] I don't know if this is ok?

@nino-filigran so we want to remove me user only for feedbacks and not for cases ? or remove all the time?

nino-filigran:
Ok, so it seems I gave incorrect direction @Archidoit @marieflorescontact & @labo-flg. My apologies for the confusion I caused. 
We need: 
- in setting, to not have the current user set as default authorized member. 
- in whatever entity (feedback, case or any future entity), have the current user **who enables the authorized member to be in the authorized member with the capa "can manage"**, the current behavior basically.

Later (in 6.5) we'll have a slot again to work on this feature. Therefore, if we need to change this logic, we'll do it then.

Archidoit:
@nino-filigran Ok thanks for the clarification, so I'm gonna remove MeUser in authorized members only in the settings.

Archidoit:
@labo-flg @marieflorescontact changes done :)

marieflorescontact (approved):

Issue #7906, opened 2024-08-01
Update views to new DataTable


labo-flg:
## Description

Gradual rollout of the new DataTable component for all the list views in the app.

PR #8367, opened 2024-09-16
[frontend] Multiple fixes on DataTables (#7906)

Kedae:


opencti-platform/opencti-front/src/components/dataGrid/components/DataTableHeaders.tsx
```diff
@@ -175,7 +175,7 @@ const DataTableHeaders: FunctionComponent<DataTableHeadersProps> = ({
                 orderAsc={!!orderAsc}
               />
             ))}
-          <span style={{ width: 42 }} />
+          <span style={{ width: 'calc( 42px + 12px )' }} />
```
labo-flg, line 178: it's for decomposing so the code is understandable ?
otherwise `width: '56px'` would be enough I guess :)

labo-flg (approved):

labo-flg (changes requested):
header margins are not good

opencti-platform/opencti-front/src/components/dataGrid/components/DataTableHeaders.tsx
labo-flg: this messes up with all the vertical alignments

PR #8400, opened 2024-09-18
[frontend] Bulk edition of status regression fix (#8393)

### LINKED ISSUE #8393: Bug/Regression - Bulk edition of status not possible
## Description

I'd like to change Vulnerability status en masse but I can't do it anymore (i.e. Select several Vulns > Update > Replace > ‘status’ option not available). I am not sure if we could do it in the past on Vulnerability, I imagine we could. One thing's for sure: we were able to at least do it on Reports, but that's no longer possible.

## Environment

OCTI 6.3.0

## Reproducible Steps

1. Select several Report and click on "update"

Screenshot 2024-09-18 003814

2. "Replace" then open the drop down menu

image

We no longer have "status"

Archidoit:
### Proposed changes
Fix a regression: it was not possible anymore to bulk edit Status via the toolbar

### Related issues
https://github.com/OpenCTI-Platform/opencti/issues/8393

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/data/DataTableToolBar.jsx
+++ b/opencti-platform/opencti-front/src/private/components/data/DataTableToolBar.jsx
@@ -720,6 +720,7 @@ class DataTableToolBar extends Component {
     const { t } = this.props;
     const { actionsInputs } = this.state;
     const disabled = R.isNil(actionsInputs[i]?.type) || R.isEmpty(actionsInputs[i]?.type);
+    const entityTypesWithoutStatus = ['Stix-Core-Object', 'Stix-Domain-Object', 'Stix-Cyber-Observable', 'Artifact', 'ExternalReference'];
     let options = [];
     if (actionsInputs[i]?.type === 'ADD') {
       options = [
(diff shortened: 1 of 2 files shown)


SouadHadjiat:
is it possible to add a non regression test for this usecase ?

opencti-platform/opencti-front/src/private/components/data/DataTableToolBar.jsx
```diff
@@ -932,9 +937,15 @@ class DataTableToolBar extends Component {
...
 
-  searchStatuses(i, event, newValue) {
+  searchStatuses(i, selectedTypes, event, newValue) {
     if (!event) return;
     const { actionsInputs } = this.state;
+    let selectedType;
+    if (selectedTypes.length === 1) {
+      [selectedType] = selectedTypes;
+    } else {
+      throw Error('It is not possible to bulk edit statuses if more than one entity type is selected.');
```
labo-flg, line 947: that would show as an error in the UI ?

labo-flg (dismissed):

labo-flg:
I can now update vulnerabilities status in mass ops from Vuln list, but not from Entities view. Is it expected ?

labo-flg:
> I can now update vulnerabilities status in mass ops from Vuln list, but not from Entities view. Is it expected ?

status is open vocab, might be different from entities to entities, so definitely not an issue.

labo-flg (approved):

Archidoit:
@labo-flg Yes it's expected because entities of different type have different possible values for 'status'

opencti-platform/opencti-front/src/private/components/data/DataTableToolBar.jsx
Archidoit: theoretically, it's not possible to have this error since we check in the code that only one entity type is selected. It's to prevent bad changing of code in the future if someone don't exactly know what he is doing ;)

Issue #8312, opened 2024-09-11
Ability to define and implement “Exclusion Lists” - POC 


romain-filigran:

nino-filigran:
This issue is closed but not finished in the current release. Closed in favor of: https://github.com/OpenCTI-Platform/opencti/issues/8941 which is the issue to follow with the same scope.

PR #8449, opened 2024-09-23
[backend] observables values implemented (#8312)

ValentinBouzinFiligran:
### Proposed changes

* observables values key created
* pattern extraction function created
* observables values filled on creation
* observables values filled on pattern edition or indicator edition when empty
* graphql schema updated to export this new key

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/8312

### Related PR 

* https://github.com/OpenCTI-Platform/client-python/pull/745

### Checklist

### Further comments

### DIFF
--- a/opencti-platform/opencti-graphql/src/modules/indicator/indicator-domain.ts
+++ b/opencti-platform/opencti-graphql/src/modules/indicator/indicator-domain.ts
@@ -212,6 +212,16 @@ export const promoteIndicatorToObservables = async (context: AuthContext, user: 
 
 export const getObservableValuesFromPattern = (pattern: string) => extractObservablesFromIndicatorPattern(pattern);
 
+const getFormattedPattern = async (context: AuthContext, user: AuthUser, pattern_type: string, pattern: string) => {
+  const patternType = pattern_type.toLowerCase();
+  const formattedPattern = cleanupIndicatorPattern(patternType, pattern);
+  const check = await checkIndicatorSyntax(context, user, patternType, formattedPattern);
+  if (check === false) {
+    throw FunctionalError(`Indicator of type ${pattern_type} is not correctly formatted.`);
+  }
+  return formattedPattern;
+};
+
 export const addIndicator = async (context: AuthContext, user: AuthUser, indicator: IndicatorAddInput) => {
   let observableType: string = isEmptyField(indicator.x_opencti_main_observable_type) ? 'Unknown' : indicator.x_opencti_main_observable_type as string;
   if (observableType === 'File') {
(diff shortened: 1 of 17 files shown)


opencti-platform/opencti-graphql/src/modules/indicator/indicator-domain.ts
```diff
@@ -305,6 +312,18 @@ export const indicatorEditField = async (context: AuthContext, user: AuthUser, i
   if (!indicator) {
     throw FunctionalError('Cannot edit the field, Indicator cannot be found.');
   }
+
+  const foundPattern = finalInput.find((item) => item.key === 'pattern');
+  if (!indicator.x_opencti_observables_values || foundPattern) {
```
SouadHadjiat: could we extract this code block in a separate function? And maybe mutualize it with the same logic we have in addIndicator?

opencti-platform/opencti-graphql/src/modules/indicator/indicator-domain.ts
```diff
@@ -231,6 +233,10 @@ export const addIndicator = async (context: AuthContext, user: AuthUser, indicat
   // find default decay rule (even if decay is not activated, it is used to compute default validFrom and validUntil)
   const decayRule = await findDecayRuleForIndicator(context, observableType);
   const { validFrom, validUntil, revoked, validPeriod } = await computeValidPeriod(indicator, decayRule.decay_lifetime);
+  const extractedObservableValues = getObservableValuesFromPattern(formattedPattern);
+  if (!extractedObservableValues.length) {
+    throw FunctionalError(`Indicator of type ${indicator.pattern_type} is not correctly formatted.`);
```
SouadHadjiat: really ? do we always extract observable values from indicator pattern ? if we don't find observable values, we don't create the indicator ? extraction works only with stix pattern, what about other pattern types ?

opencti-platform/opencti-graphql/src/modules/indicator/indicator-domain.ts
SouadHadjiat: I also suggest to move this part after the validation check about valid_until, since if we don't pass this validation check, there is no need for this part.

opencti-platform/opencti-graphql/src/modules/indicator/indicator-domain.ts
SouadHadjiat: also I think it would be best to make sure we don't extract multiple times these observable values, like what we do in createObservablesFromIndicator.

Issue #8453, opened 2024-09-23
Support of "has" relationship between a System and a Vulnerability


romain-filigran:
## Use case

To be able to correctly model Vulnerabilities associated to a System entity in OpenCTI, we need to introduce a "has" relationship between a System and a Vulnerability: 

System --> has --> Vulnerability

## Current Workaround

Use of generic "related-to" relation

## Proposed Solution

Introduce a new relation "has" between a System and a Vulnerability

PR #8458, opened 2024-09-23
[frontend + backend] added vulnerabilitties tab and 'has' relationship

stefan1anuby:
### Proposed changes

Added 'has' relationship between system and vulnerability:

image

Added 'Vulnerabilities' tab in the Knowledge sections of the Systems:

image

### Related issues

* #8453 
* #8498


SarahBocognano:
Just checked your PR. The relationship creation is working well via UI and API, but there is a little miss somewhere.
We want to be able to add the relationship in related entities too.
Capture d'écran 2024-10-02 102942

In SystemKnowledge.jsx, add the "has" relationshipTypes in the "related" route and it will work perfectly ! :)

stefan1anuby:
Would it be even better if I would create a separate route in SystemKnowledge for vulnerabilities? That also means solving #8498.

SarahBocognano (approved):

Issue #8101, opened 2024-08-19
Organisation Admins: multiple issues 


nino-filigran:
## Description

We already have one bug for Org admin: https://github.com/OpenCTI-Platform/opencti/issues/8013 => fixed
This issue aism to gather all bugs existing for ORg admins and to tackle them in one go, so that we know we do not have other issues with them. 

## Environment

Testing 

### Issue 1: Cannot Remove orga of a user that I'm admin of 
1. Log in a with a user that is Admin Orga of 2 orga & have at least one user part of these 2 orgas
2. Go to security/users
3. Select the mentioned user above & click on edit 
4. Try to remove one of the orgas you're admin of
5. you get an error: You are not allowed to do this.

### Issue 2: Cannot Add an orga to user that already has an orga even though I'm Admin org of 2 orgas
1. Log in with a user that is Admin orga of 2 orga and have a least one user pertaining to only 1 of the orgs
2. Go to security/users
3. Select the mentioned user  & click on edit
4. On the field "organisation", click on it to add the other orga that your user is not part of 
5. The other orga is not even retrieved in the dropdown

### Issue 3: Cannot Edit any property of a user when I'm Admin Orga of a user => FIXED
https://github.com/OpenCTI-Platform/opencti/pull/8030
Steps to create the smallest reproducible scenario:
1. Log in as Admin Org of a user
7. Go to security/users
8. Tap on the CTA to edit a user
9. The drawer opens 
10. Try to update any field (description, language...)
11. close the drawer
12. an error message appears "Cannot read properties of undefined (reading 'find')"

### Issue 4: Cannot Delete a user that is part of my org when I'm admin of a user => FIXED
https://github.com/OpenCTI-Platform/opencti/pull/8030
Steps to create the smallest reproducible scenario:
1. Log in as Admin Org of a user
6. Go to security/users
7. Tap on the CTA to edit a user
8. The drawer opens 
9. Try to delete the user 
10. an error message appears "Cannot read properties of undefined (reading 'find')"

### Issue 1: Cannot Remove orga of a user that I'm admin of 
You get an error: You are not allowed to do this.

### Issue 2: Cannot Add an orga to the user that already has an orga even though I'm Admin org of 2 orgas
The other orga is not even retrieved in the dropdown

### Issue 3: Cannot Edit any property of a user when I'm Admin Orga of a user
An error message appears "Cannot read properties of undefined (reading 'find')"

[...]

PR #8459, opened 2024-09-23
[backend/frontend] fix multiple issues for Organisation Admins (#8101)

marieflorescontact:
### Proposed changes
*  2 create (relations)
*  4 updates (Identity)
*  2 delete

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/8101

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/domain/user.js
+++ b/opencti-platform/opencti-graphql/src/domain/user.js
@@ -418,9 +418,16 @@ export const roleEditContext = async (context, user, roleId, input) => {
   });
 };
 
+const isUserAdministratingOrga = (user, organizationId) => {
+  return user.administrated_organizations.some(({ id }) => id === organizationId);
+};
+
 export const assignOrganizationToUser = async (context, user, userId, organizationId) => {
   if (isOnlyOrgaAdmin(user)) {
-    throw ForbiddenAccess();
+    // When user is organization admin, we make sure she is also admin of organization added
+    if (!isUserAdministratingOrga(user, organizationId)) {
+      throw ForbiddenAccess();
+    }
   }
   const targetUser = await findById(context, user, userId);
   if (!targetUser) {
@@ -1006,7 +1013,10 @@ export const userIdDeleteRelation = async (context, user, userId, toId, relation
 
 export const userDeleteOrganizationRelation = async (context, user, userId, toId) => {
   if (isOnlyOrgaAdmin(user)) {
-    throw ForbiddenAccess();
+    // When user is organization admin, we make sure she is also admin of organization removed
+    if (!isUserAdministratingOrga(user, toId)) {
+      throw ForbiddenAccess();
+    }
   }
   const targetUser = await findById(context, user, userId);
   if (!targetUser) {
--- a/opencti-platform/opencti-graphql/src/python/testing/local_synchronizer.py
+++ b/opencti-platform/opencti-graphql/src/python/testing/local_synchronizer.py
@@ -62,6 +62,9 @@ class TestLocalSynchronizer:
             logging.info("%s", f"Processing event {msg.id}")
             self.count_number += 1
             data = json.loads(msg.data)
+            type = data["data"]["type"]
+            if type == "internal-relationship":
+                return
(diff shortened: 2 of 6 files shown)


opencti-platform/opencti-graphql/src/domain/user.js
```diff
@@ -419,7 +419,11 @@ export const roleEditContext = async (context, user, roleId, input) => {
 
 export const assignOrganizationToUser = async (context, user, userId, organizationId) => {
   if (isOnlyOrgaAdmin(user)) {
-    throw ForbiddenAccess();
+    // When user is organization admin, we make sure she is also admin of organization added
+    const myAdministratedOrganizationsIds = user.administrated_organizations.map(({ id }) => id);
```
SouadHadjiat:
```suggestion
    const isAdministratedOrga = user.administrated_organizations.some(({ id }) => id === organizationId);
```

opencti-platform/opencti-graphql/src/domain/user.js
```diff
@@ -420,7 +420,11 @@ export const roleEditContext = async (context, user, roleId, input) => {
 
 export const assignOrganizationToUser = async (context, user, userId, organizationId) => {
   if (isOnlyOrgaAdmin(user)) {
-    throw ForbiddenAccess();
+    // When user is organization admin, we make sure she is also admin of organization added
+    const isAdministratedOrga = user.administrated_organizations.some(({ id }) => id === organizationId);
+    if (!isAdministratedOrga) {
+      throw ForbiddenAccess();
+    }
```
lndrtrbn: Could be nice to make a function for this one as we reuse it below. If we need to make some changes in the future we will have to make them only one time

opencti-platform/opencti-graphql/src/python/testing/local_synchronizer.py
```diff
@@ -62,6 +62,9 @@ def _process_message(self, msg):
             logging.info("%s", f"Processing event {msg.id}")
             self.count_number += 1
             data = json.loads(msg.data)
+            type = data["data"]["type"]
+            if type == "internal-relationship":
+                return
```
lndrtrbn, line 67: What's the purpose of this? We want to do nothing if internal relationship ? If I understand correctly

marieflorescontact: well... this is to avoid to count "internal-relationiship" in raw stream . am I right @SouadHadjiat ?

opencti-platform/opencti-graphql/src/domain/user.js
marieflorescontact: done

opencti-platform/opencti-graphql/src/domain/user.js
```diff
@@ -418,11 +418,14 @@ export const roleEditContext = async (context, user, roleId, input) => {
   });
 };
 
+const isAdministratedOrga = (user, organizationId) => {
```
lndrtrbn: `isUserAdministratingOrga` ? I think we miss in the name the important part that we want this orga to be administrated by this specific user, not a random one

opencti-platform/opencti-graphql/src/python/testing/local_synchronizer.py
SouadHadjiat: yes, we don't want to synchronize internal relationships : we actually never create them anyway, the issue we had is that we couldn't delete them because they weren't created.

SouadHadjiat: we however need them in the stream, because we have an inferrence rule for participates-to relationships (user member of an organization, we infere the relationship with parent organizations)

lndrtrbn: Okok thanks, could you explain to me what is the role of the local synchronizer? I think it could help me understand what we do in `process_message`

For what I understand: we do not need to synchronize internal rels as there is nothing to sync because they are not created

lndrtrbn (approved):

Issue #8284, opened 2024-09-08
Protect sensitive features and configurations from modifications


Jipegien:
## Use case

Some administrative actions and change of configuration through the UI may cause data loss during ingestion, loss of data visibility for users, breaking predefined automation, etc. without sending proper warning to users taking these administrative actions.

We must implement a way to flag elements of the platform to protect them from unwanted and unpondered modifications.

Elements to protect:
- Configuration of specific users, groups, roles
- activation/deactivation of inference rules
- Main platform Organization setting
- Default marking definitions
- Enterprise Edition deactivation
- File indexing pause/reset

Defining and Using a "Protected" element can only be done with a specific Capability. This capability is **not** bypassable.

Protected Elements can clearly be identified in the UI.
- If you don't have the right capability, the Protected Element is visible but is disabled
- If you have the right capability, the Protected Element is highlighted in strong color, you can use the feature but it needs you to confirm the action (with explicit warning), and you can unprotect the element (explicit warning)

PR #8509, opened 2024-09-25
[backend/frontend] Implementation of sensitive configuration protection (#8284)

aHenryJard:
### Proposed changes

* Add a configuration in Role to know if user is allow or not to change sensitive configuration on the platform
* Add this configuration in the capability list in the UI
* Protect initial groups/roles Administrator(s) and Connector(s) from modification when user are not allow to modifiy sensitve configuration

* Under feature flag **PROTECT_SENSITIVE_CHANGES**

### Related issues

* relates to #8284, does not close it.

### Checklist


opencti-platform/opencti-graphql/src/types/user.d.ts
```diff
@@ -4,6 +4,7 @@ import type { ConfidenceLevel } from '../generated/graphql';
 
 interface UserRole extends BasicStoreIdentifier {
   name: string;
+  patate: boolean;
```
aHenryJard:
```suggestion
```

opencti-platform/opencti-graphql/src/database/data-initialization.js
```diff
@@ -247,12 +247,21 @@ const createBasicRolesAndCapabilities = async (context) => {
...
     name: ROLE_ADMINISTRATOR,
     description: 'Administrator role that bypass every capabilities',
     capabilities: [BYPASS],
-  });
-  const connectorRole = await addRole(context, SYSTEM_USER, {
+  };
+  if(isFeatureEnabled((PROTECT_SENSITIVE_CHANGES_FF))){
+    administratorRoleInput = {
+      ...administratorRoleInput,
+      is_sensitive_changes_allow: false
```
SouadHadjiat: false ?

SouadHadjiat: need to check if CSM or platform use this group too

opencti-platform/opencti-front/src/private/Root.tsx
```diff
@@ -141,6 +141,7 @@ const meUserFragment = graphql`
       id
       name
     }
+    is_sensitive_changes_allow
```
aHenryJard: Need to add feature flag

opencti-platform/opencti-front/src/private/components/settings/roles/Role.tsx
```diff
@@ -61,6 +63,7 @@ const roleFragment = graphql`
       name
       description
     }
+    is_sensitive_changes_allow
```
aHenryJard: Need to add feature flag

opencti-platform/opencti-front/src/private/components/settings/roles/RoleEditionCapabilities.tsx
```diff
@@ -155,6 +207,7 @@ const RoleEditionCapabilities = createFragmentContainer(
     role: graphql`
       fragment RoleEditionCapabilities_role on Role {
         id
+        is_sensitive_changes_allow
```
aHenryJard: Need to add feature flag

opencti-platform/opencti-graphql/tests/01-unit/domain/user-test.ts
```diff
@@ -1,8 +1,9 @@
 import { describe, expect, it } from 'vitest';
 import { testContext } from '../../utils/testQuery';
-import { checkPasswordInlinePolicy } from '../../../src/domain/user';
+import {checkPasswordInlinePolicy, isSensitiveChangesAllowed} from '../../../src/domain/user';
+import {OPENCTI_ADMIN_UUID} from '../../../src/schema/general';
 
-describe('password checker', () => {
+describe.skip('password checker', () => {
```
aHenryJard: To be put back before merge.

opencti-platform/opencti-graphql/src/domain/grant.js
```diff
@@ -16,7 +18,19 @@ export const addRole = async (context, user, role) => {
     assoc('description', role.description ? role.description : ''),
     dissoc('capabilities'),
   )(role);
-  const { element, isCreation } = await createEntity(context, user, roleToCreate, ENTITY_TYPE_ROLE, { complete: true });
+
+  let completeRoleToCreate;
+  if(isFeatureEnabled(PROTECT_SENSITIVE_CHANGES_FF)) {
+      completeRoleToCreate = {
+        ...roleToCreate,
+        is_sensitive_changes_allow: role.is_sensitive_changes_allow ?? true, //default when undefined is true
```
SouadHadjiat: false by default ? otherwise all new roles created will have this option by default

opencti-platform/opencti-front/src/private/Root.tsx
SouadHadjiat: Suggestion of naming : `can_manage_sensitive_config` ?

opencti-platform/opencti-graphql/src/database/data-initialization.js
```diff
@@ -270,7 +279,16 @@ const createBasicRolesAndCapabilities = async (context) => {
       'SETTINGS_SETMARKINGS',
       'SETTINGS_SETLABELS',
     ],
-  });
+  };
+
+  if (isFeatureEnabled((PROTECT_SENSITIVE_CHANGES_FF))) {
+    connectorRoleInput = {
+      ...connectorRoleInput,
+      can_manage_sensitive_config: false
```
aHenryJard, line 312: I think it's not necessary anymore because for new role when feature flag is enabled the default is false now. But it can stay there, not a big deal.

opencti-platform/opencti-front/src/utils/hooks/useSensitiveModifications.ts
```diff
@@ -0,0 +1,63 @@
...
+   * @param role
+   */
+  const isRoleWithManageSensitiveConfig = (role: Role_role$data) => {
+    return role.can_manage_sensitive_config ?? true;
+  };
+
+  /**
+   * Return true when current user has the fake capa can_manage_sensitive_config enabled.
+   */
+  const isPlatformOrgaModificationAllowed = () => {
```
SouadHadjiat: why is it a function ? can't we just return the value directly ?
isPlatformOrgaModificationAllowed: me.can_manage_sensitive_config ?
I see that it's the same for other properties except ffenabled, we return functions instead of a result, why ?

Not blocking for this PR review I guess, maybe to be checked for chunk 2.

opencti-platform/opencti-graphql/tests/02-integration/01-database/elasticSearch-test.js
```diff
@@ -404,7 +404,7 @@ describe('Elasticsearch pagination', () => {
   it('should entity paginate everything', async () => {
     const data = await elPaginate(testContext, ADMIN_USER, READ_ENTITIES_INDICES, { first: ES_MAX_PAGINATION });
     expect(data).not.toBeNull();
-    expect(data.edges.length).toEqual(537);
+    expect(data.edges.length).toEqual(520 + TESTING_USERS.length + TESTING_ROLES.length + TESTING_GROUPS.length);
```
SouadHadjiat, line 407: why did this counter change ? we didn't add any object in the tests in this PR, right ?

aHenryJard: Actually we need to discuss something: as far as I see the "Administrator" group was not created on platform start, or not created anymore ? So we have added it back. The Administrator role war correctly created, only group is missing.

aHenryJard: Adding TESTING_USERS.length + TESTING_ROLES.length + TESTING_GROUPS.length is done to be less sensitive to RBAC changes in tests, it has been done previously on other counter that "paginate all the thing" in the platform, so we added that here too, just as "refactor"

opencti-platform/opencti-front/src/utils/hooks/useSensitiveModifications.ts
```diff
@@ -0,0 +1,63 @@
...
+    return me.can_manage_sensitive_config;
```
aHenryJard: I think we need the check on undefined/null here too (can_manage_sensitive_config= null)
```suggestion
    return me.can_manage_sensitive_config ?? true;
```

opencti-platform/opencti-graphql/tests/02-integration/01-database/elasticSearch-test.js
aHenryJard: test coverage added :)

opencti-platform/opencti-graphql/src/database/data-initialization.js
aHenryJard: done

opencti-platform/opencti-front/src/utils/hooks/useSensitiveModifications.ts
marieflorescontact: I replaced `isPlatformOrgaModificationAllowed` function by the value returned.
other hook functions have logic inside. is it a problm to return function in a hook instead of value?

PR #8548, opened 2024-09-30
[Backend] External reference fail creation as checked attribute is not the good one (#8545)

### LINKED ISSUE #8545: [live stream] external references are not synced
## Description

I am having trouble syncing external references of a report using live streams. 

I tested two cases, one with "perfect synchronization" and one without it:

**Without perfect synchronization**:
If I create a report, add an external reference, it is correctly sync with the destination.
However, if I remove the external reference in the report afterwards, it is not removed on the destination.
Same thing happends with the label added and then removed.

**With perfect synchronization**:
If I create a report, add an external reference, the external reference is never added to the destination.
However, the label is added correctly to the report and removed afterwards as well.

Is this expected? How can I sync two opencti instances and have the external references and labels of a report synced when being added and removed as well?

## Environment

1. OS (where OpenCTI server runs): MacOS, docker
2. OpenCTI version: 6.3.4
3. OpenCTI client: frontend
4. Other environment details:

## Reproducible Steps

1. Setup a live stream between two instances of opencti
2. Create a report, add external reference and a label
3. Remove the external reference and the label from the report

## Expected Output

I expect to have the external references and the labels added correctly, then removed.

## Actual Output

External reference is never removed from the report, or not even added with perfect sync.

## Screenshots (optional)

Example without perfect sync:
image

Example with perfect sync:
image

richard-julien:
See #8545

### DIFF
--- a/opencti-platform/opencti-graphql/src/schema/schema-attributes.ts
+++ b/opencti-platform/opencti-graphql/src/schema/schema-attributes.ts
@@ -337,7 +337,7 @@ export const validateDataBeforeIndexing = (element: any) => {
 
   Object.keys(element).forEach((elementKey) => {
     const input = element[elementKey];
-    const attributeSchemaDef = schemaAttributesDefinition.getAttributeByName(elementKey);
+    const attributeSchemaDef = schemaAttributesDefinition.getAttribute(element.entity_type, elementKey);
     if (!attributeSchemaDef) {
       return; // no validation to do, happens for meta fields like "_index"
     }


SouadHadjiat:
The code change looks all right, but I'm not sure it fixes the issue. 
I'm not sure how to reproduce the issue by the way.
I tried to run opencti locally, add a remote octi stream on testing (filtered on reports with a specific label I created). 
I create reports with this label on testing => it's well created on my local opencti
I add external ref on this report on testing => it's well added locally
I remove the external ref on this report => KO it's not removed from my local report, the external ref is still present. 
Should it be removed with this fix ? what did I miss to reproduce the issue ?

richard-julien:
> The code change looks all right, but I'm not sure it fixes the issue. I'm not sure how to reproduce the issue by the way. I tried to run opencti locally, add a remote octi stream on testing (filtered on reports with a specific label I created). I create reports with this label on testing => it's well created on my local opencti I add external ref on this report on testing => it's well added locally I remove the external ref on this report => KO it's not removed from my local report, the external ref is still present. Should it be removed with this fix ? what did I miss to reproduce the issue ?

Sometimes the external refs was not created at all. To have the deletion you need to activate the advanced option "full sync" on the client side

SouadHadjiat (approved):
change is good and should fix issues, but I couldn't make sure that the bug related is fixed.

SouadHadjiat:
With perfect sync advanced option, the external ref is well synchronized (added and removed). But without this option, the remove of external ref or label is not synchronized.

richard-julien:
> With perfect sync advanced option, the external ref is well synchronized (added and removed). But without this option, the remove of external ref or label is not synchronized.

Thats the expected behavior. Without full sync the refs are only cumulated (markings, externals refs, labels, ...)

Issue #3402, opened 2023-05-25
Outcome templates by entity types - POC


SamuelHassine:
## Use case

Outcome templates by entity types.

See special project.

## Proposed solution
- Be able to define a template for some containers (at least reports), that will generate automatically an "outcome" (a file based on this template) in the content section. The file should be editable (markdown) since templates never fits 100% of the cases.
- Be able to add some sort of variables/images of the container in the template, to simplify the creation of the template.
- Save & download in PDF.

nino-filigran:
To also be included: 

Refinement and enhancement of the rich text editor in content:

Be able to have suggestion "@" with auto completion to use data from the platform (threat actors, intrusion setc, etc.)
Be able to insert a widget in SVG format like in dashboard
Optional: graph?

nino-filigran:
This issue is not finished, but closed to acknowledge work done. The task replacing this one is: https://github.com/OpenCTI-Platform/opencti/issues/8940

PR #8553, opened 2024-09-30
[frontend/backend] Content from template for containers - Chunk 1 and 2 (#3402)

Archidoit:
### Proposed changes

POC for outcomes templates generation (frontend engine) - Chunk 1 and 2

- Feature flag
- Hardcoded templates in frontend
- Can create a container content based on one template
- POC generation of PDF with list of container entities / container attributes / a donut graph
- Can delete generated contents
- Some unit tests

No e2e tests as the UI will change in next chunks

### Related issues

- #3402


opencti-platform/opencti-front/package.json
```diff
@@ -100,6 +100,7 @@
     "@testing-library/jest-dom": "6.4.2",
     "@testing-library/react": "16.0.0",
     "@testing-library/user-event": "14.5.2",
+    "@types/html-to-pdfmake": "^2.4.4",
```
SouadHadjiat: shouldn't it be a fixed version ?

frapuks (approved):

opencti-platform/opencti-front/package.json
lndrtrbn: indeed, has been fixed thanks

opencti-platform/opencti-front/lang/front/fr.json
```diff
@@ -1656,6 +1659,7 @@
   "Max Confidence Level": "Niveau de confiance maximal",
   "Max confidence level": "Niveau de confiance maximal",
   "Max Confidence Level:": "Niveau de confiance maximum :",
+  "Max content level markings": "Marques de niveau de contenu maximum",
```
aHenryJard:
```suggestion
  "Max content level markings": "Marquage maximum du contenu",
```
it's all "marquage" in french texts

lndrtrbn: indeed, fixed in several translations, thanks

opencti-platform/opencti-front/src/utils/outcome_template/engine/apexchartUtils.ts
```diff
@@ -0,0 +1,31 @@
...
+  const options = chartOptions;
+  const canvas = document.createElement('canvas');
+  canvas.width = 800;
+  canvas.height = 400;
+  canvas.style.zIndex = '-1';
+  canvas.style.position = 'absolute';
+  document.body.appendChild(canvas);
+  const chart = new ApexCharts(canvas, options);
+  await chart.render();
+  await sleep(1000); // Wait animations are over
```
aHenryJard, line 23: Have you check if there is a way to disable animation in apexchart instead of sleep ?

aHenryJard: Maybe ? https://apexcharts.com/docs/options/chart/animations/

lndrtrbn: I tried this option but the result was still empty. The sleep option was what was used as accepted answer on a stackoverflow related question

aHenryJard (approved):
Tested without FF and with FF, all good to me.

PR #8554, opened 2024-09-30
[backend] change file upload to be case insensitive (#8547)

### LINKED ISSUE #8547: Internal document identifier are case sensitive
## Description

Internal document identifier are case sensitive

image

This situation leads to GraphQLError: Id loading expect only one response

We need to:
- Lower case the id
- Create a migration to remove the duplicate

JeremyCloarec:
### Proposed changes

* lowerCase file name on file upload to make it case insensitive
* keep existing file id casing in case the file already exists to be compatible with already uploaded files

### Related issues

* #8547 

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/tests_e2e/groupings/createGrouping.spec.ts
+++ b/opencti-platform/opencti-front/tests_e2e/groupings/createGrouping.spec.ts
@@ -31,6 +31,6 @@ test('Create a new grouping', async ({ page }) => {
   await stixDomainObjectContentTab.editFile('Test file.html', 'Test file content text');
   await stixDomainObjectContentTab.selectMainContent();
   await expect(page.getByText('Main content text')).toBeVisible();
-  await stixDomainObjectContentTab.selectFile('Test file.html');
+  await stixDomainObjectContentTab.selectFile('test file.html');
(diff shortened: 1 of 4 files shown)


opencti-platform/opencti-graphql/src/database/engine.js
```diff
@@ -1405,6 +1405,11 @@ export const elLoadById = async (context, user, id, opts = {}) => {
   const hits = await elFindByIds(context, user, id, opts);
   //* v8 ignore if */
   if (hits.length > 1) {
+    // If only files are returned, we filter a second time on their IDs to handle case sensitivity
+    if (hits.filter((e) => e.entity_type === ENTITY_TYPE_INTERNAL_FILE).length === hits.length) {
```
labo-flg:
```suggestion
    if (hits.every((e) => e.entity_type === ENTITY_TYPE_INTERNAL_FILE)) {
```

Archidoit (dismissed):

richard-julien:
Not tested (not time) but the approach looks good.
Can you add some code comments to explain why we overwrite the id after the fetch?

aHenryJard (approved):

PR #8569, opened 2024-10-01
[Backend] Enhance retention policy deletion performances / speed (#4864)

### LINKED ISSUE #4864: Enhance retention policy deletion performances / speed
### Enhance retention policy deletion performances / speed

- Scheduling moving to dynamic instead of static (that will improve the reschedule)
- Base schedule change to 30 sec / can be change with retention_manager:interval
- Pagination number move to 1500 / can be change with retention_manager:batch_size
- Introduction of deletion parallelization (10 by default) / can be change with retention_manager:max_deletion_concurrency option

### Side
- Upgrade axios dependency

### Testing
Testing was done with standard observable (multi relationships and multiple labels).
```
retentionManager.ts:95 > Deleted 1500 in 33604 (10)
retentionManager.ts:107 > retentionHandler 2024-10-01T20:08:53.567Z
retentionManager.ts:95 > Deleted 1500 in 33855 (10)
retentionManager.ts:107 > retentionHandler 2024-10-01T20:09:37.507Z
```
Lets take 35 seconds to delete 1500 observables.
Thats almost 42 delete per seconds and so almost 3.5 millions of deletion per day.

richard-julien:
See #4864

### DIFF
--- a/opencti-platform/opencti-graphql/src/manager/retentionManager.ts
+++ b/opencti-platform/opencti-graphql/src/manager/retentionManager.ts
@@ -13,9 +13,9 @@ import type { ManagerDefinition } from './managerModule';
 import { registerManager } from './managerModule';
 import type { AuthContext } from '../types/user';
 import type { FileEdge, RetentionRule } from '../generated/graphql';
+import { RetentionRuleScope, RetentionUnit } from '../generated/graphql';
 import { deleteFile } from '../database/file-storage';
 import { DELETABLE_FILE_STATUSES, paginatedForPathWithEnrichment } from '../modules/internal/document/document-domain';
-import { RetentionRuleScope, RetentionUnit } from '../generated/graphql';
 import type { BasicStoreCommonEdge, StoreObject } from '../types/store';
 import { ALREADY_DELETED_ERROR } from '../config/errors';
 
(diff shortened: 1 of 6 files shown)


opencti-platform/opencti-graphql/src/manager/retentionManager.ts
```diff
@@ -77,17 +80,22 @@ const executeProcessing = async (context: AuthContext, retentionRule: RetentionR
...
+      const { updated_at: up } = node;
+      try {
+        const humanDuration = moment.duration(utcDate(up).diff(utcDate())).humanize();
+        await deleteElement(context, scope, scope === 'knowledge' ? node.internal_id : node.id, node.entity_type);
+        logApp.debug(`[OPENCTI] Retention manager deleting ${node.id} after ${humanDuration}`);
+      } catch (e) {
+        logApp.error(e, { id: node.id, manager: 'RETENTION_MANAGER' });
+      }
+    };
+    await BluePromise.map(elements, deleteFn, { concurrency: RETENTION_MAX_CONCURRENCY });
```
labo-flg, line 88: Just for my understanding: if we batch elements that are interdependent, the first one might take a lock on another one in the concurrent batch, and when release, the second element to delete might already be deleted (for instance if it's a relationship). Am I right ?

If so, not harmful but we'll end up with a bunch of error logs.

labo-flg (dismissed):
Beside my comment on error logs, code seems ok to me !

Archidoit:
I created a retention policy with no filters and with retention = 1 day. It should delete all the entities of my platform.
I have this error in backend every few seconds:
image

And if I restart my platform : 
image

I think it's due to this modification : 
image

opencti-platform/opencti-graphql/src/manager/retentionManager.ts
```diff
@@ -42,7 +45,7 @@ export const getElementsToDelete = async (context: AuthContext, scope: string, b
   if (scope === 'knowledge') {
     const jsonFilters = filters ? JSON.parse(filters) : null;
     const queryOptions = await convertFiltersToQueryOptions(jsonFilters, { before });
-    result = await elPaginate(context, RETENTION_MANAGER_USER, READ_STIX_INDICES, { ...queryOptions, first: RETENTION_BATCH_SIZE });
+    result = await elPaginate(context, RETENTION_MANAGER_USER, READ_ENTITIES_INDICES_WITHOUT_INFERRED, { ...queryOptions, first: RETENTION_BATCH_SIZE });
```
Archidoit: why this modification?

richard-julien: Nice catch, it was a test. Will rollback

opencti-platform/opencti-graphql/src/manager/retentionManager.ts
richard-julien: Add code to prevent logging for an already deleted error

opencti-platform/opencti-graphql/src/manager/retentionManager.ts
richard-julien: done

Archidoit:
I created a retention policy on everything (Knowledge). It's indicating 512 elements will be deleting.
image


The retention rules is launched every 30s.
After 2 minutes nothing has been deleted. And when we 'test' the retention rule, the number of entities to be deleted is still the same : 
image
-->entities have not been deleted
image

This didn't happen for a retention policy with a filter, and so targeting less elements (exemple done with 2 reports to delete: the reports where deleted).

Archidoit:
The 'remaining' attribute seems not correct.
With this retention rule indicating 25 entities are 'remaining' while I only see 1 left : 
image

Archidoit:
Behavior ok 👍 

Small question: With a retention rule on Knowledge with no filters, organizations are also deleted (because they are stix domain objects)... Is it a problem?

Archidoit (dismissed):

Archidoit:
I don't really agree with the last commit that prevents empty filters for several reasons : 
- we avoid empty filters in UI but not in the API
- there is no messages in the UI informing the Create button is disabled because filters are now mandatory
- the problem is not really solved because if we add the filter: 'label different from label1' we will still delete the organization that haven't label1

--> I think that if we don't want to delete organizations we should think about it a bit more, maybe add a warning if a retention rule may delete organizations...

Archidoit (approved):

Archidoit:
> I don't really agree with the last commit that prevents empty filters for several reasons :
> 
> * we avoid empty filters in UI but not in the API
> * there is no messages in the UI informing the Create button is disabled because filters are now mandatory
> * the problem is not really solved because if we add the filter: 'label different from label1' we will still delete the organization that haven't label1
> 
> --> I think that if we don't want to delete organizations we should think about it a bit more, maybe add a warning if a retention rule may delete organizations...

Since it is not the scope, I reverted the last commit and created an issue for the organization deletion issue : https://github.com/OpenCTI-Platform/opencti/issues/8606

Issue #4230, opened 2023-09-05
Add Attack pattern list in Knowledge/Attack pattern with massive operations


Jipegien:
## Use case

Being able to use massive operation on attack pattern list associated with an SDO

## Current Workaround

None

## Proposed Solution

Add a simple list view of the attack pattern in the Knowledge/attack pattern tab, like in other tabs.

PR #8574, opened 2024-10-02
[frontend] Revert unnecessary changes in DataTableWithoutFragment (#4230)

Gwendoline-FAVRE-FELIX:
### Proposed changes

- It was planned to use the DataTableWithoutFragment component, which has been modified. In the end, it wasn't used, and unnecessary modifications were forgotten and led to errors.

- We take advantage of these ‘repairs’ to change a color.

### Related issues

- https://github.com/OpenCTI-Platform/opencti/issues/4230


Kedae:
I think the change on DataTableComponent should also be reverted ? Or I am missing something ?

Archidoit (approved):

PR #8577, opened 2024-10-02
[frontend/backend] Migrate CK Editor to 9.3 (#8151)

### LINKED ISSUE #8151: Upgrade CKeditor to latest version
## Use case

Currently using the 6.2 version while 9.00 is available.

Archidoit:
### Proposed changes

Change the way we use CKEditor in OpenCTI.

Actual usage:
We have fetched a fix version of CKEditor and build a lib ourselves with the config we want to use it in our code.

New usage:
We fetch CKEditor from yarn as any other libs of the project and use it directly inside a new React component `CKEditor` that initiate the config to keep same behavior we had with the older build.

The advantage of this is that it's easier to keep an up-to-date version of CKEditor.

### Related issues

- #8151

### DIFF
--- a/opencti-platform/opencti-front/src/components/CKEditor.tsx
+++ b/opencti-platform/opencti-front/src/components/CKEditor.tsx
@@ -1,0 +1,220 @@
+import { CKEditor as ReactCKEditor } from '@ckeditor/ckeditor5-react';
+import {
+  Editor,
+  Alignment,
+  Autoformat,
+  AutoImage,
+  AutoLink,
+  Base64UploadAdapter,
+  BlockQuote,
+  Bold,
+  ClassicEditor,
+  Code,
+  CodeBlock,
+  Essentials,
+  FontBackgroundColor,
+  FontColor,
+  FontFamily,
+  FontSize,
+  Heading,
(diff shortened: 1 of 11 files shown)


opencti-platform/opencti-front/src/components/CKEditor.tsx
```diff
@@ -0,0 +1,217 @@
...
+// eslint-disable-next-line import/extensions
+import ja from 'ckeditor5/translations/ja.js';
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore
+// eslint-disable-next-line import/extensions
+import ko from 'ckeditor5/translations/ko.js';
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore
+// eslint-disable-next-line import/extensions
+import zh from 'ckeditor5/translations/zh.js';
```
lndrtrbn, line 86: Yes, that's disgusting, I know 🤢 but cannot found a better way with how ckeditor is bundled...

SouadHadjiat (approved):
tested locally

Issue #8333, opened 2024-09-13
Missing entites in shared report


marieflorescontact:
## Description

Some entites (System, Individual) are not shared when a Report is shared with organization

## Environment

1. OS (where OpenCTI server runs): { e.g. Mac OS 10, Windows 10, Ubuntu 16.4, etc. }
2. OpenCTI version: { e.g. OpenCTI 1.0.2 }
3. OpenCTI client: { e.g. frontend or python }
4. Other environment details:

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Set a platform orga `org A`
2. create a `Report` and add entities : `System`, `Individual`,` Organization`, `Countries` etc
3. share the report with `org B`
4. log with a user from `org B`
5. Go to the `Report` > `entities tab`
6.  `System` and `Individual `are missing

## Expected Output

all entites

## Actual Output

all entities  except `System` and `Individual `

we also have erros in backend:

````
{"category":"APP","errors":[{"attributes":{"from":"Individual","genre":"BUSINESS","http_status":400,"to":"Organization","type":"granted"},"message":"The relationship is not allowed","name":"FUNCTIONAL_ERROR","stack":"GraphQLErro
r: The relationship is not allowed\n    at error (C:\\Users\\MarieFlores\\IdeaProjects\\opencti\\opencti-platform\\opencti-graphql\\src\\config\\errors.js:7:10)\n    at FunctionalError (C:\\Users\\MarieFlores\\IdeaProjects\\open
cti\\opencti-platform\\opencti-graphql\\src\\config\\errors.js:94:50)\n    at checkRelationshipRef (C:\\Users\\MarieFlores\\IdeaProjects\\opencti\\opencti-platform\\opencti-graphql\\src\\database\\stix.ts:1183:11)\n    at C:\\Us
ers\\MarieFlores\\IdeaProjects\\opencti\\opencti-platform\\opencti-graphql\\src\\utils\\modelConsistency.ts:32:9\n    at Array.forEach (<anonymous>)\n    at fn (C:\\Users\\MarieFlores\\IdeaProjects\\opencti\\opencti-platform\\op
encti-graphql\\src\\utils\\modelConsistency.ts:23:13)\n    at telemetry (C:\\Users\\MarieFlores\\IdeaProjects\\opencti\\opencti-platform\\opencti-graphql\\src\\config\\tracing.ts:121:10)\n    at checkRelationConsistency (C:\\Use
[...]

PR #8587, opened 2024-10-03
[backend] Fix missing entites in shared report (#8333)

marieflorescontact:
### Proposed changes

* check if granted ref relation is possible for `Organization` before relation creation (to avoid backend error)
*  specified that `Organization` and `Sector` are not concerned by stix ref instead of considering all `Identity`
* add organization sharing test => 2 update organizations
* add a json report file => 8create / 8delete / )
* counters: 
* 9 create ( 1 report + 8 entites )
* 6 update (orga sharing) 
* 9 delete 

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/8333

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/manager/taskManager.js
+++ b/opencti-platform/opencti-graphql/src/manager/taskManager.js
@@ -550,7 +550,7 @@ const executeProcessing = async (context, user, job, scope) => {
   return errors;
 };
 
-const taskHandler = async () => {
+export const taskHandler = async () => {
   let lock;
   try {
     // Lock the manager
--- a/opencti-platform/opencti-graphql/src/manager/taskManager.js
+++ b/opencti-platform/opencti-graphql/src/manager/taskManager.js
@@ -57,7 +57,7 @@ import { isStixCyberObservable } from '../schema/stixCyberObservable';
 import { promoteObservableToIndicator } from '../domain/stixCyberObservable';
 import { indicatorEditField, promoteIndicatorToObservables } from '../modules/indicator/indicator-domain';
 import { askElementEnrichmentForConnector } from '../domain/stixCoreObject';
-import { RELATION_GRANTED_TO, RELATION_OBJECT } from '../schema/stixRefRelationship';
+import { objectOrganization, RELATION_GRANTED_TO, RELATION_OBJECT } from '../schema/stixRefRelationship';
 import {
   ACTION_TYPE_COMPLETE_DELETE,
   ACTION_TYPE_DELETE,
@@ -79,6 +79,7 @@ import { BackgroundTaskScope } from '../generated/graphql';
 import { ENTITY_TYPE_INTERNAL_FILE } from '../schema/internalObject';
 import { deleteFile } from '../database/file-storage';
 import { checkUserIsAdminOnDashboard } from '../modules/publicDashboard/publicDashboard-utils';
+import { ENTITY_TYPE_IDENTITY_ORGANIZATION } from '../modules/organization/organization-types';
 
 // Task manager responsible to execute long manual tasks
 // Each API will start is task manager.
@@ -406,7 +407,7 @@ const executeShare = async (context, user, actionContext, element) => {
   for (let indexCreate = 0; indexCreate < values.length; indexCreate += 1) {
     const target = values[indexCreate];
     const currentGrants = element[buildRefRelationKey(RELATION_GRANTED_TO)] ?? [];
-    if (!currentGrants.includes(target)) {
+    if (!currentGrants.includes(target) && objectOrganization.isRefExistingForTypes(element.entity_type, ENTITY_TYPE_IDENTITY_ORGANIZATION)) {
       await createRelation(context, user, { fromId: element.id, toId: target, relationship_type: RELATION_GRANTED_TO });
     }
   }
--- a/opencti-platform/opencti-graphql/src/schema/stixRefRelationship.ts
+++ b/opencti-platform/opencti-graphql/src/schema/stixRefRelationship.ts
@@ -727,7 +727,9 @@ export const objectOrganization: RefAttribute = {
   multiple: true,
   upsert: true,
   isRefExistingForTypes(this, fromType, toType) {
-    return !(fromType === ENTITY_TYPE_EVENT || isStixDomainObjectIdentity(fromType)
+    return !(fromType === ENTITY_TYPE_EVENT
+        || fromType === ENTITY_TYPE_IDENTITY_ORGANIZATION
+        || fromType === ENTITY_TYPE_IDENTITY_SECTOR
         || isStixDomainObjectLocation(fromType))
       && this.toTypes.includes(toType);
   },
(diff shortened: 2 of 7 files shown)


opencti-platform/opencti-graphql/src/schema/stixRefRelationship.ts
```diff
@@ -727,7 +726,9 @@ export const objectOrganization: RefAttribute = {
   multiple: true,
   upsert: true,
   isRefExistingForTypes(this, fromType, toType) {
-    return !(fromType === ENTITY_TYPE_EVENT || isStixDomainObjectIdentity(fromType)
+    return !(fromType === ENTITY_TYPE_EVENT
+        || fromType === ENTITY_TYPE_IDENTITY_ORGANIZATION
+        || fromType === ENTITY_TYPE_IDENTITY_SECTOR
```
lndrtrbn, line 731: Not sure to understand why we replace the call to identity function and replace it with 2 conditions ?

opencti-platform/opencti-graphql/tests/03-streams/00-Raw/raw-test.js
```diff
@@ -58,8 +58,8 @@ describe('Raw streams tests', () => {
       expect(updateEventsByTypes['marking-definition'].length).toBe(2);
       expect(updateEventsByTypes['campaign'].length).toBe(7);
       expect(updateEventsByTypes['relationship'].length).toBe(8);
-      expect(updateEventsByTypes['identity'].length).toBe(16);
-      expect(updateEventsByTypes['malware'].length).toBe(16);
+      expect(updateEventsByTypes['identity'].length).toBe(18);
```
lndrtrbn, line 61: For curiosity, why here we increment by 2 instead of 4 as we have 4 identities in the bundle ? 2 already exist in the platform ?

opencti-platform/opencti-graphql/src/manager/taskManager.js
```diff
@@ -406,7 +407,7 @@ const executeShare = async (context, user, actionContext, element) => {
   for (let indexCreate = 0; indexCreate < values.length; indexCreate += 1) {
     const target = values[indexCreate];
     const currentGrants = element[buildRefRelationKey(RELATION_GRANTED_TO)] ?? [];
-    if (!currentGrants.includes(target)) {
+    if (!currentGrants.includes(target) && objectOrganization.isRefExistingForTypes(element.entity_type, ENTITY_TYPE_IDENTITY_ORGANIZATION)) {
```
lndrtrbn, line 410: Can you explain the change?

opencti-platform/opencti-graphql/tests/03-streams/00-Raw/raw-test.js
marieflorescontact: We have 6 updates due to sharing to Oganization:
- 1 Report
- 1 Indicator  + 1 Individual (`Identity`) + 1 System (`Identity`) + 1 Malware + 1 SCO

`City` `Organization` and `Sector` are entities to wich restriction dosen't not apply

opencti-platform/opencti-graphql/src/schema/stixRefRelationship.ts
marieflorescontact: `Organization` and `Sector`  are not concerned by organization segregation, we need to test for these specific Entites to exclude them from sharing

opencti-platform/opencti-graphql/tests/03-streams/00-Raw/raw-test.js
lndrtrbn: Ah okok thanks!

opencti-platform/opencti-graphql/src/schema/stixRefRelationship.ts
lndrtrbn: But was `isStixDomainObjectIdentity(fromType)` not already doing it ?

lndrtrbn: Ah, but System and and Individual are also Indentities but for them we want the sharing to be working so the function was to restrictive ?

opencti-platform/opencti-graphql/src/manager/taskManager.js
lndrtrbn: Had the answer in the description oft he PR

opencti-platform/opencti-graphql/src/schema/stixRefRelationship.ts
lndrtrbn: Had the answer in the description oft he PR

opencti-platform/opencti-graphql/src/manager/taskManager.js
lndrtrbn: Nop in fact even with description I do not understand how this line check that the relation GRANTED REF is possible between the 2 elements as we are not specifying the type of the ref anywhere ? 😅

marieflorescontact: when we share an element with an `Organization`, we create the `granted` relationship if the element's granted relationships don't include the target (= if the element isn't already shared with this orga) and if it's possible to create a `granted` relationship for this entity. 
The method `isRefExistingForTypes(element, orga)` checks that the element is not an `Event`, `Organization`, `Sector` or `Location`.

Before the modification, this check was not done and the backend sent an error, as it was not possible to create a `granted` relationship between an `Organization` and a `Sector` or between 2 Organizations. (you can see the error in the ticket).

all unrestricted entities are listed in `stixDomainObject.ts` file line 182.
```
export const STIX_ORGANIZATIONS_UNRESTRICTED = [
  ABSTRACT_INTERNAL_OBJECT,
  ABSTRACT_INTERNAL_RELATIONSHIP,
  ABSTRACT_STIX_META_OBJECT,
  ABSTRACT_STIX_REF_RELATIONSHIP,
  ENTITY_TYPE_IDENTITY_ORGANIZATION,
  ENTITY_TYPE_IDENTITY_SECTOR,
  ENTITY_TYPE_LOCATION,
  ENTITY_TYPE_WORK, // Work is defined as an history object
  ENTITY_TYPE_TAXII_COLLECTION, // TODO TaxiiCollection must be migrate to add according parent types
  ENTITY_TYPE_INTERNAL_FILE, // TODO InternalFile must be migrate to add according parent types
];
```

lndrtrbn: Thanks for the explanation.

So the function called `isRefExistingForTypes` checks if it is possible to have a `granted` ref between two types? That's quite confusing to me 😅

SouadHadjiat: yes this method is called when creating the relation and if it returns false, we throw an exception. The idea here was to call the method before trying to create a granted relationship that won't work and avoid unecessary errors in the logs.
image

lndrtrbn (approved):

PR #8592, opened 2024-10-03
[Backend] Get all objects pagination can stop for invalid global count (#8579)

### LINKED ISSUE #8579: Knowledge graph is not displaying all entities
## Description

https://demo.octi.filigran.io/dashboard/cases/incidents/2d97d1c6-d4ce-4a68-a16b-2265b734c0f7/knowledge/graph

image

=> Seems to be limited to 500

But in this incident, we have more than 2K entities:

image

richard-julien:
See #8579

### DIFF
--- a/opencti-platform/opencti-graphql/src/database/middleware-loader.ts
+++ b/opencti-platform/opencti-graphql/src/database/middleware-loader.ts
@@ -545,7 +545,13 @@ export const listEntitiesThroughRelationsPaginated = async <T extends BasicStore
   });
   return {
     edges: rebuildEdges,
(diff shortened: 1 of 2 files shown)


JeremyCloarec:
Tested OK and code is OK for me.
But did we find out why the report linked in the issue had the problem ? What caused the listEntitiesThroughRelationsPaginated to return only 499 entities, shouldn't we have access to all the object refs with our access rights? Is there an inconsistency in the database between the denormalized object relations on the report and the existing object relations?

richard-julien:
The reason is due to the fact that the rel_ are not directional in the current model.
This direction is post filtered in the query and so influence the number of result.

JeremyCloarec (approved):

Issue #8445, opened 2024-09-22
English grammar errors in the toggle labels for the "Update a group" panel 


seanthegeek:
## Description

Three of the toggle options at the bottom of the "Update a group" panel have English grammar errors that make what the option does unclear.

## Environment

1. OS (where OpenCTI server runs): Debain 12
2. OpenCTI version: 6.3.1
3. OpenCTI client: Frontend

## Reproducible Steps

1. Navigate to Settings> Security
2. Click Groups
3. Click on an existing group
4. Click on the edit/pencil button

Label text | Corrected grammar| Notes
------------|----------------------|--------
Automatically authorize this group to new marking definition || I'm not sure what this is, considering managing marking definitions are granted be a role capability 
Do not accumulate creators for the users of this group ||I found [the request](https://github.com/OpenCTI-Platform/opencti/issues/7290) that added the option, but I don't know what accumulating creators means, and I can't find a reference in the documentation 
Deny users to delete content not created by them | Do not allow users to delete content created by other users

## Screenshots (optional)

![image](https://github.com/user-attachments/assets/7cb02ed2-e55e-4bda-9bd3-be0a1d3d8d40)

nino-filigran:
Hi @seanthegeek thanks for your feedback. We'll treat your bug as soon as we can. However I can already give you some hints for your comments in the ticket:

- I'm not sure what this is, considering managing marking definitions are granted be a role capability: sometimes, data is incoming with some specific (and new) marking that are not in your platform. Therefore, these markings automatically needs to be added to the group (which the user belongs to) in order to be able to create the data with the correct marking. It is especially usefull in the context of connectors. This behavior exists on the side than the "manage marking definition" capability since as explained above, this action needs to be done automatically. You still need the capability manage marking definition on the top of it.
[...]

PR #8595, opened 2024-10-03
[frontend] English grammar errors in the toggle labels for the "Update a group" panel (#8445)

SarahBocognano:
### Proposed changes

* See issue

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/8445

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/settings/groups/GroupEditionOverview.tsx
+++ b/opencti-platform/opencti-front/src/private/components/settings/groups/GroupEditionOverview.tsx
@@ -209,7 +209,7 @@ const GroupEditionOverviewComponent: FunctionComponent<GroupEditionOverviewCompo
               type="checkbox"
               name="restrict_delete"
               label={t_i18n(
-                'Deny users to delete content not created by them',
+                'Do not allow these users to delete content created by other users',
               )}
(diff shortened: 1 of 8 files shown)


opencti-platform/opencti-front/src/private/components/settings/groups/GroupEditionOverview.tsx
```diff
@@ -209,7 +209,7 @@ const GroupEditionOverviewComponent: FunctionComponent<GroupEditionOverviewCompo
               type="checkbox"
               name="restrict_delete"
               label={t_i18n(
-                'Deny users to delete content not created by them',
```
labo-flg, line 212: you might want to remove the corresponding intl keys ?

SarahBocognano: It wasn't existing in the first place :( 
that's why I added translation for the other sentences too

seanthegeek:
What about the marking definitions text?

labo-flg (approved):

SarahBocognano:
Hi @seanthegeek , as my responsable of product said in the issue's comment part, we've decided to only change this text and not the one related to markings :)

Issue #8125, opened 2024-08-21
perf: improve the amount of entities loaded in the platform


nino-filigran:
## Use case

Currently we struggle to load more than 3000+ entities (see https://github.com/OpenCTI-Platform/opencti/issues/7398 ). 
It causes the platform to either become unresponsive or crash. 

## Current Workaround

None. 

## Proposed Solution

Improve the way we load entities & relations (batch).
Improve the technology we use ?

PR #8615, opened 2024-10-06
[backend] Improve performance by better targeting indices for query

richard-julien:
[backend] Improve performance by better targeting indices for query

* Listing are now indices computed (elPaginate)


opencti-platform/opencti-graphql/src/database/middleware-loader.ts
```diff
@@ -292,26 +292,29 @@ export const buildRelationsFilter = <T extends BasicStoreCommon>(relationTypes:
...
   }
   const paginateArgs = buildRelationsFilter(type, args);
-  return elPaginate(context, user, indices, paginateArgs);
+  return elPaginate(context, user, computedIndices, paginateArgs);
 };
 
 export const listAllRelations = async <T extends StoreProxyRelation>(context: AuthContext, user: AuthUser, type: string | Array<string>,
   args: RelationOptions<T> = {}): Promise<Array<T>> => {
-  const { indices = READ_RELATIONSHIPS_INDICES } = args;
+  const { indices } = args;
```
SouadHadjiat, line 314: why remove here the default indices READ_RELATIONSHIPS_INDICES ? written like this, indices doesn't have a default value, and if type is not specified, computeQueryIndices(indices, type); will return READ_DATA_INDICES even though here we know that we want to read only relationships

SouadHadjiat: same question for all other methods where indices had a default value

richard-julien: For me type is mandatory here. Do you see some usages without the type?

SouadHadjiat: you are right, type is always provided here

opencti-platform/opencti-graphql/tests/02-integration/01-database/document-test.ts
```diff
@@ -1,36 +0,0 @@
...
-describe('checkFileAccess coverage', () => {
-  it('should be allowed for any authenticated user to get an imported file', async () => {
-    const myUser: AuthUser = await findById(testContext, ADMIN_USER, USER_PARTICIPATE.id);
-    const opts: { entity_id?: string, filename: string, id: string } = { entity_id: 'fake-entityid', filename: 'file.json', id: 'export/report/file.json' };
-    expect(await checkFileAccess(testContext, myUser, 'read', opts)).toBeTruthy();
-  });
-
-  it('should be forbidden for any authenticated user without SUPPORT capa to get an support file', async () => {
-    const myUser: AuthUser = await findById(testContext, ADMIN_USER, USER_PARTICIPATE.id);
-    const opts: { entity_id?: string, filename: string, id: string } = { entity_id: 'fake-entityid', filename: 'anyfile.zip', id: `${SUPPORT_STORAGE_PATH}/anyfile.zip` };
```
SouadHadjiat: so since we remove this test because of the refactoring, do we have another test where we check file access ? I don't see one.

opencti-platform/opencti-graphql/src/database/engine.js
```diff
@@ -1474,11 +1476,13 @@ const computeQueryIndices = (indices, types) => {
           return READ_ENTITIES_INDICES;
         }
         // For relationships
```
SouadHadjiat, line 1489: Question for l1471: why do we have READ_INDEX_INFERRED_ENTITIES for internal object types ? do we create internal objects with inferrence rules ? from what I see, the only inferred entities we create are incidents (ENTITY_TYPE_INCIDENT)

Kedae:
Linked to #8125

opencti-platform/opencti-graphql/tests/02-integration/01-database/document-test.ts
richard-julien: As the method is now deleted, not sure its necessary

opencti-platform/opencti-graphql/src/database/engine.js
richard-julien: Yes, the only type is ENTITY_TYPE_INCIDENT

aHenryJard (commented):
Could you split the checkFileAccess changes to another PR ?

aHenryJard (approved):
Tested locally, looks good

PR #8638, opened 2024-10-08
[backend] Allow csv mapper to import entities present in several row and stream csv reading  (#8403)(#7589)

### LINKED ISSUE #8403: CSV mapper: Ability to import entities that are present in multiple rows with various data for the same attribute & hhandle larger files through streaming
## Use case

To avoid having mutliple columns or multiple values within the same fields, some users/systems create CSV having the same entity present on two lines, for instance to map several labels on the same entity.
We would like to be able to ingest entities in this situation, without skipping them.

aHenryJard:
### Proposed changes

1. Allow CSV Mapper to import entities present in several row:
- stop removing all duplicated stix id in the bundle generated from CSV, instead do a diff on all entity fields to check if it's a full duplicate or not
- since worker is also removing duplicates on bundle based on the generated stix id; when 2 csv lines generates the same stix id a new bundle is created.
- filter bundle before send to keep last version of a duplicated entity (instead of first one in the current code)

2. Stream imported csv files to avoid loading the whole file in memory:
- add a parameter in importCsv-connector for the bulk size of data to process (bulk_creation_size)
- based on https://github.com/OpenCTI-Platform/opencti/pull/7594 work
- porcess of reading : open minIO file, read the bulk_creation_size line, close minIO file, parse and run csv mapper on the bulk. reason: **Keeping the minio file open causes socket timeout issue on huge files.**
- workbench are out of stream scope, the previous code is still used (with a comment on deprecated code)
- deprecation on the test part (because API changes)

**Impacts:**
- CSV Mapper, csv mapper test
- Data > import csv file
- Entitiy > Data import
- CSV Feed (only for the entity in several rows, not concerned by stream)

**Workbench are out of scope**

### Related issues

* relates to #7400 
* relates to #7589 
* closes #8403

### Checklist


CelineSebe:
tested in local with malware analysis and IPV4 files 👌

opencti-platform/opencti-graphql/src/parser/bundle-creator.ts
```diff
@@ -2,6 +2,43 @@ import { v4 as uuidv4 } from 'uuid';
...
+          currentToCheckClone.extensions[STIX_EXT_OCTI].converter_csv = undefined;
+        }
+
+        const itemAsJson = JSON.stringify(itemClone);
+        const currentAsJson = JSON.stringify(currentToCheckClone);
+        return itemAsJson !== currentAsJson;
+      }
+      return false;
+    });
+    canAdd = canAdd && !existingObjectWithDifferentContent;
```
JeremyCloarec: If existingObjectWithDifferentContent is true, we could directly return false to not have to check all other objectsToAdd

opencti-platform/opencti-graphql/src/parser/bundle-creator.ts
```diff
@@ -2,6 +2,43 @@ import { v4 as uuidv4 } from 'uuid';
...
+ * @param objectsToAdd
+ * @param bundles
+ */
+export const canAddObjectToBundle = (objectsToAdd: StixObject[], bundles: StixObject[]): boolean => {
+  let canAdd = true;
+  for (let i = 0; i < objectsToAdd.length; i += 1) {
+    const currentToCheck = objectsToAdd[i];
+    const existingObjectWithDifferentContent = bundles.find((item: StixObject) => {
+      if (item.id === currentToCheck.id && item.type === currentToCheck.type) {
+        const itemClone = JSON.parse(JSON.stringify(item));
```
JeremyCloarec: I don't get why we do a stringify => parse to create a clone? Can't we simplify it by using structuredClone?

opencti-platform/opencti-graphql/src/parser/csv-bundler.ts
```diff
@@ -71,3 +74,127 @@ export const bundleProcess = async (
...
+  }
+  // Handle container
+  if (entity && isStixDomainObjectContainer(entity.entity_type)) {
+    for (let i = 0; i < allBundles.length; i += 1) {
+      const currentBundle = allBundles[i];
+      const refs = currentBundle.ids();
+      const stixEntity = { ...convertStoreToStix(entity), [objects.stixName]: refs };
+      currentBundle.addObject(stixEntity);
+    }
+  }
```
JeremyCloarec: We maybe need to check that the entity itself isn't already contained in the bundle also?

opencti-platform/opencti-graphql/src/connector/importCsv/importCsv-connector.ts
```diff
@@ -26,79 +32,227 @@ const connectorConfig: ConnectorConfig = {
...
+        const startingLineNumber = bulkLineCursor;
+        logApp.debug(`${logPrefix} reading line from ${bulkLineCursor} to ${BULK_LINE_PARSING_NUMBER + bulkLineCursor}`);
+        // Need an async interator to prevent blocking
+        // eslint-disable-next-line no-restricted-syntax
+        for await (const line of readStream) {
+          if (startingLineNumber <= lineNumber && lineNumber < startingLineNumber + BULK_LINE_PARSING_NUMBER) {
+            // We are in the bulk window
+            if (removeHeaderIsRequired) {
+              // Manage header removal: if csv file start with skip char, need to skipline until header is there
+              if (sanitizedMapper.skipLineChar && sanitizedMapper.skipLineChar.length === 1) {
```
JeremyCloarec, line 108: Shouldn't the skipLineChar be checked even for lines after the header? I thought the skipLineChar was used to "comment" some lines of data among all the other lines of data

opencti-platform/opencti-graphql/config/default.json
```diff
@@ -196,7 +196,8 @@
   "import_csv_built_in_connector": {
     "enabled": true,
     "interval": 10000,
-    "validate_before_import": false
+    "validate_before_import": false,
+    "bulk_creation_size": 1000
```
aHenryJard:
```suggestion
    "bulk_creation_size": 5000
```

opencti-platform/opencti-graphql/src/parser/bundle-creator.ts
aHenryJard: I have no idea on the best way to do a deep clone in js, do you have an example ? or is it the {...item } way ? I tested it's not doing a deep clone.

opencti-platform/opencti-graphql/src/connector/importCsv/importCsv-connector.ts
aHenryJard: skipLine char is managed by the csv parser for other use case than header, so I kept it. It's possible to filter lines before but not needed.

opencti-platform/opencti-graphql/src/parser/bundle-creator.ts
JeremyCloarec: I used structuredClone(item) when I needed to do a deep clone. It is used in the buildRestrictedEntity of middleware.js for exemple. Is the JSON.parse(JSON.stringify(item)) creating a deep clone or is it doing some mofidication to the original item in the process?

opencti-platform/opencti-graphql/src/connector/importCsv/importCsv-connector.ts
JeremyCloarec: My bad, you're right, I didn't see it in the parser

opencti-platform/opencti-graphql/src/database/rabbitmq.js
```diff
@@ -320,9 +320,7 @@ export const consumeQueue = async (context, connectorId, connectionSetterCallbac
         } else { // Connection success
           logApp.info('[QUEUEING] Starting connector queue consuming', { connectorId });
           conn.on('close', (onConnectError) => {
-            if (onConnectError) {
-              reject(onConnectError);
-            }
+            reject(onConnectError);
```
JeremyCloarec: Is there a reason to change this? I remember that it was intended this way: close event can be sent for regular connection close, it doesn't always mean that there is an error

opencti-platform/opencti-graphql/src/parser/bundle-creator.ts
aHenryJard: I need the same entity except for 2 fields => so I'm cloning and then removing 2 fields (that's why I need a clone)

aHenryJard: thanks trying structuredClone

opencti-platform/opencti-graphql/src/parser/bundle-creator.ts
aHenryJard: oh yes thanks, testing :)

opencti-platform/opencti-graphql/src/parser/csv-bundler.ts
aHenryJard: I think that I just moved this piece of code, I have no idea

opencti-platform/opencti-graphql/src/database/rabbitmq.js
aHenryJard: not sure on this change I think I inherited it from https://github.com/OpenCTI-Platform/opencti/pull/7594

opencti-platform/opencti-graphql/src/database/file-storage.js
```diff
@@ -78,7 +78,7 @@ export const initializeFileStorageClient = async () => {
     forcePathStyle: true,
     credentialDefaultProvider: await buildCredentialProvider(),
     logger: useAwsLogs ? logS3Debug : undefined,
-    tls: useSslConnection
+    tls: useSslConnection,
```
aHenryJard:
```suggestion
    tls: useSslConnection
```

opencti-platform/opencti-graphql/src/parser/bundle-creator.ts
```diff
@@ -2,6 +2,42 @@ import { v4 as uuidv4 } from 'uuid';
...
+  for (let i = 0; i < objectsToAdd.length; i += 1) {
+    const currentToCheck = objectsToAdd[i];
+    const existingObjectWithDifferentContent = bundles.find((item) => {
+      if (item.id === currentToCheck.id && item.type === currentToCheck.type) {
+        const itemClone = structuredClone(item);
+        if (itemClone.extensions) {
+          itemClone.extensions[STIX_EXT_OCTI].converter_csv = undefined;
+        }
+
+        const currentToCheckClone = structuredClone(currentToCheck);
```
JeremyCloarec: A small perf improvement could be to build the currentToCheckClone outside of the bundles.find(), to only clone it one time instead of on every bundle item

opencti-platform/opencti-graphql/src/parser/bundle-creator.ts
```diff
@@ -2,6 +2,42 @@ import { v4 as uuidv4 } from 'uuid';
...
+export const canAddObjectToBundle = (objectsToAdd: StixObject[], bundles: StixObject[]): boolean => {
+  const canAdd = true;
+  for (let i = 0; i < objectsToAdd.length; i += 1) {
+    const currentToCheck = objectsToAdd[i];
+    const existingObjectWithDifferentContent = bundles.find((item) => {
+      if (item.id === currentToCheck.id && item.type === currentToCheck.type) {
+        const itemClone = structuredClone(item);
+        if (itemClone.extensions) {
+          itemClone.extensions[STIX_EXT_OCTI].converter_csv = undefined;
+        }
```
JeremyCloarec: The item clones could maybe be cached? To prevent cloning it again on every call to objectsToAdd with the same bundles

aHenryJard: I think I need to see if .converter_csv is needed, because else we could just remove the whole clone part.

JeremyCloarec: That would be ideal yes

opencti-platform/opencti-graphql/tests/02-integration/07-connector/importCsv-connector-test.ts
```diff
@@ -0,0 +1,59 @@
...
+    const totalObjectsCount = await processCSVforWorkers(testContext, 'import/global/csv-file-cities.csv', mapperOpts);
+
+    // Bulk size = 5
+    //
+    // 3 first city line => same city on 2 first lines: 2 city + 'label1', 1 city + label2  = 5 objects
+    // next 5 lines => 1 skip line, 4 cities, 1 label = 5 objects
+    // next 5 lines => 5 cities + 1 label = 6 objects
+    // next 5 lines => 3 cities + 1 label ; + same city + 1 label => 6 objects (in 2 bundles)
+    // FIXME expect(totalObjectsCount).toBe(5 + 5 + 6 + 6);
+    expect(totalObjectsCount).toBe(32); // FIXME ??
```
JeremyCloarec: I don't expect 5 + 5 + 6 + 6, but 4 + 5 + 6 + 5. But anyway, I agree that I don't understand how it's 32?

opencti-platform/opencti-graphql/tests/02-integration/07-connector/importCsv-connector-test.ts
```diff
@@ -0,0 +1,59 @@
...
+      workId: work.id
+    };
+    const totalObjectsCount = await processCSVforWorkers(testContext, 'import/global/csv-file-cities.csv', mapperOpts);
+
+    // Bulk size = 5
+    //
+    // 3 first city line => same city on 2 first lines: 2 city + 'label1', 1 city + label2  = 5 objects
+    // next 5 lines => 1 skip line, 4 cities, 1 label = 5 objects
+    // next 5 lines => 5 cities + 1 label = 6 objects
+    // next 5 lines => 3 cities + 1 label ; + same city + 1 label => 6 objects (in 2 bundles)
```
JeremyCloarec: Shouldn't it be only 1 bundle if the city is a complete duplicate? So only 5 objects

aHenryJard: it's not a complete duplicate, it's same city but with different label.

opencti-platform/opencti-graphql/tests/02-integration/07-connector/importCsv-connector-test.ts
aHenryJard: Ok I found :D it's because count is before deduplication of stix id per bundle (it's done just before sending)

JeremyCloarec (approved):

PR #8644, opened 2024-10-11
[frontend] Fix the positions when adding an entity in all the Knowlege graph view (#7298)

### LINKED ISSUE #7298: In knowledge graph, do not re-apply forces when adding a new entity (working already when adding a new relations)
## Use case

In knowledge graph, do not re-apply forces when adding a new entity (working already when adding a new relations).

Very frustrating current behaviour:

1. Add dozens of entities
2. Take 10 min to properly position all the nodes and having a very beautiful graphs with manual positionning. 
3. Add 1 new entities

=> Everything you have done is lost....

Gwendoline-FAVRE-FELIX:
### Proposed changes

When you place entities manually and then add a new entity. If the forces are deactivated, make sure that the entities stay in the place they were given

### Related issues

* #7298 
*

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/workspaces/investigations/InvestigationGraph.jsx
+++ b/opencti-platform/opencti-front/src/private/components/workspaces/investigations/InvestigationGraph.jsx
@@ -1042,14 +1042,20 @@ class InvestigationGraphComponent extends Component {
   savePositions() {
     const initialPositions = R.indexBy(
       R.prop('id'),
-      R.map((n) => ({ id: n.id, x: n.fx, y: n.fy }), this.graphData.nodes),
+      R.map((n) => ({
+        id: n.id,
+        x: n.fx !== null ? n.fx : n.x,
+        y: n.fy !== null ? n.fy : n.y,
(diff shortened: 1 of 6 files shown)


opencti-platform/opencti-front/src/private/components/workspaces/investigations/InvestigationGraph.jsx
```diff
@@ -1429,7 +1440,7 @@ class InvestigationGraphComponent extends Component {
     this.graphObjects = [...this.graphObjects, stixCoreObject];
     this.graphData = buildGraphData(
       this.graphObjects,
-      decodeGraphData(this.props.workspace.graph_data),
+      decodeGraphData(this.props.workspace.x_opencti_graph_data),
```
SouadHadjiat: should be `graph_data`, not `x_opencti_graph_data`

SouadHadjiat (approved):
Pair reviewed with @Gwendoline-FAVRE-FELIX, positions were already all saved (but with null values), now we save them all with the right values, so looks good to me.

PR #8645, opened 2024-10-11
[frontend] fix ui after apexcharts update 3.54.0

labo-flg:
Latest update of apexcharts (3.54.0) introduced a styling issue.
Probably side effect of a fix for [this issue](https://github.com/apexcharts/apexcharts.js/issues/4713).

Issue only on testing platform.


marieflorescontact (approved):
locally tested ok 👌

PR #8667, opened 2024-10-15
[backend/frontend] Allow security admin to renew users token (#6643)

### LINKED ISSUE #6643: Leaked Tokens are not revokable
## Environment
1. OS (where OpenCTI server runs): debain (k8s)
2. OpenCTI version: 6.0.9

## Description
If an access token is leaked there is no possibility to revoke it or to generate a new one. The only possibility that exists is to delete the user and generate a new one. But with this approach the complete history gets lost (not associated) because it is not associated with the new user.

## Solution
A user has not automatically a token assigned. There exists the possibility for the user and for an admin to create a token. If the token is leaked the user or the admin can revoke the token (if deleting is enough depends on the current auth implementation, idk) and generate a new one, if necessary.

aHenryJard:
### Proposed changes

* Add test coverage on existing renew query
* Add frontend action in the user detail screen + popup to confirm before token revoke.
* Add an audit log on token renewal
* Add required capa SETTINGS_SETACCESSES both at UI level and graphql level (with auth) to prevent organization admin with this capa to renew other users token.
* Forbid renewal for admin token defined in configuration (either yaml or env variables)

Refactors and improvements:
- move user-test to typescript

Capture d'écran 2024-10-15 101159

### Related issues

* closes #6643 

### Checklist


opencti-platform/opencti-graphql/src/domain/user.js
```diff
@@ -1379,15 +1379,32 @@ export const resolveUserById = async (context, id) => {
...
+export const resolveUserByToken = async (context, tokenValue) => {
   const client = await elLoadBy(context, SYSTEM_USER, 'api_token', tokenValue, ENTITY_TYPE_USER);
   return buildCompleteUser(context, client);
 };
 
 export const userRenewToken = async (context, user, userId) => {
+  const userData = await storeLoadById(context, user, userId, ENTITY_TYPE_USER);
+  if (!userData) {
+    throw FunctionalError(`Cannot renew token, ${userId} user cannot be found.`);
+  }
```
SouadHadjiat, line 1405: I think we need to protect this API when it's an admin orga that calls it

labo-flg (approved):
tested ok

opencti-platform/opencti-graphql/src/domain/user.js
aHenryJard: ah yes thanks

aHenryJard: protected at graphql level with `auth`

SouadHadjiat (approved):

aHenryJard:
Change required: yaml admin token should not be renewable that way.

PR #8671, opened 2024-10-15
[backend] Improve id generation for STIX elements (#7158)

### LINKED ISSUE #7158: Limit stix_ids explosion by rewriting the standard_id in client python
## Use case

Limit stix_ids explosion by rewriting the standard_id in client python.

## Context

As described in https://github.com/OpenCTI-Platform/opencti/pull/6774 by @ckane , because of remote system or bad designed connectors, a lot of stix ids can be cumulated in an element.

`This seems to be able to result in a long sequence of entries in redis, during re-ingestion of the same entity (such as a malware STIX type) with new randomly-assigned STIX ids. My observation is that these get added to the stream once for each new STIX id encountered for the same item, and the list of stix_ids in the added/updated entity can grow infinitely. This is the source of one of the reasons why redis memory consumption can grow so high, under some circumstances, despite implementing a low TRIMMING size (such as 100000).`

## Solution

We need to generate the standard id on the client python and replace dynamically all the ids with the predicted ones.
By default, drop the base id of the source but add an option to configure in the connector if you want to consolidate them.
+ a cleanup script for existing platform to remove all the bad stix_ids already written

See https://github.com/OpenCTI-Platform/client-python/issues/659

## Why not only a task in client python?

Its also a good moment to check and enforce the ids generation on opencti side.

richard-julien:
Improve id generation for STIX elements
Add more dependencies
Adapt the graphql model
Add relationship and sighting generation

Related PR
https://github.com/OpenCTI-Platform/connectors/pull/2786 -> Merged
https://github.com/OpenCTI-Platform/client-python/pull/748


opencti-platform/opencti-graphql/tests/02-integration/05-parser/csv-parser-test.js
```diff
@@ -120,13 +120,13 @@ describe('CSV-PARSER', () => {
       .toBe(42);
     const relationshipTargets = objects.filter((o) => o.relationship_type === 'targets');
     expect(relationshipTargets.length)
-      .toBe(118);
+      .toBe(117);
     const relationshipLocatedAt = objects.filter((o) => o.relationship_type === 'located-at');
     expect(relationshipLocatedAt.length)
-      .toBe(130);
+      .toBe(125);
```
SouadHadjiat: why do we get less relationships in this test ?

opencti-platform/opencti-graphql/src/schema/identifier.js
```diff
@@ -185,27 +186,25 @@ const stixBaseEntityContribution = {
     // Entities
     [D.ENTITY_TYPE_ATTACK_PATTERN]: [[{ src: X_MITRE_ID_FIELD }], [{ src: NAME_FIELD }]],
     [D.ENTITY_TYPE_CAMPAIGN]: [{ src: NAME_FIELD }],
-    [D.ENTITY_TYPE_CONTAINER_NOTE]: [{ src: 'content' }, { src: 'created' }],
+    [D.ENTITY_TYPE_CONTAINER_NOTE]: [{ src: 'content' }, { src: 'created', dependencies: ['content'] }],
```
aHenryJard, line 189: What is the purpose of this "dependencies" information in the platform ?

aHenryJard (approved):
Tested locally along with client-python PR several connector running, everything looks good to me.

opencti-platform/opencti-graphql/src/schema/identifier.js
richard-julien: dependencies is the way to make attribute mandatory to each other

opencti-platform/opencti-graphql/tests/02-integration/05-parser/csv-parser-test.js
richard-julien: If i remember well its because the dedup is done on ids and as the id are generated now, its considered as duplicate.
Need to double check. @aHenryJard do you take a look on that in the PR review?

aHenryJard: Good point, I will look

aHenryJard: (I removed approval until I understand this point)

aHenryJard: Diff is legitimate because there was a before this PR
image

aHenryJard (approved):

Issue #6049, opened 2024-02-20
Ability to update the assignee, priority, severity, type  field through mass operation 


nino-filigran:
## Use case

In case a manager needs to assign analysts to some reports or other entities (to be defined), it would be great to be able to bulk edit all the objects in order to provide an assignee to all of them in one go, through the mass operations

PR #8693, opened 2024-10-17
[frontend] Mass operations (severity, priority, assignee...) on cases and containers (#6049)

SarahBocognano:
### Proposed changes

* Add severity/priority/assignee... on bulk edition for cases entities

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/6049


SarahBocognano:
*** Need to resolve a small little problem here, when selecting multiple "vocabularies" field, I fetch every values (exemple: severity values + priority values***

opencti-platform/opencti-front/src/private/components/data/DataTableToolBar.jsx
```diff
@@ -59,6 +59,10 @@ import Grid from '@mui/material/Grid';
 import Avatar from '@mui/material/Avatar';
 import FormControlLabel from '@mui/material/FormControlLabel';
 import Checkbox from '@mui/material/Checkbox';
+import { map, pathOr, pipe, union } from 'ramda';
```
lndrtrbn: Prefer using native JavaScript instead of Ramda.

opencti-platform/opencti-front/src/private/components/data/DataTableToolBar.jsx
```diff
@@ -721,11 +764,25 @@ class DataTableToolBar extends Component {
...
     const disabled = R.isNil(actionsInputs[i]?.type) || R.isEmpty(actionsInputs[i]?.type);
-    const typesHaveScore = selectedTypes.every((selectedType) => typesWithScore.includes(selectedType))
-      && entityTypeFilterValues.every((filterType) => typesWithScore.includes(filterType));
-    const typesHaveStatus = selectedTypes.length === 1 // the proposed statuses values depends on the entity type, so we should have only one type selected
-      && !typesWithoutStatus.includes(selectedTypes[0]);
+
+    const checkTypes = (typesList) => selectedTypes.every((type) => typesList.includes(type))
+      && entityTypeFilterValues.every((type) => typesList.includes(type));
+
+    const dynamicOptions = [
```
labo-flg: smart :)

labo-flg: you could also integrate the options depending only on the action type. This way you construct the final array directly (only need to sort)

lndrtrbn (approved):

PR #8702, opened 2024-10-18
[backend] Implement exclusion lists pattern check (#8312)

ValentinBouzinFiligran:
### Proposed changes

* algorithm to converted ipAddr to binary ( ipv4 / v6 )
* algorithm to check pattern validity
* unit tests

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/8312

### Checklist


opencti-platform/opencti-graphql/src/database/cache.ts
```diff
@@ -24,6 +26,12 @@ const STORE_ENTITIES_LINKS: Record<string, string[]> = {
   [ENTITY_TYPE_GROUP]: [ENTITY_TYPE_USER],
 };
 
+const STORE_EXCLUSION_LIST: ExclusionListProperties[] = [];
```
SouadHadjiat: what is this cache ?

SouadHadjiat: I see it's used by the tests, why define it here ?

opencti-platform/opencti-graphql/src/utils/exclusionLists.ts
```diff
@@ -0,0 +1,121 @@
...
+        startProcessingTime = new Date().getTime();
+        await new Promise((resolve) => {
+          setImmediate(resolve);
+        });
+      }
+    }
+  }
+};
+
+export const checkPatternValidity = async (extractedObservableValues: ExtractedObservableValues[]) => {
```
SouadHadjiat: this method is not used, I suggest to remove it, or change it to pass the exclusion lists in param (instead of using getExclusionListsByTypeFromCache which is empty for now)

opencti-platform/opencti-graphql/src/utils/exclusionLists.ts
```diff
@@ -0,0 +1,107 @@
...
+export const convertIpv4ToBinary = (ipv4: string, isRange?: boolean, range?: number) => {
+  const binary = ipv4.split('.').map((ip) => (parseInt(ip, 10).toString(2)).padStart(8, '0')).join('');
+  if (isRange && range) {
+    return binary.slice(0, range);
+  }
+  return binary;
+};
+
+const throwExclusionListError = (value: string, listName: string) => {
+  throw FunctionalError(`Indicator creation failed, this pattern (${value}) is contained on an exclusion list (${listName})`, { value });
```
SouadHadjiat: I don't really like this throw error part, why not just return a boolean true / false if the pattern matches or not an exclusion list ? this would need to be changed after when we'll introduce actions, but we could at start return just a boolean and we'll define the action to take when calling the method (either not create the entity, or throw an exception, or apply an action).

opencti-platform/opencti-graphql/tests/01-unit/utils/exclusionLists-test.ts
```diff
@@ -0,0 +1,274 @@
...
+export const convertAndFillExclusionList = () => {
+  STORE_EXCLUSION_LIST.forEach((item: ExclusionListProperties) => {
+    if (item.type.includes(exclusionListEntityType.IPV4_ADDR) || item.type.includes(exclusionListEntityType.IPV6_ADDR)) {
+      const newList = convertIpAddr(item.list);
+      STORE_EXCLUSION_LIST_BINARY.push({ ...item, list: newList });
+      return;
+    }
+    STORE_EXCLUSION_LIST_BINARY.push(item);
+  });
+};
```
JeremyCloarec: Since this is no longer used in the tests, it can be removed (as well as all files in data/exclusionLists)

opencti-platform/opencti-graphql/src/utils/exclusionLists.ts
JeremyCloarec: I agree. I think it should be refactored as so:
- checkExclusionList and checkIpAddressLists should be changed to return either the id/name of the exclusion list matched, or null if nothing matches
- throwExclusionListError should be changed to a applyExclusionList method, taking an indicator instance and a list name in parameter. For now, this method will throw an exception just the same way, but it will be used in the future when more actions are implemented

SouadHadjiat (approved):

Issue #8683, opened 2024-10-17
No error when updating an indicator with incorrectly formatted pattern


marieflorescontact:
## Environment

1. OS (where OpenCTI server runs): { e.g. Mac OS 10, Windows 10, Ubuntu 16.4, etc. }
2. OpenCTI version: { e.g. OpenCTI 1.0.2 }
3. OpenCTI client: { e.g. frontend or python }
4. Other environment details:

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. update Indicator Pattern with incorrectly formatted pattern. for exemple [ipv4-addr:value = '185.158.114.134' and ipv4-addr:value = '185.158.114.135'] instead of [ipv4-addr:value = '185.158.114.134' AND ipv4-addr:value = '185.158.114.135']
2. Indicator is updated without any error

## Expected Output
error from backend like we have for Indicator creation
![{F3C22B76-3FCC-4FA2-9BC3-39310A0AB73A}](https://github.com/user-attachments/assets/3721029a-6368-4ed8-82fa-b1d0517ad4f1)

## Actual Output
Indicator is updated without any error

![{54C826DC-9D01-4795-AC21-1A2044FF4DCE}](https://github.com/user-attachments/assets/0a90def9-4366-48e4-9e2c-a6596f53c7d8)

![{0E329875-289A-4204-B3CB-D3823E6D40E5}](https://github.com/user-attachments/assets/5da2827d-a6a9-498c-a960-1fbd266da61e)

nino-filigran:
@marieflorescontact I'm not able to reproduce, even on testing. Which env did you use?

marieflorescontact:
on testing:
https://testing.octi.staging.filigran.io/dashboard/observations/indicators/177b399a-1aa7-4f18-86a9-ed522909d469

nino-filigran:
Happens also on pre-release. Only at the edit however.
The fix consists in applying the same validation than at IOC creation!

PR #8717, opened 2024-10-20
[backend] check indicator pattern syntax when update (#8683)

marieflorescontact:
### Proposed changes

* check indicator pattern syntax when update as it is done for Indicator creation
* add test

### Related issues

* #8683 

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/tests/02-integration/02-resolvers/indicator-test.ts
+++ b/opencti-platform/opencti-graphql/tests/02-integration/02-resolvers/indicator-test.ts
@@ -1,6 +1,6 @@
 import { expect, it, describe } from 'vitest';
 import gql from 'graphql-tag';
-import { queryAsAdmin } from '../../utils/testQuery';
+import { adminQuery, queryAsAdmin } from '../../utils/testQuery';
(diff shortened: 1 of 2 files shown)


opencti-platform/opencti-graphql/tests/02-integration/02-resolvers/indicator-test.ts
```diff
@@ -1,6 +1,6 @@
 import { expect, it, describe } from 'vitest';
 import gql from 'graphql-tag';
-import { queryAsAdmin } from '../../utils/testQuery';
+import { adminQuery, queryAsAdmin } from '../../utils/testQuery';
```
aHenryJard, line 3: wondering: is it because queryAsAdmin was not working here ?

marieflorescontact: I don't know... I had so many problems with `queryAsAdmin` which is just a test method which does not take the context into account that I prefer use `adminQuery`

aHenryJard (approved):

ValentinBouzinFiligran (approved):

PR #8738, opened 2024-10-22
[backend] Attributes can't have same label and different names in schema (#8726)

### LINKED ISSUE #8726: [Dashboard] Two "name" filter
## Description

Two "Name" filter are available. They are different (i.e. they don't display the same result)

image

## Environment
OCTI 6.3.6

## Reproducible Steps

1. Create a "Number" widget > Entities perspective > Look for "Name" filter

Archidoit:
### Proposed changes
- Attributes can't have same label and different names check at attribute registration
- Rename label of attributes if an other attribute with same label and different name exists
- Avoid issues like having filters with same label in a list

### Related issues
https://github.com/OpenCTI-Platform/opencti/issues/8726


opencti-platform/opencti-front/lang/back/fr.json
```diff
@@ -715,9 +736,11 @@
   "Window title": "Titre de la fenêtre",
   "Windows registry values": "Valeurs du registre Windows",
   "Work": "Travail",
+  "Work status": "Statut professionnel",
```
JeremyCloarec: Maybe change to 'Statut du travail' ?

JeremyCloarec (approved):
There is just one change to do in translations but I'll approve it already so you can merge when change is done without having to review it again!

Archidoit:
Thanks for the review @JeremyCloarec :) I'm gonna do the modification and merge afterwards !

opencti-platform/opencti-front/lang/back/fr.json
Archidoit: done

PR #8741, opened 2024-10-22
[backend/frontend] Danger zone: Add config in platform settings (#8284)

marieflorescontact:
### Proposed changes

*  Add platform_protected_sensitive_config in `Default.json`
*  refactor `useSensitiveModifications` hook
* some UI alignment left, need to be done in `Settings > Customiztion > Rules `

### Related issues

* #8284 

### Checklist


opencti-platform/opencti-graphql/config/default.json
```diff
@@ -124,6 +124,33 @@
     "next_days": 30,
     "created_by_based": false
   },
+  "protected_sensitive_config": {
+    "enabled": false,
```
JeremyCloarec, line 130: Is it intended to default to false? Shouldn't it be enabled by default and disabled when the users know what they are doing?

opencti-platform/opencti-front/src/utils/hooks/useSensitiveModifications.ts
```diff
@@ -3,47 +3,35 @@ import useHelper from './useHelper';
...
+  const isSensitiveModificationEnabled = isFeatureEnable(PROTECT_SENSITIVE_CHANGES_FF) && isSensitiveConfigEnabled;
   if (!isSensitiveModificationEnabled) {
-    isAllowed = true;
+    return {
+      isAllowed: true,
+      isSensitive: false,
+      isSensitiveModificationEnabled,
+    };
+  }
+  const isAllowed = me.can_manage_sensitive_config ?? true;
```
JeremyCloarec, line 22: It probable should default to false if no configuration is found I think?

opencti-platform/opencti-graphql/config/default.json
marieflorescontact: it should be disable by default

opencti-platform/opencti-front/src/utils/hooks/useSensitiveModifications.ts
marieflorescontact: the default value to  true is a decison we made because we don't want to block users. we want to protect zones only if the config is enabled.

aHenryJard: for backward compatibility on platform, if no config then it's allowed (currently there is no protection) = same behaviors as previously.

opencti-platform/opencti-front/src/private/components/common/entreprise_edition/EnterpriseEdition.tsx
```diff
@@ -25,7 +25,7 @@ import type { Theme } from '../../../../components/Theme';
 const EnterpriseEdition = ({ message, feature }: { message?: string, feature?: string }) => {
   const theme = useTheme<Theme>();
 
-  const { isSensitiveModificationEnabled, isAllowed } = useSensitiveModifications();
+  const { isSensitive, isAllowed } = useSensitiveModifications('ce_ee_toggle');
```
JeremyCloarec, line 28: EntrepriseEditionButton at line 47 need to be updated to be disabled={!isAllowed & isSensitive}

opencti-platform/opencti-front/src/utils/hooks/useSensitiveModifications.ts
```diff
@@ -3,47 +3,35 @@ import useHelper from './useHelper';
...
+  let isSensitive = isSensitiveConfigEnabled;
+
+  if (type) {
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    const config = sensitiveConfig[type];
+    const protectedIds = config.protected_ids ?? [];
+    isSensitive = config.enabled && (!id || protectedIds.includes(id));
   }
```
JeremyCloarec: This part could be refactored to check if type parameter is equal to 'ce_ee_toggle', 'file_indexing' etc...

JeremyCloarec (changes requested):
In addition to comments left on changes, some changes are needed:
- Change DangerZoneBlock component to use isSensitive from hook (right now, file indexing, rules etc... are always disabled for not allowed users, independant of the config of these zones in the default.json config)
- Refactor useSensitiveModifcation hook => isSensitiveModificationEnabled can be removed
- During my tests, all roles were shown to have the capability to edit sensitive configs. Is it a bug?

opencti-platform/opencti-front/src/private/components/settings/Policies.tsx
```diff
@@ -213,7 +214,7 @@ const PoliciesComponent: FunctionComponent<PoliciesComponentProps> = ({
                           <EETooltip>
                             <ObjectOrganizationField
                               name="platform_organization"
-                              disabled={disabled || !isEnterpriseEdition || (isSensitiveModificationEnabled && !isAllowed)}
+                              disabled={disabled || !isEnterpriseEdition || (isSensitive && !isAllowed)}
```
JeremyCloarec: Since the DangerZoneBlock already provides the logic of disabling, you can simplify it to                               disabled={disabled || !isEnterpriseEdition}

opencti-platform/opencti-front/src/private/components/settings/Settings.jsx
```diff
@@ -428,28 +428,31 @@ const Settings = () => {
...
                                 variant="outlined"
-                                color={isSensitiveModificationEnabled ? 'dangerZone' : 'primary'}
+                                color={isSensitive ? 'dangerZone' : 'primary'}
                                 onClick={() => setOpenEEChanges(true)}
-                                disabled={disabled}
-                                style={isSensitiveModificationEnabled ? {
-                                  color: theme.palette.dangerZone.text.primary,
-                                  borderColor: theme.palette.dangerZone.main,
-                                } : undefined}
+                                disabled={isSensitive && !isAllowed}
```
JeremyCloarec: I would probably revert it back to use the disabled information given by the DangerZoneBlock

JeremyCloarec (approved):
Tested OK and code looks OK! Some minor changes could be done but it's not essential if there is no time to do it

opencti-platform/opencti-front/src/private/components/settings/Policies.tsx
marieflorescontact: done

opencti-platform/opencti-front/src/private/components/settings/Settings.jsx
marieflorescontact: done

Issue #7620, opened 2024-07-06
Domain observable with underscore is incorrectly rejected


samczsun:
## Description

This regex does not accept underscores in domains, but underscores are permitted in domains.

https://github.com/OpenCTI-Platform/opencti/blob/542b5cd9504279bc5f5c63f98725a392a35c3dd5/opencti-platform/opencti-graphql/src/utils/syntax.js#L138

## Environment

1. OS (where OpenCTI server runs): Ubuntu
2. OpenCTI version: 6.1.11
3. OpenCTI client: Frontend
4. Other environment details:

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Create domain observable with domain that contains an underscore, such as `example_domain.godaddysites.com`

## Expected Output

An observable is created

## Actual Output

An error occurs

Jipegien:
RFCs forbid use of underscore in domain names BUT some registrars allow it in subdomains. 
We need to modify the regexp to accept underscores in subdomains, but not the domain itself and the TLD

romain-filigran:
I reopened this issue as I get some ingestion error when trying to create domain observable with value: "test._mysubdomain.mydomain.com"

SarahBocognano:
@romain-filigran It's because you put the dot before the underscore. Do we want it possible ? If yes I can of course modify the regex to accept a dot followed by an underscore

aHenryJard:
Reopening: This feature has been reverted du to performance issues, we need to add a test coverage on regeprx performance issue.

PR #8768, opened 2024-10-24
[backend] Domain observable checker is incorrect (#7620)

SarahBocognano:
### Proposed changes

* Change the regex to also accept dots on subdomain 

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/7620


SouadHadjiat:
unit tests are failing on this PR, also could you add corresponding unit test for the usecase we are fixing ?

marieflorescontact (commented):
Tested ok with corrects and incorrects domains and subdomains.
the only case that seems to be not covered by test is a very long subdomain:
"very0long0subdomain0that0exceeds0the0sixty0three0character0limit.mydomain.com"
subdomain are limited to 63 characters to be valid

opencti-platform/opencti-graphql/tests/01-unit/utils/syntax-test.ts
```diff
@@ -12,6 +12,8 @@ describe('Regex Pattern Tests', () => {
     expect('example.com').toMatch(domainChecker);
     expect('sub.example.co.uk').toMatch(domainChecker);
     expect('løveskateboards.com').toMatch(domainChecker);
+    expect('test._mysubdomain.mydomain.com').toMatch(domainChecker);
+    expect('test_mysubdomain.domain.io').toMatch(domainChecker);
     expect('invalid_domain.12_3').not.toMatch(domainChecker);
```
Archidoit, line 17: maybe add a test that it's not working if there is a dot in the main name and no dots (exemple: invalid_domain)

Archidoit:
I just suggested to add a little test case. Since everything is ok, I'm gonna approve the PR so you can merge it after adding the test.

Archidoit (approved):

Issue #8575, opened 2024-10-02
Report export to STIX2 brings start_time == stop_time for relations


andurin:
## Description

While finished a report in OCTI I wanted to export the report to JSON(STIX) to import that manually into a MISP Instance.
MISP importer complained about start_time == stop_time values which could be found in the generated relations.

## Environment

1. OS (where OpenCTI server runs): Debian 11
2. OpenCTI version: OpenCTI 6.3.3
3. OpenCTI client: None used
4. Other environment details:

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Browser: /dashboard/analyses/reports
2. Choose a report which already has some relations
3. Export that report:
  - Export format: application/json
  - Export Type: Full Export
4. Download the report 
5. Open in favorite editor
6. Search for "stop_time", find a relation object and compare with start_time

## Expected Output

start_time != stop_time

## Actual Output

start_time == stop_time

## Additional information

Based on this it may be easy to just add an extra second or either microsecond to the stop_time.

https://github.com/OpenCTI-Platform/connectors/commit/a7d6a3a642685d8b50b937336e99d1a2faadd49a

nino-filigran:
@andurin indeed, I noticed this behavior but not sure it's an export problem however. 
Indeed, by default, start time and stop time are populated with the same values, which then causes the bug when exporting and importing to MISP.

nino-filigran:
The fix is therefore for us to have at least 1s of difference between two dates on our side, when creating a relation.

andurin:
Thank you for having a look into this.

Yes, it is MISP complaining about this but that's ok since also the Stix Standard says something like "if you propagate start_time and end_time you must ensure that end_time is after start_time."

What do you think - can I hope for a fix in the next Version?



03.10.2024 09:29:43 nino-filigran ***@***.***>:

nino-filigran:
@andurin I'm not sure it will be in the next release:   we have quite some high-priority bugs that also require our attention. We'll prioritize this bug in the top part of the bucket, but can't promise anything regarding when it will be fixed. 

The "milestone" will be updated once the bug will be fixed, allowing you to know when the fix will be released.

marieflorescontact:
@nino-filigran :  `start_time` and `stop_time` have same value because relationship creation form is sent with default values.

![Capture d'écran 2024-10-23 141406](https://github.com/user-attachments/assets/ad93937e-c875-4099-af56-bf79c869f55d)

[...]

PR #8782, opened 2024-10-24
[frontend] Add defaultStartTime and stopTime for StixCoreRelationship creation (#8575)

marieflorescontact:
### Proposed changes

* add default value for relation ship creation from knowledge graph and investigation graph
* remove 1 minute to make sure startime and stoptime are different

### Related issues

* #https://github.com/OpenCTI-Platform/opencti/issues/8575

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/analyses/groupings/GroupingKnowledgeGraphBar.jsx
+++ b/opencti-platform/opencti-front/src/private/components/analyses/groupings/GroupingKnowledgeGraphBar.jsx
@@ -46,7 +46,7 @@ import CommitMessage from '../../common/form/CommitMessage';
 import StixNestedRefRelationshipCreationFromKnowledgeGraph from '../../common/stix_nested_ref_relationships/StixNestedRefRelationshipCreationFromKnowledgeGraph';
 import inject18n from '../../../../components/i18n';
 import StixCoreRelationshipCreation from '../../common/stix_core_relationships/StixCoreRelationshipCreation';
-import { dateFormat, now, validStartTimeForRelationCreation } from '../../../../utils/Time';
+import { dateFormat, minutesBefore, now } from '../../../../utils/Time';
 import { truncate } from '../../../../utils/String';
 import StixCoreRelationshipEdition from '../../common/stix_core_relationships/StixCoreRelationshipEdition';
 import StixDomainObjectEdition from '../../common/stix_domain_objects/StixDomainObjectEdition';
@@ -456,6 +456,9 @@ class GroupingKnowledgeGraphBar extends Component {
         : [selectedNodes[0]];
     }
     const stixCoreObjectOrRelationshipId = (selectedNodes[0]?.id ?? null) || (selectedLinks[0]?.id ?? null);
+
+    const defaultTime = now();
+
     return (
       <UserContext.Consumer>
         {({ bannerSettings }) => (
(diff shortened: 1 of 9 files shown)


opencti-platform/opencti-front/src/utils/Time.js
```diff
@@ -82,6 +82,12 @@ export const dateFormat = (data, specificFormat = null) => {
     : '';
 };
 
+export const validStartTimeForRelationCreation = (time) => {
+  const date = new Date(time);
+  date.setMinutes(date.getMinutes() - 1);
```
lndrtrbn: In the issue Nino ask for one second less instead of one minute, was there any changes?

opencti-platform/opencti-front/src/utils/Time.js
```diff
@@ -82,6 +82,12 @@ export const dateFormat = (data, specificFormat = null) => {
     : '';
 };
 
+export const validStartTimeForRelationCreation = (time) => {
```
lndrtrbn: It would be better to have a generic function allowing to subtract `x` seconds from a date instead of a very specific function. You have an example in this same file with function `minutesBefore`

opencti-platform/opencti-front/src/private/components/analyses/groupings/GroupingKnowledgeGraphBar.jsx
```diff
@@ -986,10 +986,10 @@ class GroupingKnowledgeGraphBar extends Component {
                         fromObjects={relationFromObjects}
                         toObjects={relationToObjects}
                         startTime={
-                          lastLinkFirstSeen || dateFormat(grouping.published)
+                          lastLinkFirstSeen || validStartTimeForRelationCreation(now())
                         }
                         stopTime={
-                          lastLinkLastSeen || dateFormat(grouping.published)
+                          lastLinkLastSeen || now()
```
lndrtrbn: Calling 2 times function `now` results having two different dates. Yes the diff would be only some milliseconds. But I think it's better to have only one call to `now` to be sure we have the same date. (Same remark for other files)

opencti-platform/opencti-front/src/utils/Time.js
marieflorescontact: because second was not taken into account

marieflorescontact: I checked with @nino-filigran . 1 minute is ok.
we don't display second tht is why i used minute.

opencti-platform/opencti-front/src/utils/Time.js
marieflorescontact: As we could remove 1 minute (instead of 1 seconde as I mentioned above), I used `minutesBefore` that was already existing and I deleted `validStartTimeForRelationCreation `

lndrtrbn:
Missing default dates when creating relationships from:
- Malware analyses / Knowledge
- Incidents / Knowledge
- Observables / Knowledge / Relations
- Artifacts / Knowledge / Relations
- Indicators / Knowledge
- Infrastructures / Knowledge
- Well all entities that use the same component for create relationships than the ones above

Default start and stop are identical in:
- Observables / Knowledge / Nested objects
- Artifacts / Knowledge / Nested objects

marieflorescontact:
> Missing default dates when creating relationships from:
> 
> * Malware analyses / Knowledge
> * Incidents / Knowledge
> * Observables / Knowledge / Relations
> * Artifacts / Knowledge / Relations
> * Indicators / Knowledge
> * Infrastructures / Knowledge
> * Well all entities that use the same component for create relationships than the ones above
> 
> Default start and stop are identical in:
> 
> * Observables / Knowledge / Nested objects
> * Artifacts / Knowledge / Nested objects

ok you are right, I checked with product team, even if it is a bigger scope than the initial issue, we want to add default start time and stop time everywhere relation creation is allowed.

marieflorescontact:
PR status: all comments have been taken into account.
waiting for 2nd round of review or approval .

lndrtrbn (approved):

PR #8786, opened 2024-10-25
[backend] deduplication check is now done with SYSTEM user (#8740)

### LINKED ISSUE #8740: Data is not deduplicated when an entity under restriction (marking) is updated
## Description

It appears that data deduplication is not correctly handled when we try to update (change the name) of an allowed entity with the name of a restricted entity (under marking restriction).

## Environment
2. OpenCTI version: 6.3.6

## Reproducible Steps

1. As userA (with all allowed markings), create a Threat Actor with name = "myThreatActor" and add marking = TLP:RED
2. As userB (with only TLP:CLEAR marking authorized), create a Threat Actor with name = "myThreatActor" -> popup that indicate that the entity already exist -> OK
3. As userB (with only TLP:CLEAR marking authorized), create a Threat Actor with name = "testThreatActor" --> entity is correctly created -> OK
4. As user B (with only TLP:CLEAR marking authorized), update the threat actor ‘testThreatActor’ and replace its name with ‘myThreatActor’ -->the update is allowed, resulting in duplicate entities in the platform that have the same name, the same stix_id but not the same OCTI identifier.

Same behavior with organisation sharing/segregation.

image

JeremyCloarec:
### Proposed changes

* deduplication check was done with user, causing duplicated entities not accessible to the user to not be taken into account. Deduplication check is now done with SYSTEM user to prevent that

### Related issues

* #8740

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/database/middleware.js
+++ b/opencti-platform/opencti-graphql/src/database/middleware.js
@@ -1919,7 +1919,7 @@ export const updateAttributeMetaResolved = async (context, user, initial, inputs
     if (aliasedInputs.length > 0) {
       const aliases = R.uniq(aliasedInputs.map((a) => a.value).flat().filter((a) => isNotEmptyField(a)).map((a) => a.trim()));
       const aliasesIds = generateAliasesId(aliases, initial);
-      const existingEntities = await internalFindByIds(context, user, aliasesIds, { type: initial.entity_type });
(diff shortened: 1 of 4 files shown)


SouadHadjiat:
Do you think it would be possible to write an integration test corresponding to the issue usecase ? to make sure this issue doesn't happen again ?

SouadHadjiat (approved):

PR #8789, opened 2024-10-25
[backend] Make pattern upsertable for rename platform sync

### LINKED ISSUE #8611: Indicator pattern and standard id are not updated through stream sync
## Description

While working on another topic, we added a test backend side that updates an indicator pattern. 
This causes issue in raw sync tests.
see repro in PR #8610

Error is 

`{"category":"APP","errors":[{"attributes":{"genre":"BUSINESS","http_status":400},"message":"Cannot delete the stix element, indicator--ac07b543-765f-57b4-868d-6b2be61629a8 cannot be found.","name":"FUNCTIONAL_ERROR","stack":"GraphQLError: Cannot delete the stix element, indicator--ac07b543-765f-57b4-868d-6b2be61629a8 cannot be found.\n    at error (/tmp/raw-start-platform/opencti-graphql/src/config/errors.js:7:10)\n    at FunctionalError (/tmp/raw-start-platform/opencti-graphql/src/config/errors.js:94:50)\n    at stixDelete (/tmp/raw-start-platform/opencti-graphql/src/domain/stix.js:51:9)\n    at processTicksAndRejections (node:internal/process/task_queues:95:5)"}],"inner_relation_creation":0,"level":"error","message":"Cannot delete the stix element, indicator--ac07b543-765f-57b4-868d-6b2be61629a8 cannot be found.","operation":"Unspecified","query_attributes":[[{"arguments":[],"name":"delete"}]],"size":56,"source":"backend","time":6,"timestamp":"2024-10-04T13:30:33.689Z","type":"WRITE_ERROR","user":{"group_ids":["5fb5f176-131f-4db5-9515-34f71e9342ce"],"ip":"::ffff:192.168.192.12","organization_ids":[],"socket":"query","user_id":"88ec0c6a-13ce-5e39-b486-354fe4a7084f","user_metadata":{}},"version":"6.3.5"}`

## Environment

from master@6.3.5

## details & repro

I've reproduced this issue with 2 opencti platforms synchronizing through live streams.

* Open a stream in the Source, for indicators created by myself only (to limit the throughput) 
* Listen to this source in the Destination
* Create an indicator in the source, a domain name with stix pattern `[domain-name:value = 'www.some.test']` for instance
* The indicator appears in the Destination ✅ 
* Update the indicator **name** in the Source
* The indicator name is updated in the Destination ✅ 
* Update the indicator **pattern** in the Source (=> this causes a stix id change)
* The indicator pattern and stix id are NOT updated in the Destination ❌ 
* Delete the indicator in the Source 
* Error in the Destination `"GraphQLError: Cannot delete the stix element, indicator--ac07b543-765f-57b4-868d-6b2be61629a8 cannot be found.` ❌

labo-flg:
### Proposed changes

* make the stix pattern of the observables upsertable, to solve change in stixid not properly send into stream sync

### Related issues
Closes #8611

See this issue for repro case.

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/modules/indicator/indicator.ts
+++ b/opencti-platform/opencti-graphql/src/modules/indicator/indicator.ts
@@ -23,8 +23,8 @@ const INDICATOR_DEFINITION: ModuleDefinition<StoreEntityIndicator, StixIndicator
   attributes: [
     { name: 'name', label: 'Name', type: 'string', format: 'short', mandatoryType: 'external', editDefault: true, multiple: false, upsert: true, isFilterable: true },
     { name: 'description', label: 'Description', type: 'string', format: 'text', mandatoryType: 'customizable', editDefault: true, multiple: false, upsert: true, isFilterable: true },
-    { name: 'pattern_type', label: 'Pattern type', type: 'string', format: 'vocabulary', vocabularyCategory: 'pattern_type_ov', mandatoryType: 'external', editDefault: true, multiple: false, upsert: false, isFilterable: true },
-    { name: 'pattern_version', label: 'Pattern version', type: 'string', format: 'short', mandatoryType: 'no', editDefault: false, multiple: false, upsert: false, isFilterable: false },
+    { name: 'pattern_type', label: 'Pattern type', type: 'string', format: 'vocabulary', vocabularyCategory: 'pattern_type_ov', mandatoryType: 'external', editDefault: true, multiple: false, upsert: true, isFilterable: true },
+    { name: 'pattern_version', label: 'Pattern version', type: 'string', format: 'short', mandatoryType: 'no', editDefault: false, multiple: false, upsert: true, isFilterable: false },
     { name: 'pattern', label: 'Pattern', type: 'string', format: 'short', mandatoryType: 'external', editDefault: true, multiple: false, upsert: true, isFilterable: true },
     { name: 'indicator_types', label: 'Indicator types', type: 'string', format: 'vocabulary', vocabularyCategory: 'indicator_type_ov', mandatoryType: 'customizable', editDefault: true, multiple: true, upsert: true, isFilterable: true },
     { name: 'valid_from', label: 'Valid from', type: 'date', mandatoryType: 'customizable', editDefault: true, multiple: false, upsert: true, isFilterable: true },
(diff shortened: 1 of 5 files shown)


opencti-platform/opencti-graphql/tests/02-integration/02-resolvers/indicator-test.ts
```diff
@@ -243,6 +245,16 @@ describe('Indicator resolver standard behavior', () => {
     });
     expect(queryResult.data?.indicatorFieldPatch.name).toEqual('Indicator - test');
   });
+  it('should update indicator pattern, causing change in stix id', async () => {
```
SouadHadjiat, line 248: how this test is related to pattern upsert ? updating the pattern was already changing the standard_id, no ?

opencti-platform/opencti-graphql/src/modules/indicator/indicator.ts
```diff
@@ -25,7 +25,7 @@ const INDICATOR_DEFINITION: ModuleDefinition<StoreEntityIndicator, StixIndicator
     { name: 'description', label: 'Description', type: 'string', format: 'text', mandatoryType: 'customizable', editDefault: true, multiple: false, upsert: true, isFilterable: true },
     { name: 'pattern_type', label: 'Pattern type', type: 'string', format: 'vocabulary', vocabularyCategory: 'pattern_type_ov', mandatoryType: 'external', editDefault: true, multiple: false, upsert: false, isFilterable: true },
```
SouadHadjiat: pattern_type & pattern_version don't need to be upsertable ?

opencti-platform/opencti-graphql/tests/02-integration/02-resolvers/indicator-test.ts
SouadHadjiat: or does it impact the next tests about sync maybe ?

SouadHadjiat:
tested locally : indicator is well updated during sync ✅

opencti-platform/opencti-graphql/src/modules/indicator/indicator.ts
labo-flg: that is a good question. If we change the pattern we need to be able to change the type and version, otherwise you would change from stix to yara and keeping type as stix.... it's completly broken then.

I'll change the upsert too for these fields

opencti-platform/opencti-graphql/tests/02-integration/02-resolvers/indicator-test.ts
labo-flg: yes it's about sync issues. 
This is not upsert test here but it will cause an upsert during sync tests later.

The standard id changed on one side but not on the other.
See the repro case in the issue linked.

SouadHadjiat (approved):

PR #8802, opened 2024-10-28
[frontend] Fix delete button behaviour

marieflorescontact:
### Proposed changes

* Fix `Delete` button when FF activated => If FF off we see the `Update` button, `Delete` button is inside the `Update` component. If FF is on we dont see any `Delete` button but we see `Popover` component
* Fix `Delete` button aligment when enforce ref are activated (FF need to be off, enforce ref activated for entites sucha as "Analyses / Case and Location)

### Related issues

* this is related to PRs merged in order to remove popover component for `Analyses`, `Cases`,` Location`
* this is under` FF`, to see the problem you need to remove the `FF` (feature activated)
* ` "disabled_dev_features": ["FAB_REPLACEMENT"] => FF activated, feature deactivated`
* `"disabled_dev_features": [] => FF deactivated, feature activated`

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/analyses/reports/ReportPopoverDeletion.tsx
+++ b/opencti-platform/opencti-front/src/private/components/analyses/reports/ReportPopoverDeletion.tsx
@@ -1,0 +1,130 @@
+import { graphql } from 'react-relay';
+import DialogContent from '@mui/material/DialogContent';
+import Alert from '@mui/material/Alert';
+import AlertTitle from '@mui/material/AlertTitle';
+import FormGroup from '@mui/material/FormGroup';
+import FormControlLabel from '@mui/material/FormControlLabel';
+import Checkbox from '@mui/material/Checkbox';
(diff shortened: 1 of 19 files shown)


opencti-platform/opencti-front/src/private/components/analyses/reports/ReportPopoverDeletion.tsx
```diff
@@ -0,0 +1,130 @@
...
+      open={displayDelete}
+      TransitionComponent={Transition}
+      PaperProps={{ elevation: 1 }}
+      onClose={handleCloseDelete}
+    >
+      <DialogContent>
+        <DialogContentText>
+          {t_i18n('Do you want to delete this report?')}
+        </DialogContentText>
+        <QueryRenderer
```
lndrtrbn, line 80: QueryRenderer should not be used in new components as it is an old way of calling queries for class components. use hooks instead

lndrtrbn (approved):

PR #8811, opened 2024-10-28
[backend/frontend] Fix playbook organization sharing schema (#8767)

### LINKED ISSUE #8767: Having a large number of Organizations slows down the playbooks
## Description

When the platform ingests many many organizations, for instance through a feed like ransomware.live, it slows down the platform performances.

It might be due to the way we handle organizations in the platform, according to organization sharing (caching, playbooks on orga sharing).

## Environment

OCTI 6.3.6

SouadHadjiat:
### Proposed changes

* don't return anymore all organizations list and use ObjectOrganizationField component
* implement a resolver to parse organizations that have been already saved with only the id (we need the label)
* query organizations by id instead of using the cache

### Related issues

* #8767 

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/data/playbooks/PlaybookAddComponents.jsx
+++ b/opencti-platform/opencti-front/src/private/components/data/playbooks/PlaybookAddComponents.jsx
@@ -20,6 +20,7 @@ import Box from '@mui/material/Box';
 import Alert from '@mui/material/Alert';
 import Drawer from '../../common/drawer/Drawer';
 import ObjectMembersField from '../../common/form/ObjectMembersField';
+import ObjectOrganizationField from '../../common/form/ObjectOrganizationField';
 import CreatedByField from '../../common/form/CreatedByField';
 import Filters from '../../common/lists/Filters';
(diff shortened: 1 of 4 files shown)


opencti-platform/opencti-front/src/private/components/data/playbooks/PlaybookAddComponents.jsx
```diff
@@ -458,6 +459,18 @@ const PlaybookAddComponentsContent = ({
                       />
                     );
                   }
+                  if (k === 'organizations') {
+                    return (
+                      <ObjectOrganizationField
+                        key={k}
+                        name="organizations"
+                        style={{ marginTop: 20, width: '100%' }}
+                        label={'Target organizations'}
```
aHenryJard, line 468: There is no i18n needed here ?

SouadHadjiat: nothing is translated in playbook components for now, so I didn't work on it in this fix. you can see just above the same thing for authorized members field with label "Targets"

aHenryJard (approved):
Tested playbook with orga sharing locally, everything works as expected.

PR #8844, opened 2024-10-31
[backend/frontend] Allow to reset state of Taxii feed and csv Feed, and reset taxii state on date update (#8463)

### LINKED ISSUE #8463: Taxii feed: when updating feed to change starting date, the cursor should be reset
## Description

Date added after and next cursor are during ingestion set by the taxii server, but when the date is changed manually on taxii feed configuration the cursor should be reset.
In current implementation the cursor is reset only when all data is consume (more = false in server response).

Use case is: for some reason if users wants to restart from another date than the current one that is set up, dependeing on taxii server implementation behavior can ignore date and take cursor.

## Environment

4. Other environment details:

## Reproducible Steps

1. { e.g. Run ... }
2. { e.g. Click ... }
3. { e.g. Error ... }

## Expected Output

Update taxii feed date, cursor is empty.

## Actual Output

Update taxii feed date, cursor is not reset.

aHenryJard:
### Proposed changes

Allow to reset state of an ingestion for CSV Feed and Taxii Feed. RSS Feed are not concerned since there is only a date.

* Add "Reset state" in menu items of csv feed and taxii feed
* Add graphQL APIs to reset state
* Reset Taxii Feed state when the date from is updated

Improvements:
- Refactor IngestionTaxiiPopover.jsx => tsx and IngestionTaxiiLine.jsx => tsx
- add docker image for local smtp server for development
- removed useless field date_cursor for csv feed in UI

 
### Related issues

* closes #8463 

### Checklist

### Further comments

image


opencti-platform/opencti-graphql/config/test.json
```diff
@@ -99,6 +99,9 @@
     "enabled": true,
     "interval": 6000000
   },
+  "ingestion_manager": {
+    "enabled": false
```
aHenryJard, line 103: Because for testing pupose it's better to call it once, than wait until manager runs. I think it will avoid flakes.

JeremyCloarec:
Not related to your changes, but couldn't we imagine adding a reset to RSS feeds also? Since they also have a current state date

opencti-platform/opencti-graphql/src/modules/ingestion/ingestion-taxii-domain.ts
```diff
@@ -79,6 +80,11 @@ export const ingestionEditField = async (context: AuthContext, user: AuthUser, i
     is_running: element.ingestion_running ?? false,
     connector_user_id: element.user_id
   });
+
+  if (input.some(((editInput) => editInput.key === 'added_after_start'))) {
+    await patchTaxiiIngestion(context, user, ingestionId, { current_state_cursor: undefined });
+  }
```
JeremyCloarec: Couldn't this patch be done in the same operation as the updateAttribute above? By including the current state cursor change in the input

opencti-platform/opencti-graphql/src/modules/ingestion/ingestion-csv-domain.ts
```diff
@@ -93,6 +93,20 @@ export const ingestionCsvEditField = async (context: AuthContext, user: AuthUser
...
+    user,
+    event_type: 'mutation',
+    event_scope: 'update',
+    event_access: 'administration',
+    message: `reset state of csv ingestion ${ingestionUpdated.name}`,
+    context_data: { id: ingestionId, entity_type: ENTITY_TYPE_INGESTION_CSV, input: ingestionUpdated }
+  });
+  return notify(BUS_TOPICS[ABSTRACT_INTERNAL_OBJECT].EDIT_TOPIC, ingestionUpdated, user);
+};
+
```
JeremyCloarec, line 109: Shouldn't we also reset the state of the csv ingestion when 'added_after_start' is modified in the ingestionCsvEditField ? The same way it is done in taxii feeds

JeremyCloarec:
Does the feed state always need to be reset when the date from is updated? Doesn't it only need to be reset when the new date from is lower than the original date from?

opencti-platform/opencti-graphql/src/modules/ingestion/ingestion-csv-domain.ts
aHenryJard: As far as I know added_after_start is totally useless on csv feed, to me it's a copy paste mistake from taxii feeds since it's not used at all in ingestionManager.

aHenryJard:
> Not related to your changes, but couldn't we imagine adding a reset to RSS feeds also? Since they also have a current state date

No strong opinion on thi s one I can add it there for consistency in UI. For RSS feed you just have to update and change the date, that's why I haven't set a reset. Should a reset clear the date or start from the user's input date ? That's also a question.

aHenryJard:
> Does the feed state always need to be reset when the date from is updated? Doesn't it only need to be reset when the new date from is lower than the original date from?

In taxii feed the cursor can be anything, and the date is send as request parameter. So when date change, it's not possible to guess the cursor state, also depending on taxii server implementations there is a priority between date and cursor and once again it's impossible to know which one is taken in priority. So when date change it should clear cursor anyway.

opencti-platform/opencti-graphql/src/modules/ingestion/ingestion-taxii-domain.ts
aHenryJard: thanks looking

JeremyCloarec:
> > Not related to your changes, but couldn't we imagine adding a reset to RSS feeds also? Since they also have a current state date
> 
> No strong opinion on thi s one I can add it there for consistency in UI. For RSS feed you just have to update and change the date, that's why I haven't set a reset. Should a reset clear the date or start from the user's input date ? That's also a question.

Yes not sure exactly either what it should do. I'd go for clearing the date when reseting if I had to chose myself. But I think it could be great for consistency between all ingestion feed pages to have this reset option on all of them

opencti-platform/opencti-graphql/src/modules/ingestion/ingestion-csv-domain.ts
JeremyCloarec: You're right, I didn't dive that deep into the code so I thought it was useful in some way. 
Not related to your PR,  but when looking more at the code, I saw that the current_state_date is also completely useless now, is that intended? The user can still input an 'Import from date' date when creating a CSV feed, but since it is not used at all, maybe we should create a different issue to completely remove this field/input?

aHenryJard: I just remove from UI for now to avoid API breaking changes.

aHenryJard: I removed from UI because I think it's misleading, I haven't change the API to avoid breaking changes.

opencti-platform/opencti-graphql/src/modules/ingestion/ingestion-taxii-domain.ts
aHenryJard: done.

CelineSebe:
After reseting the CSV feed and stopping, starting the ingestion, the current state remains empty. Is it normal?
If I duplicate this specific CSV feed, after starting, the hash is displayed.
image

opencti-platform/opencti-graphql/src/modules/ingestion/ingestion-csv-domain.ts
```diff
@@ -93,6 +93,20 @@ export const ingestionCsvEditField = async (context: AuthContext, user: AuthUser
   return notify(BUS_TOPICS[ABSTRACT_INTERNAL_OBJECT].EDIT_TOPIC, element, user);
 };
 
+export const ingestionCsvResetState = async (context: AuthContext, user: AuthUser, ingestionId: string) => {
+  await patchCsvIngestion(context, user, ingestionId, { current_state_hash: '' });
+  const ingestionUpdated = await findById(context, user, ingestionId);
```
JeremyCloarec, line 98: You can probably save a call to elastic by using the return value of patchCsvIngestion instead of calling additional findById

opencti-platform/opencti-graphql/src/modules/ingestion/ingestion-taxii-domain.ts
```diff
@@ -103,3 +114,18 @@ export const ingestionDelete = async (context: AuthContext, user: AuthUser, inge
   });
   return ingestionId;
 };
+
+export const ingestionTaxiiResetState = async (context: AuthContext, user: AuthUser, ingestionId: string) => {
+  await patchTaxiiIngestion(context, user, ingestionId, { current_state_cursor: undefined });
+  const ingestionUpdated = await findById(context, user, ingestionId);
```
JeremyCloarec, line 120: You can probably save a call to elastic by using the return value of patchTaxiiIngestion instead of calling additional findById

JeremyCloarec (approved):
Apart from my last two comments for minor improvements that are not mandatory, everything looks good to me!

opencti-platform/opencti-graphql/src/modules/ingestion/ingestion-csv-domain.ts
aHenryJard: As far as I remember I added findById in purpose because the patch is not returning what I need.

aHenryJard:
> After reseting the CSV feed and stopping, starting the ingestion, the current state remains empty. Is it normal? If I duplicate this specific CSV feed, after starting, the hash is displayed. image

Have you check that the queue is empty ? Next csv ingestion is paused until the queue is empty. I do not reproduce on my side, but I need to empty queue for quick tests.

PR #8903, opened 2024-11-05
[backend] secure organization deletion (#8838)

### LINKED ISSUE #8838: Prevent deletion of the organization that is used as plateforme organization or attibuted to a user
## Description

Prevent deletion of the organization that is used as plateforme organization or attibuted to a user.

## Environment

2. OpenCTI version: { e.g. OpenCTI 1.0.2 } 6.3.7
4. Other environment details:

## Reproducible Steps

scenario 1
1. Set a platform organization (let's say "Org1")
2. Go to Entities > Organization
3. Search and delete "Org1"

scenario 2
1. In a user detail, add an organization (let's say "Org2")
2. Go to Entities > Organization
3. Search and delete "Org2"

## Expected Output

scenario 1 => in security > policy : plateform orga disapear, but in database there is still an org id in settings
scenario 2 => in user detail >  orga disappear and there is a dash instead

## Actual Output

scenario 1 => in security > policy : plateform orga should stay*
scenario 2 => in user detail >  orga should stay

## Additional information

Verify that the rentention policy still works when there is an plateform orga in the list of "old" data.

SouadHadjiat:
### Proposed changes

* add protections to prevent organization deletion if it's a platform organization, if it has an administrator, if it has members
* call organizationDelete in retention manager (however it doesn't delete permanently, it will go in trash => I don't think it's an issue, since organization delete can be quite destructive, so it's better we keep it in trash in case)
* add tests

### Related issues

* #8838 

To test : organization can't be deleted if it is used as platform organization, or it has at least one member (user part of the organization), or has an organization admin (org admin is also a member, but the error message is different). Otherwise the deletion should work. Also error message is different for users depending on their capability : if user has set access, the message will explain the reason (members or orga admin reason).

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/manager/taskManager.js
+++ b/opencti-platform/opencti-graphql/src/manager/taskManager.js
@@ -79,6 +79,7 @@ import { BackgroundTaskScope } from '../generated/graphql';
 import { ENTITY_TYPE_INTERNAL_FILE } from '../schema/internalObject';
 import { deleteFile } from '../database/file-storage';
 import { checkUserIsAdminOnDashboard } from '../modules/publicDashboard/publicDashboard-utils';
+import { organizationDelete } from '../modules/organization/organization-domain';
 import { ENTITY_TYPE_IDENTITY_ORGANIZATION } from '../modules/organization/organization-types';
 
 // Task manager responsible to execute long manual tasks
@@ -237,6 +238,9 @@ const executeDelete = async (context, user, element, scope) => {
   }
   if (scope === BackgroundTaskScope.Import) {
     await deleteFile(context, user, element.id);
+  } else if (element.entity_type === ENTITY_TYPE_IDENTITY_ORGANIZATION) {
+    // call organizationDelete which will ensure protections (for platform organization & members)
+    await organizationDelete(context, user, element.internal_id);
   } else {
     await deleteElementById(context, user, element.internal_id, element.entity_type);
   }
--- a/opencti-platform/opencti-graphql/src/modules/organization/organization-domain.ts
+++ b/opencti-platform/opencti-graphql/src/modules/organization/organization-domain.ts
@@ -1,4 +1,4 @@
-import { createEntity, patchAttribute } from '../../database/middleware';
+import { createEntity, deleteElementById, patchAttribute } from '../../database/middleware';
 import {
   type EntityOptions,
   internalFindByIds,
@@ -10,17 +10,20 @@ import {
 } from '../../database/middleware-loader';
 import { BUS_TOPICS } from '../../config/conf';
 import { notify } from '../../database/redis';
+import { getEntityFromCache } from '../../database/cache';
+import { READ_INDEX_INTERNAL_OBJECTS } from '../../database/utils';
 import { ENTITY_TYPE_IDENTITY_SECTOR } from '../../schema/stixDomainObject';
 import { RELATION_PART_OF } from '../../schema/stixCoreRelationship';
 import { ABSTRACT_STIX_DOMAIN_OBJECT } from '../../schema/general';
 import { RELATION_PARTICIPATE_TO } from '../../schema/internalRelationship';
-import { ENTITY_TYPE_USER } from '../../schema/internalObject';
+import { ENTITY_TYPE_SETTINGS, ENTITY_TYPE_USER } from '../../schema/internalObject';
 import { type BasicStoreEntityOrganization, ENTITY_TYPE_IDENTITY_ORGANIZATION } from './organization-types';
 import type { AuthContext, AuthUser } from '../../types/user';
 import type { BasicObject, OrganizationAddInput, ResolversTypes } from '../../generated/graphql';
-import { FunctionalError } from '../../config/errors';
-import { isUserHasCapability, SETTINGS_SET_ACCESSES } from '../../utils/access';
+import { AlreadyDeletedError, FunctionalError } from '../../config/errors';
+import { isUserHasCapability, SETTINGS_SET_ACCESSES, SYSTEM_USER } from '../../utils/access';
 import { publishUserAction } from '../../listener/UserActionListener';
+import type { BasicStoreSettings } from '../../types/settings';
 import type { BasicStoreCommon, BasicStoreEntity } from '../../types/store';
 import { userSessionRefresh } from '../../domain/user';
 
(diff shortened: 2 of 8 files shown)


opencti-platform/opencti-graphql/tests/utils/testQueryHelper.ts
```diff
@@ -70,6 +70,22 @@ export const queryAsUserIsExpectedForbidden = async (client: AxiosInstance, requ
...
+ * @param errorMessage
+ * @param errorName
+ */
+export const queryAsUserIsExpectedError = async (client: AxiosInstance, request: any, errorMessage?: string, errorName?: string) => {
+  const queryResult = await executeInternalQuery(client, print(request.query), request.variables);
+  logApp.info('queryAsUserIsExpectedError=> queryResult:', queryResult);
+  expect(queryResult.errors, 'error is expected.').toBeDefined();
+  expect(queryResult.errors?.length, `1 error is expected, but got ${queryResult.errors?.length} errors`).toBe(1);
+  expect(queryResult.errors[0].message, `error message: ${errorMessage} is expected, but got ${queryResult.errors[0].message}`).toBe(errorMessage);
+  expect(queryResult.errors[0].extensions.code, `error is expected but got ${queryResult.errors[0].name}`).toBe(errorName);
```
lndrtrbn: What is the behavior if we are not giving params `errorMessage` or `errorName` as they are optional?

opencti-platform/opencti-graphql/src/modules/organization/organization-domain.ts
```diff
@@ -134,4 +137,33 @@ export const childOrganizationsPaginated = async <T extends BasicStoreCommon> (c
...
+  const organization = await findById(context, user, organizationId);
+  if (!organization) {
+    throw AlreadyDeletedError({ organizationId });
+  }
+  const settings = await getEntityFromCache<BasicStoreSettings>(context, SYSTEM_USER, ENTITY_TYPE_SETTINGS);
+  if (organization.id === settings.platform_organization) {
+    throw FunctionalError('Cannot delete the platform organization.');
+  }
+  if (organization.authorized_authorities && organization.authorized_authorities.length > 0) {
+    if (isUserHasCapability(user, SETTINGS_SET_ACCESSES)) {
```
lndrtrbn, line 150: If I have BYPASS capability I see the generic error message, should I not see the detailed one?

lndrtrbn: My bad. I can see the detailed message! 😅

opencti-platform/opencti-graphql/src/modules/organization/organization-domain.ts
```diff
@@ -134,4 +137,33 @@ export const childOrganizationsPaginated = async <T extends BasicStoreCommon> (c
   return listEntitiesThroughRelationsPaginated<T>(context, user, organizationId, RELATION_PART_OF, ENTITY_TYPE_IDENTITY_ORGANIZATION, true, args);
 };
 
+export const organizationDelete = async (context: AuthContext, user: AuthUser, organizationId: string) => {
+  const organization = await findById(context, user, organizationId);
+  if (!organization) {
+    throw AlreadyDeletedError({ organizationId });
```
lndrtrbn, line 143: How can we know if it's an organization already deleted or an ID that just matches no organization? Maybe the nuance is not important

opencti-platform/opencti-graphql/tests/utils/testQueryHelper.ts
SouadHadjiat: oh right, I should have put something by default, or remove the optional

opencti-platform/opencti-graphql/src/modules/organization/organization-domain.ts
```diff
@@ -134,4 +137,33 @@ export const childOrganizationsPaginated = async <T extends BasicStoreCommon> (c
...
+  }
+  const settings = await getEntityFromCache<BasicStoreSettings>(context, SYSTEM_USER, ENTITY_TYPE_SETTINGS);
+  if (organization.id === settings.platform_organization) {
+    throw FunctionalError('Cannot delete the platform organization.');
+  }
+  if (organization.authorized_authorities && organization.authorized_authorities.length > 0) {
+    if (isUserHasCapability(user, SETTINGS_SET_ACCESSES)) {
+      throw FunctionalError('Cannot delete the organization that has an administrator.');
```
Archidoit:
```suggestion
      throw FunctionalError('Cannot delete an organization that has an administrator.');
```

opencti-platform/opencti-graphql/src/modules/organization/organization-domain.ts
```diff
@@ -134,4 +137,33 @@ export const childOrganizationsPaginated = async <T extends BasicStoreCommon> (c
...
+    }
+    // no information leakage about the organization administrators or members
+    throw FunctionalError('Cannot delete the organization.');
+  }
+  const members = await organizationMembersPaginated(context, user, organization.id, { first: 1, indices: [READ_INDEX_INTERNAL_OBJECTS] });
+  if (members.pageInfo.globalCount > 0) {
+    if (isUserHasCapability(user, SETTINGS_SET_ACCESSES)) {
+      throw FunctionalError('Cannot delete the organization that has members.');
```
Archidoit:
```suggestion
      throw FunctionalError('Cannot delete an organization that has members.');
```

opencti-platform/opencti-graphql/src/modules/organization/organization-domain.ts
SouadHadjiat: I followed the same logic we have in internalDeleteElementById

Archidoit:
Behavior ok, working well 👍

Archidoit:
⚠️  We can still delete a platform organization via a background task in Data > Entities

SouadHadjiat:
> ⚠️ We can still delete a platform organization via a background task in Data > Entities

you're right, task manager calls directly deleteElementById, I need to change it as I did with retention manager => should be fixed now

lndrtrbn (approved):

opencti-platform/opencti-graphql/src/manager/taskManager.js
```diff
@@ -237,6 +238,9 @@ const executeDelete = async (context, user, element, scope) => {
   }
   if (scope === BackgroundTaskScope.Import) {
     await deleteFile(context, user, element.id);
+  } else if (element.entity_type === ENTITY_TYPE_IDENTITY_ORGANIZATION) {
+    // call organizationDelete which will ensure protections (for platform organization & members)
+    await organizationDelete(context, user, element.internal_id);
   } else {
     await deleteElementById(context, user, element.internal_id, element.entity_type);
```
Archidoit, line 245: wouln't it be better to add the organization check in 'deleteElementById' ? to avoid forgetting to add the check when using it?

lndrtrbn: That would probably make circle dependencies as `organizationDelete` uses `deleteElementById`. Maybe having a higher level function that make those kind of checks and then redirect to the correct delete function could be a thing? But it means some global refactoring at every delete resolver, maybe not relevant inside the scope of this PR idk

PR #8905, opened 2024-11-05
[backend/frontend] Danger zone: Remove FF (#8284)

marieflorescontact:
### Proposed changes

* Removes Feature flag
* documentation PR: https://github.com/OpenCTI-Platform/docs/pull/217
* Fixes 2 css issues in RuleList

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/8284

### Checklist


opencti-platform/opencti-front/src/private/components/settings/RulesList.jsx
```diff
@@ -549,7 +549,7 @@ const RulesListComponent = ({ relay, data, keyword }) => {
                     <Paper
                       variant="outlined"
                       classes={{ root: classes.paper }}
-                      style={{ marginTop: 25 }}
+                      style={{ marginTop: 23 }}
```
SouadHadjiat: I don't think it's the right solution for the display issue, even though it seem to work. I think the rule title should be moved outside of the grid, and we should remove completly this marginTop here.

SouadHadjiat: also it would be really nice to add a overflowX: 'auto' here so that it finally displays the whole rule definition for small screens.

marieflorescontact: > I don't think it's the right solution for the display issue, even though it seem to work. I think the rule title should be moved outside of the grid, and we should remove completly this marginTop here.

I don't know how to make the change because the rule name is now part of the `DangerZoneBlock` component 😕

opencti-platform/opencti-front/src/private/components/settings/groups/Group.tsx
```diff
@@ -467,7 +467,7 @@ const Group = ({ groupData }: { groupData: Group_group$key }) => {
       </Grid>
       <GroupEdition
         groupId={group.id}
-        disabled={!isAllowed}
+        disabled={!isAllowed && isSensitive}
```
labo-flg, line 470: why this change ?
I would naively think that isAllowed is always true if isSensitive is false (a non-sensitive zone is always granted).

Archidoit (dismissed):

Archidoit:
The Danger Zone button doesn't seem well aligned with the text (it is a bit lower)

image

opencti-platform/opencti-front/src/private/components/settings/groups/Group.tsx
marieflorescontact: because the edit button was disabled for non sensitive Group/Role if user was not allowed (not allowed to modified sensitive Group/Role )

marieflorescontact:
> The Danger Zone button doesn't seem well aligned with the text (it is a bit lower)
> 
> image

well in fact this is an illusion:
Capture d'écran 2024-11-07 101153

Archidoit:
> > The Danger Zone button doesn't seem well aligned with the text (it is a bit lower)
> > image
> 
> well in fact this is an illusion: Capture d'écran 2024-11-07 101153

yes the text in the button is aligned with the text, but the button itself is not aligned... Only design question I don't know if we let that :)

marieflorescontact:
> yes the text in the button is aligned with the text, but the button itself is not aligned... Only design question I don't know if we let that :)

 I don't see what is not aligned, could you explain? 💐

Archidoit (approved):

PR #8913, opened 2024-11-05
[frontend] Fix of multiple values in massOperations (#6049)

SarahBocognano:
### Proposed changes

* Fix on 6049

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/6049

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/data/playbooks/PlaybookAddComponents.jsx
+++ b/opencti-platform/opencti-front/src/private/components/data/playbooks/PlaybookAddComponents.jsx
@@ -18,7 +18,6 @@ import MenuItem from '@mui/material/MenuItem';
 import Tooltip from '@mui/material/Tooltip';
 import Box from '@mui/material/Box';
 import Alert from '@mui/material/Alert';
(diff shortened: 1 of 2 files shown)


opencti-platform/opencti-front/src/private/components/data/playbooks/PlaybookAddComponents.jsx
```diff
@@ -314,6 +296,61 @@ const PlaybookAddComponentsContent = ({
             ])}
           />
         );
+      case 'severity':
+      case 'priority':
+      case 'response_types':
+      case 'information_types':
+      case 'takedown_types': {
+        const vocabCategories = {
```
SouadHadjiat: maybe define `vocabCategories` as a constant outside of this method ?

labo-flg (changes requested):
Frontend works as expected, but when I try to use this new feature in the playbooks, the step "Manipulate knowledge" returns an empty bundle.

It works fine with adding a label for instance. Any action on assignee or participant is not working. The history of execution gives  empty bundle on the "out" output.

To track down backend  side, we're probably missing something.

labo-flg (approved):

Issue #8884, opened 2024-11-02
Header in creation drawer is not correct on observable


SamuelHassine:
## Description

Correct header (everywhere):

![image](https://github.com/user-attachments/assets/76f29162-215f-4e14-aa65-0d8f22a1ea96)

Incorrect header in observable creation:

![image](https://github.com/user-attachments/assets/ec2f7e76-9f00-4312-8ebf-42360fa7abef)

=> Missing grey background + color of the font (should be black).

PR #8919, opened 2024-11-06
[frontend] Header in creation drawer is not correct on observable (#8884)

SarahBocognano:
### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/8884


marieflorescontact (approved):
looks good 👌

PR #8922, opened 2024-11-06
[frontend] Add types filtering for related entities in cases (#8911)

### LINKED ISSUE #8911: [Request for takedown case] Related entities weird behavior
I've added observables in my takedown case, and the related entities were automatically populated with the same observables I added in the observable section. It seems to be duplicating the information. 

and if you try to add entities manually in the related entities, they are not kept after refresh. 

image

Kedae:
Closes #8911


labo-flg (approved):
tested ok, code seems good

PR #8924, opened 2024-11-06
[frontend] Improve error message notification in the UI (#8923)

### LINKED ISSUE #8186: Error is not parsed correctly when creating a public dashboard with an already existing key
## Description

image

### LINKED ISSUE #8923: Improve error message notification in the UI
## Use case

Improve error message notification in the UI.
Sometimes the message is not the real error (already existing workbench), sometimes is directly the error from the backend without any processing.

## Repro case

1. Create a workbench that already exists.
Actual : generic error
Expected: The real error

2. Create an observable domain with bad characters
Actual: Ugly message
Expected: The part of message that is really interesting

richard-julien:
See #8923


frapuks (approved):
It will also fix this open bug : #8186

PR #8925, opened 2024-11-06
[frontend] Allow all selected nodes to move when forces are desactivated (#8783)

### LINKED ISSUE #8783: When Disabling Forces in the Knowledge Graph, only one node moves when multiple are selected
## Environment

OpenCTI version: 6.3.6

## Reproducible Steps

1. Go to a knowledge graph within a report that has multiple relationships. 
2. Enable Forces.
3. Select multiple nodes - when dragging the nodes, all the nodes can move at once.
4. Disable Forces and keep the same nodes selected - when attempting to move all the nodes, only one node moves at a time. 

## Expected Output

When multiple nodes are selected, user should be able to move all the nodes together regardless of whether or not the forces' function is enabled. 

## Actual Output

When the forces function is disabled and multiple nodes are selected, the user can only drag one node at a time rather than all nodes that are selected.

Gwendoline-FAVRE-FELIX:
### Proposed changes

* In a graph, if forces are deactivated, it should be possible to move several selected nodes at the same time.

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/8783

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/analyses/groupings/GroupingKnowledgeGraph.jsx
+++ b/opencti-platform/opencti-front/src/private/components/analyses/groupings/GroupingKnowledgeGraph.jsx
@@ -1312,33 +1312,46 @@ class GroupingKnowledgeGraphComponent extends Component {
                         this.forceUpdate();
                       }}
                       onNodeDrag={(node, translate) => {
+                        const withForces = !this.state.modeFixed;
                         if (this.selectedNodes.has(node)) {
-                          [...this.selectedNodes]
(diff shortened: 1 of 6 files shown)


lndrtrbn (commented):
Technically it's working but it has impact on moving multiple nodes when forces are applied: it's laggy and at some point we have a visual effect where selected nodes are duplicated.


https://github.com/user-attachments/assets/41856c89-9c7e-4bff-a779-6fad786176ff

Hard to see on the video as there are low FPS, it's not that laggy in real life but still not as good as it was before the fix

lndrtrbn (approved):

PR #8945, opened 2024-11-08
[frontend/backend] Outcome templates: Exports and improvements (#8940)

### LINKED ISSUE #8940: Finish Outcome Template First Step
## Use case

Finish https://github.com/OpenCTI-Platform/opencti/issues/3402 by ensuring users can create **harcoded outcomes templates.**

Archidoit:
### Proposed changes
- Export button UI revision with 2 built-in connectors
- frontend tests
- texts display
- improvements

### Related issues

https://github.com/OpenCTI-Platform/opencti/issues/8940


opencti-platform/opencti-front/package.json
```diff
@@ -34,6 +34,7 @@
     "d3-timer": "3.0.1",
     "dagre": "0.8.5",
     "date-fns": "3.6.0",
+    "dompurify": "^3.2.1",
```
SouadHadjiat: versions should be fixed (no ^)

Archidoit: fixed

opencti-platform/opencti-graphql/config/schema/opencti.graphql
```diff
@@ -3189,7 +3189,7 @@ interface Container {
     types: [String]
     viaTypes: [String]
   ): ContainerConnection
-  contentsFromTemplate(
+  filesFromTemplate(
```
SouadHadjiat, line 3192: I don't understand why we have all these arguments in this query (first, prefixMimeType, after, orderBy, orderMode, search, filters), if we only use first on the frontend ? and in the resolver we only extract first and prefixMimeType, and we don't use the other arguments.
I know it's not directly related to this PR which only renames this API, but I just noticed it.

opencti-platform/opencti-front/src/utils/outcome_template/engine/useFileFromTemplate.ts
```diff
@@ -75,6 +75,7 @@ const useFileFromTemplate = () => {
       }
     }
 
+    console.log(content);
```
labo-flg: to remove :)

opencti-platform/opencti-graphql/config/schema/opencti.graphql
Archidoit: filesFromTemplate are exactly the same nature has importFiles, so for consistancy, the same arguments have been made available

aHenryJard (approved):

Issue #7056, opened 2024-05-21
Hide password/tokens in UI


nino-filigran:
## Use case

In case of screen sharing, having the pwd/token appearing in clear in the data sharing screen can be an issue.

This feature aims to make it hidden by default everywhere a sensitive content is in the platform.

PR #9003, opened 2024-11-14
[frontend] Fix Hide token in sync edition and add token in sync creation (#7056)

Gwendoline-FAVRE-FELIX:
### Proposed changes

* It was no longer possible to add a token when creating an openCTI streams.
* Fixed forgetting to hide the token when editing the openCTI Stream.

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/7056

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/data/sync/SyncEdition.jsx
+++ b/opencti-platform/opencti-front/src/private/components/data/sync/SyncEdition.jsx
@@ -2,7 +2,7 @@ import React, { useEffect, useState } from 'react';
 import { createFragmentContainer, graphql } from 'react-relay';
 import { Field, Form, Formik } from 'formik';
 import Typography from '@mui/material/Typography';
-import { CheckCircleOutlined, Visibility, VisibilityOff, WarningOutlined } from '@mui/icons-material';
+import { CheckCircleOutlined, WarningOutlined } from '@mui/icons-material';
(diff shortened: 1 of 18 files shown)


opencti-platform/opencti-front/src/private/components/data/sync/SyncEdition.jsx
```diff
@@ -208,15 +214,39 @@ const SyncEditionContainer = ({ synchronizer }) => {
...
-            <Field
-              component={TextField}
-              variant="standard"
-              name="token"
-              label={t_i18n('Remote OpenCTI token')}
-              fullWidth={true}
-              style={{ marginTop: 20 }}
-              disabled={true}
-            />
+            <div
```
lndrtrbn: Why not using the new field dedicated to show/hide secrets?

lndrtrbn: And I think the field should be named `PasswordTextField` instead of `ToggleVisibilityField` because it is not a toggle. It's a text field where you can show/hide the content

Gwendoline-FAVRE-FELIX: Effectively, this component was created after using the modifications in the ingestion files. I'm going to apply it to these files too.

lndrtrbn (approved):

Issue #8736, opened 2024-10-22
[RSS Feed] Error 403 on accessible public feeds


Lhorus6:
## Description

Some RSS feeds return errors 403 even though the links are public and accessible.

Examples:
- https://dailydarkweb.net/feed/
- https://cybersecurity.att.com/site/blog-all-rss
- https://www.securityweek.com/feed/

## Environment

OCTI 6.3.6

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Create an RSS Feed with one of the examples in the description
2. Wait a bit and look at the platform logs.

## Expected Output

Ingestion of the RSS feed

## Actual Output

Error 403

## Screenshots

![image](https://github.com/user-attachments/assets/f06be0f9-7cf4-4feb-86f3-6087ca451089)

khalidelborai:
Same issue here.

SamuelHassine:
Maybe a problem of user agent? @romain-filigran @nino-filigran maybe critical?

romain-filigran:
@SamuelHassine: Not only a user-agent issue from my investigation, more complicated depending on the RSS source. Need to test which one is working with an external RSS tool to identify the problem.

biastogit:
it seems this issue appeared on my instance 3 days ago after adding the RSS darkreading feed.
I started disabling all the RSS feeds creating errors, without success.
Also tried to delete the darkreading reports and have multiple update indexing fail errors (UI and logs)

SamuelHassine:
Any news @romain-filigran @nino-filigran ?

nino-filigran:
None @SamuelHassine we've increased priority to ensure it's looked over by devs. So far it seems sepcific to some RSS feeds & therefore the solution not straightforward, but we need to investigate.

aHenryJard:
Update: RSS feed works fine when removing the AxiosAgent from request. I think the best is to have an option to use agent or not, but we will have to rebuild the proxy option when AxiosAgent is not use, something like (from https://axios-http.com/docs/req_config ):

```
  proxy: {
    protocol: 'https',
    host: '127.0.0.1',
    port: 9000,
    auth: {
      username: 'mikeymike',
      password: 'rapunz3l'
    }
  },
```
As side note changes in this PR could help in the "no agent" use case https://github.com/OpenCTI-Platform/opencti/pull/6451

JeremyCloarec:
[...]

PR #9006, opened 2024-11-14
[backend] modify rss http getter to a simple fetch (#8736)

JeremyCloarec:
### Proposed changes

* instead of using httpClient, use simple fetch to get RSS data

### Related issues

* #8736 

### Checklist

### Further comments

In the related issue, all linked feeds are now fetched without any 403 errors.
However, the https://cybersecurity.att.com/site/blog-all-rss feed isn't ingested properly, because items in this feed don't have any pubDate metadata, they only have a dc:date. Not sure if we want to modify the RSS parser to use dc:date if no pubDate exist in the item?

### DIFF
--- a/opencti-platform/opencti-graphql/src/manager/ingestionManager.ts
+++ b/opencti-platform/opencti-graphql/src/manager/ingestionManager.ts
@@ -121,6 +121,7 @@ interface RssItem {
   'content:encoded': { _: string }
   category: { _: string } | { _: string }[]
   pubDate: { _: string }
+  'dc:date': { _: string }
   lastBuildDate: { _: string }
   updated: { _: string }
 }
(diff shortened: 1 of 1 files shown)


lndrtrbn (commented):
I don't know what are the advantages of using our custom httpClient but are we ok with the fact to bypass it?

JeremyCloarec:
> I don't know what are the advantages of using our custom httpClient but are we ok with the fact to bypass it?

I'm not sure either, but it was the only way I was able to bypass the 403 errors. We talked about it with @romain-filigran, and the plan will be to merge it to master and keep a close eye on wether the previous RSS feeds break following this change. If that is the case, this will need to be reverted

aHenryJard:
I think that the opencti httpclient manages at least proxy configuration, have you test your PR behind a proxy ?

aHenryJard (changes requested):
What is the root cause of the 403, have you found it ? If not maybe it's because we are requesting too much this URL (DDOS protection somehow or bot protection). I don't understand how changing the http client is solving issue can you elaborate please ?

I don't see any usage of proxy and proxy CA so I'm blocking until you give me feedback on the proxy settings.

JeremyCloarec:
I didn't think about proxy settings you're right, this solution doesn't work.
I wasn't able to find the root cause, I suspect a Cloudflare bot protection, but I didn't find any proper way to understand what triggers the 403 rejections. What's weird is that even when sending the exact same request as the browser but with curl, I get a 403 error. When the same request in the browser works properly...
Would you be up for a pair debugging session to dig into it?

PR #9008, opened 2024-11-14
[frontend] Fix Kill Chain list in Matrix view (#4230)

Archidoit:
### Proposed changes
In matrix views of a container or a Knowledge view, in the kill chain list, dislay all the kill chains used in an attack pattern

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/common/stix_domain_objects/StixDomainObjectAttackPatternsKillChain.tsx
+++ b/opencti-platform/opencti-front/src/private/components/common/stix_domain_objects/StixDomainObjectAttackPatternsKillChain.tsx
@@ -246,14 +246,11 @@ const StixDomainObjectAttackPatternsKillChain: FunctionComponent<StixDomainObjec
               value={activKillChainValue}
               onChange={handleKillChainChange}
             >
-              {killChains.map((killChainName) => {
-                const stringValue = String(killChainName);
-                return (
-                  <MenuItem key={stringValue} value={stringValue}>
-                    {stringValue}
-                  </MenuItem>
-                );
-              })}
+              {killChains.map((killChainName) => (
+                <MenuItem key={killChainName} value={killChainName}>
+                  {killChainName}
+                </MenuItem>
+              ))}
             </Select>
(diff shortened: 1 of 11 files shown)


opencti-platform/opencti-front/src/private/components/common/stix_domain_objects/StixDomainObjectAttackPatternsKillChain.tsx
```diff
@@ -126,14 +133,14 @@ const StixDomainObjectAttackPatternsKillChain: FunctionComponent<StixDomainObjec
...
-    .filter((name) => typeof name === 'string')
-    : [];
+  const killChainsData = usePreloadedFragment<AttackPatternsMatrixColumnsQuery, AttackPatternsMatrixColumns_data$key>({
+    queryDef: attackPatternsMatrixColumnsQuery,
+    fragmentDef: attackPatternsMatrixColumnsFragment,
+    queryRef: killChainDataQueryRef,
+  });
+  const killChainsPhaseData = killChainsData.attackPatternsMatrix?.attackPatternsOfPhases ?? [];
+  const killChains = R.uniq(killChainsPhaseData.map((a) => a.kill_chain_name))
+    .sort((a, b) => a.localeCompare(b));
```
delemaf, line 143: This works here's just a suggestion without Ramda (you can resolve if it's not relevant/clear)
```suggestion
  const killChains = [...new Set(killChainsPhaseData.map((a) => a.kill_chain_name))]
    .sort((a, b) => a.localeCompare(b));
```

delemaf (approved):
Looks clean to me, tested functionally both in Reports and Threat Actors 🚀

opencti-platform/opencti-front/src/private/components/common/stix_domain_objects/StixDomainObjectAttackPatternsKillChain.tsx
Archidoit: we usually accept ramda for R.uniq, it is clearer with R.uniq I think

Issue #7828, opened 2024-07-28
When deleting a connector in ingestion => connectors, interface is redirected to the connector and then to the list


SamuelHassine:
## Description

When deleting a connector in ingestion => connectors, interface is redirected to the connector and then to the list.

=> It should not redirect.

PR #9024, opened 2024-11-15
[frontend] When deleting a connector in ingestion => connectors, interface is redirected to the connector and then to the list (#7828)

SarahBocognano:
### Proposed changes

* Remove the ListItemButton for a regular ListItem

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/7828


marieflorescontact (approved):
tested ok 👌

PR #9046, opened 2024-11-17
[frontend align table UI in drawer

labo-flg:
Alignment issue (not present on master).


Kedae:
Should be fixed by https://github.com/OpenCTI-Platform/opencti/pull/9044

PR #9056, opened 2024-11-18
[frontend] fetch creators on observable creation (#8686)

### LINKED ISSUE #8686: Author not assigned as Creator if object created outside Report
## Description

Issue #1389 describes assignment / prefill of Author field by either Creator (if it is new object) or Report Author (if object is created from within report) and is marked as done in release 4.5.5:

> When creating a new object or relation, it would be useful to have the "author" field automatically filled, either :
> 
> - if it is a new object, the name of the author by default should be the name of the user creating the object
> - if it is a new relation within a report, the name of the author should be by default the author of the report.

As of OpenCTI v6.3.6 second option e.g Report Author -> Object Author works, but Author remains empty when creating object outside report. Looks like a possible regression due to race condition when assigning Creator 

## Environment

1. Local Docker setup, based on https://github.com/OpenCTI-Platform/docker
2. OpenCTI 6.3.6, clean install for testing

## Reproducible Steps

1. Select Observations from left menu, then Observables
2. Create new Observable, for example IPv4 address
3. You are returned to list of Observables

## Expected Output

Based on Issue #1389 both Author and Creator fields should be filled with name of current User

## Actual Output

List of observables shows new item, both Author and Creator are empty, showing `-`. When the list is refreshed (browser page reload) or object is opened Creator has been assigned as current User, Author is still empty  showing `-`.

View before refresh:

missing-author

delemaf:
### Proposed changes

* On observable creation, **creators** are now fetched after creation
* Added some typings to `StixCyberObservableCreation` and removed useless undefined props

### Related issues

* Close https://github.com/OpenCTI-Platform/opencti/issues/8686

### Checklist

### Further comments

I've added `objectOrganization` to `me` object in `useAuth` if it may be use anywhere else


opencti-platform/opencti-front/src/private/components/observations/stix_cyber_observables/StixCyberObservableCreation.jsx
```diff
@@ -195,6 +196,12 @@ const stixCyberObservableMutation = graphql`
           entity_type
         }
       }
+      creators {
+        ... on Creator {
```
SouadHadjiat: there is no need for `... on Creator`, it can directly be : 
```
creators {
  id
  name
}
```

SouadHadjiat (approved):
tested locally ✅

opencti-platform/opencti-front/src/private/components/observations/stix_cyber_observables/StixCyberObservableCreation.jsx
delemaf: thanks, updated 👍

PR #9068, opened 2024-11-18
[backend] Prevent retention rules to delete user individuals (#7787)

### LINKED ISSUE #7787: Retention rules may periodically publish error if trying to delete individual coresponding to a user
## Description

Retention rules on Individuals may try periodically to delete some individuals even if they are linked to a user --> we have periodically errors in the backend (and in support package files)

## Reproducible Steps
- Create a retention rule on everything (or on Individuals)
- Each time the retention manager re-run the batch of deletion, he tries to delete the individuals linked to a user, and an error is published in the backend : Cannot delete an individual corresponding to a user

## Expected output
Some proposed solutions:
- The retention manager shouldn't try to delete individuals if they correspond to a user
- An error message or a warning should be written somewhere in the app if something can't be deleted by a retention rule 
- The backend error message should be clearer and indicates that the error comes from a retention rule that tries periodically to delete the object

SouadHadjiat:
### Proposed changes

* add functions to a new data-consistency.Ts file : test if organization or individual can be deleted
* use these functions in middleware in internalDeleteById
* add a check in retention manager before deleting elements (so it won't try to delete user individuals for ex)
* cleanup code for organization deletion

### Related issues

* #7787 

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/tests/02-integration/01-database/middleware-test.js
+++ b/opencti-platform/opencti-graphql/tests/02-integration/01-database/middleware-test.js
@@ -1430,7 +1430,7 @@ describe('Delete functional errors behaviors', () => {
       .rejects.toThrowError('Cannot delete an organization that has members.');
   });
   it('should not be able to delete individual associated to user', async () => {
-    const individualUserId = 'identity--d37acc64-4a6f-4dc2-879a-a4c138d0a27f'; // in DATA-TEST-STIX2_v2.json
+    const individualUserId = 'identity--cfb1de38-c40a-5f51-81f3-35036a4e3b91'; // admin individual
     await expect(() => deleteElementById(testContext, ADMIN_USER, individualUserId, ENTITY_TYPE_IDENTITY_INDIVIDUAL))
       .rejects.toThrowError('Cannot delete an individual corresponding to a user');
   });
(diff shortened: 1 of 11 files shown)


opencti-platform/opencti-graphql/tests/02-integration/01-database/middleware-test.js
```diff
@@ -1423,3 +1423,15 @@ describe('Elements deduplication behaviors', () => {
     await deleteElementById(testContext, ADMIN_USER, group2.id, ENTITY_TYPE_THREAT_ACTOR_GROUP);
   });
 });
+
+describe('Delete functional errors behaviors', () => {
+  it('should not be able to delete organization that has members', async () => {
+    await expect(() => deleteElementById(testContext, ADMIN_USER, TEST_ORGANIZATION.id, ENTITY_TYPE_IDENTITY_ORGANIZATION))
+      .rejects.toThrowError('Cannot delete an organization that has members.');
+  });
+  it.skip('should not be able to delete individual associated to user', async () => {
```
SouadHadjiat, line 1432: This test is skipped for now because it looks like admin individual is not yet created. individual-test already tests this, but I wanted to have an extra test here in middleware.

aHenryJard (approved):
All good to me

PR #9076, opened 2024-11-19
[frontend/backend] Add new markings to max_shareable_markings of groups with auto_new_marking (#7944)

### LINKED ISSUE #7944: Option "auto new marking" in groups not working on max_shareable_markings, could make exports fail
## Description

**[!!! This issue needs proper analysis to be confirmed or not !!!]**

The option "auto new marking" in groups allow to add on the fly any newly added marking to the list of allowed marking in a group.

This automated process does not add the new marking to the max_shareable_markings list.

For instance in a group dedicated to connectors, this would cause export connectors to fail exporting knowledge marked with the new marking.

## Environment

6.2.x

Archidoit:
### Proposed changes
- New marking should be automatically added to max_shareable_markings of the groups with auto_new_marking = true
- UI fixes for max shareable markings in Group overview and group edition form
- Add tests for group with auto_new_marking = true

### Related issues
https://github.com/OpenCTI-Platform/opencti/issues/7944


lndrtrbn (changes requested):
When I create a new marking with lower order than a one a group already has on max sharing, it still replaces the existing one

Archidoit:
> When I create a new marking with lower order than a one a group already has on max sharing, it still replaces the existing one

fixed :)

lndrtrbn:
In the issue it is written:
> any new marking should be added in the "max shareable markings" with the value "no restriction"

In the PR, new marking is added in "max shareable markings" with the value of the marking. Was there any discussions that changes what is described in the issue?

Archidoit:
@lndrtrbn It is the same : 'having a marking value in max shareable markings' <=> no restriction to share this marking.

What's more, what's written in the issue is partially false. Indeed : 
It is not possible to have a marking in max shareable markings with the value 'no restrictions'. It is only marking definition types which can have no restrictions (meaning all the markings of this type can be shared).

Archidoit:
It is not possible to have a marking in max shareable markings with the value 'no restrictions'. It is only marking definition types which can have no restrictions (meaning all the markings of this type can be shared).

lndrtrbn (approved):

PR #9087, opened 2024-11-19
[frontend] use context to prefill the dynamic entity mapping column

CelineSebe:
### Proposed changes

* add CsvMapperContext to pass the dynamicMappingColumn to the others representation
* add onToggleDynamicMapping to set  the dynamicMappingColumn
* add useState to modify the value of the dynamicMappingColumn

### Related issues

* 7400

### Checklist


lndrtrbn (changes requested):
- Warnings and errors in console while manipulating the fields for dynamic mapping to fix,
- You cannot empty the fields Column index and Operator,
- What column should be kept in the context ? The first one set or the last one ? Currently it is the last one

lndrtrbn (approved):

PR #9089, opened 2024-11-19
[frontend] fix the green dot after setting a default value for an attribute ref

CelineSebe:
### Proposed changes

* add configuration={value} missing when the component is called

### Related issues

* #7400 

### Checklist

### Further comments

When we set a default value, we see a green dot appear! However, it doesn't work for the ref attribute (in this case, the author).:
image


marieflorescontact:
Is it possible to have a bit of context on the bug? 💐🙂

CelineSebe:
> Is it possible to have a bit of context on the bug? 💐🙂

FYI I updated the description above :)

lndrtrbn (approved):

PR #9119, opened 2024-11-22
[frontend] clamp confidence input values between 0 and 100 (#6715)

### LINKED ISSUE #6715: When an indicator score is set to 1000000, the object does not load properly and the system hangs
## Description

We don't have any check on indicator/observable score. I mean that we can set 1000000 as the score (should be between 0 and 100).

It may cause problems. For example, if we set up an indicator with a score of 1000000, the indicator does not load properly and the system hangs.

## Environment

OCTI 6.0.9

## Reproducible Steps

1. Create an indicator with a score of 1000000 / 100
2. Try to open this Indicator

## Expected Output

Prevent users from setting a score not between 0 and 100 (on observable and indicator)

delemaf:
### Proposed changes

* clamp confidence input field values between 0 and 100 🗜️ 

### Related issues

* Close [When an indicator score is set to 1000000, the object does not load properly and the system hangs](https://github.com/OpenCTI-Platform/opencti/issues/6715)

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/observations/indicators/IndicatorCreation.tsx
+++ b/opencti-platform/opencti-front/src/private/components/observations/indicators/IndicatorCreation.tsx
@@ -142,7 +142,10 @@ export const IndicatorCreationForm: FunctionComponent<IndicatorFormProps> = ({
         return !valid_from || !value || value > valid_from;
       }),
     x_mitre_platforms: Yup.array().nullable(),
-    x_opencti_score: Yup.number().nullable(),
+    x_opencti_score: Yup.number()
+      .nullable()
+      .min(0, t_i18n('The value must be greater than or equal to 0'))
+      .max(100, t_i18n('The value must be less than or equal to 100')),
     description: Yup.string().nullable(),
     x_opencti_detection: Yup.boolean().nullable(),
     createObservables: Yup.boolean().nullable(),
--- a/opencti-platform/opencti-front/src/private/components/observations/stix_cyber_observables/StixCyberObservableEditionOverview.jsx
+++ b/opencti-platform/opencti-front/src/private/components/observations/stix_cyber_observables/StixCyberObservableEditionOverview.jsx
@@ -241,9 +241,9 @@ const StixCyberObservableEditionOverviewComponent = ({
 
   const stixCyberObservableValidation = Yup.object().shape({
     x_opencti_score: Yup.number()
-      .min(0, 'Score must be between 0 and 100')
-      .max(100, 'Score must be between 0 and 100')
-      .nullable(),
+      .nullable()
+      .min(0, t_i18n('The value must be greater than or equal to 0'))
+      .max(100, t_i18n('The value must be less than or equal to 100')),
   });
 
   return (
--- a/opencti-platform/opencti-front/src/private/components/observations/stix_cyber_observables/StixCyberObservableEditionOverview.jsx
+++ b/opencti-platform/opencti-front/src/private/components/observations/stix_cyber_observables/StixCyberObservableEditionOverview.jsx
@@ -1,9 +1,10 @@
 import React from 'react';
 import { createFragmentContainer, graphql } from 'react-relay';
 import { Field, Form, Formik } from 'formik';
+import * as R from 'ramda';
 import { assoc, difference, filter, fromPairs, head, includes, map, pick, pipe } from 'ramda';
 import { useNavigate } from 'react-router-dom';
-import * as R from 'ramda';
+import * as Yup from 'yup';
 import TextField from '../../../../components/TextField';
 import { SubscriptionFocus } from '../../../../components/Subscription';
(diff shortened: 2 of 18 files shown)


lndrtrbn (changes requested):
From live discussions :
- rename component `InputScaleField`
- remove the bloc `if(variant === edit) {...}` and adapt the one we kept
- use min max attributes from input field number instead of a callback on change

SouadHadjiat:
The issue https://github.com/OpenCTI-Platform/opencti/issues/6715 is about indicator score, not confidence, why is this PR about confidence ?

delemaf:
> The issue #6715 is about indicator score, not confidence, why is this PR about confidence ?

I think I may have confused the score with the confidence, I'll move what I've done for confidence in score

delemaf:
PR have been reworked to handle min and max in front and back.

lndrtrbn (commented):
Could we add a test in backend to try creating + editing an indicator with score > 100 ?

opencti-platform/opencti-front/src/private/components/observations/indicators/IndicatorCreation.tsx
```diff
@@ -142,7 +142,10 @@ export const IndicatorCreationForm: FunctionComponent<IndicatorFormProps> = ({
         return !valid_from || !value || value > valid_from;
       }),
     x_mitre_platforms: Yup.array().nullable(),
-    x_opencti_score: Yup.number().nullable(),
+    x_opencti_score: Yup.number()
+      .nullable()
+      .min(0, t_i18n('The value must be greater than or equal to 0'))
```
Archidoit, line 147: maybe indicate directly the range? 'value between 0 and 100' ? (only a suggestion)

delemaf: We already have theses translations and kind of validation on other data (ex: confidence) so I didn't wanted to introduce new wording :wink:

Archidoit:
When I update an indicator with a score out of 0-100, the query is sent to the backend, resulting in an error message ('The score should be between 0 and 100)
image


--> it'not the case for other fields (if the field is not validated, it is not sent to the backend)
Example: enter an invalid start date, you get no messaging error at the top (only the validator error message), the invalid start date is not sent to the backend

opencti-platform/opencti-front/src/private/components/observations/indicators/IndicatorCreation.tsx
Archidoit: I understand, perfect :)

Archidoit (approved):

Archidoit:
> When I update an indicator with a score out of 0-100, the query is sent to the backend, resulting in an error message ('The score should be between 0 and 100) image
> 
> --> it'not the case for other fields (if the field is not validated, it is not sent to the backend) Example: enter an invalid start date, you get no messaging error at the top (only the validator error message), the invalid start date is not sent to the backend

will be fixed in https://github.com/OpenCTI-Platform/opencti/pull/10501

opencti-platform/opencti-front/src/private/components/observations/stix_cyber_observables/StixCyberObservableEditionOverview.jsx
```diff
@@ -237,6 +238,14 @@ const StixCyberObservableEditionOverviewComponent = ({
       }
     }
   };
+
+  const stixCyberObservableValidation = Yup.object().shape({
+    x_opencti_score: Yup.number()
+      .min(0, 'Score must be between 0 and 100')
+      .max(100, 'Score must be between 0 and 100')
```
lndrtrbn: This is not translated

opencti-platform/opencti-graphql/tests/02-integration/02-resolvers/indicator-test.ts
```diff
@@ -230,6 +247,20 @@ describe('Indicator resolver standard behavior', () => {
       expect(indicatorCreatedEarlier).toBeDefined();
     }
   });
+  it('should not update indicator with score value outside of 0 and 100', async () => {
```
lndrtrbn: Rename to explain the test: negative value

opencti-platform/opencti-graphql/tests/02-integration/02-resolvers/indicator-test.ts
```diff
@@ -119,6 +119,23 @@ describe('Indicator resolver standard behavior', () => {
   let indicatorStixId = 'indicator--f6ad652c-166a-43e6-98b8-8ff078e2349f';
   const indicatorForTestName = 'Indicator in indicator-test';
 
+  it('should not create indicator with score value outside of 0 and 100', async () => {
```
lndrtrbn: Rename to explain the test: higher than 100

lndrtrbn (approved):

PR #9121, opened 2024-11-22
[frontend] Fix regression on connectors status lines

SarahBocognano:
### Proposed changes

* After fixing this issue: https://github.com/OpenCTI-Platform/opencti/issues/7828 a regression has been observed. In the lines of connectors status, the item were "blue" so it was the wrong color. 

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/7828


marieflorescontact (approved):
looks good

Issue #7728, opened 2024-07-17
Add assignee through automation playbook


marckto-filigran:
## Use case

automatically assign report / incident to team members depending on trigger conditions
for report could be the source / author, for incident the severity 

## Current Workaround
Manual assignment

## Proposed Solution
Add the ability to assign a user to a report or case or incident through automation playbook.

nino-filigran:
This will be done as part of this feature: https://github.com/OpenCTI-Platform/opencti/issues/6049

PR #9132, opened 2024-11-25
[backend/frontend] Add assignee through automation playbook (#7728)

SarahBocognano:
### Proposed changes

* Add ObjectAssigneeFields in Manipulate Knowledge playbook form
* Add Assignee Input into playbook components for mapping attribute

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/7728
* https://github.com/OpenCTI-Platform/client-python/pull/792

### To test
* Add a new playbook 
* Choose to "Listen Knowledge Event" block 
* Choose an event to listen (create/update or both)
* Proceed to filter to a Case (Incident/RFT/RFI/Task)
* Add "Manipulate Knowledge" block
* Select Replace => Assignee => Value of your choice
* Add "Send for Ingestion" block
* Go to a Case and and update/create
* After a moment you should be able to see the new Assignee in this case

FIY: You need to enable your worker to be able to ingest AND to be on the related branch in client-python as it's a multi repository feature


SouadHadjiat:
Why can't we add assignees, and just replace ? The first thing I tried to do to test the feature is "add" not "replace", since I want to add an assignee. @SarahBocognano @nino-filigran @romain-filigran 
image


@SarahBocognano : It's because we only do a fieldPatchAttribute on assignees. It's the same logic with mass operations.

SouadHadjiat:
When I test replace assignee, it's adds the assignees, it doens't replace them. I tried with a report I created with "admin" assignee, with a playbook replace with two assignees, and I end up with 3 assignees. => It's the same logic with labels, it doesn't really replace, but there is message explaining it, maybe we should have a similar message for assignees too ? otherwise it's just confusing
image
image

opencti-platform/opencti-front/src/private/components/data/playbooks/PlaybookAddComponents.jsx
```diff
@@ -192,6 +193,7 @@ const PlaybookAddComponentsContent = ({
         { label: t_i18n('Author'), value: 'createdBy', isMultiple: false },
         { label: t_i18n('Confidence'), value: 'confidence', isMultiple: false },
         { label: t_i18n('Score'), value: 'x_opencti_score', isMultiple: false },
+        { label: t_i18n('Assignees'), value: 'objectAssignee', isMultiple: false },
```
SouadHadjiat: I would say isMultiple should be true, right ?

SarahBocognano: You are right on this one. Changing it now !

opencti-platform/opencti-graphql/src/modules/playbook/playbook-components.ts
```diff
@@ -575,6 +576,11 @@ const attributePathMapping: any = {
     [ABSTRACT_STIX_CYBER_OBSERVABLE]: `/extensions/${STIX_EXT_OCTI_SCO}/${createdBy.stixName}`,
     [ABSTRACT_STIX_RELATIONSHIP]: `/${createdBy.stixName}`,
   },
+  [INPUT_ASSIGNEE]: {
+    [ABSTRACT_STIX_DOMAIN_OBJECT]: `/extensions/${STIX_EXT_OCTI}/assignee_ids`,
+    [ABSTRACT_STIX_CYBER_OBSERVABLE]: `/extensions/${STIX_EXT_OCTI_SCO}/assignee_ids`,
+    [ABSTRACT_STIX_RELATIONSHIP]: `/extensions/${STIX_EXT_OCTI}/assignee_ids`,
```
SouadHadjiat: assignees can only be set for some stix domain objects, not observables nor relationships

opencti-platform/opencti-graphql/src/modules/playbook/playbook-components.ts
```diff
@@ -571,10 +573,16 @@ const attributePathMapping: any = {
     [ABSTRACT_STIX_RELATIONSHIP]: `/${objectLabel.stixName}`,
   },
   [INPUT_CREATED_BY]: {
-    [ABSTRACT_STIX_DOMAIN_OBJECT]: `/${createdBy.stixName}`,
+    [ABSTRACT_STIX_DOMAIN_OBJECT]: `/extensions/${STIX_EXT_OCTI}$/${createdBy.stixName}`,
```
SouadHadjiat: why this change ? created_by_ref doesn't seem to be part of octi extensions for stix domain objects

SouadHadjiat:
Add assignee & participants work well. 

There is just an issue with the message when we select remove : I can see two messages, as soon as I select remove. I wonder if we couldn't use the same message for labels, markings, assignees, and participants @SarahBocognano @nino-filigran 
image

I also think the current message shouldn't be the same for replace and remove operations : replace will in fact add to the existing values, and remove won't work at all for values that are already present in db.

opencti-platform/opencti-front/lang/front/fr.json
```diff
@@ -2303,10 +2303,12 @@
...
   "Remove from this object": "Retirer de cet objet",
+  "Remove operation will only apply on field values added in the context of this playbook such as enrichment or other knowledge manipulations but not if values are already written in the platform.": "L'opération de suppression ne s'appliquera qu'aux valeurs de champ ajoutées dans le contexte de ce playbook, comme l'enrichissement ou d'autres manipulations de connaissances, mais pas si les valeurs sont déjà écrites dans la plateforme.",
   "Remove selected items": "Supprimer les éléments sélectionnés",
   "Remove this reaction point": "Supprimer ce point de réaction",
   "Renew": "Renouveler",
   "Replace": "Remplacer",
+  "Replace operation will effectively replace this field values added in the context of this playbook such as enrichment or other knowledge manipulations but it will only append them if values are already written in the platform.": "L'opération Replace remplacera effectivement les valeurs de ce champ ajoutées dans le contexte de ce playbook tel que l'enrichissement ou d'autres manipulations de connaissances, mais elle ne les ajoutera que si les valeurs sont déjà écrites dans la plateforme.",
```
SouadHadjiat: L'opération Remplacer ... ou L'opération de remplacement

SouadHadjiat (approved):

Issue #9133, opened 2024-11-25
[RSS] Brackets escaped in RSS feed


christiaanvaken:
When having a bracket inside a title of an article in a RSS feed, the brackets are escaped with backslashes.

Example:
Name \[1232\] - Description

Any idea if this can be solved?

romain-filigran:
Hi @christiaanvaken. Thanks. I think it's a bug that we have to investigate.

For Filigran team: 

I have the same behavior when ingested RSS item with title containing bracket. Example with redpacketsecurity RSS Feed:

**Original RSS item**
![Image](https://github.com/user-attachments/assets/f54a5958-e91f-44c2-8d24-1e35b1c677ff)

**RSS Item in OpenCTI**
![Image](https://github.com/user-attachments/assets/fa2ccf3b-377f-4ac5-ae04-940771843b7e)

PR #9152, opened 2024-11-27
[backend] remove HTML parsing in title of RSS items (#9133)

JeremyCloarec:
### Proposed changes

* remove HTML parsing of title of RSS items

### Related issues

* #9133 

### Checklist

### Further comments

* we were currently parsing the title of RSS items as html strings, causing some issues when the title wasn't an html string but started with brackets
* report name don't handle markdown anyway, so there is no need to parse html string to markdown since it wouldn't work in report name in any case


marieflorescontact (approved):
locally tested ok 👌

Issue #8941, opened 2024-11-07
Finish Exclusion Lists MVP


nino-filigran:
## Use case

Finsih the scope defined in 6.4 (continuation  of: https://github.com/OpenCTI-Platform/opencti/issues/8312 ).
When this ticket will be done, users will be able to only **block** the creation of IOCs and not perform actions such as modifying the score or editing the label .This will be tackled in another ticket mentioned in the comments.

## If the feature request is approved, would you be willing to submit a PR?

Yes / No (Help can be provided if you need assistance submitting a PR)

debelyoo:
How does this issue relate to https://github.com/OpenCTI-Platform/opencti/issues/8944 ? Is the feature usable ? Or only for some restricted use cases ?

nino-filigran:
@debelyoo I understand the confusion since https://github.com/OpenCTI-Platform/opencti/issues/8944 scope was not clear, I've added a little bit more description in this ticket & the one you mentioned.

To be clear, this feature **is already usable**. Please note that the only actiont hat you can take is to block the creation of IOCs. The mentionned ticket aims to add more actions, if an IOC matches an exclusion lists that you have created. I hope it's clearer this way.

PR #9177, opened 2024-11-28
[backend] add missing backend feature to exclusion list (#8941)

JeremyCloarec:
### Proposed changes

* add exclusion list check in indicator edit
* add handling of file update in exclusion list field patch
* restrict exclusion list file size. Limits file size to 10MB by default
* add cache status query

### Related issues

* #8941 

### Checklist


opencti-platform/opencti-front/src/schema/relay.schema.graphql
```diff
@@ -12327,6 +12328,12 @@ type ExclusionListEdge {
   node: ExclusionList!
 }
 
+type ExclusionListCacheStatus {
+  refreshAskDate: String!
+  cacheDate: String!
+  allNodeCacheUpdated: Boolean!
```
SouadHadjiat: Why String for the dates and not DateTime ? and also all these values can be null I guess ?
Also the naming is not ideal, here are my suggestions if you like them better:
```
cacheDate: DateTime
isCacheRebuildInProgress: Boolean!
```

So that the frontend can use this isCacheRebuildInProgress to display a progress status if it's true, and if it's false, then cache is up to date. Do we need more info ?

SouadHadjiat: or maybe we would like the refreshAskDate for debug purposes ?

SouadHadjiat:
Tested locally:
- creation & edition of indicator with an excluded pattern throws an error ✅
- Creation of an exlusion list with the same name replaces the previous one : I think it can be an issue, better to not replace existing exclusion lists, and create a new one with the same name.

opencti-platform/opencti-front/src/schema/relay.schema.graphql
JeremyCloarec: That's what I was thinking, that refreshAskDate could still be useful sometimes to troubleshoot a cache problem if needed. For the types as String instead of DateTime, I didn't want to be too restrictive. I know that we sometimes talked about using version numbers instead of dates for the cache handling, so typing them as String gives us some leeway on the API side. Maybe the change could be to rename 'cacheDate' to 'cacheVersion'? This way it is more generic

SouadHadjiat (approved):

Issue #9049, opened 2024-11-18
Deleting a marking definition never close the confirmation popup


SamuelHassine:
## Description

Deleting a marking definition never close the confirmation popup

## Reproducible Steps

1. Go to Settings => Security => Marking definitions
2. Create a new marking
3. Try to delete it

=> The delete confirmation never closes

PR #9196, opened 2024-12-02
[frontend] Marking definitions data tables (9049)

Archidoit:
Fix marking definition lines display

https://github.com/OpenCTI-Platform/opencti/issues/9049


CelineSebe:
Tested in local 👌

CelineSebe (approved):

Issue #8174, opened 2024-08-27
Multiple issues in markdown content files


SamuelHassine:
## Description

https://filigran.octi.filigran.io/dashboard/cases/rfis/b382a3da-550d-4220-8bd5-1a035173634a/content:

1. Preview mode => no scrolling
2. Download as PDF: raw data instead of having properly formatted content

petskratt:
Point 1 is presumably related to assigned `height` parameter, disabling it results in correct scroll:

![mui-non-scrolling](https://github.com/user-attachments/assets/3e249e8d-318d-402f-b797-d1758c96d8d9)

I would also suggest making the Preview default view for markdown files (and perhaps also "Description & Main content").

Jipegien:
In addition to the fix, put the Preview mode as the default view for files and "Description & Main content" part.

romain-filigran:
related: 
https://github.com/OpenCTI-Platform/opencti/issues/7651
https://github.com/OpenCTI-Platform/opencti/issues/8203

Lhorus6:
Related: https://github.com/OpenCTI-Platform/opencti/issues/8202

PR #9213, opened 2024-12-03
[frontend] Make markdown content "preview" by default

Kedae:
Fixes missing comment here: https://github.com/OpenCTI-Platform/opencti/issues/8174#issuecomment-2314542199

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectMappableContent.tsx
+++ b/opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectMappableContent.tsx
@@ -59,7 +59,7 @@ const StixCoreObjectMappableContent: FunctionComponent<StixCoreObjectMappableCon
 }) => {
   const { t_i18n } = useFormatter();
   let { description, contentField } = containerData;
-  const [selectedTab, setSelectedTab] = useState(editionMode ? 'write' : 'preview');
+  const [selectedTab, setSelectedTab] = useState('preview');
   const basicShape = {
     content: Yup.string().nullable(),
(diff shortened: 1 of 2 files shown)


opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectMappableContent.tsx
```diff
@@ -59,7 +59,7 @@ const StixCoreObjectMappableContent: FunctionComponent<StixCoreObjectMappableCon
 }) => {
   const { t_i18n } = useFormatter();
   let { description, contentField } = containerData;
-  const [selectedTab, setSelectedTab] = useState(editionMode ? 'write' : 'preview');
```
lndrtrbn, line 62: Not sure to understand this change. It does not seem relevant to me to ignore if we are in edition. Why not keeping this line but setting `editionMode` to false in the parent ?

Kedae: We want to always have "preview" as default, even in edition. But may be I'm missing something, can you give more insight of your thought ?

lndrtrbn: Maybe I misunderstood what is asked by the issue. I found odd to set mode view only when entering in edition mode

Kedae: That is what is expected by the product.

lndrtrbn (approved):

PR #9218, opened 2024-12-03
[backend] fix internal_id being removed during observable fullsync (#8986)

### LINKED ISSUE #8986: [livestream] renaming of observable creates a new one in perfect sync mode
## Description

When using livestream to replicate data between two instances of opencti, **with perfect sync** activated, renaming an observable creates a new one instead of renaming it. I supposed this happens when modifying information that changes the ID of the entity, possibly for all entities (not tested).

This might be linked to this previous issue: https://github.com/OpenCTI-Platform/opencti/issues/8512 which does not fix the following case.

## Environment

1. OS (where OpenCTI server runs): docker
2. OpenCTI version: 6.3.11
3. OpenCTI client: frontend
4. Other environment details: Two instances of opencti, both deployed with docker

## Reproducible Steps

1. Setup data sharing between two platforms (public live stream on the source, and "perfect synchronization" on the destination)
2. Create a report on the source, create a new domain name and add it to the report
3. Wait for the two instances to be in sync
4. Rename the observable (the one linked to the report)
5. At first, nothing happends to the destination
6. Modify another field of the observable (e.g. the description)
7. A new observable is created on the destination, and the report is still linked to the old one. Instances are not in perfect sync anymore.

Screenshots below

## Expected Output

I expect the two instances to be in perfect sync after the reproducible steps.

## Actual Output

In the following screenshots, left is the source and right is the destination.

Creation of the report with the observable: 
Image

Renaming of the observable, nothing happens on the destination:
Image

Modification of the description of the observable. This creates a new observable on the destination, as shown by the history.
Image

And the old one is still here:
Image

The report is still linked to the first observable:
Image

JeremyCloarec:
### Proposed changes

* internal_id was being upserted to an undefined value in the case of fullsync. We now only set an internal_id in addStixCyberObservable if the internal_id is not undefined

### Related issues

* #8986 

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/domain/stixCyberObservable.js
+++ b/opencti-platform/opencti-graphql/src/domain/stixCyberObservable.js
@@ -204,7 +204,6 @@ export const addStixCyberObservable = async (context, user, input) => {
     return artifactImport(context, user, { ...input, ...input[graphQLType] });
   }
   const observableInput = {
-    internal_id,
     stix_id,
     x_opencti_score,
     x_opencti_description,
@@ -216,6 +215,9 @@ export const addStixCyberObservable = async (context, user, input) => {
     update,
     ...input[graphQLType]
   };
+  if (internal_id) {
+    observableInput.internal_id = internal_id;
+  }
   if (isNotEmptyField(payload_bin) && isNotEmptyField(url)) {
     throw FunctionalError('Cannot create observable with both payload_bin and url filled.');
   }


opencti-platform/opencti-graphql/src/domain/stixCyberObservable.js
```diff
@@ -216,6 +215,9 @@ export const addStixCyberObservable = async (context, user, input) => {
     update,
     ...input[graphQLType]
   };
+  if (internal_id) {
```
lndrtrbn, line 218: Could it be possible to have a case where `internal_id` from the input is `undefined` or `null` and we want to keep this null value ? Or it is sure that in any case we don't want it ?

SouadHadjiat: I don't undesrtand in which case we would have an internal_id in the observable input, it's not part of the schema. And internal_id is not part of the stream, right ? @JeremyCloarec

SouadHadjiat (approved):

PR #9236, opened 2024-12-04
BUGFIX: [graph] Hide filtered nodes/edges in correlation graphs

ckane:
Right now, these are simply greyed out as "disabled". However, when we filter items from the correlation graphs, we want it to "clean up" the graph. So, hide the affected nodes+edges from view so we can explore just the correlated items we include in the filters.

### Proposed changes

In `buildCorrelationData` in the `utils/Graph.js`, change the "filter" behavior from disabling filtered nodes (which simply changes their color) to remove them from the graph (and their edges). This is how the Correlation feature used to work, and the behavior regressed sometime later.

Keeping the disabled nodes + edges on the graph in correlation view significantly defeats the purpose of being able to filter these nodes out, as it makes large graphs difficult to navigate.

### Checklist

### Screenshots

Example of filter behavior fixed with this PR:

image
image

### DIFF
--- a/opencti-platform/opencti-front/src/utils/Graph.js
+++ b/opencti-platform/opencti-front/src/utils/Graph.js
@@ -477,7 +477,10 @@ export const buildCorrelationData = (
     filterAdjust.markedBy,
     filterAdjust.createdBy,
   );
-  const thisReportNodes = thisReportOriginalNodes.map((n) => R.assoc('disabled', filteredNodesIds.includes(n.id), n));
+  const thisReportNodes = R.filter(
+    (n) => !filteredNodesIds.includes(n.id),
+    thisReportOriginalNodes,
+  );
   const thisReportLinkNodes = R.filter(
     (n) => n[key] && n.parent_types && n[key].edges.length > 1,
     thisReportNodes,


labo-flg:
Hi @ckane and thanks for your contribution.
We already tested different approaches in the past.
The current version (having filtered items greyed out) is actually the solution we selected in the end for good reasons.

One side effect of removing the filtered items entirely is that the 3D forces are re-applied and you lose the positioning and quickly get lost.

ckane:
> Hi @ckane and thanks for your contribution. We already tested different approaches in the past. The current version (having filtered items greyed out) is actually the solution we selected in the end for good reasons.
> 
> One side effect of removing the filtered items entirely is that the 3D forces are re-applied and you lose the positioning and quickly get lost.

I understand, especially for the primary knowledge graphs. However, in this situation, when all of the "intermediate nodes" between the current report and a linked report are "greyed out", the linked-report in that case should be "greyed out" too, but it is not, so the existing behavior doesn't solve this problem either.

image

In this chart, as an example, the "Regin Scanner" report at top-right is the "current report". All nodes connected to "OSINT Regin Samples" are greyed out, so that report should also get greyed out, in order to use the filters to "hide correlations based upon linked entity type".

Additionally, I have CTI analysts who say that the "greying the nodes out" in the correlation charts creates charts that are difficult to view when they're trying to explore correlations to other reporting in the system using the correlation graphs, and they similarly describe the experience of "easy to get lost".

I realize the Investigations feature is also one potential option, but this is a lot more time-consuming and requires creating new investigation entities for work that will ultimately get thrown away.

I understand the distinction is probably a matter of having differing use-cases as well as different procedures for research. Can we have the choice between either behavior as an option inside the graphing widget?

Issue #9198, opened 2024-12-02
fix(graph): resolve missing graph in Correlation view for reports with correlated reports


dominictory:
## Description

We have a report created by the CIRCL MISP feed connector. On the Overview page, we see multiple correlated reports:

![Image](https://github.com/user-attachments/assets/838bd8a4-71a3-469b-b9f2-38e953d43ea3)

On the Knowledge tab, we see a graph. However, when attempting to use `Correlation view`, no graph is produced:

![Image](https://github.com/user-attachments/assets/21c77a6f-e1c9-448d-9024-daf56160655b)

## Environment

6.4.0

## Reproducible Steps

- Ingest a report that is correlated with other reports
- Attempt to use `Correlation view` for that report
- Observe no graph is produced

## Expected Output

Graph is produced in `Correlation view` showing correlations

## Actual Output

No graph produced in `Correlation view`. This occurs for other reports as well.

nino-filigran:
@dominictory we know that this area of the app does not work well, that is why we are currentlyworking on a task to align the graph & the data contained in the container overview. https://github.com/OpenCTI-Platform/opencti/issues/3227. 
In short, the 2 queries are different, resulting in a different view, as you have rightfully pointed. Therefore, the feature should fix your bug cc @romain-filigran FYI

ckane:
I think that I have noticed this on graphs with a large number of correlations. I don't know what, if any, limit we might be running into. I suspect that one of the filters applied in the `buildCorrelationData` function has a logic flaw in it.

ckane:
So I added some instrumentation to my deployment and it seems like when the number of relationships goes above 500, this behavior starts to manifest. I suspect that the hard-coded `first: 500` and similar 500-entry limits is being hit somewhere along the way such that the function linked below only gets the first 500 items from the "total objects" list and it ends up not pivoting into ones with related items:
* https://github.com/OpenCTI-Platform/opencti/blob/master/opencti-platform/opencti-front/src/utils/Graph.js#L438

romain-filigran:
@Gwendoline-FAVRE-FELIX @delemaf : Can you check whether your current work on graph and correlation fix this issue ?

labo-flg:
I'm not confident latest changes would fix the issue.
Also, it's very hard to reproduce without the actual dataset (all correlated containers and their content).

IMO the problem comes from multiple stages of limitations (as the one rightfully pointed by @ckane).
The correlation graph was initially developed by taking a straight approach : 

[...]

Issue #6839, opened 2024-04-29
Reports don't manage correctly big numbers of entities


jborozco:
Created from [https://github.com/OpenCTI-Platform/opencti/issues/6499](https://github.com/OpenCTI-Platform/opencti/issues/6499
)
## Description

When there are too many entities in a report:
- Visualization in knowledge is empty
- The list of entites is empty (https://github.com/OpenCTI-Platform/opencti/issues/6499#issuecomment-2079175889)

## Environment

1. OS (where OpenCTI server runs): { e.g. Mac OS 10, Windows 10, Ubuntu 16.4, etc. }
2. OpenCTI version: { e.g. OpenCTI 1.0.2 }
3. OpenCTI client: { e.g. frontend or python }
4. Other environment details:

## Reproducible Steps

Have a report with 
I couldn't reproduce it but the are a lot of info and screenshot here [https://github.com/OpenCTI-Platform/opencti/issues/6499](https://github.com/OpenCTI-Platform/opencti/issues/6499)

In theory, open a report with 3k entities

## Expected Output

All entities are visualised 

## Actual Output

The pages are blank

jborozco:
related to [https://github.com/OpenCTI-Platform/opencti/issues/6499](https://github.com/OpenCTI-Platform/opencti/issues/6499)

richard-julien:
Can you provide a repro case on testing @jborozco? thanks

jborozco:
Reproducible here https://testing.octi.staging.filigran.io/dashboard/analyses/reports/3ba859df-b1fd-4bd1-bcf8-8307a31a8dc1/entities

Entity list 
![image](https://github.com/OpenCTI-Platform/opencti/assets/35300800/63af405c-4e7a-4654-8cdf-08cbd16604da)


Knowledge graph
![image](https://github.com/OpenCTI-Platform/opencti/assets/35300800/4b92c395-48a1-4f9b-8adc-5e05f3e7e66a)


In theory there should be around 8k entities in this graph 
![image](https://github.com/OpenCTI-Platform/opencti/assets/35300800/f10ebc6a-1579-4f2a-acff-b18fb0db6dd9)

AlexSanchezN:
Sorry to intrude... is there any prevision for this?
It makes analyzing large sets of data a pain :-)

Thanks for your effort!!

dominictory:
[...]

PR #9203, opened 2024-12-02
[frontend] Filter correlation graphs on common SCOs and indicators by default (#3227)

Gwendoline-FAVRE-FELIX:
### Proposed changes

* Modification of correlation graph queries to display filtering by indicators and stix-cyber-objects in common.
* Add a button to display all the entities in common.

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/3227

PR #9175, opened 2024-11-28
[frontend] Show all types of containers in overview correlation widget & CTA to correlation graph (#3227)

delemaf:
### Proposed changes

* Show all types of containers in overview correlation widget
* Button CTA to correlation graph
image

### Related issues

* [Related containers / correlation enhancements #3227](https://github.com/OpenCTI-Platform/opencti/issues/3227)

### Checklist

- [x] I consider the submitted work as finished
- [x] I tested the code for its functionality
- [ ] I wrote test cases for the relevant uses case (coverage and e2e)
- [ ] I added/update the relevant documentation (either on github or on notion)
- [x] Where necessary I refactored code to improve the overall quality

### Further comments

To test you have to link the same observable (Ex: URL) to two different containers (Ex: reports/groupings/caseRFI/caseRFT/incidents), then the related containers will show up on overview

PR #9240, opened 2024-12-05
BUGFIX: [graph] Increase count of related elements fetched by Correlation Graph to 500000

ckane:
The 500 upper limit is very low and results in a lot of correlation graphs displaying no correlations in the chart, while the main page of the Graph says there exist correlations. In some cases this limit was just 20. This fixes a lot of the reported problems where related reporting is seen on the Overview page of a analysis/case/etc. but the Correlation Graph is blank (because it isn't fetching all nodes, and the correlations don't show in the first 500 fetched). In some scenarios, the graph does show some items, but is missing other expected ones.

### Proposed changes

* Set `first: 500000` when fetching first-degree and second-degree related nodes for Correlation Graphs

### Related issues
* #9198 
* #6839

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/analyses/groupings/GroupingKnowledgeCorrelation.jsx
+++ b/opencti-platform/opencti-front/src/private/components/analyses/groupings/GroupingKnowledgeCorrelation.jsx
@@ -1255,7 +1255,7 @@ const GroupingKnowledgeCorrelation = createFragmentContainer(
           x_opencti_order
           x_opencti_color
         }
-        objects {
+        objects(first: 500000) {
           edges {
             node {
               ... on BasicObject {
--- a/opencti-platform/opencti-front/src/private/components/analyses/groupings/GroupingKnowledgeCorrelation.jsx
+++ b/opencti-platform/opencti-front/src/private/components/analyses/groupings/GroupingKnowledgeCorrelation.jsx
@@ -1279,7 +1279,7 @@ const GroupingKnowledgeCorrelation = createFragmentContainer(
                   x_opencti_order
                   x_opencti_color
                 }
-                groupings(first: 20) {
+                groupings(first: 500000) {
                   edges {
                     node {
                       id
@@ -1387,7 +1387,7 @@ const GroupingKnowledgeCorrelation = createFragmentContainer(
               }
               ... on StixCyberObservable {
                 observable_value
-                groupings(first: 20) {
+                groupings(first: 500000) {
                   edges {
                     node {
                       id
--- a/opencti-platform/opencti-front/src/private/components/analyses/groupings/GroupingKnowledgeCorrelation.jsx
+++ b/opencti-platform/opencti-front/src/private/components/analyses/groupings/GroupingKnowledgeCorrelation.jsx
@@ -1279,12 +1279,65 @@ const GroupingKnowledgeCorrelation = createFragmentContainer(
                   x_opencti_order
                   x_opencti_color
                 }
+                cases(first: 500000) {
+                  edges {
+                    node {
+                      id
+                      name
+                      confidence
+                      entity_type
+                      parent_types
+                      created_at
+                      createdBy {
+                        ... on Identity {
+                          id
+                          name
+                          entity_type
+                        }
+                      }
+                      objectMarking {
+                        id
+                        definition_type
+                        definition
+                        x_opencti_order
+                        x_opencti_color
+                      }
+                    }
+                  }
+                }
                 groupings(first: 500000) {
                   edges {
                     node {
                       id
                       name
                       context
+                      confidence
+                      entity_type
+                      parent_types
+                      created_at
+                      createdBy {
+                        ... on Identity {
+                          id
+                          name
+                          entity_type
+                        }
+                      }
+                      objectMarking {
+                        id
+                        definition_type
+                        definition
+                        x_opencti_order
+                        x_opencti_color
+                      }
+                    }
+                  }
+                }
+                reports(first: 500000) {
+                  edges {
+                    node {
+                      id
+                      name
+                      published
                       confidence
                       entity_type
                       parent_types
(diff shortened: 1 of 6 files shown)


labo-flg:
Thanks for your contributions !
I totally agree this limit makes the correlations views inconsistent, however it's a bit delicate to entirely remove the limit, for performance reasons.

Imagine you have a widely used entity like "APT28" or "USA", present in thousands of reports. GQL will wait for the complete resolution of this huge graph, which will take a long time. Display would suffer a lot too.

ckane:
> Thanks for your contributions ! I totally agree this limit makes the correlations views inconsistent, however it's a bit delicate to entirely remove the limit, for performance reasons.
> 
> Imagine you have a widely used entity like "APT28" or "USA", present in thousands of reports. GQL will wait for the complete resolution of this huge graph, which will take a long time. Display would suffer a lot too.

I agree, thought in my experience that behavior will probably hit an upper limit on browser or network timeout, which a user can still escape from if they don't like it - so with the low upper-limit, those "big reports" are unusable but also a bunch of medium-sized reports are also unusable. With this change, those big reports may still be unusable, but it makes the medium and small sized reports more usable and accurate.

The graphing library seems fairly performant, and here's an example of one with 1000's of items that seemed to render fine for me:
image

ckane:
"BIG Reports" like the "Operation SMN (Novetta)" report that is available via MISP Feeds are still unusable, and eventually display an error. However, the normal entity graph for that report also doesn't work, and times out w/ an error while trying to fetch its content. This report is my favorite use case when trying to test/evaluate "large report" problems.
image

In my opinion, the failure mode of "too big" graphs occurs in a much more limited set of circumstances than the current "missing data" behavior today, and between the two, I would say it is preferred to tolerate the failure where a small number of large or very-related reports might error out with a render failure.

ckane:
If you try to render a large graph and it's too big, it shows this error after about ~15secs, so IMO that behavior is preferable
image

ckane:
I took this work a step further and have the correlation traversal algorithm traverse all related nodes of type `cases`, `groupings`, and `reports` regardless of which entity type is the "current view". This allows for the exploration of related OSINT reporting from within an Incident Reponse or Grouping container's correlation graph, and vice-versa.

Example view from within an RFI object I created:
image

labo-flg:
> I took this work a step further and have the correlation traversal algorithm traverse all related nodes of type cases, groupings, and reports regardless of which entity type is the "current view". This allows for the exploration of related OSINT reporting from within an Incident Reponse or Grouping container's correlation graph, and vice-versa.

That's a good point. Part of our work on https://github.com/OpenCTI-Platform/opencti/issues/3227 for next milestone addressed this issue. It has been merged in https://github.com/OpenCTI-Platform/opencti/pull/9175 

Note that in the details view, the correlated containers are detected through common observables and indicators. To be consistent we also aligned in the graph view.
We add in [this PR](https://github.com/OpenCTI-Platform/opencti/pull/9203) the ability to switch in the graph view from "observables and indicators" to "all entities" in case user needs extended correlations.

Now, concerning the limitation & performances issues, this raised a lot of discussions.
We need to be careful about the impact of such changes : in some conditions it might generate such a huge amount of data that your backend process would just eat all the memory and die.
See this issue : https://github.com/OpenCTI-Platform/opencti/issues/9293

We think the best solution is a heavy refactoring of the graph code to fetch data bits by bits, paginating results and loading the graph incrementally.
I've drafted an issue for that https://github.com/OpenCTI-Platform/opencti/issues/9359

ckane:
> > I took this work a step further and have the correlation traversal algorithm traverse all related nodes of type cases, groupings, and reports regardless of which entity type is the "current view". This allows for the exploration of related OSINT reporting from within an Incident Reponse or Grouping container's correlation graph, and vice-versa.
> 
> That's a good point. Part of our work on #3227 for next milestone addressed this issue. It has been merged in #9175
> 
> Note that in the details view, the correlated containers are detected through common observables and indicators. To be consistent we also aligned in the graph view. We add in [this PR](https://github.com/OpenCTI-Platform/opencti/pull/9203) the ability to switch in the graph view from "observables and indicators" to "all entities" in case user needs extended correlations.
> 
> Now, concerning the limitation & performances issues, this raised a lot of discussions. We need to be careful about the impact of such changes : in some conditions it might generate such a huge amount of data that your backend process would just eat all the memory and die. See this issue : #9293
> 
> We think the best solution is a heavy refactoring of the graph code to fetch data bits by bits, paginating results and loading the graph incrementally. I've drafted an issue for that #9359

Thanks! I agree that the best implementation would be to do a paginated walk to incrementally build the graph by making smaller requests that don't overload the back-end. I don't understand well enough how to do that for these 2nd-degree related items, however (thus, my solution being to just set the limits really really high).

I do think that having the ability to have the front-end query what the "total set" size would be ahead of time, so that it can render a progress meter during incremental graph build-out, would be a helpful UI element to it, so the end users know when the data is still being retrieved. The partial graph could be shown in real-time, and have additions to it added and displayed as they're fetched from the server.

Additionally, having a "stop" and/or "pause" button that allows the data fetch/graph-build to be stopped would also be helpful. So, for larger graphs, if the end user happens to see the data they're interested in visible on the graph at some point during the paginated ingest, they could stop or pause the data fetch and then explore the portion of the graph that's been built up to that point.

labo-flg:
That's where we want to go next, we're aligned on that :)

I'm thinking step-by-step queries, all properly paginated. The frontend would query the main container, then it's content, then for each one the relate containers, etc. 
The graph would load bits by bits on screen, with a proper loading bar.

Knowing in advance the size of the full graph is challenging, but at least we can give visual feedback with all the info we can get, as soon as we get them.

aHenryJard:
Thank you for your contribution but we are currently refactoring the whole graph component so this PR is going to be superseeded by the graph refactoring so I'm closing. Should be avalaible in one of the next major release.

PR #9244, opened 2024-12-05
[backend/frontend] add configuration to have better chances to get RSS Feed contents (#8736)

### LINKED ISSUE #8736: [RSS Feed] Error 403 on accessible public feeds

aHenryJard:
### Proposed changes

Adding several configuration to get better chance to avoid HTTP 403 from public RSS Feed:
- ingestion_manager.rss_feed.min_interval_minutes : wait for at least N minutes before next HTTP call for RSS Feed
- ingestion_manager.rss_feed.user_agent: allow to overide user agent used in header for RSS Feed requests
- ingestion_manager.csv_feed.min_interval_minutes : wait for at least N minutes before next HTTP call for CSV Feed

Adding some info and warn level log:
- Info to know if an HTTP Call has been skipped because of min_interval_minutes or queue states
- warn in case of 403 coming from a cloudflare challenge to know that it's not a credential issue.

Doc PR https://github.com/OpenCTI-Platform/docs/pull/231

image

### Related issues

* closes #8736 

### Checklist


opencti-platform/opencti-graphql/src/manager/ingestionManager.ts
```diff
@@ -519,17 +536,22 @@ const csvExecutor = async (context: AuthContext) => {
   const ingestionPromises = [];
   for (let i = 0; i < ingestions.length; i += 1) {
     const ingestion = ingestions[i];
-    // If ingestion have remaining messages in the queue, dont fetch any new data
+    // If ingestion have remaining messages in the queue, or if last execution was done before CSV_FEED_MIN_INTERVAL_MINUTES minutes, dont fetch any new data
     const { messages_number, messages_size } = await queueDetails(connectorIdFromIngestId(ingestion.id));
-    if (messages_number === 0) {
+    const { last_execution_date } = ingestion;
+    const shouldExecuteIngestion = !last_execution_date || sinceNowInMinutes(last_execution_date) > CSV_FEED_MIN_INTERVAL_MINUTES;
```
SouadHadjiat: why strict > ? so if CSV_FEED_MIN_INTERVAL_MINUTES is 5min, and sinceNowInMinutes(last_execution_date) is 5, it won't execute until the next minute?

SouadHadjiat: and since this code is the same for RSS, I suggest to maybe refactor it to use a method like : 
```
const shouldExexcuteIngestion = (ingestion, min_interval_minutes) => {
   const { last_execution_date } = ingestion;
   return !last_execution_date || sinceNowInMinutes(last_execution_date) > min_interval_minutes;
}
```
so we could call it like this `shouldExecuteIngestion(ingestion, CSV_FEED_MIN_INTERVAL_MINUTES)` or `shouldExecuteIngestion(ingestion, RSS_FEED_MIN_INTERVAL_MINUTES)`

opencti-platform/opencti-front/src/private/components/data/ingestionRss/IngestionRssLine.jsx
```diff
@@ -99,6 +99,12 @@ class IngestionRssLineLineComponent extends Component {
                   status={!!node.ingestion_running}
                 />
               </div>
+              <div
+                className={classes.bodyItem}
+                style={{ width: dataColumns.last_execution_date.width }}
+              >
+                {nsdt(node.last_execution_date)}
```
lndrtrbn: No need for this `|| '-'` check here?

opencti-platform/opencti-graphql/src/utils/http-client.ts
```diff
@@ -38,8 +38,12 @@ export const getHttpClient = ({ baseURL, headers, rejectUnauthorized, responseTy
...
   const key = isNotEmptyField(certificates?.key) ? fromBase64(certificates?.key) : undefined;
   const ca = isNotEmptyField(certificates?.ca) ? fromBase64(certificates?.ca) : undefined;
-  const defaultHttpAgent = new http.Agent();
-  const defaultHttpsAgent = new https.Agent({ rejectUnauthorized: rejectUnauthorized === true, cert, key, ca });
+  const defaultHttpAgent: http.Agent | undefined = undefined;
+  let defaultHttpsAgent: https.Agent | undefined;
+
+  if (cert || key || ca) {
+    defaultHttpsAgent = new https.Agent({ rejectUnauthorized: rejectUnauthorized === true, cert, key, ca });
+  }
```
lndrtrbn: Not sure to understand the logic behind the changes. Moreover, `defaultHttpAgent` is always undefined

opencti-platform/opencti-graphql/src/manager/ingestionManager.ts
aHenryJard: Inerval is 30s so it will wait only 30s (well for default values). Anyway I'm adding an <= and ok for shouldExecuteIngestion

opencti-platform/opencti-graphql/src/utils/http-client.ts
aHenryJard: I'm going to testagain because it's the more risky change, but the issue was that defaultHttpAgent was set to new Agent() instead of undefined. Then is there is some proxy configuration the agent is correctly initialize in 
https://github.com/OpenCTI-Platform/opencti/blob/b293e2e65964e852ee3ad5eaf50971fc9db648bb/opencti-platform/opencti-graphql/src/config/conf.js#L396-L409

As you see there, if no proxy the return is undefined, which means that in the http call there is no empty httpAgent => from my test having empty httpAgent in request is what is causing being refused by cloudflare.

aHenryJard: Maybe I should just remove defaultHttpAgent in http-client

aHenryJard: reverted.

lndrtrbn (approved):

Issue #9211, opened 2024-12-03
Improve confidentiality of history


marieflorescontact:
Continuous improvement on security concerns  in the platform.

PR #9251, opened 2024-12-05
[backend] Add relatedRestrictions in stream

marieflorescontact:
### Proposed changes

* Add relatedRestrictions in stream : previously only instance's objectMarking were send into the stream. We now send  added objects's objectMarking
* Add unit test

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/9211

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/database/middleware.js
+++ b/opencti-platform/opencti-graphql/src/database/middleware.js
@@ -2871,8 +2871,7 @@ export const createRelationRaw = async (context, user, rawInput, opts = {}) => {
         message: opts.commitMessage,
         external_references: references.map((ref) => convertExternalReferenceToStix(ref))
       } : undefined;
-      const relatedRestrictions = extractObjectsRestrictionsFromInputs(inputs);
-      event = await storeUpdateEvent(context, user, previous, instance, message, { ...opts, commit, relatedRestrictions });
+      event = await storeUpdateEvent(context, user, previous, instance, message, { ...opts, commit });
       dataRel.element.from = instance; // dynamically update the from to have an up to date relation
     } else {
       const createdRelation = { ...resolvedInput, ...dataRel.element };
--- a/opencti-platform/opencti-graphql/src/database/middleware.js
+++ b/opencti-platform/opencti-graphql/src/database/middleware.js
@@ -2871,7 +2871,8 @@ export const createRelationRaw = async (context, user, rawInput, opts = {}) => {
         message: opts.commitMessage,
         external_references: references.map((ref) => convertExternalReferenceToStix(ref))
       } : undefined;
-      event = await storeUpdateEvent(context, user, previous, instance, message, { ...opts, commit });
+      const relatedRestrictions = extractObjectsRestrictionsFromInputs(inputs);
+      event = await storeUpdateEvent(context, user, previous, instance, message, { ...opts, commit, relatedRestrictions });
       dataRel.element.from = instance; // dynamically update the from to have an up to date relation
     } else {
       const createdRelation = { ...resolvedInput, ...dataRel.element };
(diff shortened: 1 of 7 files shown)


opencti-platform/opencti-graphql/src/manager/historyManager.ts
```diff
@@ -103,6 +103,10 @@ export const buildHistoryElementsFromEvents = async (context:AuthContext, events
         .map((stixId) => grantedRefsResolved.get(stixId))
         .filter((o) => isNotEmptyField(o));
     }
+    if (event.data.relatedRestrictions) {
+      const relatedMarkings = event.data.relatedRestrictions.markings ?? [];
+      eventMarkingRefs.push(...relatedMarkings);
```
SouadHadjiat: add unit test in historyManager-test to cover this use case

opencti-platform/opencti-graphql/src/database/redis.ts
```diff
@@ -523,6 +523,7 @@ export const buildStixUpdateEvent = (user: AuthUser, previousStix: StixCoreObjec
     origin: user.origin,
     data: stix,
     commit: opts.commit,
+    relatedRestrictions: opts.relatedRestrictions,
```
richard-julien: Maybe better to put that in the context

opencti-platform/opencti-graphql/src/database/utils.js
```diff
@@ -326,6 +326,19 @@ export const extractIdsFromStoreObject = (instance) => {
   return ids;
 };
 
+export const extractObjectsRestrictionsFromInputs = (inputs) => {
```
richard-julien: This method should be a bit more controlled on the type of the element.
objects are linked to container.
You also only extract the markings and not the other parts of ACL (orga sharing + authorized members)
Constant should also be available instead of using 'objects' and 'object-marking'

richard-julien (changes requested):

opencti-platform/opencti-graphql/src/database/utils.js
SouadHadjiat: I didn't think it was relevent for this case to handle orga sharing and authorized members, since the objects added always inherits the organization sharing restrictions of the container. And authorized members is not handled at all by history manager for now.

opencti-platform/opencti-graphql/src/database/redis.ts
marieflorescontact: done

opencti-platform/opencti-graphql/src/types/event.d.ts
```diff
@@ -52,6 +53,7 @@ interface UpdateEvent extends StreamDataEvent {
   context: {
     patch: Array<Operation>;
     reverse_patch: Array<Operation>;
+    relatedRestrictions?: { markings: string[] };
```
richard-julien: Can you use camel case here ? related_restrictions 
Thanks

marieflorescontact: off course. my bad

opencti-platform/opencti-graphql/src/database/utils.js
richard-julien: You're right for the sharing and the authorized members.
For the input, its more about testing if the entity type is a container in order to get the related restrictions instead of just rely on the inputs

opencti-platform/opencti-graphql/src/database/middleware.js
```diff
@@ -2200,7 +2201,8 @@ export const updateAttributeMetaResolved = async (context, user, initial, inputs
         message: opts.commitMessage,
         external_references: references.map((ref) => convertExternalReferenceToStix(ref))
       } : undefined;
-      const event = await storeUpdateEvent(context, user, initial, updatedInstance, message, { ...opts, commit });
+      const relatedRestrictions = extractObjectsRestrictionsFromInputs(updatedInputs, initial.entity_type);
+      const event = await storeUpdateEvent(context, user, initial, updatedInstance, message, { ...opts, commit, relatedRestrictions });
```
SouadHadjiat: shouldn't it be `related_restrictions` ?

SouadHadjiat (approved):

richard-julien (approved):

Issue #9065, opened 2024-11-18
[Playbook] Changing Author does not work and prevents all edits


Lhorus6:
## Description

When in my playbook I have a "manipulate knowledge" to replace the Author, nothing happens. By this I mean that nothing is created / updated in my platform while in the traces of the playbook everything works correctly (it reaches the "Send for ingestion, and the log is good)

## Environment

OCTI 6.3.12

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Create a playbook with listener on "Entity type: RFI" and only on "Create" event
2. Add a component "manipulate knowlegde" with "Replace > Author > [one organization]"
3. Add a component "send for ingestion"
4. Start the playbook
5. Create manually a RFI
6. Wait for the playbook to run, and look at the execution traces
-> everything seems good, but if you look on the RFI, the author has not been updated

![Image](https://github.com/user-attachments/assets/b6c5d060-986f-44d2-b82c-0e6686f773ba)

## Additional information

- If I put steps before, like wrapper in a container, add a label, etc. nothing happens. I.e. it's not only the author that is not updated, it's the full "send for ingestion" that doesn't seem to work.
- I don't know if this behavior is only produced with a "replace > author", but for example, an "add label" works fine.

nino-filigran:
Reproduced with the following playbook: https://testing.octi.staging.filigran.io/dashboard/data/processing/automation/2f861cab-256a-4cd6-9814-2c4c5e16f823

SarahBocognano:
After investigation, the problem occurs when the confidence of the entity is superior or equal to the one of the new Author.
Exemple: 
- User A creates a Report entity with a confidence set to 80
- We are trying to set User B who has a confidence of 50 as the new Author, we will not be able to set this new value to the Report via the playbooks as his confidence his inferior to the entity
- However, if we set a User C with a confidence of 100 via playbook of this Report, the new author will change and will be set to User C

SarahBocognano:
Update after investigation 2.0:

In the code it is explicitly specify that it's impossible to upsert an author field where the author of the RFI creation have not a confidence who is stricly superior of the one of the RFI (or entity). Considering that an entity is often created with 100 confidence (if admin rights) it will be impossible to replace that field via playbooks. 

![Image](https://github.com/user-attachments/assets/7cd765a5-c50a-47fe-b1a3-cabe0d6a56cd)

[...]

PR #9261, opened 2024-12-06
[backend] fix playbook attributPathMapping for SCOs (#9226)

### LINKED ISSUE #9226: [Playbook] Label application on Observable doesn't work
## Description

Playbooks don't add labels if the **entity already contains one**. This effect seems to occur **only on Observable** objects.

## Environment

OCTI 6.4.2

## Reproducible Steps

1. Create two label "LabelA" and "LabelB"
2. Create an Observable Text and apply the LabelA on it
3. Create the following playbook
    - Listener on "update" event and "Entity type = Text"
    - Manipulate knowledge: Add -> Label -> LabelB
    - Send for ingestion
    - (start the playbook)

4. Go to the Observable Text and modify it (add a description for example)
5. Wait for the playbook to execute
-> No label added despite the correct execution of the playbook according to the execution traces

You can redo the same experiment without having a label on the observable at the beginning (so without adding "LabelA" at the start) -> you will see that it works

## Expected Output

I expect to have both labels, i.e. to have my old label LabelA, as well as my new label LabelB

## Actual Output

The new label is not added (the entity remains unchanged)

JeremyCloarec:
### Proposed changes

* labels and createdBy weren't set properly on SCOs in playbook update knowledge component

### Related issues

* #9226 

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/modules/playbook/playbook-components.ts
+++ b/opencti-platform/opencti-graphql/src/modules/playbook/playbook-components.ts
@@ -566,13 +566,11 @@ const attributePathMapping: any = {
     [ABSTRACT_STIX_RELATIONSHIP]: `/${objectMarking.stixName}`,
   },
   [INPUT_LABELS]: {
-    [ABSTRACT_STIX_DOMAIN_OBJECT]: `/${objectLabel.stixName}`,
-    [ABSTRACT_STIX_CYBER_OBSERVABLE]: `/extensions/${STIX_EXT_OCTI_SCO}/${objectLabel.stixName}`,
+    [ABSTRACT_STIX_CORE_OBJECT]: `/${objectLabel.stixName}`,
     [ABSTRACT_STIX_RELATIONSHIP]: `/${objectLabel.stixName}`,
   },
   [INPUT_CREATED_BY]: {
-    [ABSTRACT_STIX_DOMAIN_OBJECT]: `/${createdBy.stixName}`,
-    [ABSTRACT_STIX_CYBER_OBSERVABLE]: `/extensions/${STIX_EXT_OCTI_SCO}/${createdBy.stixName}`,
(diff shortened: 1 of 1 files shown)


SarahBocognano:
We can link this one because the mapping in createdBy wasn't ok. Your PR fix it :
https://github.com/OpenCTI-Platform/opencti/issues/9065

opencti-platform/opencti-graphql/src/modules/playbook/playbook-components.ts
```diff
@@ -566,13 +566,11 @@ const attributePathMapping: any = {
     [ABSTRACT_STIX_RELATIONSHIP]: `/${objectMarking.stixName}`,
   },
   [INPUT_LABELS]: {
-    [ABSTRACT_STIX_DOMAIN_OBJECT]: `/${objectLabel.stixName}`,
-    [ABSTRACT_STIX_CYBER_OBSERVABLE]: `/extensions/${STIX_EXT_OCTI_SCO}/${objectLabel.stixName}`,
```
lndrtrbn, line 570: Could you explain the change please?

JeremyCloarec: The `PLAYBOOK_UPDATE_KNOWLEDGE_COMPONENT` that manipulates knowledge uses those `attributePathMapping` to decide where to put the added labels configured in the playbook. But it was using an incorrect path to put those labels: it was putting them in the` /extensions/${STIX_EXT_OCTI_SCO}` path of stix object, instead of putting them at the root of the stix object. Because they were at the incorrect path, they were then ignored during the bundle ingestion by the worker.
So now, instead of handling them at a different path than domain objects, we now use the same path for observable and domain objects. So we can configure the path for all stix core object instead.

SarahBocognano (approved):

Issue #9214, opened 2024-12-03
Knowledge pages Details crash if vocabulary opinion_ov is empty


lndrtrbn:
## Description

If the vocabulary opinion_ov is empty, every pages Details of knowledge crash.

## Reproducible Steps

Steps to create the smallest reproducible scenario:
> ⚠️⚠️⚠️ Follow the steps below will break your platform
1. Go to Security/Taxonomies/Vocabularies/opinion_ov
2. Delete all items
3. Try to access a knowledge

## Expected Output

The platform is still usable.

## Actual Output

Every pages Details of knowledge crash.

PR #9262, opened 2024-12-06
[frontend] Knowledge pages Details crash if vocabulary opinion_ov is empty (#9214)

SarahBocognano:
### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/9214


opencti-platform/opencti-front/src/private/components/analyses/opinions/StixCoreObjectOpinionsRadar.tsx
```diff
@@ -63,12 +64,44 @@ const StixCoreObjectOpinionsRadar: FunctionComponent<StixCoreObjectOpinionsRadar
...
+      );
+    }
+  };
+
+  if (opinionOptions.length === 0) {
+    return (
+      <Chart
+        options={{
+          noData: {
+            text: t_i18n('No data available.'),
```
lndrtrbn: When I click on "No data available" I have this error appearing:
image

Which is same behavior as when we have a graph, I think we could get rid of this on click event in this case. If easy to do, otherwise forget it

lndrtrbn (approved):

Issue #9234, opened 2024-12-04
Newly created users are always assigned to the default group even if another group is specified


romain-filigran:
## Description

When a user is created, an information message appears: “_Unless specific groups are selected, user will be created with default groups only_”, which seems to indicate that if no group is specified when the user is created, the default groups will be associated. But what happens if a group is specified when the user is created? 

In my opinion, if a group is specified when the user is created, the default groups should not be associated with the user. A second suggestion would be to pre-fill the user creation form with the default groups and let the user decide whether to leave them in or delete them.

This behavior is confusing because if you don't want to create a user associated with the default group, you have to remember to manually update the user to remove the default groups.

PR #9274, opened 2024-12-10
[frontend/backend] Possibility to prevent default group assignation at user creation (#9234)

Archidoit:
### Proposed changes
- In User creation form (UI) and user add input (back) : add the attribute prevent_default_groups
  - if prevent_default_groups = false: former behavior (default groups are added to the user at its creation)
  - if prevent_default_groups = true: default groups are not added to the user

- In the user creation form, display the default groups in a tooltip

### Related issues
https://github.com/OpenCTI-Platform/opencti/issues/9234

### Docs PR
https://github.com/OpenCTI-Platform/docs/pull/244


opencti-platform/opencti-graphql/src/domain/user.js
```diff
@@ -588,23 +588,25 @@ export const addUser = async (context, user, newUser) => {
...
+      filters: [{ key: 'default_assignation', values: [true] }],
+      filterGroups: [],
+    };
+    const defaultGroups = await findGroups(context, user, { filters: defaultAssignationFilter });
+    const defaultRelationGroups = defaultGroups.edges.map((e) => ({
+      fromId: element.id,
+      toId: e.node.internal_id,
+      relationship_type: RELATION_MEMBER_OF,
+    }));
+    relationGroups = relationGroups.concat(defaultRelationGroups);
```
lndrtrbn: I find the logic of the algorithm a bit odd here and I think it could be simplified.

Why should we need to concat two arrays here as we are in a case where we know that there is no groups set in the `newUser` property.

lndrtrbn (commented):
Now that default groups are set in the form in the UI I don't think we should check on the backend if nothing is set then set the defaults. Because, in my opinion, as a user if I clear the field with the defaults, it's not to see the user created with defaults anyway after all.

Would be nice to have product opinion

Archidoit:
> Now that default groups are set in the form in the UI I don't think we should check on the backend if nothing is set then set the defaults. Because, in my opinion, as a user if I clear the field with the defaults, it's not to see the user created with defaults anyway after all.
> 
> Would be nice to have product opinion

It was in case users are created from the API... If no groups are specified, the user should be added to the default groups (we don't create users with no groups). @nino-filigran @romain-filigran What do you think?

opencti-platform/opencti-graphql/src/domain/user.js
Archidoit: indeed ! gonna improve that

nino-filigran:
I agree that we need to also have a check at BE side in case you create through API only. Your comment makes sense @Archidoit

SouadHadjiat:
could you update user-test to test these changes ? There is at least the part "User has no capability query behavior" that removes the default_assignation, which is not necessary anymore. And maybe add a test to make sure when we create a user with a specific group, we don't assign the default group to the user.

Archidoit:
> could you update user-test to test these changes ? There is at least the part "User has no capability query behavior" that removes the default_assignation, which is not necessary anymore. And maybe add a test to make sure when we create a user with a specific group, we don't assign the default group to the user.

@SouadHadjiat  I don't understand why you want we to remove the default_assignation of 'User has no capability query behavior'. If so, the user will be affected to the default group since there are no groups specified in the user to create input.

Archidoit:
@SouadHadjiat tests written :)

Archidoit:
Waiting for Product decision about the functional breaking change (@romain-filigran )

Archidoit:
Functional decision:
- add a parameter (False by default) prevent_default_group. If false, it's the former behavior, the default groups are added. If true, the default groups are not added
- this parameter can be used in the SSO configuration

aHenryJard (approved):
Tested locally, both with local provider and SAML provider. All good to me.

Issue #9263, opened 2024-12-07
Missing button to generate Observables in container


Lhorus6:
## Description

In the context of a container, when I am in the "Entities" tab, I have access to all the contained entities. If I have Indicators inside and I want to generate observables, the button does not appear if I do not have an "Entity type = Indicator" filter

## Environment

OCTI 6.4.3

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Create a Report and go
2. Add only few Indicators in the Entities tab
3. Refresh the page
4. Select one Indicator, then "select all"
-> In both case, you don't have the button

![Image](https://github.com/user-attachments/assets/b107bb85-74cb-4957-b505-c1a99847e268)

5. Now added the filter "Entity type = Indicator"
6. Select an Indicator, or "select all"
-> In both cases, the button is now available

![Image](https://github.com/user-attachments/assets/1c407da0-d3fa-4c96-95c8-7567d8ffe6da)

## Expected Output

Have the button if I only have Indicators selected, without needing to apply a filter

Archidoit:
@Lhorus6 it's not a bug but a new feature... in a list with several entity types, if you select 'selectAll' and no filters, the entity types of the selected entities is not taken into account. Same for all the background tasks in Data>Entities for instance

Lhorus6:
Hi @Archidoit,
Okay, I can understand for the “select all”, but why is it that if I manually select one or more Indicators, I don't get the button either?

nhuber0724:
Marriott tested applying a filter, and then selecting indicators and clicking on generate observable. Marriott advised when they did this, it still does not work. The platform did not convert the indicator to an observable.

nhuber0724:
If you select only one indicator the generate observable button will not appear. For the button to show up, you have to select all.

PR #9302, opened 2024-12-11
[frontend] Missing button to generate Observables in container (#9263)

SarahBocognano:
### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/9263

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/data/DataTableToolBar.jsx
+++ b/opencti-platform/opencti-front/src/private/components/data/DataTableToolBar.jsx
@@ -1887,10 +1887,10 @@ class DataTableToolBar extends Component {
           const isOnlyStixCyberObservablesTypes = entityTypeFilterValues.length > 0
             && entityTypeFilterValues.every((id) => stixCyberObservableTypes.includes(id));
 
-          const cleanedSelectedTypes = selectedTypes.filter((type) => type !== 'Stix-Domain-Object' && type !== 'Stix-Core-Object');
+          const promotionTypesFiltered = entityTypeFilterValues.length > 0
+            && entityTypeFilterValues.every((id) => promotionTypes.includes(id));
 
-          const promotionTypesFiltered = cleanedSelectedTypes.length > 0
-            && cleanedSelectedTypes.every((type) => promotionTypes.includes(type));
+          const cleanedSelectedTypes = selectedTypes.filter((type) => type !== 'Stix-Domain-Object' && type !== 'Stix-Core-Object');
 
           const isManualPromoteSelect = !selectAll
             && cleanedSelectedTypes.length > 0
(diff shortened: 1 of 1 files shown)


opencti-platform/opencti-front/src/private/components/data/DataTableToolBar.jsx
```diff
@@ -1692,7 +1692,7 @@ class DataTableToolBar extends Component {
 
           const isManualPromoteSelect = !selectAll
             && selectedTypes.length > 0
-            && selectedTypes.every((type) => promotionTypes.includes(type));
+            && selectedTypes.some((type) => promotionTypes.includes(type));
```
Archidoit: I don't understand your fix. With it, if only some selected entities are Indicators, the button will be available.... we should have all selected entities of type Indicator/observables to have the button

Archidoit:
If I select a vulnerability and an indicator, the button is available whereas it should not 
image

Archidoit:
Suggestion: clean eventual 'useless' types (abstract types if a subtype exist) to make the .every work

exemple here: 
image

we shouldn't take Stix-Domain-Object into account
image

opencti-platform/opencti-front/src/private/components/data/DataTableToolBar.jsx
```diff
@@ -1690,9 +1690,13 @@ class DataTableToolBar extends Component {
           const promotionTypesFiltered = entityTypeFilterValues.length > 0
             && entityTypeFilterValues.every((id) => promotionTypes.includes(id));
 
+          const selectedTypesFiltered = selectedTypes.some((type) => type !== 'Stix-Domain-Object')
```
Archidoit: you should check the other types are stix domain objects before removing them.... (it can be an observable for instance, and in this case you shouldn't remove 'Stix-Domain-Object')

Archidoit:
```suggestion
          const cleanedSelectedTypes = selectedTypes.some((type) => type !== 'Stix-Domain-Object')
```

Archidoit (changes requested):

Archidoit:
In Bulk Search, when I select an observable, the generate button is not present 

image


If I select the observable in the Observable tab, the button is present : 

image

Archidoit:
If I select an indicator in Bulk Search, I do not have the button either 
image

Archidoit:
In the Bulk Search, with a filter entity type = Indicator, the Manual Promote button is not available : 
image


Whereas it is in Data > Entities:
image

Archidoit:
Working for manual multi selection of entities

Archidoit:
In Bulk Search, I select observables and indicators. The button is available. But the message in the panel seems to say there are only indicators selected ... 

Image

Is it normal? do we want to open the panel if both indicators and observables are selected?

Archidoit (approved):

PR #9304, opened 2024-12-11
[frontend] Align knowledge bar counters (#8115)

### LINKED ISSUE #8115: Incorrect display in Threat Actors menu category
## Description

Hello, 

I have created a relationship between two Threat Actors objects, being A the main one and B the secondary one. This relationship is ‘related to’ so, in the main object A, this relationship can be seen in the Other > Related Entities category menu. However, the interface seems to show the posting of this relationship in the Threat Actors menu category (as I show you in the image it indicates a (1)). 

We understand that it may be a bug in the interface, however, from our point of view, we consider that it is more intuitive in reality if there is a relationship between two actors, even though in this case it is of type ‘related to’, to appear within the Threat Actors category instead of, as it is now contemplated, in Related Entities.

Thank you very much in advance for your attention, best regards.

## Reproducible Steps

1. Create a Threat Actor object, it will be the secondary object (B).
2. Create a second Threat Actor object, it will be the primary one (A).
3. Access the primary Threat Actor object, section ‘Knowledge’. 
4. In the right context menu, go to ‘Other’ > ‘Related Entities’.
5. Create a relationship between the two Threat Actors and make sure it is of type ‘Related to’.
6. Observe the number that appears in the right context menu in the category ‘Threats’ > ‘Threat Actors ’1. Create a Threat Actor object, it will be the secretary (B)

## Expected Output

A ‘(1)’ should appear in the Related Entities category or the relationship between two actors, even if it is ‘Related To’, should appear as a relationship in the Knowledge Threat Actors part.

## Actual Output

A ‘(1)’ should appear in the Threat Actors category

## Screenshots (optional)

imagen

imagen

No.

### LINKED ISSUE #8672: Missing entities listed in Knowledge tab
## Description

Depending on the relationship, the channels are not listed in the same place in my Knowledge tab.

## Environment

OCTI 6.3.6

## Reproducible Steps

1. Create an Incident
2. Create a channel and the relation "Incident -> related to -> Channel"
3. Create another channel and the relation "Incident -> uses -> Channel"

image

4. Go on the Incident, then the Knowledge tab

image

image

## Expected Output

See all my Channels in the Channel section

## Additional information

If we define in which menu the entity is displayed based on the relationship, we may have inconsistencies like this with other entity types

### LINKED ISSUE #9057: In an Incident, the Observables count on Knowledge is one below the actual count
## Description

On the Knowledge screen of an incident, if there are 2 or more observables included, then the count in brackets will be one lower than the total

## Environment

1. OS (where OpenCTI server runs): SaaS
2. OpenCTI version: 6.3.13
3. OpenCTI client: Edge on MacOS
4. Other environment details:

## Additional information

Example in demo [here](https://demo.octi.filigran.io/dashboard/events/incidents/bf6b8c66-50fa-4e30-96ee-44a8d8b53248/knowledge/observables)

## Screenshots (optional)

Image

delemaf:
### Proposed changes

* Fix distribution counters, also add counter for related entities
* Refactor query with fragment for knowledge bar
* NB: counters update is still not sync when entity is added (it's another bug 🐛 )

### Related issues

* Close https://github.com/OpenCTI-Platform/opencti/issues/8115
* Close https://github.com/OpenCTI-Platform/opencti/issues/9057
* Close https://github.com/OpenCTI-Platform/opencti/issues/8672

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectKnowledgeBar.jsx
+++ b/opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectKnowledgeBar.jsx
@@ -33,7 +33,10 @@ const useStyles = makeStyles((theme) => ({
 }));
 
 const stixCoreObjectKnowledgeBarFragment = graphql`
-  fragment StixCoreObjectKnowledgeBar_stixCoreObject on StixCoreObject {
+  fragment StixCoreObjectKnowledgeBar_stixCoreObject on StixCoreObject
+  @argumentDefinitions(
+    relatedRelationshipTypes: { type: "[String]", defaultValue: ["related-to"] }
+  ) {
     # distribution of entities without "related to" relationship
     relationshipsWithoutRelatedToDistribution: stixCoreRelationshipsDistribution(
       field: "entity_type"
@@ -56,11 +59,11 @@ const stixCoreObjectKnowledgeBarFragment = graphql`
       label
       value
     }
-    # distribution of entities with "related to" relationship
-    relationshipsRelatedToDistribution: stixCoreRelationshipsDistribution(
+    # distribution of entities with relatedRelationshipTypes ("related to" relationship by default)
+    relationshipsRelatedDistribution: stixCoreRelationshipsDistribution(
       field: "entity_type"
       operation: count
-      relationship_type: [ "related-to" ]
+      relationship_type: $relatedRelationshipTypes
     ) {
       label
       value
@@ -87,7 +90,7 @@ const StixCoreObjectKnowledgeBar = ({
   const location = useLocation();
   const { bannerSettings, schema } = useAuth();
   const isInAvailableSection = (sections) => availableSections.some((filter) => sections.includes(filter));
-  const { relationshipsWithoutRelatedToDistribution, relationshipsRelatedToDistribution, stixCoreObjectsDistribution } = useFragment(
+  const { relationshipsWithoutRelatedToDistribution, relationshipsRelatedDistribution, stixCoreObjectsDistribution } = useFragment(
     stixCoreObjectKnowledgeBarFragment,
     data,
   );
@@ -98,7 +101,7 @@ const StixCoreObjectKnowledgeBar = ({
     : {});
 
   const statisticsWithoutRelatedToRelationship = indexEntities(relationshipsWithoutRelatedToDistribution);
-  const statisticsRelatedToRelationship = indexEntities(relationshipsRelatedToDistribution);
+  const statisticsRelatedRelationship = indexEntities(relationshipsRelatedDistribution);
   const statisticsCoreObjects = indexEntities(stixCoreObjectsDistribution);
 
   const sumEntitiesByKeys = (stats, keys) => {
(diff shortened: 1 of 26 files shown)


opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectKnowledgeBar.jsx
```diff
@@ -32,25 +33,60 @@ const useStyles = makeStyles((theme) => ({
...
+      }
+    }
+  }
+`;
+
 const StixCoreObjectKnowledgeBar = ({
   stixCoreObjectLink,
   availableSections,
-  stixCoreObjectsDistribution,
+  queryRef,
```
marieflorescontact: do we really want to call this prop `queryRef`? it seems weird as the value we use when calling `<StixCoreObjectKnowledgeBar>` is the result of `usePreloadedQuery()` that uses a queryRef as parameter.

delemaf: You're right `data` or `dataRef` may be more explicit 👍

opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectKnowledgeBar.jsx
```diff
@@ -32,25 +32,82 @@ const useStyles = makeStyles((theme) => ({
...
+  const sumEntitiesByKeys = (keys, stats) => keys
+    .map((key) => stats[key]?.value || 0)
+    .reduce((acc, val) => acc + val, 0);
+
+  const statisticsThreats = sumEntitiesByKeys(['Threat-Actor-Individual', 'Threat-Actor-Group', 'Intrusion-Set', 'Campaign', 'Incident'], statisticsRelationship);
+  const statisticsThreatActors = sumEntitiesByKeys(['Threat-Actor-Individual', 'Threat-Actor-Group'], statisticsRelationship);
+  const statisticsVictims = sumEntitiesByKeys(['Sector', 'Organization', 'Individual', 'Region', 'Country', 'City', 'Position', 'Administrative-Area'], statisticsRelationship);
+  const statisticsAttributions = sumEntitiesByKeys(attribution ?? [], statisticsRelationship);
+  const statisticsLocations = sumEntitiesByKeys(['Region', 'Country', 'City', 'Position', 'Administrative-Area'], statisticsRelationship);
+  const statisticsObservables = sumEntitiesByKeys([...schema.scos.map((s) => s.id), 'Stixfile', 'Ipv4-Addr', 'Ipv6-Addr'], statisticsCoreObjects);
```
aHenryJard: I have 2 questions on this part that are not blockers at all:
- Do we have this kind of statistic in another place ?
- This information about groups, it's not in shemas that is send from backend to frontend ?

Also I wonder if it's possible to add a frontend test on that :)

delemaf: Thanks @aHenryJard !
- This kind of statistics seems to only be in this place
- What do you mean `information about groups`, I think this kind of request is not implemented (from what I know)
I'll try to see if this could be tested easily 👍

marieflorescontact (commented):
locally tested ok 👌
just a question: do we decide to move the counter from the `Observables` side to the `Related entities` side?

opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectKnowledgeBar.jsx
aHenryJard: There is a definition of shema that is send to frontedn in setting, that you can get with the useAuth hook
https://github.com/OpenCTI-Platform/opencti/blob/caf8982c841d71cce482700e714eb50873a9b858/opencti-platform/opencti-front/src/utils/hooks/useAuth.ts#L26-L34

I don't think it's inside but I was wondering if the fact that for example this list of type "'Sector', 'Organization', 'Individual', 'Region', 'Country', 'City', 'Position', 'Administrative-Area'" is grouped as Victim is inside (in which case instead of a list we could use list from shema information)

Archidoit:
If I am in a threat actor group and add 2 relationships with other threat actor groups, their are 2 relationships in the list, but the right counter indicates 3 : 
image

Archidoit:
Relationships of tyep 'part of' are displayed when I am in the 'related to' section : 
image

delemaf:
> Relationships of tyep 'part of' are displayed when I am in the 'related to' section : image

That's normal, it looks like there're 2 exceptions in related:
- Threat Actors Group have `related-to` & `part-of` 
- Threat Actors Individual have` related-to`, `part-of`, `impersonates` & `known-as`
It now fixed, I've yet handled this case 😉

Archidoit:
I have 3 relationships in Victimology but the Victimology counter indicates 4 
image

Archidoit (approved):

PR #9316, opened 2024-12-12
[frontend] Add enroll/enrich to note & grouping popover menu (#9217)

### LINKED ISSUE #9217: Cannot manually enrich a Note or Grouping and cannot enroll them in a playbook
## Description

Cannot manually enroll a Note in a playbook (button missing).

Same for the Grouping (and maybe other).

## Environment

OCTI 6.4.2

## Reproducible Steps

1. Go on a Note entity
2. Try to enroll it in a playbook

#### On Report entity

Image

#### On Note entity

Image

## Additional information

There's no “enrich” button either. I don't know of any enrichment connector for these entity type, but to anticipate a possible problem and harmonize the platform 🤷‍♂ 

For example, with the enrichment connector enabling Splunk retrohunt as report enrichment, this could also apply to Groupings. But at present, we can't enrich Groupings (no button).

delemaf:
### Proposed changes

* Add enroll/enrich to **note** popover menu
* Add enroll/enrich to **grouping** popover menu
* Add enroll to container header (ff `FAB_REPLACEMENT `)
* Fix when you click a second time on button enroll or enrich

image
image

### Related issues

* Close https://github.com/OpenCTI-Platform/opencti/issues/9217

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/analyses/groupings/GroupingPopover.jsx
+++ b/opencti-platform/opencti-front/src/private/components/analyses/groupings/GroupingPopover.jsx
@@ -95,9 +95,7 @@ const GroupingPopover = ({ id }) => {
               {t_i18n('Enrich')}
             </MenuItem>
           </Security>
-          <Security needs={[KNOWLEDGE_KNENRICHMENT]}>
-            <MenuItem onClick={handleOpenEnroll}>{t_i18n('Enroll in playbook')}</MenuItem>
-          </Security>
+          <MenuItem onClick={handleOpenEnroll}>{t_i18n('Enroll in playbook')}</MenuItem>
           <Security needs={[KNOWLEDGE_KNUPDATE_KNDELETE]}>
             <MenuItem onClick={handleOpenDelete}>{t_i18n('Delete')}</MenuItem>
           </Security>
--- a/opencti-platform/opencti-front/src/private/components/analyses/groupings/GroupingPopover.jsx
+++ b/opencti-platform/opencti-front/src/private/components/analyses/groupings/GroupingPopover.jsx
@@ -10,6 +10,7 @@ import DialogContent from '@mui/material/DialogContent';
 import DialogContentText from '@mui/material/DialogContentText';
 import MoreVert from '@mui/icons-material/MoreVert';
 import { graphql } from 'react-relay';
+import EEChip from '../../common/entreprise_edition/EEChip';
 import StixCoreObjectEnrichment from '../../common/stix_core_objects/StixCoreObjectEnrichment';
 import StixCoreObjectEnrollPlaybook from '../../common/stix_core_objects/StixCoreObjectEnrollPlaybook';
 import { useFormatter } from '../../../../components/i18n';
@@ -20,6 +21,7 @@ import Security from '../../../../utils/Security';
 import { KNOWLEDGE_KNENRICHMENT, KNOWLEDGE_KNUPDATE_KNDELETE } from '../../../../utils/hooks/useGranted';
 import Transition from '../../../../components/Transition';
 import useHelper from '../../../../utils/hooks/useHelper';
+import useEnterpriseEdition from '../../../../utils/hooks/useEnterpriseEdition';
 
 const GroupingPopoverDeletionMutation = graphql`
   mutation GroupingPopoverDeletionMutation($id: ID!) {
@@ -30,13 +32,14 @@ const GroupingPopoverDeletionMutation = graphql`
 const GroupingPopover = ({ id }) => {
   const navigate = useNavigate();
   const { t_i18n } = useFormatter();
+  const { isFeatureEnable } = useHelper();
+  const isFABReplaced = isFeatureEnable('FAB_REPLACEMENT');
+  const isEnterpriseEdition = useEnterpriseEdition();
   const [anchorEl, setAnchorEl] = useState(null);
   const [displayDelete, setDisplayDelete] = useState(false);
   const [displayEdit, setDisplayEdit] = useState(false);
   const [displayEnrichment, setDisplayEnrichment] = useState(false);
   const [displayEnroll, setDisplayEnroll] = useState(false);
-  const { isFeatureEnable } = useHelper();
-  const isFABReplaced = isFeatureEnable('FAB_REPLACEMENT');
   const [deleting, setDeleting] = useState(false);
   const handleOpen = (event) => setAnchorEl(event.currentTarget);
   const handleClose = () => setAnchorEl(null);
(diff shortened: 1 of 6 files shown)


CelineSebe:
Buttons works 👌  , enroll/enrich are implemented to note and grouping. When I try to enroll, I don't see the playbook I previously created:
image

delemaf:
> Buttons works 👌 , enroll/enrich are implemented to note and grouping. When I try to enroll, I don't see the playbook I previously created: image

Playbook show up for me (you have to start your playbook to see it).
image

lndrtrbn (commented):
I suppose we should also have the buttons when the feature flag `FAB_REPLACEMENT` is activated? For now we have only enrich but not enroll playbook
image

opencti-platform/opencti-front/src/private/components/analyses/groupings/GroupingPopover.jsx
```diff
@@ -58,6 +62,21 @@ const GroupingPopover = ({ id }) => {
     handleClose();
   };
   const handleCloseEdit = () => setDisplayEdit(false);
+  const handleOpenEnrichment = () => {
```
Kedae, line 65: When I see a lot of duplicate code like this we are going to put in a lot of places I'm always I bit stressed : what does Landry or Staff about that ?

lndrtrbn: Well we could have multiple approaches here depending on what we want.
- A very generic hook allowing to encapsulate a `true/false` state and calling an instance of this hook for every `MenuItem` that opens a thing,
- A specific hook that encapsulate both the open state and the components to render inside the `Menu` and `Drawer`.

Not sure having a hook returning components is the easiest thing to use though.

lndrtrbn: I'll have a try on the first one and push it directly to the PR to have an idea

lndrtrbn: In fact something very similar to `useDeletion` hook but no context specific

lndrtrbn: > I'll have a try on the first one and push it directly to the PR to have an idea

Bad idea, was improving nothing. And after some thoughts I don't think I find a relevant solution that simplifies stuff. I prefer having those repetition of states rather than a complex generic component that needs to manage every cases. At least here it is simple to understand and tweak if needed.

delemaf: It might be a big refactor but could we make a generic Popover component, logic seems very similar, we could pass EditionContainer component as props, or is there tricky parts that we cannot mutualize?

lndrtrbn: Not a fan of doing it as we will obviously run into edges cases, and start adding a lot of complexity and "if"s in the generic popover. I'm not for over-engineer this thing, but keep things simple and easy to customize for every popover

delemaf:
After discussing with @labo-flg and @nino-filigran, we concluded that the views should be aligned consistently, particularly regarding the enrich and enroll popovers:
- [ ] Reports
- [ ] Groupings
- [ ] Malware Analyses
- [ ] Notes
- [ ] Incidents
- [ ] Observed Data
- [ ] Observables
- [ ] Artifact
- [ ] Indicators
- [ ] Infrastructures
- [ ] all Threats
- [ ] all Arsenal
- [ ] all Entities  (except systems)
- [ ] Systems
- [ ] all Locations

delemaf:
> I suppose we should also have the buttons when the feature flag `FAB_REPLACEMENT` is activated? For now we have only enrich but not enroll playbook image

Enroll playbook button added to container header (ff `FAB_REPLACEMENT `)

I've joined a small fix when you click a second time on button enroll or enrich drawer now open (before, it was opening on first click but not on second click after closing)

opencti-platform/opencti-front/src/private/components/analyses/notes/NotePopover.tsx
```diff
@@ -129,13 +148,23 @@ const NotePopover: FunctionComponent<NotePopoverProps> = ({
               {t_i18n('Remove from this entity')}
             </MenuItem>
           )}
+          <Security needs={[KNOWLEDGE_KNENRICHMENT]}>
+            <MenuItem onClick={handleOpenEnrichment}>
+              {t_i18n('Enrich')}
+            </MenuItem>
+          </Security>
+          <Security needs={[KNOWLEDGE_KNENRICHMENT]}>
```
labo-flg: there is no capability for the enroll action.

opencti-platform/opencti-front/src/private/components/analyses/groupings/GroupingPopover.jsx
```diff
@@ -71,10 +90,20 @@ const GroupingPopover = ({ id }) => {
         </ToggleButton>
         <Menu anchorEl={anchorEl} open={Boolean(anchorEl)} onClose={handleClose}>
           <MenuItem onClick={handleOpenEdit}>{t_i18n('Update')}</MenuItem>
+          <Security needs={[KNOWLEDGE_KNENRICHMENT]}>
+            <MenuItem onClick={handleOpenEnrichment}>
+              {t_i18n('Enrich')}
+            </MenuItem>
+          </Security>
+          <Security needs={[KNOWLEDGE_KNENRICHMENT]}>
```
labo-flg: no capa for enroll

opencti-platform/opencti-front/src/private/components/analyses/notes/NotePopover.tsx
```diff
@@ -129,13 +151,21 @@ const NotePopover: FunctionComponent<NotePopoverProps> = ({
               {t_i18n('Remove from this entity')}
             </MenuItem>
           )}
+          <Security needs={[KNOWLEDGE_KNENRICHMENT]}>
+            <MenuItem onClick={handleOpenEnrichment}>
+              {t_i18n('Enrich')}
+            </MenuItem>
+          </Security>
+          <MenuItem onClick={handleOpenEnroll}>{t_i18n('Enroll in playbook')}{!isEnterpriseEdition && <EEChip />}</MenuItem>
```
labo-flg: EEChip already return null if isEntrepriseEdition.

Also, if we add this here we need ti in every popover on all entities.

AS we wil move rapidly to quickactions I suggest we remove the chip entirely to avoid some cumbersome work that will not stay long.

labo-flg (approved):

PR #9328, opened 2024-12-12
[backend] Migrate graphQL playground to GraphiQL (#7363)

### LINKED ISSUE #7363: Apollo GraphQL Playground/v3 Deprecation
## Description

The Apollo GraphQL Playground is both still present _and_ enabled in the OpenCTI images built/distributed as of v6.1.10.  Playground has been EOL since 2022-12-31 (https://www.apollographql.com/docs/apollo-server/v2/testing/graphql-playground/) and its existence in current builds appears to just be an overlooked artifact from the upgrade from Apollo 2.x to 3.x here in the past.

Furthermore, it should be noted that Apollo 3.x is EOL as of 2024-10-22 requiring an upgrade to Apollo 4.x before that date (https://www.apollographql.com/docs/apollo-server/migration).

## Environment

1. Ubuntu 22.04.4 LTS
2. OpenCTI 6.1.10
3. Frontend
4. Other environment details:

## Reproducible Steps

In a browser just visit https[:]//youropencti[.]url **/graphql**

## Expected Output

Having Playground available in production was generally considered a security misconfiguration (and risk) in Apollo 2.x, and there is no reason to have it present in any environment in Apollo 3.x.  Given its history including high severity XSS, it should be removed completely from released builds/images.

## Actual Output

Playground is still present in distributed builds/images, providing no benefit and introducing potential risk of exploitation.

## Additional information

You can close this related 2.5 year old issue at the same time.
https://github.com/OpenCTI-Platform/opencti/issues/1835

## Screenshots (optional)

image

Archidoit:
### Related issues
closes #7363 

Replace graphQL playground (deprecated) by GraphiQL

documentation PR : https://github.com/OpenCTI-Platform/docs/pull/236


opencti-platform/opencti-front/src/private/components/profile/ProfileOverview.jsx
```diff
@@ -670,7 +670,7 @@ const ProfileOverviewComponent = (props) => {
             variant="contained"
             color="primary"
             component={Link}
-            to="/graphql"
+            to="/public/sandbox"
```
aHenryJard: to remove

opencti-platform/opencti-graphql/src/graphql/graphql.js
```diff
@@ -103,6 +101,10 @@ const createApolloServer = () => {
     csrfPrevention: false, // CSRF is handled by helmet
     tracing: DEV_MODE,
     plugins: apolloPlugins,
+    cors: {
+      origin: 'http://localhost:4000',
+      credentials: true,
+    },
```
aHenryJard: to be removed ?

opencti-platform/opencti-front/src/private/components/profile/ProfileOverview.jsx
Archidoit: done

opencti-platform/opencti-graphql/src/graphql/graphql.js
Archidoit: done

opencti-platform/opencti-graphql/src/graphql/graphql.js
```diff
@@ -103,6 +101,10 @@ const createApolloServer = () => {
     csrfPrevention: false, // CSRF is handled by helmet
     tracing: DEV_MODE,
     plugins: apolloPlugins,
+    /* cors: {
```
Archidoit: to remove

opencti-platform/opencti-graphql/src/http/httpPlatform.js
```diff
@@ -134,7 +134,7 @@ const createApp = async (app) => {
 
   const ancestorsFromConfig = nconf.get('app:public_dashboard_authorized_domains')?.trim() ?? '';
   const frameAncestorDomains = ancestorsFromConfig === '' ? "'none'" : ancestorsFromConfig;
-  const allowedFrameSrc = ["'self'"];
+  const allowedFrameSrc = ["'self'", 'https://sandbox.embed.apollographql.com'];
```
Archidoit: add only if playground enabled

opencti-platform/opencti-graphql/src/http/httpPlatform.js
```diff
@@ -164,7 +164,7 @@ const createApp = async (app) => {
   app.use(compression({}));
 
   if (ENABLED_UI) {
-    // -- Serv playground resources
+    // -- Serv playground resources // TODO
     app.use(`${basePath}/static/@apollographql/graphql-playground-react@1.7.42/build/static`, express.static('static/playground'));
```
Archidoit: to remove

opencti-platform/opencti-graphql/src/http/httpServer.js
```diff
@@ -93,13 +93,41 @@ const createHttpServer = async () => {
...
...26YoEZ14Pvz+UYyITjwHAAAAAAAAAABFY5/1TxQ2ohNvz6awEZ35j0TxIjrxfNTbpqARnXg+sv2jkBGdeA4AAAAAAAAAAIrKHtGfKkxEJ96OVGEiOvMfjcJHdOL5mLdNQSI68Xx0+0chIjrxHAAAAAAAAAAAFJ01qj/dvK9RynZIahzn/y/xdizmtDcoow5J05i/Ic3tP5S0NM/flXiel23T8XuS/wtJ4XF+J+L52PePZyXdkofvRDwHAKAUnPPim7Jqm4z/HF7/IXVfN5cNAgAAAAAA8sEa9d8Yf0Qn3o7H+CM68x+v5vYfSLo1T9+NeJ7XbTPuiE48H//+Md6ITjwHAAAAAAAAAADGWGP6W2OP6MTbfBh7RGf++ZKfiE48L8i2GXNEJ57nb/8Ya0QnngMAAAAAAAAAAKOsMf/N0Ud04m0+jT6iM/98G19EJ54XdNuMOqITz/O/f4w2ohPPAQAAAAAAAACAcda4/vbIIzrxthBGHtGZf6GMLaITz4uybUYc0Ynnhds/RhrRiecAAAAAAAAAACAQrHF/h7NHdOJtIZ09ojP/QhtdRCeeF3XbnDWiE88Lv3+cLaITzwEAAAAAAAAAQGBYefkuQ0d04m0xDB3RmX+xjCyiE8+NbJshIzrxvHj7x1ARnXgOAAAAAAAAAAACxc7Ld+k8PyaF5kqKnfJV4m2xvDGvV2HNldR9yldTsu0rmH+RdM67TdIPh/kTxHNj22buRsn6sKTMKV8lnhd3/1gi6bnTvko8BwAAAAAAAAAAgWPl9budPBO9VsTz4jt5Jvok2fYHdODCNoZSZGc+E514Hoht8/aZ6F3Ec2P7R+5MdOI5AAAAAAAAAACoEM37GtW8ey6DYP6Vuw3af6Dmdv/EP8c0ffe5DCUgZu5aoMhrNQzC5P7RcQVDAAAAAAAAAAAAACrJYEQnngMAAAAAAAAAAAAAwJnOAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAACgRP1/czqtTs2HjBMAAAAASUVORK5CYII=" alt="OpenCTI" width="250" style="vertical-align: middle; clear: both; width: 250px; max-width: 250px; padding-top: 40px; padding-bottom: 40px;"></div>
```
richard-julien: We need to find a way to use the already define logo and not use it directly in base64

richard-julien: Even better to create a simple page that will redirect auto?

Archidoit: we'll do an automatic redirection then :)

Archidoit:
Todo: 
- remove the fallback page
- some css style to avoid vertical scrollbar and see directly variables and headers

opencti-platform/opencti-front/src/public/components/PublicTopBar.tsx
```diff
@@ -36,7 +36,7 @@ const PublicTopBar = () => {
               pointerEvents: 'none',
             }}
           >
-            {t_i18n('Public dashboard')}
+            {t_i18n(title)}
```
lndrtrbn: I'm not a fan of passing string in props that are translated later on. Prefer instead giving in props directly the translated string. Two reasons:
- Better code readability because we can see directly that a string is translated or not,
- I'm not sure our script to catch missing translations will work with dynamic props like this

opencti-platform/opencti-graphql/src/http/httpServer.js
```diff
@@ -93,13 +93,26 @@ const createHttpServer = async () => {
...
   }, wsServer);
+
   apolloServer.addPlugin(ApolloServerPluginDrainHttpServer({ httpServer }));
   apolloServer.addPlugin({
     async serverWillStart() {
       return {
         async drainServer() {
           serverCleanup.dispose();
         },
+        async renderLandingPage() {
```
lndrtrbn, line 104: What is this for?

Archidoit: It handles the redirection of /graphql to public/graphql (otherwise, /graphql redirect to Sandbox, which is built in with apollo v4)

opencti-platform/opencti-front/src/public/components/PublicTopBar.tsx
Archidoit: Indeed that's better this way, it's fixed :)

labo-flg:
I'm working on some improvements to the styling.

Right now we have an ugly scrollbar because the editor takes 100%. It should be 100% minus the top bar.

Also, selecting text with current styling in the editor or result panel is not visible (wrong coloring).

Archidoit:
> I'm working on some improvements to the styling.
> 
> Right now we have an ugly scrollbar because the editor takes 100%. It should be 100% minus the top bar.
> 
> Also, selecting text with current styling in the editor or result panel is not visible (wrong coloring).

Mmm normally we fixed that with @lndrtrbn today ...

And yes good idea for the selection

labo-flg:
yes , ok for the window height !

selection is still invisible though
There is a color `!important` set on `body pre` that I cannot get rid of.

I'm pushing some minor improvements on the default texts and queries.

labo-flg:
I'm speaking about this dark background here;
image

opencti-platform/opencti-front/src/public/components/Playground.tsx
```diff
@@ -0,0 +1,109 @@
...
+      }}
+    >
+      <PublicTopBar title={t_i18n('GraphQL playground')} />
+      <GraphiQL
+        fetcher={fetcher}
+        defaultTabs={[
+          { query: placeholder },
+          { query: exampleWithVariables, variables: JSON.stringify(exampleVariables, null, '  ') },
+        ]}
+        forcedTheme={'dark'}
```
labo-flg, line 103: we do not handle light theme in the app bar, so I think this is more suitable to force dark mode. Doing so remove the option from graphiql settings too.

opencti-platform/opencti-front/src/public/components/Playground.tsx
```diff
@@ -0,0 +1,109 @@
...
+  const { t_i18n } = useFormatter();
+  return (
+    <Box
+      sx={{
+        height: '100vh',
+        display: 'flex',
+        flexDirection: 'column',
+        pre: {
+          backgroundColor: 'transparent !important',
+        },
```
labo-flg, line 93: This is strange but a css rule on background color is injected somehow as `!important` .
This makes the code block color darker and, more problematic, text selection is invisible.
I did not find where this rule comes from... So I overwrite it here.

opencti-platform/opencti-front/lang/front/fr.json
```diff
@@ -1384,6 +1384,7 @@
   "Graph": "Graph",
   "Graph type": "Type de graphique",
   "Graph view": "Vue graphe",
+  "GraphQL playground": "Terrain de jeu GraphQL",
```
marieflorescontact: should we use "Playground"? "Terrain de jeu" seems weird
image
well  this is not a big deal

marieflorescontact (commented):
works fine (and we can now select text 👌)

opencti-platform/opencti-front/lang/front/fr.json
Archidoit: fixed :)

labo-flg (approved):

Issue #9264, opened 2024-12-07
The default "Latest Reports" dashboard widget does not list recent reports


romain-filigran:
## Description

On the default/main platform dashboard, the “Latest reports” widget no longer seems to display the latest recent reports, but old reports.

Screenshot from demo public instance

![Image](https://github.com/user-attachments/assets/54ea88bd-f874-4158-8909-f469693b3444)

## Environment

2. OpenCTI version: 6.4.3

ckane:
I'm seeing this behavior too, doing browser inspection, looks like the list is sorted by "asc" instead of descending order:
![Image](https://github.com/user-attachments/assets/b644c3f5-1162-40b9-8390-9c3a1f16244d)

PR #9333, opened 2024-12-12
[frontend] The default "Latest Reports" dashboard widget list recent reports, not the oldest one

CelineSebe:
### Proposed changes

* change orderMode from 'asc' to 'desc' 

### Related issues

* #9264 

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectsList.jsx
+++ b/opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectsList.jsx
@@ -258,7 +258,7 @@ const StixCoreObjectsList = ({
           types: dataSelectionTypes,
           first: selection.number ?? 10,
           orderBy: sortBy,
-          orderMode: selection.sort_mode ?? 'desc',
+          orderMode: selection.sort_mode ?? 'asc',
           filters,
         }}
         render={({ props }) => {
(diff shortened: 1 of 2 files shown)


opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectsList.jsx
```diff
@@ -258,7 +258,7 @@ const StixCoreObjectsList = ({
           types: dataSelectionTypes,
           first: selection.number ?? 10,
           orderBy: sortBy,
-          orderMode: selection.sort_mode ?? 'asc',
```
Kedae: Not sure about that fix, I think you should change the configuration of "Latest Report" widget in Dashboard.jsx directly

Kedae (changes requested):

opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectsList.jsx
CelineSebe: Ok thanks, I look in Dashboard.jsx Until 11 november commit, it was 'desc' before, do you know why this was change?

Kedae: I don't know, but it seems "wanted" by Sam.
If we want this specific widget "Latest Reports" to be "desc" the config should be on that widget :)

Kedae (approved):

PR #9349, opened 2024-12-13
Fix some react issues

labo-flg:
This PR solves various react warnings and errors, commonly found in the app.


opencti-platform/opencti-front/src/private/components/analyses/reports/ReportDetails.jsx
```diff
@@ -242,7 +242,7 @@ const ReportDetails = ({ report }) => {
               const relatedContainer = relatedContainerEdge.node;
               return (
                 <ListItem
-                  key={reportData.id}
```
delemaf, line 245: NB: This part have been modified on 6.5, we'll have some conflicts on release

delemaf (approved):
Look good to me 😄

Issue #4961, opened 2023-11-23
Management panel for Access Restriction on Knowledge Entities


Jipegien:
## Use case

With the ability to design Access Restriction on specific Knowledge entities (https://github.com/OpenCTI-Platform/opencti/issues/3181) comes the risk that some, with time, will be forgotten, creating a Data debt.

Bu having access to a dedicated management panel, administrators will be able to retrieve any restricted Knowledge Entity and "free" it or change those who can access it.

PR #9402, opened 2024-12-18
[frontend/backend] Enabled/disabled “authorized members” in Containers' form creation + default values  (#4961)

CelineSebe:
### Proposed changes

* Add authorizedMembersField in Report, Grouping, Case incident, RFI, RFT (in the creation form)
* Add security access in the frontend (with security, canEdit and CAPA manage authorized members)
* Implement forbiden access in backend using the method in CreateEntityRaw within middleware.js
* Rework the UseDefaultValue, EntitySettings-domain, EntitySettingsEdition
* Improve accordion style

### Related issues

* #4961 

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/components/Accordion.jsx
+++ b/opencti-platform/opencti-front/src/components/Accordion.jsx
@@ -9,12 +9,15 @@ export const Accordion = styled((props) => (<MuiAccordion disableGutters elevati
   '&:before': {
     display: 'none',
   },
+  borderRadius: '4px',
+  backgroundColor: 'rgba(255, 255, 255, 0)',
 }));
 
 export const AccordionSummary = styled((props) => (
   <MuiAccordionSummary expandIcon={<ArrowForwardIosSharpIcon sx={{ fontSize: '0.9rem' }} />} {...props}/>
 ))(({ theme }) => ({
-  backgroundColor: theme.palette.mode === 'dark' ? 'rgba(255, 255, 255, .05)' : 'rgba(0, 0, 0, .03)',
+  // backgroundColor: theme.palette.mode === 'dark' ? 'rgba(255, 255, 255, .05)' : 'rgba(0, 0, 0, .03)',
+  backgroundColor: 'rgba(255, 255, 255, .05)',
   flexDirection: 'row-reverse',
   '& .MuiAccordionSummary-expandIconWrapper.Mui-expanded': {
     transform: 'rotate(90deg)',
--- a/opencti-platform/opencti-front/src/components/Accordion.jsx
+++ b/opencti-platform/opencti-front/src/components/Accordion.jsx
@@ -16,8 +16,7 @@ export const Accordion = styled((props) => (<MuiAccordion disableGutters elevati
 export const AccordionSummary = styled((props) => (
   <MuiAccordionSummary expandIcon={<ArrowForwardIosSharpIcon sx={{ fontSize: '0.9rem' }} />} {...props}/>
 ))(({ theme }) => ({
-  // backgroundColor: theme.palette.mode === 'dark' ? 'rgba(255, 255, 255, .05)' : 'rgba(0, 0, 0, .03)',
-  backgroundColor: 'rgba(255, 255, 255, .05)',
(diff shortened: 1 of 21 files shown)


SouadHadjiat (commented):
Add access verification on authorized members during creation (validateUserAccessOperation)

opencti-platform/opencti-front/src/private/components/common/form/AuthorizedMembersField.tsx
```diff
@@ -279,7 +279,7 @@ const AuthorizedMembersField = ({
             {canDeactivate && (
               <Field
                 component={SwitchField}
-                containerstyle={{ marginTop: 15 }}
+                containerstyle={{ marginTop: 15, paddingLeft: 10 }}
```
SouadHadjiat, line 282: are we sure about this padding ? 
image

CelineSebe: I don't have the choice because the switch is not completely displayed without this padding

opencti-platform/opencti-front/src/components/Accordion.jsx
```diff
@@ -9,12 +9,15 @@ export const Accordion = styled((props) => (<MuiAccordion disableGutters elevati
...
+  borderRadius: '4px',
+  backgroundColor: 'rgba(255, 255, 255, 0)',
 }));
 
 export const AccordionSummary = styled((props) => (
   <MuiAccordionSummary expandIcon={<ArrowForwardIosSharpIcon sx={{ fontSize: '0.9rem' }} />} {...props}/>
 ))(({ theme }) => ({
-  backgroundColor: theme.palette.mode === 'dark' ? 'rgba(255, 255, 255, .05)' : 'rgba(0, 0, 0, .03)',
+  // backgroundColor: theme.palette.mode === 'dark' ? 'rgba(255, 255, 255, .05)' : 'rgba(0, 0, 0, .03)',
+  backgroundColor: 'rgba(255, 255, 255, .05)',
```
JeremyCloarec: Was it validated with product team? This kinda breaks the current styling in light mode: 
image

CelineSebe: validated with product team, but you're right for the summary background in the light mode. I'll reverse this. It's an oversight

JeremyCloarec:
There is a bug on creation, when the author is different than the current user.
To reproduce:
* Go to the creation form, and select an author different than the current user
image
* Then activate restriction access:
image
We get a "Access already granted" message that is always there, and in the current list of accesses, we have a 'Deleted or restricted member' line

JeremyCloarec:
A user that doesn't have access to other Indivuals can still see the indivuals configured by default:
The indivuals list that the user sees
image
The default values (the user can see the AuthorizedMemberDelete individual):
image

JeremyCloarec:
When a user that was configured in default values is deleted, the user is still used in the authorized_members input: isn't there a way to ignore deleted members when getting the default values?

JeremyCloarec:
We can configure the default values so the Creator has no access to the entity: is it intended? 
image

marieflorescontact:
> We can configure the default values so the Creator has no access to the entity: is it intended?

yes it is intended

SouadHadjiat:
I don't understand why the user can't see other individuals, what access does the user have ? But anyway, we query members API, and it returns users, organizations, and groups (not individuals).

> A user that doesn't have access to other Indivuals can still see the indivuals configured by default: The indivuals list that the user sees image The default values (the user can see the AuthorizedMemberDelete individual): image

JeremyCloarec:
> I don't understand why the user can't see other individuals, what access does the user have ? But anyway, we query members API, and it returns users, organizations, and groups (not individuals).
> 
> > A user that doesn't have access to other Indivuals can still see the indivuals configured by default: The indivuals list that the user sees image The default values (the user can see the AuthorizedMemberDelete individual): image

I think it's more related to the org segregation config. My user isn't in the main platform org, so he doesn't see the individuals because they are not shared with his org. But I think the resolved default values of the authorized members aren't entirely checking user access? In this case I was expecting my user to see a 'Restricted' user 

Here are the capas of the user:
image
My user: 
image
And platform main org:
image

SouadHadjiat (approved):

JeremyCloarec (approved):

PR #9455, opened 2024-12-23
[backend] fix file access restriction to also target trash index (#9424)

### LINKED ISSUE #9424: Cannot delete permanently an entity from trash if a file is associated
## Description

it appears that permanent deletion is broken as long as there is a file attached to an entity

## Environment
OCTI 6.4.5

## Reproducible Steps

1. find a report that comes from a feed, with attached PDF.
2. delete it, it goes into trash
3. permanently delete the report => error `file not found or restricted`

## Additional information

The permanent deletion action should not end when a file was already deleted, maybe log a warning but do not end in error for that.

## Screenshots (optional)
Image

labo-flg:
### Proposed changes

When using `internalLoadById` to do a file access check, target also the trash index and not only the read indices.

### Related issues

* closes #9424

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/python/testing/local_exporter.py
+++ b/opencti-platform/opencti-graphql/src/python/testing/local_exporter.py
@@ -6,10 +6,11 @@ from pycti.api.opencti_api_client import OpenCTIApiClient
 
 
 class TestLocalExporter:
-    def __init__(self, api_url, api_token, entity_id, file_name, file_markings):
+    def __init__(self, api_url, api_token, entity_id, entity_type, file_name, file_markings):
         self.api_url = api_url
         self.api_token = api_token
         self.entity_id = entity_id
+        self.entity_type = entity_type
         self.file_name = file_name
         self.file_markings = file_markings
 
@@ -17,7 +18,7 @@ class TestLocalExporter:
         opencti_api_client = OpenCTIApiClient(self.api_url, self.api_token)
         # Generate a json bundle from openCTI
         bundle = opencti_api_client.stix2.get_stix_bundle_or_object_from_entity_id(
-            entity_type="Malware", entity_id=self.entity_id, mode="full"
+            entity_type=self.entity_type, entity_id=self.entity_id, mode="full"
         )
         json_bundle = json.dumps(bundle, indent=4)
         # Upload the export inside the entity to ack like an import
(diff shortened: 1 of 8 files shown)


opencti-platform/opencti-graphql/src/python/testing/local_exporter.py
```diff
@@ -6,18 +6,19 @@
...
+        self.entity_type = entity_type
         self.file_name = file_name
         self.file_markings = file_markings
 
     def upload(self):
         opencti_api_client = OpenCTIApiClient(self.api_url, self.api_token)
         # Generate a json bundle from openCTI
         bundle = opencti_api_client.stix2.get_stix_bundle_or_object_from_entity_id(
-            entity_type="Malware", entity_id=self.entity_id, mode="full"
+            entity_type=self.entity_type, entity_id=self.entity_id, mode="full"
```
labo-flg, line 21: This change is unrelated to the fix. It's just to make this script more usable in other conditions.
I have updated it during my dev, but I thought it was still something good to keep.

opencti-platform/opencti-graphql/src/python/testing/local_uploader.py
```diff
@@ -0,0 +1,37 @@
...
+        self.api_token = api_token
+        self.entity_id = entity_id
+        self.file_name = file_name
+        self.file_markings = file_markings
+
+    def upload(self):
+        opencti_api_client = OpenCTIApiClient(self.api_url, self.api_token)
+
+        # Upload the given file to the entity
+        opencti_api_client.stix_domain_object.add_file(
```
labo-flg, line 20: new utility for our tests, to upload a file to an entity

JeremyCloarec:
I don't get why the file attached to the report is deleted/sent to the trash when the report is deleted. It should simply be marked as removed with the `await elUpdateRemovedFiles(element, true);` line in `internalDeleteElementById` ? I'll investigate it more to better understand

EDIT: after testing, I understand your change better now. I thought the file was sent to the trash as well, but that's not the case,  your change is to make it possible to load the entity related to the file.
I did encounter something that no longer seems to work though: the files are no longer flagged as 'removed'. It looks like the files no longer have an entity_id, making the `elUpdateRemovedFiles` method obsolete: the query can never match the correct files. Maybe we could also fix this in this PR

SouadHadjiat (approved):

PR #9464, opened 2024-12-26
Fix dashboard header UI (under FAB replacement FF)

labo-flg:
### Proposed changes
Under FAB_REPLACMEENT feature flag, the import/create buttons are moved to the header bar, which is already crowded with buttons.

I suggest to move the time filters under the header, so that the header displays correctly when the screen size is >= 1400px wide.

image

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/workspaces/WorkspaceCreation.jsx
+++ b/opencti-platform/opencti-front/src/private/components/workspaces/WorkspaceCreation.jsx
@@ -123,7 +123,6 @@ const WorkspaceCreation = ({ paginationOptions, type }) => {
 
   const createDashboardButton = FAB_REPLACED ? (props) => (
     <>
-      <CreateEntityControlledDial entityType='Dashboard' {...props} />
       <Button
         color='primary'
         variant='outlined'
@@ -135,6 +134,7 @@ const WorkspaceCreation = ({ paginationOptions, type }) => {
       >
         {t_i18n('Import dashboard')}
       </Button>
+      <CreateEntityControlledDial entityType='Dashboard' {...props} />
     </>
   ) : ({ onOpen }) => (
     <SpeedDial
--- a/opencti-platform/opencti-front/src/private/components/workspaces/WorkspaceHeader.jsx
+++ b/opencti-platform/opencti-front/src/private/components/workspaces/WorkspaceHeader.jsx
@@ -117,6 +117,7 @@ const WorkspaceHeader = ({
   const classes = useStyles();
   const { t_i18n } = useFormatter();
   const { isFeatureEnable } = useHelper();
+  const isFABReplaced = isFeatureEnable('FAB_REPLACEMENT');
   const openTagsCreate = false;
   const [openTag, setOpenTag] = useState(false);
   const [newTag, setNewTag] = useState('');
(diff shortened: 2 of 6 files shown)


opencti-platform/opencti-front/src/private/components/workspaces/WorkspaceCreation.jsx
```diff
@@ -135,6 +134,7 @@ const WorkspaceCreation = ({ paginationOptions, type }) => {
       >
         {t_i18n('Import dashboard')}
       </Button>
+      <CreateEntityControlledDial entityType='Dashboard' {...props} />
```
labo-flg, line 137: we keep the same order everywhere : from left to right = tertiary > secondary > primary

opencti-platform/opencti-front/src/private/components/workspaces/WorkspaceHeader.jsx
```diff
@@ -221,81 +222,10 @@ const WorkspaceHeader = ({
             <WorkspacePopover workspace={workspace} />
           </div>
         </Security>
-        {variant === 'dashboard' && (
+        {variant === 'dashboard' && !isFABReplaced && (
           <Security
             needs={[EXPLORE_EXUPDATE, INVESTIGATION_INUPDATE]}
             hasAccess={canEdit}
-            placeholder={
```
labo-flg, line 228: Security placeholder was the same as the child component, so either the Security is irrelevant or we need to remove the placeholder.

opencti-platform/opencti-front/src/private/components/workspaces/dashboards/Dashboard.jsx
```diff
@@ -799,13 +802,24 @@ const DashboardComponent = ({ workspace, noToolbar }) => {
...
+          <WorkspaceHeader
+            handleAddWidget={handleAddWidget}
+            workspace={workspace}
+            variant="dashboard"
+            config={manifest.config}
+            handleDateChange={handleDateChange}
+          />
+          {isFABReplaced && (
+            <div style={{ marginTop: 8 }}>
+              <DashboardTimeFilters
```
labo-flg, line 815: I refactored the time filters, extracting the code from WorkspaceHeader and using it just here, in the dashboard.

opencti-platform/opencti-front/src/private/components/workspaces/dashboards/WidgetConfig.jsx
```diff
@@ -1452,15 +1452,15 @@ const WidgetConfig = ({ workspace, widget, onComplete, closeMenu }) => {
               <Button
                 variant='outlined'
                 disableElevation
-                style={{ marginTop: '2.5px', marginLeft: '4px' }}
+                sx={{ marginLeft: 1 }}
```
labo-flg: quick styling fix on my way

opencti-platform/opencti-front/src/private/components/workspaces/dashboards/public_dashboards/PublicDashboards.tsx
```diff
@@ -103,6 +104,9 @@ const PublicDashboards = () => {
   const { t_i18n } = useFormatter();
   const { isFeatureEnable } = useHelper();
 
+  const { setTitle } = useConnectedDocumentModifier();
+  setTitle(t_i18n('Public dashboards'));
```
labo-flg, line 108: unrelated fix on the way : tab title was `Dashboard_public`

delemaf (approved):
Tested with and without feature flag, look good to me 🚀

PR #9470, opened 2024-12-27
[frontend] Data tables for playbooks

Archidoit:
### Proposed changes
- Implement data tables for playbooks
- When creating a playbook, we are redirected automatically to the e dition page of the playbook

### Related issues
to align playbook behavior with custom dashboards and future fintel templates (in particular, the 3 dots at the end of a playbook line)


romain-filigran:
_When creating a playbook, we are redirected automatically to the e dition page of the playbook_
Perhaps we should do the same with the following to be consistent ??? @richard-julien ?? 

- dashboard
- investigation

opencti-platform/opencti-front/src/components/dataGrid/dataTableUtils.tsx
```diff
@@ -258,6 +258,13 @@ const defaultColumns: DataTableProps['dataColumns'] = {
       return defaultRender(value, helpers);
     },
   },
+  description: {
+    id: 'description',
+    label: 'Description',
+    percentWidth: 25,
+    isSortable: true,
+    render: ({ description }, helpers) => defaultRender(description, helpers),
+  },
```
frapuks, line 267: This part of code will create a conflict with release/6.5.0. @ValentinBouzinFiligran you may adapt our side in exclusion list ?

opencti-platform/opencti-front/src/private/components/data/Playbooks.tsx
```diff
@@ -15,125 +15,212 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+    messages: {
+      id: 'messages',
+      label: 'Messages',
+      percentWidth: 20,
+      isSortable: false,
+      render: ({ queue_messages }) => n(queue_messages),
+    },
+    playbook_running: {
+      id: 'playbook_running',
+      label: 'Playbook running',
```
frapuks, line 160: In exclusion list, we have choosen the label "active" to be align with Decay rules.

frapuks (commented):
Works well.

opencti-platform/opencti-front/src/private/components/data/Playbooks.tsx
Archidoit: after discussion we'll keep this label

frapuks (approved):

Issue #8146, opened 2024-08-22
Unable to display "sighting" detail when creating a sighting from an organisation entity


romain-filigran:
## Description

When creating a sighting from an organization entity with an observable, it's not possible to view the detail when clicking on it. (seems to redirect to an invalid URL).

Moreover (when it's works), I don't understand while we have the feeling to be in the "knowledge" part and not in the "sighting" part of the entity ("knowledge" tab active and not "sighting") when viewing the detail of a sighting.

![Capture d'écran 2024-08-22 194515](https://github.com/user-attachments/assets/f69bf71e-1074-4ee8-8b50-ed9f9615d506)

## Environment
OpenCTI version: 6.2.14

## Reproducible Steps
Navigate to an organization entity. from the "sighting" menu, create a sighting with an observable (ex: IP address).
Click on the created "sighting", you will redirected to an empty page.

romain-filigran:
We need to look into the possibility of always redirecting to the sighting detail view (/dashboard/events/sightings/{sighting_id}).

CelineSebe:
=> This issue has been resolved with the issue #8814 (fix Sightings tab under Observables and Indicators 
 https://github.com/OpenCTI-Platform/opencti/pull/9486).

However, there remains some doubt about the redirection. It currently points to the sighting embedded within its entity:
/dashboard/observations/indicators/{entity_id}/knowledge/sightings/{sightings_id}.
Perhaps it should instead directly redirect to:
/dashboard/events/sightings/{sighting_id}.

Archidoit:
The fix seems to be : redirect to the Sighting overview (and not to the knowledge overview of one entity involved in the sighting). @romain-filigran

PR #9486, opened 2025-01-03
[frontend] fix Sightings tab under Observables and Indicators (#8814)

### LINKED ISSUE #8814: Sightings tab under Observables and Indicators shows no data
## Description

If I create an indicator and if I create a sighting to an organization, I used to be able to see this sighting from the sighting tab under the indicator. I never done sightings for observables, but I assume it should work the same way.
I can see the sightings under Events -> Sightings, but if I drill down to the indicator, the sighting is not there.

## Environment

1. OS (where OpenCTI server runs): Ubuntu 24.04
2. OpenCTI version: 6.3.5
3. OpenCTI client: frontend

## Reproducible Steps

1. Create indicator
2. Create organization
3. Create sighting from organization
4. Try to pivot to indicator and check sighting tab

## Screenshots
image
image
image

CelineSebe:
### Proposed changes

* isTo become false in the sightings indicator and observables 

### Related issues

* #8814 

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/events/stix_sighting_relationships/EntityStixSightingRelationshipLine.tsx
+++ b/opencti-platform/opencti-front/src/private/components/events/stix_sighting_relationships/EntityStixSightingRelationshipLine.tsx
@@ -256,9 +256,10 @@ export const EntityStixSightingRelationshipLine: FunctionComponent<EntityStixSig
     node,
   );
   const entity = isTo ? data.from : data.to;
+  const isObservable = !!entity?.observable_value;
   const restricted = entity === null;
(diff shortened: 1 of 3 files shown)


marieflorescontact (commented):
when I create a `sighting` from `organization`, then click on the created `sighting` i'm redirected to the indicator knowledge tab and sighting is in sighting tab 👌.  is that correct?
Don't we want to be redirected to `/dashboard/events/sightings/{sighting_id}`. as commented in the issue:
https://github.com/OpenCTI-Platform/opencti/issues/8146
if it's ok to be redirected to the entity's knowlede page then it works fine.

marieflorescontact (approved):

PR #9515, opened 2025-01-07
[backend] Change marking values order: TLP:AMBER+STRICT to be hierarchically greater than TLP:AMBER (#9450)

### LINKED ISSUE #9450: Marking values order: change TLP:AMBER+STRICT to be hierarchically greater than TLP:AMBER
## Description

It seems more logic to differentiate the order between TLP:AMBER and TLP:AMBER+STRICT has the scope is more restricted in TLP:AMBER+STRICT definition.
TLP:AMBER+STRICT should be hierarchically superior to TLP:AMBER

## Environment

2. OpenCTI version: 6.4.5

CelineSebe:
### Proposed changes

* change marking values order during database initialization

### Related issues

* #9450 

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/database/data-initialization.js
+++ b/opencti-platform/opencti-graphql/src/database/data-initialization.js
@@ -158,13 +158,13 @@ const createMarkingDefinitions = async (context) => {
     definition_type: 'TLP',
     definition: 'TLP:AMBER+STRICT',
     x_opencti_color: '#d84315',
-    x_opencti_order: 3,
+    x_opencti_order: 4,
   });
   await addAllowedMarkingDefinition(context, SYSTEM_USER, {
     definition_type: 'TLP',
     definition: 'TLP:RED',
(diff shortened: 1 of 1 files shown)


SarahBocognano:
I tested it, and everything works fine. I have a question though : Don't we need to do a migration here ?

aHenryJard:
> I tested it, and everything works fine. I have a question though : Don't we need to do a migration here ?

No because it's very hard to guess what people have done in their platform since it's hightly customizable.

aHenryJard:
On my side I wonder if we shouldn't set PAP:RED to 5 too, when started a new platform here is what we have:

image

CelineSebe:
> On my side I wonder if we shouldn't set PAP:RED to 5 too, when started a new platform here is what we have:
> 

Is it OK not to have a rank 4 for the PAPs?

aHenryJard (approved):

Issue #9330, opened 2024-12-12
Tools entity version are not fillable


Lhorus6:
## Description

Tools entity version are not fillable

![Image](https://github.com/user-attachments/assets/774db106-16c7-4c7f-8598-3b591efa2aa4)

![Image](https://github.com/user-attachments/assets/a6cfe289-e9a5-4ce9-990b-f84392ddb5ad)

## Environment

6.4.4

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Create a Tool and try to fill in the "Version" field

PR #9525, opened 2025-01-08
[frontend] Tools entity version are not fillable (#9330)

SarahBocognano:
### Proposed changes

* Add tool_version field

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/9330


Gwendoline-FAVRE-FELIX:
It works for adding the field at creation and for modifying it.
But when modifying, the field is not pre-filled with the existing value.
Tool update

Gwendoline-FAVRE-FELIX:
I don't know if versions can be that long, but if they are, the value is not truncated and is on top of the adjacent field.
Tool truncate

opencti-platform/opencti-front/src/private/components/arsenal/tools/ToolDetails.tsx
```diff
@@ -0,0 +1,108 @@
...
+import { ToolDetails_tool$key } from '@components/arsenal/tools/__generated__/ToolDetails_tool.graphql';
+import ExpandableMarkdown from '../../../../components/ExpandableMarkdown';
+import ItemOpenVocab from '../../../../components/ItemOpenVocab';
+import StixCoreObjectKillChainPhasesView from '../../common/stix_core_objects/StixCoreObjectKillChainPhasesView';
+import { truncate } from '../../../../utils/String';
+import FieldOrEmpty from '../../../../components/FieldOrEmpty';
+import { useFormatter } from '../../../../components/i18n';
+import type { Theme } from '../../../../components/Theme';
+
+const useStyles = makeStyles<Theme>((theme) => ({
```
Gwendoline-FAVRE-FELIX: It seems to me that it's depreciated.

Gwendoline-FAVRE-FELIX: Otherwise, it works.

Gwendoline-FAVRE-FELIX (approved):

Issue #9280, opened 2024-12-10
Ordering not working in add nested objects panel


Archidoit:
## Description
The ordering by type is not working in Observable > add nested objects > entities list

## Reproducible Steps
- Go to an observable Knowledge tab
- Add nested objects
![Image](https://github.com/user-attachments/assets/03f054c1-3ce8-4a09-bdd5-0409847f9db0)

- In the entities list, sort by type, the sorting is not working
![Image](https://github.com/user-attachments/assets/4431aa7a-3d19-44c3-aafb-d5806ebb40a3)

![Image](https://github.com/user-attachments/assets/cf6c359a-e0d0-4c56-97f6-3cdfa12c2b0d)

PR #9527, opened 2025-01-08
[frontend] Enable the ordering in Observables / nested objects panel (#9280)

Gwendoline-FAVRE-FELIX:
### Proposed changes

* Changes to orderBy and orderMode to sort by type when clicked.

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/9280


JeremyCloarec:
Right now, if a search term is used, we still can't order by Types.
I think it could be great if the behavior was the same as in the Data/Entities page for exemple: when a search inputted, the sorting is reinitialized, but we can still sort by a selected column afterwards

JeremyCloarec (approved):

Issue #7295, opened 2024-06-09
Be able to select the columns to display in custom dashboard lists widgets


SamuelHassine:
## Use case

Be able to select the columns to display in custom dashboard lists widgets

PR #9532, opened 2025-01-08
[frontend] Columns customisation interface for relationship list widget (#7295)

delemaf:
### Proposed changes

* New feature flag **COLUMNS_CUSTOMIZATION**
* New custom columns selection interface for relationship list widget
image
* Relationship list widget now use DataTables

### Related issues

* Close https://github.com/OpenCTI-Platform/opencti/issues/7295

### Checklist


opencti-platform/opencti-front/src/private/components/widgets/WidgetCreationParameters.tsx
```diff
@@ -88,6 +89,30 @@ const WidgetCreationParameters: FunctionComponent<WidgetCreationParametersProps>
...
+    { attribute: 'relationship_type' },
+    { attribute: 'from.entity_type' },
+    { attribute: 'from.relationship_type' },
+    { attribute: 'to.entity_type' },
+    { attribute: 'to.relationship_type' },
+    { attribute: 'created_at' },
+    { attribute: 'createdBy.name' },
+    { attribute: 'objectMarking' },
+  ];
+  const columns = dataSelection[0]?.columns ?? availableColumns;
```
Archidoit: components of dataSelection are handled more generally with index i 

cf code after this:
image

opencti-platform/opencti-front/lang/front/es.json
```diff
@@ -268,6 +268,7 @@
   "Automation": "Automatización",
   "Autonomous system": "Sistema autónomo (AS)",
   "Autonomous systems": "Sistemas autónomos (AS)",
+  "Available Columns": "Columnas disponibles",
```
Archidoit: why a capital letter before 'columns' ?

opencti-platform/opencti-front/src/components/dashboard/WidgetListRelationships.tsx
```diff
@@ -1,189 +1,79 @@
...
-            </ListItemButton>
-          );
-        })}
-      </List>
+    const iconColumn = {
+      percentWidth: iconWidth,
+      isSortable: false,
+      label: ' ',
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      render: (stixRelationship: any) => (
```
Archidoit, line 43: remove any?

opencti-platform/opencti-front/src/components/dashboard/WidgetListRelationships.tsx
```diff
@@ -1,189 +1,79 @@
...
+        <ItemIcon
+          type={stixRelationship.entity_type}
+          color="primary"
+        />
+      ),
+    };
+    setCurrentColumns({ icon: iconColumn, ...newColumns });
+  }, [columns]);
+
+  if (!currentColumns) return null;
```
Archidoit: there is a specific widget in case there are no data, maybe because having no columns shouldn't be possible

opencti-platform/opencti-front/src/private/components/widgets/WidgetListsDefaultColumns.tsx
```diff
@@ -0,0 +1,31 @@
...
+    { attribute: 'createdBy' },
+    { attribute: 'objectMarking' },
+  ],
+};
+
+export const defaultWidgetColumns: Record<string, WidgetColumn[]> = {
+  relationships: [
+    { attribute: 'entity_type' },
+    { attribute: 'relationship_type' },
+    { attribute: 'from_entity_type', label: 'From entity type' },
```
Archidoit: From entity type is rather called 'Source type'

opencti-platform/opencti-front/src/private/components/widgets/WidgetListsDefaultColumns.tsx
```diff
@@ -0,0 +1,31 @@
...
+    { attribute: 'from_relationship_type', label: 'From relationship type' },
```
Archidoit: what is the difference between 'form entity type' and 'from relationship type' ?

Archidoit:
In a dashboard, it seems I can't move my widget list anymore , and I can't change its size either

image

opencti-platform/opencti-front/src/components/dataGrid/dataTableUtils.tsx
```diff
@@ -1034,6 +1034,36 @@ const defaultColumns: DataTableProps['dataColumns'] = {
...
   },
+  from_entity_type: {
+    id: 'from_entity_type',
+    label: 'From entity type',
+    percentWidth: 10,
+    render: ({ from }) => (<ItemEntityType showIcon entityType={from?.entity_type} inList />),
+  },
+  from_relationship_type: {
+    id: 'from_relationship_type',
+    label: 'From relationship type',
```
Archidoit: Isn't it rather 'Source representative' ?

Archidoit:
Maybe some columns names adjustements to do

Exemple: An attribute is called 'ingested' in the form but 'platform creation date' in the widget list... 

image

Archidoit:
I think we should force the user to select at least one column because it makes no sense to have no columns and because in case no columns are selected the widget is strange (and clicking on the line redirects somewhere): 
image

opencti-platform/opencti-front/src/private/components/widgets/WidgetListsDefaultColumns.tsx
delemaf: `From entity type` is the entity type and `From relationship type` is the name of the entity (you're right it's not very clear, we're going to ask the product for all the wording)
image

opencti-platform/opencti-front/lang/front/es.json
delemaf: A little mistake 😄

opencti-platform/opencti-front/src/components/dashboard/WidgetListRelationships.tsx
delemaf: We've tried to type it but it's a bit a pain because `data` in props is also typed `any`
https://github.com/OpenCTI-Platform/opencti/blob/6d89f79688dd212e79deaa84a1abc925bee75d80/opencti-platform/opencti-front/src/components/dashboard/WidgetListRelationships.tsx#L9

nino-filigran:
Hmm forcing user to have 1 column is annoying if you want to remove all columns and only select the ones you want, even though I understand the concern @Archidoit . Therefore, I'd leave it up to the user to ensure it's not ugly instead of forcing a behavior.

opencti-platform/opencti-front/src/components/dashboard/WidgetListRelationships.tsx
Archidoit: in this case we prefer to use unknown everywhere instead of any

opencti-platform/opencti-front/src/private/components/workspaces/dashboards/WidgetConfigColumnsCustomization.tsx
```diff
@@ -0,0 +1,145 @@
...
+                )}
+              </Droppable>
+            </DragDropContext>
+          </Box>
+        </Box>
+
+        <Box sx={{ display: 'flex', marginTop: 2, justifyContent: 'flex-end' }} >
+          <Button
+            variant='outlined'
+            style={{ marginTop: '2.5px', marginLeft: '4px' }}
```
Archidoit: don't know if we set such precise margin with dots... I think we recommand having only integers

Archidoit:
> In a dashboard, it seems I can't move my widget list anymore , and I can't change its size either
> 
> image

Same problem (can't resize the widget) but only for list relationship widget

Archidoit:
If I create a list relationship widget with the filter: relationship_type NOT EQUALS contains
It seems that the filter is not taken into account

Archidoit:
In a widget relationship widget, if I click on a line of a relationship of type 'contains', the redirection is not correct and I am redirected to a constantly loading page : 
image

delemaf:
> > In a dashboard, it seems I can't move my widget list anymore , and I can't change its size either
> > image
> 
> Same problem (can't resize the widget) but only for list relationship widget

This should have been improved a bit but it's still a bit laggy. It's more related to the Data Table implementation so it should be a blocker.

Archidoit:
The relationship list doesn't take the full size of the widget : 
image

Archidoit (approved):

PR #9541, opened 2025-01-09
[frontend] display virtual types in entity type filter list (#7637)

### LINKED ISSUE #7637: [Filter] Filter on "Observable" meta type
## Use case

We can't filter on "Entity type: Observables" as we used to be able to do. This was useful for:

- Share Observables via Live stream/TAXII/etc.
- Create playbooks that listen to Observables, 
- Any use case that requires filtering for all Observable types

## Current Workaround

It used to be possible to filter on "entity type: Observable", but this is no longer the case. You have to select all observable types (the list is long) to be able to do the same thing now.

## Proposed Solution

Reimplement the ability to filter on the "Observable" meta type.

Archidoit:
### Proposed changes
Add the 'Stix Domain Object', 'Stix Cyber Observable' and 'Stix core relationship' abstract types in the entity type filter when it's needed (playbooks, triggers).

### Related issues
https://github.com/OpenCTI-Platform/opencti/issues/7637

### DIFF
--- a/opencti-platform/opencti-graphql/src/schema/schemaUtils.js
+++ b/opencti-platform/opencti-graphql/src/schema/schemaUtils.js
@@ -188,10 +188,8 @@ export const keepMostRestrictiveTypes = (entityTypes) => {
   let restrictedEntityTypes = [...entityTypes];
   for (let i = 0; i < entityTypes.length; i += 1) {
     const type = entityTypes[i];
-    const parentTypes = getParentTypes(type);
-    restrictedEntityTypes = parentTypes.includes(type)
-      ? restrictedEntityTypes
-      : restrictedEntityTypes.filter((t) => !parentTypes.includes(t));
+    const parentTypes = getParentTypes(type).filter((p) => p !== type);
+    restrictedEntityTypes = restrictedEntityTypes.filter((t) => !parentTypes.includes(t));
   }
   return restrictedEntityTypes;
 };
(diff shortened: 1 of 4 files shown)


JeremyCloarec:
It's a bit of a weird use case, but it looks like the page crashes if we use a EntityType=Entity AND Observable.
Works with an OR:
image
Crashes with an AND:
image
image
Could we try to push a fix for this when this happens?

Archidoit:
> It's a bit of a weird use case, but it looks like the page crashes if we use a EntityType=Entity AND Observable. Works with an OR: image Crashes with an AND: image image Could we try to push a fix for this when this happens?

Indeed ! It's fixed :)

opencti-platform/opencti-graphql/src/schema/schemaUtils.js
```diff
@@ -188,7 +188,7 @@ export const keepMostRestrictiveTypes = (entityTypes) => {
   let restrictedEntityTypes = [...entityTypes];
   for (let i = 0; i < entityTypes.length; i += 1) {
     const type = entityTypes[i];
-    const parentTypes = getParentTypes(type);
+    const parentTypes = getParentTypes(type).filter((p) => p !== type);
```
JeremyCloarec, line 191: It looks like this change broke some other tests

Archidoit: It is the test that is not correct! It's fixed

JeremyCloarec (approved):

PR #9551, opened 2025-01-09
[frontend] Error when accessing the Observables tab in a workbench (#9543)

### LINKED ISSUE #9543: Error when accessing the Observables tab in a workbench
## Description

Error when accessing the Observables tab within a workbench

## Environment

Testing

## Reproducible Steps

1. Access any report having a workbench which contains observables
2. Click on the tab

## Expected Output

List of observables
## Actual Output

` "message": "Fail to execute engine pagination",
            "locations": [
                {
                    "line": 7,
                    "column": 3
                }
            ],
            "path": [
                "stixCyberObservables"
            ],
            "extensions": {
                "code": "DATABASE_ERROR",
                "data": {
                    "http_status": 500,
                    "genre": "TECHNICAL",
                    "cause": {
                        "message": "Engine execution fail, some shards are not available, please check your engine status",
                        "extensions": {
                            "code": "DATABASE_ERROR",`

SarahBocognano:
### Proposed changes

* Remove 'dst_port' attribute in sorting by default_value

Since it's a wrong usage of filter that is used by the workbenck when checking if an observable already exists it's hard to write a test coverage.

Root cause is that generated filter is this one with dst_port that is a number in satabase shemas :
```
{
  "types": [
    "Domain-Name"
  ],
  "search": null,
  "filters": {
    "mode": "and",
    "filters": [
      {
        "key": [
          "name",
          "value",
          "aliases",
          "x_opencti_aliases",
          "value",
          "content",
          "attribute_key",
          "path",
          "hashes.MD5",
          "hashes.SHA-1",
          "hashes.SHA-256",
          "hashes.SHA-512",
          "dst_port"
        ],
        "values": [
          "wwww.toto.fr"
        ]
      }
    ],
    "filterGroups": []
  },
  "count": 1
}
```

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/9543

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/common/files/workbench/WorkbenchFileContent.jsx
+++ b/opencti-platform/opencti-front/src/private/components/common/files/workbench/WorkbenchFileContent.jsx
@@ -307,7 +307,6 @@ const defaultValueKeys = {
     'hashes.SHA-1',
     'hashes.SHA-256',
     'hashes.SHA-512',
-    'dst_port',
   ],
 };
 


marieflorescontact (dismissed):

marieflorescontact (approved):

labo-flg:
We will discuss about adding non-regression tests separately
See https://github.com/OpenCTI-Platform/opencti/pull/9552 for a repro test, but we're snot sure we need to handle it differently (it's a misuse of the API, giving string filters values for number fields, so naturally it leads to failure).
This is not trivial (nor wise?) to do this in e2e.

PR #9558, opened 2025-01-10
[backend] handle inferred organizations in organization sharing

### LINKED ISSUE #9078: [Org segregation] Object not visible despite belonging to the correct organization
## Description

Data sharing with organizations (i.e. organization segregation) does not take into account organizations added to the user by the inference engine.

## Environment

OCTI 6.3.13

## Reproducible Steps

1. Create 3 Organization
    - "New org" organization
    - "Child" organization which is Child org of "New org" (i.e. create a relation "Child" -> part of -> "New org")
    - "Parent" organization which is Parent org of "New org" (i.e. create a relation "New org" -> part of -> "Parent")

Image

2. Activated this inference rule

Image

3. Create user A with read only capa (i.e. access knowledge) + all marking allowed + part of "New org". With the inference rule, you will have "Parent" org which will also be added automatically.

Image

4. Create 3 Report    
    - "Child" report, shared with "Child" org    
    - "New org" report, shared with "New org" org
    - "Parent" report, shared with "Parent" org

Example
Image

5. Log in with user A and see the list of accessible reports

## Expected Output

As I am part of New org (manually added), and Parent (added by inference rule), I hope to see the "New org" and "Parent" reports.

## Actual Output

I only see what is shared with "New org", not what is shared with "Parent".

## Additional information

If I remove "New org" from the user A, "Parent" is also automatically removed (which is normal). If I now manually add "Parent", then "New org" (so both are added manually, and not thanks to the inference rule) I see the two expected reports. The problem therefore seems that the inferred membership is not taken into account in the segregation by organization.

#### Additional idea

Shouldn't we have an automatic mechanism (without the need for an inference rule)? 

Example:

If I am part of "New org", I inherit from "Parent" (or "Child" I don't know). And so, even if I am not part of "Parent" (or "Child"), I still see the information shared with it.

It's an idea, I don't know if it's desired. But in any case this is currently not the case (you can redo the repro case of this issue, but without activating the inference rule, you will see)

marieflorescontact:
### Proposed changes

* replace `const shouldOrgs = user.allowed_organizations` by `const shouldOrgs = user.organizations` as infferred organization are not in `allowed_organizations`

### Related issues

* closes #9078 

### Checklist


SouadHadjiat:
we might need to change other parts where we use allowed_organizations (in access.ts)

JeremyCloarec:
I think you're right @SouadHadjiat, everything using allowed_organizations in access.ts should use organizations now.
Or maybe an other fix would be to change the way organizations and allowed_organizations are built in user.js buildCompleteUser: we could include inferences in allowed_organizations and not organizations. That's actually what was done before this change https://github.com/OpenCTI-Platform/opencti/pull/5721, maybe that was an unintended change?

SouadHadjiat (commented):
I think we want to remove completely allowed_organizations, right ? @marieflorescontact

opencti-platform/opencti-graphql/tests/01-unit/utils/logger-test.ts
```diff
@@ -129,6 +129,12 @@ describe('Logger test suite', () => {
     expect(cleanObject.cause.attributes.cause.message).toBe('embedded error');
   });
 
+  it('Log object with error correctly formatted', () => {
```
SouadHadjiat: did you change this test ?

SouadHadjiat:
you have a commit from Julien that comes from master (due to force push on master done by Julien a few days ago), you might need to drop this commit on your branch, then rebase master.

opencti-platform/opencti-graphql/tests/01-unit/utils/logger-test.ts
marieflorescontact: no

marieflorescontact:
> I think we want to remove completely allowed_organizations, right ? @marieflorescontact

I thought we wanted to use `user.organizations` but keep `user.allowed_organizations` in the first time. Am i right?

PR #9574, opened 2025-01-13
[backend] fix capability check on feeds (#9573)

### LINKED ISSUE #9573: Cannot access CSV and TAXII feed if user has only capability to manage feeds
## Description

If a user has the capability to manage feeds but not the parent capability to access feeds, then the user cannot access feeds.

## Reproducible Steps

1. Have a user with only capability to manage feeds
2. With this user try to access existing CSV or TAXII feeds

## Expected Output

The user can access the feeds.

## Actual Output

The user cannot access the feeds.

lndrtrbn:
### Proposed changes

* Fix capability checks

### Related issues

* #9573 

### Checklist


SouadHadjiat (approved):

SouadHadjiat:
just need to fix tests @lndrtrbn (adding capa to editor user changes the counters, since it adds an internal relationship between role & capa)

PR #9589, opened 2025-01-14
[frontend] reset local storage from user profile (#9572)

### LINKED ISSUE #9572: Be able to reset local storage from User Profile
Several parameters are currently stored in the local storage of the browser:
* filters
* search
* menu bar expanded or not
* ...

It might happen that some combination filters and search, stored in local storage for each and every view, lead to errors.
These errors are then fixed in a new version of the platform, through a new bugfix.
This bugfix might include removing a faulty filter option that should have not been there in the first place, but if the local storage still has this filter it will be restored automatically and the view will fail again, regardless the fix.

That's a lot of "maybe" but it happens regularly.

I suggest we add the possibility to reset user's local storage from their profile view, with a confirmation dialog..

Also, it will ease dev/test processes internally (we do this operation quite often).

labo-flg:
### Proposed changes

* new section in user profile view, with a button to clear the local storage.

### Related issues

* closes #9572 

### Checklist


CelineSebe:
image

A translation error on "local"

CelineSebe (approved):

opencti-platform/opencti-front/src/private/components/profile/ProfileLocalStorage.tsx
```diff
@@ -0,0 +1,105 @@
...
+import DialogContent from '@mui/material/DialogContent';
+import Dialog from '@mui/material/Dialog';
+import Alert from '@mui/material/Alert';
+import AlertTitle from '@mui/material/AlertTitle';
+
+import DialogActions from '@mui/material/DialogActions';
+import { useFormatter } from '../../../components/i18n';
+import Transition from '../../../components/Transition';
+
+const useStyles = makeStyles(() => ({
```
aHenryJard: @labo-flg `makeStyle ` is deprecated and should not be used in new code. Can you change it before merge ?

labo-flg: I've had a thought about it, and removed what I could.
This one is due to injection in Paper root class, I'm not sure how to do it otherwise.

@lndrtrbn do you have a better practice ?

opencti-platform/opencti-front/src/private/components/profile/ProfileLocalStorage.tsx
```diff
@@ -0,0 +1,105 @@
...
+        </DialogTitle>
+        <DialogContent>
+          <Alert
+            icon={false}
+            severity="warning"
+            variant="outlined"
+            sx={{ color: 'text.primary' }}
+          >
+            <AlertTitle style={{ marginBottom: 0, fontWeight: 400 }}>
+              {t_i18n('This will erase all the views settings you have made. All these changes will be lost. Are you sure?')}
```
aHenryJard, line 69: Not a blocker but
```suggestion
              {t_i18n('This will erase all the views settings that you have made. All these changes will be lost. Are you sure?')}
```
I thing it's either "you made" or "that you have made"

opencti-platform/opencti-front/src/private/components/profile/ProfileLocalStorage.tsx
lndrtrbn: I think replace `classes` by `sx` and `classes.paper` by the object containing the rules and you should be fine

PR #9618, opened 2025-01-16
[frontend] rename files on download to avoid UUID as names (#9634)

### LINKED ISSUE #9634: feat: use a friendly file name when downloading files (e.g. support package)
## Use case

When a file is generated by the platform, most of the time we use a generated UUID for the file name.
This is the case for the support packages, or the exclusion lists for instance.

When the user download these files, a direct link is generated to the storage, and thus the filename is preserved : it's an unfriendly UUID.
We have all the required context to improve this and use a meaningful name, such as the support package date-time, or the exclusion list name.

## Current Workaround

No workaround

## Proposed Solution

Proper solution would be to introduce a new API `downloadSupportPackage` for instance that will build the right file specs and push it for download.

labo-flg:
When downloading a file, we use direct links to the storage.
Some files are generated and named as UUID, not very easy to use once downloaded.

In this PR, I add a quick workaround to be able to override the filename at download time, and use it for exclusion lists and support packages.

Note that a proper implementation would be to do this backend side, so that we are able to fetch files with proper naming, according to data / business logic. This is significantly more work than this workaround.

closes #9634 

Issue #9635 addresses a long-term solution, for another PR.


labo-flg:
After discussions within the team, we decide not to merge this PR.

This workaround has a risky limitation due to file size. Fetching the file into  a memory blob might take a while and even OOM the browser. Some support packages might be very bug for instance.

PR #9623, opened 2025-01-16
[frontend] Add a new relationship views in entities having an Attack Pattern Knowledge view (#8835)

### LINKED ISSUE #8835: Add a view in entity/knowledge/attack pattern representing the relation
## Use case

Given we have introduced a list view to manage the attackp atterns **as entities** within the knowledge views of other entities, we have lost the ability to manage the **relations** between an entity and its attack pattern. 
The goal of this task is to add a new view "relationship", that would display the **relation** between the entity & the attack pattern, so that user can **edit the relation** or **delete the relation**

CelineSebe:
### Proposed changes

* add relationship view button 
* add dataTable view for the relationships
* add generic component to display the relationship

### Related issues

* #8835 

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/common/stix_core_relationships/StixCoreRelationships.tsx
+++ b/opencti-platform/opencti-front/src/private/components/common/stix_core_relationships/StixCoreRelationships.tsx
@@ -6,20 +6,20 @@ import { ViewColumnOutlined } from '@mui/icons-material';
 import FiligranIcon from '@components/common/FiligranIcon';
 import { ListViewIcon, SublistViewIcon } from 'filigran-icon';
 import { graphql } from 'react-relay';
-import {
-  StixCoreRelationshipsLinesPaginationQuery,
-  StixCoreRelationshipsLinesPaginationQuery$variables,
-} from '@components/common/stix_core_relationships/__generated__/StixCoreRelationshipsLinesPaginationQuery.graphql';
-import { StixCoreRelationshipsLines_data$data } from '@components/common/stix_core_relationships/__generated__/StixCoreRelationshipsLines_data.graphql';
+import { StixCoreRelationshipsLinesPaginationQuery, StixCoreRelationshipsLinesPaginationQuery$variables } from './__generated__/StixCoreRelationshipsLinesPaginationQuery.graphql';
+import { StixCoreRelationshipsLines_data$data } from './__generated__/StixCoreRelationshipsLines_data.graphql';
+import StixCoreRelationshipCreationFromEntity from './StixCoreRelationshipCreationFromEntity';
 import useAuth from '../../../../utils/hooks/useAuth';
 import { emptyFilterGroup, isFilterGroupNotEmpty, useRemoveIdAndIncorrectKeysFromFilterGroupObject } from '../../../../utils/filters/filtersUtils';
 import { usePaginationLocalStorage } from '../../../../utils/hooks/useLocalStorage';
 import DataTable from '../../../../components/dataGrid/DataTable';
 import useQueryLoading from '../../../../utils/hooks/useQueryLoading';
+import { KNOWLEDGE_KNUPDATE } from '../../../../utils/hooks/useGranted';
 import { UsePreloadedPaginationFragment } from '../../../../utils/hooks/usePreloadedPaginationFragment';
 import ItemEntityType from '../../../../components/ItemEntityType';
 import ItemIcon from '../../../../components/ItemIcon';
 import { itemColor } from '../../../../utils/Colors';
+import Security from '../../../../utils/Security';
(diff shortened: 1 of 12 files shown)


opencti-platform/opencti-front/src/private/components/common/stix_core_relationships/StixCoreRelationships.tsx
```diff
@@ -0,0 +1,381 @@
...
+        dataColumns={dataColumns}
+        resolvePath={(data: StixCoreRelationshipsLines_data$data) => data.stixCoreRelationships?.edges?.map((n) => n.node)}
+        storageKey={storageKey}
+        initialValues={initialValues}
+        toolbarFilters={contextFilters}
+        lineFragment={stixCoreRelationshipsFragment}
+        preloadedPaginationProps={preloadedPaginationProps}
+        exportContext={{ entity_type: 'Attack-Pattern' }}
+        additionalHeaderButtons={[
+          (<ToggleButton key="matrix" value="matrix" aria-label="matrix">
```
SouadHadjiat: would be nice if we could extract this list of buttons to avoid duplication with StixDomainObjectAttackPatternsKillChainMatrixInLine

marieflorescontact (commented):
Locally tested:
- There is a scroll bar when list is empty:
image

- Need to full refresh to see in the list newly created relationship 
- Sorting by `type`/`creators`/`authors` makes the platform crash
image

opencti-platform/opencti-front/src/private/components/common/stix_core_relationships/StixCoreRelationships.tsx
```diff
@@ -0,0 +1,381 @@
...
+
+  // Filters due to screen context
+  const userFilters = useRemoveIdAndIncorrectKeysFromFilterGroupObject(filters, ['stix-core-relationship']);
+  const predefinedFilters = [{ key: 'relationship_type', values: relationshipTypes }];
+  if (direction === 'all') {
+    predefinedFilters.push({ key: 'fromOrToId', values: [entityId] });
+    predefinedFilters.push({ key: 'elementWithTargetTypes', values: targetTypes });
+  } else if (direction === 'toEntity') {
+    predefinedFilters.push({ key: 'toId', values: [entityId] });
+    // if (role) predefinedFilters.push({ key: 'toRole', values: [role] });
```
marieflorescontact: Did you forget comments (line 302, 306)

CelineSebe:
> * There is a scroll bar when list is empty:

I don't reproduce this issue on my side, but it is linked to the datatable. 
image
Are you in full screen? You may need to reset the columns.

I'll look into the other issues. thanks !

SouadHadjiat (changes requested):
Select all relationships targets all entities in the platform, there should be filters 
image

marieflorescontact (commented):
functionally tested ok 👌

opencti-platform/opencti-front/src/utils/hooks/useQueryLoading.ts
```diff
@@ -21,4 +22,22 @@ const useQueryLoading = <T extends OperationType>(
...
+  if (!equals(variables, varRef.current)) {
+    varRef.current = variables;
+  }
+  // refetch when variables change
+  useEffect(() => {
+    loadQuery(variables, { ...opts, fetchPolicy: 'store-and-network' });
+  }, [varRef.current]);
+
+  return [queryRef, loadQuery];
+};
```
marieflorescontact, line 41: Do we really need to create a  new method here, as the  code is exactly the same as in useQueryLoading except for the return ?

opencti-platform/opencti-front/src/utils/hooks/useQueryLoading.ts
```diff
@@ -21,4 +22,22 @@ const useQueryLoading = <T extends OperationType>(
   return queryRef;
 };
 
+export const useQueryLoadingWithLoadQuery = <T extends OperationType>(
+  query: GraphQLTaggedNode | PreloadableConcreteRequest<T>,
```
marieflorescontact, line 26: Do we really need to create a new method here, as the code is exactly the same as in useQueryLoading except for the return ?

CelineSebe: it's not exactly the same. We have returned loadQuery in addition to queryRef.

marieflorescontact: yes, the return is different, loadQuery is returned in `useQueryLoadingWithLoadQuery` but except that the method is exectly the same. I was wondering if we really need to have 2 methods

SouadHadjiat (approved):

PR #9632, opened 2025-01-17
Floating Action Button Replacement in Settings

Bonsai8863:
### Proposed changes

* Replaced the FABs for creation and edition in Settings > Security
* Replaced the FABs for creation and edition in Settings > Customization
* Replaced the FABs for creation and edition in Settings > Taxonomies
* Replaced the FABs for creation and edition in Settings > Alerting

### Checklist


labo-flg:
Thanks for your contribution ! I've done some tests and found some issues:

`FAB_REPLACEMENT = false`
Expected result : no change compared to current version.

Security > Roles, Groups, Users
* Delete action is visible in the update drawer, it should be visible only under FF

`FAB_REPLACEMENT = true`
Expected result : new UI

* Delete button in drawer is present at the bottom of all tabs, when the drawer has tabs => I would suggest showing it only in the first, Overview tab.

opencti-platform/opencti-front/src/components/UpdateEntityControlledDial.tsx
```diff
@@ -0,0 +1,44 @@
...
+      <div style={{ display: 'flex' }}>
+        {buttonValue}
+      </div>
+    </Button>
+  );
+};
+
+const DefaultUpdateControlledDial = (props: DrawerControlledDialProps) => (
+  <UpdateEntityControlledDial {...props} />
+);
```
labo-flg, line 42: why the need for this export ?

just `export default UpdateEntityControlledDial;` , no ?

opencti-platform/opencti-front/src/private/components/settings/groups/GroupEditionContainer.tsx
```diff
@@ -63,6 +69,85 @@ const GroupEditionContainerFragment = graphql`
...
+
+const GroupDeletionDialog: FunctionComponent<GroupDeletionDialogProps> = ({
+  groupId,
+}) => {
+  const { t_i18n } = useFormatter();
+  const navigate = useNavigate();
+  const [deleting, setDeleting] = useState<boolean>(false);
+  const [displayDelete, setDisplayDelete] = useState<boolean>(false);
+  const deleteSuccessMessage = t_i18n('', {
+    id: '... successfully delete',
```
labo-flg:
```suggestion
    id: '... successfully deleted',
```

opencti-platform/opencti-front/src/components/UpdateEntityControlledDial.tsx
Bonsai8863: `DefaultUpdateControlledDial` is a function that returns the controlled dial. This is passed to `Drawer` components which expect such a function, rather than the actual component itself. The `Drawer` uses this function to generate its controlled dial with its specific `onOpen` and `onClose` functions.

If we default export `UpdateEntityControlledDial` and pass that to `Drawer`s as its `controlledDial` prop, we will get a type error because it is expecting a function to create its controlled dial rather than a component.

Bonsai8863: This follows a similar structure of the `CreateEntityControlledDial`. There, we do export default the `CreateEntityControlledDial`, but everywhere we use it we need to create a function to generate the controlled dial with the given entity type name in its label. You can see an example of that in `ExternalReferenceCreation.tsx`:
```
const CreateExternalReferenceControlledDial = (props: DrawerControlledDialProps) => (
  <CreateEntityControlledDial entityType='External-Reference' {...props} />
);
```

Since the Update controlled dial doesn't need the string in the label, I've moved this boilerplate into `DefaultUpdateControlledDial` so we don't need to rewrite this code in each of the Role, User, and Group edition files.

labo-flg: ok understood, I remember our first discussions on CreateEntityControlledDial

opencti-platform/opencti-front/src/private/components/settings/users/SettingsOrganizationUserCreation.jsx
```diff
@@ -126,19 +139,21 @@ const SettingsOrganizationUserCreation = ({
...
+    <Fab
+      onClick={() => setOpenAddUser(true)}
+      color="secondary"
+      aria-label="Add"
+      className={classes.createButtonFab}
+    >
+      <Add />
+    </Fab>
+  );
+  if (isFABReplaced) { createButton = CreateOrgUserControlledDial; }
```
labo-flg: This will not work actually.

When FF is disabled, `createButton` is a `<Fab ..>` component.
When FF is enabled, `createButton` is a function component, you cannot instantiate it in the JSX the same way below.

If you test this feature under FF you won't see any button available.

To test, enable a platform organization in the Policies, and create a user with no admin capabilities, only access to knowledge, but who is admin of the platform organization.
This user will have access to the security > Users panel and see only users of his organization, and he should be able to create users in this organization.

labo-flg:
```suggestion
 if (isFABReplaced) {
    createButton = (
      <CreateOrgUserControlledDial
        onOpen={() => setOpenAddUser(true)}
        onClose={() => setOpenAddUser(false)}
      />
    );
  }
```

labo-flg (approved):
all tested ok, let's merge this :)
Thanks for your contributions !

I'll probably do a follow-up PR to prepare for the FF removal, tweak few things.

Issue #4248, opened 2023-09-07
Select attributes to be displayed in Dashboard widget list


Jipegien:
## Use case

Users want to custom their Dashboard lists with relevant attributes or ref.

Display should be adapted to the number of columns (not croped columns, no large blank)

![image](https://github.com/OpenCTI-Platform/opencti/assets/104078945/fda4c9d1-8ac5-4bd0-88f2-38319df606e4)

sweet-mentat:
I would also like to be able to edit columns displayed in visualizations and the various display tables in OpenCTI

sweet-mentat:
Can you please add this to this epic? https://github.com/OpenCTI-Platform/opencti/issues/3863

romain-filigran:
related to https://github.com/OpenCTI-Platform/opencti/issues/7295

nino-filigran:
Closing this issue given the above PR merged.

PR #9636, opened 2025-01-17
[frontend] Columns customisation for list entities widget (#7295)

delemaf:
### Proposed changes
* Extend columns customisation interface to entities list widget
* When a entity_ type filter is selected, propose corresponding entity attributes

### Related issues

* Close https://github.com/OpenCTI-Platform/opencti/issues/7295
* Close https://github.com/OpenCTI-Platform/opencti/issues/4248

### Checklist


opencti-platform/opencti-front/src/components/dataGrid/dataTableUtils.tsx
```diff
@@ -920,6 +1016,39 @@ const defaultColumns: DataTableProps['dataColumns'] = {
...
+    render: ({ x_opencti_epss_score }) => <ItemScore score={x_opencti_epss_score} />,
+  },
+  x_opencti_epss_percentile: {
+    id: 'x_opencti_epss_percentile',
+    label: 'EPSS Percentile',
+    percentWidth: 15,
+    render: ({ x_opencti_epss_percentile }) => (
+      <Tooltip title={x_opencti_epss_percentile}>
+        <Truncate>{x_opencti_epss_percentile || '-'}</Truncate>
+      </Tooltip>
```
lndrtrbn: Why not using `defaultRender` here ?

opencti-platform/opencti-front/src/private/components/widgets/WidgetListsDefaultColumns.tsx
```diff
@@ -11,12 +11,142 @@ export const defaultWidgetColumns: Record<string, WidgetColumn[]> = {
...
   relationships: [
     ...defaultWidgetColumns.relationships,
     { attribute: 'start_time' },
     { attribute: 'stop_time' },
   ],
+  common: [
+    ...defaultWidgetColumns.common,
+  ],
+  Report: [
+    { attribute: 'report_types', label: 'Report type' },
```
lndrtrbn, line 38: How the translation is made on labels?

opencti-platform/opencti-front/src/private/components/widgets/WidgetListsDefaultColumns.tsx
```diff
@@ -11,12 +11,142 @@ export const defaultWidgetColumns: Record<string, WidgetColumn[]> = {
...
+    { attribute: 'objectAssignee' },
+    { attribute: 'objectParticipant' },
```
lndrtrbn, line 40: Why some does not have labels?

opencti-platform/opencti-front/src/private/components/workspaces/dashboards/WidgetConfigColumnsCustomization.tsx
lndrtrbn, line 1: Why this file only used in WidgetConfigParameters is not in its folder?

lndrtrbn: Also, to help understand this component is in fact a kind of an input used in a form I would name it with `Input` or `Field` at the end. And replace props `columns` by `value` and `setColumns` by `onChange` to keep consistent API between form inputs, even it is a very specific use case

opencti-platform/opencti-front/src/private/components/workspaces/dashboards/WidgetConfigColumnsCustomization.tsx
```diff
@@ -25,6 +25,14 @@ const WidgetConfigColumnsCustomization: FunctionComponent<WidgetConfigColumnsCus
...
   const theme = useTheme<Theme>();
 
+  // Ensure columns only include available columns
+  useEffect(() => {
+    const filteredColumns = columns.filter((col) => availableColumns.some((availableCol) => availableCol.attribute === col.attribute));
+    if (filteredColumns.length !== columns.length) {
+      setColumns(filteredColumns);
+    }
+  }, [availableColumns, columns]);
+
```
lndrtrbn: I don't know why yet, but I can only see incoming bugs with those kind of useEffect that modifies the same data it has in dependencies

opencti-platform/opencti-front/src/private/components/widgets/WidgetListsDefaultColumns.tsx
ValentinBouzinFiligran: Some attributes are something_type and are simply translated as "Types".
and we had to have a different header name

opencti-platform/opencti-front/src/private/components/widgets/WidgetListsDefaultColumns.tsx
ValentinBouzinFiligran: Some attributes are something_type and are simply translated as "Types".
and we had to have a different header name

opencti-platform/opencti-front/src/components/dataGrid/dataTableUtils.tsx
ValentinBouzinFiligran: done

opencti-platform/opencti-front/src/private/components/workspaces/dashboards/WidgetConfigColumnsCustomization.tsx
delemaf: `availableColumns` is not modified in this component so no major risk but this could be also filtered directly in Selected Columns but I less clearer to me

Archidoit:
When creating an entities list widgets with filters: entity type = campaign, it's written 'restricted' as value in the column 'Aliases' : 
image

Archidoit:
The 'creators' and 'modification date' columns are always empty even if there are values : 
image

image

opencti-platform/opencti-front/src/components/dashboard/WidgetListCoreObjects.tsx
```diff
@@ -1,62 +1,53 @@
...
-  };
-  if (sortBy && !['entity_type', 'name', 'value', 'observable_value', 'createdBy', 'objectLabel', 'x_opencti_workflow_id', 'objectMarking'].includes(sortBy)) {
-    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-    // @ts-ignore
-    delete dataColumns.x_opencti_workflow_id;
-    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-    // @ts-ignore
-    dataColumns[sortBy] = { percentWidth: 15, isSortable: false };
-  }
+  const buildColumns = useCallback((): DataTableProps['dataColumns'] => {
```
lndrtrbn: Not sure to understand why using a `useCallback` here. Yes the function will be redefined only when `columns` changes but as the function is called line 42, it will be called at every render of the component anyway.

I think if you want to avoid some computation here, a `useMemo` is more what you want to use.

Archidoit:
The column 'alias' is present for every entity type, even for the ones that have no aliases (exemple: reports). The entity types having an alias can be found with this function in backend : isStixObjectAliased

Archidoit:
Attributes are not well selected according to the entity type filter if the main mode is OR.
Exemple, in this case, the entities are not necessarily Report : 
image

but the attributes proposed are the ones of the reports:
image

which can lead to a table with a 'report type' column for entities that are not report : 
image

opencti-platform/opencti-front/src/private/components/widgets/WidgetCreationParameters.tsx
```diff
@@ -771,15 +772,33 @@ const WidgetCreationParameters = () => {
...
-              columns={[...(columns ?? defaultWidgetColumns[perspective])]}
-              setColumns={(newColumns) => setColumns(index, newColumns)}
-            />
-          ) : null))}
+          && dataSelection.map(({ perspective, columns, filters }, index) => {
+            if (perspective === 'relationships' || perspective === 'entities') {
+              const getEntityTypeFromFilters = (filterGroup?: FilterGroup | null) => {
+                if (!filterGroup) return null;
+
+                const entityTypeFilters = filterGroup?.filters?.filter((filter) => filter.key === 'entity_type');
```
Archidoit: maybe you can reuse getEntityTypeTwoFirstLevelsFilterValues or adapt it

delemaf: Done 😄

opencti-platform/opencti-front/src/components/dataGrid/dataTableUtils.tsx
```diff
@@ -947,7 +1084,7 @@ const defaultColumns: DataTableProps['dataColumns'] = {
           classes={{ root: classes.chipInList }}
           color="primary"
           variant="outlined"
-          label={x_opencti_organization_type ?? 'Unknown'}
+          label={x_opencti_organization_type || 'Unknown'}
```
Archidoit, line 1087: translate for Unknown

opencti-platform/opencti-front/src/private/components/widgets/WidgetCreationParameters.tsx
```diff
@@ -771,15 +772,41 @@ const WidgetCreationParameters = () => {
...
+          && dataSelection.map(({ perspective, columns, filters }, index) => {
+            if (perspective === 'relationships' || perspective === 'entities') {
+              const getEntityTypeFromFilters = (filterGroup?: FilterGroup | null): string | undefined => {
+                if (!filterGroup) return undefined;
+
+                const entityTypeFilters = getEntityTypeTwoFirstLevelsFilterValues(filterGroup);
+                const hasSingleEntityType = entityTypeFilters.length === 1;
+                const otherFiltersLength = filterGroup?.filters?.filter((filter) => filter.key !== 'entity_type')?.length;
+
+                if (filterGroup.mode === 'and' && hasSingleEntityType && otherFiltersLength >= 0) {
```
Archidoit, line 784: useless to check otherFiltersLength here since it is always >= 0

Archidoit:
Your strategy to get the entity type filter is not working.
With exemple like this one : 
image


you get the attributes of channels : 
image

Archidoit (approved):

PR #9653, opened 2025-01-20
[worker] Pool thread control

### LINKED ISSUE #9709: Improve worker thread control using thread pool
We need to make worker predictable in term of api calls concurrency.

By default a worker will create consumer thread for each queue but message execution will integrate a thread pool of only 1 slot.
Pool size can be configured,

This will allow us to scale the worker resources properly.

richard-julien:
In order to make worker predictable in term of api calls concurrency
By default a worker will create consumer thread for each queue but message execution will integrate a thread pool of only 1 slot

### DIFF
--- a/opencti-worker/src/worker.py
+++ b/opencti-worker/src/worker.py
@@ -10,6 +10,7 @@ import os
 import sys
 import threading
 import time
+from concurrent.futures import ThreadPoolExecutor
 from dataclasses import dataclass, field
 from threading import Thread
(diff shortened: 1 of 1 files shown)


richard-julien:
Need also to take a look at https://github.com/OpenCTI-Platform/opencti/commit/f55e541da3800cb10aa7a4f803e6c28f64146098

SouadHadjiat (approved):

PR #9654, opened 2025-01-20
[backend] Introduce child fork to handle locking get/release and renewal

### LINKED ISSUE #9711: Move resource lock handling to a separate process
Extending lock on resources during operation on these resources is important in OpenCTI.

Locking in a dedicated process will improve any cpu burn preventing locking to renew.

richard-julien:
Extending lock is important in OpenCTI.
Locking in a dedicated process will improve any cpu burn preventing locking to renew.

### DIFF
--- a/opencti-platform/opencti-graphql/src/boot.js
+++ b/opencti-platform/opencti-graphql/src/boot.js
@@ -1,3 +1,4 @@
+import { fork } from 'child_process';
 import { environment, getStoppingState, logApp, setStoppingState } from './config/conf';
 import platformInit, { checkFeatureFlags, checkSystemDependencies } from './initialization';
 import cacheManager from './manager/cacheManager';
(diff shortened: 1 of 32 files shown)


opencti-platform/opencti-graphql/src/lock/master-lock.js
```diff
@@ -0,0 +1,83 @@
...
+  if (!lockProcess.forked) {
+    throw UnsupportedError('Lock child fork not initialize');
+  }
+  const operation = crypto.randomUUID(); // Use crypto to fast ramdom generation
+  const controller = new AbortController();
+  const { signal } = controller;
+  return new Promise((resolve, reject) => {
+    // Set up the abort callback
+    lockProcess.callbacks.set(`${operation}-abort`, () => {
+      controller.abort({ name: TYPE_LOCK_ERROR });
```
labo-flg, line 75: AbortController abort method does not accept any parameter, from what I see in the docs.

richard-julien: The code tells the opposite :)

```
    interface AbortController {
        /**
         * Returns the AbortSignal object associated with this object.
         */

        readonly signal: AbortSignal;
        /**
         * Invoking this method will set this object's AbortSignal's aborted flag and signal to any observers that the associated activity is to be aborted.
         */
        abort(reason?: any): void;
    }
```

labo-flg: wow ok. 🤷

opencti-platform/opencti-graphql/src/lock/master-lock.js
```diff
@@ -0,0 +1,93 @@
...
+const childLockResources = async (ids, args = {}) => {
+  if (!lockProcess.forked) {
+    throw UnsupportedError('Lock child fork not initialize');
+  }
+  const operation = crypto.randomUUID(); // Use crypto to fast ramdom generation
+  const controller = new AbortController();
+  const { signal } = controller;
+  return new Promise((resolve, reject) => {
+    // Set up the abort callback
+    lockProcess.callbacks.set(`${operation}-abort`, () => {
```
SouadHadjiat, line 74: how does abort message arrives ? I don't see it in child-lock.manager. 
controller.abort will stop completly the child process, right ? in which case do we do that ?

opencti-platform/opencti-graphql/src/boot.js
```diff
@@ -16,6 +17,13 @@ export const platformStart = async () => {
       logApp.error('[OPENCTI] System dependencies check failed', { cause: dependencyError });
       throw dependencyError; //  Re-throw the error to exit the main try block
     }
+    // Init the lock manager
+    try {
+      initLockFork();
```
SouadHadjiat, line 22: do we have a shutdown method that we could call in platformStop ?

opencti-platform/opencti-graphql/src/lock/master-lock.js
```diff
@@ -0,0 +1,93 @@
+import { fork } from 'child_process';
+import * as crypto from 'crypto';
+import { TYPE_LOCK_ERROR, UnsupportedError } from '../config/errors';
+import conf, { booleanConf, logApp } from '../config/conf';
+import { lockResource } from '../database/redis';
+
+// Global variable for the child process
+const USE_CHILD_LOCK = booleanConf('app:child_locking_process:enabled', true);
```
SouadHadjiat: shouldn't we set it to false by default, and give us time to test it in our internal platforms before deploying it ?

SouadHadjiat:
what happens if the child process is killed ?

opencti-platform/opencti-graphql/src/lock/master-lock.js
richard-julien: Good catch, i miss something. Working on it

richard-julien: Fixed

opencti-platform/opencti-graphql/src/boot.js
richard-julien: Code pushed to improve that.
The principle is the fork is an attached process, so if the parent is killed, the child will be also killed by the os.

opencti-platform/opencti-graphql/src/lock/master-lock.js
richard-julien: For me this option is more to have a fallback is something strange happens but so helpfull for the platform that I prefer to go with this option activated by default.

Issue #8906, opened 2024-11-05
Import/export an Outcome Template


nino-filigran:
## Use case

Ability to import/export an outcome template

PR #9706, opened 2025-01-24
[frontend/backend] Export/import Fintel templates (#8906)

lndrtrbn:
### Proposed changes

* Possibility to export a fintel template in a JSON file
* Possibility to import a fintel template with a JSON file

### Related issues
* #8906 

### Checklist


opencti-platform/opencti-graphql/src/modules/fintelTemplate/fintel-template-resolvers.ts
```diff
@@ -1,10 +1,13 @@
 import type { Resolvers } from '../../generated/graphql';
-import { addFintelTemplate, findById, fintelTemplateDelete, fintelTemplateEditField } from './fintelTemplate-domain';
+import { addFintelTemplate, findById, fintelTemplateDelete, fintelTemplateEditField, fintelTemplateExport, fintelTemplateImport } from './fintelTemplate-domain';
 
 const fintelTemplateResolvers: Resolvers = {
   Query: {
     fintelTemplate: (_, { id }, context) => findById(context, context.user, id),
   },
+  FintelTemplate: {
+    toConfigurationExport: (fintelTemplate, _, context) => fintelTemplateExport(context, context.user, fintelTemplate),
```
labo-flg, line 9: why this name and not `fintelTemplateExport` to be consistent ?

Archidoit: it's to be consistant with workspaces, whose configuration export attribute is toConfigurationExport

opencti-platform/opencti-graphql/src/modules/fintelTemplate/fintelTemplate-domain.ts
```diff
@@ -204,3 +209,91 @@ export const initFintelTemplates = async (context: AuthContext, user: AuthUser)
...
+    }))
+  }));
+  await convertWidgetsIds(context, user, widgets, 'internal');
+  const exportWidgets = fintel_template_widgets.map(({ variable_name }, i) => ({
+    variable_name,
+    widget: widgets[i]
+  }));
+
+  return JSON.stringify({
+    openCTI_version: pjson.version,
```
labo-flg, line 242: it should be `octi_version`, openCTI is non-conventional.
I see its already like this in workspace and it's sad.

lndrtrbn: Well as you said it was to keep consistency with workspaces. Julien wanted us to keep the same `toConfigurationExport`, I don't know if it is also the case here

marieflorescontact (commented):
locally tested ok 👌

opencti-platform/opencti-graphql/src/modules/workspace/workspace-utils.ts
```diff
@@ -0,0 +1,110 @@
...
+      if (isNotEmptyField(selection.dynamicTo)) {
+        const dynamicToIds = extractFiltersIds(selection.dynamicTo as FilterGroup, from);
+        resolvingIds.push(...dynamicToIds);
+      }
+    });
+  });
+  // Resolve then second iteration to replace the ids
+  const resolveOpts = { baseData: true, toMap: true, mapWithAllIds: true };
+  const resolvedMap = await internalFindByIds(context, user, resolvingIds, resolveOpts);
+  const idsMap = resolvedMap as unknown as { [k: string]: BasicStoreObject };
```
marieflorescontact, line 96: as unknown as something : don't we have a type or should we create one?

lndrtrbn: The issue with `internalFindByIds` is that it defines a return type that is not valid when we set the option `toMap: true` because the value returned is then a map and not an array. That's why there is a hard cast here.

Maybe there is a better solution but to be fair all this file is just an extract of code that was before in `workspace-domain` and I didn't look into it very much.

marieflorescontact (approved):

labo-flg (approved):

PR #9720, opened 2025-01-26
[frontend] Fix for user deletion on cache + prevent loop on error on Private Root (#9719)

### LINKED ISSUE #9719: Platform crashes after user deletion
## Description

1. First issue, on testing platform:

```
[
    {
        "message": "Cannot read properties of undefined (reading 'internal_id')",
        "locations": [
            {
                "line": 190,
                "column": 3
            }
        ],
        "path": [
            "me",
            "effective_confidence_level"
        ],
        "extensions": {
        },
    }
]
```

2. Second issue, when something like this happens, it reload forever instead of just crash the frontend with the error message:

https://github.com/user-attachments/assets/6c2f32b2-4e47-4849-ab99-a0c865390fb6

Kedae:
Closes #9719 

1 - User deletion doesn't trigger cache reset, hence we are trying to get a user from the cache that is deleted
2 - The default error boundary component is not usable is RootPrivate fails because it requires Intl

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/Error.jsx
+++ b/opencti-platform/opencti-front/src/private/components/Error.jsx
@@ -9,6 +9,11 @@ import ErrorNotFound from '../../components/ErrorNotFound';
 import { useFormatter } from '../../components/i18n';
 import { commitMutation } from '../../relay/environment';
 
+// Highest level of error catching, do not rely on any tierce (intl, theme, ...) pure fallback
+export const HighLevelError = () => (
+  <Alert severity="error">An unknown error occurred. Please contact your administrator or OpenCTI maintainers</Alert>
+);
+
 // Really simple error display
 export const SimpleError = () => {
   const { t_i18n } = useFormatter();
@@ -78,7 +83,7 @@ class ErrorBoundaryComponent extends React.Component {
       const types = map((e) => e.extensions.code, [...baseErrors, ...retroErrors]);
       // Specific error catching
       if (includes('COMPLEX_SEARCH_ERROR', types)) {
-        return <DedicatedWarning title={'Complex search'} description={'Your search have too much terms to be executed. Please limit the number of words or the complexity'}/>;
+        return <DedicatedWarning title={'Complex search'} description={'Your search have too much terms to be executed. Please limit the number of words or the complexity'} />;
       }
       // Access error must be forwarded
       if (includes('FORBIDDEN_ACCESS', types) || includes('AUTH_REQUIRED', types)) {
@@ -91,6 +96,7 @@ class ErrorBoundaryComponent extends React.Component {
     return this.props.children;
   }
 }
+
 ErrorBoundaryComponent.propTypes = {
   display: PropTypes.object,
   children: PropTypes.node,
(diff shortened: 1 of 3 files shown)


opencti-platform/opencti-front/src/private/components/Error.jsx
```diff
@@ -9,6 +9,11 @@ import ErrorNotFound from '../../components/ErrorNotFound';
 import { useFormatter } from '../../components/i18n';
 import { commitMutation } from '../../relay/environment';
 
+// Highest level of error catching, do not rely on any tierce (intl, theme, ...) pure fallback
+export const HighLevelError = () => (
+  <Alert severity="error">An unknown error occurred. Please contact your administrator or OpenCTI maintainers</Alert>
```
delemaf, line 14: Should we translate this (i18n) ?

opencti-platform/opencti-front/src/private/components/Error.jsx
```diff
@@ -78,7 +83,7 @@ class ErrorBoundaryComponent extends React.Component {
       const types = map((e) => e.extensions.code, [...baseErrors, ...retroErrors]);
       // Specific error catching
       if (includes('COMPLEX_SEARCH_ERROR', types)) {
-        return <DedicatedWarning title={'Complex search'} description={'Your search have too much terms to be executed. Please limit the number of words or the complexity'}/>;
+        return <DedicatedWarning title={'Complex search'} description={'Your search have too much terms to be executed. Please limit the number of words or the complexity'} />;
```
delemaf, line 86: Should we also translate this ?

delemaf (changes requested):
I works for me apart from the translation question

opencti-platform/opencti-front/src/private/components/Error.jsx
Kedae: Nop, that's the whole point of not depending on formatter (hence on settings of the plateform)

opencti-platform/opencti-front/src/private/components/Error.jsx
Kedae: I don't think I can inject formatter in ErrorBoundary

opencti-platform/opencti-graphql/src/manager/cacheManager.ts
```diff
@@ -141,7 +141,7 @@ const platformUsers = (context: AuthContext) => {
   const reloadUsers = async () => {
     const users = await listAllEntities(context, SYSTEM_USER, [ENTITY_TYPE_USER], { connectionFormat: false });
     const allUserIds = users.map((user) => user.internal_id);
-    return Bluebird.map(allUserIds, (userId: string) => resolveUserById(context, userId), { concurrency: ES_MAX_CONCURRENCY });
+    return Bluebird.map(allUserIds, (userId: string) => resolveUserById(context, userId), { concurrency: ES_MAX_CONCURRENCY }).filter((u) => u != null);
```
SouadHadjiat, line 144: why adding this filter ? can it happen ? since we just loaded the users from the database.

Kedae: Yes, there could be a weird race-case scenario where the cache is reset while a user is being deleted -> users contains an ID that exists no longer in elastic.

delemaf (approved):

PR #9730, opened 2025-01-27
[frontend] align the card selector UI in export form

labo-flg:
From

image

To

image

to align icons and colors.

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/common/form/StixCoreObjectFileExportForm.tsx
+++ b/opencti-platform/opencti-front/src/private/components/common/form/StixCoreObjectFileExportForm.tsx
@@ -3,7 +3,7 @@ import { Field, Form, Formik } from 'formik';
 import Dialog from '@mui/material/Dialog';
 import DialogTitle from '@mui/material/DialogTitle';
 import Tooltip from '@mui/material/Tooltip';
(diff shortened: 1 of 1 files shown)


opencti-platform/opencti-front/src/private/components/common/form/StixCoreObjectFileExportForm.tsx
```diff
@@ -76,13 +76,22 @@ export interface StixCoreObjectFileExportFormProps {
...
       return <HtmlOutlined fontSize="large" color="primary"/>;
     case 'text/markdown':
       return <LanguageMarkdownOutline fontSize="large" color="primary"/>;
+    case 'text/plain':
+      return <AbcOutlined fontSize="large" color="primary"/>;
+    case 'text/csv':
+      return <NumbersOutlined fontSize="large" color="primary"/>;
+
     default:
       return <div />;
```
Kedae: Can you find something nice for default fallback ? Just to be consistent in display ?

Kedae (approved):

opencti-platform/opencti-front/src/private/components/common/form/StixCoreObjectFileExportForm.tsx
labo-flg: yes ! using as default the same icon as the export button

PR #9791, opened 2025-01-31
[backend] now also send event_id for events sent to worker coming from stream (#9824)

### LINKED ISSUE #9824: An infinite update loop can happen when two instances listen to each other streams
## Description

When two platforms are interconnected with two streams on each side, a scenario can happen where updates are sent in an infinite loop between the instances.
This can occur when one of the platform makes two updates in a short amount of time.

## Environment

4. Other environment details:

## Reproducible Steps

1. { e.g. Run ... }
2. { e.g. Click ... }
3. { e.g. Error ... }

JeremyCloarec:
### Proposed changes

* out of date events coming from stream could be processed even if they should be ignored
* now send originating eventId in streams if current event resulted from an external event

### Related issues

* #9824

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/types/user.d.ts
+++ b/opencti-platform/opencti-graphql/src/types/user.d.ts
@@ -57,4 +57,5 @@ interface AuthContext {
   tracing: TracingContext
   user: AuthUser | undefined
   draft_context?: string | undefined
+  eventId?: string | undefined
 }
--- a/opencti-platform/opencti-graphql/src/database/middleware.js
+++ b/opencti-platform/opencti-graphql/src/database/middleware.js
@@ -1870,6 +1870,10 @@ const updateAttributeRaw = async (context, user, instance, inputs, opts = {}) =>
   };
 };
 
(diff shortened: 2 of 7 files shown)


opencti-platform/opencti-graphql/src/types/user.d.ts
```diff
@@ -57,4 +57,5 @@ interface AuthContext {
   tracing: TracingContext
   user: AuthUser | undefined
   draft_context?: string | undefined
+  eventId?: string | undefined
```
lndrtrbn, line 60: other fields seem to be snake case, why not this one?

lndrtrbn: and in the other files you changed event_id is snake case

SouadHadjiat: It's a field that existed already in AuthContext, it has been declared here for typing (it was only used in js files)

aHenryJard (approved):

PR #9797, opened 2025-01-31
[frontend] Modified left padding for submenus to align with parent text

Bonsai8863:
### Proposed changes

* Modifies submenus to align with the left side of the parent menu text. Easier to navigate and scan quickly.

### Checklist

- [ ] ~~I wrote test cases for the relevant uses case (coverage and e2e)~~

### Further comments

Several users complained about the ambiguity of the submenus and that it was difficult to quickly read and scan the submenus. Left-aligning the submenus with the text of the parent helps users differentiate what is a top-level menu item, and what is a submenu item.

https://github.com/user-attachments/assets/63b02047-ced8-43c8-9b7e-e86d06f0f144

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/nav/LeftBar.jsx
+++ b/opencti-platform/opencti-front/src/private/components/nav/LeftBar.jsx
@@ -142,12 +142,13 @@ const useStyles = makeStyles((theme) => createStyles({
     fontSize: 14,
   },
   menuSubItem: {
+    paddingLeft: 44,
     height: 25,
     fontWeight: 500,
     fontSize: 12,
   },
   menuSubItemWithIcon: {
-    paddingLeft: 20,
+    paddingLeft: 40,
     height: 25,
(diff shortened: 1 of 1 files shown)


opencti-platform/opencti-front/src/private/components/nav/LeftBar.jsx
```diff
@@ -142,12 +142,13 @@ const useStyles = makeStyles((theme) => createStyles({
     fontSize: 14,
   },
   menuSubItem: {
+    paddingLeft: 44,
     height: 25,
     fontWeight: 500,
     fontSize: 12,
   },
   menuSubItemWithIcon: {
-    paddingLeft: 20,
```
CelineSebe: This change doesn't seem necessary to me. 20px or 30 px is sufficient and allows us to see more of the submenu before it gets truncated.

CelineSebe (approved):

PR #9801, opened 2025-02-03
[frontend] Detect datatable parent height changes to auto adapt

lndrtrbn:
### Proposed changes

* Use an observer to detect when the height of the parent has changed to recompute datatable own height.

### Checklist


delemaf (approved):
Perfect 👏

PR #9804, opened 2025-02-03
[frontend] translate types in CSV feeds list (#9771)

### LINKED ISSUE #9771: Missing CSV feeds entity types translations
## Description
Missing translations in CSV feeds entity types

## Reproducible Steps
- Create a CSV feeds
- In the CSV feeds list, the 'entity types' are not translated 

Image

- Whereas there are translated in the creation and edition forms:

Image

Archidoit:
### Proposed changes
translate types in CSV feeds list

### Related issues
https://github.com/OpenCTI-Platform/opencti/issues/9771

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/data/feeds/FeedLine.jsx
+++ b/opencti-platform/opencti-front/src/private/components/data/feeds/FeedLine.jsx
@@ -61,7 +61,7 @@ const styles = (theme) => ({
 
 class FeedLineLineComponent extends Component {
   render() {
-    const { classes, node, dataColumns, paginationOptions } = this.props;
+    const { classes, node, dataColumns, paginationOptions, t } = this.props;
     const filters = deserializeFilterGroupForFrontend(node.filters);
     return (
       <ListItem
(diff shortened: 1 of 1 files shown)


opencti-platform/opencti-front/src/private/components/data/feeds/FeedLine.jsx
```diff
@@ -88,7 +88,7 @@ class FeedLineLineComponent extends Component {
                 className={classes.bodyItem}
                 style={{ width: dataColumns.feed_types.width }}
               >
-                {node.feed_types.join(', ')}
+                {node.feed_types.map((type) => t(`entity_${type}`)).join(', ')}
```
lndrtrbn, line 91: any reason to use `t` instead of `t_i18n` ? If no, prefer use `t_i18n` as good practice because it is the key used to detect missing translations (even if here, the key is dynamic)

Archidoit: in the platform, we use t in components that are not in pure function

lndrtrbn: Well ok, at least also update types `FeedLineLineComponent.propTypes`

Archidoit: done

lndrtrbn (approved):

PR #9817, opened 2025-02-04
[frontend] migrate/refactor StixCoreObjectKnowledgeBar component

delemaf:
### Proposed changes

* Migrate StixCoreObjectKnowledgeBar (you can test in Arsenal, Entities, Events, Locations, Techinques & Threats)

### Checklist


CelineSebe:
I'm not sure if this problem is linked to this PR, but I didn't encounter it on testing or prerelease, so it might be related. As you see below, there's a large space on the left side of the knowledge bar :
image

The knowledge bar also is narrower in comparison with testing. The problem must be due to that. 
Is this expected?
image

opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectKnowledgeBar.tsx
```diff
@@ -0,0 +1,423 @@
...
+                  {section.title}
+                </ListSubheader>
+              )}
+              {section.items.map(({ path, label, iconType, count }) => (
+                <KnowledgeBarItem
+                  key={label}
+                  to={`${stixCoreObjectLink}/${path}`}
+                  iconType={iconType}
+                  label={label}
+                  count={count}
```
CelineSebe, line 420: Is it possible to update the counts for each item without refreshing the page easily? (with a refetch or a state)

delemaf: Not yet, the cause is probably the buggy relay query in the Root components and the way the subscription is done, even the counter of the table is not refreshed, it's not due to the refactor of this component. But I'll open a [issue](https://github.com/OpenCTI-Platform/opencti/issues/9838) to handle this bug 🙏

CelineSebe (approved):

opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectKnowledgeBar.tsx
```diff
@@ -0,0 +1,423 @@
...
+  title: string;
+  items: {
+    label: string;
+    iconType: string;
+    path: string;
+    count: number;
+  }[];
+}
+
+const KnowledgeBarItem = ({ to, iconType, label, count }: { to: string; iconType: string; label: string; count: number }) => {
```
lndrtrbn: Prefer create an interface to type the props

opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectKnowledgeBar.jsx
```diff
@@ -1,833 +0,0 @@
...
-import { useFormatter } from '../../../../components/i18n';
-import useAuth from '../../../../utils/hooks/useAuth';
-import { useSettingsMessagesBannerHeight } from '../../settings/settings_messages/SettingsMessagesBanner';
-import ItemIcon from '../../../../components/ItemIcon';
-
-// Deprecated - https://mui.com/system/styles/basics/
-// Do not use it for new code.
-const useStyles = makeStyles((theme) => ({
-  drawer: {
-    minHeight: '100vh',
```
lndrtrbn, line 21: You don't need it anymore?

lndrtrbn: It seems no

delemaf: I've add it again to make sure all item are displayed 👍

lndrtrbn (approved):

PR #9874, opened 2025-02-07
[backend] change how we check if need to fetch rss feed (#9836)

### LINKED ISSUE #9836: RSS Feed connectors always say they are buffering even when nothing is going on
## Description

Image
Image

In page Ingestion/Connectors, when I select a connector responsible of a RSS feed, I have always the message saying that there is buffering. Event if nothing is going on with this connector and nothing is being ingested.

## Reproducible Steps

1. Go to Data/Ingestion/Connectors
2. Select a connector responsible for RSS feed
3. The message buffering is displayed even if it should not

## Expected Output

The message is displayed only when there is some ingestion waiting.

## Actual Output

The message is always displayed.

lndrtrbn:
### Proposed changes

* Fix issue when buffering was always true

### Related issues

* #9836 

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/manager/ingestionManager.ts
+++ b/opencti-platform/opencti-graphql/src/manager/ingestionManager.ts
@@ -276,9 +276,14 @@ const rssExecutor = async (context: AuthContext, turndownService: TurndownServic
   const ingestionPromises = [];
   for (let i = 0; i < ingestions.length; i += 1) {
     const ingestion = ingestions[i];
-    // If ingestion have remaining messages in the queue, or if last execution was done before RSS_FEED_MIN_INTERVAL_MINUTES minutes, dont fetch any new data
     const { messages_number, messages_size } = await queueDetails(connectorIdFromIngestId(ingestion.id));
-    if (messages_number === 0 && shouldExecuteIngestion(ingestion, RSS_FEED_MIN_INTERVAL_MINUTES)) {
+    // If last execution was done before RSS_FEED_MIN_INTERVAL_MINUTES minutes, dont fetch any new data
+    if (!shouldExecuteIngestion(ingestion, RSS_FEED_MIN_INTERVAL_MINUTES)) {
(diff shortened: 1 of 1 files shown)


opencti-platform/opencti-graphql/src/manager/ingestionManager.ts
```diff
@@ -276,9 +276,14 @@ const rssExecutor = async (context: AuthContext, turndownService: TurndownServic
...
   for (let i = 0; i < ingestions.length; i += 1) {
     const ingestion = ingestions[i];
-    // If ingestion have remaining messages in the queue, or if last execution was done before RSS_FEED_MIN_INTERVAL_MINUTES minutes, dont fetch any new data
     const { messages_number, messages_size } = await queueDetails(connectorIdFromIngestId(ingestion.id));
-    if (messages_number === 0 && shouldExecuteIngestion(ingestion, RSS_FEED_MIN_INTERVAL_MINUTES)) {
+    // If last execution was done before RSS_FEED_MIN_INTERVAL_MINUTES minutes, dont fetch any new data
+    if (!shouldExecuteIngestion(ingestion, RSS_FEED_MIN_INTERVAL_MINUTES)) {
+      logApp.info(`[OPENCTI-MODULE] INGESTION Rss, skipping ${ingestion.name} - last run is more recent than ${RSS_FEED_MIN_INTERVAL_MINUTES} minutes.`);
+      const ingestionPromise = updateBuiltInConnectorInfo(context, ingestion.user_id, ingestion.id, { messages_size });
+      ingestionPromises.push(ingestionPromise);
```
JeremyCloarec: I think the if/elseif/else should be reordered: in this version, the connector will not be shown as buffering even if there are messages in the queue (because we will enter in the first If before checking for messages number).
I think the if/elseif/else order should be: 
- if(messages_number !==0)
- elseif(!shouldExecuteIngestion)
- else

JeremyCloarec:
Could you also apply the same changes to the csvExecutor? It uses the same shouldExecuteIngestion mechanism as the rssExecutor

opencti-platform/opencti-graphql/src/manager/ingestionManager.ts
lndrtrbn: Indeed!

JeremyCloarec (approved):

Issue #9626, opened 2025-01-16
Add a 21 organization to a user, delete another one automatically


Lhorus6:
## Description

When I add a 21st organization to a user, another organization is deleted.

EDIT: After retesting, I have the impression that it's just a display problem (i.e. it removes the org from the screen, but not completely from the user). To be seen

## Environment

OCTI 6.4.7

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Create a User
2. Add 20 organizations to the User
3. Add one more Organization and see what happens
-> it's added but another is deleted

## Expected Output

I hope to be able to add my organization without another being deleted.

## Additional information

If we cap at 20 Organizations per User:
- I expect to be blocked, not silently deleted another Org.

nino-filigran:
I also reproduce (partially) the issue: I do not get an org deleted, however in the UI I don't see the 21th added as an org. I only see it through the activity. After sharing an entity with this 21st org only, the user could access it, meaning that he's part of this 21th org, confirming the UI display issue.

PR #9920, opened 2025-02-10
[backend/frontend] Add a 21 organization to a user, delete another one automatically (#9626)

SarahBocognano:
### Proposed changes

* Add "first" args in graphQL schema and queries to allowed custom pagination

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/9626

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/settings/users/User.tsx
+++ b/opencti-platform/opencti-front/src/private/components/settings/users/User.tsx
@@ -137,6 +137,7 @@ const UserFragment = graphql`
     groupsOrderMode: { type: "OrderingMode", defaultValue: asc }
     organizationsOrderBy: { type: "OrganizationsOrdering", defaultValue: name }
     organizationsOrderMode: { type: "OrderingMode", defaultValue: asc }
+    organizationsCount: { type: "Int", defaultValue: 500 }
     rolesOrderBy: { type: "RolesOrdering", defaultValue: name }
     rolesOrderMode: { type: "OrderingMode", defaultValue: asc }
   ) {
@@ -200,6 +201,7 @@ const UserFragment = graphql`
       }
     }
     objectOrganization(
+      first: $organizationsCount
       orderBy: $organizationsOrderBy
       orderMode: $organizationsOrderMode
     ) {
--- a/opencti-platform/opencti-front/src/schema/relay.schema.graphql
+++ b/opencti-platform/opencti-front/src/schema/relay.schema.graphql
@@ -1679,16 +1679,16 @@ type User implements BasicObject & InternalObject {
   theme: String
   language: String
   external: Boolean
-  roles(orderBy: RolesOrdering, orderMode: OrderingMode): [Role!]!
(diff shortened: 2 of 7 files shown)


opencti-platform/opencti-front/src/schema/relay.schema.graphql
```diff
@@ -1679,16 +1679,16 @@ type User implements BasicObject & InternalObject {
   theme: String
   language: String
   external: Boolean
-  roles(orderBy: RolesOrdering, orderMode: OrderingMode): [Role!]!
+  roles(first: Int, orderBy: RolesOrdering, orderMode: OrderingMode): [Role!]!
```
SouadHadjiat: why do we have this change ? it's not present in opencti.graphql, but it's here in the generated schema.

opencti-platform/opencti-front/src/schema/relay.schema.graphql
```diff
@@ -1679,16 +1679,16 @@ type User implements BasicObject & InternalObject {
...
   capabilities: [Capability]!
   default_hidden_types: [String!]!
   user_confidence_level: ConfidenceLevel
   effective_confidence_level: EffectiveConfidenceLevel
   no_creators: Boolean
   restrict_delete: Boolean
-  groups(orderBy: GroupsOrdering, orderMode: OrderingMode): GroupConnection
-  objectOrganization(orderBy: OrganizationsOrdering, orderMode: OrderingMode): OrganizationConnection
-  objectAssignedOrganization(orderBy: OrganizationsOrdering, orderMode: OrderingMode): OrganizationConnection
+  groups(first: Int, orderBy: GroupsOrdering, orderMode: OrderingMode): GroupConnection
```
SouadHadjiat: same question here

opencti-platform/opencti-front/src/private/components/settings/users/User.tsx
```diff
@@ -137,6 +137,7 @@ const UserFragment = graphql`
     groupsOrderMode: { type: "OrderingMode", defaultValue: asc }
     organizationsOrderBy: { type: "OrganizationsOrdering", defaultValue: name }
     organizationsOrderMode: { type: "OrderingMode", defaultValue: asc }
+    organizationsCount: { type: "Int", defaultValue: 500 }
```
Gwendoline-FAVRE-FELIX, line 140: Why have we chosen to hard-code 500 as the value for organizationsCount?

opencti-platform/opencti-front/src/schema/relay.schema.graphql
SarahBocognano: My bad, I didn't regenerate the schema. At first I've added this first into roles and groups as well but I wasn't required for this fix, so I deleted it

opencti-platform/opencti-front/src/private/components/settings/users/User.tsx
SarahBocognano: To be comfortable in pagination. I could have put 100 but 500 is way larger and will prevent to have another pagination issue

Gwendoline-FAVRE-FELIX (approved):

PR #9924, opened 2025-02-10
[backend] Improve deletion of element with hundreds of thousands relationships (#9923)

### LINKED ISSUE #9923: Improve deletion of element with hundreds of thousands relationships
## Use case

Improve deletion of element with hundreds of thousands relationships.
In some massive deletions, the Ramda merge all can fail for RangeError: Maximum call stack size exceeded

## What to do

- Change the implementation to prevent mergeAll usage that use _curry internally that destructure elements
- Add setImmediate splitter to improve event loop management

richard-julien:
See #9923


aHenryJard (approved):
Looks good to me, tested locally I don't see any side effect.

Issue #9300, opened 2024-12-11
[Placeholder] Ensure draft readiness for release


nino-filigran:
## If the feature request is approved, would you be willing to submit a PR?

Yes / No (Help can be provided if you need assistance submitting a PR)

PR #9935, opened 2025-02-11
[backend] enable file upload and file import in a draft (#9300)

JeremyCloarec:
### Proposed changes

* enable contextual & global file upload in a draft
* enable file import in a draft
* added a files tab in draft overview to see draft global files
* handle file import properly during draft validation

### Related issues

* #9300 

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/database/middleware.js
+++ b/opencti-platform/opencti-graphql/src/database/middleware.js
@@ -39,7 +39,7 @@ import {
   READ_INDEX_HISTORY,
   READ_INDEX_INFERRED_RELATIONSHIPS,
   READ_RELATIONSHIPS_INDICES,
-  READ_RELATIONSHIPS_INDICES_WITHOUT_INFERRED,
+  READ_RELATIONSHIPS_INDICES_WITHOUT_INFERRED, toBase64,
   UPDATE_OPERATION_ADD,
   UPDATE_OPERATION_REMOVE,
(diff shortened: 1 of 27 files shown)


opencti-platform/opencti-graphql/src/database/middleware.js
```diff
@@ -3226,7 +3251,7 @@ const draftInternalDeleteElement = async (context, user, draftElement) => {
     // Try to get the lock in redis
     lock = await lockResources(participantIds, { draftId: getDraftContext(context, user) });
 
-    await elDeleteElements(context, user, [draftElement]);
+    await elMarkElementsAsDraftDelete(context, user, [draftElement]);
```
SouadHadjiat, line 3254: why these changes in draftInternalDeleteElement ?

JeremyCloarec: This change is mostly a refactor: since we know we are already in a draft, we can use elMarkElementsAsDraftDelete directly. Before, it was elDeleteElements that was redirecting to elMarkElementsAsDraftDelete by also checking that we are in a draft context.
The other change in middleware that was not a refactor but a fix is at line 3274: before, we only checked for the index of the element instead of the current draft context. This could have caused some live entities to have their files marked as remove, or even worse, to have their live files deleted

SouadHadjiat (approved):

PR #9942, opened 2025-02-12
[backend/frontend] Improve readability of works in the connector interface

### LINKED ISSUE #8015: Unable to see "active" job in connector interface when there are more than 50 jobs in progress.
## Description

When a connector has more than 50 jobs "in Progress", we can't have a vision of the "active" job as the page only list the 50 most recent jobs. Related connectors: Crowdstrike.

As a first step, it might be a good idea to add a counter for the number of jobs in progress.
Then see if you can paginate, filter or display more than 50 items.
@richard-julien @Kedae @labo-flg : Any suggestions ?

Maybe we should also take a look at the Crowdstrike connector to understand why it's generating so many jobs.

OpenCTI version: 6.2.13

## Reproducible Steps

https://www.notion.so/filigran/All-Connectors-1709057d28914ed293ea867e9cc2e18b

CelineSebe:
### Proposed changes

* add created_at for the in progress works, to see the oldest
* add counter for the 'In progress works' and 'Completed works'

### Related issues

* #8015

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/data/connectors/ConnectorWorks.tsx
+++ b/opencti-platform/opencti-front/src/private/components/data/connectors/ConnectorWorks.tsx
@@ -80,7 +80,6 @@ interface ConnectorWorksComponentProps {
   options: ConnectorWorksQuery$variables[]
   relay: RelayRefetchProp
   inProgress?: boolean
-  isCompleted?: boolean
 }
 
 const ConnectorWorksComponent: FunctionComponent<ConnectorWorksComponentProps> = ({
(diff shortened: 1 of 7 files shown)


opencti-platform/opencti-front/src/private/components/data/connectors/ConnectorWorks.tsx
```diff
@@ -130,8 +134,19 @@ const ConnectorWorksComponent: FunctionComponent<ConnectorWorksComponentProps> =
...
+      {inProgress && (
+      <Typography variant="h4" gutterBottom={true}>
+        {t_i18n('In progress works')}{` (${works.length})`}
+      </Typography>
+      )}
+      {isCompleted && (
+        <Typography variant="h4" gutterBottom={true}>
+          {t_i18n('Completed works')}{` (${works.length})`}
+        </Typography>
+      )}
```
lndrtrbn: Only the call to `t_i18n` needs to be conditioned. You will avoid some duplicate code that way

lndrtrbn: You could also pass the label through props to not make any condition at all

Gwendoline-FAVRE-FELIX (approved):

lndrtrbn (approved):

PR #9968, opened 2025-02-13
[backend] Side deletions of ref relationships can lead to updated_at upgrade (#9944)

### LINKED ISSUE #9944: Side deletions of ref relationships can lead to updated_at upgrade
## Steps

- 1. Create indicator
- 2. Put indicator in a report -> updated_at of indicator do not move => OK
- 3. Remove indicator of the report -> updated_at of indicator change => KO.

## Expected

Updated_at must remains identical

## Hints

Side remove function doest not check the direction.
See elRemoveRelationConnection function

```
          if (isStixRefRelationship(relationType)) {
            source += 'ctx._source[\'updated_at\'] = params.updated_at;';
          }
```

richard-julien:
See #9944

### DIFF
--- a/opencti-platform/opencti-graphql/src/database/engine.js
+++ b/opencti-platform/opencti-graphql/src/database/engine.js
@@ -3653,11 +3653,12 @@ const elRemoveRelationConnection = async (context, user, elementsImpact) => {
           if (isEmptyField(fromIndex)) { // No need to clean up the connections if the target is already deleted.
             return updates;
           }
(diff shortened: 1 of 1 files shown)


lndrtrbn:
Does "remove the indicator from the report" should update the `updated_at` of the report? From what I see in the code of the PR it should, but in my tests it does not

richard-julien:
> Does "remove the indicator from the report" should update the `updated_at` of the report? From what I see in the code of the PR it should, but in my tests it does not

It should work yes. Thats the target

PR #9969, opened 2025-02-13
[backend] Prevent event loop CPU blocking when processing applyUpdate in container rules (#9936)

### LINKED ISSUE #9936: Prevent event loop CPU blocking when processing applyUpdate in container rules
## Description

When processing extremely large report, the rule engine can consume too much cpu

Image

## Hints

Code responsible

```
      // AddedRefs are ids not includes in previous data
      const addedRefs: Array<StixId> = newRefIds.filter((newId) => !previousRefIds.includes(newId));
      // RemovedRefs are ids not includes in current data
      const removedRefs: Array<StixId> = previousRefIds.filter((newId) => !newRefIds.includes(newId));
```

## Todo

Change the filter function to use a loop aware of event loop

## Extra performance detection

Image

richard-julien:
See #9936

### DIFF
--- a/opencti-platform/opencti-graphql/tests/01-unit/utils/data-processing-test.ts
+++ b/opencti-platform/opencti-graphql/tests/01-unit/utils/data-processing-test.ts
@@ -1,0 +1,15 @@
+import { describe, expect, it } from 'vitest';
+import { asyncFilter } from '../../../src/utils/data-processing';
+
+describe('Data processing tests', () => {
+  it('Filter processing data', async () => {
+    const elements = ['Report', 'City', 'Region', 'Malware'];
+    const predicate = (item: string) => item === 'Report';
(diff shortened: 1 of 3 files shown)


opencti-platform/opencti-graphql/tests/01-unit/utils/data-processing-test.ts
```diff
@@ -0,0 +1,15 @@
+import { describe, expect, it } from 'vitest';
+import { asyncFilter } from '../../../src/utils/data-processing';
+
+describe('Data processing tests', () => {
+  it('Filter processing data', async () => {
```
lndrtrbn, line 5: Is it possible in the test to mock `MAX_EVENT_LOOP_PROCESSING_TIME` to 0 to test the function `asyncFilter` with the Promise being called?

Issue #9970, opened 2025-02-13
Editing Arsenal > Tools is not possible


ParamConstructor:
## Description

BUG - On Master - 6.5.1 (also reflected in release/6.6.0) with the editing for various fields on Arsenal --> Tool entities.

## Environment

1. OS (where OpenCTI server runs): MAC OS
2. OpenCTI version: 6.5.1 (also reflected in release/6.6.0)
3. OpenCTI client: Frontend
4. Other environment details:

## Reproducible Steps

Steps to create the smallest reproducible scenario:

BUG - On Master - 6.5.1 (also reflected in release/6.6.0)

1. Create a Arsenal --> Tool
2. Fill in Kill Chain, Author, Markings, Tool Type. 
3. Create - This will save. 
4. Click Update -> Edit the Kill Chain to be empty - form will updated to be blank (you can see in background on report) - try to add a new Kill Chain - nothing happens.
5. Same behavior for Author, and Markings if you empty them then try to add a new one. 
6. Tool type, try to edit it - it will never change - even to blank.

## Expected Output

Kill Chain, Author, Markings, Tool Type should be editable

## Actual Output

1. Kill Chain, Author, Markings can be edited back to "-" (empty), but on edit never set to a new real value
2. Tool Type is never edited

ParamConstructor:
This Bug - "Blocks" - https://github.com/OpenCTI-Platform/opencti/pull/9201 - ([frontend] added configurable required fields to Arsenal)

nino-filigran:
Been able to reproduce

PR #9975, opened 2025-02-14
[frontend] fix tool edition (#9970)

Archidoit:
### Related issues
https://github.com/OpenCTI-Platform/opencti/issues/9970

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/arsenal/tools/ToolEditionOverview.tsx
+++ b/opencti-platform/opencti-front/src/private/components/arsenal/tools/ToolEditionOverview.tsx
@@ -277,13 +277,14 @@ const ToolEditionOverview: FunctionComponent<ToolEditionOverviewProps> = ({
             onSubmit={handleSubmitField}
             containerStyle={{ marginTop: theme.spacing(2) }}
             editContext={context}
+            variant="edit"
           />
           <KillChainPhasesField
             name="killChainPhases"
             setFieldValue={setFieldValue}
             style={{ marginTop: theme.spacing(2) }}
             helpertext={<SubscriptionFocus context={context} fieldName="killChainPhases" />}
-            onChange={editor.changeField}
+            onChange={editor.changeKillChainPhases}
           />
           {tool.workflowEnabled && (
             <StatusField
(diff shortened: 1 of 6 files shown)


opencti-platform/opencti-front/src/private/components/arsenal/tools/ToolEditionOverview.tsx
```diff
@@ -102,6 +102,7 @@ const toolEditionOverviewFragment = graphql`
     }
     killChainPhases {
       id
+      kill_chain_name
```
labo-flg, line 105: was causing `undefined` in the input once we select something

opencti-platform/opencti-front/src/private/components/arsenal/tools/ToolEditionOverview.tsx
```diff
@@ -174,22 +175,15 @@ const ToolEditionOverview: FunctionComponent<ToolEditionOverviewProps> = ({
...
               id: tool.id,
-              input: [
-                {
-                  key: name,
-                  value: Array.isArray(value)
-                    ? value.map((o) => (o as Option).value)
-                    : value,
-                },
-              ],
+              input: [{ key: name, value }],
```
labo-flg, line 186: input of handleSubmitField is never an array of options, but an array of strings
This was causing a crash because it cannot be casted into Option.

opencti-platform/opencti-front/src/private/components/arsenal/tools/ToolEditionOverview.tsx
```diff
@@ -224,19 +218,16 @@ const ToolEditionOverview: FunctionComponent<ToolEditionOverviewProps> = ({
...
     description: tool.description ?? '',
     confidence: tool.confidence,
     tool_types: tool.tool_types ?? [],
     tool_version: tool.tool_version ?? '',
-    createdBy: createdBy as Option,
+    createdBy: convertCreatedBy(tool) as Option,
     killChainPhases: convertKillChainPhases(tool),
     objectMarking: convertMarkings(tool),
-    x_opencti_workflow_id: status as Option,
+    x_opencti_workflow_id: convertStatus(t_i18n, tool) as Option,
```
labo-flg, line 230: this was causing an empty field even when a status is set

labo-flg (approved):

Issue #9723, opened 2025-01-27
Switching between card view and list view introduce size change in the toggle buttons group of a few pixels


SamuelHassine:
## Description

This should not show any difference when clicking:

https://github.com/user-attachments/assets/2dae0fa8-07da-4e40-94d8-861c2c2f7c05

On list views, the toggle group is a little bigger.

PR #10000, opened 2025-02-18
[frontend] Fix button size in cards views (#9723)

Gwendoline-FAVRE-FELIX:
### Proposed changes

* Fix button size in cards views
* Fix Line Button's color in malware view

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/9723

### DIFF
--- a/opencti-platform/opencti-front/src/components/list_cards/ListCards.jsx
+++ b/opencti-platform/opencti-front/src/components/list_cards/ListCards.jsx
@@ -24,10 +24,14 @@ import { export_max_size } from '../../utils/utils';
 const styles = (theme) => ({
   parameters: {
     display: 'flex',
-    alignItems: 'center',
     gap: theme.spacing(1),
     marginBottom: theme.spacing(2),
     flexWrap: 'wrap',
+  },
+  sortFieldContainer: {
+    display: 'flex',
+    gap: theme.spacing(1),
+    alignItems: 'center',
   },
   cardsContainer: {
     margin: '10px 0 0 -15px',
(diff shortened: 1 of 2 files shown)


Archidoit:
The buttons are changing place when we switch between card view and list view in Arsenal > Malware
image

The behavior is ok in Threat Actor Groups
image

Gwendoline-FAVRE-FELIX:
> The buttons are changing place when we switch between card view and list view in Arsenal > Malware image
> 
> The behavior is ok in Threat Actor Groups image

This is another existing bug caused by the scroll bar moving across lines.

opencti-platform/opencti-front/src/components/list_cards/ListCards.jsx
```diff
@@ -101,43 +105,45 @@ class ListCards extends Component {
...
                 />
                 )}
-                <InputLabel classes={{ root: classes.sortFieldLabel }}>
-                  {t('Sort by')}
-                </InputLabel>
-                <FormControl>
-                  <Select
-                    name="sort-by"
-                    value={sortBy}
+                <div className={classes.sortFieldContainer}>
```
labo-flg, line 108: I do not understand how this fixes the button size ?
When I test, the button height is still different between the two modes (38px and a bit less).

Gwendoline-FAVRE-FELIX: Removing the alignItems: 'center' makes it possible to readjust the buttons a little. Before they were 36px high, now they're 37.6979px. Not identical, but close enough to the 38px of the line view.

aHenryJard (approved):

PR #10009, opened 2025-02-19
[backend/frontend] Move basic auth and bearer authentication out of platform sessions (#8877)

### LINKED ISSUE #8877: Move basic auth and bearer authentication out of platform sessions
## Use case

Today the platform act the same for a "real" user login or a token/bearer authentication. In both use case a platform session is created. However maintaining sessions for machine to machine queries through bearer create problematic situation when the client is not able to manage the session.

## Todo

Change the login to have 2 modes.
- By UI -> Create a user session and maintains the current platform creation.
- By Basic Auth / Bearer -> No more session, only use the cache to match the user
- Remove the special option in the user to force the no session mode.
- Remove the deprecated metrics of sessions user count
- Session is now ultra light, only containing OTP information. (important testing point)

Introduce cache change.
- Change completeUser function to be mutliple and predictive in term of performance
- Introduce cache partial update

richard-julien:
See #8877

- [x] Add migation to cleanup stateless_session
- [x] Rebase


opencti-platform/opencti-graphql/tests/02-integration/01-database/stixCyberObservable-domain-test.ts
```diff
@@ -9,7 +9,7 @@ import { requestFileFromStorageAsAdmin } from '../../utils/testQueryHelper';
 import { paginatedForPathWithEnrichment, findById as documentFindById } from '../../../src/modules/internal/document/document-domain';
 
 describe('Testing Artifact merge with files on S3', () => {
-  const adminContext: AuthContext = { user: ADMIN_USER, tracing: undefined, source: 'stixCyberObservableDomain-test', otp_mandatory: false };
+  const adminContext: AuthContext = { user: ADMIN_USER, tracing: undefined, source: 'stixCyberObservableDomain-test', otp_mandatory: false, user_inside_platform_organization: false };
```
SouadHadjiat, line 12: why do we want user_inside_platform_organization inside the context instead of user ?

richard-julien: The idea is to have attributes recomputed at each call in the context, and more stable attributes in the user.
Instead of invalidating the cache for all users with an orga platform is setup, user_inside_platform_organization in context will allow to have a dynamic approach

opencti-platform/opencti-graphql/src/modules/attributes/internalObject-registrationAttributes.ts
```diff
@@ -291,7 +291,6 @@ const internalObjectsAttributes: { [k: string]: Array<AttributeDefinition> } = {
     { name: 'theme', label: 'Theme', type: 'string', format: 'short', mandatoryType: 'no', editDefault: false, multiple: false, upsert: false, isFilterable: true },
     { name: 'language', label: 'Language', type: 'string', format: 'short', mandatoryType: 'no', editDefault: false, multiple: false, upsert: false, isFilterable: true },
     { name: 'external', label: 'External', type: 'boolean', mandatoryType: 'no', editDefault: false, multiple: false, upsert: false, isFilterable: true },
-    { name: 'stateless_session', label: 'Stateless session', type: 'boolean', mandatoryType: 'no', editDefault: false, multiple: false, upsert: false, isFilterable: true },
```
SouadHadjiat, line 294: shouldn't we remove this field in a migration ?

opencti-platform/opencti-graphql/src/migrations/1740426699455-remove-stateless-session.js
```diff
@@ -0,0 +1,33 @@
...
+      bool: {
+        must: [
+          { term: { 'entity_type.keyword': { value: 'User' } } },
+        ],
+      },
+    },
+  };
+  await elUpdateByQueryForMigration(
+    message,
+    READ_DATA_INDICES,
```
SouadHadjiat: we could target only READ_INDEX_INTERNAL_OBJECTS

SouadHadjiat (approved):

PR #10010, opened 2025-02-19
[frontend] Only accessible markings visible in filters (#10006)

### LINKED ISSUE #10006: Marking not accessible which are visible in the filters
## Description

When I filter with the "Marking" filter, I can see all the available markings, even the ones I can't access

## Environment

OCTI 6.5.2

## Reproducible Steps

1. Create a User A with all marking allowed except TLP:RED
2. Create a TLP:RED Report
3. Log in with User A and go to "Analyses > Reports"
4. Add the "Markings" filter and look for TLP:RED 
-> you will find it

## Expected Output

I hope to see in the list only the markings that are accessible to me

## Actual Output

Currently I see in the filter list all the available markings, i.e. all the markings applied to the reports (whether I have access to them or not).

This does not ultimately allow me to see the entities in question, but I still have the possibility of seeing the all markings existing on the platform.

#### The values (markings) available to me in the list

Image

#### The allowed markings to me

Image

Archidoit:
### Proposed changes
In 'markings' filter, the user should only be able to see the markings he has access to

### Related issues
https://github.com/OpenCTI-Platform/opencti/issues/10006


marieflorescontact (approved):
locally tested ok 👌

Issue #9724, opened 2025-01-27
Multiple spacing issues in the "Victimology" screen


SamuelHassine:
## Description

Multiple spacing issues in the "Victimology" screen.

![Image](https://github.com/user-attachments/assets/629e7abf-f6d5-4317-9f13-15d357ccfbd1)

* Red spacing is not the standard one between papers, please refer to "Overview"
* Blue / Yellow lines should be equal, the correct one is the blue one (shortest)

Also, when switching between relationships view and nested view, search field and toggle buttons group move (should stay static):

https://github.com/user-attachments/assets/abbfa8ee-46cd-456b-a1e9-3a8182790645

PR #10012, opened 2025-02-19
[frontend] Fix spacing issues and buttons movement in Victimology view (#9724)

Gwendoline-FAVRE-FELIX:
### Proposed changes

* Fix margin issues

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/9724

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/common/stix_domain_objects/StixDomainObjectVictimology.jsx
+++ b/opencti-platform/opencti-front/src/private/components/common/stix_domain_objects/StixDomainObjectVictimology.jsx
@@ -13,7 +13,7 @@ import StixDomainObjectVictimologyMap from './StixDomainObjectVictimologyMap';
 
 const styles = (theme) => ({
   container: {
-    marginTop: 25,
+    marginTop: theme.spacing(3),
   },
   bottomNav: {
     zIndex: 1,
--- a/opencti-platform/opencti-front/src/private/components/common/stix_domain_objects/StixDomainObjectVictimology.jsx
+++ b/opencti-platform/opencti-front/src/private/components/common/stix_domain_objects/StixDomainObjectVictimology.jsx
@@ -13,7 +13,7 @@ import StixDomainObjectVictimologyMap from './StixDomainObjectVictimologyMap';
 
 const styles = (theme) => ({
   container: {
-    marginTop: 40,
+    marginTop: 25,
   },
   bottomNav: {
     zIndex: 1,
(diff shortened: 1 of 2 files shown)


opencti-platform/opencti-front/src/private/components/common/stix_domain_objects/StixDomainObjectVictimologySectors.jsx
```diff
@@ -302,15 +302,15 @@ class StixDomainObjectVictimologySectorsComponent extends Component {
...
+          <div style={{ float: 'left', marginRight: 20 }}>
             <SearchInput
               variant="small"
               onSubmit={this.handleSearch.bind(this)}
               keyword={searchTerm}
             />
           </div>
           <div className={classes.views}>
-            <div style={{ float: 'right', marginTop: -20 }}>
+            <div style={{ float: 'right', marginTop: -10 }}>
```
marieflorescontact, line 313: Not sure if this is possible but could we avoid negative margin?

marieflorescontact (commented):
looks good but not sure about negative margin

opencti-platform/opencti-front/src/private/components/common/stix_domain_objects/StixDomainObjectVictimologySectors.jsx
Gwendoline-FAVRE-FELIX: I can't do it any other way.

opencti-platform/opencti-front/src/private/components/common/stix_domain_objects/StixDomainObjectVictimology.jsx
```diff
@@ -13,7 +13,7 @@ import StixDomainObjectVictimologyMap from './StixDomainObjectVictimologyMap';
 
 const styles = (theme) => ({
   container: {
-    marginTop: 40,
+    marginTop: 25,
```
aHenryJard: @lndrtrbn told me to use theme.spacing() instead of hard coded value in one of my PR, can it be used here too ?

lndrtrbn: Sometimes it's difficult to use, not technically but because everywhere on the page margins are arbitrary and not multiples of 8. So using `theme.spacing` partially could lead to inconsistent spacings (because in theme it's only multiples of 8). 

But here, for 1px diff, you could have a try

CelineSebe (approved):

Issue #9298, opened 2024-12-11
Graph Rework step 1: Refactor, Improve loading & add enriching capability


nino-filigran:
## Use case

Graphs require a heavy rework.
The goal of the first step is to imrpove the way we load graph to be able to support bigger graphs & to be able to enrich directly form investigations.

ckane:
I just want to say, thank you so much for this change - I just updated to 6.6.1 and was extremely happy to see the cross-container correlation Graphs appear to be working out-of-the-box really well now. Really great addition for these two controls in the graph now:

![Image](https://github.com/user-attachments/assets/300187de-de07-4486-be58-ee175b264859)

labo-flg:
Thanks @ckane 🙏 

We have made a huge refactoring that will allow us to finally tackle some bugs and performance issues in the graphs.
Additionally to the correlation graph switch you mention, we added proper pagination and a loading bar for very large graphs, so now the platform won't try to load everything at once in memory and kill itself.

Plus, the UI has been improved a bit so we better see what's selected in the graphs.

We'll tackle more bugs and improvements in the future, now the code is a bit more clean (and more importantly, not duplicated like crazy anymore)

PR #10017, opened 2025-02-19
[frontend] Refactoring of graphs - chunk 1 : Report knowledge graph

lndrtrbn:
### Proposed changes

* refactoring of graphs - step 1
  * new components architecture
  * use them for Report Knowledge graph

The refactoring is not finished but the PR is to be merged to be able to test on a deployed platform.

### Related issues

* #9298

### Checklist


opencti-platform/opencti-front/src/utils/graph/Graph.tsx
```diff
@@ -0,0 +1,180 @@
...
+      modeTree,
+      selectFree,
+      selectFreeRectangle,
+      withForces,
+    },
+  } = useGraphContext();
+
+  useGraphFilter();
+
+  const shouldDisplayLinks = graphData?.links.length ?? 0 < 200;
```
aHenryJard, line 72: Does this magic number "200" requires a constant, where does it coming from ?

opencti-platform/opencti-front/src/utils/graph/utils/useGraphPainter.ts
```diff
@@ -0,0 +1,224 @@
...
+  };
+
+  /**
+   * Draws a link for 3D mode.
+   *
+   * @param link Link to draw.
+   */
+  const linkThreePaint = (link: GraphLink) => {
+    const sprite = new SpriteText(link.label);
+    sprite.color = 'lightgrey';
```
aHenryJard, line 193: don't we need to use theme colors ?

lndrtrbn: It's copy paste from old code. For now I leave it that way to keep same state as before. Maybe it will change in later chunks

opencti-platform/opencti-front/src/utils/graph/Graph.tsx
lndrtrbn: It's copy paste from old code. I could make a constant, I'm not sure it will stay at all in next chunks anyway.

Gwendoline-FAVRE-FELIX:
Problem: with forces activated, they deactivate for all nodes after a move. Only reactivates after deactivation if no movement has taken place in the meantime.

Gwendoline-FAVRE-FELIX (approved):

PR #10050, opened 2025-02-24
[frontend] Refacto graphs - chunk 2 : all knowledge graphs (#9298)

lndrtrbn:
### Proposed changes

* Refactor all knowledge graphs: grouping, incident, rfi, rft
* Some fixes discovered after chunk 1 merged

### Related issues

* #9298 

### Checklist


marieflorescontact (commented):
Tested ok for basic operations (add entity, create relation, select entity/relation, filtering etc)

We can select an entity that has been filtered, is it ok?
here a filter created by is applied but the filtered entity can still be selected:
image

lndrtrbn:
> Tested ok for basic operations (add entity, create relation, select entity/relation, filtering etc)
> 
> We can select an entity that is been filtered, is it ok? here a filter created by is applied but the filtered entity can still be selected

Same on testing so yes

Gwendoline-FAVRE-FELIX (approved):
Everything works

opencti-platform/opencti-front/src/private/components/cases/case_incidents/IncidentKnowledge.jsx
CelineSebe, line 1: is it possible to convert this component into a functional component ?

CelineSebe:
Functionally, everything works 👌

PR #10103, opened 2025-02-27
[frontend/backend] add new filter capabilities to playbooks

### LINKED ISSUE #9144: [Playbook]: Support for more filtering capabilities to trigger a playbook
## Use case

At the moment, playbooks doesn't purpose many filtering capabilities in the trigger component or in the other playbook components.
It would be great to be able to filter on many cases types to handle more use cases like: 
- note type
- incident response case type
- incident type
- request for information type
- request for Takedown type
....

## Proposed Solution

Do the same as for report types

labo-flg:
Closes #9144 

Added STIX filtering for fields `response_types`, `information_types`, `takedown_types`, `note_types`, `incident_type`

### DIFF
--- a/opencti-platform/opencti-front/src/utils/filters/filtersUtils.tsx
+++ b/opencti-platform/opencti-front/src/utils/filters/filtersUtils.tsx
@@ -96,6 +96,11 @@ export const stixFilters = [
   'x_opencti_cvss_base_score',
   'x_opencti_cvss_base_severity',
   'report_types',
+  'response_types',
+  'information_types',
+  'takedown_types',
+  'note_types',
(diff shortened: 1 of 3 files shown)


Archidoit:
The added keys should also be added in documentation here : https://docs.opencti.io/latest/reference/filters/?h=filters#limited-support-in-stream-events-filtering

Archidoit (approved):

labo-flg:
> The added keys should also be added in documentation here : https://docs.opencti.io/latest/reference/filters/?h=filters#limited-support-in-stream-events-filtering

https://github.com/OpenCTI-Platform/docs/pull/257

Issue #7458, opened 2024-06-24
[Dashboard] White color in "Tree" visualization makes text illegible


Lhorus6:
## Description

When using "Tree" visualization, if a square is white, we can't see what it written (it's written in white on white).

![image](https://github.com/OpenCTI-Platform/opencti/assets/70956706/9ae697bf-c401-4911-add5-10beb5f8a8cb)

## Environment

OCTI 6.1.12

## Reproducible Steps

As we don't choose the color, I don't really have a repro case to give. But the screenshot in the description illustrates the problem perfectly. We just need to prevent the color white.

Steps to create the smallest reproducible scenario:
1. Create a widget "Tree" visualization
2. Hope for a white square

nino-filigran:
Solution: remove white from possible options if possible.

SarahBocognano:
I haven't figured out to reproduce. Everything is in "blue" in my dashboard treemap widget

Lhorus6:
@SarahBocognano 
1. To obtain color, you have to switch on the "Distributed" toggle in the 4th step ("parameter") of the widget configuration.
2. You can't choose the color (I guess it's random), so reproducing it is complicated. Either you're lucky and get it straight away, or you spend your days making lots of them until you end up with a widget that has a white square.

SarahBocognano:
@Lhorus6 Thanks a lot ! A think I've found a way to remove the white one as @nino-filigran requested, but I wanted to verify the possibility first haha

Lhorus6:
Ah I see, indeed, it will be difficult to check :/
But we have a customer with a widget like that (the one in the screenshot), so after the update I'll be able to tell whether it worked or not (not really practical, but at least we'll know for sure if it worked ;) ).

PR #10122, opened 2025-03-03
[frontend] [Dashboard] White color in "Tree" visualization makes text illegible (#7458)

SarahBocognano:
As it is almost impossible to test (I've never been able to have a white square in many attempts), we just need to trust the code in this. I need your advice frontend geniuses tag in reviewers

### Proposed changes

* Filter the possibility to have white color

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/7458

### DIFF
--- a/opencti-platform/opencti-front/src/utils/Charts.jsx
+++ b/opencti-platform/opencti-front/src/utils/Charts.jsx
@@ -24,7 +24,7 @@ export const colors = (temp) => [
   C.grey[temp],
 ];
 
-const isColorTooLight = (hex) => {
+const isColorCloseToWhite = (hex, threshold = 0.9) => {
   if (!hex) return true;
   const c = hex.replace('#', '');
   const r = parseInt(c.substr(0, 2), 16);
--- a/opencti-platform/opencti-front/src/utils/Charts.jsx
+++ b/opencti-platform/opencti-front/src/utils/Charts.jsx
@@ -910,7 +910,7 @@ export const treeMapOptions = (
     },
     colors: [
       theme.palette.primary.main,
-      ...colors(theme.palette.mode === 'dark' ? 400 : 600).filter((n) => n !== '#ffffff'),
+      ...colors(theme.palette.mode === 'dark' ? 400 : 600),
     ],
     fill: {
       opacity: 1,
@@ -945,7 +945,7 @@ export const treeMapOptions = (
       style: {
         fontFamily: '"IBM Plex Sans", sans-serif',
         fontWeight: 600,
-        colors: [theme.palette.text.primary],
+        colors: [theme.palette.text.primary, theme.palette.text.secondary],
       },
       background: {
         enabled: false,
--- a/opencti-platform/opencti-front/src/utils/Charts.jsx
+++ b/opencti-platform/opencti-front/src/utils/Charts.jsx
@@ -1,6 +1,7 @@
 import * as C from '@mui/material/colors';
 import { resolveLink } from './Entity';
 import { truncate } from './String';
+import { isColorCloseToWhite } from './Colors';
 
 export const colors = (temp) => [
   C.red[temp],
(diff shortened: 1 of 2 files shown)


opencti-platform/opencti-front/src/utils/Charts.js
```diff
@@ -910,7 +910,7 @@ export const treeMapOptions = (
     },
     colors: [
       theme.palette.primary.main,
-      ...colors(theme.palette.mode === 'dark' ? 400 : 600),
+      ...colors(theme.palette.mode === 'dark' ? 400 : 600).filter((n) => n !== '#ffffff'),
```
lndrtrbn: I can't see how this can work. We are calling `colors` function with either value 400 or 600, which are never a full white. Do you have manage to reproduce?

lndrtrbn: For me the easiest solution would be to set background to true in `dataLabels` few lines below

lndrtrbn: Otherwise try to have a look if we can pass a function that allows to know at runtime which is the color of the background to apply a contrasted color.

By the time I am writing this comment, I see that `dataLabels.style.colors` is an array, and we are giving only one color, which is probably the white one, maybe just add a black one here and ApexChart automatically manages which one to use depending of the background

SarahBocognano: You're the best thank you so much

opencti-platform/opencti-front/src/utils/Charts.jsx
```diff
@@ -24,6 +24,16 @@ export const colors = (temp) => [
   C.grey[temp],
 ];
 
+const isColorTooLight = (hex) => {
```
lndrtrbn:
```suggestion
const isColorCloseToWhite = (hex, threshold = 0.9) => {
```

opencti-platform/opencti-front/src/utils/Charts.jsx
```diff
@@ -24,6 +24,16 @@ export const colors = (temp) => [
...
+  if (!hex) return true;
```
lndrtrbn:
```suggestion
  if (!hex) return false;
```

opencti-platform/opencti-front/src/utils/Charts.jsx
```diff
@@ -24,6 +24,16 @@ export const colors = (temp) => [
...
+  const c = hex.replace('#', '');
+  const r = parseInt(c.substr(0, 2), 16);
+  const g = parseInt(c.substr(2, 2), 16);
+  const b = parseInt(c.substr(4, 2), 16);
+  const luminance = (0.299 * r + 0.587 * g + 0.114 * b) / 255;
+  return luminance > 0.9; // filter on too white colors
```
lndrtrbn:
```suggestion
  return luminance >= threshold; // filter on too white colors
```

opencti-platform/opencti-front/src/utils/Charts.jsx
```diff
@@ -908,10 +918,9 @@ export const treeMapOptions = (
     theme: {
       mode: theme.palette.mode,
     },
-    colors: [
-      theme.palette.primary.main,
-      ...colors(theme.palette.mode === 'dark' ? 400 : 600),
-    ],
+    colors: distributed
+      ? colors(theme.palette.mode === 'dark' ? 400 : 600).filter((c) => !isColorTooLight(c))
```
lndrtrbn:
```suggestion
      ? colors(theme.palette.mode === 'dark' ? 400 : 600).filter((c) => !isColorCloseToWhite(c))
```

opencti-platform/opencti-front/src/utils/Charts.jsx
lndrtrbn: Move this function in `utils/Colors.js`

lndrtrbn (approved):

PR #10138, opened 2025-03-04
[frontend] regardingOf filter display in tasks details (#10099)

### LINKED ISSUE #10099: Error in tasks filter groups details if regardingOf filter
## Description
Can't open the filter groups details pop up in background task description in the filter groups contain a 'in regards of' filter

## Reproducible Steps
- In data/entities, add a 'in regards of filter'
- Select all the checkboxes

Image

- Do a background task to delete the selected entities (ie Click on the 'delete' button in the tool bar)
- In the Pop-up that opens, click on 'filters are not fully displayed' orange button to see the not displayed filter groups

Image

- An error occurs

## Expected output
The filter groups containing regardingOf filter is displayed

## Comment
No error if the filter is not a regardingOf filter, because the format of this filter is special

Image

clicking in the 'filters are not fully displayed button' is working as expected : 

Image

Archidoit:
### Proposed changes
Special display for 'in regards of' filters in background tasks filter groups pop-up

### Related issues
https://github.com/OpenCTI-Platform/opencti/issues/10099

### DIFF
--- a/opencti-platform/opencti-front/src/components/filters/DisplayFilterGroup.tsx
+++ b/opencti-platform/opencti-front/src/components/filters/DisplayFilterGroup.tsx
@@ -124,9 +124,9 @@ const DisplayFilterGroup: FunctionComponent<DisplayFilterGroupProps> = ({
               {key === 'regardingOf'
                 ? <>
                   {values.filter((v) => v.key === 'relationship_type').flat().map((value) => {
-                    return (<>
+                    return (<span key={'relationship_type'}>
                       {displayValues(value.values)}
-                    </>);
+                    </span>);
                   })}
                   <Box
                     sx={{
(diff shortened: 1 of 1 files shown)


opencti-platform/opencti-front/src/components/filters/DisplayFilterGroup.tsx
```diff
@@ -37,66 +37,97 @@ const DisplayFilterGroup: FunctionComponent<DisplayFilterGroupProps> = ({
   const handleClose = () => {
     setOpen(false);
   };
+  const displayValues = (values: string[], mode?: string) => {
```
lndrtrbn: It's a good idea to separate the logic as it is used multiple times :+1: but I'm not a fan of making an internal function returning JSX. It's a better practice to make an other component instead. You can keep it in the same file

Archidoit: fixed !!

lndrtrbn (approved):

PR #10154, opened 2025-03-05
[backend] Bug correction for setting x_opencti_score on SCO. Uses confidence factor value from User/Group

### LINKED ISSUE #10153: Confidence value on SCO (x_opencti_score) via connectors/backend is not properly set via user/group max confidence value.
## Description
Maybe x_opencti_score isn't a direct corollary to "confidence" for an SCO, however it is required for decay rules, so just in case:

The new PR#(https://github.com/OpenCTI-Platform/connectors/pull/3526) removes the setting of confidence value directly via a connector. It is expected to inherit the value via the User/Group setting of the users running the connector. However, SCO objects use `x_opencti_score` and not `confidence` value as the key in the DB that tracks this value. Thus, all SCOs will not be created with a Confidence value (x_opencti_score) - it will be set to None. Thinking there should be a similar behavior for `x_opencti_score`, based on the confidence level that the User/Group creating the record is rated.

I believe the None is a bug in itself, since the docs seem to indicate all should have 50 to start at a minimum or be set by the connector/datasource. (https://docs.opencti.io/latest/usage/indicators-lifecycle/#score-decay) I would suspect that many users set a confidence value for the connector and expect that to fall through to`x_opencti_score` where appropriate.

`middleware.js` needs to be updated to account for this difference. See PR https://github.com/OpenCTI-Platform/opencti/pull/10154

## Environment

1. OS (where OpenCTI server runs): Mac
2. OpenCTI version: openCTI 6.5.3
3. OpenCTI client: Backend
4. Other environment details:

## Reproducible Steps

1. Using PyCTI or a Connector and running as a specific user create an SCO

Example code:

```
                        response = self.helper.api.stix_cyber_observable.create(
                            type="Artifact",
                            createIndicator=self.create_indicator,
                            x_opencti_additional_names=sha256,
                            objectMarking=stix2.TLP_WHITE["id"],
                            observableData={
                                "type": "Artifact",
                                "mime_type": mime_type,
                                "x_opencti_additional_names": [file_name],
                                "hashes": {
                                    "sha-1": sha1,
                                    "sha-256": sha256,
                                    "sha-512": sha512,
                                    "md5": md5,
                                },
                                # Connector OpenCTI 6.0+ - Enforced by OpenCTI Connector User Settings
                                # "x_opencti_score": self.helper.connect_confidence_level,
                                "x_opencti_description": f"Created by some direct code",
                                "payload_bin": base64.b64encode(file_contents),
                            },
```

The score/ confidence value will be None

2. Using PyCTI or a Connector and running as a specific user create an SCO and pass the score....

Example code:
```
                        response = self.helper.api.stix_cyber_observable.create(
                            type="Artifact",
                            createIndicator=self.create_indicator,
                            x_opencti_additional_names=sha256,
                            objectMarking=stix2.TLP_WHITE["id"],
                            observableData={
                                "type": "Artifact",
                                "mime_type": mime_type,
                                "x_opencti_additional_names": [file_name],
                                "hashes": {
                                    "sha-1": sha1,
                                    "sha-256": sha256,
                                    "sha-512": sha512,
                                    "md5": md5,
                                },
                                # Will now have a Confidence value / Score.....
                                "x_opencti_score": 57, # self.helper.connect_confidence_level is None by default now....
                                "x_opencti_description": f"Created by some direct code",
                                "payload_bin": base64.b64encode(file_contents),
                            },
```
Will have a Confidence value / score since x_opencti_score is set.

3. Update GraphQL backend with the patch proposed in (https://github.com/OpenCTI-Platform/opencti/pull/10154). Re-execute Option 1, again - not passing an x_opencti_score 

The SCO will have the confidence/x_opencti_score value inherited from the user/group as expected.

## Actual Output

Option 1 - has no score. Since `middleware.js` is using `input.confidence` to set the confidence value. And a SCO doesn't have a confidence field, it uses `input.x_opencti_score`. Option 2 will have a score, since we set the `x_opencti_score`.

## Expected Output

Option 2 will have a score, since we set the `x_opencti_score`. Option 1 and 3 will have a score inherited since it was fixed by the behavior in the PR for `middleware.js`

ParamConstructor:
Maybe x_opencti_score isn't a direct corollary to "confidence" for an SCO, however it is required for decay rules, so just in case:

The new PR#(https://github.com/OpenCTI-Platform/connectors/pull/3526) removes the setting of confidence value directly via a connector. It is expected to inherit the value via the User/Group setting of the users running the connector. However, SCO objects use `x_opencti_score` and not `confidence` value as the key in the DB that tracks this "similar" value. Thus, all SCOs would not be created with a "confidence" value (x_opencti_score) (i.e. would be set to None). This PR seeks to create a similar behavior for `x_opencti_score`, based on the confidence level that the User/Group creating the record is rated.

`opencti-graphql/src/domain/stixCyberObservable.js` has been updated to account for this difference.

#### Platform UI Front Level
NOTE: The Artifact create drawer within the Frontend off - `/dashboard/observations/artifacts` - does **not** have a score field on it, so artifacts created via this panel will now inherit the score of the creator. However, the drawer off - `/dashboard/observations/observables` - to create and select Artifact - does have a score field, defaulted to 50.

Previous behavior would be a score of `None` set off  `/dashboard/observations/artifacts` and a score of 50 (or whatever the user defined) set off `/dashboard/observations/observables`. I believe the `None` is a bug in itself, since the docs seem to indicate all should have 50 to start at a minimum. (https://docs.opencti.io/latest/usage/indicators-lifecycle/#score-decay)

To fix / make the behavior the same (setting to 50 or a selected value) - **artifactImport** would require the support of `x_opencti_score`, which it currently doesn't support. This is viewed as a separate tech debt PR should this behavior correction be required. They system will now off this screen (`/dashboard/observations/artifacts`) set the score to that of the user creating versus `None`

#### Connector Level
Connectors that pass bundles that do **not** have an `x_opencti_score` in them will set the `x_opencti_score` to the `confidence level` of the user that the connector is currently running under. So, if the connector has a `confidence level` of `90` and **no** `x_opencti_score` is passed, the observable will be created with an `x_opencti_score` of `90`. The previous behavior would be that the `x_opencti_score` would be empty. This empty behavior is a violation of the intended documentation pattern which indicates all should have `50` to start at a minimum. (https://docs.opencti.io/latest/usage/indicators-lifecycle/#score-decay)

 If an `x_opencti_score` is passed, the value passed will be the setting upon creation of the observable.

### Checklist

### Further comments

See Issue - https://github.com/OpenCTI-Platform/opencti/issues/10153

### DIFF
--- a/opencti-platform/opencti-graphql/src/database/middleware.js
+++ b/opencti-platform/opencti-graphql/src/database/middleware.js
@@ -3156,10 +3156,6 @@ const createEntityRaw = async (context, user, rawInput, type, opts = {}) => {
   const input = { ...rawInput };
   const { confidenceLevelToApply } = controlCreateInputWithUserConfidence(user, input, type);
   input.confidence = confidenceLevelToApply; // confidence of new entity will be capped to user's confidence
-  // StixCyberObservables use x_opencti_score versus having a confidence value field. Creates will not
-  // default populate this field, unless x_opencti_score is set. Connectors cannot set this value based on
-  // default user it is running due to PR#3526 in OpenCTI Connectors removed the passing of these settings.
-  input.x_opencti_score = confidenceLevelToApply; // x_opencti_score of new entity will be capped to user's confidence
   // authorized_members renaming
   if (input.authorized_members?.length > 0) {
     input.restricted_members = input.authorized_members;
--- a/opencti-platform/opencti-graphql/src/database/middleware.js
+++ b/opencti-platform/opencti-graphql/src/database/middleware.js
@@ -3156,6 +3156,10 @@ const createEntityRaw = async (context, user, rawInput, type, opts = {}) => {
   const input = { ...rawInput };
   const { confidenceLevelToApply } = controlCreateInputWithUserConfidence(user, input, type);
   input.confidence = confidenceLevelToApply; // confidence of new entity will be capped to user's confidence
+  // StixCyberObservables use x_opencti_score versus having a confidence value field. Creates will not
+  // default populate this field, unless x_opencti_score is set. Connectors cannot set this value based on
+  // default user it is running due to PR#3526 in OpenCTI Connectors removed the passing of these settings.
+  input.x_opencti_score = confidenceLevelToApply; // x_opencti_score of new entity will be capped to user's confidence
   // authorized_members renaming
   if (input.authorized_members?.length > 0) {
     input.restricted_members = input.authorized_members;
--- a/opencti-platform/opencti-graphql/src/domain/stixCyberObservable.js
+++ b/opencti-platform/opencti-graphql/src/domain/stixCyberObservable.js
@@ -42,7 +42,7 @@ import { checkScore, now, observableValue } from '../utils/format';
 import { stixObjectOrRelationshipAddRefRelation, stixObjectOrRelationshipDeleteRefRelation } from './stixObjectOrStixRelationship';
 import { addFilter } from '../utils/filtering/filtering-utils';
 import { ENTITY_TYPE_INDICATOR } from '../modules/indicator/indicator-types';
-import { controlUserConfidenceAgainstElement } from '../utils/confidence-level';
+import { controlCreateInputWithUserConfidence, controlUserConfidenceAgainstElement } from '../utils/confidence-level';
 import { uploadToStorage } from '../database/file-storage-helper';
 import { isNumericAttribute } from '../schema/schema-attributes';
 
@@ -178,7 +178,6 @@ export const addStixCyberObservable = async (context, user, input) => {
   const {
     internal_id,
     stix_id,
-    x_opencti_score,
     x_opencti_description,
     createdBy,
     objectMarking,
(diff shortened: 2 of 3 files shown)


romain-filigran:
@ParamConstructor : Thank you for the contribution and sorry for the late reply.

We discussed this pull request internally, particularly regarding the mapping of the confidence level associated with a user to the score of an observable or indicator. We are aligned on the approach, which consists of applying the confidence level of the user who created the observable or indicator only when it is not provided by the source or connector.

However, the code proposed in this pull request appears to be too generic and should ideally be restricted to entities such as Indicators and Observables. For instance, we are currently updating the Organization model to include a new 'x_opencti_score' field, and in this context, mapping the confidence level to the score does not seem appropriate.

@richard-julien : Can you add some recommendations on PR?

opencti-platform/opencti-graphql/src/database/middleware.js
```diff
@@ -3087,6 +3087,10 @@ const createEntityRaw = async (context, user, rawInput, type, opts = {}) => {
   const input = { ...rawInput };
   const { confidenceLevelToApply } = controlCreateInputWithUserConfidence(user, input, type);
   input.confidence = confidenceLevelToApply; // confidence of new entity will be capped to user's confidence
+  // StixCyberObservables use x_opencti_score versus having a confidence value field. Creates will not
+  // default populate this field, unless x_opencti_score is set. Connectors cannot set this value based on
+  // default user it is running due to PR#3526 in OpenCTI Connectors removed the passing of these settings.
+  input.x_opencti_score = confidenceLevelToApply; // x_opencti_score of new entity will be capped to user's confidence
```
aHenryJard: There are several entity with score, please move the changes to https://github.com/OpenCTI-Platform/opencti/blob/087cb5356bf99b046b703b89a4e0c3d82a4a793e/opencti-platform/opencti-graphql/src/domain/stixCyberObservable.js#L166 to target only observables.

Changes might be will be required in the stixCyberObservableEditField in the same file too.

ParamConstructor: @aHenryJard / @romain-filigran - This change has been made. The change was tested with connectors and within the platform as documented above in the description. This is ready for a review at your convenience.

Also, a slight change was made to the order of operations in the validation pattern of checkScore(). Believe it best to verify the value is a `Number` first, then if a `Number` check the range on it.

richard-julien:
Hi @ParamConstructor,
After looking at this PR, i think its not the correct approach to solve this problem.
Using the max_confidence level of a user as a default value could work but it will generate difficulties to manage the correct confidence in front of the expected score for the same user.
I think this problem must be firstly addressed in the connectors that send the information. It was previously link to confidence but as its no longer the same concern, bring back a default score in the configuration make sens.
Do you have the list of the connectors we need to take care to bring back this option?

Then to be able to do it without impacting the connector, we need more a kind of pre processing engine where following some rules, the platform will be able to adapt data before ingestion.

ParamConstructor:
@richard-julien - Yes - adding x_opencti_score back to specific connectors - would solve our initial issue. So - that would be an acceptable approach from our side.

There is still the random issue in the platform side - see the Platform UI Front Level note on the PR description. The Artifact create drawer within the Frontend off - `/dashboard/observations/artifacts` - does not have a score field on it, so artifacts created via this panel get a `None` score versus a base of `50` like is suggested in the documentation  (https://docs.opencti.io/latest/usage/indicators-lifecycle/#score-decay).

richard-julien:
@ParamConstructor, do you have the list of the connectors we need to take care to bring back this option?

PR #10163, opened 2025-03-06
[frontend/backend] Dynamic 'me' filter value (#9128)

### LINKED ISSUE #9128: [Filters]: Dynamic "Me'" Filter
## Use case

When developing dashboards or “saved searches” (in the future), it would be interesting to have the equivalent of a “Me” filter that could be used on attributes of type “Assignee”, “Participant”, “Author”. 
This would make it possible to create dynamic dashboards integrating visualizations linked to the connected user (ex: List of incidents assigned to me)

## Current Workaround

N/A

## Proposed Solution

Introduce a dynamic "Me" filter

Archidoit:
### Proposed changes
For filters involving users id as value, add the possibility to set a dynamic value @me representing the current user

to test:
- dynamic filtering (lists, exports, fintel templates, widgets, public dashboards)
- stix filtering (triggers, stream, feed, playbook etc)

### Related issues
https://github.com/OpenCTI-Platform/opencti/issues/9128

### DIFF
--- a/opencti-platform/opencti-front/src/utils/filters/useSearchEntities.tsx
+++ b/opencti-platform/opencti-front/src/utils/filters/useSearchEntities.tsx
@@ -505,7 +505,7 @@ const useSearchEntities = ({
       type: string,
       data: IdentitySearchCreatorsSearchQuery$data['creators'], // this type is actually the same for the different queries we use, not only creators
     ) => {
-      const newOptions: { label: string, value: string, type: string, color?: string }[] = (data?.edges ?? []).map((n) => ({
+      const newOptions: EntityValue[] = (data?.edges ?? []).map((n) => ({
         label: n?.node.name ?? '',
         value: n?.node.id ?? '',
         type,
--- a/opencti-platform/opencti-front/src/utils/filters/useSearchEntities.tsx
+++ b/opencti-platform/opencti-front/src/utils/filters/useSearchEntities.tsx
@@ -35,7 +35,7 @@ import { fetchQuery } from '../../relay/environment';
 import { useSearchEntitiesSchemaSCOSearchQuery$data } from './__generated__/useSearchEntitiesSchemaSCOSearchQuery.graphql';
 import type { Theme } from '../../components/Theme';
 import useAttributes, { containerTypes } from '../hooks/useAttributes';
-import { contextFilters, entityTypesFilters } from './filtersUtils';
+import { contextFilters, entityTypesFilters, ME_FILTER_VALUE } from './filtersUtils';
 import { useSearchEntitiesDashboardsQuery$data } from './__generated__/useSearchEntitiesDashboardsQuery.graphql';
 import { convertMarking } from '../edition';
 
@@ -266,6 +266,12 @@ const useSearchEntities = ({
     const currentMap = schema.filterKeysSchema.get(entityType);
     currentMap?.forEach((value, key) => filterKeysMap.set(key, value));
   });
+  const meEntity = {
+    label: ME_FILTER_VALUE,
+    value: ME_FILTER_VALUE,
+    type: 'User',
+    color: theme.palette.primary.main,
+  };
   const unionSetEntities = (key: string, newEntities: EntityValue[]) => setEntities((c) => ({
     ...c,
     [key]: [...newEntities, ...(c[key] ?? [])].filter(
@@ -476,6 +482,9 @@ const useSearchEntities = ({
             group: n?.node.entity_type,
           }));
           unionSetEntities(key, membersEntities);
+          if (entityTypes.includes('User')) { // add @me filter value
+            unionSetEntities(key, [meEntity]);
+          }
           const membersSystems = (
             (data as ObjectAssigneeFieldMembersSearchQuery$data)?.systemMembers
               ?.edges ?? []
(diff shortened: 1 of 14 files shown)


opencti-platform/opencti-front/src/utils/filters/useSearchEntities.tsx
```diff
@@ -496,7 +505,7 @@ const useSearchEntities = ({
       type: string,
       data: IdentitySearchCreatorsSearchQuery$data['creators'], // this type is actually the same for the different queries we use, not only creators
     ) => {
-      const newOptions = (data?.edges ?? []).map((n) => ({
+      const newOptions: { label: string, value: string, type: string, color?: string }[] = (data?.edges ?? []).map((n) => ({
```
lndrtrbn: We do not have a defined type somewhere for this?

opencti-platform/opencti-front/src/utils/filters/useSearchEntities.tsx
```diff
@@ -563,7 +576,7 @@ const useSearchEntities = ({
               .toPromise()
               .then((response) => {
                 const data = response as ObjectAssigneeFieldAssigneesSearchQuery$data;
-                buildCachedOptionsFromGenericFetchResponse(filterKey, 'Individual', data?.assignees);
+                buildCachedOptionsFromGenericFetchResponse(filterKey, 'User', data?.assignees);
```
lndrtrbn, line 579: Could you explain the reason of this change?

opencti-platform/opencti-graphql/src/utils/filtering/filtering-constants.ts
```diff
@@ -155,3 +155,14 @@ export const specialFilterKeysWhoseValueToResolve = [
   RELATION_FROM_FILTER,
   RELATION_TO_FILTER
 ];
+
+// special filter values
+export const ME_FILTER_VALUE = '@me';
+export const filterKeysWithMeValue = [
```
lndrtrbn, line 161: As it's a const, should we not use only uppercase?

opencti-platform/opencti-graphql/src/utils/filtering/filtering-utils.ts
```diff
@@ -279,6 +281,26 @@ const getConvertedRelationsNames = (relationNames: string[]) => {
...
+export const replaceMeValuesInFilters = (filterGroup: FilterGroup, userId: string) => {
+  const filtersResult = { ...filterGroup };
+  filtersResult.filters.forEach((filter) => {
+    const { key } = filter;
+    const arrayKeys = Array.isArray(key) ? key : [key];
+    if (arrayKeys.some((filterKey) => filterKeysWithMeValue.includes(filterKey))) {
+      // replace ME_FILTER_VALUE with the id of the user
+      if (filter.values.includes(ME_FILTER_VALUE)) {
+        // eslint-disable-next-line no-param-reassign
+        filter.values = filter.values.map((v) => (v === ME_FILTER_VALUE ? userId : v));
```
lndrtrbn, line 296: You could avoid using an eslint ignore by mapping through `filterGroup.filters` instead of making a `forEach`

```js
const filtersResult = {
    filters: filterGroup.filters.map(filter => {
        // ...
        return {...}
    })
};
```

lndrtrbn: Ok I didn't see we have recursivity below. That would be still possible to do it without reasigning but probably harder to read. So you can ignore my comment.

opencti-platform/opencti-graphql/src/utils/filtering/filtering-resolution.ts
```diff
@@ -98,6 +100,14 @@ const resolveFilter = async (
     newFilterValues = statusIds;
   }
 
+  // 3. handle the special case of dynamic @me value
+  if (key.some((k) => filterKeysWithMeValue.includes(k))) {
+    if (filter.values.includes(ME_FILTER_VALUE)) {
+      // eslint-disable-next-line no-param-reassign
+      filter.values = filter.values.map((v) => (v === ME_FILTER_VALUE ? user.id : v));
+    }
```
lndrtrbn, line 108: I do not understand why we need to do this here. You already made a function `replaceMeValuesInFilters`, why this new function is not called somewhere to avoid those lines?

opencti-platform/opencti-front/src/utils/filters/useSearchEntities.tsx
Archidoit: There is the type Option but it is not exactly that

opencti-platform/opencti-front/src/utils/filters/useSearchEntities.tsx
Archidoit: It's not individuals, but users that are fetched for assignees, so it was a mistake in the code

opencti-platform/opencti-graphql/src/utils/filtering/filtering-utils.ts
Archidoit: ok :)

opencti-platform/opencti-graphql/src/utils/filtering/filtering-resolution.ts
Archidoit: This is for stix filtering (triggers, stream etc), whereas replaceMeValuesInFilters is for dynamic filtering (list of entities, widgets)

opencti-platform/opencti-graphql/src/utils/filtering/filtering-constants.ts
Archidoit: the other constants with filter keys are in lower case so I keep the same logic (example: specialFilterKeysWhoseValueToResolve, specialFilterKeys, etc)

opencti-platform/opencti-front/src/utils/filters/useSearchEntities.tsx
Archidoit: done

lndrtrbn (approved):

Issue #9871, opened 2025-02-07
Missing observable name when creating sighting from knowledge graph


labo-flg:
## Description

Missing representative of the observable in the sighting creation drawer.

## Environment

OCTI 6.5.1

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. open the knowledge graph of a report
2. add an indicator URL and an observable URL
3. select the two and create a sighting with the toolbar button at the bottom right

![Image](https://github.com/user-attachments/assets/85774401-ea29-4191-a80d-410fc3e03aaf)

PR #10164, opened 2025-03-06
[frontend] Missing observable name when creating sighting from knowledge graph (#9871)

SarahBocognano:
### Proposed changes

* Add mainRepresentative

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/9871

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/events/stix_sighting_relationships/StixSightingRelationshipCreationForm.js
+++ b/opencti-platform/opencti-front/src/private/components/events/stix_sighting_relationships/StixSightingRelationshipCreationForm.js
@@ -180,15 +180,17 @@ const StixSightingRelationshipCreationForm = ({
                   </div>
                 </div>
                 <div className={classes.content}>
-                  <Tooltip title={getMainRepresentative(fromEntity)}>
+                  {isMultipleFrom ? (
                     <span className={classes.name}>
-                      {isMultipleFrom
-                        ? (<em>{t_i18n('Multiple entities selected')}</em>)
-                        : (
-                          truncate(getMainRepresentative(fromEntity), 20)
-                        )}
+                      <em>{t_i18n('Multiple entities selected')}</em>
                     </span>
-                  </Tooltip>
+                  ) : (
+                    <Tooltip title={getMainRepresentative(fromEntity)}>
+                      <span className={classes.name}>
+                        {truncate(getMainRepresentative(fromEntity), 20)}
+                      </span>
+                    </Tooltip>
+                  )}
                 </div>
               </div>
               <div className={classes.middle} style={{ paddingTop: 25 }}>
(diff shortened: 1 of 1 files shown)


opencti-platform/opencti-front/src/private/components/events/stix_sighting_relationships/StixSightingRelationshipCreationForm.js
```diff
@@ -242,7 +230,7 @@ const StixSightingRelationshipCreationForm = ({
                   <span className={classes.name}>
                     {isMultipleTo
                       ? (<em>{t_i18n('Multiple entities selected')}</em>)
-                      : (defaultName(toEntity))}
+                      : (getMainRepresentative(toEntity))}
```
Archidoit: the value displayed is not troncated to 20 anymore... Problems with long representatives

Archidoit (changes requested):

Archidoit:
Problem if representative is too long -> troncate the representative to 20 

image

Archidoit:
The representative displayed do not seem correct (a date is added after the name 'Fedora' for instance) : 
image


Note : It is also the case when creating classic relationships, --> is it a regression or something we want @romain-filigran  ? 
image

SarahBocognano:
Thanks for your feedback ! I truncated it as required, but still waiting for PO's point of view regarding of the date next to some observables by using main representative

nino-filigran:
For me, date should not be present: indeed, in the sighting creation we have already some dates, so it should be fine without it.

nino-filigran:
After discussion, date is evrywhere. to avoid any regression, we'll keep it as-is.

opencti-platform/opencti-front/src/private/components/events/stix_sighting_relationships/StixSightingRelationshipCreationForm.js
```diff
@@ -239,11 +233,13 @@ const StixSightingRelationshipCreationForm = ({
                   </div>
                 </div>
                 <div className={classes.content}>
-                  <span className={classes.name}>
-                    {isMultipleTo
-                      ? (<em>{t_i18n('Multiple entities selected')}</em>)
-                      : (defaultName(toEntity))}
-                  </span>
+                  <Tooltip title={getMainRepresentative(toEntity)}>
```
Archidoit: if there are several entities selected (multipleTo = true), we should display all the entities representatives in the tooltip, not only the first one

opencti-platform/opencti-front/src/private/components/events/stix_sighting_relationships/StixSightingRelationshipCreationForm.js
```diff
@@ -190,11 +180,15 @@ const StixSightingRelationshipCreationForm = ({
                   </div>
                 </div>
                 <div className={classes.content}>
-                  <span className={classes.name}>
-                    {isMultipleFrom
-                      ? (<em>{t_i18n('Multiple entities selected')}</em>)
-                      : (defaultName(fromEntity))}
-                  </span>
+                  <Tooltip title={getMainRepresentative(fromEntity)}>
```
Archidoit: same here

Archidoit (changes requested):

opencti-platform/opencti-front/src/private/components/events/stix_sighting_relationships/StixSightingRelationshipCreationForm.js
SarahBocognano: As it's not present right now in the platform (both for relationships and sightings) and that the product didn't ask for it, I will get a tooltip only if there is one entity. Nothing will be displayed if there is more than one except for the "Multiple entities selected" to keep the same behavior.

opencti-platform/opencti-front/src/private/components/events/stix_sighting_relationships/StixSightingRelationshipCreationForm.js
SarahBocognano: As it's not present right now in the platform (both for relationships and sightings) and that the product didn't ask for it, I will get a tooltip only if there is one entity. Nothing will be displayed if there is more than one except for the "Multiple entities selected" to keep the same behavior.

Archidoit:
When creating a relationship from knowledge graph : 
image


When creating a sighting : 
image


--> align the troncate number of sighting to the one of relationships

Archidoit:
Troncate value is not align for nested relationships created in graphs: 
image

Archidoit (approved):

Issue #9751, opened 2025-01-29
[POC]: Natural Language Query capabilities in OpenCTI


romain-filigran:
## Use case

Develop a POC for a Natural Language Query (NLQ) feature in OpenCTI. 
The goal is to allow users to input queries in natural language and convert them into a set of OpenCTI filters.

PR #10177, opened 2025-03-07
[frontend] Fix filters edition after NLQ search (#9751)

Archidoit:
### Proposed changes
Fix filters edition after NLQ search (frontend)
- Be able to modify the filters after a NLQ search
- The entities list should be updated according to these filters changes


CelineSebe:
Filters edition works 👌

CelineSebe (approved):

PR #10215, opened 2025-03-10
[backend] Check date filters values validity (#9751)

Archidoit:
### Proposed changes
Improve filters validity checking in backend:
- date filters values should be either in relative date math format, either a valid date understood by elastic

Enables not to make the screen crashes when doing a NLQ search returning filters with a date filter in bad format

Exemple: 
Do a NLQ search with the question: 'Give me the 10 most recent reports'
--> The screen should not throw an error anymore. You should now have a message like this : 
image


Archidoit:
Cancelled for the moment with the last changes in NLQ model

Issue #9171, opened 2024-11-26
Adding external references is failing when the user has a lower confidence level than the entity


gileri:
## Description

[The sentinel-intel connector adds a relationship after publishing data to the Sentinel API](https://github.com/OpenCTI-Platform/connectors/blob/fbdab76582dff61262068d17687292d20861c6fc/stream/sentinel-intel/src/sentinel_intel_connector/connector.py#L115-L131
).

However, it fails when it does so on a entity with an higher Confidence Level than the maximum Confidence Level set on the connector user.

## Environment

1. OS (where OpenCTI server runs): Official docker image
2. OpenCTI version: 6.4.0
3. OpenCTI client: python
4. Other environment details:

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Run the connector
2. Create an indicator that is picked up by the connector stream

## Expected Output

No error, external reference is created

## Actual Output

```json
{"timestamp": "2024-11-26T11:35:04.191941Z", "level": "ERROR", "name": "Sentinel Intel",
[REDACTED]
"exc_info": "Traceback (most recent call last):\n  File \"/opt/opencti-connector-sentinel-intel/sentinel_intel_connector/connector.py\", line 256, in process_message\n    self._handle_create_event(data)\n  File \"/opt/opencti-connector-sentinel-intel/sentinel_intel_connector/connector.py\", line 188, in _handle_create_event\n    self._create_sentinel_indicator(observable)\n  File \"/opt/opencti-connector-sentinel-intel/sentinel_intel_connector/connector.py\", line 123, in _create_sentinel_indicator\n    self.helper.api.stix_domain_object.add_external_reference(\n  File \"/usr/local/lib/python3.11/site-packages/pycti/entities/opencti_stix_domain_object.py\", line 1703, in add_external_reference\n    self.opencti.query(\n  File \"/usr/local/lib/python3.11/site-packages/pycti/api/opencti_api_client.py\", line 368, in query\n    raise ValueError(value_error)\nValueError: {'name': 'FUNCTIONAL_ERROR', 'error_message': 'User effective max confidence level is insufficient to update this element'}"}
```

SamuelHassine:
Transferring to OpenCTI as adding an external reference to an entity with a lower confidence level should be possible. Also, I'm seeing a risk here to have the external reference created but not associated to anything.

nino-filigran:
[...]

PR #10217, opened 2025-03-10
[backend] do not check confidence when adding external ref or label (#9171)

lndrtrbn:
### Proposed changes

* 

### Related issues

* #9171 

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/database/middleware.js
+++ b/opencti-platform/opencti-graphql/src/database/middleware.js
@@ -203,7 +203,8 @@ import {
   adaptUpdateInputsConfidence,
   controlCreateInputWithUserConfidence,
   controlUpsertInputWithUserConfidence,
-  controlUserConfidenceAgainstElement
+  controlUserConfidenceAgainstElement,
+  shouldCheckConfidenceOnRefRelationship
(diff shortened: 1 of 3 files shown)


opencti-platform/opencti-graphql/src/database/middleware.js
```diff
@@ -2257,7 +2258,13 @@ export const updateAttributeFromLoadedWithRefs = async (context, user, initial,
     throw FunctionalError('Cant update undefined element');
   }
   // region confidence control
-  controlUserConfidenceAgainstElement(user, initial);
+  const checkConfidence = (Array.isArray(inputs) ? inputs : [inputs]).some(({ key, operation }) => {
```
aHenryJard, line 2261: If inputs is an array and only one key can skip the confidence check, does it skip for all ?

lndrtrbn: Nop, the `some` function returns `true` if at least one element matches. So here if at least one needs to be checked then the const will be `true` also

aHenryJard (approved):
Wouls be great if you find a way to add test coverage in a follow up PR. Tested locally, all good to me.

PR #10236, opened 2025-03-13
[frontend] Scrollable User History

ParamConstructor:
The current User profile view could use two improvements for display purposes:

First, it does not support fetching and listing more than 10 Activity History entries which isn't really enough history to determine what the user has been doing recently. [ENHANCEMENT]

Secondly, the Session History display area is not bounded, so if you have say 10-20 session entries they make the form expand in height, versus just limiting a height and making the area scrollable. [BUG]

Here is an example of a user with a "large amount" of active sessions to demonstrate this issue.
image

### Proposed changes

* **Fetch the user's last 500 actions to display in a scrollable (if needed) Activity History Area after 13 entries (Aligned to left Operations widget in same row):**

`**User that requires Scrolling History**`
image

`**User not requiring scrolling**`
image

* **Allow Session History to display as scrollable, if required:**

`**Example requiring scrolling:**`

image

`**Example not requiring scrolling:**`

image

### Related issues

* None

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/settings/users/User.tsx
+++ b/opencti-platform/opencti-front/src/private/components/settings/users/User.tsx
@@ -608,7 +608,7 @@ const User: FunctionComponent<UserProps> = ({ data, refetch }) => {
                 </Security>
                 <div className="clearfix" />
                 <FieldOrEmpty source={orderedSessions}>
-                  <List style={{ marginTop: -2 }}>
+                  <List style={{ marginTop: -2, width: '100%', maxHeight: 400, overflowY: 'auto'}}>
                     {orderedSessions
                       && orderedSessions.map((session: Session) => (
                         <ListItem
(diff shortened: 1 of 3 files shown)


opencti-platform/opencti-front/src/private/components/settings/users/User.tsx
```diff
@@ -676,7 +676,7 @@ const User: FunctionComponent<UserProps> = ({ data, refetch }) => {
         </Grid>
         <Triggers recipientId={user.id} filterKey="authorized_members.id" />
         <Grid item xs={6} style={{ marginTop: 10 }}>
-          <Typography variant="h4" gutterBottom={true}>
+          <Typography variant="h4" gutterBottom={true} style={{ paddingBottom: '22px' }}>
```
ParamConstructor, line 672: Needed for top alignment of the Operation widget with the History Widget "Paper" Area. In master, this is off alignment slightly since the History widget has a search bar and 2 action icons that shift it down.

labo-flg:
@ParamConstructor closing release/6.6.0 closed this PR automatically.
You can rebase on `release/current` and reopen this PR to target `release/current`.

As for the review, I had a look.
All good for the fix on the scrollable area.
As for the history, I'm not sure it's the right approach.
Today this query is apparently very slow, even for 10 entries (which is a problem...) Pushing to 500 can only get it worse.
The query is also repeated every few seconds while we're on this screen, to update latest history.

Ideally, I would go for this approach:

*  Keep the last 10 actions in the main view. This panel is just to see what's happened recently, for a quick view.
* add a button on top of the Paper to "Browse user's history", that would open a drawer with a paginated data table to query the full history, with search and filters (time range for instance, we have a new operator 'within' in 6.6 for that).

Now I understand this is significantly more work, we can discuss it separately and already merge the fix on the scrollable area first.

PR #10253, opened 2025-03-17
Taxii timestamp requires YYYY-MM-DDTHH:MM:SS.ssssssZ format instead of YYYY-MM-DDTHH:MM:SS.sssZ

### LINKED ISSUE #10090: fix(ingestion): resolve TAXII download problems due to wrong time resolution for added-after query
## Description

We provide a Taxi 2.1 server instance for our customers, and we encountered a problem while openCPTI consumed data from the server.

The server limits the returned objects during a query to 100 objects and sets the more to true and the next flag in the response. Opencti will page through the responses until we respond with more false. The following query triggered by the workers uses the X-TAXII-Date-Added-Last Value returned with the last response but uses a wrong time resolution.

The server returns X-TAXII-Date-Added-Last: 2025-02-13T19:41:56.302472Z  but the timestamp then used 2025-02-13T19:41:56.302Z during the query.

This behavior can lead the fetching process to consume data again and again.
On our server, we are doing batch exports for several hundred objects, and the timestamps are very close to each other, so it happens from time to time that if this difference is just some microseconds, an amount of data gets requested again, and again because of the millisecond resolution. 

The taxii 2.1 spec specifies that the timestamp needs a microsecond resolution to avoid this kind of problem. 
https://docs.oasis-open.org/cti/taxii/v2.1/os/taxii-v2.1-os.html#_Toc31107512 exactly

## Environment

4. Other environment details:

## Reproducible Steps
descript above

## Expected Output

Store and use the X-TAXII-Date-Added-Last in microsecond resolution 

## Implementation notes

The solution is to use nano date https://www.elastic.co/guide/en/elasticsearch/reference/current/date_nanos.html and also https://opensearch.org/docs/latest/field-types/supported-field-types/date-nanos/

We don't use nano second at all in the platform. So we need to update the shemas to use nano date for the `added_after_start` data that is store in database for Taxii Feeds. It will require a new field and a maybe migration of existing taxii feed configuration for added_after_start.

https://github.com/OpenCTI-Platform/opencti/blob/ceea6cadcab85e2ec4c1a0560fdf6bfe41eb5065/opencti-platform/opencti-graphql/src/modules/ingestion/ingestion-taxii.ts#L49

aHenryJard:
See https://docs.oasis-open.org/cti/taxii/v2.1/os/taxii-v2.1-os.html#_Toc31107512

### Proposed changes

Just a test to reproduce issue for now, still requires implementation.

*
*

### Related issues

* closes #10090

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/tests/02-integration/04-manager/ingestionManager-test.ts
+++ b/opencti-platform/opencti-graphql/tests/02-integration/04-manager/ingestionManager-test.ts
@@ -235,6 +235,50 @@ describe('Verify taxii ingestion', () => {
     // Delete the ingest
     await ingestionDelete(testContext, ADMIN_USER, ingestionPaginatedWithStartDate.internal_id);
   });
+
+  it('Should taxii response with more=false and next date be accurate including microsecond', async () => {
+    // 1. Create ingestion in opencti
+    const input3 : IngestionTaxiiAddInput = {
+      authentication_type: IngestionAuthType.None,
+      collection: 'testcollectionMicrosecond',
+      ingestion_running: true,
+      name: 'taxii ingestion with last page, more = false and date',
+      uri: 'http://test.invalid',
+      version: TaxiiVersion.V21,
+    };
+    const ingestionPaginatedWithNextDate = await addTaxiiIngestion(testContext, ADMIN_USER, input3);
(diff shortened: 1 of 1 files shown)


aHenryJard:
It's just a test to reproduce issue, feel free to reopen if you work on the bugfix.

PR #10265, opened 2025-03-18
[backend/frontend] Check events restrictions in notifications and remove stream message for creation/deletion notifications

Archidoit:


labo-flg:
This one could be covered with some tests I think.

Archidoit:
> This one could be covered with some tests I think.

done :)

opencti-platform/opencti-graphql/src/manager/publisherManager.ts
```diff
@@ -177,8 +197,12 @@ const processLiveNotificationEvent = async (
     let notificationMessage = message;
     if (streamUser && (event as KnowledgeNotificationEvent).streamMessage) {
       const { streamMessage } = event as KnowledgeNotificationEvent;
-      const streamBuiltMessage = `\`${streamUser.name}\` ${streamMessage}`;
+      let streamBuiltMessage = `\`${streamUser.name}\` ${streamMessage}`;
       if (streamBuiltMessage !== notificationMessage) {
+        // for relationships, modify the stream message to keep the parts the stream user has the right to see
+        if ('type' in event.data && event.data.type && event.data.type === 'relationship') {
+          streamBuiltMessage = restrictedStreamMessageForRelationship(streamBuiltMessage, message);
+        }
```
JeremyCloarec: It probably needs to be done for sightings also ?

opencti-platform/opencti-graphql/src/manager/publisherManager.ts
```diff
@@ -165,6 +165,26 @@ const processNotificationEvent = async (
...
+      return `\`${splittedStreamBuiltMessage[1]}\`${splittedStreamBuiltMessage[2]}\`Restricted\` to \`Restricted\``;
+    } if (fromRestricted) {
+      return `\`${splittedStreamBuiltMessage[1]}\`${splittedStreamBuiltMessage[2]}\`Restricted\` to \`${splittedStreamBuiltMessage[5]}\`${splittedStreamBuiltMessage[6]}`;
+    } if (toRestricted) {
+      return `\`${splittedStreamBuiltMessage[1]}\`${splittedStreamBuiltMessage[2]}\`${splittedStreamBuiltMessage[3]}\`${splittedStreamBuiltMessage[4]}\`Restricted\``;
+    }
+  }
+  return streamMessage;
+};
+
```
JeremyCloarec: It's good functionnaly, but it's not robust: if we decide to change the way messages are generated, this code will break or will need to be updated also.
I don't have a full solution in mind, but I'd say we need to refactor message generation in generate-message.js so that we can use the same method to generate the 'restricted' message here, and the original stream message

opencti-platform/opencti-graphql/src/manager/publisherManager.ts
```diff
@@ -177,8 +197,12 @@ const processLiveNotificationEvent = async (
...
     if (streamUser && (event as KnowledgeNotificationEvent).streamMessage) {
       const { streamMessage } = event as KnowledgeNotificationEvent;
-      const streamBuiltMessage = `\`${streamUser.name}\` ${streamMessage}`;
+      let streamBuiltMessage = `\`${streamUser.name}\` ${streamMessage}`;
       if (streamBuiltMessage !== notificationMessage) {
+        // for relationships, modify the stream message to keep the parts the stream user has the right to see
+        if ('type' in event.data && event.data.type && event.data.type === 'relationship') {
+          streamBuiltMessage = restrictedStreamMessageForRelationship(streamBuiltMessage, message);
+        }
         notificationMessage = `${message} - ${streamBuiltMessage}`;
```
SouadHadjiat: the issue comes from this part where we add the streamBuiltMessage instead of keeping only the notification message. We need to understand why we did this, and change it.

Archidoit: it was done by @Kedae in Mai 2024

Archidoit: any idea why you added the stream message @Kedae ?

labo-flg:
Is there any news on this PR ?

Archidoit:
The chosen solution has been completely modified for this PR. See issue.

opencti-platform/opencti-graphql/src/utils/access.ts
```diff
@@ -627,6 +628,25 @@ export const isUserCanAccessStixElement = async (context: AuthContext, user: Aut
   const hasPlatformOrg = !!settings.platform_organization;
   return checkUserCanAccessStixElement(context, user, instance, hasPlatformOrg);
 };
+
+const checkUserCanAccessMarkings = async (user: AuthUser, markingIds: string[]) => {
+  if (isBypassUser(user)) {
+    return true;
+  }
+  const userMarkingIds = (user.allowed_marking || []).map((m) => m.internal_id);
+  if (markingIds.every((m) => userMarkingIds.includes(m))) {
```
SouadHadjiat, line 637: need to check that markingIds is not empty

opencti-platform/opencti-graphql/src/utils/access.ts
```diff
@@ -627,6 +628,25 @@ export const isUserCanAccessStixElement = async (context: AuthContext, user: Aut
...
+  }
+  return false;
+};
+
+export const isUserCanAccessStreamUpdateEvent = async (user: AuthUser, updateEvent: UpdateEvent) => {
+  const relatedRestrictions = updateEvent.context.related_restrictions;
+  if (!relatedRestrictions) {
+    return true;
+  }
+  return checkUserCanAccessMarkings(user, relatedRestrictions.markings ?? []);
```
SouadHadjiat: maybe we don't need to call this function if relatedRestrictions.markings is empty

opencti-platform/opencti-graphql/src/utils/access.ts
Archidoit: done

opencti-platform/opencti-graphql/src/utils/access.ts
Archidoit: done

SouadHadjiat (approved):

PR #10285, opened 2025-03-19
[frontend/backend] Entities ids mapping by entity type for NLQ filters result (#9751)

Archidoit:
### Proposed changes
Improvement of the values mapping for ids filters applied on NLQ filters result:
- take the filter key entity types ids into account in the values mapping search
- enables to look only for users for the filters creators, assignee and participants
- enables to resolve stix meta objects values (like the one for the Markings filter)

Add a message if no filters have been found for a NLQ search in frontend

### Examples of resolution that are now possible

image

image

### Related issues
#9751

### DIFF
--- a/opencti-platform/opencti-front/lang/front/en.json
+++ b/opencti-platform/opencti-front/lang/front/en.json
@@ -2982,6 +2982,7 @@
   "The Max Confidence Level is currently defined at the user level. It overrides Max Confidence Level from user's groups.": "The Max Confidence Level is currently defined at the user level. It overrides Max Confidence Level from user's groups.",
   "The Max Confidence Level is currently inherited from...": "The Max Confidence Level is currently inherited from the group {link}.",
   "The min value cannot be greater than max value": "The min value cannot be greater than max value",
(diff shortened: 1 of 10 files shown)


opencti-platform/opencti-front/lang/front/en.json
```diff
@@ -2982,6 +2982,7 @@
   "The Max Confidence Level is currently defined at the user level. It overrides Max Confidence Level from user's groups.": "The Max Confidence Level is currently defined at the user level. It overrides Max Confidence Level from user's groups.",
   "The Max Confidence Level is currently inherited from...": "The Max Confidence Level is currently inherited from the group {link}.",
   "The min value cannot be greater than max value": "The min value cannot be greater than max value",
+  "The NLQ model didn't find filters corresponding to your question": "The NLQ model didn't find filters corresponding to your question",
```
aHenryJard, line 2985: I'm not sure that NLQ is an "end user" word, at least it should be the whole word like :
```suggestion
  "The NLQ model didn't find filters corresponding to your question": "The Native Language Query model didn't find filters corresponding to your question",
```

aHenryJard (approved):
Since we are using Feature Flag text can be updated later. All good to me.

PR #10302, opened 2025-03-20
[backend] Update decay score when expiration manager revoke an indicator. (#10174)

### LINKED ISSUE #10174: Decay algorithm doesn't update the score for revoked Indicators
## Description

The revoke mechanism works well but  the decay algorithm doesn't  update the score. 

## Environment

OpenCTI version: 6.5.4

## Reproducible Steps

While I couldn't identify any specification on when this behavior occurs, you can observe it when
1. On Indicators view
2. Add the filters Revoked = True AND Score > 20
3. You can now observe the issue

## Expected Output

The “Valid until” date corresponds to the date on which the decay curve reaches the revocation level/score 20. When this is reached, the “Revoked” field is set to True, and decay must update the score. 

It seems that in some cases the score is not updated.

## Actual Output

I observe Indicators objects with Revoked = True, the “Valid until” date has passed, but the score is not at 20

## Implementation target after product feedback

- When an indicator with decay rules is revoked by expiredManager, compute and update score on revoke
- Use a dedicated user in expiredManager
- add detection false in revoke by decay to be consistent with expiredManager

aHenryJard:
### Proposed changes

* Update decay score when expiration manager revoke an indicator.
* Use a dedicated user for the manager for better readability of activity logs

### Related issues

* closes #10174 

### Checklist


marieflorescontact (approved):
tested ok 👌

Issue #10300, opened 2025-03-20
feat(decay-rules): when valid_until is updated, recompute score and revoke value


aHenryJard:
## Description

Decay computation and indicator trio {score, valid until, revoked} requires more consistency.

Extracted from https://github.com/OpenCTI-Platform/opencti/issues/10174

When valid until is updated, compute score back & revoke value

## Environment

1. OS (where OpenCTI server runs): { e.g. Mac OS 10, Windows 10, Ubuntu 16.4, etc. }
2. OpenCTI version: { e.g. OpenCTI 1.0.2 }
3. OpenCTI client: { e.g. frontend or python }
4. Other environment details:

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. { e.g. Run ... }
2. { e.g. Click ... }
3. { e.g. Error ... }

aHenryJard:
This change more complex than expected, since the score is an integer and not a float. It means that a naïve approach would change the valid until that is enter by the user because of math computation. This requires product team refinement before moving on. I keep the draft of current PR but it's paused for now.

PR #10304, opened 2025-03-20
[backend] upsert indicator score if revoked (#10298)

### LINKED ISSUE #10298: Indicator scores not updated during upsert
## Description

When ingesting an Indicator that already exists in the platform, an upsert occurs. During this upsert, if the score of the ingested Indicator is equal to the starting score of the existing Indicator, i.e. its initial score (aka base score), then its score is not updated.

This is very problematic, as revoked Indicators remain revoked even though they are still reported as malicious, only because the score is the equal to the original one.

## Environment

OCTI 6.5.8

## Reproducible Steps

Create this CSV mapper:
- Representation of Indicator
    - Name = A
    - Pattern = B
    - Pattern type = C
    - Score = D

1. Find a Indicator in your platform an Indicator that has been revoked by decay
2. See what its initial score was
3. Create a CSV file with “<name_of_this_indicator>,<pattern_of_this_indicator>,stix,<initial_score>” inside.
4. Import the CSV file using the CSV mapper
5. Look at the score of the Indicator -> it has not been changed, the Indicator remains revoked

## Expected Output

Update the Indicator score if I import it again without taking into account that the score is equal to the initial score / base score.

SouadHadjiat:
### Proposed changes

* if indicator has been revoked, decay score needs to be reset during upsert
* Change the upsert condition to : if the indicator present in db has been revoked, or if it has a different decay base score, or if it has a different score, we upsert the indicator fully with decay. Otherwise we don't change the score & decay attributes.

### Related issues

* #10298 

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/database/middleware.js
+++ b/opencti-platform/opencti-graphql/src/database/middleware.js
@@ -2603,7 +2603,7 @@ const upsertElement = async (context, user, element, type, basePatch, opts = {})
       updatePatch.valid_from = element.valid_from;
       updatePatch.valid_until = element.valid_until;
       // don't reset decay attributes
-      updatePatch.decay_base_score = element.decay_base_score;
+      // updatePatch.decay_base_score = element.decay_base_score; // no need since it's the same score
       updatePatch.revoked = element.revoked;
       updatePatch.decay_base_score_date = element.decay_base_score_date;
       updatePatch.decay_applied_rule = element.decay_applied_rule;
--- a/opencti-platform/opencti-graphql/src/database/middleware.js
+++ b/opencti-platform/opencti-graphql/src/database/middleware.js
@@ -2592,7 +2592,28 @@ const upsertElement = async (context, user, element, type, basePatch, opts = {})
       updatePatch.number_observed = element.number_observed + updatePatch.number_observed;
     }
   }
-
+  if (type === ENTITY_TYPE_INDICATOR) {
+    // Do not compute decay again when base score does not change
+    // if the element was revoked, we need to update the score to reactivate the indicator
+    if (!element.revoked && updatePatch.decay_applied_rule
+      && (updatePatch.decay_base_score === element.decay_base_score && updatePatch.decay_base_score === element.x_opencti_score)) {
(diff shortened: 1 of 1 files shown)


aHenryJard (commented):
On local instance: I created an indicator (with no score, so default = 50) and then revoked it by doing an upsert with a valid until in the past and the score is still 50 despite revoked being true.

How do you test this ?

Also maybe we should find a place to add a test coverage.

Lhorus6:
Hi @aHenryJard,
I think what you are referring to is the test for the fix of this issue: : https://github.com/OpenCTI-Platform/opencti/issues/10300

Here the problem solved by this PR is different. It is the score update that is not taken into account during the upsert. You can find a repro case here: https://github.com/OpenCTI-Platform/opencti/issues/10298

SouadHadjiat:
> On local instance: I created an indicator (with no score, so default = 50) and then revoked it by doing an upsert with a valid until in the past and the score is still 50 despite revoked being true.
> 
> How do you test this ?
> 
> Also maybe we should find a place to add a test coverage.

I tested that an indicator that is revoked already on the platform that had a base score of 50 can be upsert with the same indicator that has a base score of 50 @aHenryJard

opencti-platform/opencti-graphql/src/database/middleware.js
```diff
@@ -2594,14 +2594,16 @@ const upsertElement = async (context, user, element, type, basePatch, opts = {})
   }
   if (type === ENTITY_TYPE_INDICATOR) {
     // Do not compute decay again when base score does not change
-    if (updatePatch.decay_applied_rule && (updatePatch.decay_base_score === element.decay_base_score || updatePatch.decay_base_score === element.x_opencti_score)) {
+    // if the element was revoked, we need to update the score to reactivate the indicator
+    if (!element.revoked && updatePatch.decay_applied_rule
+      && (updatePatch.decay_base_score === element.decay_base_score && updatePatch.decay_base_score === element.x_opencti_score)) {
```
JeremyCloarec, line 2599: I don't fully understand why the condition changed from an 'or' to an 'and'? Won't it change the intended previous behavior to not change the decay even if current score is different?

JeremyCloarec (approved):

Issue #10268, opened 2025-03-18
Delete button bad margin in Relationship edition form


Archidoit:
## Reproducible Steps
Go to a relationship overview and click on the button to edit it.
The relationship edition panel opens.
At the bottom, the 'delete' button is too high.

![Image](https://github.com/user-attachments/assets/99d2c420-6855-423e-af41-f86ba1605b59)

Compared to other edition forms (example here with a fintel template edition form): 

![Image](https://github.com/user-attachments/assets/8a1a1119-3f63-4509-b200-c34fd96dde65)

PR #10331, opened 2025-03-24
[frontend] Delete button bad margin in Relationship edition form (#10268)

SarahBocognano:
### Proposed changes

* Add marginTop to delete Button in relationship edition form
* Get out the delete button from formik to have a float "right" in sighting edition form

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/10268

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/common/stix_core_relationships/StixCoreRelationshipEditionOverview.tsx
+++ b/opencti-platform/opencti-front/src/private/components/common/stix_core_relationships/StixCoreRelationshipEditionOverview.tsx
@@ -54,7 +54,7 @@ const useStyles = makeStyles<Theme>((theme) => ({
     float: 'left',
   },
   button: {
-    float: 'right',
+    float: 'left',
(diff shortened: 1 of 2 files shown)


Archidoit:
To uniformize the delete button should be on the left.

Like for entity edition form and fintel templates edition form : 

image


Whereas in relationship creation form, you put it on the right : 

image

SarahBocognano:
Ok I didn't know we had change the button place. Modifying it

Archidoit:
If I click on Update from this screen
image

I have the delete button : 
image


---

Whereas if I click from this screen (3 dots at the end of the line, then Update) : 
image

I have no Delete button in the Update form
image

maybe add it to uniformize?

SarahBocognano:
There is no delete button from this particular update form because there is directly a "delete" option in the three dots popup. It makes no sense to add a delete button in the update when there is a delete option directly.

Gwendoline-FAVRE-FELIX (approved):

Archidoit (approved):

Issue #10296, opened 2025-03-20
Spacing of left submenu without icons is incorrect


SamuelHassine:
## Description

Spacing of left submenu without icons is incorrect

Spacing of submenu is correct with icons (left padding):

![Image](https://github.com/user-attachments/assets/7cf1b7ff-07ef-417f-bf1b-4bb7f0b3d4fc)

Spacing of submenu is incorrect without icons (missing left padding):

![Image](https://github.com/user-attachments/assets/2b673fef-81ee-4569-9b7d-0df2cd4cbd78)

PR #10332, opened 2025-03-24
[frontend] Spacing of left submenu without icons is incorrect (#10296)

SarahBocognano:
### Proposed changes

* Add a padding left

### Related issues

* [Spacing of left submenu without icons is incorrect
](https://github.com/OpenCTI-Platform/opencti/issues/10296)

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/nav/LeftBar.jsx
+++ b/opencti-platform/opencti-front/src/private/components/nav/LeftBar.jsx
@@ -169,6 +169,14 @@ const useStyles = makeStyles((theme) => createStyles({
     whiteSpace: 'nowrap',
     overflow: 'hidden',
     textOverflow: 'ellipsis',
+    padding: '1px 0 0 0px',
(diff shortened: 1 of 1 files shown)


opencti-platform/opencti-front/src/private/components/nav/LeftBar.jsx
```diff
@@ -389,7 +397,7 @@ const LeftBar = () => {
                     </ListItemIcon>
                   )}
                   <ListItemText
-                    classes={{ primary: (submenu_show_icons && entry.icon) ? classes.menuSubItemText : classes.menuSubItemTextWithoutIcon }}
+                    classes={{ primary: (submenu_show_icons && entry.icon) ? classes.menuSubItemText : classes.menuSubItemTextWithoutIconCollapseOpen }}
```
lndrtrbn: Try using `sx` props instead to avoid duplicating classes

lndrtrbn (approved):

Issue #9952, opened 2025-02-13
[Authorized Member]: Enhance Authorized Member feature to support intersection between organizations and groups


romain-filigran:
## Use case

Enhance the existing Authorized Member feature by introducing the ability to restrict access based on the intersection of Organizations and Groups. This enhancement enables more precise access control by requiring users to be members of both specified Organizations and Groups simultaneously.

PR #10356, opened 2025-03-26
[backend/frontend] Authorized members organizations and groups intersection (#9952)

marieflorescontact:
### Proposed changes

* update authorized members dialog with group restriction
*  add member_id because of relay store
* 

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/9952

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/analyses/reports/ReportCreation.tsx
+++ b/opencti-platform/opencti-front/src/private/components/analyses/reports/ReportCreation.tsx
@@ -177,7 +177,7 @@ export const ReportCreationForm: FunctionComponent<ReportFormProps> = ({
         authorized_members: values.authorized_members.map(({ value, accessRight, groupsRestriction }) => ({
           id: value,
           access_right: accessRight,
-          groups_restriction_ids: groupsRestriction.map((g) => g.value),
+          groups_restriction_ids: groupsRestriction ? groupsRestriction.map((g) => g.value) : [],
         })),
       }),
     };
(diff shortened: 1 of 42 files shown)


opencti-platform/opencti-graphql/config/schema/opencti.graphql
```diff
@@ -1539,6 +1539,7 @@ type MemberGroupRestriction {
 
 type MemberAccess {
   id: ID! # internal_id
+  member_id: String!
```
SouadHadjiat: update comment above. member_id is the internal_id, and id is a unique id "internal_id + index". And id should be of type String

opencti-platform/opencti-front/src/private/components/analyses/reports/ReportCreation.tsx
```diff
@@ -167,9 +174,10 @@ export const ReportCreationForm: FunctionComponent<ReportFormProps> = ({
       externalReferences: values.externalReferences.map(({ value }) => value),
       file: values.file,
       ...(isEnterpriseEdition && canEditAuthorizedMembers && values.authorized_members && {
-        authorized_members: values.authorized_members.map(({ value, accessRight }) => ({
+        authorized_members: values.authorized_members.map(({ value, accessRight, groupsRestriction }) => ({
           id: value,
           access_right: accessRight,
+          groups_restriction_ids: groupsRestriction.map((g) => g.value),
```
SouadHadjiat: needs to be done for other containers creation forms (cases, groupings)

richard-julien (approved):

PR #10362, opened 2025-03-26
FAB_REPLACEMENT: add missing screens and realign styling

labo-flg:
This PR addresses some required changes before we can remove the FAB_REPLACEMENT feature flag:

* remove FAB from Data > ingestion, Data > processing, Data > data sharing
* Update button is now primary CTA (from `outlined` to `contained` button variant), while we wait for _Create relationship_ CTA
* enroll in playbook action bar button is displayed with primary color like any other button

Also, I fixed some styling and simplify some code 

* no need for entityType prefix prop for `CreateEntityControlledDial`; we just need to add the missing translation keys
* align create buttons styles across views (create buttons for Sectors and Decay Rules were too small)
* fix some styling on the way (unrelated to FAB work)


opencti-platform/opencti-front/src/private/components/common/entreprise_edition/EETooltip.tsx
```diff
@@ -104,7 +104,7 @@ const EETooltip = ({
   return (
     <>
       <Tooltip title={title ? t_i18n(title) : undefined}>
-        <span onClick={(e) => {
+        <span onClickCapture={(e) => {
```
labo-flg, line 107: onClick was apparently called after propagation to children, drawer was still opening.

onClickCapture is called before propagation.

opencti-platform/opencti-front/src/private/components/data/connectors/Connector.jsx
```diff
@@ -305,7 +305,9 @@ const ConnectorComponent = ({ connector, relay }) => {
                           minWidth: '6rem',
                         }}
                       >
-                        Reset
+                        <span style={{ zIndex: 2 }}>
```
labo-flg, line 308: fixing z-index problem
image

Kedae:
The update button in Security > User is outlined instead of contained

Kedae:
image
The button in Indicators > Sightings is too small

labo-flg:
> The update button in Security > User is outlined instead of contained

this one ?
image

labo-flg:
> The update button in Security > User is outlined instead of contained

seems ok

CelineSebe:
Tested locally, data and settings ✅ 

FYI, I see a slight difference in the height of the buttons between Data | Settings and the others containers:
create-ingester
create-report-button
The 'create report' label is not perfectly centered (same for the 'update' label) but this has not been done in this PR

CelineSebe (approved):

labo-flg:
> Tested locally, data and settings ✅
> 
> FYI, I see a slight difference in the height of the buttons between Data | Settings and the others containers: create-ingester create-report-button The 'create report' label is not perfectly centered (same for the 'update' label) but this has not been done in this PR

The difference in size is due to List Vs DataTable. We will align everything to DataTable in long term.
As for the centering, looking at it right no :)

PR #10371, opened 2025-03-27
[backend] Fix cache refresh multiple times (#8877)

SouadHadjiat:
### Proposed changes

* cache is refreshed multiple times : once when editing a user, once in notify, then another time in cache manager. For user cache refresh this can lead to perf issues (when entering a draft for ex) => we keep the direct cache refresh in notify

### Related issues

* #8877 
* #9300 

### Checklist


aHenryJard (approved):
Tested locally looks good to me, refresh is correct after confidence changes for example.

PR #10372, opened 2025-03-27
[frontend] fix External reference search field

### LINKED ISSUE #9776: External Reference Search Field Missing Options
## Description

In an entity creation form, the external references field does not show all the options returned by the GraphQL query.

## Environment

1. OS (where OpenCTI server runs): Mac OS Sequoia 15.2
2. OpenCTI version: release-6.5.0 (commit: 377e9e873)
3. OpenCTI client: frontend
4. Other environment details:

## Reproducible Steps

1. Insert test:dev test data
2. Navigate to an entity creation form with the external reference field
3. Enter "pivy" into the external reference field

## Expected Output

Expect to see two options in the autocomplete field related to "pivy"

## Actual Output

No options available

## Screenshots (optional)

The console log in the video below shows the transformation on the query results. After typing "pivy", there should be two options available to the user. This can be recreated with a console log of `newExternalReferences` at line 162 in `ExternalReferencesField.tsx`.

https://github.com/user-attachments/assets/ce886414-561d-45d5-a30a-aa3d63528512

CelineSebe:
### Proposed changes

* optionLength changed

### Related issues

* #9776 

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/common/form/ExternalReferencesField.tsx
+++ b/opencti-platform/opencti-front/src/private/components/common/form/ExternalReferencesField.tsx
@@ -102,11 +102,11 @@ ExternalReferencesFieldProps
     value: string;
     entity?: {
       created?: string;
-      description?: string | null;
-      external_id?: string | null;
+      description: string | null;
+      external_id: string | null;
       id: string;
-      source_name?: string;
-      url?: string | null;
+      source_name: string;
+      url: string | null;
     };
   }[]
   >([]);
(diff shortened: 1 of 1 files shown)


opencti-platform/opencti-front/src/private/components/common/form/ExternalReferencesField.tsx
```diff
@@ -171,6 +171,7 @@ ExternalReferencesFieldProps
         name={name}
         required={required}
         multiple={true}
+        optionLength={200}
```
lndrtrbn: This props is responsible to truncate the label of the option. Changing the default value (40) to 200 is not a fix to the initial bug.

If indeed it appears that the "pivy" case described in the issue is solved with this 200 trick it probable means that the search is made on the truncated label and not the full one, that's the real issue to fix. The search should be made on the complete option, not the truncated version

lndrtrbn (changes requested):

lndrtrbn (approved):

Issue #9760, opened 2025-01-29
Improve the Import Workflow


nino-filigran:
## Use case

Make the import workflow seamless. 
Support for multiple file uploads

nino-filigran:
In the context of import worklfow, the import screen needs to be reworked: https://github.com/OpenCTI-Platform/opencti/issues/10482

PR #10382, opened 2025-03-28
[frontend/backend] Import Workflow Chunk 4 - Add import mode to files dialog

delemaf:
### Proposed changes
* Implement Automatic Import and Step-by-step Import

### Related issues

* Close https://github.com/OpenCTI-Platform/opencti/issues/9760

### Checklist


opencti-platform/opencti-graphql/src/database/file-storage.js
```diff
@@ -436,6 +436,7 @@ export const uploadJobImport = async (context, user, file, entityId, opts = {})
     bypassValidation = false,
     validationMode = defaultValidationMode,
     forceValidation = false
+
```
SouadHadjiat: looks like there is no change here

SouadHadjiat:
I just tried to import files directly, then when I arrive at the last step, the import button is disabled. It gets enabled only if I select a marking or an entity, but these should be optional :

Capture d'écran 2025-03-28 113308

image

opencti-platform/opencti-front/lang/front/fr.json
```diff
@@ -13,6 +13,7 @@
   "2FA state": "État de l'ADF",
   "30 days": "30 jours",
   "A FINTEL export will contain extra information like markings and creation date": "Une exportation FINTEL contiendra des informations supplémentaires telles que les marquages et la date de création",
+  "A guided workflow that streamlines files import, selection of connectors and allows the creation of a workbench or draft for review before final import": "Un flux de travail guidé qui rationalise l'importation des fichiers, la sélection des connecteurs et permet la création d'un plan de travail ou d'une ébauche à réviser avant l'importation finale",
```
JeremyCloarec: Small changes to keep it consistent with other translations: 'espace de travail' instead of 'plan de travail', 'brouillon' instead of 'ébauche'. And 'à valider' instead of 'à réviser'

JeremyCloarec (approved):

PR #10384, opened 2025-03-28
[frontend] fix export of Indicators in Knowledge tab (#10381)

### LINKED ISSUE #10008: Export on multiple indicators from knowledge view throws weird error
## Description

The following error is thrown when exporting data:

`{
  "timestamp": "2025-02-19T00:15:51.484389Z",
  "level": "ERROR",
  "name": "ExportFileCsv",
  "message": "Error in message processing, reporting error to API",
  "exc_info": "Traceback (most recent call last):\n  File \"/usr/local/lib/python3.12/site-packages/pycti/connector/opencti_connector_helper.py\", line 360, in _data_handler\n    message = self.callback(event_data)\n              ^^^^^^^^^^^^^^^^^^^^^^^^^\n  File \"/opt/opencti-connector-export-file-csv/export-file-csv.py\", line 279, in _process_message\n    entities_list = self.helper.api_impersonate.stix2.export_entities_list(\n                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\n  File \"/usr/local/lib/python3.12/site-packages/pycti/utils/opencti_stix2.py\", line 2322, in export_entities_list\n    return do_list(\n           ^^^^^^^^\n  File \"/usr/local/lib/python3.12/site-packages/pycti/entities/opencti_indicator.py\", line 114, in list\n    result = self.opencti.query(\n             ^^^^^^^^^^^^^^^^^^^\n  File \"/usr/local/lib/python3.12/site-packages/pycti/api/opencti_api_client.py\", line 371, in query\n    raise ValueError(value_error)\nValueError: {'name': 'INTERNAL_SERVER_ERROR', 'error_message': 'Cannot return null for non-nullable field Indicator.objectMarking.'}",
  "taskName": null
}`

This error has been observed several times in several different situations (see "Additional information" section) and seems related to the amount of data to export.

## Environment

OCTI 6.5.2

## Reproducible Steps

1. Find an Intrusion set or a Malware with several thousand of Indicators (e.g. [here](https://filigran.octi.filigran.io/dashboard/threats/intrusion_sets/3bd73b78-e1fa-4eb5-ab6c-01ee13e2ac77/knowledge/indicators))
2. Go to the knowledge tab > Indicators
3. Select all and export in "text/csv" + simple export
-> Wait, you should get the error shared in the description 

Image

## Expected Output

I expect to get the export of my data

## Additional information

This error had already been observed in the past at a customer but we had not managed to reproduce it. We initially thought it was due to malformed data in DB (i.e. a kind of corrupted "marking" field), but after investigation, we never identified this. We had managed to solve the problem by exporting the data in several smaller slices which led us to think it was related to the size of the Elastic queries behind the scenes.

When we first observed these errors, it was when exporting data through the API. Now we observe the same thing with the export connectors.

### LINKED ISSUE #10381: export stix of IOC would export all IOCs of the platform
## Description

If you trigger an export stix of IOC, all IOC of the platform would be included. 
## Environment

Testing
## Reproducible Steps

1. Go in a an entity
2. Go to its knowledge tab, indicator section
3. Select IOCs that you would like to download
4. Select stix 

## Expected Output

Only export the selected IOCs. 
## Actual Output

All IOC of the platform exported.

lndrtrbn:
### Proposed changes

* Add a new props in DataTable to be able to pass extra filters that are not saved in local storage. Those extra filter are for example used by the export components when we are in a context of entities inside a container to specify which container it is. 

### Related issues

* #10008 
* #10381 

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/components/dataGrid/DataTableFilters.tsx
+++ b/opencti-platform/opencti-front/src/components/dataGrid/DataTableFilters.tsx
@@ -72,6 +72,7 @@ export const DataTableDisplayFilters = ({
 };
 
 const DataTableFilters = ({
+  additionalFilters,
   availableFilterKeys,
   searchContextFinal,
   availableEntityTypes,
@@ -110,6 +111,22 @@ const DataTableFilters = ({
   const hasFilters = availableFilterKeys && availableFilterKeys.length > 0;
 
   const hasToggleGroup = additionalHeaderButtons || redirectionModeEnabled || !exportDisabled;
+
+  const exportFilterGroups = [];
+  if (isFilterGroupNotEmpty(additionalFilters)) {
+    exportFilterGroups.push(additionalFilters);
+  }
+  if (isFilterGroupNotEmpty(paginationOptions.filters)) {
+    exportFilterGroups.push(paginationOptions.filters);
+  }
+  const exportPaginationOptions = {
+    ...paginationOptions,
+    filters: {
+      mode: 'and',
+      filters: [],
+      filterGroups: exportFilterGroups,
+    },
+  };
 
   return (
     <ExportContext.Provider value={{ selectedIds: Object.keys(selectedElements) }}>
@@ -194,7 +211,7 @@ const DataTableFilters = ({
             <StixDomainObjectsExports
               open={!!openExports}
               handleToggle={helpers.handleToggleExports}
-              paginationOptions={paginationOptions}
+              paginationOptions={exportPaginationOptions}
               exportContext={exportContext}
             />
           </Security>
@@ -205,7 +222,7 @@ const DataTableFilters = ({
             <StixCoreRelationshipsExports
               open={openExports}
               handleToggle={helpers.handleToggleExports}
-              paginationOptions={paginationOptions}
+              paginationOptions={exportPaginationOptions}
               exportContext={exportContext}
             />
           </Security>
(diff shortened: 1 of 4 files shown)


opencti-platform/opencti-front/src/components/dataGrid/DataTableFilters.tsx
```diff
@@ -111,6 +112,22 @@ const DataTableFilters = ({
...
+    exportFilterGroups.push(paginationOptions.filters);
+  }
+  const exportPaginationOptions = {
+    ...paginationOptions,
+    filters: {
+      mode: 'and',
+      filters: [],
+      filterGroups: exportFilterGroups,
+    },
+  };
```
SouadHadjiat, line 129: I don't really understand this part. additionalFilters are filters that are added to the paginationOptionFilters?

aHenryJard (approved):
Tested locally, all good to me both for stix export and csv export.

There is still an issue in some use case, but will be fixed with https://github.com/OpenCTI-Platform/connectors/issues/3617

CelineSebe:
Tested with export file stix, good to me too ✅

opencti-platform/opencti-front/src/components/dataGrid/DataTableFilters.tsx
lndrtrbn: Yes. In the case of Knowledge tab in the page of an entity, when fetching data we are manually adding a filter `inRegardsOf` to be able to fetch only objects correlated to this entity (you can have a look at `EntityStixCoreRelationshipsIndicatorsEntitiesView` line 165).

When using old way of displaying lines it works fine because we are directly given those custom filters to the `ListLines` component. But in case of `DataTable` it's different, the DataTable get the filter itself from local storage. But in local storage we do not have this manual trick using `inRegardsOf` so it was missing when exporting.

The fix now allows to give some extra filters that are not saved in local storage but needed to export correctly depending of the context

Issue #8832, opened 2024-10-30
refactor: rework STIX export connector


marieflorescontact:
## Use case

The stix connector convert into STIX 2.0 with old and messy code. We need to leverage the API capability to export in STIX. 

## If the feature request is approved, would you be willing to submit a PR?

Yes / No (Help can be provided if you need assistance submitting a PR)

PR #10392, opened 2025-03-28
[backend] STIX 2.0 converter: Malware (#8832)

marieflorescontact:
### Proposed changes

* implement STIX 2.0 converter for Malware entity
* tests

### Related issues

* [#8832](https://github.com/OpenCTI-Platform/opencti/issues/8832)

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/database/stix-2-0-converter.ts
+++ b/opencti-platform/opencti-graphql/src/database/stix-2-0-converter.ts
@@ -131,8 +131,6 @@ export const convertMalwareToStix = async (instance: StoreEntity, type: string):
     implementation_languages: instance.implementation_languages,
     capabilities: instance.capabilities,
     operating_system_refs: (instance[INPUT_OPERATING_SYSTEM] ?? []).map((m) => m.standard_id),
-    samples: (instance[INPUT_SAMPLE] ?? []).map((m) => { return { id: m.id, }; }),
     sample_refs: (instance[INPUT_SAMPLE] ?? []).map((m) => m.standard_id),
-    x_opencti_sample_refs_ids: (instance[INPUT_SAMPLE] ?? []).map((m) => m.internal_id),
   };
 };
--- a/opencti-platform/opencti-graphql/src/database/stix-2-0-converter.ts
+++ b/opencti-platform/opencti-graphql/src/database/stix-2-0-converter.ts
@@ -13,7 +13,7 @@ const buildKillChainPhases = (instance: StoreEntity | StoreRelation): Array<SMO.
     const data: SMO.StixInternalKillChainPhase = {
       kill_chain_name: k.kill_chain_name,
       phase_name: k.phase_name,
-      x_opencti_order: k.x_opencti_order, // TODO external_id and hashes ?
+      x_opencti_order: k.x_opencti_order,
     };
     return cleanObject(data);
   });
@@ -25,6 +25,8 @@ const buildExternalReferences = (instance: StoreObject): Array<SMO.StixInternalE
       source_name: e.source_name,
       description: e.description,
       url: e.url,
+      hash: e.hashes, // key is "hash" in client python but should be "hashes"
+      external_id: e.external_id,
     };
     return cleanObject(data);
   });
@@ -51,9 +53,9 @@ const buildStixDomain = (instance: StoreEntity | StoreRelation): S.StixDomainObj
     revoked: instance.revoked,
     confidence: instance.confidence,
     // lang: instance.lang,
-    labels: (instance[INPUT_LABELS] ?? []).map((m) => m.value),
+    labels: (instance[INPUT_LABELS] ?? []).map((m) => m.value), // memo : x_opencti_labels for SCO
     object_marking_refs: (instance[INPUT_MARKINGS] ?? []).map((m) => m.standard_id),
-    created_by_ref: instance[INPUT_CREATED_BY]?.standard_id,
+    created_by_ref: instance[INPUT_CREATED_BY]?.standard_id, // memo: x_opencti_created_by_ref for SCO
     external_references: buildExternalReferences(instance),
   };
 };
@@ -75,5 +77,6 @@ export const convertMalwareToStix2 = (instance: StoreEntity, type: string): SDO.
     capabilities: instance.capabilities,
     operating_system_refs: (instance[INPUT_OPERATING_SYSTEM] ?? []).map((m) => m.standard_id),
     sample_refs: (instance[INPUT_SAMPLE] ?? []).map((m) => m.standard_id),
+    samples: (instance[INPUT_SAMPLE] ?? []).map((m) => { return { id: m.id, }; }),
   };
 };
--- a/opencti-platform/opencti-graphql/src/database/stix-2-0-converter.ts
+++ b/opencti-platform/opencti-graphql/src/database/stix-2-0-converter.ts
@@ -73,7 +73,7 @@ const buildExternalReferences = (instance: StoreObject): Array<SMO.StixInternalE
       source_name: e.source_name,
       description: e.description,
       url: e.url,
-      hash: e.hashes, // key is "hash" in client python but should be "hashes"
+      hash: e.hashes,
       external_id: e.external_id,
     };
     return cleanObject(data);
@@ -108,9 +108,9 @@ const buildStixDomain = async (instance: StoreEntity | StoreRelation): Promise<S
     revoked: instance.revoked,
     confidence: instance.confidence,
     // lang: instance.lang,
-    labels: (instance[INPUT_LABELS] ?? []).map((m) => m.value), // memo : x_opencti_labels for SCO
+    labels: (instance[INPUT_LABELS] ?? []).map((m) => m.value),
     object_marking_refs: (instance[INPUT_MARKINGS] ?? []).map((m) => m.standard_id),
-    created_by_ref: instance[INPUT_CREATED_BY]?.standard_id, // memo: x_opencti_created_by_ref for SCO
+    created_by_ref: instance[INPUT_CREATED_BY]?.standard_id,
     external_references: buildExternalReferences(instance),
   };
 };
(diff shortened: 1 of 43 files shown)


opencti-platform/opencti-graphql/src/database/stix-2-0-converter.ts
```diff
@@ -0,0 +1,83 @@
...
+    aliases: instance.aliases,
+    kill_chain_phases: buildKillChainPhases(instance),
+    first_seen: convertToStixDate(instance.first_seen),
+    last_seen: convertToStixDate(instance.last_seen),
+    architecture_execution_envs: instance.architecture_execution_envs,
+    implementation_languages: instance.implementation_languages,
+    capabilities: instance.capabilities,
+    operating_system_refs: (instance[INPUT_OPERATING_SYSTEM] ?? []).map((m) => m.standard_id),
+    sample_refs: (instance[INPUT_SAMPLE] ?? []).map((m) => m.standard_id),
+    samples: (instance[INPUT_SAMPLE] ?? []).map((m) => { return { id: m.id, }; }),
```
richard-julien: Must be x_opencti_sample_refs_ids i think

Do you take the time to compare your format with the result of the stix file exporter?
Not sure but maybe some elements are missing

opencti-platform/opencti-graphql/src/database/stix-2-0-converter.ts
```diff
@@ -0,0 +1,83 @@
...
+    confidence: instance.confidence,
+    // lang: instance.lang,
+    labels: (instance[INPUT_LABELS] ?? []).map((m) => m.value), // memo : x_opencti_labels for SCO
+    object_marking_refs: (instance[INPUT_MARKINGS] ?? []).map((m) => m.standard_id),
+    created_by_ref: instance[INPUT_CREATED_BY]?.standard_id, // memo: x_opencti_created_by_ref for SCO
+    external_references: buildExternalReferences(instance),
+  };
+};
+
+export const convertMalwareToStix2 = (instance: StoreEntity, type: string): SDO.StixMalware => {
```
richard-julien: convertMalwareToStix could be a better name

richard-julien (changes requested):

richard-julien:
Exposing the new format by adding an option to toStix Graphql method is plan in another PR?

marieflorescontact:
> Exposing the new format by adding an option to toStix Graphql method is plan in another PR?

yes @richard-julien , this will be in chunk 2

opencti-platform/opencti-graphql/src/database/stix-2-0-converter.ts
marieflorescontact: done

opencti-platform/opencti-graphql/src/database/stix-2-0-converter.ts
marieflorescontact: Yes I compare with actual export-stix-file (see test` malware-stix-2-0-converter-test.ts`)

@richard-julien this is actual STIX export, we do  have  `sample_refs`  and I can't find any ` x_opencti_sample_refs_ids`  in the code 🤔: 
image

richard-julien: Thats an error in the python converter. For me sample in 2.0 must no exists like this.
It should be like workflow and follow the pattern rule and so be x_opencti_sample_refs_ids

marieflorescontact: just to be sure @richard-julien  :
`sample_refs` actual attribute should become `x_opencti_sample_refs_ids` ? 
we currently have `samples` (id) and `sample_refs` (standard id). should we keep both except that `sample_refs` would become  `x_opencti_sample_refs_ids`

richard-julien: No, sample_refs must remains the same.

if we can, **samples** must be removed. There is no reason to have this attribute.

opencti-platform/opencti-graphql/src/database/stix-2-0-converter.ts
```diff
@@ -0,0 +1,83 @@
...
+  });
+};
+
+// Builders
+const buildStixObject = (instance: StoreObject): S.StixObject => {
+  return {
+    id: instance.standard_id,
+    x_opencti_id: instance.id,
+    spec_version: '2.0',
+    x_opencti_type: ENTITY_TYPE_MALWARE,
```
JeremyCloarec: Probably a copy/paste error, but this should be instance.entity_type instead of malware

opencti-platform/opencti-graphql/src/database/stix-2-0-converter.ts
```diff
@@ -0,0 +1,83 @@
...
+      url: e.url,
+      hash: e.hashes, // key is "hash" in client python but should be "hashes"
+      external_id: e.external_id,
+    };
+    return cleanObject(data);
+  });
+};
+
+// Builders
+const buildStixObject = (instance: StoreObject): S.StixObject => {
```
JeremyCloarec, line 83: An attribute is missing from current python export: x_opencti_files. It can be added in buildStixObject

opencti-platform/opencti-graphql/src/database/stix-2-0-converter.ts
```diff
@@ -0,0 +1,83 @@
...
+  return {
+    id: instance.standard_id,
+    x_opencti_id: instance.id,
+    spec_version: '2.0',
+    x_opencti_type: ENTITY_TYPE_MALWARE,
+    type: instance.entity_type.toLowerCase(),
```
JeremyCloarec: This works for malware, but some entity types will need to be adapted. For exemple, Threat Actor Group have a type 'threat-actor' when exported. This is because Threat Actor Group doesn't exist in stix 2.0 and need to be changed when exported. 
You can find this type logic at the beginning of the generate_export method in opencti_stix2.py

opencti-platform/opencti-graphql/src/database/stix-2-0-converter.ts
marieflorescontact: no it was just because for now only Malware Entitype is converted, but I will change it, for the future 🚀

marieflorescontact: done

opencti-platform/opencti-graphql/src/database/stix-2-0-converter.ts
marieflorescontact: in current python export a file will be exported like this (in external ref):
Capture d’écran 2025-05-28 165530

wich is what we have:
````
const buildStixDomain = (instance: StoreEntity | StoreRelation): S.StixDomainObject => {
  return {
    ...buildStixObject(instance),
    created: convertToStixDate(instance.created),
    modified: convertToStixDate(instance.modified),
    revoked: instance.revoked,
    confidence: instance.confidence,
    // lang: instance.lang,
    labels: (instance[INPUT_LABELS] ?? []).map((m) => m.value), // memo : x_opencti_labels for SCO
    object_marking_refs: (instance[INPUT_MARKINGS] ?? []).map((m) => m.standard_id),
    created_by_ref: instance[INPUT_CREATED_BY]?.standard_id, // memo: x_opencti_created_by_ref for SCO
    external_references: buildExternalReferences(instance),
  };
};
````

File: opencti-platform/opencti-graphql/src/database/stix-2-0-converter.ts (hunk 5 above)
Comment (marieflorescontact) [reply opencti-platform/opencti-graphql/src/database/stix-2-0-converter.ts]: Yes, it was the same reason as before: for now, it's only for Malware entity. I'll fix that for the future 🚀

Comment (marieflorescontact) [reply opencti-platform/opencti-graphql/src/database/stix-2-0-converter.ts]: done

File: opencti-platform/opencti-graphql/src/database/stix-2-0-converter.ts (hunk 4 above)
Comment (JeremyCloarec) [reply opencti-platform/opencti-graphql/src/database/stix-2-0-converter.ts:83]: They can be referenced in external refs when the Create external reference at upload is configured on the entity, but they can also have the x_opencti_files otherwise. An example of this here: https://testing.octi.staging.filigran.io/dashboard/arsenal/malwares/8e5bb5a8-5fa4-43b4-a04e-954286a5bb69/files

**Review by JeremyCloarec — APPROVED** [review]

File: opencti-platform/opencti-graphql/src/database/stix-2-0-converter.ts (hunk 1 above)
Comment (marieflorescontact) [reply opencti-platform/opencti-graphql/src/database/stix-2-0-converter.ts]: we've decided to keep `samples` to avoid breaking changes.
and `x_opencti_sanple_resf_ids` has been added

Comment (richard-julien) [reply opencti-platform/opencti-graphql/src/database/stix-2-0-converter.ts]: Im not agree with this. Thats needs to be change.

Comment (richard-julien) [reply opencti-platform/opencti-graphql/src/database/stix-2-0-converter.ts]: Please remove samples and also x_opencti_sample_refs_ids (i finally think its not needed). Thanks

File: opencti-platform/opencti-graphql/src/database/stix-2-0-converter.ts [hunk 6]
```diff
@@ -0,0 +1,136 @@
...
+    id: instance.standard_id,
+    x_opencti_id: instance.id,
+    spec_version: '2.0',
+    x_opencti_type: instance.entity_type,
+    type: convertTypeToStix2Type(instance.entity_type),
+    x_opencti_granted_refs: (instance[INPUT_GRANTED_REFS] ?? []).map((m) => m.standard_id),
+    x_opencti_workflow_id: instance.x_opencti_workflow_id,
+    x_opencti_files: await Promise.all((instance.x_opencti_files ?? []).map(async (file: StoreFileWithRefs) => ({
+      name: file.name,
+      data: await getFileContent(file.id, 'base64'),
```
richard-julien: Are you sure about this? I dont think the file data content should be exposed like this

marieflorescontact: In pycti, x_opencti_files always have the data attached. It may be better to remove it, but we won't keep the same behavior as pycti if we do so. Do we prioritize keeping the same behavior to prevent breaking changes, or do we remove it to improve export format?

marieflorescontact: data has been removed @richard-julien

marieflorescontact:
PR status: all comment have been taken into account. waiting for approval or another round of review @richard-julien 🙂

richard-julien (approved):

PR #10428, opened 2025-03-31
[worker] fix starting pingAlive thread once

sbocahu:
### Proposed changes

* Start pingAlive once in the main loop
* Use consume() generator instead of callable + basic_consume to handle graceful stops
* handle graceful stops for sigint (control-c) or sigterm
* improved exception handling
* rejection of faulty messages in AMQP listening mode (if message is not json)

### Checklist

### Further comments

Before this change, a worker would create a pingAlive thread per rabbitmq queue, which is useless & consumes resources.
Also, on exceptions (eg. a message which is not valid json), the consuption would fail and restart, creating a new pingAlive thread. Exceptions during ingestion would then lead to an ever increasing number of these pingAlive threads.
A comment suggested graceful stop was handled but it was incomplete/incorrect.

### DIFF
--- a/opencti-worker/src/worker.py
+++ b/opencti-worker/src/worker.py
@@ -49,44 +49,6 @@ bundles_processing_time_gauge = meter.create_histogram(
 )
 
 
-class PingAlive(threading.Thread):
-    def __init__(self, worker_logger, api) -> None:
-        threading.Thread.__init__(self)
-        self.worker_logger = worker_logger
-        self.api = api
-        self.exit_event = threading.Event()
-        self.in_error = False
-
-    def ping(self) -> None:
-        while not self.exit_event.is_set():
-            try:
-                self.worker_logger.debug("PingAlive running.")
-                self.api.query(
-                    """
-                    query workerPing {
-                      about {
-                        version
-                      }
-                    }
-                  """
-                )
-            except Exception as e:  # pylint: disable=broad-except
-                self.in_error = True
-                self.worker_logger.error(
-                    "Error pinging the API",
-                    {"reason": str(e), "headers": str(self.api.get_request_headers())},
-                )
-            self.exit_event.wait(30)
-        self.worker_logger.info("Thread for PingAlive terminated")
-
-    def run(self) -> None:
-        self.worker_logger.info("Starting PingAlive thread")
-        self.ping()
-
-    def stop(self) -> None:
-        self.exit_event.set()
-
-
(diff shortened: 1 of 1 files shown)


sbocahu:
just got an exception that looks not great. converting to draft

opencti-worker/src/worker.py
```diff
@@ -702,9 +721,10 @@ def start(self) -> None:
                             self.opencti_ssl_verify,
                             self.opencti_json_logging,
                         )
+                        self.consumer_threads[push_queue].name = push_queue
                         self.consumer_threads[push_queue].start()
                     # Listen for webhook message
-                    if connector["config"].get("listen_callback_uri") is not None:
+                    if "connector_user" in connector and connector["config"].get("listen_callback_uri") is not None:
```
sbocahu: did not test much this part but I ran into issues with external import connectors. looking at the code, it seems to be wrong to use listen_callback_uri as a condition to create an ApiConsumer because it is always set up by the connector helper (https://github.com/OpenCTI-Platform/client-python/blob/073609f05b968455e9ecb905612bdedb523c14dd/pycti/connector/opencti_connector_helper.py#L1105)

sbocahu:
> just got an exception that looks not great. converting to draft

the exception was due to a control-C/sigterm & introduced earlier with the addition of the threadpool.
I reworked the Consumer object to handle graceful stop, which was not the case previously (despite the comment saying so).

TODO:
- [x] rework apiConsumer for graceful stop
- [x] check the condition around launching an ApiConsumer
- [x] handle sigkill similarly to sigint

sbocahu:
sibbling PR in pycti: https://github.com/OpenCTI-Platform/client-python/pull/867

opencti-worker/src/worker.py
sbocahu: resolving with https://github.com/OpenCTI-Platform/client-python/pull/867

aHenryJard:
@sbocahu I was about to test locally but there is a conflict with current master, I'm not sure on how to resolve it. Could you rebase and manage the conflit ?

sbocahu:
done

aHenryJard (approved):
Tested locally, all good to me. I used some connectors, and also stopped opencti for a while to see if the worker is correctly taking back work when opencti is up again and it's all good.

richard-julien:
Please dont merge until my review

richard-julien (changes requested):
Need review and change

PR #10443, opened 2025-04-02
[frontend] UX fixes for NLQ (#9751)

Archidoit:
### Proposed changes
If a question is asked in the top search input and the user clicks on the 'ask ai' button, the NLQ query should be launched


labo-flg (approved):
all tested ok, LGTM

PR #10458, opened 2025-04-03
[frontend] Import Workflow - Fixes & import dialog in entities (#9760)

delemaf:
### Proposed changes

* Fix active step
* Fix entity preselection
* Fix no global import capa upload
* Add new import in entity

### Related issues

* Close https://github.com/OpenCTI-Platform/opencti/issues/9760

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/common/files/import_files/ImportFilesStepper.tsx
+++ b/opencti-platform/opencti-front/src/private/components/common/files/import_files/ImportFilesStepper.tsx
@@ -9,7 +9,7 @@ const ImportFilesStepper = () => {
   const hasSelectedFiles = files.length > 0;
 
   return (
-    <Stepper nonLinear activeStep={activeStep} sx={{ marginInline: 10 }}>
+    <Stepper nonLinear activeStep={canSelectImportMode ? activeStep : activeStep - 1} sx={{ marginInline: 10 }}>
       {canSelectImportMode && (<Step key={'import_mode'}>
         <StepButton color="inherit" onClick={() => setActiveStep(0)}>
           {t_i18n('Import mode')}
--- a/opencti-platform/opencti-front/src/private/components/common/files/import_files/ImportFilesStepper.tsx
+++ b/opencti-platform/opencti-front/src/private/components/common/files/import_files/ImportFilesStepper.tsx
@@ -9,6 +9,8 @@ const ImportFilesStepper = () => {
   const hasSelectedFiles = files.length > 0;
 
   return (
+    // If canSelectImportMode is true activeStep is initialised to 1 instead of 0 (we have 2 step instead of 3)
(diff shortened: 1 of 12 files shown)


opencti-platform/opencti-front/src/private/components/nav/TopBar.tsx
```diff
@@ -322,17 +321,13 @@ const TopBarComponent: FunctionComponent<TopBarProps> = ({
...
-                      <IconButton
-                        size="medium"
-                        aria-haspopup="true"
-                        onClick={() => setOpenImportFilesDialog(true)}
-                        color="inherit"
-                      >
-                        <CloudUploadOutlined/>
-                      </IconButton>
-                    </Tooltip>
+                  <Security needs={[KNOWLEDGE_KNASKIMPORT]}>
```
lndrtrbn, line 324: Why not feature flag here ?

opencti-platform/opencti-front/src/private/components/common/files/import_files/ImportFilesStepper.tsx
```diff
@@ -9,7 +9,7 @@ const ImportFilesStepper = () => {
   const hasSelectedFiles = files.length > 0;
 
   return (
-    <Stepper nonLinear activeStep={activeStep} sx={{ marginInline: 10 }}>
+    <Stepper nonLinear activeStep={canSelectImportMode ? activeStep : activeStep - 1} sx={{ marginInline: 10 }}>
```
lndrtrbn, line 14: Can you explain the trick?

opencti-platform/opencti-front/src/private/components/nav/TopBar.tsx
delemaf: Feature flag isImportWorkflowEnabled is already checked

opencti-platform/opencti-front/src/private/components/common/files/import_files/ImportFilesStepper.tsx
delemaf: If canSelectImportMode is true activeStep is initialised to 1 instead of 0 (we have 2 step instead of 3) so we 'lie' to the stepper so the right step is displayed active. I can write a little comment about that

lndrtrbn: Ok, yes a small comment please then

JeremyCloarec:
We still have the 'Navigate to entity' button at the end: shouldn't we only have the close button, when import is done from an entity?
image

lndrtrbn:
[vertical_scroll.webm](https://github.com/user-attachments/assets/c5c1e5eb-5eca-41cf-93ef-d9e1b40e087f)
 Vertical scroll in Dialog

lndrtrbn:
image
Not aligned on Analyses / External References / X

lndrtrbn (approved):

Issue #10466, opened 2025-04-03
In Mass "replace" "Detection" property on indicator, replace to false disabled


alice-debra:
## Description

In Mass "replace" "Detection" property on indicator, replace to false disabled at first. I need to switch it to true and back to false.  (Button update is disabled)

<img width="722" alt="Image" src="https://github.com/user-attachments/assets/8b7a7101-bc17-446e-be1c-300beb3b6815" />

Initialise to false when clicking on detection.

alice-debra:
This problem has been encountered in other fields too.

Issue #8245, opened 2024-09-04
Enhance massive operations on Indicators


romain-filigran:
Ability to update in mass some indicators properties. To take into account: 

- Mass "replace" "Detection" property
- Mass "add/replace/remove" "kill chains" properties
- Mass "add/replace/remove" "indicator types" properties
- Mass "add/replace/remove" "processing status" properties

romain-filigran:
related: https://github.com/OpenCTI-Platform/opencti/issues/7815

blackpint:
Hi there! I just want to vote for this feature.

We rely on "detection" flag to control which indicators should be checked across our logs. Sometimes things go wrong and bunch of indicators became incorrectly "detectable" (or vice versa) and it is such pain in the ass to fix it as there is no way to simply switch this parameter for selected entities. So, this capability is highly desirable :)

nino-filigran:
I added two other tickets related to this one if we add it to (low prio)

PR #10474, opened 2025-04-04
[frontend] In Mass Op to Indicator: Fix toggle display for detection field (#10466)

Gwendoline-FAVRE-FELIX:
### Proposed changes

* Take into account the toggle value when selecting the field.
* Idem in playbook

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/10466
* https://github.com/OpenCTI-Platform/opencti/issues/8245

This PR is based on release/6.6.0 because the bug is the result of a feature on release.

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/data/DataTableToolBar.jsx
+++ b/opencti-platform/opencti-front/src/private/components/data/DataTableToolBar.jsx
@@ -538,8 +538,12 @@ class DataTableToolBar extends Component {
 
     actionsInputs[i] = R.assoc(key, value, actionsInputs[i] || {});
     if (key === 'field') {
-      const values = [];
-      actionsInputs[i] = R.assoc('values', values, actionsInputs[i] || {});
+      if (value === 'x_opencti_detection') {
+        actionsInputs[i] = R.assoc('values', ['false'], actionsInputs[i] || {});
+      } else {
+        const values = [];
+        actionsInputs[i] = R.assoc('values', values, actionsInputs[i] || {});
+      }
       if (
         value === 'object-marking'
         || value === 'object-label'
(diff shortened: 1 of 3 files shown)


opencti-platform/opencti-front/src/private/components/data/DataTableToolBar.jsx
```diff
@@ -538,8 +538,12 @@ class DataTableToolBar extends Component {
 
     actionsInputs[i] = R.assoc(key, value, actionsInputs[i] || {});
     if (key === 'field') {
-      const values = [];
-      actionsInputs[i] = R.assoc('values', values, actionsInputs[i] || {});
+      if (value === 'x_opencti_detection') {
+        actionsInputs[i] = R.assoc('values', ['false'], actionsInputs[i] || {});
```
marieflorescontact, line 542:
```suggestion
        actionsInputs[i] = { …actionsInputs[i], ['values']: ['false'], }
```

marieflorescontact: If time, could be great to remove Ramda

opencti-platform/opencti-front/src/private/components/data/DataTableToolBar.jsx
```diff
@@ -538,8 +538,12 @@ class DataTableToolBar extends Component {
...
+      } else {
+        const values = [];
+        actionsInputs[i] = R.assoc('values', values, actionsInputs[i] || {});
```
marieflorescontact, line 545: same

marieflorescontact (commented):
locally tested ok 👌

opencti-platform/opencti-front/src/private/components/data/DataTableToolBar.jsx
labo-flg: That's true !
However this file is 3k line long and has 51 usages of Ramda. 
Let's do this in another context, we need this bugfix for 6.6.0.

marieflorescontact (approved):

Issue #10467, opened 2025-04-03
In playbooks, I add a type indicator, choose it, then change add by replace. I then get an object that cannot be modified.


alice-debra:
## Description

In playbooks, I add a type indicator, choose it, then change add by replace. I then get an object that cannot be modified.

## Reproducible Steps

- Go to playbook
- Add Block to manipulate knowledge
- add filter
- add
- Type indicator
- choose one
- change add to replace

## Expected Output

Can select or keep the indicator type;

## Actual Output

I get an object that cannot be modified.

## Screenshots (optional)

<img width="1102" alt="Image" src="https://github.com/user-attachments/assets/695549aa-e9fd-4df2-82cd-4c4f25febd18" />

alice-debra:
This problem has been encountered in other fields too.

labo-flg:
FYI: Fixed in release branch 6.6

PR #10478, opened 2025-04-04
[frontend] In playbook, preserve field value when switching action type (#10467)

Gwendoline-FAVRE-FELIX:
### Proposed changes

* Allows you to modify the value while keeping the field defined if you change the type of action.

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/10467


marieflorescontact (approved):
tested ok  👌

Issue #4950, opened 2023-11-22
[Filters]: Relative date-time


nino-filigran:
## Use case

Currently, it is possible filter on "created at" for a specific value only, but not for a dynamic value (ex: - 15d ).
## Current Workaround

None.

## Proposed Solution

Either add the possibility to filter on a dynamic value or ability to provide a date range for the same filter. To be refined. 

## If the feature request is approved, would you be willing to submit a PR?

No

nino-filigran:
Requested again on 04/29: https://filigran-community.slack.com/archives/C06CETVDJ5R/p1714379373471889

PR #10492, opened 2025-04-07
[frontend/backend] fix within date checks (#4950)

Archidoit:
### Proposed changes
- Fix for : within filters with some incorrect values throw error
exemple:
image

### DIFF
--- a/opencti-platform/opencti-front/src/components/filters/RelativeDateInput.tsx
+++ b/opencti-platform/opencti-front/src/components/filters/RelativeDateInput.tsx
@@ -55,9 +55,10 @@ const RelativeDateInput: FunctionComponent<RelativeDateInputProps> = ({
     }
     return undefined;
   };
-  const isValidValue = (value: string, values: string[]) => {
-    const isValidString = RELATIVE_DATE_REGEX.test(value) || isValidDate(value);
-    if (value && values[0] !== values[1] && isValidString) {
+  const isValidValues = (values: string[]) => {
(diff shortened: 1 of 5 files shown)


opencti-platform/opencti-front/src/components/filters/RelativeDateInput.tsx
```diff
@@ -55,11 +55,18 @@ const RelativeDateInput: FunctionComponent<RelativeDateInputProps> = ({
     }
     return undefined;
   };
+  const isValidValue = (value: string, values: string[]) => {
+    const isValidString = RELATIVE_DATE_REGEX.test(value) || isValidDate(value);
+    if (value && value !== '' && values[0] !== values[1] && isValidString) {
```
CelineSebe: CI issue here:
Condition 'value !== ''' is always true at this point because the true branch of the condition 'value' at line 60 has been taken

labo-flg: empty string is false, indeed :)

Archidoit: fixed

CelineSebe:
Do we have to raise an error when the date is later than today?
image

CelineSebe:
Locally tested 👌

Archidoit:
> Do we have to raise an error when the date is later than today? image

--> @CelineSebe no we don't :)

CelineSebe (approved):

PR #10496, opened 2025-04-07
[backend/frontend] align MIME type detection in import workflow with backend behavior

labo-flg:
### Proposed changes

* add new query in file module to `guessMimeType` of a filename, using same mechanics as when a file is uploaded and imported automatically. This uses the platform config, for instance remapping mime type oif `.yar` file to `test/plain+yara`
* call this API when selecting a file in the import wizard, to accurately detect the right connector(s) to suggest

### Related issues

* #9760 

### Checklist


labo-flg:
All tested OK with yara and import-document.
Added unit tests for the new query, this initiate a test suite for `file.js` resolvers (underlying code is still unit tested for most part, but we could improve on that).

delemaf (approved):

PR #10499, opened 2025-04-07
[frontend] fix update note

### LINKED ISSUE #10327: Trying to update a specific Note raises an error
## Description

When I try to update a specific note, I get an error. The Note is this one : https://testing.octi.staging.filigran.io/dashboard/analyses/notes/05dbac7b-4632-4781-81e0-5d3a82b9a66d

This doesn't happen with other notes.

NB: I'm a platform admin and I created this Note through the API

## Environment

OCTI 6.5.8

## Reproducible Steps

1. Go on this Note : https://testing.octi.staging.filigran.io/dashboard/analyses/notes/05dbac7b-4632-4781-81e0-5d3a82b9a66d
2. Click on the button with the three dots (i.e. the burger icon) on top right corner, then "update"
-> Error 

Image

Image

In the front-end platform logs I get this log: `TypeError: e is null`

marieflorescontact:
### Proposed changes
* Note with no abstract raised an error

### Related issues

* closes: https://github.com/OpenCTI-Platform/opencti/issues/10327

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/common/form/TextFieldAskAI.tsx
+++ b/opencti-platform/opencti-front/src/private/components/common/form/TextFieldAskAI.tsx
@@ -239,7 +239,7 @@ const TextFieldAskAI: FunctionComponent<TextFieldAskAiProps> = ({
           <IconButton
             size="small"
             onClick={(event) => ((isEnterpriseEdition && enabled && configured) ? handleOpenMenu(event) : null)}
-            disabled={disabled || currentValue?.length < 10}
+            disabled={disabled || currentValue.length < 10}
             style={{ marginTop: 4, color: theme.palette.ai.main }}
           >
             <AutoAwesomeOutlined fontSize='small' />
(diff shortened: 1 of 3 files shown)


JeremyCloarec (dismissed):

opencti-platform/opencti-front/src/private/components/common/form/TextFieldAskAI.tsx
```diff
@@ -239,7 +239,7 @@ const TextFieldAskAI: FunctionComponent<TextFieldAskAiProps> = ({
           <IconButton
             size="small"
             onClick={(event) => ((isEnterpriseEdition && enabled && configured) ? handleOpenMenu(event) : null)}
-            disabled={disabled || currentValue.length < 10}
+            disabled={disabled || currentValue?.length < 10}
```
lndrtrbn: Not sure of the fix. `currentValue` is defined as a string that cannot be null in props type. So probably that the fix is more to be sure `currentValue` is defined when given to the component.

For most of component called it is already done (for example `ResponseDialog.tsx` l.190) but there are still some calls without this check. Can you try to make this check for all calls and verify that there is no crash?

lndrtrbn (approved):

Issue #10290, opened 2025-03-19
In read only mode, users are not able to mass delete their notifications


SamuelHassine:
## Description

1. Create a user A with default role (access knowledge)
2. Create a trigger
3. Get some notifications

The user can:

* Ack notifications individually
* Delete notification individually
* Ack notification in bulk

But when trying to delete in bulk:

![Image](https://github.com/user-attachments/assets/7e5e1803-d931-44e5-8f57-911ed92210f2)

nino-filigran:
@SamuelHassine which environment/version did you use?
I tried both on testing & prod, and could not reproduce.

Also I noticed that:
- the screen in the background is different: where did you trigger this background task from? 
-  the background task modal is different from yours: 

![Image](https://github.com/user-attachments/assets/7a221203-5477-40ed-ba1d-53fc9ead08f1)

SamuelHassine:
@nino-filigran, all platforms (master branch and dev):

https://dev.octi.staging.filigran.io/dashboard/profile/notifications

1. Select all
2. Click on the trash

![Image](https://github.com/user-attachments/assets/0cda42b2-9c51-47a4-97bf-b24205096c7a)

nino-filigran:
Ok reproduced: important for repro case and the person who will fix it: you need to **select all** (top of the data table) to reproduce. If you do bulk deletion, by only clicking one line & trigger a background task, no error is thrown.

PR #10503, opened 2025-04-07
[frontend] every user should be able to manage its own notifications (#10290)

Archidoit:
### Proposed changes
A user should be able to manage its own notifiations (ie mark them as read / delete them), even if he only has the read only capability

### Related issues
https://github.com/OpenCTI-Platform/opencti/issues/10290


SouadHadjiat (approved):
tested locally ✅

PR #10520, opened 2025-04-09
[frontend] Use new FieldOption interface (#10514)

### LINKED ISSUE #10514: [frontend] Replace usage of ReferenceField.Option
### Actual behavior

Inside the frontend code, in a lot of forms when we need to type the data given to an `AutocompleteField` we are using the interface `Option` from the file `ReferenceField`. This type is as following:

```ts
export interface Option {
  value: string;
  label: string;
  color?: string;
  [key: string]: ReactNode;
  type?: string;
  standard_id?: string;
}
```

### Issue with actual behavior

The line `[key: string]: ReactNode;` is problematic because it is a generic declaration that tells "any other key in the object should have a value that is a ReactNode". And this is quite annoying when we want to extend this interface when we have specific use cases because this `[key: string]` will collide with the extra names we are given the new interface extending `Option`.

This can lead to complicated TypeScript issues to work with.

### Alternative solution

Use an interface that represents only what is expected in an `AutocompleteField` and get rid of the rest. When extra information is needed, just extend this simple interface to fit the needs. This interface has already been created in the file `utils/field.ts` and looks like this:

```ts
export interface AutoCompleteOption {
  value: string;
  label: string;
  color?: string;
  type?: string;
  standard_id?: string;
}
```

The problematic line has been removed. The next step is to use this new interface everywhere we are currently using the one from `ReferenceField`. And if necessary, create specific interfaces extending `AutoCompleteOption` to match particular needs.

lndrtrbn:
### Proposed changes

* Replace all usage of interface Option from file ReferenceField by new interface FieldOption.

### Related issues

* #10514 

### Checklist


marieflorescontact (commented):
missing type or Interface `Option` in `Triggersfield`, `AlertsField` et `NestedMenuButton`

marieflorescontact (approved):

PR #10531, opened 2025-04-09
[frontend/backend] add bulk search in activity logs (#9403)

### LINKED ISSUE #9403: [Activity log] Missing logs when a user perform a bulk search
## Description

When a user does a bulk search, no log is raised (unlike a normal search)

## Environment

6.4.4

## Reproducible Steps

1. In the platform, search using the search bar at the top
2. Then, search using the bulk search
3. Go to "Settings > Activity"
4. Filter on the user with whom you did the search
-> you will see the logs for the normal search (i.e. Advanced search), but not the bulk search

Image

Archidoit:
### Proposed changes
In the extended case (ie the user has an activity trigger on it or is in the activty configuration), there should be an activity log for bulk searches (for the moment, log only for advanced searches)

### Related issues
https://github.com/OpenCTI-Platform/opencti/issues/9403

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/Search.tsx
+++ b/opencti-platform/opencti-front/src/private/components/Search.tsx
@@ -84,59 +84,6 @@ const searchStixCoreObjectsLinesQuery = graphql`
   }
 `;
 
-export const searchStixCoreObjectsLinesSearchQuery = graphql`
-  query SearchStixCoreObjectsLinesSearchQuery(
-    $types: [String]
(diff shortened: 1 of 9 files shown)


aHenryJard:
It seems to be very similar to https://github.com/OpenCTI-Platform/opencti/pull/10539

Archidoit:
Closed because fixed by @richard-julien  with : https://github.com/OpenCTI-Platform/opencti/pull/10539#pullrequestreview-2755791113

PR #10539, opened 2025-04-09
[frontend/backend] Missing logs when a user perform a bulk search (#9403)

### LINKED ISSUE #9403: [Activity log] Missing logs when a user perform a bulk search

richard-julien:
See #9403

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/Search.tsx
+++ b/opencti-platform/opencti-front/src/private/components/Search.tsx
@@ -84,59 +84,6 @@ const searchStixCoreObjectsLinesQuery = graphql`
   }
 `;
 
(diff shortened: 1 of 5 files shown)


aHenryJard:
It seem to be very similar to https://github.com/OpenCTI-Platform/opencti/pull/10531

Archidoit (approved):

Issue #10257, opened 2025-03-17
[Bug] Backend decay test mingles UTC/Z time and "local time" in the date calculations


ParamConstructor:
## Description

The backend test **tests/01-unit/domain/decayRule-test.ts** fails when run locally in a timezone that is UTC-X (i.e. a negative from UTC offset), because the the base time is set to Z/UTC 

![Image](https://github.com/user-attachments/assets/105cbf2e-e40b-4ca7-99bc-a2e0a4fd3e1b)

But the test to manipulate this time use the moment() module and default to manipulation of "local" time, not UTC time. This results in the date expected being the day before where the system running is in a minus UTC offset.

**FAIL** tests/01-unit/domain/decayRule-test.ts > Decay chart data generation > should compute live score serie correctly
**AssertionError**: Base core 100 should be at start date: expected '14/12/2023' to be '15/12/2023' // Object.is equality

Expected: "15/12/2023"
Received: "14/12/2023"

❯ tests/01-unit/domain/decayRule-test.ts:400:104
    398|
    399|     expect(result[0].score).toBe(100);
    400|     expect(moment(result[0].updated_at).format('DD/MM/YYYY'), 'Base core 100 should be at start date').toBe('15/12/2023');
       |                                                                                                       ^
    401|    
    402|     console.log(result[0].score);

The "simple fix" in this case is to just remove the "Z" in the `startDate` initialization, then it will be using "local time" throughout the test.

## Environment

1. OS (where OpenCTI server runs): Mac running UTC-5 timezone
2. OpenCTI version: OpenCTI 6.5.6
3. OpenCTI client: frontend
4. Other environment details: Timezone setting for the test should be ina timezone that is UTC offset to the negative. Like - UTC-5

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. yarn test

## Expected Output

Expect tests should match required dates in the test

## Actual Output

Expect tests are reflecting a date 1 day earlier, due to the negative UTC offset in the current timezone.

## Additional information

None

PR #10547, opened 2025-04-10
[backend] fix the base time set in decayRule test

CelineSebe:
### Proposed changes

* change the base time

### Related issues

* #10257 

### Checklist

### Reproducible Steps

To test this PR, you need to:
* change your time zone: setting for the test should be in a timezone that is UTC offset to the negative. Like - UTC-5
* run the test with 'yarn test decayRule'
*  Results: check that all tests are passed


delemaf (approved):
Tested with UTC-5

SarahBocognano (approved):
working ! ✅

PR #10549, opened 2025-04-10
[backend] fix clean up of denormalized rel if multiple rel of same type exists (#7208)

### LINKED ISSUE #7208: When having 2 relations with the same type between the 2 entities, deleting the first one leads to purge _rel
## Description

When having 2 relations with the same type between the 2 entities, deleting the first one leads to purge _rel

JeremyCloarec:
### Proposed changes

* keep 'duplicates' in computeDeleteElementsImpacts. If a 'duplicate' id is found, it is there for a good reason: it needs to be taken into account, because it means that there are 2 relations of the same type and same target
* change script in elRemoveRelationConnection to only remove first occurence of cleanUpIds

### Related issues

* #7208 
*

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/database/engine.js
+++ b/opencti-platform/opencti-graphql/src/database/engine.js
@@ -3731,7 +3731,7 @@ export const elRemoveRelationConnection = async (context, user, elementsImpact) 
           let source = `if(ctx._source[params.rel_key] != null){
               for (int i=params.cleanupIds.length-1; i>=0; i--) {
                 def cleanupIndex = ctx._source[params.rel_key].indexOf(params.cleanupIds[i]);
-                if(cleanupIndex !== 1){
+                if(cleanupIndex !== -1){
                   ctx._source[params.rel_key].remove(cleanupIndex);
                 }
             }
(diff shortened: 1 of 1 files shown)


opencti-platform/opencti-graphql/src/database/engine.js
```diff
@@ -3728,18 +3728,31 @@ export const elRemoveRelationConnection = async (context, user, elementsImpact)
           const [relationType, relationIndex, side, sideType] = typeAndIndex.split('|');
           const refField = isStixRefRelationship(relationType) && isInferredIndex(relationIndex) ? ID_INFERRED : ID_INTERNAL;
           const rel_key = buildRefRelationKey(relationType, refField);
-          let source = `if (ctx._source['${rel_key}'] != null) ctx._source['${rel_key}'] = ctx._source['${rel_key}'].stream().filter(id -> !params.cleanupIds.contains(id)).collect(Collectors.toList())`;
```
richard-julien, line 3731: Why not just removing the first occurrence?

JeremyCloarec: That's because we don't always want to only remove the first occurence: if we delete the target for example, we will want to delete 2 occurences if we had two relations of the same type to this target (cleanupIds will contain the target ID twice in this case)

SouadHadjiat (approved):

PR #10563, opened 2025-04-10
[backend] Enforce two-factor authentication option prevent connectors to communicate with the platform (#10557)

### LINKED ISSUE #10557: Enforce two-factor authentication option prevent connectors to communicate with the platform
## Description

 Enforce two-factor authentication option prevent connectors to communicate with the platform

## Environment

OpenCTI 6.6.1

richard-julien:
See #10557


aHenryJard (approved):
Tested locally all good to me

PR #10656, opened 2025-04-14
[frontend] Detect changes in CK Editor when in Source mode (#8694)

### LINKED ISSUE #8694: Container: copy/pasting content in a HTML doc in source mode does not detect the text and therefore document cannot be saved
## Description

Not possible to save in Source mode 
## Environment

pre release, demo

## Reproducible Steps

1. Go in report
2. go to content
3. turn to source mode
4. copy a small enough text so that you don't have to scroll since scroll is broken (cf: https://github.com/OpenCTI-Platform/opencti/issues/8687) 
5. paste your text
6. you cannot save

## Expected Output

I can also save in source mode 
## Actual Output

I cannot save 
## Additional information

see video below, starting around 24secs, since at first, I copied a text too long & could not scroll.

## Screenshots (optional)

https://github.com/user-attachments/assets/eca6de02-0a13-4500-a2ee-ae38ff3b2e67

Archidoit:
### Proposed changes
In the Content tab of a Container, in the text area to edit an html file, detect changes when in Source mode to be able to save the changes in Source mode.

### Related issues
https://github.com/OpenCTI-Platform/opencti/issues/8694

### DIFF
--- a/opencti-platform/opencti-front/src/components/CKEditor.tsx
+++ b/opencti-platform/opencti-front/src/components/CKEditor.tsx
@@ -221,6 +221,22 @@ const CKEditor = (props: CKEditorProps<ClassicEditor>) => {
       editor={ClassicEditor}
       config={config}
       {...props}
+      onReady={(editor) => { // to detect changes in source editing mode
+        props.onReady?.(editor); // call parent-provided onReady if defined
+        if (editor.plugins.has('SourceEditing')) {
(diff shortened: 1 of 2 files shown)


SarahBocognano (approved):
The saving works correctly. The UX is not incredible to see that the text is actually saved, but it's not the subject of the bug here

PR #10657, opened 2025-04-14
[backend] remove status update if status is not in entity statuses configuration (#9071)

### LINKED ISSUE #9071: [Playbook] Playbooks apply status to entities that shouldn't have them
## Description

When I change statuses on entities with my playbook, it adds the status on all entities in the manipulated bundle, without taking into account which ones should have it.

## Environment

OCTI 6.2.12

## Reproducible Steps

1. Create a playbook:
2. 1st component: listening on "create" + "Entity type=Note"
3. 2nd component: Apply predefined rule - Resolve container references
4. 3rd component: Manipulate knowledge - "Replace > Status > [an Incident status]", with "Manipulate all elements included in the bundle" turn on
5. 4th component: Send for ingestion
 
Image

6. Go on an Indicent
7. Create a Note

## Expected Output

I want the status to be changed on my Incident **only**

## Actual Output

My status is changed on my Incident, but it is also applied to my Note (because it is included in the bundle manipulated by the playbook). However:

1. The Note has no workflow (i.e. the "processing status" field is normally "disable"), so should not have a status.
2. The status applied is an Incident status, it should not apply to entity types other than Incident.

Image

Image

JeremyCloarec:
### Proposed changes

* remove status update if status is not in entity statuses configuration
*

### Related issues

* #9071 
*

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/database/middleware.js
+++ b/opencti-platform/opencti-graphql/src/database/middleware.js
@@ -106,7 +106,6 @@ import {
   INPUT_CREATED_BY,
   INPUT_LABELS,
   INPUT_MARKINGS,
-  INPUT_WORKFLOW_STATUS,
   INTERNAL_IDS_ALIASES,
   INTERNAL_PREFIX,
   REL_INDEX_PREFIX,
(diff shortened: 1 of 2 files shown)


SarahBocognano (approved):

opencti-platform/opencti-graphql/src/database/middleware.js
```diff
@@ -2004,6 +2005,17 @@ export const updateAttributeMetaResolved = async (context, user, initial, inputs
...
   }
+  // Validate workflow status value
+  const statusInput = updates.find((update) => update.key === INPUT_WORKFLOW_STATUS);
+  if (statusInput && statusInput.value?.length > 0) {
+    const statusInputValue = statusInput.value[0];
+    const platformStatuses = await getEntitiesListFromCache(context, user, ENTITY_TYPE_STATUS);
+    const entitySettings = platformStatuses.filter((status) => status.type === initial.entity_type);
+    if (entitySettings?.length === 0 || !entitySettings.some((entityStatus) => entityStatus.internal_id === statusInputValue)) {
+      logApp.warn('Entity status to modify was not found in entity status configuration', { statusInputValue, entitySettings });
+      updates = updates.filter((update) => update.key !== INPUT_WORKFLOW_STATUS);
```
SouadHadjiat: to move in prepareAttributesForUpdate as discussed

PR #10658, opened 2025-04-14
[backend] Error message when using NLQ with bad AI configuration (#10652)

### LINKED ISSUE #10652: NLQ AI should give a well-handled error when no AI endpoint is configured
## Use case

As an analyst, I see an API error when trying to use NLQ when I have no AI API configured.

## Current Workaround

None; error appearsa

## Proposed Solution

Give an error as for AI Insights. ie.
"Enable AI powered platform
To use AI, please enable it in the configuration of your platform."

Archidoit:
### Proposed changes
When using NLQ with a bad AI configuration, display a specific error message : 'Incorrect AI configuration for NLQ', instead of the generic NLQ error : 'Error when calling the NLQ model'

### Related issues
https://github.com/OpenCTI-Platform/opencti/issues/10652

### DIFF
--- a/opencti-platform/opencti-graphql/src/modules/ai/ai-domain.ts
+++ b/opencti-platform/opencti-graphql/src/modules/ai/ai-domain.ts
@@ -14,9 +14,9 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 */
 
 import { callWithTimeout } from '@opentelemetry/sdk-metrics/build/esnext/utils';
-import { GraphQLError } from 'graphql/index';
+import { TimeoutError } from '@opentelemetry/sdk-metrics';
 import { logApp } from '../../config/conf';
-import { FunctionalError, UnknownError, UNSUPPORTED_ERROR, UnsupportedError } from '../../config/errors';
+import { FunctionalError, UnknownError } from '../../config/errors';
 import { queryAi, queryNLQAi } from '../../database/ai-llm';
 import { elSearchFiles } from '../../database/file-search';
 import { storeLoadById } from '../../database/middleware-loader';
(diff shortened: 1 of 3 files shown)


opencti-platform/opencti-graphql/src/modules/ai/ai-domain.ts
```diff
@@ -421,7 +422,11 @@ export const generateNLQresponse = async (context: AuthContext, user: AuthUser,
   try {
     rawResponse = await callWithTimeout(queryNLQAi(promptValue), NLQ_TIMEOUT);
   } catch (error) {
-    throw UnknownError('Error when calling the NLQ model', { error, promptValue });
+    if (error instanceof GraphQLError && error.extensions.code === UNSUPPORTED_ERROR) {
+      throw UnsupportedError('Incorrect AI configuration for NLQ', { error, promptValue });
+    } else {
+      throw UnknownError('Error when calling the NLQ model', { error, promptValue });
+    }
```
JeremyCloarec: Since queryNLQAi already handles it's own error throwing, we only need to catch the TimeoutError thrown from callWithTimeout and transform it into an UnknownError

Archidoit: Not exactly since the error coming from queryNLQAi can be from several reasons... we only want to catch the case when the error cause is a bad AI configuration

aHenryJard (changes requested):
I see that in default file the ai.enabled is true ; ai token and AI url are empty.

=> this could be catch even before calling AI, if there is no API endpoint I don't see how it can work at all.

Also I tested locally, for a wrong token and a wrong AI url I still get the message "Error when calling the NLQ model"

I wonder if we shouldn't consider AI as disabled when at least endpoint is empty.

opencti-platform/opencti-graphql/src/modules/ai/ai-domain.ts
JeremyCloarec: Yes, but queryNLQAi can already only raise two errors: an UnsupportedError or an UnknownError (outside of potential errors from redis). The condition you added `if (error instanceof GraphQLError && error.extensions.code === UNSUPPORTED_ERROR)` will catch the already thrown UnsupportedError thrown by queryNLQAi, it's redundant. It's equivalent to this code:
```
try {
   throw UnsupportedError();
}
catch (error) {
   if (error instanceof GraphQLError && error.extensions.code === UNSUPPORTED_ERROR) {
      throw UnsupportedErro();
   }
}
```

JeremyCloarec:
Changes are good, but it is missing some front end changes I think: the target is to have this dialog open when clicking on AskAI, like it is done on AI Insights. The userAI hook's fullyActive boolean can be used to check if AI is fully configured
image

JeremyCloarec (approved):

aHenryJard (approved):

Issue #10614, opened 2025-04-12
When left menu opened, the active marker is not working in drafts section


SamuelHassine:
## Description

Correct with left menu closed:
![Image](https://github.com/user-attachments/assets/81e12ce8-61d0-400f-bdc7-dcc7a2a7323a)

Not correct with left menu opened:
![Image](https://github.com/user-attachments/assets/01920e3f-626d-42da-8df2-25a8f931d8de)

Same in investigation:

![Image](https://github.com/user-attachments/assets/c8d5c2a2-6bd3-4592-85fa-25ecaf348dac)

romain-filigran:
Same for Trash menu

PR #10669, opened 2025-04-14
[frontend] When left menu opened, the active marker is not working in drafts section (#10614)

SarahBocognano:
### Proposed changes

* Add icon in draft / investigation & trash

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/10614

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/nav/LeftBar.jsx
+++ b/opencti-platform/opencti-front/src/private/components/nav/LeftBar.jsx
@@ -544,6 +544,7 @@ const LeftBar = () => {
                   primary={t_i18n('Investigations')}
                 />
                 )}
+                {navOpen && (selectedMenu.includes('investigations') ? <ExpandLessOutlined /> : <ExpandMoreOutlined />)}
               </MenuItem>
             </StyledTooltip>
             )}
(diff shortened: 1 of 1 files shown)


opencti-platform/opencti-front/src/private/components/nav/LeftBar.jsx
```diff
@@ -544,6 +544,7 @@ const LeftBar = () => {
                   primary={t_i18n('Investigations')}
                 />
                 )}
+                {navOpen && (selectedMenu.includes('investigations') ? <ExpandLessOutlined /> : <ExpandMoreOutlined />)}
```
SouadHadjiat: I don't understand this fix : the issue is that the menu does not appear as selected, we don't want to add an arrow to it. We want these menus (investigations, trash, draft) to appear selected as it does with the home menu.
image

SarahBocognano: Sorry. I didn't understood the issue at all. Rolling back and fixing it properly lol

Gwendoline-FAVRE-FELIX (approved):

SouadHadjiat (approved):

PR #10689, opened 2025-04-15
Remove Floating Action Button Replacement Feature Flag

Bonsai8863:
### Proposed changes

* Removed FAB_REPLACEMENT feature flag platform-wide.

### Checklist


opencti-platform/opencti-front/src/private/components/analyses/groupings/GroupingPopover.jsx
```diff
@@ -33,10 +32,8 @@ const GroupingPopover = ({ id }) => {
   const [anchorEl, setAnchorEl] = useState(null);
   const [displayDelete, setDisplayDelete] = useState(false);
   const [displayEdit, setDisplayEdit] = useState(false);
-  const [displayEnrichment, setDisplayEnrichment] = useState(false);
+  const [_, setDisplayEnrichment] = useState(false);
```
Archidoit: setDisplayEnrichment is useless if displayEnrichment is no more used ?

opencti-platform/opencti-front/src/private/components/analyses/groupings/GroupingPopover.jsx
```diff
@@ -77,69 +71,67 @@ const GroupingPopover = ({ id }) => {
...
-                  open={displayEdit}
-                />
-              );
-            }
-            return <div />;
-          }}
-        />
-      </>
-    );
+  return (
```
Archidoit: your code was : isFABReplaced
    ? (<></>)
    : (

So it seems we should rather have: return (<></>);

opencti-platform/opencti-front/src/private/components/analyses/malware_analyses/MalwareAnalysisPopover.tsx
```diff
@@ -77,57 +74,55 @@ const MalwareAnalysisPopover = ({ id }: { id: string }) => {
     });
   };
 
-  return isFABReplaced
-    ? (<></>)
-    : (
+  return (
```
Archidoit: same for all the containers

opencti-platform/opencti-front/src/private/components/analyses/notes/NotePopover.tsx
```diff
@@ -118,89 +112,87 @@ const NotePopover: FunctionComponent<NotePopoverProps> = ({
...
-          </IconButton>
-        ) : (
-          <ToggleButton
-            value="popover"
-            size="small"
-            onClick={handleOpen}
-          >
-            <MoreVert fontSize="small" color="primary" />
-          </ToggleButton>
+  return (
```
Archidoit, line 101: same, should return <></> --> these files can be deleted

Archidoit (changes requested):

opencti-platform/opencti-front/src/private/components/analyses/notes/NotePopover.tsx
Bonsai8863: I removed most of the other popovers. Note and External Reference are still used in-line for components. For NotePopover, I removed the option for enrichment since that was what was using FAB for some reason and there is a dedicated enrichment button next to the popover now.

opencti-platform/opencti-front/src/private/components/common/containers/ContainerAddStixCoreObjectsInGraph.tsx
```diff
@@ -1,32 +1,15 @@
...
- * FAB_REPLACEMENT feature flag is enabled, returns a ContainerAddStixCoreObjectsInLine
- * component with a blue plus controlled dial.
- *
- * Since many of the components that call ContainerAddStixCoreObjects are
- * class components, we cannot use the useHelper hook. So, I've abstracted the
- * switching logic into this function component until the FABs have been
- * removed from the system entirely and the FAB_REPLACEMENT feature flag is
- * redundant.
- */
 const ContainerAddStixCoreObjectsInGraph = (
```
Archidoit: this component is useless, as you can directly call ContainerAddStixCoreObjectsInLine instead of it

opencti-platform/opencti-front/src/private/components/settings/activity/alerting/AlertCreation.tsx
```diff
@@ -54,10 +30,7 @@ const AlertCreation: FunctionComponent<TriggerCreationProps> = ({
 }) => {
   const { t_i18n } = useFormatter();
   const theme = useTheme<Theme>();
-  const { isFeatureEnable } = useHelper();
-  const isFABReplaced = isFeatureEnable('FAB_REPLACEMENT');
-  const classes = useStyles();
-  const [openSpeedDial, setOpenSpeedDial] = useState(false);
+  const [_, setOpenSpeedDial] = useState(false);
```
Archidoit: useless

Archidoit:
Different design (button margin size) between
- entities creation
image

image


- activity digest/trigger and workbench creations (buttons are smaller)
image

image

Archidoit:
The speed dial has not been replaced for triggers and digest 
image


but it has been for activity digest and activity triggers
image

Archidoit:
Speed dial is still present for Vocabularies 
image

Archidoit:
When I open the 'create live trigger' tab from the Triggers & Digest section, the panel opens to 3/4 of the page : 
image


Whereas before and in from the Activity tab, it opens smaller:
image

Archidoit (approved):

PR #10694, opened 2025-04-16
[frontend] Display all filters inline with a single tooltip(#9806)

### LINKED ISSUE #9806: feat(ui): display all filters inline with a tooltip
## Description

We need to see all filters inline in Data Sharing > Live Streams, TAXII Collections, or CSV Feeds

## Environment

4. Other environment details:

## Expected Output

Example for the markings:
Image

## Actual Output

Image

CelineSebe:
### Proposed changes

* one tooltip
* refacto filterIconButtonContainer

### Related issues

*  #9806 

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/components/FilterIconButtonContainer.tsx
+++ b/opencti-platform/opencti-front/src/components/FilterIconButtonContainer.tsx
@@ -271,7 +271,7 @@ FilterIconButtonContainerProps
     }
   }
   const allFilterKeys = displayedFilters.map((filter) => filter.key);
-  const allRegardingOf = allFilterKeys.every((key) => key === 'regardingOf');
+  const allRegardingOfFilterKeys = allFilterKeys.every((key) => key === 'regardingOf');
   const buildKeyLabel = (filterOperator: string, filterLabel: string, values: string[]) => {
     const isOperatorDisplayed = filterOperatorsWithIcon.includes(filterOperator);
     return (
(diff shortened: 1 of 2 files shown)


Archidoit:
The tooltip doesn't take the global mode into account

image

--> same tooltip for these two filters whereas they are different (see global mode: and/or)

Archidoit:
The tooltip is poping when I modify filters in entities list : 
image

opencti-platform/opencti-front/src/components/FilterIconButtonContainer.tsx
```diff
@@ -267,157 +267,167 @@ FilterIconButtonContainerProps
       };
     }
   }
-  return (
-    <Box sx={boxStyle}>
+  const allFilterKeys = displayedFilters.map((filter) => filter.key);
+  const allRegardingOf = allFilterKeys.every((key) => key === 'regardingOf');
```
Archidoit:
```suggestion
  const allRegardingOfFilerKeys = allFilterKeys.every((key) => key === 'regardingOf');
```

Archidoit:
In trigger creation form, the tooltip is displayed at the middle of the form instead of near the filter : 
image


And it could block the filter edition: 
image


--> Same in an entity Knowledge tab : 
image

Archidoit (changes requested):

Archidoit:
When I edit a filter, there should be no tooltip.
Example, in this case, it avoids changing the operator: 
image


image

Issue #9811, opened 2025-02-03
Introduce a score field on Organization entities


romain-filigran:
## Use case

Introduce a score field to Organization entities to enable supply chain and organizational  risk assessment . This score will use to track potential risks associated with organizations.
The score value must be contains between 0 and 100.

PR #10703, opened 2025-04-16
[backend/frontend] Add score to Organizations (#9811)

marieflorescontact:
### Proposed changes

* Add score attribute for Organization
*  Add score to graphQL schema
*  Add score in STIX extensions
* Add resolver test
* Add score in creation/update UI form
* Add score in Organization details screen

### Related issues

* closes: https://github.com/OpenCTI-Platform/opencti/issues/9811
* Client python PR: https://github.com/OpenCTI-Platform/client-python/pull/884

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/entities/organizations/OrganizationCreation.tsx
+++ b/opencti-platform/opencti-front/src/private/components/entities/organizations/OrganizationCreation.tsx
@@ -59,7 +59,7 @@ interface OrganizationAddInput {
   confidence: number | null
   x_opencti_reliability: string | undefined
   x_opencti_organization_type: string | undefined
-  x_opencti_score: number | null
+  x_opencti_score: string | null
   createdBy: FieldOption | null
   objectMarking: FieldOption[]
   objectLabel: FieldOption[]
@@ -147,7 +147,7 @@ export const OrganizationCreationForm: FunctionComponent<OrganizationFormProps> 
         description: values.description,
         x_opencti_reliability: values.x_opencti_reliability,
         x_opencti_organization_type: values.x_opencti_organization_type,
-        x_opencti_score: parseInt(String(values.x_opencti_score), 10),
+        x_opencti_score: values.x_opencti_score ? parseInt(values.x_opencti_score, 10) : null,
         createdBy: values.createdBy?.value,
         confidence: parseInt(String(values.confidence), 10),
         objectMarking: values.objectMarking.map((v) => v.value),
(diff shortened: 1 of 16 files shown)


opencti-platform/opencti-front/src/private/components/entities/organizations/OrganizationCreation.tsx
```diff
@@ -100,6 +102,11 @@ export const OrganizationCreationForm: FunctionComponent<OrganizationFormProps>
       .nullable(),
     x_opencti_reliability: Yup.string()
       .nullable(),
+    x_opencti_score: Yup.number()
+      .nullable()
+      .min(0, t_i18n('The value must be greater than or equal to 0'))
+      .max(100, t_i18n('The value must be less than or equal to 100'))
+      .test('Digits only', t_i18n('The value must contain digits only'), (val) => /^\d+$/.test(String(val))),
```
SouadHadjiat: do we really need this last test 'Digits only' ? isn't it already tested by Yup.number() ?

marieflorescontact: this is asked by product to avoid having value such as "4+4"  wich is not taken into acount anyway but does not have error message

opencti-platform/opencti-front/src/private/components/entities/organizations/OrganizationCreation.tsx
```diff
@@ -141,6 +148,7 @@ export const OrganizationCreationForm: FunctionComponent<OrganizationFormProps>
         description: values.description,
         x_opencti_reliability: values.x_opencti_reliability,
         x_opencti_organization_type: values.x_opencti_organization_type,
+        x_opencti_score: parseInt(String(values.x_opencti_score), 10),
```
delemaf: It's strange to use `String()` here
```suggestion
        x_opencti_score: parseInt(values.x_opencti_score, 10),
```

marieflorescontact: as `values.x_opencti_score` could be null or undefined, it needs to be transformed as string , see confidence line 152

opencti-platform/opencti-front/src/private/components/entities/organizations/OrganizationCreation.tsx
marieflorescontact: I removed the .test()

opencti-platform/opencti-front/src/private/components/entities/organizations/OrganizationCreation.tsx
delemaf: image
Same issue will occur, either with String and parseInt maybe the best workaround would be to initialize (line 189) it  to 50 (like we do with Indicator creation score)

marieflorescontact: we don't want score to be initialized as Indicator or Observable. we want it to be null when we open the creationForm.

marieflorescontact: if I remove the String() I get this type error:
image

delemaf: Patched together :+1:

delemaf (approved):
Pair tested !

PR #10711, opened 2025-04-17
[frontend]  Sorting should not trigger full reload (#10252)

### LINKED ISSUE #10252: Sorting should not trigger full reload / disappearing of the table header but just showing skeletons
## Description

1. Go to any list
2. Then just select a new ordering

During the loading, everything is disappearing. Sorting in table should not display a full screen loader but skeletons without disappearing the header of the table (this was the previous behavior).

https://github.com/user-attachments/assets/74fa53cd-935e-4de9-a817-c21ce785a866

+ the fact that there is, during this loading, a useless scrollbar on the right...

ValentinBouzinFiligran:
### Proposed changes
* useDatatable removed from props and called on datatable body
* data from datatableWithoutFragment passed on props

### Related issues

* #10252 

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/components/dataGrid/components/DataTableBody.tsx
+++ b/opencti-platform/opencti-front/src/components/dataGrid/components/DataTableBody.tsx
@@ -39,14 +39,11 @@ const DataTableBody = ({
   } = useDataTableContext();
 
   const {
-    data: paginatedData,
+    data: queryData,
     isLoading,
     loadMore,
     hasMore,
-  } = useDataTable(dataQueryArgs);
-
-  // data is from datatableWithoutFragment
-  const queryData = data ?? paginatedData;
+  } = data ? { data } : useDataTable(dataQueryArgs); // data is from datatableWithoutFragment
 
   const resolvedData = useMemo(() => {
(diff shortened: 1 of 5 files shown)


opencti-platform/opencti-front/src/components/dataGrid/components/DataTableBody.tsx
```diff
@@ -26,21 +27,27 @@ const DataTableBody = ({
...
+  const {
+    data: paginatedData,
+    isLoading,
+    loadMore,
+    hasMore,
+  } = useDataTable(dataQueryArgs);
+
+  // data is from datatableWithoutFragment
+  const queryData = data ?? paginatedData;
+
```
delemaf: In case we only use data (DataTableWithoutFragment) calling useDataTable doesn't work. 
```suggestion
  const {
    data: queryData,
    isLoading,
    loadMore,
    hasMore,
  } = data ? { data } : useDataTable(dataQueryArgs); // data is from datatableWithoutFragment
```

delemaf (changes requested):

delemaf (approved):

PR #10732, opened 2025-04-18
Do not start child-lock.manager before every backend test

aHenryJard:
### Proposed changes

*
*

### Related issues

*
*

### Checklist


aHenryJard:
no it's not good either.... it's reset at some point. Need to dig deeper.

PR #10763, opened 2025-04-21
[frontend] Scrollable User History

ParamConstructor:
This is a PR that was originally submitted under - https://github.com/OpenCTI-Platform/opencti/pull/10236. The original was auto-closed when `release/6.6.0` branch was recently closed and the new `release/current` approach was adopted.

This submission has been rebased to `release/current` and reopened.

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/settings/users/User.tsx
+++ b/opencti-platform/opencti-front/src/private/components/settings/users/User.tsx
@@ -609,7 +609,7 @@ const User: FunctionComponent<UserProps> = ({ data, refetch }) => {
                 </Security>
                 <div className="clearfix" />
                 <FieldOrEmpty source={orderedSessions}>
-                  <List style={{ marginTop: -2, width: '100%', maxHeight: 400, overflowY: 'auto'}}>
+                  <List style={{ marginTop: -2, width: '100%', maxHeight: 400, overflowY: 'auto' }}>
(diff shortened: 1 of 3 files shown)


ParamConstructor:
Copying @labo-flg 's last review comment from base PR that was closed:

@ParamConstructor closing release/6.6.0 closed this PR automatically.
You can rebase on release/current and reopen this PR to target release/current.

As for the review, I had a look.
All good for the fix on the scrollable area.
As for the history, I'm not sure it's the right approach.
Today this query is apparently very slow, even for 10 entries (which is a problem...) Pushing to 500 can only get it worse.
The query is also repeated every few seconds while we're on this screen, to update latest history.

Ideally, I would go for this approach:

- Keep the last 10 actions in the main view. This panel is just to see what's happened recently, for a quick view.
- add a button on top of the Paper to "Browse user's history", that would open a drawer with a paginated data table to query the full history, with search and filters (time range for instance, we have a new operator 'within' in 6.6 for that).

Now I understand this is significantly more work, we can discuss it separately and already merge the fix on the scrollable area first.

ParamConstructor:
This is being reworked to reflect a Tab approach recommend by the Filigran team. A new PR will be opened with this coded approach. The overflow bug in this PR was merged into master via a separate PR (https://github.com/OpenCTI-Platform/opencti/pull/10787)

PR #10802, opened 2025-04-23
[backend] Update cache after renew api key (10762)

### LINKED ISSUE #10762: API Key Renewal Does Not Persist After Refresh
## Description
The UI generates a new key when a user attempts to renew an API key via the OpenCTI interface. However, upon refreshing the page or navigating away and back, the API key reverts to its previous value. 

## Environment
SaaS

## Reproducible Steps
1. Navigate to the user profile.
2. Click the “Renew API Key” to generate a new key.
3. Observe the newly displayed API key.
4. Refresh the browser or navigate away and return to the same page.

## Expected Output
The newly generated API key should be persisted and displayed after refreshing the page or navigating back. The old key should be invalidated and no longer visible or usable.

## Actual Output
After refreshing the page, the system displays the old API key instead of the newly generated one.

frapuks:
### Proposed changes

* Add notify function to update cache ater renew api kay

### Related issues

* close #10762 

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/domain/user.js
+++ b/opencti-platform/opencti-graphql/src/domain/user.js
@@ -1425,8 +1425,7 @@ export const userRenewToken = async (context, user, userId) => {
     throw FunctionalError(`Cannot renew token, ${userId} user cannot be found.`);
   }
   const patch = { api_token: uuid() };
-  await patchAttribute(context, user, userId, ENTITY_TYPE_USER, patch);
-  const result = await storeLoadById(context, user, userId, ENTITY_TYPE_USER);
+  const { element } = await patchAttribute(context, user, userId, ENTITY_TYPE_USER, patch);
 
   const actionEmail = ENABLED_DEMO_MODE ? REDACTED_USER.user_email : userData.user_email;
   await publishUserAction({
@@ -1438,7 +1437,7 @@ export const userRenewToken = async (context, user, userId) => {
     context_data: { id: userId, entity_type: ENTITY_TYPE_USER }
   });
 
-  return notify(BUS_TOPICS[ENTITY_TYPE_USER].EDIT_TOPIC, result, user);
+  return notify(BUS_TOPICS[ENTITY_TYPE_USER].EDIT_TOPIC, element, user);
 };
 
 /**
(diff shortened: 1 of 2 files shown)


Gwendoline-FAVRE-FELIX (approved):
It works

JeremyCloarec (changes requested):
Will need to be looked at further. 
Fetching the me user from the cache and not from elastic is intended for performance reasons, and we do not want to change that. We need to find a way to fix the refresh of the user in the cache instead

SouadHadjiat:
> Will need to be looked at further. Fetching the me user from the cache and not from elastic is intended for performance reasons, and we do not want to change that. We need to find a way to fix the refresh of the user in the cache instead

yes I agree, the refresh should have happened, we need to understand why the cache was not refreshed

frapuks:
ready for 2nd review

opencti-platform/opencti-graphql/src/resolvers/user.js
```diff
@@ -54,7 +54,7 @@ import {
   userOrganizationsPaginated,
   userOrganizationsPaginatedWithoutInferences,
   userRenewToken,
-  userWithOrigin
+  userWithOrigin,
```
JeremyCloarec: Small detail, but can you remove the comma to prevent a change here?

frapuks: corrected

JeremyCloarec (approved):

opencti-platform/opencti-graphql/src/domain/user.js
```diff
@@ -1426,7 +1426,7 @@ export const userRenewToken = async (context, user, userId) => {
   }
   const patch = { api_token: uuid() };
   await patchAttribute(context, user, userId, ENTITY_TYPE_USER, patch);
-  const result = storeLoadById(context, user, userId, ENTITY_TYPE_USER);
+  const result = await storeLoadById(context, user, userId, ENTITY_TYPE_USER);
```
SouadHadjiat: we could avoid this call by returning directly element from patchAttribute

JeremyCloarec: I was thinking that the call to storeLoadById was really intended for something? (even though I can't see any reason myself). But it's been like that for at least 3 years

frapuks: Updated
- Remove storeLoadById
- Use the result of patchAttribute

SouadHadjiat (approved):

Issue #10700, opened 2025-04-16
When only one export connector is available for the desired format, it should be directly selected


SamuelHassine:
## Description

When only one export connector is available for the desired format, it should be directly selected.

It was the case before (at least a few month ago).

https://github.com/user-attachments/assets/e0f6aa02-b6e3-41f1-a10a-77fb38c1e4b5

PR #10806, opened 2025-04-23
[frontend] When only one export connector is available for the desired format, it should be directly selected (#10700)

Gwendoline-FAVRE-FELIX:
### Proposed changes

* pre-select connector if only one matches selected format

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/10700


SarahBocognano:
https://github.com/user-attachments/assets/7155d7ae-2ec3-4119-8fb9-df591c71334a

It works for some, but not for the pdf one :(

Gwendoline-FAVRE-FELIX:
That's because you have several choices available. The aim is that the pre-selection is only made if there is only one choice available.

SarahBocognano:
Oh ok great then ! It works perfectly :)

SarahBocognano (approved):

PR #10807, opened 2025-04-23
[frontend] Removing popover should not show cascade delete in investigations (#10793)

### LINKED ISSUE #10793: In investigation, popover for removing is showing checkbox to delete element if no container contains it
## Description

In investigation, popover for removing is showing checkbox to delete element if no container contains it => investigations are not containers.

Image

Archidoit:
### Proposed changes
When removing an entity from an investigation graph, the 'cascade delete' part should not be displayed

### Related issues
https://github.com/OpenCTI-Platform/opencti/issues/10793


SarahBocognano (approved):
Tested ✅

PR #10809, opened 2025-04-23
[frontend] Make deletion calls in knowledge graphs sequential (#10784)

### LINKED ISSUE #10784: Removing more than 10 entities at the same time from a container is not working anymore
## Description

Removing more than 10 entities at the same time from a container is not working anymore.

1. Go to a report that contains 100 indicators
2. Select all indicators with select by type
3. Remove them

This removes only a few, then timing out.

This breaking change is related to graph refactoring:

Previously: R.forEach((n) / .then(async (data) => { 

==> Meaning we are doing this in sequence

Now: [...selectedNodes, ...selectedLinks].forEach(({ id }) / .toPromise().then((data) => {

==> Meaning we are doing this in parallel, which was explicitely forbidden to avoid concurrency issue on the report.

lndrtrbn:
### Proposed changes

* Call delete mutation sequentially
* Remove all elements in the graph at the end
* Add a progress bar

### Related issues

* #10784 

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/components/graph/components/GraphToolbarRemoveConfirm.tsx
+++ b/opencti-platform/opencti-front/src/components/graph/components/GraphToolbarRemoveConfirm.tsx
@@ -127,6 +127,8 @@ const GraphToolbarRemoveConfirm = ({
     setTotalToDelete(allSelection.length + associatedLinks.length);
 
     // Remove selected nodes and links
+    // /!\ We are voluntary using await in loop to call API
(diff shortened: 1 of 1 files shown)


Archidoit:
Working well !

Is it normal that the box is already checked if I previously did a cascade delete in the report for other entities ? Shouldn't it get back ton uncheck after each deletion? 
image

delemaf (approved):
Test ok :+1:

Archidoit (approved):

Issue #9646, opened 2025-01-18
[Playbook] Problem displaying “label” filter


Lhorus6:
## Description

When I use the “label” filter in a “Query knowledge on a regular basis” component, after refreshing the page, the display is not good (we have the key and the ID). See the repro case

## Environment

OCTI 6.4.7

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Create a playbook
2. Add a “Query knowledge on a regular basis” component with a filter on a label
3. Refresh the page
4. Open the “Query knowledge on a regular basis” component to see the filter

![Image](https://github.com/user-attachments/assets/c4845a7b-2097-4fdc-a12f-68a7ca914ae6)

## Additional information

This does not seem to happen with a “listen” component. 

![Image](https://github.com/user-attachments/assets/14d398e6-77fc-4bb6-9ed4-f03684132da4)

romain-filigran:
reproduced only in the "Query knowledge on a regular basis" playbook component

PR #10810, opened 2025-04-23
[backend] fix cron playbook node filter saving (#9646)

JeremyCloarec:
### Proposed changes

* instead of converting filter when saving cron playbook node, convert it in playbook manager when node is run instead
* keep @me conversion during cron playbook node save: we can't resolve @me dynamically in playbook manager because we no longer know what to resolve @me to

### Related issues

* #9646
*

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/utils/filtering/filtering-utils.ts
+++ b/opencti-platform/opencti-graphql/src/utils/filtering/filtering-utils.ts
@@ -349,17 +349,21 @@ export const checkFiltersValidity = (filterGroup: FilterGroup, noFiltersChecking
  * - check that the key is available with respect to the schema, throws an Error if not
  * - convert relation refs key if any
  */
-export const checkAndConvertFilters = (inputFilterGroup: FilterGroup | null | undefined, userId: string, opts: { noFiltersChecking?: boolean } = {}) => {
+export const checkAndConvertFilters = (
+  inputFilterGroup: FilterGroup | null | undefined,
+  userId: string,
+  opts: { noFiltersChecking?: boolean, noFiltersConvert?: boolean } = {}
+) => {
   if (!inputFilterGroup) {
     return undefined;
   }
   // 01. check filters validity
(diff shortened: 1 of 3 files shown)


Archidoit:
Code is ok, but we also need to do a migration to convert playbook filters that were modified ... 
image

opencti-platform/opencti-graphql/src/utils/filtering/filtering-utils.ts
```diff
@@ -349,17 +349,21 @@ export const checkFiltersValidity = (filterGroup: FilterGroup, noFiltersChecking
...
+) => {
   if (!inputFilterGroup) {
     return undefined;
   }
   // 01. check filters validity
-  const { noFiltersChecking = false } = opts;
+  const { noFiltersChecking = false, noFiltersConvert = false } = opts;
   checkFiltersValidity(inputFilterGroup, noFiltersChecking);
   // 02. replace dynamic @me value
   const filterGroup = replaceMeValuesInFilters(inputFilterGroup, userId);
```
Archidoit, line 364: We shouln't replace 'me' value if noFiltersConvert = true

Archidoit (changes requested):

opencti-platform/opencti-graphql/src/utils/filtering/filtering-utils.ts
JeremyCloarec: In the case of playbooks, we do need to convert the me filter immediately: we have no way of converting the me filter properly when it is used in the playbookManager since we can't resolve the user who configured the playbook filter

Archidoit: mmm Okay, so when we reload the filter page of the playbook, we have the user name and not 'me' anymore. It's not a problem but maybe confusing, we can let this I don't know

Archidoit (approved):

Issue #8685, opened 2024-10-17
Spamming user when modifying a report assigned to someone


SamuelHassine:
## Description

We need to re-think the default notification to avoid spamming (and being blacklisted).

1. Create a incident response case and assign to you
2. Go to knowledge and start adding 15 entities

=> You receive 15 notifications by email:

![image](https://github.com/user-attachments/assets/d46d22ec-9adc-42ee-a8e2-5e20f438c4d8)

SamuelHassine:
It could be great to brainstorm about default notifiers and modifications of containers. Maybe just we need to buffer modifications especially when someone is manipulating a container.

ups1decyber:
Hi, just to add to this: I receive email notifications of my own changes and I don't think they are of value for me at all. If you could prevent notifications of your own actions, that would be nice :)

PR #10832, opened 2025-04-24
[backend] notif buffering implementation (#8685)

JeremyCloarec:
### Proposed changes

* buffer live notifications per targeted entities in publisher manager.
When PUBLISHER_ENABLE_BUFFERING is enabled, notifcation events are stored in memory of publisher manager instead of being sent immediately. 
When the buffer time is older than the time configured in PUBLISHER_BUFFERING_MINUTES, we send the content of the buffer for an entity.
  

### Related issues

* #8685 
*

### Checklist

### Further comments

Possible things to discuss/improve:
* we currently handle the buffers in the same loop as the stream handling (in `notificationHandler` of the publisherManager). We might want to create a separate cron scheduler, the same way it is done in other managers (notificationManager for exemple)
* we might need to handle event loop locking in  `handleEntityNotificationBuffer` method

### DIFF
--- a/opencti-platform/opencti-graphql/config/default.json
+++ b/opencti-platform/opencti-graphql/config/default.json
@@ -302,7 +302,9 @@
   "publisher_manager": {
     "enabled": true,
     "lock_key": "publisher_manager_lock",
-    "interval": 10000
+    "interval": 10000,
+    "enable_buffering": false,
+    "buffering_minutes": 3
   },
   "sync_manager": {
     "enabled": true,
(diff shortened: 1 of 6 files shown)


opencti-platform/opencti-graphql/config/default.json
```diff
@@ -299,7 +299,9 @@
   "publisher_manager": {
     "enabled": true,
     "lock_key": "publisher_manager_lock",
-    "interval": 10000
+    "interval": 10000,
+    "enable_buffering": false,
+    "buffering_minutes": 3
```
SarahBocognano: In seconds, and 30sec or 1min could be enough

SouadHadjiat (commented):
Tested locally, notifications are sent after 1min delay like expected, with all notification data stored in db. 
Would need to test that everything displays well after rebase (once the release build is fixed).

SouadHadjiat (approved):

Issue #10819, opened 2025-04-23
Columns not aligned in status template list


SamuelHassine:
## Description

Columns not aligned in status template list

![Image](https://github.com/user-attachments/assets/911e00b4-a712-4222-a9f3-c4d4c20313fc)

https://filigran.octi.filigran.io/dashboard/settings/vocabularies/status_templates

PR #10845, opened 2025-04-25
[frontend] Columns not aligned in status template list (#10819)

Gwendoline-FAVRE-FELIX:
### Proposed changes

* Refacto Status templates view with datatable

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/10819


opencti-platform/opencti-front/src/private/components/settings/status_templates/StatusTemplates.tsx
```diff
@@ -1,118 +1,173 @@
...
   const { t_i18n } = useFormatter();
   const { setTitle } = useConnectedDocumentModifier();
   setTitle(t_i18n('Status Templates | Taxonomies | Settings'));
-  const classes = useStyles();
-  const { viewStorage, helpers, paginationOptions } = usePaginationLocalStorage<StatusTemplatesLinesPaginationQuery$variables>(
+
+  const initialValues = {
+    searchTerm: '',
+    sortBy: 'name',
+    orderAsc: false,
```
Archidoit: should be true, else the status are not sorted by alphabetical order by default: 
image

opencti-platform/opencti-front/src/private/components/settings/status_templates/StatusTemplates.tsx
```diff
@@ -1,118 +1,173 @@
...
+
+  const queryRef = useQueryLoading<StatusTemplatesLinesPaginationQuery>(
+    statusTemplatesQuery,
+    queryPaginationOptions,
+  );
+
+  const dataColumns: DataTableProps['dataColumns'] = {
+    name: {
+      id: 'name',
+      label: 'Name',
```
Archidoit: adding precisions like label and isSortable is useless here since name is already declared in dataTable utils and don't need to be overwritten in this case - same for usages and color

Archidoit (changes requested):

Archidoit (approved):

PR #10859, opened 2025-04-28
[frontend] Provide all entity types for source/target type filters (#10852)

### LINKED ISSUE #10852: In Knowledge => All relationships view, source type / target type are incorrect
## Description

In Knowledge => All relationships view, source type / target type are incorrect.

Showing relationship type instead of entity types:
Image

Archidoit:
### Proposed changes
For the 'source type' and 'target type' filters: provide all the possible entity type values

### Related issues
(https://github.com/OpenCTI-Platform/opencti/issues/10852)


SarahBocognano (approved):
Tested ✅

PR #10862, opened 2025-04-28
[backend] Revert playbook cron filter rel keys conversion (#9646)

Archidoit:
### Proposed changes
Migration to convert playbook filters for Cron components (these filters have been converted unwantedely with checkAndConvertFilters before the fix : https://github.com/OpenCTI-Platform/opencti/pull/10810 )

### Related issues
https://github.com/OpenCTI-Platform/opencti/issues/9646

### DIFF
--- a/opencti-platform/opencti-graphql/src/migrations/1742823297615-fix-playbook-cron-filters-conversion.js
+++ b/opencti-platform/opencti-graphql/src/migrations/1742823297615-fix-playbook-cron-filters-conversion.js
@@ -1,4 +1,4 @@
-import { logApp } from '../config/conf';
+import { logMigration } from '../config/conf';
 import { listAllEntities } from '../database/middleware-loader';
 import { executionContext, SYSTEM_USER } from '../utils/access';
 import { ENTITY_TYPE_PLAYBOOK } from '../modules/playbook/playbook-types';
@@ -15,7 +15,7 @@ import { schemaRelationsRefDefinition } from '../schema/schema-relationsRef';
 const message = '[MIGRATION] Fix playbook cron filter rel keys conversion';
 
 export const up = async (next) => {
-  logApp.info(`${message} > started`);
+  logMigration.info(`${message} > started`);
   const context = executionContext('migration', SYSTEM_USER);
 
   // -- step 0: define the utils functions --
@@ -147,7 +147,7 @@ export const up = async (next) => {
     READ_DATA_INDICES,
     playbooksUpdateQuery
   );
-  logApp.info(`${message} > done`);
+  logMigration.info(`${message} > done`);
   next();
 };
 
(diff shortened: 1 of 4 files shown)


JeremyCloarec:
Could you rename the migration file? To prevent interfering with migrations pushed on release/current, the naming convention to use for migrations targeting master is to use the previous migration timestamp + 1

opencti-platform/opencti-graphql/src/migrations/1745486086578-fix-playbook-cron-filters-conversion.js
```diff
@@ -0,0 +1,156 @@
...
+import { REL_INDEX_PREFIX } from '../schema/general';
+import { schemaRelationsRefDefinition } from '../schema/schema-relationsRef';
+
+// In playbooks components of type PLAYBOOK_INTERNAL_DATA_CRON
+// revert the conversion of rel filter keys that were made by an unwanted call to checkAndConvertFilters at playbook creation
+
+const message = '[MIGRATION] Fix playbook cron filter rel keys conversion';
+
+export const up = async (next) => {
+  logApp.info(`${message} > started`);
```
JeremyCloarec: Instead of logApp, we should use logMigration in migrations now

Archidoit: done

opencti-platform/opencti-graphql/src/migrations/1742823297615-fix-playbook-cron-filters-conversion.js
```diff
@@ -0,0 +1,156 @@
...
+  const revertRelFilterKeyConversion = (key) => {
+    if (isKeyFromRelConversion(key)) {
+      const relDatabaseName = key.replace(REL_INDEX_PREFIX, '').replace('.*', '');
+      const relInputName = inputNameToDatabaseNameMap.get(relDatabaseName);
+      if (relInputName) {
+        return relInputName;
+      }
+      return key;
+    }
+    return key; // no conversion for key that are not rel
```
JeremyCloarec, line 43: We only revert ref filter, but I think we also need to revert special filter keys & core relationship keys? Since they are converted in `convertRelationRefsFilterKeys` of `filtering-utils.ts`

Archidoit: indeed but these filter keys can't be in playbook filters since playbook filters are limited to stix filtering on stix core objects

Archidoit: (no stix core relationships keys and no special filter keys available for playbooks)

JeremyCloarec (approved):

PR #10871, opened 2025-04-28
[backend] fix unresolved elements in upsert (#10856)

### LINKED ISSUE #10856: In some cases, we have multiple values in "author" (rel_created-by.internal_id)
## Description

In some cases, we have multiple values in "author" (rel_created-by.internal_id).

To determine how this can happen, but consequences are the fact that updating the author is not possible in the UI, also inconsistency in history at upsert etc.

```
POST opencti_*/_search
{
  "query": {
    "script": {
      "script": {
        "source": "doc['rel_created-by.internal_id.keyword'].length > 1",
        "lang": "painless"
      }
    }
  },
  "track_total_hits": true
}
```

Results:

* 206 in filigran instance
* 108 in se2 instance
* 323 in testing

JeremyCloarec:
### Proposed changes

* some upsertElement call were done with elements without their refs resolved. This could cause` multiple: false` refs to not be deleted when upserting the ref.
The ref resolving is now done in upsertElement if needed
* add a migration to only keep one author per entity with multiple authors

### Related issues

* #10856 
*

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/migrations/1742823297617-remove_multiple_authors.js
+++ b/opencti-platform/opencti-graphql/src/migrations/1742823297617-remove_multiple_authors.js
@@ -1,5 +1,5 @@
 import { logMigration } from '../config/conf';
-import { elRawDeleteByQuery, elRawSearch, elRawUpdateByQuery } from '../database/engine';
+import { elRawDeleteByQuery, elRawSearch, elUpdate } from '../database/engine';
 import { READ_DATA_INDICES, READ_INDEX_STIX_META_RELATIONSHIPS } from '../database/utils';
 import { executionContext, SYSTEM_USER } from '../utils/access';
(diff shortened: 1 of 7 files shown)


opencti-platform/opencti-graphql/src/migrations/1742823297617-remove_multiple_authors.js
```diff
@@ -0,0 +1,119 @@
...
+        params: { authorIdToKeep },
+        lang: 'painless'
+      },
+      query: {
+        match: {
+          _id: currentEntityWithMultipleAuthors._id
+        }
+      }
+    };
+    await elRawUpdateByQuery({
```
SouadHadjiat: use elUpdate directly

SouadHadjiat (approved):

PR #10895, opened 2025-04-30
[backend/frontend] Introducing JSON feed + mapper in beta with feature flag

### LINKED ISSUE #8258: Custom scheduling for CSV and RSS Feed Ingesters
## Use case

Currently, built-in Feed ingesters (TAXII, RSS, CSV) are scheduled to run every 30 seconds by default.
Not all built-in ingester configurations can have the same scheduling requirement. 
Also, some information sources do not accept to fetch data too often (e.x. TOR Node List allows to fetch data only every 15 or 30 minutes).

## Current Workaround

N/A

## Proposed Solution

Add the ability to set an execution interval per built-in ingester RSS or CSV configuration.
TAXII ingester not concerned in this feature.

### LINKED ISSUE #11062: Add support for JSON Feed and JSON Mapper
## Use case

New feature that enables support for JSON feeds and a corresponding JSON mapper, following the same logic and structure we previously used for CSV files/feeds.

richard-julien:
Concept of this PR is to introduce a first version of JSON feed along with associated mapper.
The aim is to connect to any JSON api and convert the feed to STIX 2.1 data to be ingested in the platform.
JSON mapping can be very complex, only the "simple" mode (no formula and variables) is introduced.
This feature is linked under the feature flag "JSON_MAPPER"

Issues: 

- https://github.com/OpenCTI-Platform/opencti/issues/11062
- https://github.com/OpenCTI-Platform/opencti/issues/8258


opencti-platform/opencti-front/src/private/components/common/CodeBlock.tsx
```diff
@@ -7,15 +7,16 @@ import type { Theme } from '../../../components/Theme';
 interface CodeBlockProps {
   code: string;
   language: string;
+  customHeight?: string;
 }
 
-const CodeBlock: FunctionComponent<CodeBlockProps> = ({ language, code }) => {
+const CodeBlock: FunctionComponent<CodeBlockProps> = ({ language, code, customHeight }) => {
```
jpkha:
```suggestion
const CodeBlock: FunctionComponent<CodeBlockProps> = ({ language, code, customHeight = '400px' }) => {
```

opencti-platform/opencti-front/src/private/components/common/CodeBlock.tsx
```diff
@@ -7,15 +7,16 @@ import type { Theme } from '../../../components/Theme';
...
   const theme = useTheme<Theme>();
   return (
     <SyntaxHighlighter
       language={language}
       style={theme.palette.mode === 'dark' ? a11yDark : coy}
-      customStyle={{ height: '400px', minWidth: '550px' }}
+      customStyle={{ height: customHeight ?? '400px', minWidth: '550px' }}
```
jpkha:
```suggestion
      customStyle={{ height: customHeight, minWidth: '550px' }}
```

Follow the first suggestion you can just use customHeight by giving '400px' by default

opencti-platform/opencti-graphql/src/back.js
```diff
@@ -16,9 +16,10 @@ import './manager/index';
 import './config/tracing';
 // endregion
 import { SEMRESATTRS_SERVICE_NAME } from '@opentelemetry/semantic-conventions';
-import { platformStart } from './boot';
+// import { platformStart } from './boot';
```
jpkha: it's commented here but added on the next line

opencti-platform/opencti-graphql/src/back.js
```diff
@@ -59,3 +60,22 @@ if (ENABLED_EVENT_LOOP_MONITORING) {
...
+// trino.query(
+//   'select * from customer'
+// ).then(async (iter) => {
+//   for await (const queryResult of iter) {
+//     console.log(`iter result ${(queryResult.data ?? []).length}`);
+//   }
+// });
+
+// testJsonMapper({ externalUri: 'https://4.233.151.63:444' }, json4, mispJsonMapper4);
+// jsonExecutor().then(() => jsonExecutor().then(() => jsonExecutor()));
```
aHenryJard: Do you want to keep this comment ? Is it interesting to keep that as test maybe ?

romain-filigran:
@richard-julien : The “schedule period” update doesn't work on a RSS configuration. (Maybe the same for CSV Feed).
Similarly, it seems that updating the “description” doesn't work either.

Issue #10897, opened 2025-05-01
[BUG] Multiple persona tags on Threat Actor when linked via multiple knowledge graphs


ParamConstructor:
## Description

Persona label is displayed multiple times on Threat Actor report when linked via multiple reports or other platform entities.

## Environment

1. OS (where OpenCTI server runs):  Mac OS
2. OpenCTI version:  OpenCTI 6.6.8
3. OpenCTI client: frontend
4. Other environment details:

## Reproducible Steps

Steps to create the smallest reproducible scenario:

Creating an "Also Known As" relationship directly on a Threat Actor, then navigating to a Report (adding the Threat Actor and Persona in report, then creating a "Also Known As" relationship within the Knowledge graph of the report) causes the Threat Actor display page to list the Persona Label multiple times.

## Expected Output

Persona link shows only one label

## Actual Output

Persona linked shows label multiple times

## Additional information

Correction PR: https://github.com/OpenCTI-Platform/opencti/pull/10898

## Screenshots (optional)

![Image](https://github.com/user-attachments/assets/ef76c342-ea9d-4742-852d-573a5e5f06ad)

PR #10898, opened 2025-05-01
[frontend] Patch to correct duplicate persona chips

ParamConstructor:
See Issue - https://github.com/OpenCTI-Platform/opencti/issues/10897

### Proposed changes

* Deduplicates erroneous push for seen persona chips on a Threat Actor, resulting in duplicate chips displayed

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/10897

### Checklist

### Further comments

N/A

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/threats/threat_actors_individual/ThreatActorIndividualDetailsChips.tsx
+++ b/opencti-platform/opencti-front/src/private/components/threats/threat_actors_individual/ThreatActorIndividualDetailsChips.tsx
@@ -33,6 +33,9 @@ const typeMappings: Record<SupportedTypes, MappingFields> = {
   },
 };
 
+interface SeenPersonaID<T> {
+  [Key:string]: T;
+}
 interface ThreatActorIndividualDetailsChipsProps {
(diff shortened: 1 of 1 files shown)


opencti-platform/opencti-front/src/private/components/threats/threat_actors_individual/ThreatActorIndividualDetailsChips.tsx
```diff
@@ -47,10 +50,14 @@ ThreatActorIndividualDetailsChipsProps
   const { title, field, path, AddComponent } = typeMappings[relType];
 
   const getRelationshipsOfType = (rel_type: SupportedTypes) => {
+    const seen_persona_id_dict: SeenPersonaID<number> = {};
```
lndrtrbn: You can do it by using a Set, it will avoid using a Map where we are not interested in the values but only in keys.

ParamConstructor: @lndrtrbn - Swapped to a Set object and rebased.

lndrtrbn (approved):

Issue #10818, opened 2025-04-23
Vocabularies screen layout is broken


SamuelHassine:
## Description

Vocabularies screen layout is broken.

* Columns not aligned
* Hover effect not correct
* etc.

https://filigran.octi.filigran.io/dashboard/settings/vocabularies/fields/attack_resource_level_ov

![Image](https://github.com/user-attachments/assets/29f5c302-ffeb-4ec1-9711-c0110b990375)

PR #10960, opened 2025-05-07
[frontend] Vocabularies screen layout is broken (#10818)

Gwendoline-FAVRE-FELIX:
### Proposed changes

* Refacto Vocabularies view with Datatable

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/10818


opencti-platform/opencti-front/src/private/components/settings/Vocabularies.tsx
```diff
@@ -1,199 +1,214 @@
...
+    entity_types: {
+      label: 'Used in',
+      percentWidth: 20,
+      isSortable: false,
+      render: (node: useVocabularyCategory_Vocabularynode$data) => (
+        <>
+          {node.category.entity_types.map((type) => (
+            <Chip
+              key={type}
+              style={{
```
jbmeslin: Never use inline style

opencti-platform/opencti-front/src/private/components/settings/Vocabularies.tsx
```diff
@@ -1,199 +1,214 @@
 import React from 'react';
-import makeStyles from '@mui/styles/makeStyles';
 import Chip from '@mui/material/Chip';
 import { useParams } from 'react-router-dom';
-import ListLines from '../../../components/list_lines/ListLines';
+import { graphql } from 'react-relay';
+import { ShortTextOutlined } from '@mui/icons-material';
+import VocabularyPopover from '@components/settings/attributes/VocabularyPopover';
+import { VocabulariesLinesPaginationQuery, VocabulariesLinesPaginationQuery$variables } from '@components/settings/__generated__/VocabulariesLinesPaginationQuery.graphql';
+import { useTheme } from '@mui/material';
```
jbmeslin, line 7: It is better to have the external libraries first, then the generated ones, and finally the project ones.
This allows us to see the hierarchy of dependencies between what we use and what we create.

CelineSebe:
I found several issues:
- Aliases are not displayed in the table
- After creation, we need to refresh the page to display the new vocabulary
- The IconButton doesn't need to have size= 'large' prop to avoid this effect:
image
- I encountered this error when creating a new vocabulary:
"message":"Error: Cant insert node on not found connection Pagination_vocabularies with filters {\"categor
y\":\"account_type_ov\",\"orderMode\":\"asc\",\"orderBy\":\"name\"}"

Gwendoline-FAVRE-FELIX:
I'll fix the rest, but the aliases are displayed. Did you give the same value to the alias and the name? The back filter if so.
Aliases

Gwendoline-FAVRE-FELIX:
I can change it so that there's a hyphen if the field is empty.

CelineSebe:
> I'll fix the rest, but the aliases are displayed. Did you give the same value to the alias and the name? The back filter if so. !

Thanks for this. I looked again, and I can no longer reproduce the issue with the alias

SarahBocognano (approved):

PR #10961, opened 2025-05-07
[frontend] Author field cannot be emptied anymore (#10857)

### LINKED ISSUE #10857: Author field cannot be emptied anymore
## Description

Author field cannot be emptied anymore

1. Put an author
2. Try to remove it

=> Nothing happen

Archidoit:
### Proposed changes
We should be able to empty the Author field in stix domain objects via the edition form

### Related issues
https://github.com/OpenCTI-Platform/opencti/issues/10857

### DIFF
--- a/opencti-platform/opencti-front/src/utils/hooks/useFormEditor.ts
+++ b/opencti-platform/opencti-front/src/utils/hooks/useFormEditor.ts
@@ -140,7 +140,7 @@ const useFormEditor = (
   };
 
   // Simple
-  const changeCreated = (name: string, value: Option | string) => {
+  const changeCreated = (name: string, value: Option | '') => {
     if (!enableReferences) {
       validate(name, value !== '' ? value : null, () => {
         const finalValue = value !== '' ? (value as Option).value : null;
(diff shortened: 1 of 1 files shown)


opencti-platform/opencti-front/src/utils/hooks/useFormEditor.ts
```diff
@@ -62,7 +62,7 @@ const useFormEditor = (
 
   const validate = (
     name: string,
-    values: number | number[] | string | Date | Option | Option[],
+    values: number | number[] | string | Date | Option | Option[] | null,
```
jbmeslin, line 65: and after we could add undefined

aHenryJard (approved):

opencti-platform/opencti-front/src/utils/hooks/useFormEditor.ts
```diff
@@ -140,16 +140,17 @@ const useFormEditor = (
   };
 
   // Simple
-  const changeCreated = (name: string, value: Option) => {
+  const changeCreated = (name: string, value: Option | string) => {
```
lndrtrbn: `value: Option | ''`
We do not want any string, the only case is empty string when we are emptying the field

lndrtrbn (changes requested):

opencti-platform/opencti-front/src/utils/hooks/useFormEditor.ts
Archidoit: fixed :)

lndrtrbn (approved):

PR #10967, opened 2025-05-07
[backend] Improving code and test coverage on data sharing

aHenryJard:
### Proposed changes

* Add a check on data sharing > TAXIIAPI & CSV Feed and Stream
* Add test coverage on all data > sharing resolvers

### Related issues

*
*

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/domain/stream.js
+++ b/opencti-platform/opencti-graphql/src/domain/stream.js
@@ -4,11 +4,12 @@ import { createEntity, deleteElementById, updateAttribute, } from '../database/m
 import { listEntities, storeLoadById } from '../database/middleware-loader';
 import { delEditContext, notify, setEditContext } from '../database/redis';
 import { BUS_TOPICS } from '../config/conf';
-import { MEMBER_ACCESS_RIGHT_VIEW, SYSTEM_USER, TAXIIAPI_SETCOLLECTIONS } from '../utils/access';
+import { isUserHasCapability, MEMBER_ACCESS_RIGHT_VIEW, SYSTEM_USER, TAXIIAPI_SETCOLLECTIONS } from '../utils/access';
 import { publishUserAction } from '../listener/UserActionListener';
 import { addFilter } from '../utils/filtering/filtering-utils';
 import { validateFilterGroupForStixMatch } from '../utils/filtering/filtering-stix/stix-filtering';
 import { authorizedMembers } from '../schema/attribute-definition';
+import { TAXIIAPI } from './user';
 
 // Stream graphQL handlers
 export const createStreamCollection = async (context, user, input) => {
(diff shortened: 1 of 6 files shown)


opencti-platform/opencti-graphql/src/domain/stream.js
```diff
@@ -40,7 +41,7 @@ export const findById = async (context, user, collectionId) => {
 };
 export const findAll = (context, user, args) => {
   // If user is logged, list all streams where the user have access.
-  if (user) {
+  if (user && isUserHasCapability(user, TAXIIAPI)) {
```
jbmeslin, line 44: It would be better to put the return of the function in a variable and/or test the User variable in the function

opencti-platform/opencti-graphql/tests/02-integration/02-resolvers/csvFeed-test.ts
```diff
@@ -19,7 +22,7 @@ describe('CSV Feed resolver standard behavior', () => {
       query: CREATE_FEED,
       variables: {
         input: {
-          name: 'List of created cities',
+          name: 'List of created cities - internal csv feed',
```
jbmeslin, line 25: When doing i18n it is better to give keys ie:  "domain.screen.label" it is better for refactoring

here it a test so it ok but if your are testing a key coming from the real view, 
when you change the label of your screen, will you change the value and also the key?

opencti-platform/opencti-graphql/tests/02-integration/02-resolvers/stream-test.ts
```diff
@@ -0,0 +1,280 @@
...
+                },
+            `,
+      variables: { input: restrictedStreamInput }
+    });
+
+    logApp.info('amberRestrictedStreamResponse:', restrictedStreamResponse);
+    expect(restrictedStreamResponse?.data?.streamCollectionAdd?.id).toBeDefined();
+    restrictedStreamId = restrictedStreamResponse?.data?.streamCollectionAdd?.id;
+
+    expect(restrictedStreamResponse?.data?.streamCollectionAdd?.name).toBe('Not public stream with empty auth member for resolver tests');
```
jbmeslin, line 124: maybe extract restrictedStreamResponse?.data?.streamCollectionAdd?  


for better readability

opencti-platform/opencti-graphql/tests/02-integration/02-resolvers/csvFeed-test.ts
aHenryJard: I don't understand this feedback, it's an graphQL API call there is never i18n in that context.

opencti-platform/opencti-graphql/tests/02-integration/02-resolvers/stream-test.ts
aHenryJard: ok, since CI is green for now, I will do if there is some production code changes requires

opencti-platform/opencti-graphql/src/domain/stream.js
aHenryJard: how is it better ? I prefer explicit checks on my side: "user is defined and has required capability" is more readable because both conditions are needed.

marieflorescontact (approved):
👌

PR #10971, opened 2025-05-11
[backend] Various performance improvments

richard-julien:
Various performance improvments
- Usage of filter instead of direct bool to prevent scoring computing
- Prevent ordering if not needed
- Prevent usage track_total_hits if not needed
- Implement a generic asyncMap with optional transform to prevent event loop blocking
- Add a buffer time in redis stream listening to prevent too much call and use bulk execution


opencti-platform/opencti-graphql/src/database/engine.js
```diff
@@ -3059,9 +3069,7 @@ const elQueryBodyBuilder = async (context, user, options) => {
       }
     }
     // Add standard_id if not specify to ensure ordering uniqueness
```
lndrtrbn: So this comment is not relevant anymore?

opencti-platform/opencti-graphql/src/database/engine.js
```diff
@@ -3072,8 +3080,8 @@ const elQueryBodyBuilder = async (context, user, options) => {
         script: { source, params },
       };
     }
-  } else { // If not ordering criteria, order by standard_id
-    ordering.push({ 'standard_id.keyword': 'asc' });
+  } else { // If not ordering criteria, order by _doc
+    ordering.push({ _doc: 'asc' });
```
lndrtrbn: I'm not sure to understand this `else` case. We are doing the same thing we did line 3072, so is it useful?

SouadHadjiat: I guess it could be simplified by adding `ordering.push({ _doc: 'asc' });` all the time, by moving line 3072 to 3084 and removing the else condition.

opencti-platform/opencti-graphql/src/database/engine.js
richard-julien: Removed

opencti-platform/opencti-graphql/src/database/engine.js
richard-julien: Fixed

opencti-platform/opencti-graphql/src/database/engine.js
```diff
@@ -3072,9 +3078,9 @@ const elQueryBodyBuilder = async (context, user, options) => {
         script: { source, params },
       };
     }
-  } else { // If not ordering criteria, order by standard_id
-    ordering.push({ 'standard_id.keyword': 'asc' });
   }
+  // Add _doc as default order
+  ordering.push({ _doc: 'asc' });
```
SouadHadjiat, line 3127: are we sure this _doc sorting is also supported by opensearch ? I can't find it in documentation

SouadHadjiat (approved):

opencti-platform/opencti-graphql/src/database/engine.js
SouadHadjiat: just tested locally with opensearch, everything works well

Issue #10838, opened 2025-04-25
Style in sightings history is different


alice-debra:
History classique style 

<img width="577" alt="Image" src="https://github.com/user-attachments/assets/217dcf9e-5e39-4beb-9219-42a5ce2adb42" />

History style in sightings

<img width="1491" alt="Image" src="https://github.com/user-attachments/assets/50dbda19-6292-4dbd-a130-44f330b5c618" />

<img width="655" alt="Image" src="https://github.com/user-attachments/assets/97055363-a39f-4175-94de-965c2efa43d3" />

PR #11001, opened 2025-05-14
[frontend] Style in sightings history is different (#10838)

SarahBocognano:
### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/10838


lndrtrbn (commented):
I don't know if it should be done in this PR but in my opinion we should have a component HistoryLine that is reused everywhere we need to display lines of history instead of duplicating code and styles through the different pages displaying history

SarahBocognano:
You are right @lndrtrbn it's the exact same component, I refactored it

Archidoit (approved):

PR #11004, opened 2025-05-14
[backend] now also send stream message in digest notifications (#8685)

JeremyCloarec:
### Proposed changes

* now also send stream message in digest notifications
*

### Related issues

* #8685 
*

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/manager/publisherManager.ts
+++ b/opencti-platform/opencti-graphql/src/manager/publisherManager.ts
@@ -196,18 +196,12 @@ const processLiveNotificationEvent = async (
   notificationMap: Map<string, BasicStoreEntityTrigger>,
   event: KnowledgeNotificationEvent | ActivityNotificationEvent | ActionNotificationEvent
 ) => {
-  const { targets, data: instance, origin: { user_id } } = event;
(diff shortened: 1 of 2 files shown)


opencti-platform/opencti-graphql/src/manager/publisherManager.ts
```diff
@@ -175,31 +175,45 @@ const processNotificationEvent = async (
   }
 };
 
+const createFullNotificationMessage = (notificationMessage: string, usersMap: Map<string, AuthUser>, streamMessage?: string, origin?: Partial<UserOrigin>) => {
+  let fullMessage = notificationMessage;
+  if (origin && streamMessage) {
+    const { user_id } = origin;
+    const streamUser = usersMap.get(user_id as string) as AuthUser;
```
lndrtrbn: `as AuthUser` is not needed

lndrtrbn: not a fan neither of `as string` to bypass the fact that TypeScript is telling that user_id can be `undefined`
it would be better to trust TypeScript and use `.get(user_id ?? '')` to be sure to pass a string as argument

JeremyCloarec: I'll improve it yes you're right, this was copy pasted and refactored from what was done in `processLiveNotificationEvent` before

lndrtrbn (approved):

PR #11019, opened 2025-05-15
[frontend] fix broken layout on open import dialog (#10761)

### LINKED ISSUE #10761: Layout is broken when loading the import panel
## Description

Layout is broken when loading the import panel

1. Search bar is still displayed
2. Loader on the right misplaced / wired

https://github.com/user-attachments/assets/200adaa4-8ba9-4f85-a5cc-c02445f4e7e6

delemaf:
### Proposed changes

* Fix flashing import dialog

### Related issues

* Closes https://github.com/OpenCTI-Platform/opencti/issues/10761

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/common/files/import_files/ImportFilesContext.tsx
+++ b/opencti-platform/opencti-front/src/private/components/common/files/import_files/ImportFilesContext.tsx
@@ -5,7 +5,6 @@ import { ImportFilesContextQuery } from '@components/common/files/import_files/_
 import { ImportFilesContextGuessMimeTypeQuery$data } from '@components/common/files/import_files/__generated__/ImportFilesContextGuessMimeTypeQuery.graphql';
 import useGranted from '../../../../../utils/hooks/useGranted';
 import useQueryLoading from '../../../../../utils/hooks/useQueryLoading';
-import Loader, { LoaderVariant } from '../../../../../components/Loader';
 import useDraftContext from '../../../../../utils/hooks/useDraftContext';
 import { fetchQuery } from '../../../../../relay/environment';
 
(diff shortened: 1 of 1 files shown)


opencti-platform/opencti-front/src/private/components/common/files/import_files/ImportFilesDialog.tsx
```diff
@@ -478,8 +478,17 @@ const ImportFiles = ({ open, handleClose }: ImportFilesDialogProps) => {
...
 const ImportFilesDialog = ({ open, entityId, handleClose }: ImportFilesDialogProps) => {
+  const [dialogKey, setDialogKey] = useState(0);
+
+  useEffect(() => {
+    // Resets all dialog state on close by forcing a complete component remount via key change
+    if (!open) {
+      setDialogKey((prev) => prev + 1);
+    }
+  }, [open]);
+
```
lndrtrbn: This solution looks more like a workaround to a bad use of some React concepts somewhere.
Can you explain the issue that this fix is supposed to solve ? (I mean technically, not functionally)

Archidoit:
Working well 👍

opencti-platform/opencti-front/src/private/components/common/files/import_files/ImportFilesDialog.tsx
delemaf: It's a key-controlled reset (it doesn't really unmount the component) but it's more a workaround, I'll try to reset the state instead :+1:

delemaf: After investigation it was the loader that broke the topbar, so I'll keep the ImportFilesDialog unmounted instead of resetting state

frapuks (approved):

Issue #10598, opened 2025-04-11
History is duplicated when uploading a new version of the file. (upsert)


SamuelHassine:
## Description

History is duplicated when uploading a new version of the file (upsert)

![Image](https://github.com/user-attachments/assets/190bd642-eda1-4209-8e40-8375cfdd8328)

We want this but we have to change the wording of the message:

* [C] mandiant adds 23-00042229.pdf in files and external_references
* [C] mandiant uploads a new version of 23-00042229.pdf in files and external_references

PR #11038, opened 2025-05-19
[frontend/backend] History message for new version of a file upload (#10598)

Archidoit:
### Proposed changes
Different history message when uploading a file in an entity : 
- When uploading a file in an entity: 'add file_name in files'
- When uploading a file that already exist in an entity (file upsert): 'adds a new version of file_name in files'

### Related issues
https://github.com/OpenCTI-Platform/opencti/issues/10598

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectHistoryLine.jsx
+++ b/opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectHistoryLine.jsx
@@ -216,7 +216,7 @@ class StixCoreObjectHistoryLineComponent extends Component {
           </Avatar>
         );
       }
-      if (eventScope === 'update' && (eventMesage.includes('adds') || eventMesage.includes('new version'))) {
+      if (eventScope === 'update' && (eventMesage.includes('adds'))) {
         return (
           <Avatar
             sx={{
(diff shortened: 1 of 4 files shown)


opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectHistoryLine.jsx
```diff
@@ -216,7 +216,7 @@ class StixCoreObjectHistoryLineComponent extends Component {
           </Avatar>
         );
       }
-      if (eventScope === 'update' && eventMesage.includes('adds')) {
+      if (eventScope === 'update' && (eventMesage.includes('adds'))) {
```
marieflorescontact: is it realyy necessary to add () in this case?

marieflorescontact (commented):
locally tested ok 👌

opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectHistoryLine.jsx
Archidoit: indeed, I forgot to remove this, thanks :)

marieflorescontact (approved):

PR #11054, opened 2025-05-20
[backend] add correct user_inside_platform_organization in taskManager context (#11043)

### LINKED ISSUE #11043: Inability to Add Observables from Report to Another Container (Grouping) for Platform Org Users without "Bypass all capabilities"
## Description
In the OpenCTI version 6.6.6, users who are part of the platform organization and possess necessary capabilities but lack the "Bypass all capabilities" cannot add observables from a report to a container, despite having the appropriate permissions. This operation fails silently without any errors in the UI or logs, yet the task is marked as successfully completed. When an additional organization is leveraged or the "Bypass all capabilities" option is enabled, the operation proceeds as expected, indicating a bug in the handling of permissions for platform organization users.

## Environment
6.6.6

## Reproducible Steps

1. Create a grouping.
2. Navigate to Analysis → Reports and select a report containing observables.
3. Select one or more observables from the report.
4. Click "Add to container" and select the new grouping.
5. Wait for the task to indicate completion.

## Expected Output
The selected observables should be successfully added to the chosen container without requiring additional organization memberships or the "Bypass all capabilities" option.

## Actual Output
Observables are not added to the container when using a user who is part of only the platform organization without the "Bypass all capabilities" permission. No errors are displayed, and logs report successful task completion.

JeremyCloarec:
### Proposed changes

* user_inside_platform_organization was always set to false in taskManager. It is now computed from the task initiator user
*

### Related issues

* #11043 
*

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/http/httpAuthenticatedContext.js
+++ b/opencti-platform/opencti-graphql/src/http/httpAuthenticatedContext.js
@@ -1,4 +1,4 @@
-import { executionContext, isBypassUser, SYSTEM_USER } from '../utils/access';
+import { executionContext, isUserInPlatformOrganization, SYSTEM_USER } from '../utils/access';
 import { getEntityFromCache } from '../database/cache';
 import { ENTITY_TYPE_SETTINGS } from '../schema/internalObject';
 import { authenticateUserFromRequest, userWithOrigin } from '../domain/user';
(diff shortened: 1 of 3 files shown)


SouadHadjiat (approved):
All good for task manager fix.
I'm wondering if we would need to change executionContext where user_inside_platform_organization is set to false by default, to check with @richard-julien

PR #11055, opened 2025-05-20
[frontend] Move page Rules engine to TypeScript (#10590)

### LINKED ISSUE #10590: Rule engine page keeps refreshing
## Description

Since 6.6 the rule engine page refreshes itself at random interval making it difficult to navigate through.

lndrtrbn:
### Proposed changes

* Move page Rules engine from JavaScript to TypeScript
* Separate code into multiple smaller components
* Improve UI (spacings, font, no more truncated text inside rules boxes)
* Fix initial issue about page refresh

### Related issues

* #10590 

### Checklist


opencti-platform/opencti-front/src/utils/hooks/useInterval.ts
```diff
@@ -0,0 +1,25 @@
+import { useEffect, useRef } from 'react';
+
+// https://www.geeksforgeeks.org/reactjs-useinterval-custom-hook/
```
SarahBocognano: Do we want to keep that ?

lndrtrbn: idk, I added explicitly because sometimes I find useful when copy pasting some complicated code but this one is quite simple, not sure the source is required

opencti-platform/opencti-front/src/private/components/settings/rules/RulesStatusChangeDialog.tsx
```diff
@@ -0,0 +1,109 @@
...
+  return (
+    <Dialog
+      keepMounted
+      open={status !== undefined}
+      slotProps={{ paper: { elevation: 1 } }}
+      slots={{ transition: Transition }}
+      onClose={close}
+    >
+      <DialogContent>
+        <DialogContentText>
```
labo-flg: `<Alert>` is a `<div>,` `<DialogContentText>` is a `<p>`
This causes "`<div>` cannot appear as a child of `<p>`" error in the console.

I suspect we can just get rid of the `<DialogContentText>` wrapper.

labo-flg:
Tested OK locally, but I did not really have the refresh issue in the first place.
How did you solve it actually ?

lndrtrbn:
> Tested OK locally, but I did not really have the refresh issue in the first place. How did you solve it actually ?

Replacing the QueryRenderer + using a ref for query variables fix the issue. I cannot explain the original issue, I also had difficulties to reproduce it locally.

opencti-platform/opencti-front/src/private/components/settings/rules/Rules.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+  const ruleEngineEnabled = platformModuleHelpers.isRuleEngineEnable();
+
+  const breadcrumb = [
+    { label: t_i18n('Settings') },
+    { label: t_i18n('Customization') },
+    { label: t_i18n('Rules engine'), current: true },
+  ];
+
+  const queryVariables = useRef({ startDate: yearsAgo(1), endDate: dayAgo() });
```
labo-flg, line 55: just a quick comment would be nice

labo-flg (approved):

opencti-platform/opencti-front/src/private/components/settings/rules/Rules.tsx
lndrtrbn:
```suggestion
  // Use a ref for variables otherwise we have new variables at each calls (with different seconds)
  // which can cause blink effects with slower networks.
  const queryVariables = useRef({ startDate: yearsAgo(1), endDate: dayAgo() });
```

Issue #10798, opened 2025-04-22
feat(ux): automatically add created entities to a container


SamuelHassine:
## Description

In the shoes of a user, simply add a list of 40 IPs addresses to a container

https://github.com/user-attachments/assets/860bf9e0-d3e7-4845-8874-ea5fd2ad53dc

We do do better.... But... the worst case scenario is if you would like to add the indicators in addition to observables and you check the box. That is just impossible.

nino-filigran:
Related to: https://github.com/OpenCTI-Platform/opencti/issues/7779

nino-filigran:
After a discussion today, even though we understand the pain, it's still a feature (CF ticket opened above), and even though the identified fix is easy from a technical standpoint, it needs to be thoroughly checked: indeed, all create APIs needs to be update to incorporate a Container ID, this change needs also to be added in python client avoid to create any tech debt and we need a minor change in UI to let user decide wheteher or not the created entities/observables should be added in container. 

On the top of this, as a side work, we also need to transform the inline component into a datable. The benefit of this would be to allow a Shift+select, to at least, provide a "simpler" to your issue.

Issue #7779, opened 2024-07-23
feat(graph): automatically add the entity to the report once created in the context of a container


nino-filigran:
## Use case

In the context of a container, when creating an entity within the knowledge graph (or within the list), the entity is not directly added to the container?
User always have to to do an extra step to add the entity within the container/graph, as shown in the screens below.
The goal would be that this entity is added automatically to the container.

- Creation of an entity
![image](https://github.com/user-attachments/assets/f909f790-2f07-484e-86f3-d0b7c93f3ab8)

- Entity create and added to the list 
![image](https://github.com/user-attachments/assets/71bc4de6-1c28-4175-bd1f-da840a385e3b)

nino-filigran:
Follow up conversation today: will require some work to ensure all API, python client are aligned.

PR #11056, opened 2025-05-20
[frontend] Automatically add created entities in a container (#10798)

ValentinBouzinFiligran:
### Proposed changes

* after creation, create relation between new entity and container

### Related issues

* #10798 
* #7779 

### Checklist


JeremyCloarec:
It only works for entities, but it doesn't look like it was implemented for observables: we probably also want to do this for observables

JeremyCloarec:
It looks like the updater of the Add entities drawer doesn't always work properly: I have some created entities that are not selected (even though they were added correctly)
image

PR #11066, opened 2025-05-21
[frontend] remove loader in importFile

frapuks:
### Proposed changes

* delete the loader in importFile to avoid breaking the UI when displaying the loader
* issue metionned in notion (only in release/current)
* Linked to https://github.com/OpenCTI-Platform/opencti/pull/11019 into master

### Before
The home page during the first loading
image


frapuks:
The PR linked was merged into master. We are waiting for a sync between master and release to check if the issue in release will be fixed.

Issue #10350, opened 2025-03-25
Ability to import/export CSV feeds 


EllynBsc:
## Use case
We're developing a CSV Feeds library in XTM Hub.

Linked to these 3 issues:

-  [import CSV feed](https://github.com/orgs/FiligranHQ/projects/17/views/23?pane=issue&itemId=105469264&issue=FiligranHQ%7Cxtm-hub%7C492) in the XTM Hub backlog

- [export CSV feed](https://github.com/orgs/FiligranHQ/projects/17/views/23?pane=issue&itemId=105469238&issue=FiligranHQ%7Cxtm-hub%7C493) in the XTM Hub backlog

- [Ability to edit a CSV Feed with its inline CSV mapper](https://github.com/FiligranHQ/xtm-hub/issues/614) in the XTM backlog

- [add 'Import from Hub' CTA in CSV feed](https://github.com/orgs/FiligranHQ/projects/17/views/23?pane=issue&itemId=103538673&issue=FiligranHQ%7Cxtm-hub%7C450) in the XTM Hub backlog

We need to develop the ability to import/export CSV feeds in order to populate our csv feed library in the hub.

PR #11085, opened 2025-05-22
[Backend/Frontend] Add CSV Feed export functionality

jpkha:
### Proposed changes

* Add the possibility to export CSV Feed configuration in json. It will be always inline

### Checklist

### Further comments

Related #10350

### DIFF
--- a/opencti-platform/opencti-graphql/src/modules/ingestion/ingestion-csv-domain.ts
+++ b/opencti-platform/opencti-graphql/src/modules/ingestion/ingestion-csv-domain.ts
@@ -6,9 +6,9 @@ import { createEntity, deleteElementById, patchAttribute, updateAttribute } from
 import { publishUserAction } from '../../listener/UserActionListener';
 import { type CsvMapperTestResult, type EditInput, type IngestionCsvAddInput, IngestionCsvMapperType } from '../../generated/graphql';
 import { notify } from '../../database/redis';
-import { BUS_TOPICS, isFeatureEnabled } from '../../config/conf';
+import { BUS_TOPICS, isFeatureEnabled, PLATFORM_VERSION } from '../../config/conf';
 import { ABSTRACT_INTERNAL_OBJECT } from '../../schema/general';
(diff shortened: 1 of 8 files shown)


opencti-platform/opencti-graphql/src/modules/ingestion/ingestion-csv-domain.ts
```diff
@@ -217,3 +218,59 @@ export const csvFeedGetCsvMapper = (context: AuthContext, ingestionCsv: BasicSto
...
+    openCTI_version: PLATFORM_VERSION,
+    type: 'csvFeeds',
+    configuration: {
+      name,
+      description,
+      uri,
+      authentication_type,
+      authentication_value: '',
+      markings,
+      csv_mapper_type: 'inline',
```
lndrtrbn, line 267: I do not understand why we set always the type as `inline` here knowing that the function `getCsvMapper` above seems to not return only inline mappers. Can you explain it ? Maybe I am missing something because I don't know what an inline csv mapper means

jpkha: Yes, it's always set to inline for export because we can't guarantee that a csv_mapper exists between each OCTI on different platforms and we don't want to create csv mapper each time. So the idea is to "inline" it, basically, that means embedding the csv_mapper config directly into the CSVFeed document instead of just referencing it by ID. 

The getCsvMapper function above just gives you a basic CSV mapper config. If it comes from an ID, we pull out the csv_mapper config so we can include it in the CSVFeed export. If it's marked as "inline", we just use the existing csv_mapper as-is.

Let me know if it's still not clear, we can take 10 minutes to chat about the use case if that helps.

lndrtrbn: Thanks, it's clear

lndrtrbn (approved):
Tested ok. 

In terms of UX it is confusing to see the name of the inline mapper in update form of csv feed but it does not match any existing mapper in the list. Would be nice to have some improvements here to explain what is going on

jpkha:
Yes @lndrtrbn you are right it's complex UI, it will be done in the next chunk.

Issue #10985, opened 2025-05-13
Revamp floating action buttons (FAB) and header menus


aHenryJard:
## Use case

Improve UI/UX on floating button actions and 3 dots menu.

## If the feature request is approved, would you be willing to submit a PR?

Yes / No (Help can be provided if you need assistance submitting a PR)

PR #11115, opened 2025-05-23
[frontend] update button alignments and placements (#10985)

natalie-natsu:
### Related issues

* [frontend] move dashboard settings to user profile page (#10985)

### Checklist


opencti-platform/opencti-front/src/private/components/common/stix_core_relationships/StixCoreRelationshipOverview.jsx
```diff
@@ -241,13 +241,31 @@ class StixCoreRelationshipContainer extends Component {
...
+            <Typography variant="h4" gutterBottom={false} sx={{ float: 'left' }}>
               {t('Relationship')}{stixCoreRelationship.draftVersion && (<DraftChip/>)}
             </Typography>
+            {!stixCoreRelationship.is_inferred && (
+            <Security needs={[KNOWLEDGE_KNUPDATE]}>
+              <IconButton
+                color="primary"
+                onClick={this.handleOpenEdition.bind(this)}
+                size="large"
+                sx={{ margin: '-15px 0px 0px -2px' }}
```
natalie-natsu, line 253: This is ugly, but the all grid use highly unstable css with float and all. ^^ I tried to put a simple `<Stack> `component and it was breaking all layout 💥

opencti-platform/opencti-front/src/private/components/common/stix_core_relationships/StixCoreRelationshipOverview.jsx
```diff
@@ -241,13 +241,31 @@ class StixCoreRelationshipContainer extends Component {
...
+                handleClose={this.handleCloseEdition.bind(this)}
+                handleDelete={this.handleOpenDelete.bind(this)}
+              />
+            </Security>
+            )}
             <Paper
               classes={{ root: classes.paperWithoutPadding }}
               variant="outlined"
-              style={{ position: 'relative' }}
+              style={{ position: 'relative', marginTop: '-4px' }}
```
natalie-natsu, line 268: Ugly because there is somewhere another inline style overriding every `sx` props we put on the component.

opencti-platform/opencti-front/src/private/components/analyses/groupings/GroupingEditionOverview.jsx
```diff
@@ -289,7 +289,7 @@ const GroupingEditionOverviewComponent = (props) => {
               setFieldValue={setFieldValue}
               onChange={editor.changeMarking}
             />
-            <div style={{ display: 'flex', justifyContent: 'space-between', flex: 1 }}>
+            <div style={{ display: 'flex', justifyContent: 'flex-end', flex: 1, gap: '16px' }}>
```
lndrtrbn: Instead of adding `16px` use `theme.spacing(2)` which is multiple of 8 (same for all other components in the PR)

opencti-platform/opencti-front/src/private/components/profile/ProfileOverview.jsx
```diff
@@ -486,6 +490,16 @@ const ProfileOverviewComponent = (props) => {
           )}
         </Formik>
       </Paper>
+      { hasKnowledgeAccess ? (
+        <Paper classes={{ root: classes.paper }} variant="outlined">
+          <Typography variant="h1" gutterBottom={true}>
+            {t('Dashboard settings')}
+          </Typography>
+          <Box py="20px">
```
lndrtrbn, line 498: I don't think we use `Box` coupled to `sy` or `sx` props to make spacing elsewhere is the code but why not 👍 Just, can we use a multiple of 8 as spacing ? Or does it break the page layout ?

opencti-platform/opencti-front/src/private/components/analyses/groupings/GroupingEditionOverview.jsx
natalie-natsu: Of course, I wasn't sure if the theme was used or nor. But I agree, better use the theming system. :-)

opencti-platform/opencti-front/src/private/components/profile/ProfileOverview.jsx
natalie-natsu: I can change for a `py={2.5}` but if the theme change for any reason, yeah it breaks layout. I wouldn't try for this one ^^

lndrtrbn (approved):

PR #11119, opened 2025-05-26
[backend] Language entity type should not be exposed in UI (#10698)

### LINKED ISSUE #10698: fix(filters): language entity type filter should not exist
## Description
'Language' should not be present in the possible values for the entity type filter

## Reproducible Steps
Go under Data> Entities
Add an 'entity type' filter.
In the possible values, we should not have 'Language' since it's not an entity type in the platform : 

Image

Archidoit:
### Proposed changes
Language entity type should not be exposed in the UI
- in the entity type filter values list
- in Customization > Entity types

### Related issues
https://github.com/OpenCTI-Platform/opencti/issues/10698

### DIFF
--- a/opencti-platform/opencti-graphql/src/domain/subType.ts
+++ b/opencti-platform/opencti-graphql/src/domain/subType.ts
@@ -11,13 +11,17 @@ import { ENTITY_HASHED_OBSERVABLE_ARTIFACT } from '../schema/stixCyberObservable
 import { telemetry } from '../config/tracing';
 import type { AuthContext, AuthUser } from '../types/user';
 import { ENTITY_TYPE_EXTERNAL_REFERENCE } from '../schema/stixMetaObject';
+import { ENTITY_TYPE_LANGUAGE } from '../modules/language/language-types';
+
+const NOT_EXPOSED_TYPES = [ENTITY_TYPE_LANGUAGE];
 
 // -- ENTITY TYPES --
 
 export const queryDefaultSubTypes = async (context: AuthContext, user: AuthUser, search : string | null = null) => {
   const queryDefaultSubTypesFn = async () => {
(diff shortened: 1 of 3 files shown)


opencti-platform/opencti-graphql/src/domain/subType.ts
```diff
@@ -11,13 +11,17 @@ import { ENTITY_HASHED_OBSERVABLE_ARTIFACT } from '../schema/stixCyberObservable
 import { telemetry } from '../config/tracing';
 import type { AuthContext, AuthUser } from '../types/user';
 import { ENTITY_TYPE_EXTERNAL_REFERENCE } from '../schema/stixMetaObject';
+import { ENTITY_TYPE_LANGUAGE } from '../modules/language/language-types';
+
+const NOT_EXPOSED_TYPES = [ENTITY_TYPE_LANGUAGE];
```
lndrtrbn, line 16: As far as I know there are already some entity types that are not listed in the Customization page. So I am wondering why we need a new logic `NOT_EXPOSED_TYPES` for the specific case of Language.

Don't we already have a logic somewhere that does the same thing?

SouadHadjiat (commented):
I don't feel confortable with hiding the entity type "Language" in the filters and customization, knowing that we can create it using graphql API, and it will be listed in Data -> entities.

Archidoit:
@romain-filigran Indeed we can create Language entities via the API, and they will be listed in Data>Entities. So maybe we want to keep the Language entity type filter?
image


Do we also want to keep Language in Customization? (if yes, there is no bug :)

Archidoit:
Note: If I click on the Language line, I'm redirected to a not-existing page : 

image

Archidoit:
Closed because the scope has changed: we want to add Language in Entities.

Issue #4948, opened 2023-11-22
Be able to overlay two ATT&CK matrices


nino-filigran:
## Use case

Compare the TTPs used by an actor/malware and the TTPs covered by an Organisation/security platform

## Current Workaround

None. 

## Additional Information

Feature to refine. 

## If the feature request is approved, would you be willing to submit a PR?

Yes / No (Help can be provided if you need assistance submitting a PR)

alice-debra:
https://github.com/OpenCTI-Platform/opencti/issues/3007

PR #11126, opened 2025-05-26
[frontend] Be able to overlay multiple attack pattern matrices (#4948)

delemaf:
### Proposed changes

* New select security posture
* Able to compare security platform entity matrix to threat
* Hide only covered attack pattern

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/4948

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/common/form/EntitySelect.tsx
+++ b/opencti-platform/opencti-front/src/private/components/common/form/EntitySelect.tsx
@@ -67,10 +67,6 @@ const EntitySelectComponent = ({
     type: sco.node.entity_type,
   }));
 
-  const handleChange = (newValue: EntityOption | EntityOption[] | null) => {
-    onChange?.(newValue);
-  };
-
   return (
     <Autocomplete
       value={value}
@@ -80,7 +76,7 @@ const EntitySelectComponent = ({
       noOptionsText={t_i18n('No available options')}
       isOptionEqualToValue={(option, val) => option.value === val.value}
       onInputChange={(_, val) => throttleSearch(val)}
-      onChange={(_, val) => handleChange(val)}
+      onChange={(_, val) => onChange?.(val)}
       renderInput={(params) => (
         <TextField
           {...params}
--- a/opencti-platform/opencti-front/src/private/components/common/form/EntitySelect.tsx
+++ b/opencti-platform/opencti-front/src/private/components/common/form/EntitySelect.tsx
@@ -1,4 +1,4 @@
-import { Autocomplete, Chip, TextField, TextFieldProps, TextFieldVariants } from '@mui/material';
+import { Autocomplete, Checkbox, Chip, TextField, TextFieldProps, TextFieldVariants } from '@mui/material';
 import React, { Suspense, useMemo, useState } from 'react';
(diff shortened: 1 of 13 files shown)


opencti-platform/opencti-front/src/private/components/common/form/EntitySelect.tsx
```diff
@@ -57,37 +67,63 @@ const EntitySelectComponent = ({
     type: sco.node.entity_type,
   }));
 
+  const handleChange = (newValue: EntityOption | EntityOption[] | null) => {
+    onChange?.(newValue);
+  };
```
lndrtrbn: Can we not use directly onChange instead of this function?

opencti-platform/opencti-front/src/private/components/common/stix_domain_objects/StixDomainObjectAttackPatternsKillChain.tsx
```diff
@@ -123,6 +150,61 @@ const StixDomainObjectAttackPatternsKillChain: FunctionComponent<StixDomainObjec
...
+        },
+      },
+    ).toPromise() as StixDomainObjectAttackPatternsKillChainOverlapQuery$data;
+  };
+
+  useEffect(() => {
+    if (selectedSecurityPlatforms.length > 0) {
+      getAttackPatternIdsToOverlap(selectedSecurityPlatforms.map(({ value }) => value))
+        .then((result) => {
+          setAttackPatternIdsToOverlap(result?.stixCoreObjects?.edges?.map(({ node }: StixDomainObjectAttackPatternsKillChainOverlapEdges) => node.id));
```
lndrtrbn: as a global rule, we prefer using async await syntax

delemaf: I also changed the way we fetch data using a callback function instead of useEffect hook :+1:

SouadHadjiat (approved):

opencti-platform/opencti-front/src/private/components/common/stix_domain_objects/StixDomainObjectAttackPatternsKillChain.tsx
```diff
@@ -123,6 +148,58 @@ const StixDomainObjectAttackPatternsKillChain: FunctionComponent<StixDomainObjec
     setSelectedKillChain(event.target.value as string);
   };
 
+  const getAttackPatternIdsToOverlap = async (entityIdsToOverlap: string[]) => {
+    if (entityIdsToOverlap.length === 0) return undefined;
+
+    const { stixCoreObjects } = await fetchQuery(
+      stixDomainObjectAttackPatternsKillChainOverlapQuery,
+      {
+        count: 1000,
```
SouadHadjiat, line 157: so we have a limitation of 1000 attack patterns covered by security platforms that can be displayed in the matrix (for overlap) ?

delemaf: We could increase limitation, it's just that I have to give a count to  the query

PR #11157, opened 2025-05-28
Implemented automatic Ioc de-sanitizing before performing searches

### LINKED ISSUE #11213: Implemented automatic Ioc de-sanitizing/defang before performing searches
## Use case

implemented auto "un-sanitizing" when performing a search in the top bar, pretty much like what can be seen when using VirusTotal's search bar.

That way, a search like "hxxp[:][//]domain[.]tld/admin[@]filigran[.]io" will be automatically converted in the frontend to http://domain.tld/admin@filigran.io
This is specifically useful because someimes, searching for sanitized IoCs don't yield the correct indicator in the results

DreadFog:
### Proposed changes
implemented auto "un-sanitizing" when performing a search in the top bar, pretty much like what can be seen when using VirusTotal's search bar.

That way, a search like "**hxxp[:][//]domain[.]tld/admin[@]filigran[.]io**" will be automatically converted in the frontend to **http://domain.tld/admin@filigran.io**
This is specifically useful because someimes, searching for sanitized IoCs don't yield the correct indicator in the results

### Related issues

- closes https://github.com/OpenCTI-Platform/opencti/issues/11213

### Checklist

### Further comments
Maybe the "auto de-sanitize" functionnality could be activated/deactivated through the user settings, but this is not mandatory.


aHenryJard:
Thank you for your proposal @DreadFog, actually several people are interested by this feature so I opened an related issue in case there is some discussions and others examples.

opencti-platform/opencti-front/src/private/components/nav/TopBar.tsx
```diff
@@ -254,7 +263,7 @@ const TopBarComponent: FunctionComponent<TopBarProps> = ({
         },
       });
     } else {
-      handleSearchByKeyword(searchKeyword, 'knowledge', navigate);
+      handleSearchByKeyword(unsanitizedKeyword, 'knowledge', navigate); // Use unsanitized keyword
```
aHenryJard: Unsanitizing before sending search will not cover graphQL api direct usages and client python usage, this replacement must be done in the backend side. Propably at database level to benefit from score ordering on results.

Also, the search should be sent with both sanitize and unsanitize terms because it's possible that both exists in the platform.

PR #11178, opened 2025-06-02
[backend] don't unpack xlsx and docx files in artifact import (#10380)

### LINKED ISSUE #10380: OpenCTI Unpacks Office Files Instead of Preserving The File
## Description

When uploading an Office document, such as .docx and .xlsx, as an artifact object, OpenCTI interprets it as an archive file type, unpacks it, grabs the first file (“[file_name].xml”), and then stores that file with the appropriate hashes in OCTI. 

## Environment

SaaS 6.5.9

## Reproducible Steps

1. Navigate to the Artifacts tab under Observations
2. Create a new artifact and upload an Office file.

## Expected Output

The artifact is the Office File.

## Actual Output

The artifact is the [Content Types].xml of the Office file.

JeremyCloarec:
### Proposed changes

*  don't unpack xlsx and docx files in artifact import
*

### Related issues

* #10380 
*

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/domain/stixCyberObservable.js
+++ b/opencti-platform/opencti-graphql/src/domain/stixCyberObservable.js
@@ -382,8 +382,8 @@ export const artifactImport = async (context, user, args) => {
   let resolvedFile = await file;
   // Checking infected ZIP files
 
-  if (file.mimeType !== 'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet'
-      && file.mimeType !== 'application/vnd.openxmlformats-officedocument.wordprocessingml.document') {
+  if (resolvedFile.mimetype !== 'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet'
+      && resolvedFile.mimetype !== 'application/vnd.openxmlformats-officedocument.wordprocessingml.document') {
     try {
       resolvedFile = await extractInfectedZipFile(resolvedFile);
     } catch {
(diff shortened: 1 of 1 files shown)


opencti-platform/opencti-graphql/src/domain/stixCyberObservable.js
```diff
@@ -381,11 +381,16 @@ export const artifactImport = async (context, user, args) => {
   const { file, x_opencti_description: description, createdBy, objectMarking, objectLabel } = args;
   let resolvedFile = await file;
   // Checking infected ZIP files
-  try {
-    resolvedFile = await extractInfectedZipFile(resolvedFile);
-  } catch {
-    // do nothing
+
+  if (resolvedFile.mimetype !== 'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet'
```
aHenryJard: I wonder about doing the opposite: extract only explicit zip file. I think there are other compress file types like docx (pptx ? and maybe similar openoffice files also).

JeremyCloarec: I thought about doing the opposite also, but I'm not super familiar with this part of the code: I'm not sure why we are trying to extract all files as archives in the first place? But I didn't want to change the current behavior, thinking that it was there for a reason.
You're right about pptx and openoffice files, I'll try it

JeremyCloarec: I discussed it with @alice-debra, and I created a new issue to discuss further improvements that could be implemented https://github.com/OpenCTI-Platform/opencti/issues/11228

aHenryJard (approved):

PR #11219, opened 2025-06-05
Unique on Users in the Dashboard Active Users Widget

daimyo007:
### Proposed changes

* Adding option to display just the number of unique users on the dashboard

### Checklist


nino-filigran:
@daimyo007, thanks for your contribution! I remember the conversation we had around the need, so this part is pretty clear.

However, in the current implementation, we've noticed few issues:
- The first problem is that the results of a distribution list are limited to 100. Therefore, if you have more than 100 users, they would be excluded from this count.
- The second problem is that if you restrict the result to let's say 2, it will only display the first entries: if the first results that you get are two actions for the same user, you will get only 1 in your unique count, while you would expect 2.
- The distribution list is, as its name says, a list of distribution and not a number (I understand however, that this was the only way for you to get this result, since only the distribution list would allow you to get the list of active users on the first hand).

We have tried to describe in this issue, the way we would like to tackle this topic: https://github.com/OpenCTI-Platform/opencti/issues/11229 

Please let me know you thoughts, if you feel like able to do it or not, and if it make sense to you.

Anyway, thanks again for pushing this contribution!

daimyo007:
modified and moved into [PR 11511](https://github.com/OpenCTI-Platform/opencti/pull/11511)

PR #11241, opened 2025-06-09
[frontend] don't show subscribers number when there are no ones (#10985)

Archidoit:
Don't show the 'xx subscribers' when there are no subscribers

### Related issues
https://github.com/OpenCTI-Platform/opencti/issues/10985


opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectSubscribers.tsx
```diff
@@ -53,14 +53,18 @@ const StixCoreObjectSubscribers: FunctionComponent<ContainerHeaderSharedProps> =
 
   const [{ triggersKnowledge, triggersKnowledgeCount }] = useRefetchableFragment<TriggerQuery, FragmentKey>(stixCoreObjectTriggersFragment, triggerData);
 
+  if (!triggersKnowledgeCount || triggersKnowledgeCount === 0) {
+    return (<></>);
```
ValentinBouzinFiligran: I think you can just : `return </>` or `return </div>` ?

Archidoit: done :)

ValentinBouzinFiligran (approved):

Issue #10452, opened 2025-04-02
Create security platform entity (relation & import/export python included)


nino-filigran:
## Use case

Goal: be able to declare a theoretical security coverage, import it, export it in STIX & create a relation with Attack Pattern.

PR #11257, opened 2025-06-10
[backend/frontend] Fixes on Security Platforms (#10452)

Gwendoline-FAVRE-FELIX:
### Proposed changes

* Remove aliases
* Fix types on Security Platforms list
* Forgotten parenthesis
* Fix Organization sharing
* Fix notes
* Fix update types form
* Fix name, icon in graph
* Fix hidden type
* Fix name in investigation

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/10452


ValentinBouzinFiligran:
When I update the type, it does not appear on my drawer : 

image

Archidoit:
Working well :) 
Small other fix eventyally : Security platform is the only display in singular in the right menu : 
image


image

Archidoit:
in other entities overview, the type is displayed with this design  (example: report, malwares, etc) : 
image

image


Is it normal it is different for security platform?
image

Archidoit:
Security platform is the only entity for which the Notes section is only displayed in half of the screen (instead of the whole screen width)
image

lndrtrbn:
We can create a Note from SecurityPlatform overview, when going to the note page, the app crashes

Archidoit:
If I add an external reference to a Security platform, and then click on the external ref, an error occurs instead of redirecting to the external ref: 
image

image

Archidoit:
I don't see Security Platform under the Platform hidden types list (Parameters > Hidden entity types)

image


image

lndrtrbn:
In graph, it misses the icon of Security Platforms (I added a security platform in a report)
image

Archidoit:
In the 'Add entities' panel of an investigation, security platforms are not well displayed:
- the name is always displayed as Unknown
- we should have '-' if there are no authors

image

It is also displayed as Unknown in the Details tab
image

Archidoit:
The security platform name is not correctly displayed in the overview of a relationship involving a security platform (relationship_SecurityPlatform instead of its name)
image

Archidoit:
It's not possible to update the processing status of a Security platform, even if Workflow is enabled in the Customization (no 'status' field in the edition form)
image

opencti-platform/opencti-front/src/components/dataGrid/dataTableUtils.tsx
```diff
@@ -1033,7 +1033,17 @@ const defaultColumns: DataTableProps['dataColumns'] = {
     percentWidth: 10,
     isSortable: false,
     render: ({ secondary_motivations }) => {
-      const value = secondary_motivations ? secondary_motivations.join(', ') : '-';
+      const value = secondary_motivations || '-';
```
SouadHadjiat: why this change on secondary_motivations ?

lndrtrbn:
image
image

It's not the same icon used for the entity type in graphs

lndrtrbn (approved):

Issue #11231, opened 2025-06-06
Error at Audit list widget creation


Archidoit:
## Description
All the Audit list widgets have errors

![Image](https://github.com/user-attachments/assets/ed8f67b8-9906-4486-8053-3b590712318d)

![Image](https://github.com/user-attachments/assets/5b262d26-47dd-430e-b536-27dbe8a8c98a)

## Reproducible Steps
Create a widget in a dashboard with:
- visualization = list
- perspective = audit
- filters: no filters, or what you want

The generated widget is displayed with an error, whereas the query is correctly done (frontend error).

PR #11298, opened 2025-06-12
[frontend/backend] Fix Audits list widgets errors (#11231)

Archidoit:
### Proposed changes
- Fix error at audits list display
- Take into account the ordering mode in audits list
- Audit List component in tsx

### Related issues
https://github.com/OpenCTI-Platform/opencti/issues/11231


lndrtrbn (commented):
Maybe take example as other reworked widgets for the structure. Otherwise works well

Archidoit:
> Maybe take example as other reworked widgets for the structure. Otherwise works well

done :)

opencti-platform/opencti-front/src/private/components/common/audits/AuditsList.tsx
```diff
@@ -0,0 +1,167 @@
...
+  startDate,
+  endDate,
+  dataSelection,
+  parameters,
+}) => {
+  const { t_i18n } = useFormatter();
+  const isGrantedToSettings = useGranted([SETTINGS_SETACCESSES, SETTINGS_SECURITYACTIVITY, VIRTUAL_ORGANIZATION_ADMIN]);
+  const isEnterpriseEdition = useEnterpriseEdition();
+
+  const renderContent = () => {
```
lndrtrbn: You do not need this function, just put the code at the root of the component. Moreover, hooks should be called at the top of a component, before any conditions. So move the queryLoading before any ifs (https://react.dev/reference/rules/rules-of-hooks)

Archidoit: done :)

lndrtrbn (approved):

Issue #11273, opened 2025-06-11
[Groups] Group ids are not identified in activity monitoring


NiQuintin:
## Description

Groups actions are not identified in the activity section. It makes it impossible to filter by group in a widget or directly in activity/events. (NB: it's working for the users of the group).

## Environment

OpenCTI version: 6.6.15

## Reproducible Steps

Steps to create the smallest reproducible scenario:

1. Take a group ID and look for it activity/events

## Expected Output

We should see activities related to group members

## Actual Output

Nothing is displayed

## Additional information

see .gif 👇

## Screenshots (optional)

![Image](https://github.com/user-attachments/assets/dd12862a-83a2-4087-a4b2-901ee45fca53)

jw-NYC-001:
@NiQuintin - I noticed that Group IDs are still not searchable in Activity Section in v6.7.17 (we are currently testing for upgrade). Can you please provide more details on if this issues is being solved in a different ticket?

PR #11313, opened 2025-06-13
[backend] removing too heavy cleanup migration script (#11273)

labo-flg:
With the fix #11275 for #11273 , we introduced a migration script to clean up activity logs.
This migration is not mandatory for the platform to operate, it is a matter of having consistent activity logs that were missing some data.

However in very large platforms, this migration can be very long to run.
We are removing the migration script for the time being, and will address this operation in a non-blocking manner.


JeremyCloarec:
We'll have to be careful to not push a migration with the same timestamp if we delete this one.
This shouldn't be that much of a problem though if release is merged into master soon, since it will have migrations with higher timestamps

JeremyCloarec (approved):

labo-flg:
I'm keeping the migration file to avoid any timestamp issue indeed. It will simply do nothing.

Kedae (approved):

Issue #7574, opened 2024-07-03
Filters are back on the right handside for containers 


nino-filigran:
## Description

In containers, on entity & observable tabs, we have the right panel to filter which is back. We don't want it. Instead, we should use the classic filters entity type = , already selected

Archidoit:
Not a regression since it was present before. The removing of all the entity types filter right panels was deprioritized, so only some of them have been removed for the moment.

nino-filigran:
I've switched it to feature. If it's a tiny, then I'll push in 6.4

PR #11316, opened 2025-06-13
[frontend] Remove entity type right bar in containers (#7574)

Archidoit:
### Proposed changes
remove entity type right bar in Entities and Observables tabs of Containers 
The user should use the entity type filter instead

### Related issues
https://github.com/OpenCTI-Platform/opencti/issues/7574


lndrtrbn (commented):
navigation e2e test seems to be in error

Archidoit:
> navigation e2e test seems to be in error

it's fixed :)

lndrtrbn (approved):

PR #11335, opened 2025-06-14
[backend/frontend] Introduce dynamic filters for relationships and in regards of (#11357)

### LINKED ISSUE #11357: Introduce dynamic filters for relationships and in regards of
## Use case

Introduce dynamic filters for relationships and in regards of

richard-julien:
Introduce dynamic filters for relationships and in regards of

Closes #11357.


opencti-platform/opencti-front/src/utils/tests/FilterUtilsConstants.ts
```diff
@@ -662,6 +662,39 @@ filterKeysSchema.set('Stix-Core-Object', new Map([
       ],
     },
   ],
+  [
+    'dynamicRegardingOf',
+    {
+      filterKey: 'regardingOf',
```
Archidoit:
```suggestion
      filterKey: 'dynamicRegardingOf',
```

Archidoit:
'no label' appears as 'label = deleted' in a dynamic regarding of filter

image

Archidoit:
If I create a dynamic regarding of filter, than remove all the values in it, I have an error

image


image

it's because the filter looks this way instead of being undefined:
image

Archidoit:
With regardingOf, we only support the EQ operator, and it's the same for dynamic regardingOf (not working with NOT_EQ). So we should only propose the EQ operator in the UI and throw an error via the API if an other operator is used for dynamic regarding Of

image

Archidoit:
If I create a dynamic regarding Of filter with only relationship types, and then click on the filter to edit it : frontend error
image

image

opencti-platform/opencti-front/src/utils/tests/FilterUtilsConstants.ts
SamuelHassine: Done

SamuelHassine:
> 'no label' appears as 'label = deleted' in a dynamic regarding of filter
> 
> image

Fixed

Issue #10284, opened 2025-03-19
Add platform logo in first page of a Fintel Template


nino-filigran:
## Use case

Be able to replace OCTI logo by the platform logo in the template.

PR #11356, opened 2025-06-16
[frontend] fintel design available for built in html to pdf connector (#10284)

Archidoit:
### Proposed changes
In the content tab of a container, when doing a file export with the connector 'html file to pdf' : the 'fintel design' field should appear if the selected file is originated from a fintel template

### Related issues
https://github.com/OpenCTI-Platform/opencti/issues/10284


labo-flg (approved):

labo-flg:
tested ok

PR #11279, opened 2025-06-11
[backend/frontend] Auto creation of users on feed ingestion (#11047)

hervyt:
### Proposed changes

* Automatically create a user when option is selected on CSV Feed ingestion

Counter updates:
- (2 times) when adding a user to an org, there is a relation "participate-to" in stream (same as user modified with UI in settings)

### Related issues

* #11047

### Checklist

- [X] I consider the submitted work as finished
- [X] I tested the code for its functionality
- [X] I wrote test cases for the relevant uses case (coverage and e2e)
- [ ] I added/update the relevant documentation (either on github or on notion)
- [ ] Where necessary I refactored code to improve the overall quality

Issue #11047, opened 2025-05-19
Ability to create a user of a CSV Feed on the fly


EllynBsc:
## Use case

Ability to create a user of a CSV Feed on the fly 

## Context 

The automatic creation of users for the CSV feeds would require specific configuration to assign the users to the correct group, which is a recurring issue in Open CTI.

- **Automatic User Creation:** To automatically create users for the feeds, the platform must be configured to assign them to the correct group. This requires specific setup to ensure proper user assignment.
- **Permission Issues:** Managing permissions and user groups is a recurring issue in Open CTI. This functionality is essential to ensure data security and compliance

PR #11370, opened 2025-06-17
[backend/frontend]-(AutoIngestionUser) call existing user only on auto_user (#11047)

hervyt:
### Proposed changes
outside of FF:
* bug fix usage of SYSTEM user in CSV feed

Under CSV_FEED FF
* add condition before calling backend to check user existence

### Related issues

* #11279 

### Checklist


opencti-platform/opencti-front/src/private/components/data/ingestionCsv/IngestionCsvCreation.tsx
```diff
@@ -228,26 +228,29 @@ const IngestionCsvCreation: FunctionComponent<IngestionCsvCreationProps> = ({ pa
...
   ) => {
     // Check if user does not already exist.
-    const existingUsers = await fetchQuery(ingestionCsvCreationUsersQuery, { filters: {
-      mode: 'and',
-      filters: [
-        {
-          key: ['name'],
-          values: [
-            (values.user_id as FieldOption).value,
+    if (values.automatic_user !== false) {
```
aHenryJard: this if should be only if feature flag CSV_FEED is enabled no ?

hervyt: No, we need to have the switch automatic_user enabled too, otherwise, we request a user with name "null" (if we add no user for SYSTEM user cases :)  
And if we select an already existing user, the request is useless, because of course this user already exists

opencti-platform/opencti-graphql/src/modules/ingestion/ingestion-csv-domain.ts
```diff
@@ -65,8 +57,19 @@ export const defaultIngestionGroupsCount = async (context: AuthContext) => {
   return defaultGroupLength.length ?? 0;
 };
 
-export const userAlreadyExists = async (context: AuthContext, args: QueryUserAlreadyExistsArgs) => {
-  const users = await findAll(context, context.user, args) as UserConnection;
+export const userAlreadyExists = async (context: AuthContext, name: string) => {
+  // We use SYSTEM_USER because manage ingestion should be enough to create an ingestion Feed
+  const users: BasicStoreCommon[] = await findAll(context, SYSTEM_USER, {
```
aHenryJard: Could you rename to =>
```suggestion
  const users: BasicStoreCommon[] = await findAllUser(context, SYSTEM_USER, {
```

aHenryJard (dismissed):
All good to me

aHenryJard (approved):
All good to me

PR #11371, opened 2025-06-17
[frontend] fixes for dynamic regardingOf filter (#11357)

Archidoit:
Bug following this PR : https://github.com/OpenCTI-Platform/opencti/pull/11335

## Bug 1
don't display 'Dynamic filter' if there is no dynamic filter selected
image

image

## Bug 2
fix : when editing a dynamic filter with only a relationship type, the screen crashes
image

## Bug 3
fix: when we add only a dynamic filter, then remove it, the screen crashes

## Bug 4
add a tooltip for 'dynamic filter' content

- before: 
image

- after: 
image

## Bug 5
Do a query background task with a dynamic regarding of filter
image

In the opened pop up, click on the orange chip
image

An error occurs 
image

## Bug 6

If the filter in 'dynamic' is targeting too many ids, the list of ids put in the final query is too long and the screen crashes
image

### DIFF
--- a/opencti-platform/opencti-front/src/components/filters/DisplayFilterGroup.tsx
+++ b/opencti-platform/opencti-front/src/components/filters/DisplayFilterGroup.tsx
@@ -13,7 +13,6 @@ import Chip from '@mui/material/Chip';
 import { useFormatter } from '../i18n';
 import { FilterRepresentative } from './FiltersModel';
 import { Filter, FilterGroup } from '../../utils/filters/filtersHelpers-types';
-import TasksFilterValueContainer from '../TasksFilterValueContainer';
 
 interface DisplayFiltersValuesProps {
   filtersRepresentativesMap: Map<string, FilterRepresentative>,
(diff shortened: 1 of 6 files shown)


CelineSebe:
image

It seems that we can choose 'In regard of' ad infinitum

CelineSebe:
The fixes works well 👌

CelineSebe:
the Dynamic filter tooltip is not the same size, it's too big:

image

the others filters:
image

image

Archidoit:
@CelineSebe  It's because in the second case, it's more complex: it's not a tooltip to see the a single filter part, it's a tooltip to see an inner filter group that canc contains several filters : 
image

Archidoit:
> image
> 
> It seems that we can choose 'In regard of' ad infinitum

@CelineSebe Thanks for pointing that. Indeed it's not really nice, but it's not a blocking thing according to @nino-filigran :)

aHenryJard (approved):
LGTM

CelineSebe (approved):

Archidoit:
remaining fix to do : bug 6

PR #11373, opened 2025-06-17
[frontend] Add in container popup has broken search (#11265)

### LINKED ISSUE #11265: Add in container popup has broken search
## Description

"Add in container" search box in the pop up seems broken, does not handle update after changing search terms.
## Environment

OCTI 6.6.15

## Reproducible Steps

1. Go into any entity, like a malware
2. click on "add in container quick action" in the header
3. try to search for a specific container name
4. list is not updated, still show unrelated items

## Expected Output

List is filtered with the search terms

## Actual Output

list does not change

## Additional information

As seen in video below, if you enter invalid search (search with 0 result) the list is updated. Also (end of the video) if you hit enter key, it seems like the right item was selected somehow in the list.

## Screenshots (optional)

https://github.com/user-attachments/assets/f7202d0d-dc10-4445-a338-a38da6de2e10

ValentinBouzinFiligran:
### Proposed changes

* StixCoreObjectContainer updated from JSX to TSX
* component refactored to be more efficient

### Related issues

* #11265 

### Checklist


lndrtrbn (commented):
From what I see it cannot work, but maybe I am missing something.

The list of containers is only fetched at component init and that's it. But The query made to fetch containers is paginated so potentially, if you have a lot of containers, you won't have all the containers retrieved by the query. So if you filter with the text input, you will filter only the first page of containers, and potentially miss some results.

The search query needs to be done also when the search input changes IMO.

ValentinBouzinFiligran:
ok, I'll put this behavior back
I'll set a timer so as not to spam while writing

ValentinBouzinFiligran:
⚠️ there is a little difference between the search value and the result: I think the result contains rows with the search value that are in their data but only the representative is displayed

Here,it seems that it is inconsistent, but the search value "azea" is contain on the feedback description

image

image

lndrtrbn (approved):

PR #11394, opened 2025-06-18
[frontend] Improve managing of widgets positioning in Dashboards (#10730)

### LINKED ISSUE #10730: Dashboard widget tiling issue
## Description
When I add a widget on a custom dashboard, it changes the general display of the dashboards by moving everything to the right side.

## Environment
1. 6.4.5 SaaS

## Reproducible Steps
1. Start to add widgets and align as desired, some centred.
2. Add new widgets, layout is not respected and often moves previous widgets to the right.
3. If you have a full dashboard and add another widget, it can move all widgets to the right, in a vertical column.

## Expected Output
Adding new widgets to not displace previous and add them below others, or in available free space

## Actual Output
Bunches up and moves existing widgets

## Screenshots (optional)
Image
Image

lndrtrbn:
### Proposed changes

* Refactor some code
* Always add a new widget in a new line

### Related issues

* #10730 

### Checklist

### Further comments

The code might seem a bit over-engineered, the issue we face is that we save in the same object the config for the Chart that is displayed inside the grid and the config of the grid position itself. But for performance issue we do not want to recompute all charts when we just moved or resized an element in the grid. So we need to separate the case of updating the widget and the case of modifying the grid. That's why we have a local state containing the grid layout that complicate a bit the code.

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/workspaces/dashboards/Dashboard.jsx
+++ b/opencti-platform/opencti-front/src/private/components/workspaces/dashboards/Dashboard.jsx
@@ -1,6 +1,6 @@
 import React, { useEffect, useMemo, useState } from 'react';
 import * as R from 'ramda';
-import { createFragmentContainer, graphql } from 'react-relay';
+import { graphql, useFragment } from 'react-relay';
 import RGL, { WidthProvider } from 'react-grid-layout';
 import Paper from '@mui/material/Paper';
 import { v4 as uuid } from 'uuid';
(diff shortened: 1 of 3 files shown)


Archidoit:
Working well for the fix !! :)

Archidoit:
A small problem is remaining: the last added widget is resized to its default size when we reload the page or leave and come back to the dashbaord.

Exemple:
dashboard 1: 
image


After reload:
image

opencti-platform/opencti-front/src/private/components/workspaces/dashboards/Dashboard.jsx
```diff
@@ -108,39 +159,23 @@ const DashboardComponent = ({ workspace, noToolbar }) => {
...
         ...manifest.widgets,
-        [widgetManifest.id]: {
-          ...widgetManifest,
-          layout: { i: widgetManifest.id, x: maxX, y: maxY, w: 4, h: 2 },
+        [widgetConfig.id]: {
+          ...widgetConfig,
+          layout: {
+            i: widgetConfig.id,
+            x: 0,
+            y: 1000, // 1000 will be replaced automatically by a new row at the end of existing ones.
```
lndrtrbn: In fact not sure it's a good idea, will go back to a computation

delemaf (approved):
Need a little rebase but works for me :+1:

Issue #11391, opened 2025-06-18
Improve retention manager execution


richard-julien:
## Use case

Improve retention manager execution
* Allow manager to stop correctly when platform needs to shutdown (remove BluePromise)
* Limit the number of parrallelisation by default
* Allow deletion to be done without a shards direct synchronization for performance
* Increase bulk timeout and number of retry
* Bulk not throwing anymore exception for document_missing_exception

PR #11395, opened 2025-06-18
[backend] Improve retention manager execution (#11391)

richard-julien:
See #11391


opencti-platform/opencti-graphql/src/database/engine.js
```diff
@@ -208,10 +208,11 @@ const ES_PRIMARY_SHARD_SIZE = conf.get('elasticsearch:max_primary_shard_size') |
 const ES_MAX_DOCS = conf.get('elasticsearch:max_docs') || 75000000;
 
 const TOO_MANY_CLAUSES = 'too_many_nested_clauses';
-export const BULK_TIMEOUT = '5m';
+const DOCUMENT_MISSING_EXCEPTION = 'document_missing_exception';
+export const ES_RETRY_ON_CONFLICT = 30;
```
SouadHadjiat, line 212: 30 retries ? isn't it too much ? this parameter is used in multiple places, not just for delete

richard-julien: I dont think its too much. In case of conflicts, we really want to finally be capable to apply the operation.

Issue #11378, opened 2025-06-17
No Label filter displayed as deleted


Archidoit:
## Reproducible Steps
Add a filter : label = No label
The filter is displayed as: label= deleted (instead of : label is empty)

![Image](https://github.com/user-attachments/assets/e03cff8f-4618-4caf-bfb7-33dafb66b49a)

![Image](https://github.com/user-attachments/assets/a4ba1086-5dce-44d9-9710-2f60bab61321)

PR #11400, opened 2025-06-18
[frontend] No label filter should be displayed as 'label is empty' (#11378)

Archidoit:
### Proposed changes
In entities list, when a user add a filter via 'label = No label', the filter added should be 'label is empty' (instead of label = deleted)

Same logic when the user add a filter via 'label != No label', the filter added should be 'label is not empty'.

### Related issues
https://github.com/OpenCTI-Platform/opencti/issues/11378


SarahBocognano (approved):

SarahBocognano:
Need rebase but otherwise it's ok :)

PR #11416, opened 2025-06-19
[backend] launch container sharing task with system user instead of context user (#11570)

### LINKED ISSUE #11570: Propagating container orga sharing to elements added in container is done with context user instead of internal user
## Description

Because createContainerSharingTask is done with context user instead of an internal bypass user, createContainerSharingTask can fail if the context user doesn't have the proper capabilities and/or organizations

## Environment

4. Other environment details:

## Reproducible Steps

1. { e.g. Run ... }
2. { e.g. Click ... }
3. { e.g. Error ... }

JeremyCloarec:
### Proposed changes

* launch container sharing task with system user instead of context user
*

### Related issues

* #11570 
*

### Checklist


opencti-platform/opencti-graphql/src/database/middleware.js
```diff
@@ -960,7 +960,7 @@ const createContainerSharingTask = (context, type, element, relations = []) => {
   if (targetGrantIds.length > 0) {
     const sharingDescription = `${type} organizations of ${element.name} to contained objects`;
     const input = { ids: targetGrantIds, scope: 'KNOWLEDGE', actions: [{ type, context: { values: elementGrants } }], description: sharingDescription };
-    taskPromise = createListTask(context, context.user, input);
+    taskPromise = createListTask(context, SYSTEM_USER, input);
```
aHenryJard: Could you please create a new AuthUser dedicated to this in https://github.com/OpenCTI-Platform/opencti/blob/master/opencti-platform/opencti-graphql/src/utils/access.ts

JeremyCloarec: Are you sure we want a dedicated internal user for this? We usually use the `SYSTEM_USER` in these kind of cases of data consistency (we use it in mergeEntitiesRaw for deleting old relations, in loadElementsWithDependencies to clean up invalid relations etc..)

aHenryJard (approved):
All good.

Issue #10902, opened 2025-05-02
AI Insights sometimes returns content in Markdown instead of expected HTML


romain-filigran:
## Description

The AI Insights feature intermittently returns output formatted in Markdown even when the expected format is HTML. 

![Image](https://github.com/user-attachments/assets/31483c2b-98d4-4f07-aa33-fa9ec4a8af4f)

## Environment

OpenCTI version: 6.6.9

delemaf:
@romain-filigran I need a little more context, on which platform and which entity (cmd.exe?) did you have this issue ?

SamuelHassine:
@delemaf @romain-filigran we need to prioritize this one please. Basically it is everywhere, go into a report, just click AI Insight. It is very painful for multiple customers and during demos. Thanks!

romain-filigran:
It's not linked to any particular entity. The problem seems to occur randomly wherever AI Insight is used.
For example on an intrusion-set, the output for the "Containers digest" could sometime return HTML or markdown formatted response. However, the prompts all seem to indicate that the answer must be in HTML.

SamuelHassine:
There are new functions to pass parameters to API to explicitely ask for an output format, we should use this I think.

PR #11444, opened 2025-06-20
[frontend/backend] force container summary to output HTML and handle empty entity (#10902)

delemaf:
### Proposed changes

* Force output to HTML
* Handle entity without any container
image

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/10902

### Checklist


richard-julien:
Sam was talking a specific option instead of a prompt instruction, do you check it?

delemaf:
> Sam was talking a specific option instead of a prompt instruction, do you check it?

I thought Sam was talking about the functions that use **longchain** (like the _queryNLQIa_ one) but the we can't use them without some modifications because they don't handle the streaming like _queryIa_. Plus after checking with data team, fixing prompt was the quickest way to handle this behavior without getting into a big refacto.

Issue #11445, opened 2025-06-20
[backend] Improve label filtering to support name instead of only id


richard-julien:
## Use case

To allow a simpler usage of the label filtering, allow to use direct label name instead of the id in the filtering.

PR #11448, opened 2025-06-21
[backend] Improve label filtering to support name instead of only id (#11445)

richard-julien:
see #11445

### DIFF
--- a/opencti-platform/opencti-graphql/src/utils/filtering/filtering-utils.ts
+++ b/opencti-platform/opencti-graphql/src/utils/filtering/filtering-utils.ts
@@ -13,6 +13,7 @@ import {
   filterKeysWithMeValue,
   INSTANCE_DYNAMIC_REGARDING_OF,
   INSTANCE_REGARDING_OF,
+  LABEL_FILTER,
   ME_FILTER_VALUE,
   MEMBERS_GROUP_FILTER,
   MEMBERS_ORGANIZATION_FILTER,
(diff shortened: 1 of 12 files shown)


opencti-platform/opencti-graphql/src/utils/filtering/filtering-utils.ts
```diff
@@ -409,13 +472,14 @@ export const checkAndConvertFilters = (
   // 01. check filters validity
   const { noFiltersChecking = false, noFiltersConvert = false } = opts;
   checkFiltersValidity(inputFilterGroup, noFiltersChecking);
-  // 02. replace dynamic @me value
-  const filterGroup = replaceMeValuesInFilters(inputFilterGroup, userId);
-  // 03. convert relation refs
+  // 02. If label filtered by name, try to resolve it.
+  const resolvedLabels: Record<string, string> = await computeFilterLabelMap(context, user, inputFilterGroup, idsFinder);
+  // 03. replace dynamic @me value
```
Archidoit:
```suggestion
  // 03. replace dynamic @me value and label values
```

opencti-platform/opencti-graphql/src/utils/filtering/filtering-utils.ts
```diff
@@ -330,10 +334,27 @@ const getConvertedRelationsNames = (relationNames: string[]) => {
...
+  });
+  filtersResult.filterGroups.forEach((fg) => {
+    const vals = extractFilterKeyValues(filterKey, fg);
+    values.push(...vals);
+  });
+  return values;
+};
+
 /**
  * Replace @me by the user id in filter whose values can contain user ids
```
Archidoit:
```suggestion
 * Replace @me by the user id in filter whose values can contain user ids, and replace eventual label values with label ids  
```

Archidoit (approved):

Issue #11447, opened 2025-06-20
[backend] Improve input resolve refs performance not fetching the denormalization


richard-julien:
## Use case

Improve input resolve refs performance not fetching the denormalization

PR #11449, opened 2025-06-21
[backend] Improve input resolve refs performance not fetching the denormalization #11447

richard-julien:
See #11447

### DIFF
--- a/opencti-platform/opencti-front/tests_e2e/navigation/navigation.spec.ts
+++ b/opencti-platform/opencti-front/tests_e2e/navigation/navigation.spec.ts
@@ -15,7 +15,7 @@ import ObservablesPage from '../model/observable.pageModel';
 import ObservableDetailsPage from '../model/observableDetails.pageModel';
 
 /**
- * Goal: validate that everything is opening wihtout errors in Analyses > Malware analyses.
+ * Goal: validate that everything is opening without errors in Analyses > Malware analyses.
  * @param page
  */
 const navigateMalwareAnalyses = async (page: Page) => {
@@ -36,7 +36,7 @@ const navigateMalwareAnalyses = async (page: Page) => {
   await page.getByLabel('relationships', { exact: true }).click();
   await expect(page.getByRole('link', { name: 'related to Malware Spelevo EK' })).toBeVisible();
   await page.getByLabel('entities', { exact: true }).click();
-  await expect(page.getByRole('link', { name: 'Malware Spelevo EK admin ryuk' })).toBeVisible();
+  await expect(page.getByRole('link', { name: 'Malware Spelevo EK admin covid-19' })).toBeVisible();
 
   // -- Content
   await malwareAnalysesDetailsPage.tabs.goToContentTab();
--- a/opencti-platform/opencti-graphql/src/database/engine.js
+++ b/opencti-platform/opencti-graphql/src/database/engine.js
@@ -1812,7 +1812,8 @@ export const elBatchIds = async (context, user, elements) => {
 export const elBatchIdsWithRelCount = async (context, user, elements) => {
   const ids = elements.map((e) => e.id);
   const types = elements.map((e) => e.type);
-  const hits = await elFindByIds(context, user, ids, { type: types, includeDeletedInDraft: true, relCount: true });
+  const opts = { type: types, includeDeletedInDraft: true, relCount: true, baseData: true };
+  const hits = await elFindByIds(context, user, ids, opts);
   return ids.map((id) => R.find((h) => h.internal_id === id, hits));
 };
 
--- a/opencti-platform/opencti-graphql/src/database/engine.js
+++ b/opencti-platform/opencti-graphql/src/database/engine.js
@@ -1804,7 +1804,13 @@ export const elLoadById = async (context, user, id, opts = {}) => {
 export const elBatchIds = async (context, user, elements) => {
   const ids = elements.map((e) => e.id);
   const types = elements.map((e) => e.type);
-  const hits = await elFindByIds(context, user, ids, { type: types, includeDeletedInDraft: true });
+  const allContextOverride = elements.map((e) => e.contextOverride).filter((c) => c);
+  let contextToUse = context;
+  if (allContextOverride.length > 0) {
+    const contextOverride = allContextOverride[0];
+    contextToUse = contextOverride;
+  }
+  const hits = await elFindByIds(contextToUse, user, ids, { type: types, includeDeletedInDraft: true });
   return ids.map((id) => R.find((h) => h.internal_id === id, hits));
 };
 
--- a/opencti-platform/opencti-graphql/src/database/engine.js
+++ b/opencti-platform/opencti-graphql/src/database/engine.js
@@ -1677,6 +1677,7 @@ const REL_COUNT_SCRIPT_FIELD = {
 export const elFindByIds = async (context, user, ids, opts = {}) => {
   const { indices, baseData = false, baseFields = [] } = opts;
   const { withoutRels = true, toMap = false, mapWithAllIds = false, type = null } = opts;
+  const { relCount = false } = opts;
   const { orderBy = null, orderMode = 'asc' } = opts;
   const idsArray = Array.isArray(ids) ? ids : [ids];
   const types = (Array.isArray(type) || isEmptyField(type)) ? type : [type];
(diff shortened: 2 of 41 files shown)


opencti-platform/opencti-front/tests_e2e/navigation/navigation.spec.ts
```diff
@@ -36,7 +36,7 @@ const navigateMalwareAnalyses = async (page: Page) => {
   await page.getByLabel('relationships', { exact: true }).click();
   await expect(page.getByRole('link', { name: 'related to Malware Spelevo EK' })).toBeVisible();
   await page.getByLabel('entities', { exact: true }).click();
-  await expect(page.getByRole('link', { name: 'Malware Spelevo EK admin ryuk' })).toBeVisible();
+  await expect(page.getByRole('link', { name: 'Malware Spelevo EK admin covid-19' })).toBeVisible();
```
JeremyCloarec, line 39: I don't quite get why this needed to be changed?

opencti-platform/opencti-graphql/src/resolvers/stix.js
```diff
@@ -10,8 +10,8 @@ import { ENABLED_DEMO_MODE } from '../config/conf';
 import { ENTITY_TYPE_USER } from '../schema/internalObject';
 
 const creatorsLoader = batchLoader(batchCreators);
-const relBatchLoader = batchLoader(batchInternalRels);
 const internalLoadThroughDenormalized = (context, user, element, inputName, args = {}) => {
+  const relBatchLoader = batchLoader(batchInternalRels);
```
JeremyCloarec: Do we know how this change impacts performances? While the DataLoader used in batchLoader does state in its documentation `consider creating a new instance per web request`, I wonder what the overhead is on instantiating a new DataLoader each time. What was the reason for moving this into internalLoadThroughDenormalized instead of keeping it ouside?

opencti-platform/opencti-graphql/src/database/middleware.js
```diff
@@ -488,7 +488,7 @@ export const loadElementsWithDependencies = async (context, user, elements, opts
   return loadedElements;
 };
 const loadByIdsWithDependencies = async (context, user, ids, opts = {}) => {
-  const elements = await elFindByIds(context, user, ids, { ...opts, withoutRels: true, connectionFormat: false });
+  const elements = await elFindByIds(context, user, ids, { ...opts, connectionFormat: false });
```
JeremyCloarec, line 490: (not related to this line) There is still a `withoutRels: true` used in `loadElementMetaDependencies` that can be removed

opencti-platform/opencti-graphql/src/database/engine.js
```diff
@@ -326,7 +326,7 @@ export const searchEngineInit = async () => {
       apiKey: conf.get('elasticsearch:api_key') || null,
     },
     maxRetries: conf.get('elasticsearch:max_retries') || 3,
-    requestTimeout: conf.get('elasticsearch:request_timeout') || 30000,
+    requestTimeout: conf.get('elasticsearch:request_timeout') || 3600000,
```
JeremyCloarec, line 336: What's the reason for a 1h timeout instead of 30s? Isn't 1h too much?

opencti-platform/opencti-graphql/src/database/engine.js
```diff
@@ -3478,18 +3529,24 @@ export const elPaginate = async (context, user, indexName, options = {}) => {
...
-  const { types = null, connectionFormat = true } = options;
-  const body = await elQueryBodyBuilder(context, user, options);
+  const { withoutRels = true, types = null, connectionFormat = true } = options;
+  const body = await elQueryBodyBuilder(context, user, { ...options, relCount: true });
   if (body.size > ES_MAX_PAGINATION && !bypassSizeLimit) {
     logApp.info('[SEARCH] Pagination limited to max result config', { size: body.size, max: ES_MAX_PAGINATION });
     body.size = ES_MAX_PAGINATION;
   }
+  const _source = { excludes: [] };
+  if (withoutRels) _source.excludes.push('rel_*');
```
JeremyCloarec: We could reuse the `REL_INDEX_PREFIX` contant here

opencti-platform/opencti-graphql/src/database/engine.js
```diff
@@ -1628,11 +1624,51 @@ export const computeQueryIndices = (indices, typeOrTypes, withInferences = true)
...
+  'rel_sample*.keyword',
+];
+
+const REL_COUNT_SCRIPT_FIELD = {
+  script: {
+    lang: 'painless',
+    source: `
+          int totalElements = 0;
+          for (String fieldName : params['_source'].keySet()) {
+            if (fieldName.startsWith('rel_')) {
```
JeremyCloarec, line 1659: We could reuse `REL_INDEX_PREFIX` instead of the hardocded rel_ here

JeremyCloarec:
When testing, I had an error thrown when the `createSharingTask` method was used (when sharing a report with an orga for example). The error was thrown in `elDataConverter` (line 1492).
When `elPaginate`/`elFindByIds` is called with custom baseData and baseFields (like in `createSharingTask`), the esHit doesn't always contain an entity_type This causes an error when calling `isSingleRelationsRef` in `elDataConverter`.

To fix this, I think we could always include entity_type in the baseFields

opencti-platform/opencti-front/tests_e2e/navigation/navigation.spec.ts
richard-julien: labels must be sort by name. This test showed me that there is a a problem with the loader.
I patch the bulk loader init and fix the test

opencti-platform/opencti-graphql/src/database/engine.js
richard-julien: Aligned with the bulk elastic timeout configuration.

opencti-platform/opencti-graphql/src/database/engine.js
richard-julien: Yep, i will change that to use the prefix and the rel constant.

opencti-platform/opencti-graphql/src/database/engine.js
richard-julien: yes

opencti-platform/opencti-graphql/src/resolvers/stix.js
richard-julien: Was due to the problem of labels ordering.
It could maybe have an impact, difficult to say.
This change was to validate this theory. Now could be great to rollback this change and find how to prevent this sorting issue

richard-julien:
> baseFields

Not sure to understand how it can be related to the PIR. Interested to have a demo :)

JeremyCloarec (dismissed):

opencti-platform/opencti-graphql/src/database/engine.js
```diff
@@ -1681,6 +1725,10 @@ export const elFindByIds = async (context, user, ids, opts = {}) => {
     // Handle draft
     const draftMust = buildDraftFilter(context, user, opts);
     const body = {
+      sort: [{ [orderBy]: orderMode }],
+      script_fields: {
+        script_field_denormalization_count: REL_COUNT_SCRIPT_FIELD
```
SouadHadjiat: to check if we need it here all the time or if we can condition it

JeremyCloarec: We do need to condition it to fix performances issues, but not sure how to. 
My first idea is to change the resolver for numberOfConnectedElement in stixCoreObject.js to batch reload the object with the script included, and never include the script outside of this case. Since that's the only place where we need the result of the script

opencti-platform/opencti-graphql/src/database/engine.js
```diff
@@ -3040,6 +3103,7 @@ const completeSpecialFilterKeys = async (context, user, inputFilters) => {
 const elQueryBodyBuilder = async (context, user, options) => {
   // eslint-disable-next-line no-use-before-define
   const { ids = [], after, orderBy = null, orderMode = 'asc', noSize = false, noSort = false, intervalInclude = false } = options;
+  const { relCount = false } = options;
```
SouadHadjiat, line 3106: when is relCount used ? I can only see it used for elFindByIds, otherwise it is always false

JeremyCloarec: you're right, it's no longer used following the rework of the previous commit. I think it's still useful to keep it available for future uses if necessary

Issue #10095, opened 2025-02-27
fix(data-sharing): users not in the platform organization don't receive events when using live streams with platform organizations


samczsun:
## Description

I just spent 2 hours debugging this.

I have an instance with the platform organization set to `Acme Inc`. I have two users who are both part of `Contonso Corp`. User A is creating knowledge, while User B is consuming knowledge using a live stream.

When User A creates knowledge, it is automatically shared with `Contonso Corp`. This is calculated here: https://github.com/OpenCTI-Platform/opencti/blob/8573c5c61e8052cc77c4ae06a72f0ad0583959d1/opencti-platform/opencti-graphql/src/database/data-builder.js#L29

Note that this function returns an object where `element` is the STIX entity, and `relations` is the relationships to create. The entire `dataEntity` is then passed to `indexCreatedElement` which persists the STIX entity and its relationships into Elastic.

However, the returned `createdElement` only contains `dataEntity.element`, not `dataEntity.relations`. This means that when the created entity is pushed into the Redis stream, it does not contain the correct `granted_refs`.

Now, when User B attempts to load the stream, `checkUserCanAccessStixElement` returns false due to the missing `granted_refs`: https://github.com/OpenCTI-Platform/opencti/blob/8573c5c61e8052cc77c4ae06a72f0ad0583959d1/opencti-platform/opencti-graphql/src/utils/access.ts#L572

## Environment

1. OS (where OpenCTI server runs): Ubuntu
2. OpenCTI version: 6.4.11
3. OpenCTI client: N/A
4. Other environment details: N/A

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Set platform organization
2. Create two users who are not in the platform organization
3. Create a live stream that filters on some knowledge
4. Subscribe to the live stream from User B
5. Create knowledge from User A, observe that the knowledge is automatically restricted to the user's organization
6. Observe that the live stream does not push a create or update event
7. Grant User B bypass permission, repeat steps 4-5, observe that the live stream now contains the create event

## Expected Output

Live streams should work while using platform organizations

## Actual Output

Live streams do not work

nino-filigran:
@samczsun I've not been able to reproduce your bug. Maybe I did not use the correct repro steps? 
Let me write here the steps I have taken, to see if there is any variation with what you mentioned. 


**Stream set up**
On my platform 1: 
- I've set the org segregation, by setting filigran as the main platform org
- created two users A & B, both belonging to a another org
[...]

PR #11459, opened 2025-06-22
[backend] set granted refs on input (#10095)

samczsun:
### Proposed changes
The problem is that if the user does not specify an `INPUT_GRANTED_REFS`, then the backend automatically computes the default value without updating it on the original input object. This means that ElasticSearch and Redis see two different objects:
- ElasticSearch stores the entity and the relationships separately and merges them every read, so the relationships are always correct
- Redis will only see the original input without considering the relationship objects

The end result is that Redis streams are not aware of the default-applied `INPUT_GRANTED_REFS`. This PR updates the calculation so that the default value is applied on the input, instead of being passed straight into `relsToCreate`.

### Related issues

* #10095
*

### Checklist

I'm not really sure how to write a test for this

### DIFF
--- a/opencti-platform/opencti-graphql/src/modules/entitySetting/entitySetting-utils.ts
+++ b/opencti-platform/opencti-graphql/src/modules/entitySetting/entitySetting-utils.ts
@@ -5,6 +5,7 @@ import {
   ABSTRACT_STIX_DOMAIN_OBJECT,
   ENTITY_TYPE_CONTAINER,
   INPUT_AUTHORIZED_MEMBERS,
+  INPUT_GRANTED_REFS,
   INPUT_MARKINGS
 } from '../../schema/general';
 import { STIX_SIGHTING_RELATIONSHIP } from '../../schema/stixSightingRelationship';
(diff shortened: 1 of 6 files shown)


samczsun:
Drone seems to be a little flakey because I don't understand how my changes could affect that test

opencti-platform/opencti-graphql/src/modules/entitySetting/entitySetting-utils.ts
```diff
@@ -144,52 +152,58 @@ export const getDefaultValues = (attributeConfiguration: AttributeConfiguration,
...
-
-        if (attr.name === INPUT_AUTHORIZED_MEMBERS && parsedValue) {
-          const defaultAuthorizedMembers = (parsedValue as string[]).map((v) => JSON.parse(v));
-          // Replace dynamic creator rule with the id of the user making the query.
-          const creatorRule = defaultAuthorizedMembers.find((v) => v.id === MEMBER_ACCESS_CREATOR);
-          if (creatorRule) {
-            creatorRule.id = user.id;
+  if (!input[INPUT_GRANTED_REFS] && entitySetting?.target_type) {
+    const isSegregationEntity = !STIX_ORGANIZATIONS_UNRESTRICTED.some((o) => getParentTypes(entitySetting.target_type).includes(o))
+        || STIX_ORGANIZATIONS_RESTRICTED.some((o) => o === entitySetting.target_type || getParentTypes(entitySetting.target_type).includes(o));
```
OctaveLaventure, line 159: **1. Refactoring**
This could be extracted to an `isSegregationEntity` function and re-used in `data-builder.js`

OctaveLaventure (commented):
Hello @samczsun! Thank you very much for your pull request!
Can you please rebase it on the latest master and add a few tests covering your changes?
Apart from that, looks good to me.
@aHenryJard or @SouadHadjiat can one of you review as well?

OctaveLaventure:
Closing in favor of the updated PR with changes from this PR + tests and refactoring: https://github.com/OpenCTI-Platform/opencti/pull/13691

PR #11461, opened 2025-06-22
[backend/frontend] Improve SSO error UI feedback when local strategy is disabled (#11353)

### LINKED ISSUE #11353: Cookie error is never shown if local strategy is not used
## Description

`opencti_flash` is only shown in `LoginForm`, which is not rendered if no `FORM` strategies are enabled (for example, if only OIDC strategies are enabled)

## Environment

4. Other environment details:

## Reproducible Steps

1. { e.g. Run ... }
2. { e.g. Click ... }
3. { e.g. Error ... }

samczsun:
### Proposed changes
The cookie error is shown with the `LoginForm`, which is only shown if at least one strategy with a form is enabled. In some configurations (such as those that only use OIDC, including demo.opencti.io), this is not the case, and errors will never be shown.

Also, it seems like sometimes the cookie was expiring before the page could fully load, even on localhost, so I bumped the expiry up to 10 seconds
*
*

### Related issues

* #11353
*

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/public/components/Login.tsx
+++ b/opencti-platform/opencti-front/src/public/components/Login.tsx
@@ -1,4 +1,5 @@
 import React, { FunctionComponent, useState } from 'react';
+import { useCookies } from 'react-cookie';
 import Button from '@mui/material/Button';
 import { useTheme } from '@mui/styles';
 import { Facebook, Github, Google, KeyOutline } from 'mdi-material-ui';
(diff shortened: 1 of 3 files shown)


aHenryJard (commented):
Adding a screenshot for next reviewer:
Screenshot from 2025-07-04 17-25-48

@samczsun  thank you for this contribution but this project requires signed commit. Could you please [sign your commits](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) ?

aHenryJard (commented):
Tested locally with OpenID and SAML, looks good to me.

Thank you for this contribution, I will merge it soon.

aHenryJard (approved):
Tested locally on SAML and OpenID, all good to me. Thank !

Issue #11476, opened 2025-06-24
Regression Bug - Inability to easily add external references to an object


nhuber0724:
## Description

With the new upgraded version, the ability to easily add external references to an object no longer exists. When attempting to add a newly created external reference to an object, the user must first create the external reference, then close out the external reference list, refresh OpenCTI, and then go back in and search for the external reference (which the newly created external reference doesn't show up at the top of the list). 

## Environment

OpenCTI v6.7

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Navigate to a Report Object and click on + next to external references on the Overview tab
2. Create a new External Reference - the newly created external reference won't show up in the external reference list
3. Close out the list and refresh
4. Go back in to add the external reference. 
5. The new external reference won't be listed at the top of the page immediately

## Expected Output

Upon creating a new external reference, the external reference should immediately be listed at the top of the external reference list within the object for easy addition to the object. 

## Actual Output

The newly created external reference is not immediately listed within the external reference list nor is it listed at the top of the list.

PR #11495, opened 2025-06-25
[frontend] External ref creation updater in Add external ref list (#11476)

Archidoit:
### Proposed changes
When creating an external reference from an 'Add external reference' form, the external reference should be added at the top of the external references list

### Related issues
https://github.com/OpenCTI-Platform/opencti/issues/11476


SarahBocognano (approved):
Niquel !

Issue #11249, opened 2025-06-09
Entity historic events are not periodically reloaded


Archidoit:
## Description
Since 6.6.13, in the overview of an entity, the historic is not reloaded every 5 seconds like before

Due to https://github.com/OpenCTI-Platform/opencti/pull/11042/files

PR #11498, opened 2025-06-25
[frontend] refetch entity history every 5s (#11249)

Archidoit:
### Proposed changes
Refetch history in entities overview every 5s

### Related issues
https://github.com/OpenCTI-Platform/opencti/issues/11249

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectHistoryLines.tsx
+++ b/opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectHistoryLines.tsx
@@ -1,10 +1,15 @@
-import React, { FunctionComponent } from 'react';
-import { graphql, PreloadedQuery, useFragment, usePreloadedQuery } from 'react-relay';
+import React, { FunctionComponent, useEffect } from 'react';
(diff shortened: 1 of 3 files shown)


opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectHistoryLines.tsx
```diff
@@ -40,15 +46,33 @@ const StixCoreObjectHistoryLinesFragment = graphql`
...
+
+  useEffect(() => {
+    // Refresh the history every interval
+    const subscription = interval$.subscribe(() => {
+      refetch(paginationOptions, { fetchPolicy: 'store-and-network' });
+    });
+    return function cleanup() {
+      subscription.unsubscribe();
+    };
+  }, [refetch]);
```
lndrtrbn: Best effort: we have a hook `useInterval` to simplify loop logic instead of rxjs.interval+useEffect, would be nice to replace it

lndrtrbn (approved):

opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectHistoryLines.tsx
Archidoit: done :) thanks !!

Issue #11507, opened 2025-06-25
[CI] Improve Drone pipeline


efaure:
## Use case

Improve Drone pipeline

## Proposed Solution

Gathering small changes that may improve and help to monitor CI.

 - Display drone runner name at the beginning of a pipeline
 - Switch to Kubernetes runners

PR #11508, opened 2025-06-25
[ci] Display drone runner name (#11507)

efaure:
### Proposed changes

* Display drone runner name as a first step of the pipeline
*

### Related issues

* #11507

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -10,7 +10,7 @@ steps:
   - name: Runner information
     image: alpine:3.19
     commands:
-    - echo DRONE_STAGE_MACHINE  ${DRONE_STAGE_MACHINE}
+    - echo DRONE_STAGE_MACHINE ${DRONE_STAGE_MACHINE}
 
   - name: dependencies-checkout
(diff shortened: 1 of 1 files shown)


.drone.yml
```diff
@@ -3,6 +3,11 @@ kind: pipeline
 name: opencti-tests
 
 steps:
+  - name: Runner information
+    image: alpine:3.19
+    commands:
+    - echo DRONE_STAGE_MACHINE ${DRONE_STAGE_MACHINE}
```
aHenryJard, line 9: could it be the first line of `dependencies-checkout` step instead ?

efaure: I added it in a separate step to ensure to always have it, even if the pull of there is an issue with secret or image used in `dependencies-checkout`. And it seems to add only 4 seconds.
But no strong opinion on this, I can move it.

aHenryJard (approved):
Let's go

aHenryJard (commented):
This PR need to be rebase on top of master please.

Issue #10970, opened 2025-05-11
FINTEL cannot produce PDFs with Japanese characters


damians-filigran:
## Description

When creating a PDF from HTML with the FINTEL feature, from a Japanese language content, the Japanese characters are replaced with non-printable blocks on the PDF output.

## Environment

1. OS: SaaS
2. OpenCTI version: 6.10
3. OpenCTI client: Edge on MacOS
4. Other environment details:

## Reproducible Steps

See [internal video](https://www.notion.so/filigran/FINTEL-PDFs-cannot-reproduce-Japanese-characters-1f08fce17f2a8088af2df6a7796bb9b6?pvs=4)

PR #11539, opened 2025-06-27
[frontend] Fintel cannot produce PDFs with Japanese characters (#10970)

SarahBocognano:
### Proposed changes

* Add missing fonts, add regex to detect pdf language and apply good font (also did it for Chinese & Korean)

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/10970

### DIFF
--- a/opencti-platform/opencti-front/src/utils/htmlToPdf/utils/pdfFonts.ts
+++ b/opencti-platform/opencti-front/src/utils/htmlToPdf/utils/pdfFonts.ts
@@ -3,7 +3,7 @@ import { APP_BASE_PATH } from '../../../relay/environment';
 const { protocol, hostname, port } = window.location;
 const url = `${protocol}//${hostname}:${port || ''}`;
 
-const FONTS = {
+export const FONTS = {
   Roboto: {
     normal: `${url}${APP_BASE_PATH}/static/ext/Roboto-Regular.ttf`,
     bold: `${url}${APP_BASE_PATH}/static/ext/Roboto-Bold.ttf`,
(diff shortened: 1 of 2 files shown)


opencti-platform/opencti-front/src/utils/htmlToPdf/utils/pdfFonts.ts
```diff
@@ -22,6 +22,27 @@ const FONTS = {
...
 };
 
-export default FONTS;
+const isJapanese = (htmlData: string) => /[\u3000-\u303F\u3040-\u309F\u30A0-\u30FF\uFF00-\uFFEF\u4E00-\u9FAF\u3400-\u4DBF]/.test(htmlData);
+const isKorean = (htmlData: string) => /[\u1100-\u11FF\u3130-\u318F\uAC00-\uD7AF]/.test(htmlData);
+const isChinese = (htmlData: string) => /[\u4E00-\u9FFF]/.test(htmlData);
+
+export const detectLanguage = (htmlData: string) => {
+  if (isJapanese(htmlData)) return 'NotoSansJp';
+  if (isKorean(htmlData)) return 'NotoSansKr';
```
aHenryJard, line 40: I tested locally, it's all good for Japanese but I see that you check Korean and Chinese and actually it's not good for them.
Maybe this PR should focus on Japanese and we can open an issue on other language ?

Screenshot from 2025-07-02 09-19-10

Screenshot from 2025-07-02 09-17-35

SarahBocognano: Probably because you put all the languages in the same pdf ^^. The Korean one is working when only Korean language is detected in the pdf.
And for Chinese it's because I put a simplified chinese fonts and not a complex one. So that depends of which type of chinese you used. I can add the complexe one for this one
image

SarahBocognano: Buuuuuut after investigation there is a font that covers all 3 of them. So I will add it right away !

aHenryJard: ah ! ok

aHenryJard (approved):
Tested locally all good.
Would be great to add in documentation which languages are currently supported by fintel and the limitation on first language detected.

PR #11540, opened 2025-06-27
[ci] Improve drone workflow

efaure:
### Proposed changes

* Remove cache shared on the runner between multiple build
* Add image pull secret


efaure:
@aHenryJard  The PR to switch to Kubernetes runner is ready. I made 20 test build, all successful. 

I tested with and without node cache. It seems that there is no difference in flakiness or build duration. 
For now, we can keep the cache, but if we think that it's an issue we know we can remove it.

Feel free to merge when you want.

aHenryJard (approved):

PR #11552, opened 2025-06-30
[Backend/frontend] Add generation licences file + verification that no "unsupported" licenses are installed

Kedae:
During the CI we generate a .txt file containing all the licenses of both back and front, displaying the corresponding librairies.
Additionnaly, we check that all licenses used are contained in an array of "valid" licenses.

-> One improvement could be to have the count of licenses on top of the file, I'm looking into it as it is not trivial. But review 
can be done nonetheless


.drone.yml
```diff
@@ -19,6 +19,33 @@ steps:
       - echo "[INFO] using client-python on branch $(git branch --show-current)"
       - git log -n 1
 
+  - name: backend-generate-licenses
+    image: nikolaik/python-nodejs:python3.11-nodejs20-alpine
+    volumes:
+      - name: cache-node-backend
+        path: /drone/src/opencti-platform/opencti-graphql/node_modules
```
aHenryJard: Please remove the cache usage, using another step cache is bringing flakiness in builds. We are about to remove them anyway, it's not faster with a cache.

.drone.yml
```diff
@@ -19,6 +19,33 @@ steps:
...
+      - pip install -e .[dev,doc]
+      - cd "$DRONE_WORKSPACE/opencti-platform/opencti-graphql"
+      - yarn install
+      - yarn verify-licenses
+
+  - name: frontend-generate-licenses
+    image: node:22.14.0
+    volumes:
+      - name: cache-node-frontend-e2e
+        path: /drone/src/opencti-platform/opencti-front/node_modules
```
aHenryJard: Please remove cache

labo-flg:
I have tested the script `generate-license.sh`
Result seems ok.

Could you put the generated file (under `licenses` folder) in gitignore ?

scripts/generate-licenses.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+
+mkdir ./licenses
+pwd
+ls
```
labo-flg: debug lines ?
I'd say either remove or add a meaningful log instead.

.drone.yml
```diff
@@ -19,6 +19,33 @@ steps:
...
+    commands:
+      - chmod 777 scripts/*
+      - ./scripts/generate-licenses.sh
```
labo-flg: I would add another step that depends on both front and back `verify-licenses`, that would generate the licenses.
This scripts checks both frontend and backend and is in a "frontend" step, which is misleading.
Also, it generates _before_  actually checking. In case it's wrong we do not want to generate at all.

.drone.yml
Kedae: Done

.drone.yml
Kedae: Done

scripts/generate-licenses.sh
Kedae: Done

.drone.yml
Kedae: Done

labo-flg (approved):

Issue #11441, opened 2025-06-20
Error Displayed When Creating a Relationship Between a Malware Node and Two (2) Campaign Nodes


ericWadeFord:
## Description
When creating a relationship between a malware node and two (2) campaign nodes in a Report or Investigation in the Knowledge Graph, the error “The relationship type uses is not matched between campaign and campaign" is displayed.

## Environment
6.6.18

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Navigate to a report or start an Investigation
2. Select a malware node. Then, two (2) campaigns. If starting an investigation, add the entities first, then, select them.
3. Create the relationship "uses" by clicking the REVERSE relationship icon.
4. Observe that one (1) relationship is created, but the error “The relationship type uses is not matched between campaign and campaign" is displayed.

## Expected Output
A relationship edge is created between each campaign node and the malware node.

## Actual Output
Only one relationship edge is created and the error “The relationship type uses is not matched between campaign and campaign" is displayed.

## Additional information
This behavior may occur in other Knowledge Graph views. Confirmed in Reports and in an Investigation.

## Screenshots (optional)

![Image](https://github.com/user-attachments/assets/8a3f2cf9-2980-4152-b728-cf090d2af616)
_Error displayed after creating relationships_

PR #11569, opened 2025-07-02
[frontend] Multiple relationships creation from graph after Reverse (#11441)

Archidoit:
### Proposed changes
Select nodes from a container graphs (exemple: a malware and 2 campaigns).
Click on the button 'create relationship'
Reverse the relationship
Click on create.
The relationships should be created. (Before the bug resolution: only the first relation was created and an error message was displayed)

### Related issues
https://github.com/OpenCTI-Platform/opencti/issues/11441


opencti-platform/opencti-front/src/private/components/common/stix_core_relationships/StixCoreRelationshipCreation.tsx
```diff
@@ -643,62 +694,36 @@ class StixCoreRelationshipCreation extends Component {
...
-          anchor="right"
-          elevation={1}
-          sx={{ zIndex: 1202 }}
-          classes={{ paper: classes.drawerPaper }}
-          onClose={this.handleClose.bind(this)}
-        >
-          {step === 0
+  return (
+    <>
+      <Drawer
```
delemaf, line 701: For lighter DOM you could unmount Drawer if not open
```suggestion
      {open && (<Drawer
```

delemaf (approved):
Works for me, just a little suggestion but no blocker :smiley:

PR #11571, opened 2025-07-02
[backend] Improve management of hashed observable IDs (#11301)

### LINKED ISSUE #11301: Hashed observables sometimes doest cumulate correctly stix ids that fail absorptions of related relationships
## Description

In a CSV file, I have modelized some incidents, that needs to be related to some File Observables. I have 2 incidents that need to be linked to a single, however, my file has 3 different names and is represented by some hashes (MD5, Sha1, Sha256). When trying to import my file using draft, it fails.
## Environment

Testing 

## Reproducible Steps

1. Use the CSV file attached (3 lines, 2 incidents, 3 files (all having a different name, but on line 1 and 2 sha1/MD5/sha256 are filled while on line 3 only the file name & sha1 are filled for the file)
2. Use the  CSV mapper attached -*On testing, you can use the Mapper called **File&Incident Repro Bug**
3. Launch import (in draft, easier to notice the issue)

## Expected Output

All three files are deduplicated, creating a single file & this file has 2 relations (related to) to incidents.

## Actual Output

The three files are deduplicated. However a single relation is created & a "missing reference error" is raised. 

## Additional information

It seems that reading the error message, the missing relation is due to the fact that the relation tries to be created with the file having a single hash, identified by a specific Stix-ID. However given this file gets deduplicated, the stix-ID does not exists (and therefore not found), but it's an assumption only. 

## Screenshots (optional)

Image

[20250612_csvMapper_File&Incident Repro Bug.json](https://github.com/user-attachments/files/20709199/20250612_csvMapper_File.Incident.Repro.Bug.json)

[File&Incident.csv](https://github.com/user-attachments/files/20709209/File.Incident.csv)

lndrtrbn:
### Proposed changes

* During attribute update, modify the logic manipulating x_opencti_stix_ids because it was not correct in case of observables with ids generated by priority instead of composition
* Skip some merge logic in case of hashes observables
* Take all hashes id in account while looking for entities
* Add tests to validate behavior of different scenarios

### Related issues

* #11301

### Checklist


SouadHadjiat (approved):

opencti-platform/opencti-graphql/src/schema/identifier.js
```diff
@@ -548,6 +571,7 @@ export const getInputIds = (type, input, fromRule) => {
   }
   ids.push(...generateAliasesIdsForInstance(input));
   ids.push(...getHashIds(type, input.hashes));
```
richard-julien, line 573: Not sure this getHashIds function really do something interesting

opencti-platform/opencti-graphql/src/domain/stixCyberObservable.js
```diff
@@ -225,6 +225,15 @@ export const addStixCyberObservable = async (context, user, input) => {
   // Convert hashes to dictionary if needed.
   if (isStixCyberObservableHashedObservable(type) && observableInput.hashes) {
     observableInput.hashes = inputHashesToStix(observableInput.hashes);
+    const standardId = generateStandardId(type, observableInput);
```
richard-julien: Not sure about the code location.
Can you check if we always go through this function when we create observable from indicator?
This stix ids gen could maybe be move to buildEntityData in createEntityRaw?

opencti-platform/opencti-graphql/src/database/middleware.js
```diff
@@ -1840,7 +1841,28 @@ const updateAttributeRaw = async (context, user, instance, inputs, opts = {}) =>
...
+        const newStixIds = [...instanceOtherStixIds, ...updatedInstanceStandardIds].filter((id) => id !== standardId);
+        const stixIdsHaveNotChanged = instanceStixIds.length === newStixIds.length
+          && newStixIds.every((id) => instanceStixIds.includes(id));
+
+        if (!stixIdsHaveNotChanged) {
+          const stixInput = R.find((e) => e.key === IDS_STIX, preparedElements);
+          if (stixInput) {
+            // If update already contains a change of the other stix ids
+            // we need to impact directly the impacted and updated related input
+            stixInput.operation = UPDATE_OPERATION_REPLACE;
```
richard-julien: Are you sure about this? What about if its come from a field patch using remove operation?

opencti-platform/opencti-graphql/src/schema/identifier.js
```diff
@@ -347,11 +347,16 @@ export const fieldsContributingToStandardId = (instance, keys) => {
...
+    return null;
   }
+  return R.map((t) => t.src, R.flatten(properties));
+};
+
+export const fieldsContributingToStandardId = (instance, keys) => {
+  const allFields = allFieldsContributingToStandardId(instance);
+  if (allFields === null) return keys;
   const targetKeys = R.map((k) => (k.includes('.') ? R.head(k.split('.')) : k), keys);
-  const propertiesToKeep = R.map((t) => t.src, R.flatten(properties));
```
richard-julien, line 353: Can you explain this modification? thanks

opencti-platform/opencti-graphql/tests/02-integration/02-resolvers/observables-hashes-test.ts
```diff
@@ -0,0 +1,596 @@
...
+      query: DELETE_STIX_FILE_QUERY,
+      variables: { id: file4Id, },
+    });
+    await queryAsAdmin({
+      query: DELETE_STIX_FILE_QUERY,
+      variables: { id: file5Id, },
+    });
+  });
+
+  it('should replace standard_id and add old one in x_opencti_stix_ids if prior data arrives by Upsert', async () => {
```
richard-julien, line 143: Does this test cumulate the name in the other_stix_ids?

opencti-platform/opencti-graphql/src/domain/stixCyberObservable.js
SouadHadjiat: ah you're right, I though we called stixCyberObservable domain to create from indicator but we don't.. so we need to move it indeed

opencti-platform/opencti-graphql/tests/02-integration/02-resolvers/observables-hashes-test.ts
SouadHadjiat: the name doesn't generate other stix ids

opencti-platform/opencti-graphql/src/schema/identifier.js
lndrtrbn: Yes we hesitated to remove it to limit the risk of side effects

opencti-platform/opencti-graphql/src/schema/identifier.js
lndrtrbn: At some intermediate point we were in need of most of the logic of the function `fieldsContributingToStandardId` so we extracted it inside an other one (`allFieldsContributingToStandardId`). At the end we do not need this new function anymore but we still let the separation because we have written unit tests on the new one.

We can rollback if you think it is wiser

opencti-platform/opencti-graphql/src/database/middleware.js
lndrtrbn: The objective was to always compute the new array of ids and replace the old one with it no matter the operation. But indeed maybe for remove operation this line
```
stixInput.value = R.uniq([...stixInput.value, ...newStixIds]);
```
is not correct because the input value is what we want to remove, we need to test the input operation to do the right thing

lndrtrbn: In fact I think there is also an issue if the input operation is "replace", this current code works only for add operation

lndrtrbn: Fixed

opencti-platform/opencti-graphql/src/domain/stixCyberObservable.js
lndrtrbn: Fixed

PR #11575, opened 2025-07-02
[backend] Fix perf issue due to large sync operation (#11390)

### LINKED ISSUE #11390: Performance issue due to large synchronous operation
## Description

We identified in `sseMiddleware.js` a bottleneck that might make the platform unresponsive while stuck in a synchronous operation.

Faulty line is : https://github.com/OpenCTI-Platform/opencti/blob/e9bf7949cf8c1ebf8ee101b8556f18f076a4c7bb/opencti-platform/opencti-graphql/src/graphql/sseMiddleware.js#L210

## Additional information

Dev can be adfded into the PR #10971 that already introduces the function asyncMap to do the job.

SouadHadjiat:
### Proposed changes

* use asyncMap & asyncFilter

### Related issues

Issue discovered with Pyroscope
* #11390 

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/graphql/sseMiddleware.js
+++ b/opencti-platform/opencti-graphql/src/graphql/sseMiddleware.js
@@ -1,4 +1,3 @@
-import * as R from 'ramda';
 import * as jsonpatch from 'fast-json-patch';
 import { Promise } from 'bluebird';
 import { LRUCache } from 'lru-cache';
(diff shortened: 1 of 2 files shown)


opencti-platform/opencti-graphql/src/graphql/sseMiddleware.js
```diff
@@ -206,9 +207,9 @@ const createSseMiddleware = () => {
     if (refsToResolve.length > 0) {
       const resolvedStoreElements = await storeLoadByIdsWithRefs(context, req.user, refsToResolve);
       missingElements.push(...resolvedStoreElements);
-      const resolvedMissingIds = R.uniq(missingElements.map((elem) => extractIdsFromStoreObject(elem)).flat());
-      const parentRefs = resolvedStoreElements.map((r) => stixRefsExtractor(convertStoreToStix(r)))
-        .flat().filter((parentId) => !resolvedMissingIds.includes(parentId));
+      const resolvedMissingIds = R.uniq((await asyncMap(missingElements, (elem) => extractIdsFromStoreObject(elem))).flat());
```
richard-julien: Seems good but i'm a bit worried about the R.uniq and .flat() impact as this 2 methods a still event loop blocking potential point. Have you been able to test the performance and the blocking aspect of the method to validate that this way we can safely handle a very large anount of resolvedStoreElements?

SouadHadjiat: well, I can change asyncMap function to introduce options for uniq and flat and implement them in the function. 

How exactly can I test ? I tried to enable event_loop_logs locally and open a stream, I don't see anything in the logs, but I don't have enough data and also I don't see event_loop_logs used anywhere (when did we remove it ? it's still present in default.config but it doesn't work anymore)

opencti-platform/opencti-graphql/src/utils/data-processing.ts
```diff
@@ -20,3 +20,31 @@ export const asyncFilter = async <T>(elements: T[], predicate: (value: T, index:
...
+export const asyncMap = async <T, Z>(elements: T[], transform: (value: T) => Z, filter?: (value: Z) => boolean, opts: { flat?: boolean, unique?: boolean } = {}) => {
+  const { flat = false, unique = false } = opts;
+  const transformed: Z[] = [];
+  let startProcessingTime = new Date().getTime();
+  for (let index = 0; index < elements.length; index += 1) {
+    const element = elements[index];
+    const item = transform(element); // can be one element or array
+    if (!filter || filter(item)) {
+      if (flat && Array.isArray(item)) {
+        transformed.push(...item);
```
richard-julien: this approach could throw an execption if item lengh > 65000 elements.
I think its better to have a standard for loop to add element.
This loop could also be protected by a set immediate.

opencti-platform/opencti-graphql/src/utils/data-processing.ts
```diff
@@ -20,3 +20,31 @@ export const asyncFilter = async <T>(elements: T[], predicate: (value: T, index:
...
+    // Prevent event loop locking more than MAX_EVENT_LOOP_PROCESSING_TIME
+    if (new Date().getTime() - startProcessingTime > MAX_EVENT_LOOP_PROCESSING_TIME) {
+      startProcessingTime = new Date().getTime();
+      await new Promise((resolve) => {
+        setImmediate(resolve);
+      });
+    }
+  }
+  if (unique) {
+    return Array.from(new Set(transformed));
```
richard-julien: Honestly maybe this way can also be loop blocking.
Maybe creating a asyncUniqMap taking only string and directly adding in a Set will be more efficient

richard-julien (approved):

Issue #3343, opened 2023-05-17
Massive operations on users


SamuelHassine:
## Use case

For users, with checkboxes, be able to:

* Massively associate users to organizations
* Massively associate users to groups
* Massively delete users

nino-filigran:
Additionally, be able to massively add an expiration date through mass operation & modify Account status massively, to prevent users to login

PR #11601, opened 2025-07-04
[backend/frontend] add new action types for handling users in background tasks (#3343)

JeremyCloarec:
### Proposed changes

* moved Users and Users in organization to DataTable
* added mass operation on users

client PR: https://github.com/OpenCTI-Platform/client-python/pull/932

### Related issues

* #3343 
*

### Checklist


opencti-platform/opencti-front/src/private/components/settings/Users.tsx
```diff
@@ -0,0 +1,249 @@
...
+    );
+  } else if (!isSetAccess && isAdminOrganization && isEnterpriseEdition) {
+    userCreateButton = (
+      <SettingsOrganizationUserCreation
+        paginationOptions={queryPaginationOptions}
+        organization={organization}
+        variant="controlledDial"
+      />
+    );
+  }
```
delemaf: The **Suspense fallback** cause some flash/height issue during loading, we don't really the fallback. This is enough:
```
      <React.Suspense>
```

It could also be good to wrap this in a `useMemo` since isSetAccess, isAdminOrganization, etc. change rarely and it might be lighter for the query loading (and also using immutable :smile:)
```suggestion
const userCreateButton = useMemo(() => {
  if (isSetAccess && defaultGroupsQueryRef) {
    return (
      <React.Suspense fallback={<Loader variant={LoaderVariant.inElement} />}>
        <UserCreation 
          paginationOptions={queryPaginationOptions} 
          defaultGroupsQueryRef={defaultGroupsQueryRef} 
        />
      </React.Suspense>
    );
  }
  if (!isSetAccess && isAdminOrganization && isEnterpriseEdition) {
    return (
      <SettingsOrganizationUserCreation
        paginationOptions={queryPaginationOptions}
        organization={organization}
        variant="controlledDial"
      />
    );
  }
  return null;
}, [
  isSetAccess, 
  defaultGroupsQueryRef, 
  isAdminOrganization, 
  isEnterpriseEdition,
  queryPaginationOptions,
  organization
]);```

File: opencti-platform/opencti-front/src/components/dataGrid/components/DataTableHeaders.tsx [hunk 2]
```diff
@@ -32,7 +32,6 @@ const DataTableHeaders: FunctionComponent<DataTableHeadersProps> = ({
     removeSelectAll,
     startsWithAction,
     startsWithIcon,
-    startColumnWidth,
```
delemaf, line 35: This should not be removed, it break the header alignment.
image

opencti-platform/opencti-front/src/components/dataGrid/components/DataTableHeaders.tsx
```diff
@@ -64,7 +63,6 @@ const DataTableHeaders: FunctionComponent<DataTableHeadersProps> = ({
     background: hasSelectedElements && !removeSelectAll
       ? theme.palette.background.accent
       : 'transparent',
-    minWidth: startColumnWidth,
```
delemaf, line 67: Same here for header alignment.

delemaf (changes requested):
Code looks good overall but little front-end fix need to be done

opencti-platform/opencti-front/src/components/dataGrid/DataTable.tsx
```diff
@@ -148,6 +150,7 @@ const DataTableInternalToolbar = ({
     >
 
       <DataTableToolBar
+        isUserDatatable={isUserDatatable}
```
Kedae: I really don't like that. This is for customization of the toolbar, we need to find a more generic way.

ValentinBouzinFiligran: taskScope is now used

delemaf (approved):

Issue #11188, opened 2025-06-03
feat: make it possible to provide platform ID at initial deployment or when cloning


sbocahu:
## Use case

OpenCTI has an internal identifier for each instance, which is automatically generated at bootstrap / initial deployment (UUIDv4).
 
For some usecases (cloning an existing instance, for example) we need to be able to change an existing ID, and for some others it would even be better to provide the ID as soon as bootstrapping the instance.

## Proposed Solution

Could be introducing an env variable

PR #11602, opened 2025-07-04
[backend] allow to set/update platform id (#11188)

labo-flg:
This feature allows to set a platform identifier instead of using the default uuid v4.
Note that technically, the platform id is the id of the Platform Settings object in elastic.

### Proposed changes

* new config token `platform_id`
* When booting a brand new platform, if `platform_id` is defined in the configuration, we use it instead of a random uuid
* When booting an existing platform, we change the platform identifier during startup, by patching the platform settings document in elastic.

### Related issues
* #11188 

### Checklist

### Further comments

To test this feature I suggest to:
* set a new `elasticsearch.index_prefix` if you do not want to pollute your main dev index
* set `platform_id` to a fixed uuidv4 of your choice
* start the platform
* checkout the platform identifier in the app settings (or with a quick query in ES, get the only doc with `Settings` as entityType)
* stop the platform
* change the setting `platform_id` with another value
* restart the backend
* checkout the platform identifier once again
* delete your ES indices to remove your test data

### DIFF
--- a/opencti-platform/opencti-graphql/src/database/data-initialization.js
+++ b/opencti-platform/opencti-graphql/src/database/data-initialization.js
@@ -14,9 +14,12 @@ import { GROUP_DEFAULT, groupAddRelation } from '../domain/group';
 import { TAXIIAPI } from '../domain/user';
 import { KNOWLEDGE_COLLABORATION, KNOWLEDGE_DELETE, KNOWLEDGE_FRONTEND_EXPORT, KNOWLEDGE_MANAGE_AUTH_MEMBERS, KNOWLEDGE_UPDATE } from '../schema/general';
 import { ENTITY_TYPE_CONTAINER_CASE_RFI } from '../modules/case/case-rfi/case-rfi-types';
-import { updateAttribute } from './middleware';
+import { loadEntity, updateAttribute } from './middleware';
 import { ENTITY_TYPE_ENTITY_SETTING } from '../modules/entitySetting/entitySetting-types';
-import { logApp } from '../config/conf';
+import conf, { logApp } from '../config/conf';
+import { INDEX_INTERNAL_OBJECTS, isNotEmptyField } from './utils';
+import { ENTITY_TYPE_SETTINGS } from '../schema/internalObject';
+import { elRawDelete, elRawGet, elRawIndex } from './engine';
 
 // region Platform capabilities definition
 const KNOWLEDGE_CAPABILITY = 'KNOWLEDGE';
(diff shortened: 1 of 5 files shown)


opencti-platform/opencti-graphql/src/database/data-initialization.js
```diff
@@ -381,3 +390,33 @@ export const initializeData = async (context, withMarkings = true) => {
   logApp.info('[INIT] Platform default initialized');
   return true;
 };
+
+export const patchPlatformId = async (context) => {
+  const platformId = conf.get('platform_id') || undefined;
+  if (isNotEmptyField((platformId))) {
+    const platformSettings = await loadEntity(context, SYSTEM_USER, [ENTITY_TYPE_SETTINGS]);
+    if (platformSettings.id !== platformId) {
+      logApp.info(`[INIT] Switching platform identifier from [${platformSettings.id}] to [${platformId}]`);
```
aHenryJard: I wonder about audit log here, it looks like more that just an info to me. And so we want to be able to change it afterward and not only on first startup ?

opencti-platform/opencti-graphql/src/database/data-initialization.js
```diff
@@ -362,7 +365,13 @@ const createBasicRolesAndCapabilities = async (context) => {
 export const initializeData = async (context, withMarkings = true) => {
   logApp.info('[INIT] Initialization of settings and basic elements');
   // Create default elements
+  const platformId = conf.get('platform_id') || undefined;
```
aHenryJard, line 370: Maybe we should check a minimal length, like at least I don't know 5 char, or even better like the admin token. Do we expect an UUID there ?

https://github.com/OpenCTI-Platform/opencti/blob/f11f1bed07bdaa1de79a591d7d2cc27c242c9ad4/opencti-platform/opencti-graphql/src/config/providers.js#L43

richard-julien: Agree. We need to check that is a valid uuid

opencti-platform/opencti-graphql/src/database/data-initialization.js
richard-julien: Yes we want to change it afterward. (duplicate a database then change the id)

opencti-platform/opencti-graphql/src/database/data-initialization.js
labo-flg: yes, you're right !

opencti-platform/opencti-graphql/src/database/data-initialization.js
labo-flg: I'm not sure about audit logs. We cannot really track who did it, it's a settings in the config file.
You want to log it as SYSTEM ?

aHenryJard: or maybe just use warn level instead of info

labo-flg: OK, I'll put warn level.

aHenryJard (approved):
Tested locally, all good to me, thank you for the test coverage.

labo-flg:
I added one more test for setting an invalid UUID

Issue #11008, opened 2025-05-14
Update button greyed out on existing dashboard 'number' widget


fruitcakej:
## Description
In a number widget, if you click on 'Update', add a filter (or not) and click on 'Validate', on the next page the ‘Update’ button is greyed out and it is impossible to update this widget. Similarly, if I duplicate the widget, I have the same problem.
However, if I recreate a widget from scratch, using the same criteria, it's valid and the changes work.

## Environment
1. SaaS 6.6.11

## Reproducible Steps
Steps to create the smallest reproducible scenario:
1. Try to update an 'existing' number widget after validation
2. Update button is greyed out

## Expected Output
Update button to not be greyed out / disabled

## Actual Output
Button greyed out

## Additional information
Link to the widget titled 'PRO Iran hacktivivts CAMPAIGNS' in testing (exported from existing platform).
https://testing.octi.staging.filigran.io/dashboard/workspaces/dashboards/e13d4bfb-27e0-446b-9b3c-c0025c4ba72b
Scroll down to the bottom of the dashboard.
A newly created number widget titled 'TEST 'NEW' UPDATE' works without issue

alice-debra:
I can't reproduce the bug. Can you send me your username? So I can check the permissions.

fruitcakej:
The dashboard has 'can edit' for all users. 
Remember this only affects existing 'number' widgets, if you create a new one, it is fine. That is why I exported the widget  'PRO Iran hacktivivts CAMPAIGNS' to the DB linked above. Try to update that one

fruitcakej:
Perhaps it is an issue in the platform upgrade cycle

PR #11613, opened 2025-07-07
[frontend] fix update button greyed out on existing dashboard 'number' widget (#11008)

CelineSebe:
### Proposed changes

* isDataSelectionNumberValid changes in useWidgetConfigValidateForm

### Related issues

* #11008 

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/widgets/useWidgetConfigValidateForm.tsx
+++ b/opencti-platform/opencti-front/src/private/components/widgets/useWidgetConfigValidateForm.tsx
@@ -1,5 +1,5 @@
 import { useWidgetConfigContext } from './WidgetConfigContext';
-import { getCurrentAvailableParameters } from '../../../utils/widget/widgetUtils';
+import { getCurrentAvailableParameters, isDataSelectionNumber } from '../../../utils/widget/widgetUtils';
 
 export const fintelTemplateVariableNameChecker = /^[A-Za-z0-9_-]+$/;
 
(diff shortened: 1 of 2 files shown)


opencti-platform/opencti-front/src/private/components/widgets/useWidgetConfigValidateForm.tsx
```diff
@@ -32,13 +32,11 @@ const useWidgetConfigValidateForm = () => {
...
   // Check we are at the last step
   const isLastStep = step === 3;
-
   // Check there is a type
   const isTypeFilled = !!type && type !== '';
 
   // Check the number of results is lower than 100 for lists
-  const isDataSelectionNumberValid = dataSelection.every((selection) => !selection.number || selection.number <= 100);
-
+  const isDataSelectionNumberValid = config.widget.type === 'lists' ? dataSelection.every((selection) => !selection.number || selection.number <= 100) : true;
```
lndrtrbn: Despite the comment above, checking the number of results is not made only for list widgets but for all of the ones below:
```
list
distribution-list
timeline
donut
horizontal-bar
radar
polar-area
tree
map
wordcloud
```
So  this condition breaks the checking for them

lndrtrbn (changes requested):

opencti-platform/opencti-front/src/private/components/widgets/useWidgetConfigValidateForm.tsx
lndrtrbn: You can create a function in `widgetUtils.tsx` that returns the list of widgets requiring number limitation and use it in the checking hook

opencti-platform/opencti-front/src/utils/widget/widgetUtils.tsx
```diff
@@ -17,6 +17,8 @@ import {
   ViewListOutline,
 } from 'mdi-material-ui';
 import React from 'react';
+// eslint-disable-next-line import/extensions
+import { WidgetDataSelection } from './widget';
```
lndrtrbn:
```suggestion
import type { WidgetDataSelection } from './widget';
```

opencti-platform/opencti-front/src/utils/widget/widgetUtils.tsx
```diff
@@ -269,3 +271,20 @@ export const renderWidgetIcon = (key: string, fontSize: 'large' | 'small' | 'med
       return <div />;
   }
 };
+
+export const isDataSelectionNumber = (type: string, dataSelection: WidgetDataSelection[]) => {
```
lndrtrbn:
```suggestion
export const isDataSelectionNumberValid = (type: string, dataSelection: WidgetDataSelection[]) => {
```
I think it's better that way to explain what the function does, it does not check if there is a selection number but if it is valid

lndrtrbn (approved):

PR #11620, opened 2025-07-07
[backend] save session when opt login (#11619)

### LINKED ISSUE #11619: "otpValidated" user status is not persisted across opencti-graphql instances
## Description

When deploying multiple GraphQL servers, some state needs to be persisted across all instances.
It seems to be the case for user login, but not user OTP session "activation". In turn, human or programmatic users may be asked to enter their OTP login again, one for each OpenCTI instance behind a load balancer.

I don't have experience in the code, but it looks like that adding req.session.save() [here](https://github.com/OpenCTI-Platform/opencti/blob/87d635b1a473a709bfa65f1191a4f83d762213bf/opencti-platform/opencti-graphql/src/domain/user.js#L1221) could perhaps solve this issue.

## Environment

1. OS (where OpenCTI server runs): Linux
2. OpenCTI version:  6.7.3
3. OpenCTI client: graphql
4. Other environment details:
  - Multiple opencti-graphql API running behind a round-robin load balancer 

## Reproducible Steps

1. Run multiple opencti-graphql API running behind a round-robin load balancer
2. Enable 2FA on a user
3. Login, enter the 2FA code once, and the API+frontend will ask for it again.

## Actual Output

One OTP input should be sufficient.

JeremyCloarec:
### Proposed changes

* save session when otp login 
*

### Related issues

* #11619 
*

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/domain/user.js
+++ b/opencti-platform/opencti-graphql/src/domain/user.js
@@ -1219,6 +1219,7 @@ export const otpUserLogin = async (req, user, { code }) => {
     throw AuthenticationFailure();
   }
   req.session.user.otp_validated = isValidated;
+  req.session.save();
   return isValidated;
 };
 


Gwendoline-FAVRE-FELIX (approved):

gileri:
I can confirm that it works perfectly, thanks !

PR #11628, opened 2025-07-08
[backend] replace ruleApply in background task to direct inferred creation (#11626)

### LINKED ISSUE #11626: Change rule activation in background tasks
## Description

Currently, rule activation in background tasks use the ruleApply API. But a single ruleApply can lead to a lot of infferred 
creation, leading to requests that can timeout.
We want to change this behavior: instead of sending the entities targeted by the rule to the worker, we want to provision the inferred creations in the task manager, and send these inferred creation to the worker, so that each operation in the worker is a unique inferred creation

## Environment

4. Other environment details:

## Reproducible Steps

1. { e.g. Run ... }
2. { e.g. Click ... }
3. { e.g. Error ... }

JeremyCloarec:
### Proposed changes

* now send inferred creation operation to worker instead of rule apply

### Related issues

* Fix #11626 

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/config/schema/opencti.graphql
+++ b/opencti-platform/opencti-graphql/config/schema/opencti.graphql
@@ -15552,4 +15552,8 @@ type Mutation {
   stixSightingRelationshipAdd(input: StixSightingRelationshipAddInput!): StixSightingRelationship
   @auth(for: [KNOWLEDGE_KNUPDATE])
   stixSightingRelationshipEdit(id: ID!): StixSightingRelationshipEditMutations @auth(for: [KNOWLEDGE_KNUPDATE])
+
+  ######## INTERNAL INFERRED CREATION
+  inferredRelationAdd(jsonInput: String!): ID @auth(for: [KNOWLEDGE_KNUPDATE])
+  inferredEntityAdd(jsonInput: String!): ID @auth(for: [KNOWLEDGE_KNUPDATE])
 }
--- a/opencti-platform/opencti-graphql/src/domain/inferredObject.ts
+++ b/opencti-platform/opencti-graphql/src/domain/inferredObject.ts
@@ -8,7 +8,6 @@ export const createInternalInferredRelation = async (context: AuthContext, user:
   if (user.id !== RULE_MANAGER_USER.id) {
     throw ForbiddenAccess();
   }
-  // TODO: JSON input validation? Maybe we don't need it since it's only coming from task manager?
   const { input, ruleContent, opts } = JSON.parse(jsonInput);
   const createdInferredRelation = await createInferredRelation(context, input, ruleContent, opts);
   return createdInferredRelation.id;
@@ -18,7 +17,6 @@ export const createInternalInferredEntity = async (context: AuthContext, user: A
   if (user.id !== RULE_MANAGER_USER.id) {
     throw ForbiddenAccess();
   }
-  // TODO: JSON input validation? Maybe we don't need it since it's only coming from task manager?
   const { input, ruleContent, type } = JSON.parse(jsonInput);
   const createdInferredEntity = await createInferredEntity(context, input, ruleContent, type);
   return createdInferredEntity.id;
--- a/opencti-platform/opencti-graphql/src/domain/inferredObject.ts
+++ b/opencti-platform/opencti-graphql/src/domain/inferredObject.ts
@@ -1,0 +1,25 @@
+import { createInferredEntity, createInferredRelation } from '../database/middleware';
+import type { AuthContext, AuthUser } from '../types/user';
+import { RULE_MANAGER_USER } from '../utils/access';
+import { ForbiddenAccess } from '../config/errors';
+
+export const createInternalInferredRelation = async (context: AuthContext, user: AuthUser, jsonInput: string) => {
+  // This API should only be available to task manager user
+  if (user.id !== RULE_MANAGER_USER.id) {
+    throw ForbiddenAccess();
+  }
+  // TODO: JSON input validation? Maybe we don't need it since it's only coming from task manager?
+  const { input, ruleContent, opts } = JSON.parse(jsonInput);
+  const createdInferredRelation = await createInferredRelation(context, input, ruleContent, opts);
+  return createdInferredRelation.id;
+};
+export const createInternalInferredEntity = async (context: AuthContext, user: AuthUser, jsonInput: string) => {
+  // This API should only be available to task manager user
+  if (user.id !== RULE_MANAGER_USER.id) {
+    throw ForbiddenAccess();
+  }
+  // TODO: JSON input validation? Maybe we don't need it since it's only coming from task manager?
+  const { input, ruleContent, type } = JSON.parse(jsonInput);
+  const createdInferredEntity = await createInferredEntity(context, input, ruleContent, type);
+  return createdInferredEntity.id;
+};
--- a/opencti-platform/opencti-graphql/src/manager/ruleManager.ts
+++ b/opencti-platform/opencti-graphql/src/manager/ruleManager.ts
@@ -6,7 +6,7 @@ import { createStreamProcessor } from '../database/stream/stream-handler';
 import { redisGetManagerEventState, redisSetManagerEventState } from '../database/redis';
 import { lockResources } from '../lock/master-lock';
 import conf, { booleanConf, logApp } from '../config/conf';
-import { createEntity, patchAttribute, stixLoadById, storeLoadByIdWithRefs } from '../database/middleware';
+import { createEntity, createInferredRelation, createInferredEntity, patchAttribute, stixLoadById, storeLoadByIdWithRefs } from '../database/middleware';
 import { EVENT_TYPE_CREATE, EVENT_TYPE_DELETE, EVENT_TYPE_MERGE, EVENT_TYPE_UPDATE, isEmptyField, isNotEmptyField, READ_DATA_INDICES } from '../database/utils';
 import { ABSTRACT_STIX_RELATIONSHIP, RULE_PREFIX } from '../schema/general';
 import { ENTITY_TYPE_RULE_MANAGER } from '../schema/internalObject';
@@ -15,7 +15,7 @@ import { getParentTypes } from '../schema/schemaUtils';
 import { isBasicRelationship, isStixRelationship } from '../schema/stixRelationship';
 import { isStixSightingRelationship } from '../schema/stixSightingRelationship';
 import { internalLoadById, fullRelationsList } from '../database/middleware-loader';
-import type { RuleDefinition, RuleRuntime, RuleScope } from '../types/rules';
+import type { CreateInferredEntityCallbackFunction, CreateInferredRelationCallbackFunction, RuleDefinition, RuleRuntime, RuleScope } from '../types/rules';
 import type { BasicManagerEntity, BasicStoreBase, BasicStoreCommon, BasicStoreEntity, BasicStoreRelation, StoreObject } from '../types/store';
 import type { AuthContext, AuthUser } from '../types/user';
 import type { RuleManager } from '../generated/graphql';
@@ -81,7 +81,6 @@ const isAttributesImpactDependencies = (rule: RuleDefinition, operations: Operat
     .map((a) => a.name);
   const operationAttributes = R.uniq(operations.map((o) => {
     const parts = o.path.substring(1).split('/');
-
     return parts.filter((p) => isNaN(Number(p))).join('.');
   }));
   return operationAttributes.filter((f) => rulesAttributes.includes(f)).length > 0;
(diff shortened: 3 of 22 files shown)


opencti-platform/opencti-graphql/src/manager/taskManager.js
```diff
@@ -497,6 +504,56 @@ const sharingOperationCallback = async (context, user, task, actionType, operati
...
+const ruleApplyCallback = async (context, user, task, ruleId) => {
+  return async (elements) => {
+    for (let index = 0; index < elements.length; index += 1) {
+      const inferredObjectsBundle = [];
+      const element = elements[index];
+      // in case of rule apply we need to fake the apply of the rule and get the resulting inferred creations
+      const inferredRelations = [];
+      const inferredEntities = [];
+      const createInferredRelationOverride = (_context, input, ruleContent, opts) => {
+        inferredRelations.push({ input, ruleContent, opts });
```
richard-julien: Accumulate everything before sending bundle could cause on OOM. Can you change that to send bundle to queue incrementaly ?

opencti-platform/opencti-graphql/config/schema/opencti.graphql
```diff
@@ -14889,4 +14889,8 @@ type Mutation {
   stixSightingRelationshipAdd(input: StixSightingRelationshipAddInput!): StixSightingRelationship
   @auth(for: [KNOWLEDGE_KNUPDATE])
   stixSightingRelationshipEdit(id: ID!): StixSightingRelationshipEditMutations @auth(for: [KNOWLEDGE_KNUPDATE])
+
+  ######## INTERNAL INFERRED CREATION
+  inferredRelationAdd(jsonInput: String!): ID @auth(for: [KNOWLEDGE_KNUPDATE])
```
richard-julien, line 15557: Maybe we can protect more as it must only be created by the rule user?

opencti-platform/opencti-graphql/src/domain/inferredObject.ts
```diff
@@ -0,0 +1,13 @@
+import { createInferredEntity, createInferredRelation } from '../database/middleware';
+import type { AuthContext, AuthUser } from '../types/user';
+
+export const createInternalInferredRelation = async (context: AuthContext, _user: AuthUser, jsonInput: string) => {
+  const { input, ruleContent, opts } = JSON.parse(jsonInput);
```
richard-julien, line 11: We should validate that the json is correct for the creation

opencti-platform/opencti-graphql/src/manager/ruleManager.ts
```diff
@@ -158,7 +158,14 @@ const applyCleanupOnDependencyIds = async (deletionIds: Array<string>, rules: Ar
   await elList(context, RULE_MANAGER_USER, READ_DATA_INDICES, opts);
 };
 
-export const rulesApplyHandler = async (context: AuthContext, user: AuthUser, events: Array<DataEvent>, forRules: Array<RuleRuntime> = []) => {
+export const rulesApplyHandler = async (
+  context: AuthContext,
+  user: AuthUser,
+  events: Array<DataEvent>,
+  forRules: Array<RuleRuntime> = [],
+  createInferredEntityOverride?: createInferredEntityOverrideFunction | undefined,
```
richard-julien: Maybe creationCallback could be a better semantic?

opencti-platform/opencti-graphql/config/schema/opencti.graphql
JeremyCloarec: I used the same capability as for ruleApply, but I can change it to a more restrictive capability

JeremyCloarec: I added a new protection on the domain layer: I only accept calls made by the rule user

opencti-platform/opencti-graphql/src/rules/containerWithRefsBuilder.ts
```diff
@@ -87,7 +101,7 @@ const buildContainerRefsRule = (ruleDefinition: RuleDefinition, containerType: s
       }
     }
     // endregion
-    if (createdTargets.length > 0 || deletedTargetRefs.length > 0) {
+    if ((createdTargets.length > 0 || deletedTargetRefs.length > 0) && !createInferredRelationCallback) {
```
JeremyCloarec: Only drawback where I can't find a solution: we no longer send update event on report to stream... Need to find a way to do this: add a publishToStream operation in bundle?

opencti-platform/opencti-graphql/src/domain/inferredObject.ts
```diff
@@ -0,0 +1,32 @@
...
+import { ForbiddenAccess } from '../config/errors';
+import { publishStixToStream } from '../database/redis';
+
+export const createInternalInferredRelation = async (context: AuthContext, user: AuthUser, jsonInput: string) => {
+  // This API should only be available to task manager user
+  if (user.id !== RULE_MANAGER_USER.id) {
+    throw ForbiddenAccess();
+  }
+  // TODO: JSON input validation? Maybe we don't need it since it's only coming from task manager?
+  const { input, ruleContent, opts } = JSON.parse(jsonInput);
```
xfournet, line 11: would it possible to have these structure in the GraphQL schema ? or it's a bad idea ?

opencti-platform/opencti-graphql/src/rules/indicate-sighted/IndicateSightedRule.ts
```diff
@@ -13,14 +13,18 @@ import { createInferredRelation, deleteInferredRuleElement } from '../../databas
...
 import { executionContext, RULE_MANAGER_USER } from '../../utils/access';
 import type { AuthContext } from '../../types/user';
 
 const indicateSightedRuleBuilder = (): RuleRuntime => {
   // Execution
-  const applyFromStixRelation = async (context: AuthContext, data: StixRelation): Promise<void> => {
+  const applyFromStixRelation = async (
+    context: AuthContext,
+    data: StixRelation,
+    createInferredRelationCallback?: createInferredRelationCallbackFunction | undefined
```
xfournet: can be simplified with
`createInferredRelationCallback: createInferredRelationCallbackFunction = createInferredRelation`

so then createInferredRelationCallback can be directly called
same apply for other methods/rules

but in fact it can be probably resolved at an higher level in rulesApplyHandler

opencti-platform/opencti-graphql/src/types/rules.d.ts
```diff
@@ -52,9 +53,16 @@ interface RuleDefinition {
   scopes: Array<RuleScope>;
 }
 
+type createInferredRelationCallbackFunction = (context: AuthContext, input: any, ruleContent: any, opts?: any | undefined) => Promise<void>;
+type createInferredEntityCallbackFunction = (context: AuthContext, input: any, ruleContent: any, type: string) => Promise<void>;
```
xfournet: types should have first letter in upper case

opencti-platform/opencti-graphql/src/rules/indicate-sighted/IndicateSightedRule.ts
```diff
@@ -103,21 +119,28 @@ const indicateSightedRuleBuilder = (): RuleRuntime => {
...
   };
   // Contract
   const clean = async (element: StoreObject, deletedDependencies: Array<string>): Promise<void> => {
     await deleteInferredRuleElement(def.id, element, deletedDependencies);
   };
-  const insert = async (element: StixRelation): Promise<void> => {
-    return applyUpsert(element);
+  const insert = async (
+    element: StixRelation,
+    _createInferredEntityCallback?: createInferredEntityCallbackFunction | undefined,
```
xfournet: it look likes we must ignore some parameters in many places, it would be more readable to have an object that contains the different methods

moreover we are replicating the method signature in many places (it was already the case before this change and also apply to other methods like clean & update). It would be more efficient to reuse types, either:
* typing insert : `  const insert: RuleRuntime['insert'] = async (element, _createInferredEntityCallback, createInferredRelationCallback) => {
` 
* inlining insert in the return so it can be infered by TS:
```
  return {
    ...def,
    insert: async (element, _createInferredEntityCallback, createInferredRelationCallback) => applyUpsert(element, createInferredRelationCallback),
    update: async (element) => applyUpsert(element),
    clean: async (element: StoreObject, deletedDependencies: Array<string>) => {
      await deleteInferredRuleElement(def.id, element, deletedDependencies);
    },
  };

```

opencti-platform/opencti-graphql/src/manager/ruleManager.ts
```diff
@@ -158,7 +158,14 @@ const applyCleanupOnDependencyIds = async (deletionIds: Array<string>, rules: Ar
...
 };
 
-export const rulesApplyHandler = async (context: AuthContext, user: AuthUser, events: Array<DataEvent>, forRules: Array<RuleRuntime> = []) => {
+export const rulesApplyHandler = async (
+  context: AuthContext,
+  user: AuthUser,
+  events: Array<DataEvent>,
+  forRules: Array<RuleRuntime> = [],
+  createInferredEntityCallback?: createInferredEntityCallbackFunction | undefined,
+  createInferredRelationCallback?: createInferredRelationCallbackFunction | undefined
```
xfournet: default values can be resolved here, so it rules implementation will be simple and avoid repetition of default case handling

```
  createInferredEntityCallback: createInferredEntityCallbackFunction = createInferredEntity,
  createInferredRelationCallback: createInferredRelationCallbackFunction = createInferredRelation
```

opencti-platform/opencti-graphql/src/manager/taskManager.js
```diff
@@ -265,16 +266,14 @@ const sendResultToQueue = async (context, user, task, objects, opts = {}) => {
     no_split: opts.forceNoSplit ?? false
   });
 };
-
-const buildBundleElement = (element, actionType, operations) => {
+const buildBaseBundleElement = (element, standard_id) => {
```
xfournet: the two usage of this method add additional information in extensions[STIX_EXT_OCTI], it could directly added here via a second args that will receive the additional information to be added for this ext

opencti-platform/opencti-graphql/src/manager/taskManager.js
xfournet, line 1: To make sure I understand correctly:

In the new approach, bundles are split using RULE_APPLY_MAX_BUNDLE_SIZE. The same logic should also be applied to standardOperationCallback, which currently doesn’t split and creates a single (potentially very large) bundle.
Adding queues introduces an extra step and increases code complexity.

Is this additional complexity intended to take advantage of the worker queues’ consumption regulation?

JeremyCloarec: standardOperationCallback bundles are already split (at least in query tasks) during the elList with MAX_TASK_ELEMENTS (it can be found in the opts in taskQuery).

As for your question, just to make sure I understand it correctly also: you're asking specifically about the new RULE_APPLY_MAX_BUNDLE_SIZE, and not the overall design of sending inferred objects in the bundle right?
The RULE_APPLY_MAX_BUNDLE_SIZE was used to prevent very large bundle from being sent to the worker, since a single ruleApply call can lead to a very large amount of inferred data. Only sending a bundle at the end of the ruleApply could lead to a bundle having >100k objects, which in theory could be handled by the worker, but it would put unnecessary load on the splitting of the bundle

opencti-platform/opencti-graphql/src/manager/ruleManager.ts
xfournet: need to add createInferredEntityCallback / createInferredRelationCallback

opencti-platform/opencti-graphql/src/manager/taskManager.js
```diff
@@ -285,6 +285,15 @@ const buildBundleElement = (element, actionType, operations) => {
   }
   return baseObject;
 };
+const buildBundleElement = (element, actionType, operations) => {
+  const additionalExtensions = baseOperationBuilder(actionType, operations, element);
+  const bundleObject = buildBaseBundleElement(element, additionalExtensions);
+  bundleObject.extensions[STIX_EXT_OCTI] = {
+    ...bundleObject.extensions[STIX_EXT_OCTI],
+    ...baseOperationBuilder(actionType, operations, element)
+  };
```
xfournet: It should be removed since it has already been done

opencti-platform/opencti-graphql/src/domain/inferredObject.ts
JeremyCloarec: Since it's only used internally for now, and since the validation isn't that easy to implement, we'll skip it until it's used externally

opencti-platform/opencti-graphql/src/domain/inferredObject.ts
JeremyCloarec: Since it's only used internally for now, and since the validation isn't that easy to implement, we'll skip it until it's used externally

Issue #11519, opened 2025-06-26
Improve e2e tests


aHenryJard:
## Description

e2e test issues and flakes.
Feel free to add screenshot of flake if you got some.

## Environment

1. OS (where OpenCTI server runs): { e.g. Mac OS 10, Windows 10, Ubuntu 16.4, etc. }
2. OpenCTI version: { e.g. OpenCTI 1.0.2 }
3. OpenCTI client: { e.g. frontend or python }
4. Other environment details:

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Have a build failing on e2e step
2. Re-Run it's green

PR #11645, opened 2025-07-10
[ci] improve e2e dashboard and improve retry support and local run (#11519)

aHenryJard:
### Proposed changes

* Allow some test to retry without require to cleanup data (better chances that drone retries works)
    * make sure that dashboard can be run several times
    * make sure that enableReference can run several times
    * always login as admin for start, if we allow retry, we also allow to fail in the middle of a login or with another user.
       * it's not really faster, and locally half the time I have to delete the auth files
    * and then move logout to the end
* Attempt to fix dashboard flake.
   * moved the restricted test to a dedicated spec, because was making dashboard CRUD failing also 

### Related issues

* relates to #11519 

### Checklist


opencti-platform/opencti-front/tests_e2e/dashboard/dashboard.spec.ts
```diff
@@ -214,7 +220,7 @@ test('Dashboard CRUD', async ({ page }) => {
   await dashboardDetailsPage.getEditButton().click();
   await dashboardDetailsPage.getDeleteButton().click();
   await dashboardDetailsPage.getConfirmButton().click();
-
+  await page.waitForTimeout(1000);
```
aHenryJard:
```suggestion
  await page.waitForTimeout(1000); // After delete need to wait a bit
```

lndrtrbn (approved):

PR #11648, opened 2025-07-10
[frontend] prevent backdrop interaction when filter popover is open (#11594)

### LINKED ISSUE #11594: Widgets in background are still movable when editing a filter
## Description

When editing a filter, widgets in the background can still be moved around. This can be an issue when selecting some text but missclicking and moving a widget by accident

## Environment

1. Ubuntu LTS 22.04
2. Opencti 6.7.2
4. Running on a proxmox server

## Reproducible Steps

1. On a dasboard, have two widgets next to each other
2. Go to the filter tabs, click on one of the filter to display a text entry
3. While on the popup, move the other

## Expected Output

The widgets would not move

## Actual Output

The widgets are moving

## Screenshots (optional)

Image

maelv-filigran:
### Proposed changes

* Fixed backdrop interaction issue in FilterChipPopover component where users could interact with elements behind the popup when the filter selection popup was open
* Resolved React warning about keys being passed through spread props in Autocomplete's renderOption

### Related issues

* #11594
* Fixes issue where focusing on the Autocomplete input field within the filter popup did not prevent interaction with backdrop elements
* Addresses nested popup behavior when FilterChipPopover is opened from within another modal/dialog

### Checklist

### Further comments

The main issue was that the FilterChipPopover component, when opened from within another popup (nested popups scenario), was not properly blocking interactions with the backdrop. This allowed users to interact with elements behind the popup when clicking outside of it, particularly when the Autocomplete component was focused.

**Solution implemented:**
- Modified the Popover component to explicitly handle backdrop behavior using `hideBackdrop={false}` and custom backdrop props
- Added proper event handlers (`onClick` and `onMouseDown`) to the backdrop to prevent event propagation while maintaining the ability to close the popup
- Fixed a secondary issue where React keys were being passed through spread props in the Autocomplete's renderOption, which was causing React warnings

### DIFF
--- a/opencti-platform/opencti-front/src/components/filters/FilterChipPopover.tsx
+++ b/opencti-platform/opencti-front/src/components/filters/FilterChipPopover.tsx
@@ -401,6 +401,22 @@ export const FilterChipPopover: FunctionComponent<FilterChipMenuProps> = ({
           onChange={(event) => handleChangeOperator(event, finalFilterDefinition)}
           style={{ marginBottom: 15 }}
           disabled={disabled}
+          MenuProps={{
+            // Force MUI to use a backdrop
+            hideBackdrop: false,
+            BackdropProps: {
+              style: {
+                // Make the backdrop invisible because we already have one
+                backgroundColor: 'rgba(0, 0, 0, 0)',
+              },
+              // Prevent clicks from going through
+              onClick: (e) => {
+                stopEvent(e);
+                handleClose();
+              },
+              onMouseDown: stopEvent,
+            },
+          }}
         >
           {availableOperators.map((value) => (
             <MenuItem key={value} value={value}>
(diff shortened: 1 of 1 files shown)


opencti-platform/opencti-front/src/components/filters/FilterChipPopover.tsx
```diff
@@ -102,6 +102,43 @@ export const FilterChipPopover: FunctionComponent<FilterChipMenuProps> = ({
...
   const { typesWithFintelTemplates } = useAttributes();
+  const popoverContentRef = React.useRef<HTMLDivElement>(null);
+
+  React.useEffect(() => {
+    if (!open) return;
+
+    const handleMouseDown = (e: MouseEvent) => {
+      const target = e.target as HTMLElement;
+
+      // Vérifier si le click est dans notre popover spécifique
```
aHenryJard: Could you please use English comment in code ?

maelv-filigran: of course :)

lndrtrbn (commented):
Works well when manipulating the autocomplete of values.
But still have the issue when manipulating the autocomplete of operator.
image

maelv-filigran:
@lndrtrbn good catch! The issue is now corrected also on the select operator component (same fix)

lndrtrbn (approved):

Issue #11246, opened 2025-06-09
Revamp sharing organization


alice-debra:
## Use case

Organization sharing should work like subscribers with a drawer.

An intuitive sharing interface should offer at-a-glance comprehension of access permissions, with ordered lists that make it effortless to manage entities.

labo-flg:
I have opened https://github.com/OpenCTI-Platform/opencti/pull/12288 to reopen the subject and find a proper solution.
In this stage, there are too much limitations and bugs that we need  to tackle.
We're back to the drawing board on this.

PR #11653, opened 2025-07-10
[frontend] New shared Organisations drawer (#11246)

delemaf:
### Proposed changes

* Add new sharing organisations button and counter
image

* Add new sharing organisations drawer
image

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/11246

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/common/containers/StixCoreObjectSharedOrganisationsDrawer.tsx
+++ b/opencti-platform/opencti-front/src/private/components/common/containers/StixCoreObjectSharedOrganisationsDrawer.tsx
@@ -1,0 +1,218 @@
+import React, { FunctionComponent, Suspense, useState } from 'react';
+import { PreloadedQuery, useFragment, usePreloadedQuery } from 'react-relay';
+import {
+  StixCoreObjectSharingListFragment$data,
+  StixCoreObjectSharingListFragment$key,
+} from '@components/common/stix_core_objects/__generated__/StixCoreObjectSharingListFragment.graphql';
+import { objectOrganizationFragment } from '@components/common/stix_core_objects/StixCoreObjectSharingList';
+import { useFormatter } from '../../../../components/i18n';
+import ListLines from '../../../../components/list_lines/ListLines';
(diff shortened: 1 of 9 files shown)


CelineSebe:
the loader is blocked after sharing orga:
image

CelineSebe:
I can select the same orga several times:
image

Archidoit:
Design comments in the Add organization panel

- if there is no author, the convention is to display author as '-' instead of empty
- there is '..' at the end of each line

image

Archidoit:
> the loader is blocked after sharing orga: image

@CelineSebe did you run your worker?

Archidoit:
I have still the old design when sharing an observable with an organization : 
image

Archidoit:
The sharing is also different for sightings sharing : 
image

Archidoit:
When sharing/unsharing entities with organization from the toolbar (background task), it's also the old design. Maybe consider opening a drawer like it is done for bulk update via the tool bar

image

opencti-platform/opencti-front/src/private/components/common/containers/StixCoreObjectSharedOrganisationsDrawer.tsx
```diff
@@ -0,0 +1,212 @@
...
+  };
+  const keyword = (searchTerm ?? '').length === 0 ? selectedText : searchTerm;
+  const buildColumns = () => {
+    return {
+      entity_type: {
+        label: 'Type',
+        width: '15%',
+        isSortable: true,
+      },
+      value: {
```
Archidoit, line 141: Organization name should be sortable. Maybe if you display the organization name instead of its value if value is not sortable. (See organizations list, the name is sortable)

delemaf:
> The sharing is also different for sightings sharing : image

This case was in the initial scope, it needs more reflexion and work since it's not the same pattern.

delemaf:
> When sharing/unsharing entities with organization from the toolbar (background task), it's also the old design. Maybe consider opening a drawer like it is done for bulk update via the tool bar
> 
> image

Same thing for this one, it needs a bit more relexion and work in another PR.

marieflorescontact (approved):
locally tested ok 👌

PR #11656, opened 2025-07-11
[frontend] feat(markdown): add breakWords prop for long string handling (#10430)

### LINKED ISSUE #10430: Description Field Truncates Long Unbroken Strings in Relationship Objects
## Description

When displaying the description field within the context of a relationship (e.g., between a Threat Actor, Intrusion Set, or Campaign and an Attack Pattern), OpenCTI truncates long unbroken strings (such as Windows Registry paths or URLs). 

## Environment

SaaS

## Reproducible Steps

1. Create a relationship (e.g., Threat Actor → Attack Pattern).
2. Add a detailed description containing a long unbroken strings (e.g., Windows Registry paths).
3. View the relationship in the UI.
4. Observe how the description field is truncated or poorly formatted.

## Expected Output

Long unbroken strings should either wrap correctly without truncation, or be displayed in a scrollable or expandable text block.

## Actual Output

Long unbroken strings are truncated. 

## Screenshots:

Image
Image

maelv-filigran:
### Proposed changes

* Add optional `breakWords` prop to MarkdownDisplay component to control text wrapping behavior for long strings
* Implement CSS styles 
* Default behavior remains unchanged (truncation) to maintain backward compatibility

### Related issues

* #10430

### Checklist

### Further comments

By default, the component maintains its current behavior (truncating long strings), but when `breakWords={true}` is passed, it will wrap long strings naturally instead of truncating them.

This is particularly useful for displaying URLs, long technical identifiers, or any content where preserving the full text is more important than maintaining strict layout constraints.

The implementation uses standard CSS properties for maximum browser compatibility

### DIFF
--- a/opencti-platform/opencti-front/src/components/MarkdownDisplay.tsx
+++ b/opencti-platform/opencti-front/src/components/MarkdownDisplay.tsx
@@ -57,6 +57,7 @@ interface MarkdownWithRedirectionWarningProps {
   remarkPlugins?: MarkdownProps['remarkPlugins'];
   emptyStringIfUndefined?: boolean;
   disableWarningAtLinkClick?: boolean;
+  breakWords?: boolean;
 }
 
 const MarkdownDisplay: FunctionComponent<
@@ -73,6 +74,7 @@ MarkdownWithRedirectionWarningProps
   remarkPlugins,
   emptyStringIfUndefined,
   disableWarningAtLinkClick,
+  breakWords = true,
 }) => {
   const theme = useTheme<Theme>();
   const [displayExternalLink, setDisplayExternalLink] = useState(false);
@@ -90,9 +92,17 @@ MarkdownWithRedirectionWarningProps
   if (removeLineBreaks) {
     disallowedElements.push('p');
   }
+  const markdownStyle: React.CSSProperties = breakWords
+    ? {
+      overflowWrap: 'break-word',
+      wordBreak: 'break-word',
+      hyphens: 'auto',
+    }
+    : {};
+
   const markdownElement = () => {
     return (
-      <div className="markdown">
+      <div style={markdownStyle}>
         <Markdown
           disallowedElements={disallowedElements}
           unwrapDisallowed={true}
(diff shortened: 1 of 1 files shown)


opencti-platform/opencti-front/src/components/MarkdownDisplay.tsx
```diff
@@ -90,9 +92,17 @@ MarkdownWithRedirectionWarningProps
...
+      overflowWrap: 'break-word',
+      wordBreak: 'break-word',
+      hyphens: 'auto',
+    }
+    : {};
+
   const markdownElement = () => {
     return (
-      <div className="markdown">
+      <div className="markdown" style={markdownStyle}>
```
Archidoit: className is useless here?

Archidoit:
It's resolved for Relationships description. But same problem for other entities description (exemple below for a report description). 
----> Maybe we want breakWords to be true by default.

image

Archidoit:
You can also consider adding the arrow the develop long descriptions (present for Reports and other entities description, but not for relationships descriptions).

maelv-filigran:
@alice-debra As suggested by @Archidoit, should long unbroken strings always wrap without truncation by default, or should we apply this approach only in specific cases?

alice-debra:
I think it's a good idea, but we need something to show users when the world breaks.

alice-debra:
Maybe you can ask for Sandy's input?

alice-debra:
Break the world in all descriptions, without visual indication.

opencti-platform/opencti-front/src/components/MarkdownDisplay.tsx
maelv-filigran: indeed, there is no .markdown class in the application's style sheet. I'll remove this!

opencti-platform/opencti-front/src/components/MarkdownDisplay.tsx
```diff
@@ -73,6 +74,7 @@ MarkdownWithRedirectionWarningProps
   remarkPlugins,
   emptyStringIfUndefined,
   disableWarningAtLinkClick,
+  breakWords = true,
```
Archidoit: if this props is never used, I suggest to remove it and always give the new styling to the div(s)

Archidoit (approved):

Issue #10831, opened 2025-04-24
Add Drafts into Data tab of entities like workbench


nino-filigran:
## Use case

With the appearance of drafts in the platform, we want to add a dedicated space within the entities where user can also create drafts directly within an entity, like it exists for workbench. 

## Proposed Solution

Add a section "Drafts" in each Data Tab of any entities.

PR #11698, opened 2025-07-17
[frontend] Have a Draft section in Data section of each entities (like workbench) (#10831)

SarahBocognano:
### Proposed changes

* Add draft section in Data section of each entities

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/10831

### Checklist


opencti-platform/opencti-front/src/private/components/common/files/draftWorkspace/DraftWorkspaceDialogCreation.tsx
```diff
@@ -0,0 +1,133 @@
...
+  );
+
+  const updater = (store: RecordSourceSelectorProxy) => insertNode(
+    store,
+    'Pagination_draftWorkspaces',
+    paginationOptions,
+    'draftWorkspaceAdd',
+  );
+
+  const draftValidation = () => Yup.object().shape({
```
lndrtrbn: No need to use a function

opencti-platform/opencti-front/src/private/components/common/files/draftWorkspace/DraftWorkspaceViewer.tsx
```diff
@@ -0,0 +1,72 @@
...
+    <div style={{
+      display: 'table',
+      height: '100%',
+      width: '100%',
+      textAlign: 'center',
+    }}
+    >
+      {t_i18n('No draft for the moment')}
+    </div>
+  );
```
lndrtrbn: I would prefer having a separate component in a dedicated file, under datatable folder for example. This component take the title as props and so we can reuse it anywhere if needed

opencti-platform/opencti-front/src/private/components/drafts/DraftContextBanner.tsx
```diff
@@ -202,14 +202,13 @@ const DraftContextBannerComponent: FunctionComponent<DraftContextBannerComponent
...
   const [queryRef, loadQuery] = useQueryLoader<DraftContextBannerQuery>(draftContextBannerQuery);
-
   if (!draftContext) {
     return null;
   }
 
   useEffect(() => {
     loadQuery({ id: draftContext.id }, { fetchPolicy: 'store-and-network' });
-  }, []);
+  }, [draftContext.name]);
```
lndrtrbn: Why are we depending of the name of the context but we do not use it in the useEffect?

opencti-platform/opencti-front/src/private/components/drafts/Drafts.tsx
```diff
@@ -113,14 +118,25 @@ const computeValidationProgress = (validationWork: Drafts_node$data['validationW
   return `${Math.floor(100 * (validationWork.tracking.import_processed_number / validationWork.tracking.import_expected_number))}%`;
 };
 
-const Drafts: React.FC = () => {
+interface DraftProps {
```
lndrtrbn: rename DraftsProps

opencti-platform/opencti-front/src/private/components/drafts/Drafts.tsx
```diff
@@ -203,32 +220,86 @@ const Drafts: React.FC = () => {
...
+        )}
+        <DraftWorkspaceDialogCreation
+          paginationOptions={queryPaginationOptions}
+          handleCloseCreate={setOpenCreate}
+          entityId={entityId}
+          openCreate={openCreate}
+        />
+      </>
+    );
+  };
```
lndrtrbn: I find this component a bit complicated the fact it has to manage 2 different usages.
It would be preferable to have a common component for the datatable/queries and then 2 separates components for each context (full page of in entity)

opencti-platform/opencti-front/src/private/components/common/files/draftWorkspace/DraftWorkspaceViewer.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+  return (
+    <Grid item xs={6}>
+      <div style={{ height: '100%' }}>
+        <Typography variant="h4" gutterBottom={true} style={{ float: 'left', marginBottom: theme.spacing(2) }}>
```
lndrtrbn: Can't we avoid the floating left and arbitrary margins? Ping me if needed

JeremyCloarec: Margins are removed, difficult to avoid the floating left though

opencti-platform/opencti-front/src/private/components/drafts/DraftContextBanner.tsx
SarahBocognano: Because when changing the draft context, the name of the banner doesn't reload itself. We need to reload the query when the name is changing

opencti-platform/opencti-front/src/private/components/drafts/Drafts.tsx
SarahBocognano: Agree in theory, but we kinda want to merged it asap, however when merged I'll try to rework on it to refactor and have a common component

opencti-platform/opencti-front/src/private/components/common/files/draftWorkspace/DraftWorkspaceViewer.tsx
SarahBocognano: cannot avoid the floating left without breaking everything, but we did remove the marginBottom though ! :D

JeremyCloarec (approved):

Issue #7328, opened 2024-06-09
[Composer] Be able to change connector configuration from the OpenCTI UI


SamuelHassine:
## Use case

Be able to change connector configuration from the OpenCTI UI

damians-filigran:
Note that this should include secure best-practice secrets management of API keys, etc. as part of this issue. See [here](https://www.notion.so/Sentinel-DIscussion-1ff8fce17f2a80daab9bedd849a5ad51?d=1ff8fce17f2a80bb9962001c3b4a4b56&pvs=4#1ff8fce17f2a80a6a11adeb48523f2b1) for details.

PR #11703, opened 2025-07-17
[backend/frontend] xtm composer - get and use manifest (#7328)

frapuks:
### Proposed changes

* New Catalog page (data/ingestion/catalog)
* New Connector page (data/ingestion/catalog/NAME)
* New Form to deploy
* Use the new manifest (generated by Integration team)

### Related issues

* #7328 

### Checklist


maelv-filigran:
I updated the tests to take into account the changes to the manifest and GraphQL.

opencti-platform/opencti-front/src/private/components/data/IngestionCatalog/IngestionCatalogConnectorCreationUserHandling.tsx
```diff
@@ -42,7 +42,7 @@ const IngestionCatalogConnectorCreationUserHandlingComponent = ({ queryRef, max_
       'user_id',
       values.automatic_user === false
         ? ''
-        : { label: `[F] ${values.name}`, value: `[F] ${values.name}` },
+        : { label: `[C] ${values.name}`, value: `[F] ${values.name}` },
```
maelv-filigran: value should be also `[C] ${values.name}`

opencti-platform/opencti-front/src/private/components/data/IngestionCatalog/IngestionCatalogCard.tsx
```diff
@@ -0,0 +1,162 @@
...
+    color: 'success',
+  },
+};
+
+const IngestionCatalogCard = ({ node, dataListId }: IngestionCatalogCardProps) => {
+  const { t_i18n } = useFormatter();
+  const navigate = useNavigate();
+  const theme = useTheme<Theme>();
+  const [openCreation, setOpenCreation] = useState(false);
+  const connector: IngestionConnector = JSON.parse(node);
```
maelv-filigran: I think JSON.parse should be at the parent component level :
Misplaced responsibility: A presentational component (Card) should not perform data transformation.
Computational cost: JSON.parse() is synchronous and blocking; 50 connector cards mean 50 executions.
Repeated parsing: Re-rendering / filtering components triggers repeated parsing.
Impact on the main thread: May cause UI micro-freezes, particularly on less powerful devices.

Maybe add Error Handling -> A single malformed connector definition crashes the entire catalog page

opencti-platform/opencti-front/src/private/components/data/IngestionCatalog/IngestionCatalogConnectorCreationUserHandling.tsx
frapuks: fixed

maelv-filigran:
Not mandatory but you may want to consider this: there is a tight coupling between UI components and the data structure.

You could add a data access layer or create custom hooks such as:

```
// Create custom hooks
const useConnectorCatalog = () => {
  // Handle data fetching, parsing, and state management
};

const useCatalogDeployment = () => {
  // Handle deployment logic
};
```

opencti-platform/opencti-front/src/private/components/data/IngestionCatalog/IngestionCatalogConnectorCreation.tsx
```diff
@@ -0,0 +1,234 @@
...
+          automatic_user: true,
+          ...connector.default,
+        }}
+        onSubmit={() => {
+        }}
+      >
+        {({ values, setFieldValue, isSubmitting, setSubmitting, resetForm, isValid }) => {
+          // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+          // @ts-ignore
+          if (!compiledValidator || compiledValidator.schema.container_image !== connector.container_image) {
```
maelv-filigran: This code is executed directly within the render function, causing an infinite loop:
- The component renders.
- The condition is true, leading to calls to setCompiledValidator and setFieldValue.
- These calls change the state, triggering a re-render.
- The cycle repeats indefinitely.

Solution: Move this code into a useEffect so it only executes when dependencies change, not on every render.

maelv-filigran:
if I create a new connector with the default confidence level I have this error
image

opencti-platform/opencti-graphql/src/domain/connector.ts
```diff
@@ -178,7 +179,16 @@ export const managedConnectorAdd = async (
   user:AuthUser,
   input: AddManagedConnectorInput
 ) => {
-  const connectorUser: any = await storeLoadById(context, user, input.connector_user_id, ENTITY_TYPE_USER);
+  if (input.user_id.length < 2) {
+    throw FunctionalError('You have not chosen a user responsible for data creation', {});
+  }
+  let finalUserId = input.user_id;
+  if (input.automatic_user) {
+    const onTheFlyCreatedUser = await createOnTheFlyUser(context, user, { userName: input.user_id, confidenceLevel: input.confidence_level });
```
SouadHadjiat, line 198: so user_id can be a name ?

opencti-platform/opencti-graphql/src/domain/connector.ts
```diff
@@ -178,7 +179,16 @@ export const managedConnectorAdd = async (
...
+    finalUserId = onTheFlyCreatedUser.id;
+  }
+
+  const connectorUser = await storeLoadById(context, user, finalUserId, ENTITY_TYPE_USER);
```
SouadHadjiat, line 201: I think that targetContract should be first checked before trying to create a user on the fly

opencti-platform/opencti-graphql/src/domain/connector.ts
```diff
@@ -178,7 +179,16 @@ export const managedConnectorAdd = async (
...
   if (isEmptyField(connectorUser)) {
     throw UnsupportedError('Connector user not found');
   }
```
SouadHadjiat, line 204: Do we check somewhere that the connector contract is "manager_supported" ?

opencti-platform/opencti-graphql/src/domain/connector.ts
frapuks: Yes...
This part of the code is inherited from the CSV feed, so I didn't question what had been done.
It currently works. Should I change this? If so, can I change it in the next PR?

opencti-platform/opencti-graphql/src/domain/connector.ts
frapuks: fixed

opencti-platform/opencti-graphql/src/domain/connector.ts
frapuks: fixed

opencti-platform/opencti-graphql/src/domain/connector.ts
frapuks: It will be fixed into th next chunk

frapuks:
Ready for second cycle of review

opencti-platform/opencti-front/script/get-connectors-config.js
```diff
@@ -0,0 +1,50 @@
+import { compile } from 'json-schema-to-typescript';
+import fs from 'fs/promises';
+import path from 'path';
+import _ from 'lodash';
+
+const SCHEMA_URL = 'https://raw.githubusercontent.com/OpenCTI-Platform/connectors/refs/heads/oob/connector_manager/manifest.json'; // TODO set SCHEMA_URL in parameters
```
maelv-filigran: Maybe something like this for the next PR
const SCHEMA_URL = process.env.CONNECTOR_SCHEMA_URL || config.get('connector:schema_url') || 'https://raw.githubusercontent.com/OpenCTI-Platform/connectors/refs/heads/main/connector_manager/manifest.json';

frapuks: This file is a work-in-progress. It works, but the result is not used. We need to discuss about it in the next chunk with Natalie.
I can remove it from this PR and replace it on the next PR, if it's better.

natalie-natsu: Can you just get the JSON and write it in sources instead of copy paste the manifest ? And specify that if you want to get the new manifest, you need to run the yarn script. :-)

maelv-filigran:
While testing this branch, I encountered an interesting runtime error that might be worth investigating:

```javascript
TypeError: Cannot read properties of undefined (reading 'container_version')
    at computeManagerConnectorImage (src/database/repository.js:59:96)
```

This error appears during the `GetConnectors` GraphQL query.  The code assumes contracts.get(cn.manager_contract_image) always returns a valid contract. But what if it doesn't?

1. __Have you tested scenarios where a connector's `manager_contract_image` might not have a corresponding contract in the catalog?__

   - What happens when `getSupportedContractsByImage()` doesn't contain a specific image?

2. __Could there be a timing issue or missing data scenario?__

   - Perhaps during connector initialization or when contracts haven't been loaded yet?

This could help make the code more resilient to edge cases. What do you think would be the appropriate behavior when a contract isn't found - should it return `null`, throw a specific error, or handle it differently?

natalie-natsu (approved):

PR #11769, opened 2025-07-22
[frontend] improve investigation rollbacks (#11565)

### LINKED ISSUE #11565: [Knowledge Graph] Unable to Expand Nodes in Investigation if the Investigation is Large
## Description
When attempting to expand nodes within an Investigation, the system locks up preventing the nodes from being expanded. This behavior is not consistent and only occurs when the Investigation is really large and you are attempting to expand the nodes that have a large number of targets and relationships.

## Environment
6.7.1

## Reproducible Steps

1. Navigate to an Investigation that has numerous nodes and edges
2. Select several nodes to expand
3. Expand the nodes selecting the targets and relationships that have several entities associated with them.
4. Observe that the platform locks up, forcing the user to exit the platform and return.

## Expected Output
Nodes are expanded, regardless of how many are selected. 

## Actual Output
Nodes are not expanded, and the user is forced to exit the platform and return.

lndrtrbn:
### Proposed changes

We have size limit issue with session storage with the feature of investigation rollbacks as we were saving the entire investigation state at each expansion.

* Refacto to save the operation instead

### Related issues

* #11565

### Checklist


ValentinBouzinFiligran:
looks good to me !
 ( difficult to test with a large graph )

ValentinBouzinFiligran (approved):

PR #11775, opened 2025-07-23
[backend] Display all internal connectors (#11666)

### LINKED ISSUE #11666: New background task in queues create a misunderstanding in the connectors screen
## Description

New background task in queues create a misunderstanding in the connectors screen:
Image

6K bundles in queues, all queues to 0.

SouadHadjiat:
### Proposed changes

* Get all internal queues to list internal connectors (background tasks, playbooks, syncs). We have a dedicated queue for each playbook and sync.
* Update internal queue names for better display in list

### Related issues

* #11666

### Checklist

### Further comments
image

### DIFF
--- a/opencti-platform/opencti-graphql/tests/02-integration/02-resolvers/connector-test.ts
+++ b/opencti-platform/opencti-graphql/tests/02-integration/02-resolvers/connector-test.ts
@@ -131,12 +131,13 @@ const DELETE_CONNECTOR_QUERY = gql`
 `;
 
 const TEST_CN_ID = '5ed680de-75e2-4aa0-bec0-4e8e5a0d1695';
+const TEST_CN_NAME = 'TestConnector';
 
 beforeAll(async () => {
   const CONNECTOR_TO_CREATE = {
     input: {
       id: TEST_CN_ID,
-      name: 'TestConnector',
+      name: TEST_CN_NAME,
       type: 'EXTERNAL_IMPORT',
       scope: 'Observable',
       auto: true,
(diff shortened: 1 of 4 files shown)


opencti-platform/opencti-graphql/tests/02-integration/02-resolvers/connector-test.ts
```diff
@@ -135,6 +155,19 @@ beforeAll(async () => {
 
 describe('Connector resolver standard behaviour', () => {
   let workId: string;
+  it('should list all connectors', async () => {
+    const queryResult = await queryAsUserWithSuccess(USER_CONNECTOR.client, { query: LIST_CONNECTORS_QUERY, variables: {} });
+    expect(queryResult.data.connectors).toBeDefined();
+    expect(queryResult.data.connectors.length).toEqual(7); // 1 created (TestConnector) + 6 built-in connectors (4 background tasks + import csv + draft validation)
```
aHenryJard: Could test some expected values instead of adding a count ? Something like =>
expect(connectors.find((connector) => connector.name === 'TestConnector')).toBeThruthy();

SouadHadjiat: well it's important to know exactly which connectors this API returns

SouadHadjiat: if we add a new built-in connector, we need to make sure that it's returned by this list. I'm going to make some changes to this test.

aHenryJard (commented):
Should this count be taken in account ?
https://github.com/OpenCTI-Platform/opencti/blob/6b2fea73534b7a25fdb6e5e8d9eb79ca26c54a58/opencti-platform/opencti-graphql/src/manager/taskManager.js#L76

I tried to change it locally to 15 and I still have 4 "task connector" in the list. But not sure if it's what I think it is.

SouadHadjiat:
> Should this count be taken in account ?
> 
> https://github.com/OpenCTI-Platform/opencti/blob/6b2fea73534b7a25fdb6e5e8d9eb79ca26c54a58/opencti-platform/opencti-graphql/src/manager/taskManager.js#L76
> 
> I tried to change it locally to 15 and I still have 4 "task connector" in the list. But not sure if it's what I think it is.

It's not the right config in your screenshot. You tried to change 'max_queues_breakdown' config to 15 and it didn't change the number of background tasks queues ? queues are initialized at startup during platform init. I just tested to change it to 10 and I do have the right number of background tasks queues.

`export const BACKGROUND_TASK_QUEUES = parseInt(conf.get('app:task_scheduler:max_queues_breakdown') ?? '4', 10);`

image

aHenryJard (approved):
All good to me :)

I just wonder about starting at #1 instead of #0 but since it's a technical screen it's all fine to me.

PR #11789, opened 2025-07-24
[backend] replace the “Too many values” log error with a warn only (#11682)

### LINKED ISSUE #11682: [backend] Decay manager log issue
## Description

We get the following log as an error, while it's should only be a Warn. 
It does not block the ingestion process

Image

Image

## Environment

1.  OpenCTI version: 6.7.3

CelineSebe:
### Proposed changes

* warn log replacing log error

### Related issues

* #11682 

### Checklist


richard-julien (approved):
Im fine with the change but need to find why CI/CD is failing :)

Issue #11634, opened 2025-07-08
Unable to select a relationship type in Observable nested objects


delemaf:
## Description
Unable to select a relationship type in Observable nested objects

## Environment

1. OS (where OpenCTI server runs): { e.g. Mac OS 10, Windows 10, Ubuntu 16.4, etc. }
2. OpenCTI version: { e.g. OpenCTI 1.0.2 }
3. OpenCTI client: { e.g. frontend or python }
4. Other environment details:

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Go to a observable knowledge
2. Add a nested object
3. Try to add a relationship type

## Expected Output

You often cannot pick a relationship type

## Actual Output

You should be able to add object only with an allowed entity type

PR #11796, opened 2025-07-24
[frontend/backend] Remove not allowed types in Nested Relation creation list (#11634)

Archidoit:
### Proposed changes
In the panel to select an entity with which to create a nested ref relationship, keep only the entities with which it is possible to create the nested ref

### Related issues
https://github.com/OpenCTI-Platform/opencti/issues/11634


opencti-platform/opencti-graphql/src/resolvers/stixRefRelationship.js
```diff
@@ -25,6 +26,7 @@ const stixRefRelationshipResolvers = {
     stixRefRelationships: (_, args, context) => findAll(context, context.user, args),
     stixNestedRefRelationships: (_, args, context) => findNested(context, context.user, args),
     stixSchemaRefRelationships: (_, { id, toType }, context) => schemaRefRelationships(context, context.user, id, toType),
+    stixSchemaRefRelationshipsPossibleTypes: (_, { type }, __) => schemaRefRelationshipsPossibleTypes(type),
```
aHenryJard: Could you please add a test on this new resolver API ?

Archidoit: done :)

delemaf (approved):
Looks good to me :+1:

PR #11799, opened 2025-07-24
[backend](fix)Vulnerabilities: support "_" & "-" in CVSS values 

### LINKED ISSUE #11649: Vulnerabilities: support "_" & "-" in CVSS values
## Description

Currently, the implementation of CVSS does not support "_" & "-" (not in the norm), resulting in throwing errors when vulnerabilties are ingested with this values. To be clear, these "_" & "-" are in the context of CVSS data: ex "ADJACENT_NETWORK"

But it seems that some users get data source where data has these values are formated like this. Therefore, we should support this. 

## Environment

6.7.3

## Reproducible Steps

1. Try to ingest a vulnerability with "ADJACENT_NETWORK" OR ADJACENT-NETWORK as attack vector

## Expected Output

Vulnerability is ingested without any issue

## Actual Output

Vulnerability is not ingested an issue araise.

maelv-filigran:
### Proposed changes

* Add support for multiple Attack Vector input formats in CVSS vulnerability mapping
* Handle "ADJACENT_NETWORK", "ADJACENT-NETWORK", and "Adjacent Network" variations using regex pattern matching
* Ensure proper mapping to CVSS version-specific values ("Adjacent Network" for CVSS2, "Adjacent" for CVSS3/4)

### Related issues

* #11649
* Fixes issue where Attack Vector values with underscores or hyphens were not properly recognized
* Improves CVSS data import compatibility with various external sources

### Checklist

### Further comments

This change improves the robustness of CVSS Attack Vector parsing by accepting multiple input formats that are commonly found in vulnerability databases and security tools. The implementation uses a regex pattern `/^adjacent[\s_-]?network$/i` to match all variations while maintaining backward compatibility.

The refactoring also addresses ESLint warnings by avoiding parameter reassignment and using proper dot notation for object property access.

### DIFF
--- a/opencti-platform/opencti-graphql/src/utils/vulnerabilities.ts
+++ b/opencti-platform/opencti-graphql/src/utils/vulnerabilities.ts
@@ -262,7 +262,7 @@ const cvss2OutputKeyCase: Record<string, string> = {
 
 // --- Helpers ---
 
-const getFullValue = (
+export const getFullValue = (
   metric: string | undefined,
   value: string | null,
   config: CvssConfig
@@ -289,7 +289,7 @@ const getFullValue = (
   return value;
 };
 
-const getCodeValue = (
+export const getCodeValue = (
   metric: string,
   value: string,
   config: CvssConfig
--- a/opencti-platform/opencti-graphql/src/utils/vulnerabilities.ts
+++ b/opencti-platform/opencti-graphql/src/utils/vulnerabilities.ts
@@ -297,22 +297,38 @@ const getCodeValue = (
   const map = config.fullToCode[metric];
   if (!map) return value;
 
+  let processedValue = value;
+
+  // Special handling for Attack Vector (AV) with variations of "Adjacent Network"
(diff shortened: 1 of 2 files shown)


opencti-platform/opencti-graphql/src/utils/vulnerabilities.ts
```diff
@@ -297,22 +297,38 @@ const getCodeValue = (
   const map = config.fullToCode[metric];
   if (!map) return value;
 
+  let processedValue = value;
```
Archidoit, line 300: can you add tests for the getCodeValue function?

Archidoit (changes requested):

SarahBocognano (approved):

opencti-platform/opencti-graphql/src/utils/vulnerabilities.ts
maelv-filigran: done :)

Archidoit (approved):

Issue #11798, opened 2025-07-24
Massive Operations: Prevent making 2 actions in the same tasks on the same key


nino-filigran:
## Description

If trying to perform the following task (see screenshot), it will fail. We should prevent users from selecting the same key once selected once.

<img width="951" height="317" alt="Image" src="https://github.com/user-attachments/assets/d112cee0-d382-4517-81f3-9fc887266bbf" />

## Environment

Testing

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. In any screen supporting massive operation, select at least 1 entity
2. click on edit
3. select the add action 
4. select label
5. add a label 
6. add another acton (add/label/value of the label)

## Expected Output

Success of the task.

In fact, the idea would be to prevent user from selecting the same "attribute" once already selected. Therefore, if for instance user has already selected ADD/LABEL/value of the label, if users adds a new task, selects ADD, "LABEL" should not be offered.

This should be the case for any operations & any attribute "ADD"/"REPLACE"/"REMOVE".

## Actual Output

Failure of the task.

PR #11806, opened 2025-07-25
[frontend/backend] Prevent making several actions on the same key in the same task (#11798)

Archidoit:
### Proposed changes
Prevent doing several actions on the same key in a background task (because it can cause task errors):
- prevent selecting several times the same field when creating a task from the UI
- raise an error at the creation of a task with several actions on the same field in the backend for tasks created from the API

### Related issues
https://github.com/OpenCTI-Platform/opencti/issues/11798

### DIFF
--- a/opencti-platform/opencti-graphql/tests/01-unit/domain/backgroundTask-common-test.ts
+++ b/opencti-platform/opencti-graphql/tests/01-unit/domain/backgroundTask-common-test.ts
@@ -61,8 +61,8 @@ describe('Background task validity check (checkActionValidity)', () => {
     const type = TASK_TYPE_QUERY;
     const input = {
       actions: [
-        { type: ACTION_TYPE_ADD, field: 'label', value: ['label1'] },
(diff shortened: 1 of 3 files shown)


opencti-platform/opencti-graphql/tests/01-unit/domain/backgroundTask-common-test.ts
```diff
@@ -55,6 +56,21 @@ describe('Background task validity check (checkActionValidity)', () => {
     { name: 'INVESTIGATION_INUPDATE_INDELETE' },
   ]);
 
+  it('should throw an error if task QUERY and filter is not PublicDashboard', async () => {
```
lndrtrbn: Is the title of the test a copy paste? I do not understand the link between the title and what the test is doing

Archidoit: indeed :) I renamed it :) done

lndrtrbn (approved):

Issue #11780, opened 2025-07-23
[Dashboard] Activity & History Donut Widget Also Displays Sectors When Relationship Exists With an Organization


ericWadeFord:
## Description
When creating a widget to display a donut visualization displaying login events for organizations linked to users assigned to a specific group, if these organizations have a relationship with a sector, that sector is also shown in the widget.

## Environment
6.7.7

## Reproducible Steps
Steps to create the smallest reproducible scenario:
1. Create a widget using the Donut visualization, with an Activity & History perspective.
2. Set the filters to a specific group and the login Event scope, and validate.
3. Under parameters, set the Attribute to Organization.
4. Observe in the output that Sector(s) and Organizations are shown.

## Expected Output
Only Organizations are shown in the output

## Actual Output
If an organization has a relationship with a sector, that sector is also displayed in the widget's output.

## Screenshots

<img width="682" height="476" alt="Image" src="https://github.com/user-attachments/assets/f2f1fd68-35fb-4ab8-82b5-f0941b19fe18" />

<img width="981" height="301" alt="Image" src="https://github.com/user-attachments/assets/0a06f138-0d6b-452b-8c56-0ba60690f75d" />

aHenryJard:
Here is the root cause, but this fix may not be enough to get a correct Donut data.
https://github.com/OpenCTI-Platform/opencti/issues/11824

PR #11823, opened 2025-07-29
[backend] Restrict Rule ParticipateToPartsRule to Organization and restrict build of user organizationIds to org (#11824)

### LINKED ISSUE #11824: The Rule ParticipateToPartsRule creates participate to relations with Sector.
## Description

The Rule ParticipateToPartsRule should creates 'participate' relation when Organization are part of other Organization.
But the rule is taking also Sector when Organisation are port of sector.

The issue is that Users ends with a list or organizationsId that contains both Organizations and Sector, and this list of ids ends in Activity logs.

Also users should not list other entity than Organization in organization list.

Note: This is the root cause of https://github.com/OpenCTI-Platform/opencti/issues/11780 but it will not fix on existing platform.

## Environment

4. Other environment details:

## Reproducible Steps

1. { e.g. Run ... }
2. { e.g. Click ... }
3. { e.g. Error ... }

aHenryJard:
### Proposed changes

* Fix rule engine rule execution for ParticipateToPartsRule by adding a verification on the toType
* Add test coverage on the bug use case

### Related issues

* part of #11780 
* closes #11824

### Checklist


opencti-platform/opencti-graphql/src/rules/relationToRelationBuilder.ts
```diff
@@ -79,7 +97,10 @@ const buildRelationToRelationRule = (ruleDefinition: RuleDefinition, relationTyp
             stop_time: range.end,
             objectMarking: elementMarkings,
           });
-          await createInferredRelation(context, input, ruleContent);
+
+          if (await isTargetTypeValid(context, RULE_MANAGER_USER, ruleDefinition, toId)) {
+            await createInferredRelation(context, input, ruleContent);
+          }
```
JeremyCloarec, line 103: Instead of checking the type inside the callback of listAllRelations, I think the fix should be to apply more filters in listAllRelations.
We should add fromTypes and toTypes from the filters in the scopes of the rule to the listToArgs and listFromArgs

aHenryJard:
Way too old PR

Issue #11295, opened 2025-06-12
Ability to enroll my openCTI platform into XTM Hub


EllynBsc:
## Use case
Ability to enroll my openCTI platform into XTM Hub

### Detailed Description
The issue pertains to the lack of a feature that allows the integration of an openCTI platform with the XTM Hub. This is a problem as it prevents seamless data exchange and workflow synchronization between these two platforms which can greatly enhance productivity and efficiency.

SamuelHassine:
@aHenryJard Why this is not closed?

PR #11846, opened 2025-07-30
[backend/frontend] feat(hub): add registration connectivity check

Kroustille:
### Proposed changes

* add hub registration manager
* check for XTM hub registration connectivity
* update registration status if there is an issue
* display this lost connectivity to user on the XTM Hub settings page

### Related issues

* #11295 

### Checklist

### Further comments

I will update enroll to register every in a future PR to be able to merge this one fast.


opencti-platform/opencti-graphql/config/default.json
```diff
@@ -154,6 +154,10 @@
       "to_email": "no-reply@filigran.io",
       "max_list_size": 500
     },
+    "enrollment": {
+      "to_email": "no-reply@filigran.io",
```
aHenryJard: I think it's better to group in "xtm" block. Maybe xtmhub_to_email : if we send other email for xtm purpose we could re-use the same config.

About max_list_size I wonder if we shouldn't have one for the whole platform.... like "smtp.email_max_cc_size"

opencti-platform/opencti-graphql/src/manager/hubEnrollmentManager.ts
```diff
@@ -0,0 +1,123 @@
...
+const HUB_ENROLLMENT_MANAGER_ENABLED = booleanConf('hub_enrollment_manager:enabled', true);
+const HUB_ENROLLMENT_MANAGER_KEY = conf.get('hub_enrollment_manager:lock_key') || 'hub_enrollment_manager_lock';
+const SCHEDULE_TIME = conf.get('hub_enrollment_manager:interval') || 60 * 60 * 1000; // 1 hour
+const MAX_ENROLLMENT_LIST_SIZE = conf.get('app:enrollment:max_list_size') || 500;
+const TO_EMAIL = conf.get('app:enrollment:to_email') || 'no-reply@filigran.io';
+
+const EMAIL_BODY = `
+  <h2>Hi,</h2>
+  
+  <p>We wanted to inform you that the connectivity between OCTI and the XTM Hub has been lost. As a result, the integration is currently inactive.</p>
```
aHenryJard: no OCTI please XD

opencti-platform/opencti-graphql/src/manager/hubEnrollmentManager.ts
```diff
@@ -0,0 +1,123 @@
...
+  const settings = await getEntityFromCache<BasicStoreSettings>(context, SYSTEM_USER, ENTITY_TYPE_SETTINGS);
+  if (!settings.xtm_hub_token) {
+    return;
+  }
+
+  const status = await hubClient.loadEnrollmentStatus({ platformId: settings.id, token: settings.xtm_hub_token });
+  if (status === 'active') {
+    await settingsEditField(
+      context,
+      SYSTEM_USER,
```
aHenryJard: we try to avoid using SYSTEM_USER as much as possible, please use a dedicated manager user, such as https://github.com/OpenCTI-Platform/opencti/blob/8d3382900a7b29db7245badd50889ed08ba58e38/opencti-platform/opencti-graphql/src/utils/access.ts#L183

aHenryJard: and you need to register the user here also
https://github.com/OpenCTI-Platform/opencti/blob/8d3382900a7b29db7245badd50889ed08ba58e38/opencti-platform/opencti-graphql/src/utils/access.ts#L469

opencti-platform/opencti-graphql/src/utils/hub-client.ts
```diff
@@ -0,0 +1,37 @@
+import { getHttpClient } from './http-client';
+import conf, { logApp } from '../config/conf';
+
+type EnrollmentStatus = 'active' | 'inactive';
+
+const HUB_BACKEND_URL = conf.get('xtm:xtmhub_backend_url');
+
+export const hubClient = {
```
aHenryJard: I see that we don't have an "xtm hub" module yet, but we have an "xtm" one. Could you move this file to src/modules/xtm please ?

opencti-platform/opencti-graphql/src/modules/xtm/hub/xtm-hub-client.ts
```diff
@@ -0,0 +1,37 @@
...
+    const httpClient = getHttpClient({
+      baseURL: HUB_BACKEND_URL,
+      responseType: 'json'
+    });
+
+    try {
+      const response = await httpClient.post('/graphql-api', { query, variables });
+      return response.data.data.octiPlatformEnrollmentStatus.status;
+    } catch (error) {
+      logApp.warn(error);
```
aHenryJard:
```suggestion
      logApp.warn('XTM Hub is not reachable', { reason: error });
```
The warn log is way to much verbose, you need to add a message (I proposed one but feel free to change)and put the error object in metadata.

opencti-platform/opencti-graphql/src/manager/hubRegistrationManager.ts
```diff
@@ -0,0 +1,123 @@
...
+export const hubRegistrationManager = async () => {
+  const context = executionContext('hub_registration_manager');
+  const settings = await getEntityFromCache<BasicStoreSettings>(context, HUB_REGISTRATION_MANAGER_USER, ENTITY_TYPE_SETTINGS);
+  if (!settings.xtm_hub_token) {
+    return;
+  }
+
+  const status = await xtmHubClient.loadRegistrationStatus({ platformId: settings.id, token: settings.xtm_hub_token });
+  if (status === 'active') {
+    await settingsEditField(
```
aHenryJard: Would be more performant to store in elastic only when connectivity changes. Can you add a check previous <-> new status before calling settingsEditField please ?

aHenryJard: ah I just think on date 'xtm_hub_last_connectivity_check' ... so instead we need to use another way to update setting without having an activity log all the time.

opencti-platform/opencti-graphql/config/default.json
```diff
@@ -232,6 +232,10 @@
     "openrm_token": "",
     "openmtd_url": "",
     "openmtd_token": "",
+    "hub": {
```
aHenryJard, line 235: There is a consistency issue, some part of xtm hub config is under "hub" and some part is outside "xtmhub_url".
Let me think about it.

opencti-platform/opencti-graphql/src/manager/hubRegistrationManager.ts
```diff
@@ -0,0 +1,123 @@
...
+    <a href="${getBaseUrl()}">Access OpenCTI</a><br />
+    Best,<br />
+    Filigran Team<br />
+  </p>
+`;
+
+const sendAdministratorsLostConnectivityEmail = async (context: AuthContext, settings: BasicStoreSettings) => {
+  const administrators = (await findUserWithCapabilities(context, HUB_REGISTRATION_MANAGER_USER, [BYPASS, SETTINGS_SETMANAGEXTMHUB])) as AuthUser[];
+  if (administrators.length > MAX_EMAIL_LIST_SIZE) {
+    logApp.error(`You cannot have more than ${MAX_EMAIL_LIST_SIZE} e-mail addresses`);
```
aHenryJard: I wonder about sending to the 500 first and log a warn message here. Instead of sending nothing.

delemaf (approved):
Looks good to me :+1:

Issue #11748, opened 2025-07-21
[ChatBot] Make AI features disableable


alice-debra:
Add in the platform configuration specific fields to disable or enable the AI features.

Kedae:
We need to add this in the conf of the platform, not in UI

labo-flg:
We actually made it possible to configure AI feature through the UI, in the _Filigran Experience_ view.

PR #11848, opened 2025-07-30
[frontend] Make AI features disableable with configuration (#11748)

CelineSebe:
### Proposed changes

* add enable and configured condition to display AI CTA

### Related issues

* #11748 

### Checklist


opencti-platform/opencti-front/lang/front/fr.json
```diff
@@ -3605,7 +3605,7 @@
   "To prevent people seeing all the data...": "Pour empêcher les gens de voir toutes les données du tableau de bord public, sélectionnez une définition de marquage pour restreindre les données partagées.",
   "to terminate your session.": "pour mettre fin à votre session.",
   "To type": "Type cible",
-  "To use AI, please enable it in the configuration of your platform.": "Pour utiliser l'IA, veuillez l'activer dans la configuration de votre plateforme.",
+  "To use this AI Enterprise edition feature, please add a token": "Pour utiliser cette fonctionnalité de l'édition AI Enterprise, veuillez ajouter un jeton",
```
lndrtrbn: This translation seems not very good

opencti-platform/opencti-front/src/private/components/common/ai/AIInsights.tsx
lndrtrbn, line 1: You can make an early returns `if (!isAIConfigured) return null` instead of adding the same condition in two different ifs. You won't need the second if at all after.

opencti-platform/opencti-front/src/private/components/common/form/TextFieldAskAI.tsx
```diff
@@ -241,103 +242,105 @@ const TextFieldAskAI: FunctionComponent<TextFieldAskAiProps> = ({
...
   const renderButton = () => {
-    return (
-      <>
-        <EETooltip forAi={true} title={t_i18n('Ask AI')}>
-          <IconButton
-            size="small"
-            onClick={(event) => ((isEnterpriseEdition && enabled && configured) ? handleOpenMenu(event) : null)}
-            disabled={disabled || currentValue.length < 10}
-            style={{ color: theme.palette.ai.main }}
+    if (isAIConfigured || !isEnterpriseEdition) {
```
lndrtrbn: In my opinion, a function called `renderButton` should return the button JSX anyway. If the button should not be displayed then the condition should be set where we are calling the function, not inside the function.

opencti-platform/opencti-front/src/private/components/common/form/TextFieldAskAI.tsx
```diff
@@ -241,6 +240,7 @@ const TextFieldAskAI: FunctionComponent<TextFieldAskAiProps> = ({
   };
 
   const renderButton = () => {
+    if (!isAIConfigured) return null;
```
lndrtrbn: Prefer not put any condition in the function, instead do not call the function at all if not needed.

I suppose you have something like this in the JSX : `{renderButton()}`
Replace it with: `{isAIConfigured && renderButton()}`

That way the function render button always have the same behavior, we just do not call it when we do not need it.

opencti-platform/opencti-front/src/private/components/common/ai/AIInsights.tsx
```diff
@@ -174,48 +167,9 @@ const AIInsights = ({
...
-            handleCloseDrawer={() => setDisplayEEDialog(false)}
-            initialValue={{
-              description: t_i18n('I would like to use a EE feature AI Summary but I don\'t have EE activated.\nI would like to discuss with you about activating EE.'),
-            }}
-          />
-        )}
-      </>
-    );
-  }
+  const isAIConfigured = enabled && configured;
```
lndrtrbn: You do not need to create a variable `isAIConfigured`. In hook `useAI` there is an attribute `fullyActive` that already does what you need.

You can replace in all modified files.

aHenryJard (approved):
Tested locally with CE and without CE, all good to me.

Issue #4936, opened 2023-11-20
Very low consumer ACK rate


MaxwellDPS:
## Description

Having a very hard time loading data to the platform, uploading any data takes a long time to process. 50K items in the queue can take > 3 hours, this leads to seeing an ack rate sustained of ~.7/s. There is not a manageable way to import this data before it begins to backlog

## Environment

1. OS (where OpenCTI server runs): CentOS stream 9
2. OpenCTI version: 5.11.14
3. OpenCTI client: Frontend
5. Other environment details:
>        15x Platform (No Web) @ 1c 6Gb (Reserved)
>        25x Worker            @ 1c 4Gb (Reserved)
>        k8s Clustered deployment
>
>        Elastic Cluster: 3x Data Node @ 16Gb 4c per 
>        RabbitMQ Cluster: 3x Node @ 4Gb 1c per 
>        Redis: 1x 16Gb 3
>               
>        Host Networking: 25Gbps x 3 Hosts

None of these seem resource saturated, I am as scaled as I can and the platform is nearly not able to import data

## Reproducible Steps

Steps to create the smallest reproducible scenario:
N/A 

## Expected Output

ACK rate that imports data fast enough to not buildup a backlog

## Additional information

# Resource usage
```
NAME                                                    CPU(cores)   MEMORY(bytes)   
elastic-es-data-0                                       1923m        14645Mi         
elastic-es-data-1                                       884m         11464Mi         
elastic-es-data-2                                       1443m        14217Mi         
elastic-es-leaders-0                                    131m         1931Mi          
elastic-es-leaders-1                                    74m          1833Mi          
elastic-es-leaders-2                                    46m          1927Mi                
minio-66787954dd-9g7sn                                  1m           180Mi           
opencti-api-59cc685c95-84qnp                            1398m        1401Mi          
opencti-api-59cc685c95-9j4vr                            502m         550Mi           
opencti-api-59cc685c95-bs6sk                            2131m        5267Mi          
opencti-api-59cc685c95-dhczm                            2272m        3755Mi          
opencti-api-59cc685c95-g6hnt                            953m         3376Mi          
opencti-api-59cc685c95-qzwjq                            389m         535Mi           
opencti-api-59cc685c95-z9t5g                            398m         543Mi           
opencti-api-59cc685c95-zkvfs                            1829m        1929Mi                   
opencti-web-7c45ccb866-66wnj                            810m         760Mi           
[...]

PR #11849, opened 2025-07-30
[worker] Worker performance increase (#11889)

### LINKED ISSUE #11889: [worker] Improve worker performance
## Description

Workers process only handle 1 message at a time per active queue, which defeats part of the intended parallelism implemented in the worker.

## Environment

1. OS (where OpenCTI server runs): N/A
2. OpenCTI version: OpenCTI 6.7.9
3. OpenCTI client: python
4. Other environment details:

## Reproducible Steps

1. { e.g. Run ... }
2. { e.g. Click ... }
3. { e.g. Error ... }

gileri:
### Proposed changes

* Align AMQP prefetch with requested worker thread pool size
* Removed busy polling for thread termination. Instead, the above [AMQP prefetch limits](https://www.rabbitmq.com/docs/consumer-prefetch) will limit the throughput.

Those two changes does allow having better utilization of resources, limiting the need of running multiple worker processes.

Remaining ideas to implement in another PR:

* [Raise pycti Requests module maximum connections, to allow more than 10 connections in parallel](https://docs.python-requests.org/en/latest/api/#requests.adapters.HTTPAdapter)
* Finer prefetch adjustment for additional performance

### Related issues

* #4936
* https://github.com/OpenCTI-Platform/opencti/issues/11889

### Checklist

### Further comments

This PR may increase the load on the OpenCTI API and backends, up to 5 times on single RabbitMQ active queue workloads (Only one queue has messages), down to ~5-10% on systems with more than 5 active queues

Fixes #11889

### DIFF
--- a/opencti-worker/src/message_queue_consumer.py
+++ b/opencti-worker/src/message_queue_consumer.py
@@ -15,12 +15,13 @@ class MessageQueueConsumer:  # pylint: disable=too-many-instance-attributes
     pika_parameters: pika.ConnectionParameters
     execution_pool: ThreadPoolExecutor
     handle_message: Callable[[str], Literal["ack", "nack", "requeue"]]
+    queue_concurrency_enabled: bool
     should_stop: bool = field(default=False, init=False)
 
     def __post_init__(self) -> None:
         self.pika_connection = pika.BlockingConnection(self.pika_parameters)
         self.channel = self.pika_connection.channel()
-        self.channel.basic_qos(prefetch_count=1)
+        self.channel.basic_qos(prefetch_count=(self.execution_pool._max_workers + 1) if self.queue_concurrency_enabled else 1)
         self.thread = Thread(target=self.consume_queue, name=self.queue_name)
         self.thread.start()
 
@@ -83,9 +84,10 @@ class MessageQueueConsumer:  # pylint: disable=too-many-instance-attributes
                     method.delivery_tag,
                     body,
                 )
-                while task_future.running():  # Loop while the thread is processing
-                    self.pika_connection.sleep(0.05)
-                self.logger.info("Message processed, thread terminated")
+                if not self.queue_concurrency_enabled:
+                    while task_future.running():  # Loop while the thread is processing
+                        self.pika_connection.sleep(0.05)
+                    self.logger.info("Message processed, thread terminated")
         except Exception as e:
             self.logger.error("Unhandled exception", {"exception": e})
         finally:
--- a/opencti-worker/src/worker.py
+++ b/opencti-worker/src/worker.py
@@ -146,6 +146,14 @@ class Worker:  # pylint: disable=too-few-public-methods, too-many-instance-attri
             True,
             0,
         )
+        self.worker_queue_concurrency_enabled = get_config_variable(
+            "WORKER_QUEUE_CONCURRENCY_ENABLED",
+            ["worker", "queue_concurrency_enabled"],
+            config,
+            False,
+            False,
+        )
+
         # Telemetry
         if self.telemetry_enabled:
             self.prom_httpd, self.prom_t = start_http_server(
@@ -257,6 +265,7 @@ class Worker:  # pylint: disable=too-few-public-methods, too-many-instance-attri
                             pika_parameters,
                             push_execution_pool,
                             push_handler.handle_message,
+                            self.worker_queue_concurrency_enabled,
                         )
 
                     # Listen for webhook message
@@ -281,6 +290,7 @@ class Worker:  # pylint: disable=too-few-public-methods, too-many-instance-attri
                                 self.build_pika_parameters(connector_config),
                                 listen_execution_pool,
                                 listen_handler.handle_message,
+                                self.worker_queue_concurrency_enabled,
                             )
 
                 # Check if some consumer must be stopped
--- a/opencti-worker/src/push_handler.py
+++ b/opencti-worker/src/push_handler.py
@@ -1,6 +1,7 @@
 import base64
 import datetime
 import json
+import threading
 from dataclasses import dataclass
 from typing import Any, Dict, Union, Literal
 
@@ -29,13 +30,19 @@ class PushHandler:  # pylint: disable=too-many-instance-attributes
     objects_max_refs: int
 
     def __post_init__(self) -> None:
-        self.api = OpenCTIApiClient(
-            url=self.opencti_url,
-            token=self.opencti_token,
-            log_level=self.log_level,
-            json_logging=self.json_logging,
-            ssl_verify=self.ssl_verify,
+        self.local_api = threading.local()
+
+    # OpenCTIClient is not thread safe, use a thread local to ensure to work on a dedicated client when creating and sending a request
+    def get_api_client(self) -> OpenCTIApiClient:
+        if not hasattr(self.local_api, "client"):
+            self.local_api.client = OpenCTIApiClient(
+                url=self.opencti_url,
+                token=self.opencti_token,
+                log_level=self.log_level,
+                json_logging=self.json_logging,
+                ssl_verify=self.ssl_verify,
         )
+        return self.local_api.client
 
     def send_bundle_to_specific_queue(
         self,
@@ -76,17 +83,18 @@ class PushHandler:  # pylint: disable=too-many-instance-attributes
         imported_items = []
         start_processing = datetime.datetime.now()
         try:
+            api = self.get_api_client()
             # Set the API headers
-            self.api.set_applicant_id_header(data.get("applicant_id"))
-            self.api.set_playbook_id_header(data.get("playbook_id"))
-            self.api.set_event_id(data.get("event_id"))
-            self.api.set_draft_id(data.get("draft_id"))
-            self.api.set_synchronized_upsert_header(data.get("synchronized", False))
-            self.api.set_previous_standard_header(data.get("previous_standard"))
+            api.set_applicant_id_header(data.get("applicant_id"))
+            api.set_playbook_id_header(data.get("playbook_id"))
+            api.set_event_id(data.get("event_id"))
+            api.set_draft_id(data.get("draft_id"))
+            api.set_synchronized_upsert_header(data.get("synchronized", False))
+            api.set_previous_standard_header(data.get("previous_standard"))
             work_id = data.get("work_id")
             # Check if work is still valid
             if work_id is not None:
-                is_work_alive = self.api.work.get_is_work_alive(work_id)
+                is_work_alive = api.work.get_is_work_alive(work_id)
                 # If work no longer exists, bundle can be acked without doing anything
                 if not is_work_alive:
                     return "ack"
@@ -107,7 +115,7 @@ class PushHandler:  # pylint: disable=too-many-instance-attributes
                 if len(content["objects"]) == 1 or data.get("no_split", False):
                     update = data.get("update", False)
                     imported_items, too_large_items_bundles = (
-                        self.api.stix2.import_bundle_from_json(
+                        api.stix2.import_bundle_from_json(
                             raw_content, update, types, work_id, self.objects_max_refs
                         )
                     )
@@ -159,7 +167,7 @@ class PushHandler:  # pylint: disable=too-many-instance-attributes
                             )
                             # Add expectations to the work
                             if work_id is not None:
-                                self.api.work.add_expectations(work_id, expectations)
+                                api.work.add_expectations(work_id, expectations)
                             # For each split bundle, send it to the same queue
                             for bundle in bundles:
                                 self.send_bundle_to_specific_queue(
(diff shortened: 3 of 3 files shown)


JustWats:
I've worked with this a bit myself on a custom build, played with tuning on the worker.py env settings since I have AMQP prefetch hooked up on an env-variable.

A lot of the times I was running into database_errors, lock_errors, etc.

Specs were typically a 20Gi maxmem on 5 node redis-cluster and 24Gi JVM Heap on a 3 data-node elastic-cluster.

Was able to stabilize on 1 worker.py process at about 40 bundles/per sec instead of the typical 0.5-0.9 bps. Although anything much higher than that led to the connection size issue which throttled it.

gileri:
I have tested with multiple active queue, and a worker size of 20, it works fine.

Here's the rate with, then without the PR changes (on 10 workers, pool size of 30):

Screenshot From 2025-07-31 17-34-00
Screenshot From 2025-07-31 17-32-32

Screenshot From 2025-07-31 17-32-29

victorhery-filigran:
Good job thanks!

Just be careful, as discussed with @aHenryJard , something that should be tested are bundle that create messages by themselves
Sometimes, when bundle is consumed, it creates new messages in the queue because it has some entanglement. And these messages should be consumed in the good order after that

Angélique, if you know a connector doing that we could specifically use for test, it would be cool 😊 
Thanks!

opencti-worker/src/worker.py
```diff
@@ -643,7 +645,8 @@ def __post_init__(self) -> None:
         # Initialize variables
         self.connectors: List[Any] = []
         self.queues: List[Any] = []
-        self.execution_pool = ThreadPoolExecutor(max_workers=self.opencti_pool_size)
+        global execution_pool
+        execution_pool = ThreadPoolExecutor(max_workers=self.opencti_pool_size)
         self.listen_api_execution_pool = ThreadPoolExecutor(
```
xfournet: self.listen_api_execution_pool is no more really used since self.execution_pool is no more used in ApiConsumer.
The global variable may add confusion, it would be clearer to remove self.list_api_execution_pool and provide self.execution_pool instead to ApiConsumer constructors

JustWats:
image

Just an update, @gileri been running stable about a few days with the changes. Testing a bunch of connectors at the moment to test thoroughput.

image

Running on 30-30-3 settings at the moment, and it feels pretty good.

image


Running on 10 node cluster for OCTI. 2 frontend, 2 ingestion, 5 backend, 1 api. 3 Workers per backend.
Running on elasticsearch cluster 3-3 master-data.  13Gi JVM heaps, redis-cluster with 25Gi RAM.

opencti-worker/src/worker.py
gileri: Indeed, that was a pretty bad implementation here. Fixed, and we still use two separate execution queues to avoid head-of-line blocking if a callback URL is slow.

gileri:
Thanks for the reviews and suggestions @xfournet!

I've cleaned up the thread pools handling, and keep the two separate. Also reverted the pool size reduction: Performance would increase up to 5 times on single-queue workloads, but on systems with 5 or more active queues it would be only about 5-10% (AMQP prefetch added, 25ms wait average wait between message per queue removed).

Amending the PR description to avoid confusion.

xfournet:
> Just be careful, as discussed with @aHenryJard , something that should be tested are bundle that create messages by themselves Sometimes, when bundle is consumed, it creates new messages in the queue because it has some entanglement. And these messages should be consumed in the good order after that

@victorhery-filigran @aHenryJard  is it a strong requirement for theses messages to be consumed in order ? currently, in multiple worker mode, messages may be processed out of order

opencti-worker/src/worker.py
```diff
@@ -493,16 +490,13 @@ def run(self) -> None:
...
+                    self.execution_pool.submit(
                         self.data_handler,
                         self.pika_connection,
                         self.channel,
                         method.delivery_tag,
                         data,
                     )
-                    while task_future.running():  # Loop while the thread is processing
-                        self.pika_connection.sleep(0.05)
-                    self.worker_logger.info("Message processed, thread terminated")
```
xfournet: I identified a potential concurrency issue that we had not initially considered. Indeed, this change allows multiple messages per queue to be processed concurrently. In the Consumer class, the OpenCTI API client (self.api) is modified at the beginning of data_handler (lines 320 to 329) and used later (lines 346, 372, 403, 424, and 447), which leaves room for a race condition between concurrent threads.
Note: In ApiConsumer, this appears to be safe.

xfournet (changes requested):

opencti-worker/src/worker.py
gileri: Good catch ! This seems a blocker to me, even if perhaps the GIL saves us.

aHenryJard:
> > Just be careful, as discussed with @aHenryJard , something that should be tested are bundle that create messages by themselves Sometimes, when bundle is consumed, it creates new messages in the queue because it has some entanglement. And these messages should be consumed in the good order after that
> 
> @victorhery-filigran @aHenryJard is it a strong requirement for theses messages to be consumed in order ? currently, in multiple worker mode, messages may be processed out of order

I'm not sure at all that the order matter, my wondering is about the "split bundle" system, I don't remember out of my head if a bundle in split with an order that should not be changed or if each small bundle are independent. So the part to check for concurrency is in this part https://github.com/OpenCTI-Platform/client-python/blob/a784e89fb0aba21a6cbba009ff457b0af65c29c4/pycti/utils/opencti_stix2_splitter.py#L204

That part will need to be tested. If all splitted bundle are inside the same queue will this code change have impact on queue order ?

aHenryJard:
Could you target release/current branch for performance PRs instead of master please ? This will let more time to test on internal validation platform before releasing.

opencti-worker/src/worker.py
gileri: Fixed in https://github.com/OpenCTI-Platform/opencti/pull/11849/commits/770a291aea4394cba39040362439df4db8ffa1ad

gileri:
> bundle in split with an order that should not be changed or if each small bundle are independent. So the part to check for concurrency is in this part https://github.com/OpenCTI-Platform/client-python/blob/a784e89fb0aba21a6cbba009ff457b0af65c29c4/pycti/utils/opencti_stix2_splitter.py#L204

[RabbitMQ Queues are FIFO, and this bundle-split system requeues sub-bundles in independent AMQP messages](https://github.com/OpenCTI-Platform/opencti/blob/f2053d0b0e38ba79b31089e3656d52b2d6fad20b/opencti-worker/src/worker.py#L369-L390), so order of processing for those is already not guaranteed nor respected. And the split operation itself seem pretty safe, now that HTTP headers are correctly sent.

>Could you target release/current branch for performance PRs instead of master please ? This will let more time to test on internal validation platform before releasing.

Done, thanks.

opencti-worker/src/worker.py
```diff
@@ -316,17 +316,28 @@ def data_handler(  # pylint: disable=too-many-statements, too-many-locals
...
         try:
+            api = OpenCTIApiClient(
+                url=self.opencti_url,
+                token=self.opencti_token,
+                log_level=self.log_level,
+                ssl_verify=self.ssl_verify,
+                json_logging=self.json_logging,
+                perform_health_check=False,
+            )
+            api.logger_class("worker")
```
xfournet: this line can be removed since the result is not assigned

opencti-worker/src/worker.py
```diff
@@ -316,17 +316,28 @@ def data_handler(  # pylint: disable=too-many-statements, too-many-locals
...
         start_processing = datetime.datetime.now()
         try:
+            api = OpenCTIApiClient(
+                url=self.opencti_url,
+                token=self.opencti_token,
+                log_level=self.log_level,
+                ssl_verify=self.ssl_verify,
+                json_logging=self.json_logging,
+                perform_health_check=False,
+            )
```
xfournet: depending on the perf impact of creating a client for each call, it may be needed to cache it with a thread local

xfournet (changes requested):

opencti-worker/src/worker.py
gileri: I didn't check the implications of not calling that, but [it seems that it can be safely removed](https://github.com/OpenCTI-Platform/client-python/blob/master/pycti/utils/opencti_logger.py).

opencti-worker/src/worker.py
gileri: [I've benchmarked, it takes about 30ms CPU time per 1 000 messages](https://github.com/OpenCTI-Platform/opencti/pull/11849/commits/a3970b0c06aac5f7ac26a821001e89158de5148a). The tradeoff seems acceptable for the simplicity, but a API client pool could be implemented, that would be better.

We would have needed a pool if we couldn't disable `perform_health_check`, but we can. An healthcheck is already done at worker start up, like before. And if platform becomes unavailable later, nothing changes how it's handled.

gileri:
I've done 2 full import of the MITRE dataset with original worker code, and two with the code in the PR.
Both result in the same number of OCTI entities created. Running multiple instances of the worker did not negatively impact the test either.

gileri:
[This test](https://github.com/OpenCTI-Platform/opencti/blob/b8595a371aad1879ff0c2af6446fce97a0843725/opencti-platform/opencti-graphql/tests/03-streams/00-Raw/raw-test.js#L63-L66) fails with the changes, due to 4 label edits present in stream instead of the expect 1 defined [here](https://github.com/OpenCTI-Platform/opencti/blob/b8595a371aad1879ff0c2af6446fce97a0843725/opencti-platform/opencti-graphql/tests/utils/syncCountHelper.ts#L56-L88).

However I see 4 labelEdit mutations [here](https://github.com/OpenCTI-Platform/opencti/blob/b8595a371aad1879ff0c2af6446fce97a0843725/opencti-platform/opencti-graphql/tests/02-integration/02-resolvers/label-test.js). What is the expected number here ? Should the syncCountHelper.ts file be updated ?

gileri:
Implemented a feature flag for busy-polling as asked by @xfournet.

I've benchmarked clean OpenCTI instances with 1 MITRE, 1CVE connector, 2 workers processes. It should be representative of a regular running instance. Here are the results: 

- Stable worker container: 22 minutes to import full MITRE (with CVE running too)
- PR connector image (w/ fixes), busy wait enabled (like in stable worker): 21 minutes for MITRE (5% improvement)
- PR connector image (w/ fixes), busy wait disabled: 20 minutes for MITRE (10%)
- PR connector image (w/ fixes), busy wait disabled, 15 thread pool: 14 minutes for MITRE (36%)

By default, additional performance/overhead of this PR is expected to be 10%, or 5% if we keep the busy-wait enabled.

gileri:
I've updated the value, tests pass. Ready for review.

xfournet (approved):

gileri:
Thanks @xfournet for the review!

opencti-worker/src/message_queue_consumer.py
```diff
@@ -83,9 +84,10 @@ def consume_queue(self) -> None:
                     method.delivery_tag,
                     body,
                 )
-                while task_future.running():  # Loop while the thread is processing
-                    self.pika_connection.sleep(0.05)
-                self.logger.info("Message processed, thread terminated")
+                if not self.queue_concurrency_enabled:
+                    while task_future.running():  # Loop while the thread is processing
+                        self.pika_connection.sleep(0.05)
+                    self.logger.info("Message processed, thread terminated")
```
JeremyCloarec, line 90: Do we need to keep this logic, when self.queue_concurrency_enabled = false? Shouldn't it already be handled when setting prefetch_count to 1?

opencti-worker/src/message_queue_consumer.py
```diff
@@ -15,12 +15,13 @@ class MessageQueueConsumer:  # pylint: disable=too-many-instance-attributes
...
     execution_pool: ThreadPoolExecutor
     handle_message: Callable[[str], Literal["ack", "nack", "requeue"]]
+    queue_concurrency_enabled: bool
     should_stop: bool = field(default=False, init=False)
 
     def __post_init__(self) -> None:
         self.pika_connection = pika.BlockingConnection(self.pika_parameters)
         self.channel = self.pika_connection.channel()
-        self.channel.basic_qos(prefetch_count=1)
+        self.channel.basic_qos(prefetch_count=(self.execution_pool._max_workers + 1) if self.queue_concurrency_enabled else 1)
```
JeremyCloarec, line 24: I don't get the `+1` in `self.execution_pool._max_workers + 1`, shouldn't it be just `self.execution_pool._max_workers` ?

xfournet: the +1 is needed to be ensure to have always a message ready to be consumed by the next worker thread.
Currently there is a roundtrip with RabbitMQ for each consumed message.

opencti-worker/src/message_queue_consumer.py
xfournet: you're probably right, but i wanted to minimize the impact of the changes when the settings is not enabled. Especially in regards of flaky test, i notice that every changes around that introduce a different behavior on sync test events count.
i propose to reassess the need of this active wait later

JeremyCloarec (approved):

opencti-worker/src/message_queue_consumer.py
gileri: Indeed, it's a naive "buffer" to avoid starving the execution threads. It may be tuned further, but should be acceptable as-is.

Issue #11027, opened 2025-05-16
[ChatBot] First Implementation of a ChatBot/AI Assistant


romain-filigran:
## Use case

First implementation of a ChatBot/AI Assistant

PR #11878, opened 2025-08-01
[frontend/backend] First version of chatbot (#11027)

Kedae:
## Description
- Uses filigran-chatbot lib with a git ref
- Enable chatbot or not depending on parameters and settings:
  - if CE => chatbot disable but visible with a EE chips
  - if EE:
    - when url is configured, and chatbot is enabled => show chatbot
    - when url missing or chatbot disabled => hide it

## Related issue
- relates to #11027


aHenryJard (approved):
Let's send a first version

Issue #11024, opened 2025-05-16
Marking color missing in Task


marieflorescontact:
## Environment

1. OS (where OpenCTI server runs): { e.g. Mac OS 10, Windows 10, Ubuntu 16.4, etc. }
2. OpenCTI version: { e.g. OpenCTI 1.0.2 }
3. OpenCTI client: { e.g. frontend or python }
4. Other environment details:

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Create an incident respons with a marking definition 
2. Apply a Case Template
3. Click on one of the Task and look at marking chip in the right panel

## Expected Output

marking chip should have color

## Actual Output

![Image](https://github.com/user-attachments/assets/8ec01cab-8540-4e39-a826-ffb8160e8cbd)

PR #11907, opened 2025-08-05
[frontend] Marking color missing in Task (#11024)

SarahBocognano:
### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/11024

### DIFF
--- a/opencti-platform/opencti-front/src/components/ItemMarkings.tsx
+++ b/opencti-platform/opencti-front/src/components/ItemMarkings.tsx
@@ -68,13 +68,17 @@ const ChipMarking = ({
         case 'DR':
         case 'DR-SF':
         case 'TLP:RED':
+        case 'PAP:RED':
           color = '#c62828';
           break;
         case 'TLP:AMBER':
+        case 'TLP:AMBER+STRICT':
+        case 'PAP:AMBER':
           color = '#d84315';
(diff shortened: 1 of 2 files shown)


Archidoit (dismissed):

Archidoit:
Not working for markings I created myself and that are not mentionned in ItemMarkings > getStyle

Archidoit:
Not working if I modified the color of a marking.

Example: I modify the color of the TLP:AMBER+STRICT marking to pink.
image

The marking color correctly changed in the platform, but not in the task overview (still orange) : 

image

opencti-platform/opencti-front/src/components/ItemMarkings.tsx
```diff
@@ -68,13 +68,17 @@ const ChipMarking = ({
         case 'DR':
         case 'DR-SF':
         case 'TLP:RED':
+        case 'PAP:RED':
```
Archidoit, line 71: Maybe consider fetching the markingDefinition color in ItemMarkings (via a query) and use it directly instead of using a static list

Archidoit (changes requested):

SarahBocognano:
Changes requested done

Archidoit (approved):

Issue #11195, opened 2025-06-04
Missing translations in the rule configuration page


aHenryJard:
## Description

All rule title and descriptions are not translated.

## Environment

1. OS (where OpenCTI server runs): { e.g. Mac OS 10, Windows 10, Ubuntu 16.4, etc. }
2. OpenCTI version: 6.6.14
3. OpenCTI client: { e.g. frontend or python }
4. Other environment details:

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Go on setting > customization > rule engine

## Actual Output

![Image](https://github.com/user-attachments/assets/0e35d395-725d-4a66-8933-b83491f8b055)

PR #11909, opened 2025-08-05
[frontend] Missing translations in the rule configuration page (#11195)

SarahBocognano:
### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/11195


lndrtrbn (commented):
Seems to miss:
- Report propagation
- Parent-child propagation

Which are titles of categories

CelineSebe (approved):

PR #11924, opened 2025-08-06
node24 support

xfournet:
yarn install fail when using node 24

### Proposed changes
* upgrade node-gyp transitive dependency

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/yarn.lock
+++ b/opencti-platform/opencti-graphql/yarn.lock
@@ -3070,6 +3070,15 @@ __metadata:
   languageName: node
   linkType: hard
 
+"@isaacs/fs-minipass@npm:^4.0.0":
+  version: 4.0.1
+  resolution: "@isaacs/fs-minipass@npm:4.0.1"
+  dependencies:
+    minipass: "npm:^7.0.4"
+  checksum: 10/4412e9e6713c89c1e66d80bb0bb5a2a93192f10477623a27d08f228ba0316bb880affabc5bfe7f838f58a34d26c2c190da726e576cdfc18c49a72e89adabdcf5
+  languageName: node
+  linkType: hard
+
 "@istanbuljs/schema@npm:^0.1.2":
   version: 0.1.3
   resolution: "@istanbuljs/schema@npm:0.1.3"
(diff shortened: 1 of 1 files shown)


aHenryJard (commented):
Don't we need to fix the version in the package.json ?

xfournet:
> Don't we need to fix the version in the package.json ?

node-gyp is an indirect dependency, it is not referenced in the package.json, but only in the lock file

aHenryJard:
> > Don't we need to fix the version in the package.json ?
> 
> node-gyp is an indirect dependency, it is not referenced in the package.json, but only in the lock file

I'm quite new to yarn & node ecosystem, but then anyone that redo the yarn.lock with a local nodejs 22 version will erase this change no ? I think that for transitive dependencies you can fix versions in this part 
https://github.com/OpenCTI-Platform/opencti/blob/789d37346574846bf3ccc771003b868cb223ffd9/opencti-platform/opencti-graphql/package.json#L222-L241

xfournet:
> I'm quite new to yarn & node ecosystem, but then anyone that redo the yarn.lock with a local nodejs 22 version will erase this change no ? I think that for transitive dependencies you can fix versions in this part
> 

The Node.js version does not influence the outcome when updating the yarn.lock file. This commit is the result of the `yarn up -R node-gyp` command, and it would have been the same regardless of the Node.js version used to execute the command.

Using resolutions is generally discouraged because they are difficult to maintain and could potentially break dependencies. Resolutions should only be considered as a last resort, typically when dealing with unmaintained libraries or libraries that do not use semver ranges correctly or when a patch is required.

Best practice is to keep the yarn.lock file up to date. For indirect dependencies like node-gyp, this can be achieved manually with the yarn up command or through an automated tool like Renovate. Currently, our Renovate setup is not configured to handle this, it's only address direct dependencies that are referenced in package.json files. This needs to be adjusted.

Once automatic lock file updates are implemented, the need for resolution overrides will be significantly reduced.

lndrtrbn:
I have a question that come to my mind about merge conflicts. When merging master into our current release branch we have regularly conflicts inside this yarn lock file. Most of the time the conflicts are a mess and the solution we use is:
- accepting one version no matter the side,
- after merge is finished, delete yarn lock
- run yarn command to regenerate it.

But using this solution will probably break those kind of manual fixes. How can we manage this case properly?

aHenryJard (approved):

Issue #11149, opened 2025-05-28
[PIR] Deploy Priority Intelligence Requirements feature (PIR phase 2)


romain-filigran:
## Use case

Implement first releasable version of the Priority Intelligence Requirements feature, after a first PoC phase.

PR #11946, opened 2025-08-08
[frontend/backend] PIR processes (#11149)

Archidoit:
### Proposed changes
- Pir processes (works) informations


opencti-platform/opencti-front/src/private/components/pir/PirProcessingStatus.tsx
```diff
@@ -0,0 +1,69 @@
...
+      <Drawer
+        title={t_i18n('PIR processes')}
+        open={displayProcesses}
+        onClose={() => { setDisplayProcesses(false); }}
+      >
+        <>
+          <Alert severity="info">{t_i18n('This page lists the most recent works and tasks of the current PIR')}</Alert>
+          <Tabs style={{ paddingBottom: 10 }} value={tabValue} onChange={(_, newValue) => setTabValue(newValue)}>
+            <Tab label={t_i18n('Works')} value="Works" />
+            <Tab label={t_i18n('Tasks')} value="Tasks" />
```
SouadHadjiat, line 60: are there tasks for PIRs ? I don't think so, we should have only works

lndrtrbn:
Replaced by https://github.com/OpenCTI-Platform/opencti/pull/12049

Issue #11932, opened 2025-08-07
Bulk update of label not available on the second entry


fruitcakej:
## Description
In the batch update feature:
When we select “add” or “remove” label
Add a new line
The label field is now greyed out (see screenshot)

This used to work last week

After doing the remove, the selection we did is lost so we cannot rerun a second batch update to add a label on the same selection

## Environment
1. SaaS 6.7.9

## Actual Output
Label is greyed out on the 2nd entry

## Additional information
We cannot use the “replace” feature as we would lose existing labels that we don’t want to remove

## Screenshots (optional)

<img width="1289" height="426" alt="Image" src="https://github.com/user-attachments/assets/214d8cb4-197f-44b1-8bcf-754e46363832" />

Archidoit:
This is a new functionnality implemented to avoid some issues. We didn't want to be able to do several different bulk action on the same attribute. @nino-filigran ?

Here was the initial issue leading to this change : https://github.com/OpenCTI-Platform/opencti/issues/11798

nino-filigran:
Yes indeed @Archidoit, this was exactly the reason why.

DucNg:
Hello, I was actually using this feature to remove a label from a batch observable and adding another one on the same batch of observable.

I understand that this could be an issue from a technical stand point but we actually need this in our daily workflow.

Could you at least propose a workaround?

I reiterate what was said in the original issue, we cannot use the **replace** option as it removes all existing labels that we would like to keep (plus we have no way to find out what they were, thus, they're lost).

Archidoit:
@nino-filigran Maybe we can authorize doing several actions on the same key if the actions are remove/add

nino-filigran:
If you think it's not an issue, I'm fine with it @Archidoit !

PR #11955, opened 2025-08-11
[backend/frontend] authorize several bulk actions on the same key if no replace action (#11932)

Archidoit:
### Proposed changes
Authorize bulk update on the same key if there is no 'replace' actions among them

### Related issues
https://github.com/OpenCTI-Platform/opencti/issues/11932


CelineSebe:
It works well when trying to replace the same key, to add and remove differents labels  👌

CelineSebe (approved):

Issue #10846, opened 2025-04-25
Args not taken into account for rel relationship creation


Archidoit:
## Description
When creating a rel relationship, the args existing in StixRefRelationshipAddInput are not used

### Scenario 1
Create a rel relationship via the API
Only the fromId, toId and relationship_type are taken into account among the arguments (the confidence, or start_time are not for instance)

![Image](https://github.com/user-attachments/assets/0bf7d03e-b223-4904-a139-75c192a93dbf)

### Scenario 2
In an observable, go in the Knowledge tab and create a nested ref with a start_time and a stop_time

![Image](https://github.com/user-attachments/assets/2dcb981e-1b93-4d6f-8659-3e207ccf374b)

The nested ref is created with no start_time and no stop_time
![Image](https://github.com/user-attachments/assets/4c29983b-18fb-4a92-b864-58cea803f306)

nino-filigran:
I guess scenario 2 is linked to your other bug: https://github.com/OpenCTI-Platform/opencti/issues/10778

Archidoit:
> I guess scenario 2 is linked to your other bug: [#10778](https://github.com/OpenCTI-Platform/opencti/issues/10778)

No it's not the same issue and no the same root cause. The other issue is about the edition of a stix ref and is linked to a validator in frontend

PR #11956, opened 2025-08-11
[backend/frontend] Take args into account at ref relationship creation (#10846)

Archidoit:
### Proposed changes
- When creating a nested relation from an observable, start_time and stop_time must be taken into account
- When using stixRefRelationshipAdd from API, the input start_time, stop_time and confidence must be taken into account
- In an observable Knowledge tab, in the Nested objects panel, the creator of the nested relationship should be displayed

### Related issues
https://github.com/OpenCTI-Platform/opencti/issues/10846


marieflorescontact (approved):
locally tested ok 👌

PR #11961, opened 2025-08-11
[backend/frontend] Draft into data tabs of an entity (#10831)

SarahBocognano:
### Proposed changes

* Fix from product feedback

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/10831

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/drafts/DraftContextBanner.tsx
+++ b/opencti-platform/opencti-front/src/private/components/drafts/DraftContextBanner.tsx
@@ -99,8 +99,9 @@ const DraftContextBannerComponent: FunctionComponent<DraftContextBannerComponent
       onCompleted: () => {
         if (entity_id) {
           navigate(`/dashboard/id/${entity_id}`);
+        } else {
+          navigate('/dashboard/data/import/draft');
         }
-        navigate('/dashboard/data/import/draft');
       },
     });
   };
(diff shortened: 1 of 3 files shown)


opencti-platform/opencti-front/src/private/components/drafts/DraftContextBanner.tsx
```diff
@@ -96,6 +97,9 @@ const DraftContextBannerComponent: FunctionComponent<DraftContextBannerComponent
         input: { key: 'draft_context', value: '' },
       },
       onCompleted: () => {
+        if (entity_id) {
+          navigate(`/dashboard/id/${entity_id}`);
+        }
         navigate('/dashboard/data/import/draft');
       },
```
JeremyCloarec: I'm not sure if this could cause a problem or not, but to be safe, it'd be better to do the other navigate in a 'else' block, to be sure that we don't run it when we have an entiy_id

opencti-platform/opencti-front/src/private/components/drafts/DraftContextBanner.tsx
```diff
@@ -116,6 +120,9 @@ const DraftContextBannerComponent: FunctionComponent<DraftContextBannerComponent
             onCompleted: () => {
               setApproving(false);
               MESSAGING$.notifySuccess('Draft validation in progress');
+              if (entity_id) {
+                navigate(`/dashboard/id/${entity_id}`);
+              }
               navigate('/dashboard/data/import/draft');
             },
```
JeremyCloarec: Same as above, should be rewritten as
```suggestion
              if (entity_id) {
                navigate(`/dashboard/id/${entity_id}`);
              }
              else {
                 navigate('/dashboard/data/import/draft');
              }
            },
```

JeremyCloarec (approved):

Issue #11976, opened 2025-08-12
Add to container tasks can lead to errors when including neighbors is activated


JeremyCloarec:
## Description

Add to containers tasks can lead to a 'too_many_clauses: too_many_clauses: maxClauseCount is set to 5386' error. This is due to a bug in the provisioning logic.

## Environment

1. OS (where OpenCTI server runs): { e.g. Mac OS 10, Windows 10, Ubuntu 16.4, etc. }
2. OpenCTI version: { e.g. OpenCTI 1.0.2 }
3. OpenCTI client: { e.g. frontend or python }
4. Other environment details:

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. { e.g. Run ... }
2. { e.g. Click ... }
3. { e.g. Error ... }

PR #11977, opened 2025-08-12
[backend] fix issue in buildContainersElementsBundle when withNeighbours was activated (#11976)

JeremyCloarec:
### Proposed changes

* when withNeighbours is activated, we now list the neighbours outside the for loop instead of inside. Doing it inside the for loop was incorrect: we were refetching the relations each time we added a new element to the list of container elements to add, causing too many neighbors to be fetched
*

### Related issues

* #11976
*

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/manager/taskManager.js
+++ b/opencti-platform/opencti-graphql/src/manager/taskManager.js
@@ -319,19 +319,19 @@ export const buildContainersElementsBundle = async (context, user, containers, e
     const element = elements[index];
     elementIds.add(element.internal_id);
     elementStandardIds.add(element.standard_id);
-    if (withNeighbours) {
-      if (element.fromId) elementIds.add(element.fromId);
-      if (element.toId) elementIds.add(element.toId);
-      const callback = (relations) => {
-        relations.forEach((relation) => {
-          elementIds.add(relation.fromId);
-          elementIds.add(relation.toId);
-          elementIds.add(relation.id);
-        });
-      };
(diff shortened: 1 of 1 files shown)


opencti-platform/opencti-graphql/src/manager/taskManager.js
```diff
@@ -319,19 +319,19 @@ export const buildContainersElementsBundle = async (context, user, containers, e
...
-          elementIds.add(relation.fromId);
-          elementIds.add(relation.toId);
-          elementIds.add(relation.id);
-        });
-      };
-      const args = { fromOrToId: Array.from(elementIds), baseData: true, callback };
-      await listAllRelations(context, user, ABSTRACT_STIX_CORE_RELATIONSHIP, args);
-    }
+    if (element.fromId) elementIds.add(element.fromId);
+    if (element.toId) elementIds.add(element.toId);
```
SouadHadjiat, line 323: so here even if `withNeighbours` is false, when an element is a relationship, we will add the relationship and the `from` and `to` to the container, is it right ?

JeremyCloarec: yes, this wasn't the case before and it's not related to the bug, but I thought that it would be better like this: if a user targets relationships in his tasks, and wants to add the relationships to a container, I figured that even if `withNeighbours` isn't selected, it would make sense to add the targets of the relationships. Otherwise adding them to the container doesn't really do anything

SouadHadjiat (approved):

PR #11988, opened 2025-08-12
[backend] optimize checkFilterKeys function (#11987)

### LINKED ISSUE #11987: Improve performance of checkFilterKeys function
## Description

Function checkFilterKeys can be improved.

Currently required a lot of CPU in some situations

Image

xfournet:
See #11987

Changed:
- algo was O(n²) on keys array, now is O(n)
- cache the available keys set since it's not updated during the execution lifetime

### DIFF
--- a/opencti-platform/opencti-graphql/src/utils/filtering/filtering-utils.ts
+++ b/opencti-platform/opencti-graphql/src/utils/filtering/filtering-utils.ts
@@ -386,15 +386,9 @@ export const replaceEnrichValuesInFilters = (filterGroup: FilterGroup, userId: s
   return filtersResult;
 };
 
-/**
- * Check the filter keys exist in the schema
- */
-const checkFilterKeys = (filterGroup: FilterGroup) => {
-  // TODO improvement: check filters keys correspond to the entity types if types is given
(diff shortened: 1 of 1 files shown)


opencti-platform/opencti-graphql/src/utils/filtering/filtering-utils.ts
```diff
@@ -408,14 +402,22 @@ const checkFilterKeys = (filterGroup: FilterGroup) => {
...
+};
+
+/**
+ * Check the filter keys exist in the schema
+ */
+const checkFilterKeys = (filterGroup: FilterGroup) => {
+  // TODO improvement: check filters keys correspond to the entity types if types is given
+  const incorrectKeys = extractFilterKeys(filterGroup)
+    .map((k) => k.split('.')[0]) // keep only the first part of the key to handle composed keys
+    .filter((k) => !(getAvailableKeys().has(k) || k.startsWith(RULE_PREFIX)));
```
SouadHadjiat, line 417: I wonder if we could initialize the `availableKeysCache` Set outside of this method, but otherwise it's a good improvement

SouadHadjiat (approved):

Issue #11964, opened 2025-08-12
Prevent users from seing other user's from different org than their own orgs in assignee & participant field


nino-filigran:
## Use case

To ensure proper data segregation, we need to ensure that a user can only see users of its own organization in the field Assignee & Participant.

PR #12012, opened 2025-08-13
[backend] Prevent users from seing other user's from different org than their own orgs in assignee & participant field  (#11964)

Gwendoline-FAVRE-FELIX:
### Proposed changes

* Filter members by user orgs if platform org active and user outside

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/11964

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/domain/user.js
+++ b/opencti-platform/opencti-graphql/src/domain/user.js
@@ -57,6 +57,7 @@ import {
   isBypassUser,
   isOnlyOrgaAdmin,
   isUserHasCapability,
+  isUserInPlatformOrganization,
   REDACTED_USER,
   SETTINGS_SET_ACCESSES,
   SYSTEM_USER,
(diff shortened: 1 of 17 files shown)


opencti-platform/opencti-graphql/src/domain/user.js
```diff
@@ -206,24 +207,26 @@ export const findAll = async (context, user, args) => {
   return listEntities(context, user, [ENTITY_TYPE_USER], args);
 };
 
-export const findCreators = (context, user, args) => {
+export const findCreators = async (context, user, args) => {
```
marieflorescontact: why do we had async but no await?

opencti-platform/opencti-graphql/src/resolvers/log.js
```diff
@@ -14,7 +15,14 @@ const logResolvers = {
...
   },
   Log: {
-    user: (log, _, context) => context.batch.creatorBatchLoader.load(log.applicant_id || log.user_id),
+    user: async (log, _, context) => {
+      const realUser = await context.batch.creatorBatchLoader.load(log.applicant_id || log.user_id);
+      if (!realUser) {
+        return null;
+      }
+      const filteredUser = await filterMembersWithUsersOrgs(context, context.user, [realUser]);
+      return filteredUser[0];
```
marieflorescontact, line 24: Could we avoid `filteredUser[0]` ?

marieflorescontact (commented):
functionnally tested ok 👌

opencti-platform/opencti-graphql/src/domain/user.js
marieflorescontact: seems to be fixed now 🙂

marieflorescontact (approved):

Issue #11985, opened 2025-08-12
Decay manager is causing high CPU consumption on Indicator with more than 1000 decay_history


aHenryJard:
## Description

For some reason, some indicator can have more than 1k (we saw some with 22k) decay history. It's the data that store all the point that are displayed in the lifecycle popup.

It's causing CPU issue on decay manager, the related error looks like
```
{
  "category": "APP",
  "key": "decay_history",
  "level": "error",
  "message": "[ENGINE] Element preparation too many values",
  "size": 22393,
  "source": "backend",
  "timestamp": "2025-08-12T15:01:23.792Z",
  "version": "6.7.9"
}
```

It's not usefull to have tha many point, we want to keep the 25 first and 25 last when decay_history size is > 50 elements.

## Environment

1. OS (where OpenCTI server runs): { e.g. Mac OS 10, Windows 10, Ubuntu 16.4, etc. }
2. OpenCTI version: { e.g. OpenCTI 1.0.2 }
3. OpenCTI client: { e.g. frontend or python }
4. Other environment details:

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. { e.g. Run ... }
2. { e.g. Click ... }
3. { e.g. Error ... }

PR #12022, opened 2025-08-14
[backend] trim decay history size (#11985)

JeremyCloarec:
### Proposed changes

* trim decay history size to only keep 50 points max
*

### Related issues

* #11985 
*

### Checklist


opencti-platform/opencti-graphql/src/modules/indicator/indicator-domain.ts
```diff
@@ -344,6 +344,22 @@ export const addIndicator = async (context: AuthContext, user: AuthUser, indicat
   return notify(BUS_TOPICS[ABSTRACT_STIX_DOMAIN_OBJECT].ADDED_TOPIC, created, user);
 };
 
+export const MAX_DECAY_HISTORY_POINTS = 50;
```
SouadHadjiat, line 347: would it need to be configurable ?

SouadHadjiat (approved):
I think it will solve most of the issues when the indicator is updated (by decay manager or directly), I just wonder if this will work when upserting the indicator (from connectors). But we could I guess fix this in a separate PR.

Issue #12000, opened 2025-08-13
User overview header layout is broken


SamuelHassine:
## Description

User overview header layout is broken (too much space between the name and the tabs):

Correct:

<img width="670" height="195" alt="Image" src="https://github.com/user-attachments/assets/d35fd982-0b47-426e-ba80-fe8ec4e3cbc5" />

Incorrect:

<img width="552" height="294" alt="Image" src="https://github.com/user-attachments/assets/d3cf158c-63b8-4033-9fb6-30c0a72d4530" />

PR #12042, opened 2025-08-18
[frontend] user header layout (#12000)

Archidoit:
### Proposed changes
Fix layout regression: too many margin between user header and user tabs in user overview.

### Related issues
https://github.com/OpenCTI-Platform/opencti/issues/12000

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/settings/users/Root.jsx
+++ b/opencti-platform/opencti-front/src/private/components/settings/users/Root.jsx
@@ -1,12 +1,11 @@
 import React, { useEffect, useMemo, useState } from 'react';
 import * as PropTypes from 'prop-types';
-import { Link, Route, Routes, useParams, useLocation } from 'react-router-dom';
+import { Link, Route, Routes, useLocation, useParams } from 'react-router-dom';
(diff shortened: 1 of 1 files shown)


opencti-platform/opencti-front/src/private/components/settings/users/Root.jsx
```diff
@@ -119,7 +109,7 @@ const RootUserComponent = ({ queryRef, userId, refetch }) => {
   const UserHeader = styled('div')({
     display: 'flex',
     justifyContent: 'space-between',
-    marginBottom: 24,
+    marginBottom: 8,
```
ValentinBouzinFiligran, line 112: Do we need this line now ?
I tried to remove it and it looks better for me 🤔

ValentinBouzinFiligran: With margin 8:
image


Without: 
image

Archidoit: I put 8 to have exactly the same visual as for Attack patterns and other entities

ValentinBouzinFiligran (approved):

Issue #12066, opened 2025-08-19
Add scrolling on the support package component in Filigran Experience


EllynBsc:
## Description

At the moment, we cannot scroll the support package, we have to scroll the entire page.

For a better UX, in the settings > filigran experience, implement a scroll on the support package section.

SamuelHassine:
I think this one is just applying new data table to the support package list with inline scrolling and pagination

EllynBsc:
cc @Kroustille

PR #12082, opened 2025-08-20
[frontend] fix: add scroll to support packages

Kroustille:
### Proposed changes

* fix height of support packages block
* add overflow scroll to enable scrolling only on this block

### Related issues

* #12066

### Checklist

### Further comments

Another solution was to update component to use a datatable but it does not have a great ROI since this component won't often have a lot of values and this solution is much faster.

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/settings/support/SupportPackages.tsx
+++ b/opencti-platform/opencti-front/src/private/components/settings/support/SupportPackages.tsx
@@ -146,7 +146,8 @@ const SupportPackages = () => {
       <div className="clearfix"/>
       <Paper className="paper-for-grid" variant="outlined" sx={{
         height: '100%',
-        minHeight: '100%',
+        maxHeight: '600px',
+        overflowY: 'auto',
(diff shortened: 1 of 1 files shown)


opencti-platform/opencti-front/src/private/components/settings/support/SupportPackages.tsx
```diff
@@ -146,7 +146,8 @@ const SupportPackages = () => {
       <div className="clearfix"/>
       <Paper className="paper-for-grid" variant="outlined" sx={{
         height: '100%',
-        minHeight: '100%',
+        maxHeight: '600px',
+        overflow: 'scroll',
```
delemaf: Scrollbar are displayed all the time, both in X and Y. 
We only need in Y and when the list overflow the max height.
image

This would fix the overflow: 
```suggestion
        overflowY: 'auto',
```

delemaf (changes requested):

delemaf (approved):

PR #12092, opened 2025-08-21
[Backend/Frontend] Secure credentials in composer (#12013)

Gwendoline-FAVRE-FELIX:
### Proposed changes

* Encrypt credentials of type "password" in contract

Composer PR: https://github.com/FiligranHQ/xtm-composer/pull/13

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/12013

### Checklist


opencti-platform/opencti-graphql/src/modules/attributes/internalObject-registrationAttributes.ts
```diff
@@ -408,6 +408,7 @@ const internalObjectsAttributes: { [k: string]: Array<AttributeDefinition> } = {
   [ENTITY_TYPE_CONNECTOR_MANAGER]: [
     { ...updatedAt, update: true }, // Allow change of updated_at for connector ping
     { featureFlag: COMPOSER_FF, name: 'name', label: 'Name', type: 'string', format: 'short', mandatoryType: 'no', editDefault: false, multiple: false, upsert: false, isFilterable: true },
+    { featureFlag: COMPOSER_FF, name: 'public_key', label: 'PublicKey', type: 'string', format: 'short', mandatoryType: 'external', editDefault: false, multiple: false, upsert: false, isFilterable: true },
```
aHenryJard, line 412: it's one public key per connector ? I understood that it's more one composer public key for the whole platform. So I was expecting to see the public key in settings.

JeremyCloarec: this public_key here is on the connector_manager (aka the composer). And there is only one connector_manager per platform (when a connector_manager registers, it replaces the old one)

maelv-filigran:
The public key is stored as a plain string in the database without any validation or format checking.

__Questions to consider__:

- What happens if an invalid public key format is provided during registration?
- Should there be a mechanism to validate the public key format before storing it?
-  If we store a path to a key file rather than the key content (as discussed with @JeremyCloarec), should we still validate that the file contains a well-formed, usable RSA public key?

__Suggestion__: Consider adding validation for the RSA public key format in the `registerConnectorsManager` function. You might want to verify that the key can actually be used for encryption before accepting it.

opencti-platform/opencti-graphql/src/modules/catalog/catalog-domain.ts
```diff
@@ -77,20 +78,40 @@ const getCatalogs = () => {
   return catalogMap;
 };
 
-export const computeConnectorTargetContract = (configurations: any, targetContract: CatalogContract) => {
+const encryptValue = (publicKey: string, value: string) => {
```
maelv-filigran, line 82: How do you handle encryption failures gracefully?
Suggestion: Wrap the encryption logic in a try-catch block and provide meaningful error messages.

opencti-platform/opencti-graphql/src/modules/catalog/catalog-domain.ts
```diff
@@ -77,20 +78,40 @@ const getCatalogs = () => {
...
+  const buffer = Buffer.from(value, 'utf8');
+  const encrypted = crypto.publicEncrypt(
+    {
+      key: publicKey,
+      padding: crypto.constants.RSA_PKCS1_PADDING,
```
maelv-filigran, line 87: RSA with PKCS#1 v1.5 padding is considered legacy.
Should we switch to RSA-OAEP for encryption to improve security?

opencti-platform/opencti-graphql/src/modules/catalog/catalog-domain.ts
```diff
@@ -77,20 +78,40 @@ const getCatalogs = () => {
...
+    },
+    buffer
+  );
+  return encrypted.toString('base64');
+};
+
+export const computeConnectorTargetContract = (configurations: any, targetContract: CatalogContract, publicKey: string) => {
```
maelv-filigran: ### Critical Issue: Re-encryption of Already Encrypted Values

__The code blindly encrypts any value marked as a password field, without checking if it's already encrypted.__

This means:

1. __First save__: Password gets encrypted once ✓
2. __Second save__: Already encrypted password gets encrypted again ❌
3. __Third save__: Double-encrypted password gets encrypted again ❌❌

This creates a cascading encryption problem where each update adds another layer of encryption, making the value completely unusable.

## Why This Is Critical

1. __Data Corruption__: After the first update, the password becomes double-encrypted and unusable
2. __Irreversible__: Without the private keys from each encryption layer, you can't decrypt back to the original
3. __Silent Failure__: The system won't throw errors during encryption, but the connector will fail when trying to use the password

The system needs to detect if a value is already encrypted before applying encryption. Here are the key questions:

- How can you differentiate between a plain text password and an encrypted one?
- Should the backend avoid sending encrypted values to the frontend and instead return only a flag (e.g., isSecret) indicating that a value exists?

maelv-filigran:
tested locally, works :)

maelv-filigran (approved):

Issue #4523, opened 2023-10-09
Be able to export / import playbooks


SamuelHassine:
## Use case

Be able to export / import playbooks

Lhorus6:
Note: Consider adding a "duplicate" button too

cmandich:
👋 Hi folks, looks like this issue will soon be open for close to 2 years. We just had this feature request internally as well. It'd be great to see some traction or get an estimated release date for this request as it appears multiple folks have requested / taken note of it. Best, Chris.

PR #12126, opened 2025-08-25
[backend/frontend] Be able to duplicate / import and export playbooks (#4523)

richard-julien:
See #4523


SarahBocognano:
image


Only when left menu is open

SarahBocognano:
The PR works perfectly ! Import, Export and Duplicate are fine. Error message for types/version as well 

Only the small UI display with the add of the import button in our regular pc screen

PR #12131, opened 2025-08-25
[frontend] add filters catalog composer (#7328)

esrevi:
### Proposed changes

* Add filters on title, description, types and use_cases on ingestion catalogue contracts
* Handle empty state when no catalogue match the filters criteria

https://github.com/user-attachments/assets/45c0bd1e-5f4d-40cb-93b6-e8ded6c0f84b

### Related issues
https://github.com/OpenCTI-Platform/opencti/issues/7328

### Checklist


opencti-platform/opencti-front/src/private/components/data/IngestionCatalog/IngestionCatalogFilters.tsx
```diff
@@ -0,0 +1,156 @@
...
+  const handleClearFilters = () => {
+    setSearchInput('');
+    onFiltersChange({ search: '', type: '', useCase: '' });
+  };
+
+  const hasActiveFilters = filters.search || filters.type || filters.useCase;
+
+  return (
+    <Stack flexDirection="row" gap={2} flex={1}>
+      <TextField
```
frapuks: You can use our SearchInput component, to be align with the other screen.

opencti-platform/opencti-front/src/private/components/data/IngestionCatalog/IngestionCatalogFilters.tsx
```diff
@@ -0,0 +1,156 @@
...
+}) => {
+  const { t_i18n } = useFormatter();
+  const [searchInput, setSearchInput] = useState(filters.search);
+
+  useEffect(() => {
+    const handler = setTimeout(() => {
+      if (searchInput !== filters.search) {
+        onFiltersChange({ ...filters, search: searchInput });
+      }
+    }, DEBOUNCE_DELAY);
```
frapuks: We don't have automatically search after delay in the rest of the platform. Currently, we submit with Enter button.

opencti-platform/opencti-front/src/private/components/data/IngestionCatalog/IngestionCatalogFilters.tsx
```diff
@@ -0,0 +1,156 @@
...
+        }}
+      />
+
+      <Autocomplete
+        size="small"
+        sx={{ width: INPUT_WIDTH }}
+        options={filterOptions.types}
+        value={filterOptions.types.find((o) => o.value === filters.type) || null}
+        onChange={(event, option) => handleFilterChange('type', option?.value || '')}
+        getOptionLabel={(option) => option.label}
```
frapuks: You don't translate labels. This list of 4 types is already translated in our languages files, you can just use t_i18n.

opencti-platform/opencti-front/src/private/components/data/IngestionCatalog/IngestionCatalogFilters.tsx
```diff
@@ -0,0 +1,156 @@
...
+        clearOnEscape
+      />
+
+      <Autocomplete
+        size="small"
+        sx={{ width: INPUT_WIDTH }}
+        options={filterOptions.useCases}
+        value={filterOptions.useCases.find((o) => o.value === filters.useCase) || null}
+        onChange={(event, option) => handleFilterChange('useCase', option?.value || '')}
+        getOptionLabel={(option) => option.label}
```
frapuks: For use cases, maybe ask product team to know what they want about the translation.

frapuks (commented):
Overall, it works, and I'm okay with it after the changes made to my comments. But I think it would be better if a senior front-end developer also reviewed it.

opencti-platform/opencti-front/src/private/components/data/IngestionCatalog/IngestionCatalogFilters.tsx
esrevi: ok, this behavior will be handled with the use of SearchInput i guess

opencti-platform/opencti-front/src/private/components/data/IngestionCatalog/IngestionCatalogFilters.tsx
esrevi: for the record, product agrees on not translating the labels of use cases

opencti-platform/opencti-front/src/private/components/data/IngestionCatalog/IngestionCatalogFilters.tsx
```diff
@@ -0,0 +1,143 @@
...
+        label: type,
+      })),
+      useCases: useCases.sort().map((useCase) => ({
+        value: useCase,
+        label: useCase,
+      })),
+    };
+  }, [contracts]);
+
+  const handleFilterChange = (key: keyof FilterState, value: string) => {
```
Archidoit, line 67: There is already a generic logic to handle filters (add, change, remove, apply on data, etc) you can reuse. Did you reimplemant the logic on purpose for your case ?

Archidoit (changes requested):

opencti-platform/opencti-front/src/private/components/data/IngestionCatalog/IngestionCatalogFilters.tsx
Archidoit: You can look how filters have been implemented for ListCards of malwares for instance

esrevi: as discussed with @Archidoit: The reason of the implementation is the contract items are json strings, and it cannot work as it is with the Filters component without more dev I think

Archidoit: I didn't follow this feature, so I don't know why contracts are stored as string. But just as a remark, I think that if we want to filter / order by contracts attributes, we should have stored contracts as nested objects instead of string... @natalie-natsu @ValentinBouzinFiligran @frapuks

natalie-natsu: This is uncommon, but the manifest (the config file of all connectors) is a just a json file that we fetch in the back-end from the connectors repo. So we don't really store theses informations. Maybe it would be great to normalized this manifest in the DB one day, but for now we keep it straight forward like that. :-) Thanks for your review Cathia, I resolve this one comment ;-)

Archidoit (approved):

Archidoit:
Filters are working well

PR #12133, opened 2025-08-26
[frontend/backend] Proxify chatbot queries (#11027)

delemaf:
### Proposed changes

* Proxify chatbot queries and remove url/token/certificate from frontend calls

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/11027

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/config/conf.js
+++ b/opencti-platform/opencti-graphql/src/config/conf.js
@@ -348,7 +348,7 @@ const baseUrl = BasePathUrl.endsWith('/') ? BasePathUrl.slice(0, -1) : BasePathU
 
 export const getBaseUrl = (req) => {
   // If base url is defined, take it in priority
-  if (baseUrl) {
+  if (baseUrl && !baseUrl.includes('localhost')) {
     // Always append base path to the uri
     return baseUrl + basePath;
   }
--- a/opencti-platform/opencti-graphql/src/config/conf.js
+++ b/opencti-platform/opencti-graphql/src/config/conf.js
@@ -348,7 +348,7 @@ const baseUrl = BasePathUrl.endsWith('/') ? BasePathUrl.slice(0, -1) : BasePathU
 
 export const getBaseUrl = (req) => {
   // If base url is defined, take it in priority
-  if (baseUrl && !baseUrl.includes('localhost')) {
+  if (baseUrl) {
     // Always append base path to the uri
     return baseUrl + basePath;
   }
@@ -361,6 +361,16 @@ export const getBaseUrl = (req) => {
   }
   // If no base url and no request, send only the base path
   return basePath;
+};
+
+export const getChatbotUrl = (req) => {
+  if (req) {
+    const [, port] = req.headers.host ? req.headers.host.split(':') : [];
+    const isCustomPort = port !== '80' && port !== '443';
+    const httpPort = isCustomPort && port ? `:${port}` : '';
+    return `${req.protocol}://${req.hostname}${httpPort}${basePath}`;
+  }
+  throw UnknownError('Missing request for chatbot');
 };
 
 export const configureCA = (certificates) => {
--- a/opencti-platform/opencti-graphql/src/domain/settings.js
+++ b/opencti-platform/opencti-graphql/src/domain/settings.js
@@ -103,10 +103,7 @@ export const getProtectedSensitiveConfig = async (context, user) => {
 export const getSettings = async (context) => {
   const platformSettings = await loadEntity(context, SYSTEM_USER, [ENTITY_TYPE_SETTINGS]);
   const clusterInfo = await getClusterInformation();
-  const {
-    license_raw_pem: _,
-    ...eeInfo
-  } = getEnterpriseEditionInfoFromPem(platformSettings.internal_id, platformSettings.enterprise_license);
+  const eeInfo = getEnterpriseEditionInfoFromPem(platformSettings.internal_id, platformSettings.enterprise_license);
 
   return {
     ...platformSettings,
--- a/opencti-platform/opencti-graphql/src/domain/settings.js
+++ b/opencti-platform/opencti-graphql/src/domain/settings.js
@@ -103,7 +103,11 @@ export const getProtectedSensitiveConfig = async (context, user) => {
 export const getSettings = async (context) => {
   const platformSettings = await loadEntity(context, SYSTEM_USER, [ENTITY_TYPE_SETTINGS]);
   const clusterInfo = await getClusterInformation();
-  const eeInfo = getEnterpriseEditionInfoFromPem(platformSettings.internal_id, platformSettings.enterprise_license);
+  const {
+    license_raw_pem: _,
+    ...eeInfo
+  } = getEnterpriseEditionInfoFromPem(platformSettings.internal_id, platformSettings.enterprise_license);
+
   return {
     ...platformSettings,
     platform_url: getBaseUrl(context.req),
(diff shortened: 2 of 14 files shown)


opencti-platform/opencti-graphql/src/http/httpChatbotProxy.ts
```diff
@@ -0,0 +1,88 @@
...
+    }
+    if (!chatbotUrl) {
+      res.status(400).json({ error: 'Chatbot proxy not properly configured' });
+      return;
+    }
+
+    const vars = {
+      OPENCTI_URL: settings.platform_url,
+      OPENCTI_TOKEN: context.user?.api_token,
+      OPENCTI_CERTIFICATE: Buffer.from(license_raw_pem, 'utf-8').toString('base64'),
```
aHenryJard: Can you add X-API-KEY header please ? Platform team need it as soon as possible.

aHenryJard (commented):
Please remove the PEM exposed from settings too, not sure but I think it's this part
(in frontend => `platform_enterprise_edition.license_raw_pem`)

https://github.com/OpenCTI-Platform/opencti/blob/ae493b4a34a1b5dd62c2aaaace7d46aa994524ba/opencti-platform/opencti-graphql/src/domain/settings.js#L109

opencti-platform/opencti-graphql/src/http/httpChatbotProxy.ts
```diff
@@ -0,0 +1,88 @@
...
+    if (!isChatbotEnabled || !license_raw_pem) {
+      return;
+    }
+    if (!chatbotUrl) {
+      res.status(400).json({ error: 'Chatbot proxy not properly configured' });
+      return;
+    }
+
+    const vars = {
+      OPENCTI_URL: settings.platform_url,
```
richard-julien: You forget what we said about the default platform_url on localhost.
If localhost is configured you must try to use the referer domain of the request

richard-julien: If localhost, use getBaseUrl(req)

JeremyCloarec: getBaseUrl doesn't handle the localhost case currently, this logic will need to be added. Currently, it always uses the base_url if it is set, even if it is localhost

JeremyCloarec: also, I didn't remember that, but settings.platform_url is always set using getBaseUrl. So if getBaseUrl is changed to handle the localhost case, we can keep using settings.platform_url in any case
image

richard-julien: Ok so in fact is simpler than that.

only getBaseUrl must be a bit change to not relay on the base_path is its the localhost default config

richard-julien: Concurrent messaging :)

opencti-platform/opencti-graphql/src/domain/settings.js
```diff
@@ -103,7 +103,11 @@ export const getProtectedSensitiveConfig = async (context, user) => {
 export const getSettings = async (context) => {
   const platformSettings = await loadEntity(context, SYSTEM_USER, [ENTITY_TYPE_SETTINGS]);
   const clusterInfo = await getClusterInformation();
-  const eeInfo = getEnterpriseEditionInfoFromPem(platformSettings.internal_id, platformSettings.enterprise_license);
+  const {
+    license_raw_pem: _,
+    ...eeInfo
+  } = getEnterpriseEditionInfoFromPem(platformSettings.internal_id, platformSettings.enterprise_license);
+
```
JeremyCloarec: Instead of removing the license_raw_pem here, you can remove it from the return of getEntrepriseEditionInforFromPem.
Also, it needs to be removed from PlatformEE in opencti.graphql

opencti-platform/opencti-graphql/src/http/httpChatbotProxy.ts
```diff
@@ -0,0 +1,88 @@
...
+      res.sendStatus(403);
+      return;
+    }
+
+    const chatbotUrl = nconf.get('ai:chatbot:url');
+    const isChatbotEnabled = nconf.get('ai:chatbot:enabled');
+    const settings = await getEntityFromCache<BasicStoreSettings>(context, context.user, ENTITY_TYPE_SETTINGS);
+    const { license_raw_pem } = getEnterpriseEditionInfo(settings);
+    if (!isChatbotEnabled || !license_raw_pem) {
+      return;
```
JeremyCloarec, line 26: We'll probably want to return an error in this case, but I'm not sure which one yet. We can already send a 400 error even if we'll need to discuss further the error message

opencti-platform/opencti-graphql/src/http/httpChatbotProxy.ts
```diff
@@ -0,0 +1,88 @@
...
+    const headers = {
+      'Content-Type': 'application/json',
+      Accept: 'text/event-stream',
+      ...vars,
+    };
+
+    const enhancedBody = {
+      ...req.body,
+      overrideConfig: {
+        ...req.body.overrideConfig,
```
JeremyCloarec: I think body could be null here, so to be safe we could null check this. I think it'd be even better to check the content of the request before, and send an error if there is no body

opencti-platform/opencti-graphql/src/config/conf.js
```diff
@@ -348,7 +348,7 @@ const baseUrl = BasePathUrl.endsWith('/') ? BasePathUrl.slice(0, -1) : BasePathU
 
 export const getBaseUrl = (req) => {
   // If base url is defined, take it in priority
-  if (baseUrl) {
+  if (baseUrl && !baseUrl.includes('localhost')) {
```
aHenryJard: I don't understand this change, I use basePath locally sometime for some specific tests. And also it will not work for 127.0.0.1.

opencti-platform/opencti-graphql/src/http/httpChatbotProxy.ts
```diff
@@ -0,0 +1,92 @@
...
+    if (!context.user) {
+      res.sendStatus(403);
+      return;
+    }
+
+    const chatbotUrl = nconf.get('ai:chatbot:url');
+    const isChatbotEnabled = nconf.get('ai:chatbot:enabled');
+    const settings = await getEntityFromCache<BasicStoreSettings>(context, context.user, ENTITY_TYPE_SETTINGS);
+    const license_pem = getEntrepriseEditionActivePem(settings.enterprise_license);
+    if (!isChatbotEnabled || !license_pem) {
```
aHenryJard, line 24: We need to check if license is valid also no ? I mean if a platform has an expired license the request to chatbot will be send. I think you need to add isEnterpriseEdition from ee.ts.

Not sure but maybe also a check on FF ?

opencti-platform/opencti-graphql/src/http/httpChatbotProxy.ts
aHenryJard: ok I understand above change now, so we don't want to be able to use a basePath different from empty locally anymore ?

aHenryJard (approved):
Looks good to me

opencti-platform/opencti-graphql/src/http/httpChatbotProxy.ts
delemaf: We'll do this in another PR

opencti-platform/opencti-graphql/src/http/httpChatbotProxy.ts
delemaf: To prevent side effect on getBaseUrl I've created a getChatbotUrl to isolate this logic

PR #12143, opened 2025-08-26
[frontend] add tooltip over relationships names (#11934)

### LINKED ISSUE #11934: When hovering over a relation, view its name
## Use case

Often, in list, on a small screen, the relation type is not fully visible. When relation type names are close, it would be useful to see them to understand what the relation is without having to click on the relation to view it, and then click back to go back to the page we were.

## Proposed Solution

Add a hovering effect on the relation type icon in views: 
- entity overview (displaying some relations)
- knowledge view of entities (relation view)
- knowledge view of observables
- data/relation

labo-flg:
### Proposed changes

* add tooltip over relationship and entity chip item
* add missing translation over the relationship name

### Related issues

* closes #11934 

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/components/ItemEntityType.tsx
+++ b/opencti-platform/opencti-front/src/components/ItemEntityType.tsx
@@ -8,7 +8,6 @@ import { useFormatter } from './i18n';
 import ThemeLight from './ThemeLight';
 import ThemeDark from './ThemeDark';
 import ItemIcon from './ItemIcon';
-import { truncate } from '../utils/String';
 import { chipInListBasicStyle } from '../utils/chipStyle';
 
 // Deprecated - https://mui.com/system/styles/basics/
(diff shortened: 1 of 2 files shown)


opencti-platform/opencti-front/src/components/ItemEntityType.tsx
```diff
@@ -99,23 +98,23 @@ const ItemEntityType: FunctionComponent<ItemEntityTypeProps> = ({
   };
   const getLabel = () => {
     if (isRestricted) return t_i18n('Restricted');
-    const label = t_i18n(isRelationship ? `relationship_${entityType}` : `entity_${entityType}`);
-    if (maxLength) return truncate(label, maxLength);
```
JeremyCloarec, line 103: Shouldn't we keep this maxLength config for the label of the chip still? I'm not sure if it could break something or not. 
We could keep the trucating for the label of the chip, but not truncate in the tooltip

labo-flg: I started by doing just so, but after search, this prop appears to be unused in the app.
Also, I do not understand why we would have to truncate the label in the chip ; by default mui already truncate with "..." when the label is too big for the space (which is exactly what the truncate utility is doing).

JeremyCloarec (approved):

PR #12146, opened 2025-08-26
[backend] update get-connectors-manifest script (#7328)

natalie-natsu:
Update of the script to get the connectors manifest.


.circleci/config.yml
```diff
@@ -49,6 +49,8 @@ jobs:
           command: yarn install
       - run:
           working_directory: ~/opencti/opencti-platform/opencti-graphql
+          environment:
+            CONNECTORS_MANIFEST_REF: stable
```
aHenryJard: This will not guaranty that we will get the right manifest version, do you think it's possible to use the version (${CIRCLE_TAG} maybe) instead ? Better to fail release than do an inconsistent release IMO.

opencti-platform/opencti-graphql/package.json
```diff
@@ -7,8 +7,8 @@
     "install:python": "pip3 install -r src/python/requirements.txt",
     "build:types": "graphql-codegen --config graphql-codegen.yml",
     "build:schema": "yarn build:types && node builder/schema/schema.js && node build/script-generate-schema.js",
-    "build:prod": "yarn build:types && yarn check-ts && node builder/prod/prod.js",
-    "build:dev": "yarn build:schema && node builder/dev/dev.js",
+    "build:prod": "yarn build:types && yarn get-connectors-manifest && yarn check-ts && node builder/prod/prod.js",
+    "build:dev": "yarn build:schema && yarn get-connectors-manifest && node builder/dev/dev.js",
```
frapuks: I try to run and i get an error :
image

It seems to need to run the get-connectors-manifest befor the build:schema

.circleci/config.yml
natalie-natsu: I can do that yes. :-)

opencti-platform/opencti-graphql/package.json
natalie-natsu: Oh, thanks, I will put it before schemas.

.circleci/config.yml
natalie-natsu: I removed it to target only master for now

opencti-platform/opencti-graphql/package.json
natalie-natsu: Done

aHenryJard (approved):
All good to me, just need to update the corresponding test to get CI green and it's fine.

maelv-filigran:
I saw you had the catalog file hardcoded in the tests. That did help confirm the downloaded file matched the expected format. You’re right it isn’t the test’s responsibility, so I’m good with your change. 
My concern: we may no longer catch corruption during transfer. What do you think about validating the manifest against the schema with Ajv in get-connectors-manifest.js?

maelv-filigran (approved):

Issue #12044, opened 2025-08-13
Note Field Is not hidden


DKboyzz:
## Description

When trying to hide the Notes for Users using the "hide notes" function under Groups or Organizations it successfully hides the user's ability to create and edit notes. But when someone that can access the Note type creates a Note it becomes visible to those we want to hide it for.
 
Screenshot 1) shows that the Analyst can't see, create or access notes.
then the manager with note access creates a note.
Screenshot 2) shows that the analyst can now see, edit and update the notes from someone else.
## Environment

OpenCTI version: OpenCTI 6.7.10 (SaaS)

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1) Create two accounts within the same organization. Account A has "hide notes" enabled and Account B has no section hidden.
2) Account B creates a Note in a report.
3) Account A can view and edit this note in the report, possibly changing its status, label and triggering a workflow if necessary and built based on the note type.

## Expected Output

The notes should stay hidden, or at least the user (Account A) shouldn't be able to edit them and change critical elements of it.

## Actual Output

Account A can view and edit this note in the report, possibly changing its status, label and triggering a workflow if necessary and built based on the note type.

## Screenshots (optional)

View of Account A not being able to view notes or edit. 
<img width="1310" height="367" alt="Image" src="https://github.com/user-attachments/assets/0150d590-52be-438f-a7b1-01bb873914b7" />

After Account B creates a note, Analyst A has not only view but edit rights.
<img width="1308" height="848" alt="Image" src="https://github.com/user-attachments/assets/4e27dc57-e7ad-40f9-9aae-e543f0b2d648" />

SarahBocognano:
I'm sorry I don't understand very well where you found the "hide notes" function.  Can you explain to me the steps to reproduce this part please ?

DKboyzz:
@SarahBocognano ,

Steps (using privileged account):
1. Settings
2. Security
3. Organization
4. For any given Organization, select: "Hidden Entity Type":

<img width="2159" height="418" alt="Image" src="https://github.com/user-attachments/assets/87dac846-3d2a-4c2d-9483-230605e1638f" />

Second method and steps:
1. Settings
2. Security
3. Groups
4. For any given group, select: "Hidden Entity Type"

<img width="1086" height="308" alt="Image" src="https://github.com/user-attachments/assets/12007ce7-bf0c-487f-8124-e4d214744075" />

SarahBocognano:
[...]

PR #12182, opened 2025-08-28
[frontend] Note Field Is not hidden (#12044)

SarahBocognano:
### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/12044


CelineSebe (approved):

CelineSebe:
Everything seems to be going well :)

lndrtrbn (changes requested):
I think it would be nice to avoid duplicating the logic in every files calling the Not component

lndrtrbn (approved):

Issue #12193, opened 2025-08-29
One-click deploy works even if the connectivity is lost


Kroustille:
Refers to https://github.com/FiligranHQ/xtm-hub/issues/882

hervyt:
Done on XTM Hub side :)

PR #12194, opened 2025-08-29
[frontend] feat: prevent one click deploy when connectivity is lost

Kroustille:
### Proposed changes

* refactor way we download a file with new hook and client
* add platform id and token in headers when downloading a file
* prevent user from downloading a document when connectivity is lost
  * display a modal inviting user to reregister when he has capabilities
  * display a modal with cancel only otherwise 

### Related issues

* #12193

### Checklist


marieflorescontact (approved):
locally tested ok 👌

PR #12216, opened 2025-09-01
[backend/frontend] : add connector health metrics and uptime display

maelv-filigran:
### Proposed changes

* Add health metrics tracking for managed connectors including restart count, start time, and reboot loop detection
* Implement GraphQL schema extensions for connector health metrics (new `ConnectorHealthMetrics` type)
* Add Redis storage layer for health metrics with 5-minute TTL
* Display connector uptime in the UI when connector is running
* Add comprehensive integration tests for health metrics functionality

Composer PR:  https://github.com/FiligranHQ/xtm-composer/pull/18

### Related issues
* #7328

### Checklist

### Further comments

This implementation adds health monitoring capabilities for managed connectors, allowing operators to track connector uptime and detect potential issues like reboot loops. The health metrics are stored in Redis with a TTL to ensure data freshness, and the UI displays the uptime in a human-readable format when a connector is running.

The solution includes:
- Backend API for updating and retrieving health metrics
- Frontend display of uptime using moment.js for duration formatting
- Integration tests covering health metrics updates 
- Mock XTM Composer extensions to support health metrics in tests


opencti-platform/opencti-graphql/config/schema/opencti.graphql
```diff
@@ -14794,7 +14812,7 @@ type Mutation {
   updateConnectorRequestedStatus(input: RequestConnectorStatusInput!): ManagedConnector @auth(for: [MODULES_MODMANAGE])
   updateConnectorCurrentStatus(input: CurrentConnectorStatusInput!): ManagedConnector @auth(for: [CONNECTORAPI])
   updateConnectorLogs(input: LogsConnectorStatusInput!): ID! @auth(for: [CONNECTORAPI])
-  updateConnectorTrigger(id: ID!, input: [EditInput]!): Connector @auth(for: [MODULES_MODMANAGE])
+  updateConnectorHealth(input: HealthConnectorStatusInput!): ID! @auth(for: [CONNECTORAPI])  updateConnectorTrigger(id: ID!, input: [EditInput]!): Connector @auth(for: [MODULES_MODMANAGE])
```
SouadHadjiat: looks like there is a missing line break between updateConnectorHealth and updateConnectorTrigger

opencti-platform/opencti-graphql/src/domain/connector.ts
```diff
@@ -350,6 +359,31 @@ export const connectorUpdateLogs = async (_context: AuthContext, _user: AuthUser
...
+// Health metrics input interface
+interface HealthConnectorStatusInput {
+  id: string;
+  restart_count: number;
+  started_at: string;
+  is_in_reboot_loop: boolean;
+}
+
+// Health metrics update function
+export const connectorUpdateHealth = async (_context: AuthContext, _user: AuthUser, input: HealthConnectorStatusInput) => {
```
SouadHadjiat, line 397: the input type should be imported from graphql

opencti-platform/opencti-front/src/private/components/data/connectors/Connector.tsx
```diff
@@ -254,6 +255,51 @@ const ConnectorComponent: FunctionComponent<ConnectorComponentProps> = ({ connec
...
 
+  // Calculate uptime from started_at using moment duration
+  const calculateUptime = (startedAt: string | null | undefined): string => {
+    if (!startedAt) return t_i18n('Not available');
+
+    // Handle potential Unix timestamp in seconds (if it's a number string)
+    let start;
+    if (!Number.isNaN(Number(startedAt)) && Number(startedAt) < 10000000000) {
+      // If it's a Unix timestamp in seconds (less than 10 billion), convert to milliseconds
+      start = moment(Number(startedAt) * 1000);
```
esrevi: I think we should not use moment since this lib is deprecated, it could be replaced with date-fns that is already present in the codebase ^^

opencti-platform/opencti-graphql/src/domain/connector.ts
maelv-filigran: Thanks for spotting this one: remaining code from debug...

opencti-platform/opencti-front/src/private/components/data/connectors/Connector.tsx
maelv-filigran: nice catch!

opencti-platform/opencti-graphql/src/domain/connector.ts
```diff
@@ -350,6 +360,23 @@ export const connectorUpdateLogs = async (_context: AuthContext, _user: AuthUser
...
+    started_at: input.started_at,
+    is_in_reboot_loop: input.is_in_reboot_loop,
+    last_update: new Date().toISOString()
+  };
+  await redisSetConnectorHealthMetrics(input.id, metrics);
+  return input.id;
+};
+
+// Get health metrics function
+export const connectorGetHealth = async (_context: AuthContext, _user: AuthUser, connectorId: string): Promise<ConnectorHealthMetrics | null> => {
```
JeremyCloarec, line 409: This method is currently unused, but should be used in connector resolver to keep the architecture of resolver/domain cleaner

opencti-platform/opencti-graphql/src/resolvers/connector.js
```diff
@@ -90,13 +91,15 @@ const connectorResolvers = {
     connector_queue_details: (cn) => queueDetails(cn.id),
     connector_user: (cn, _, context) => connectorUser(context, context.user, cn.connector_user_id),
     manager_connector_logs: (cn) => redisGetConnectorLogs(cn.id),
+    manager_health_metrics: (cn) => redisGetConnectorHealthMetrics(cn.id),
```
JeremyCloarec: Can you use the connectorGetHealth method from the domain instead?

opencti-platform/opencti-graphql/src/resolvers/connector.js
```diff
@@ -90,13 +91,15 @@ const connectorResolvers = {
...
     manager_contract_hash: (cn, _, context) => computeManagerContractHash(context, context.user, cn),
     manager_contract_definition: (cn, _, context) => computeManagerConnectorContract(context, context.user, cn),
     manager_contract_configuration: (cn, _, context) => computeManagerConnectorConfiguration(context, context.user, cn, true),
     manager_contract_image: (cn) => computeManagerConnectorImage(cn),
   },
   ManagedConnector: {
     manager_connector_logs: (cn) => redisGetConnectorLogs(cn.id),
+    manager_health_metrics: (cn) => redisGetConnectorHealthMetrics(cn.id),
```
JeremyCloarec: Same as above

opencti-platform/opencti-front/src/private/components/data/connectors/Connector.tsx
```diff
@@ -254,6 +255,64 @@ const ConnectorComponent: FunctionComponent<ConnectorComponentProps> = ({ connec
...
+    if (minutes > 0) parts.push(`${minutes} ${t_i18n(minutes === 1 ? 'minute' : 'minutes')}`);
+
+    // If uptime is less than a minute, show seconds
+    if (parts.length === 0) {
+      parts.push(`${seconds} ${t_i18n(seconds === 1 ? 'second' : 'seconds')}`);
+    }
+
+    return parts.join(', ');
+  };
+
```
JeremyCloarec: I think this could be great to move the uptime compute to the backend: you could introduce a new `manager_connector_uptime` in the API and compute the uptime in the resolver of connectors. What do you think? And on the frontend side, you'll just need to keep the formatting to string part of the code

maelv-filigran:
### What's Changed

Added a new `manager_connector_uptime` field to provide connector uptime directly from the backend instead of calculating it on the frontend.

#### Key Changes:
- **Backend**: Added `connectorGetUptime` function to calculate uptime in seconds from the connector's `started_at` timestamp
- **GraphQL Schema**: Added `manager_connector_uptime` field (Float) to both `Connector` and `ManagedConnector` types
- **Resolvers**: Integrated the new uptime calculation in connector resolvers
- **Frontend**: Refactored uptime display to use the pre-calculated backend value instead of client-side date calculations

opencti-platform/opencti-front/src/private/components/data/connectors/Connector.tsx
```diff
@@ -254,6 +254,30 @@ const ConnectorComponent: FunctionComponent<ConnectorComponentProps> = ({ connec
...
```
JeremyCloarec: You can use the `formatSeconds` method in Time.js to fromat from a timestamp to a string. This won't give you the same formating, but I think it'd be better to align the display on all views

opencti-platform/opencti-front/src/private/components/data/connectors/Connector.tsx
```diff
@@ -641,6 +665,16 @@ const ConnectorComponent: FunctionComponent<ConnectorComponentProps> = ({ connec
                   )
               }
               </Grid>
+              {connector.is_managed && connector.manager_current_status === 'started' && connector.manager_connector_uptime !== null && connector.manager_connector_uptime !== undefined && (
```
JeremyCloarec: Was there a reason to check for null & undefined separately instead of just checking `!!connector.manager_connector_uptime` ?

opencti-platform/opencti-graphql/config/schema/opencti.graphql
```diff
@@ -2113,7 +2121,8 @@ type Connector implements BasicObject & InternalObject {
   manager_contract_configuration: [ManagerContractConfiguration!]
   manager_connector_logs: [String!]
   manager_contract_hash: String
-  ## Built in Connector
+  manager_health_metrics: ConnectorHealthMetrics
+  manager_connector_uptime: Float
```
JeremyCloarec: Is it a Float? Can't it be typed as Int?

maelv-filigran: The Math.floor() ensures we're always getting an integer value (number of seconds). Therefore, the GraphQL schema should indeed use Int instead of Float. This field represents uptime in seconds, which is always a whole number. You're right !

opencti-platform/opencti-front/src/private/components/data/connectors/Connector.tsx
maelv-filigran: I'll update this

opencti-platform/opencti-front/src/private/components/data/connectors/Connector.tsx
maelv-filigran: formatSeconds returns a format like "01:23:45" (HH:MM:SS format)
formatUptime returns a human-readable format like "2 days, 3 hours, 15 minutes"

Since connectors are intended to run for several days, I believe the current implementation improves readability. However, these functions could also be added to the utility library Time.js. What do you think?

JeremyCloarec: You're right, I didn't dig into the differences between formatSeconds & formatUptime format, but formatUptime format should be kept for readability.
But as you said, I think it's a good idea to add it in Time.js to make it available elsewhere also 👍

JeremyCloarec (approved):

PR #12222, opened 2025-09-01
[backend/frontend] Add more granularity to "Manage taxonomies" permission (#8102)

### LINKED ISSUE #8102: Add more granularity to "Manage taxonomies" permission
## Use case

By default, we limit the creation of new labels to avoid duplications (command-and-control, c2, c2c, c-and-c, c&c, etc.). However, we have what we call intelligence curators who are expert analysts that work within the organization that may require the creation of certain labels under very concrete circumstances. These labels may be brand-new labels and may need to be created.

imagen

We have added a permission to these high-level profiles giving access to "Manage Taxonomies" permission.

However, we still are not comfortable with them changing or deleting by mistake other aspects which are not expected to change that often, specifically, phases, vocabularies and templates.

imagen

## Current Workaround

No workaround. Simply limiting the management of the Taxonomies to a very limited amount of people and adding a procedure to request new creation. 

## Proposed Solution

Adding a sub-sub-permission to allow or deny the management of the subsections under taxonomies as happens with other permissions which also have "sublevels":

imagen

## Additional Information

None.

No.

richard-julien:
See #8102


SarahBocognano:
Translations of roles are missing in backend
image

SarahBocognano:
I have trouble accessing "case templates", "status template" & "kill chain phases" alone.
Let me explain : 

If I grant only case template capability in taxonomies for exemple I'm redirect in settings and not in the case template page.
Same for status template & kill chain phases altought the path are correct in the code.
But it works perfectly when I select only labels or vocabs capability alone.

I don't know if it's the expected behavior ?

richard-julien:
> I have trouble accessing "case templates", "status template" & "kill chain phases" alone. Let me explain :
> 
> If I grant only case template capability in taxonomies for exemple I'm redirect in settings and not in the case template page. Same for status template & kill chain phases altought the path are correct in the code. But it works perfectly when I select only labels or vocabs capability alone.
> 
> I don't know if it's the expected behavior ?

Thanks for the feedbacks @SarahBocognano .
Not normal. Both of your comment are now fixed
As I rename a bit one capability description, you need to reset completely the testing platform if already migrated.

SarahBocognano (approved):

Issue #11284, opened 2025-06-11
[BUG] Edge Condition Bug - PayloadTooLargeError error for users with a 'very large' representation of the user's profile


ParamConstructor:
## Description

Using a SAML/SSO configuration, the returned representation of the user's profile/group/etc was `too large` and we were receiving:

```
at next (/opt/opencti/build/node_modules/express/lib/router/route.js:145:7)"
       },
       {
           "message": "request entity too large",
           "name": "PayloadTooLargeError",
           "stack": "PayloadTooLargeError: request entity too large
   at readStream (/opt/opencti/build/node_modules/raw-body/index.js:163:17)
``` 

We attempted to use the APP__MAX_PAYLOAD_BODY_SIZE setting, but this didn't seem to correct the issue.

By default the limit for bodyParser is 100kb. We resolved it by adding a `limit: '5mb'` (our edge case resolution) to https://github.com/OpenCTI-Platform/opencti/blob/master/opencti-platform/opencti-graphql/src/http/httpPlatform.js#L417-L418

![Image](https://github.com/user-attachments/assets/e7062be3-7123-42e8-8f46-f2ae4a9a1603)

## Environment

1. OS (where OpenCTI server runs): Linux
2. OpenCTI version:  OpenCTI 6.6.15
3. OpenCTI client: Frontend
4. Other environment details:

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Attempt to login with a user, via SAML, that has a `very large` profile returned (greater than 100kb)
2. Receive error - `"stack": "PayloadTooLargeError: request entity too large`

## Expected Output

Login user via SAML

## Actual Output

Failed login - `"stack": "PayloadTooLargeError: request entity too large`

## Additional information

This is more of an "edge condition" - so just mentioning it as a bug, versus submitting a PR with a configurable variable to control this payload size handling. This is for tracking should Filigran determine a path and add a configurable variable, or just implement a 5MB default limit.

PR #12248, opened 2025-09-03
[backend] Set a configurable SSO/SAML payload body size to handle enterprises that have large envelopes of data (#11284)

ParamConstructor:
By default the Express body-parse payload limit is set to '100kb' (https://expressjs.com/en/resources/middleware/body-parser.html). This limit will cause the issue (PayloadTooLargeError) noted in (https://github.com/OpenCTI-Platform/opencti/issues/11284) if your Enterprise has a large amount of data in the SSO/SAML payload body envelope.

This PR seeks to solve this issue in a configurable manner. The default is set to the current value of `'100kb'`, but if you choose to - you can now use the `auth_payload_body_size` setting to increase the payload size to your required need for your Enterprise.

### Proposed changes

* Add configurable 'auth_payload_body_size' setting for the APP.
* Solves (PayloadTooLargeError) for Enterpises with payloads that exceed the `'100kb'` default limit.

### Related issues
* (https://github.com/OpenCTI-Platform/opencti/issues/11284) 

### Checklist

- [X] I added/update the relevant documentation (either on github or on notion)

### Further comments

None

### DIFF
--- a/opencti-platform/opencti-graphql/src/http/httpPlatform.js
+++ b/opencti-platform/opencti-graphql/src/http/httpPlatform.js
@@ -13,7 +13,7 @@ import validator from 'validator';
 import archiverZipEncrypted from 'archiver-zip-encrypted';
 import rateLimit from 'express-rate-limit';
 import contentDisposition from 'content-disposition';
-import { basePath, booleanConf, DEV_MODE, ENABLED_UI, logApp, OPENCTI_SESSION } from '../config/conf';
+import { basePath, booleanConf, DEV_MODE, ENABLED_UI, logApp, OPENCTI_SESSION, AUTH_PAYLOAD_BODY_SIZE } from '../config/conf';
 import passport, { isStrategyActivated, STRATEGY_CERT } from '../config/providers';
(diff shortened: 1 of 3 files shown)


opencti-platform/opencti-graphql/src/http/httpPlatform.js
```diff
@@ -415,7 +415,8 @@ const createApp = async (app) => {
   });
 
   // -- Passport callback
-  const urlencodedParser = bodyParser.urlencoded({ extended: true });
+  // -- Default limit is '100kb' based on https://expressjs.com/en/resources/middleware/body-parser.html
+  const urlencodedParser = bodyParser.urlencoded({ extended: true, limit: AUTH_PAYLOAD_BODY_SIZE });
```
JeremyCloarec: Could you change it to not set the limit parameter if no auth_payload_body_size is configured? By not setting a default value for AUTH_PAYLOAD_BODY_SIZE, and checking if AUTH_PAYLOAD_BODY_SIZE is defined here before setting the limit parameter.
This way, even if the default limit is changed on the side of express, the default value of '100kb' won't have to be updated on our side since the library will handle the default value

ParamConstructor:
@JeremyCloarec - Rebased to master and made requested changes.

JeremyCloarec (approved):

Issue #12245, opened 2025-09-03
Can't get list of External References in Draft mode


leitosama:
## Description

If you created external reference while you in Draft mode you wouldn't be able to get a list of external references.

## Environment

1. OS (where OpenCTI server runs): Fedora (Docker)
2. OpenCTI version: 6.7.16
3. OpenCTI client: frontend
4. Other environment details: -

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Create Draft and open
3. (as example) Create Report Entity with new external reference (using "+" button in "Create an entity" window) 
4. Try to link other external references to the new entity 
## Expected Output

Get a list of existed entities to link.

## Actual Output

Frontend error.
```
opencti-1  | {"category":"APP","cause":{"attributes":{},"message":"Cannot return null for non-nullable field ExternalReference.source_name.","name":"GraphQLError","stack":"Error: Cannot return null for non-nullable field ExternalReference.source_name.\n    at completeValue (/opt/opencti/build/back.js:181:11234)\n    at executeField (/opt/opencti/build/back.js:181:10023)\n    at executeFields (/opt/opencti/build/back.js:181:8900)\n    at completeObjectValue (/opt/opencti/build/back.js:181:16364)\n    at completeValue (/opt/opencti/build/back.js:181:11760)\n    at completeValue (/opt/opencti/build/back.js:181:11135)\n    at executeField (/opt/opencti/build/back.js:181:10023)\n    at executeFields (/opt/opencti/build/back.js:181:8900)\n    at completeObjectValue (/opt/opencti/build/back.js:181:16364)\n    at completeValue (/opt/opencti/build/back.js:181:11760)\n    at completeValue (/opt/opencti/build/back.js:181:11135)\n    at /opt/opencti/build/back.js:181:12541\n    at Function.from (<anonymous>)\n    at completeListValue (/opt/opencti/build/back.js:181:12284)\n    at completeValue (/opt/opencti/build/back.js:181:11426)\n    at completeValue (/opt/opencti/build/back.js:181:11135)"},"level":"error","message":"Cannot return null for non-nullable field ExternalReference.source_name.","operation":"AddExternalReferencesLinesQuery","size":80,"source":"backend","time":12,"timestamp":"2025-09-03T10:41:56.297Z","type":"READ_ERROR","user_id":"88ec0c6a-13ce-5e39-b486-354fe4a7084f","version":"6.7.16"}
```

## Additional information

[...]

PR #12250, opened 2025-09-03
[backend] add filter on base_type for external references (#12245)

JeremyCloarec:
### Proposed changes

Both external reference entity and external reference ref have the same entity_type name: "External-Reference" for the entity and "external-reference" for the ref. 
This can particularly cause an issue when in a draft: queryting the entity_type "External-Reference" will return both the entities and the refs, because they are all stored in the same index. This usually doesn't cause an issue outside of drafts, because we only target the READ_INDEX_STIX_META_OBJECTS index when looking up external references.

The fix here is to also add a filter on base_type when querying external references, this way we don't retrieve the external reference refs when loading in the draft index

### Related issues

* #12245
*

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/domain/externalReference.js
+++ b/opencti-platform/opencti-graphql/src/domain/externalReference.js
@@ -17,7 +17,8 @@ export const findById = (context, user, externalReferenceId) => {
 };
 
 export const findAll = (context, user, args) => {
-  return listEntities(context, user, [ENTITY_TYPE_EXTERNAL_REFERENCE], args);
+  const filters = addFilter(args.filters, 'base_type', 'ENTITY');
+  return listEntities(context, user, [ENTITY_TYPE_EXTERNAL_REFERENCE], { ...args, filters });
 };
 
 export const references = async (context, user, externalReferenceId, args) => {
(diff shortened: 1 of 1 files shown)


opencti-platform/opencti-graphql/src/domain/externalReference.js
```diff
@@ -17,7 +17,8 @@ export const findById = (context, user, externalReferenceId) => {
 };
 
 export const findAll = (context, user, args) => {
-  return listEntities(context, user, [ENTITY_TYPE_EXTERNAL_REFERENCE], args);
+  const filters = addFilter(args.filters, 'base_type', 'ENTITY');
```
Archidoit:
```suggestion
  const filters = addFilter(args.filters, 'base_type', BASE_TYPE_ENTITY);
```

Archidoit: you can use the constant here :)

SouadHadjiat (commented):
ok for the fix, try to fix task manager too ?

Archidoit (approved):

JeremyCloarec:
> ok for the fix, try to fix task manager too ?

I'll create an issue to track it, because I can't find an easy way to fix it :/ And I think it'll be better to already merge this fix for now

PR #12251, opened 2025-09-03
[backend] add more logs to debug rule engine stalling

JeremyCloarec:
### Proposed changes

* add more logs to debug rule engine stalling
*

### Related issues

*
*

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/manager/ruleManager.ts
+++ b/opencti-platform/opencti-graphql/src/manager/ruleManager.ts
@@ -167,7 +167,7 @@ export const rulesApplyHandler = async (context: AuthContext, user: AuthUser, ev
   for (let index = 0; index < events.length; index += 1) {
     const event = events[index];
     const { type, data } = event;
-    logApp.info('[RULE_MANAGER_DEBUG] Handling event in ruleManager', { event });
+    logApp.info('[RULE_MANAGER_DEBUG] Handling event in ruleManager', { eventType: event.type, eventMessage: event.message });
     try {
       // In case of merge convert the events to basic events and restart the process
       if (type === EVENT_TYPE_MERGE) {
@@ -200,7 +200,7 @@ export const rulesApplyHandler = async (context: AuthContext, user: AuthUser, ev
           }
           // Rule match, need to apply
           if (isCurrentMatched) {
-            logApp.info('[RULE_MANAGER_DEBUG] Applying rule update on rule', { rule });
+            logApp.info('[RULE_MANAGER_DEBUG] Applying rule update on rule', { ruleId: rule.id, ruleName: rule.name });
             await rule.update(data, updateEvent);
           }
         }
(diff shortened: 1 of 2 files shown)


opencti-platform/opencti-graphql/src/rules/containerWithRefsBuilder.ts
```diff
@@ -140,6 +143,7 @@ const buildContainerRefsRule = (ruleDefinition: RuleDefinition, containerType: s
       };
       const listArgs = isSource ? { fromId: originIds, toTypes: [relationTypes.rightType] } : { toId: originIds, fromTypes: [relationTypes.leftType] };
       const fullListArgs = { ...listArgs, callback: listAddedRefsCallback };
+      logApp.info('[RULE_MANAGER_DEBUG] Listing all rels', { fullListArgs });
```
SouadHadjiat: do we need the fullListArgs here with the callback ?

opencti-platform/opencti-graphql/src/rules/containerWithRefsBuilder.ts
```diff
@@ -159,6 +163,7 @@ const buildContainerRefsRule = (ruleDefinition: RuleDefinition, containerType: s
         }
       };
       const args = { fromId: report.extensions[STIX_EXT_OCTI].id, filters, noFiltersChecking: true, indices: READ_DATA_INDICES, callback: listRemovedRefsCallback };
+      logApp.info('[RULE_MANAGER_DEBUG] Listing all remove rels', { args });
```
SouadHadjiat: same comment about the args

opencti-platform/opencti-graphql/src/rules/containerWithRefsBuilder.ts
JeremyCloarec: probably not you're right

opencti-platform/opencti-graphql/src/manager/ruleManager.ts
```diff
@@ -199,6 +200,7 @@ export const rulesApplyHandler = async (context: AuthContext, user: AuthUser, ev
           }
           // Rule match, need to apply
           if (isCurrentMatched) {
+            logApp.info('[RULE_MANAGER_DEBUG] Applying rule update on rule', { rule });
```
SouadHadjiat: I also wonder about rule, it looks like it could be a big object, why not log only the rule name and id

SouadHadjiat (approved):

maelv-filigran:
All debug logs are added with logApp.info() instead of logApp.debug().

In production, these logs will generate a massive volume of data.
Risk of disk saturation and performance degradation.
Difficulty identifying real problems in the noise of the logs.
Have you considered the impact of logging every event at the INFO level in production, where thousands of events can be processed per minute?

Logs are always executed even if the debug level is not enabled.

Perhaps something like:
```typescript
// Add a check for the logging level
if (logApp.isDebugEnabled()) {
  logApp.debug(‘[RULE_MANAGER_DEBUG] Processing event in ruleManager’, { 
    eventType: event.type,
    dataId: event.data?.id 
  });
}
```

maelv-filigran:
Logging the entire event object may expose sensitive data.

logApp.info(‘[RULE_MANAGER_DEBUG] Handling event in ruleManager’, { event });

Consider filter logged data to include only essential metadata:

```
logApp.debug(‘[RULE_MANAGER_DEBUG] Handling event in ruleManager’, { 
  eventType: event.type,
  dataId: event.data?.id,
  dataType: event.data?.type
});
```

PR #12258, opened 2025-09-03
[backend] Expose graphql schema through express route (#12240)

### LINKED ISSUE #12240: Expose graphql schema through express route
## Use case

For remote system it really useful to be able to easily know the schema definition.
As introspection is disable by default for security/performance, this endpoint will only expose the raw graphql definition

richard-julien:
See #12240

### DIFF
--- a/opencti-platform/opencti-graphql/src/http/httpPlatform.js
+++ b/opencti-platform/opencti-graphql/src/http/httpPlatform.js
@@ -13,7 +13,8 @@ import validator from 'validator';
 import archiverZipEncrypted from 'archiver-zip-encrypted';
 import rateLimit from 'express-rate-limit';
 import contentDisposition from 'content-disposition';
-import { basePath, DEV_MODE, ENABLED_UI, logApp, OPENCTI_SESSION } from '../config/conf';
+import { printSchema } from 'graphql/utilities';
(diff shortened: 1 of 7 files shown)


opencti-platform/opencti-graphql/src/http/httpPlatform.js
```diff
@@ -179,6 +180,18 @@ const createApp = async (app) => {
...
   archiver.registerFormat('zip-encrypted', archiverZipEncrypted);
 
+  // -- API schema
+  app.get(`${basePath}/schema`, async (req, res) => {
+    const context = await createAuthenticatedContext(req, res, 'schema_get');
+    if (!context.user) {
+      res.sendStatus(403);
+      return;
+    }
+    res.set('Cache-Control', 'public, max-age=12000'); // 1 hour cache
```
xfournet: 12000 is 3h20 not 1h

opencti-platform/opencti-graphql/src/http/httpPlatform.js
```diff
@@ -179,6 +180,18 @@ const createApp = async (app) => {
...
+    res.set('Vary', 'X-OPENCTI-SCHEMA-VARY-CACHE'); // Way for client to invalidate cache
```
xfournet, line 191: there is no corresponding header, doesn't seems to have any effect

xfournet (approved):

PR #12276, opened 2025-09-04
[backend] Fix Redis connections leaks when using SSE live stream in recover mode

### LINKED ISSUE #12273: Redis connection can leak when using SSE live stream in recover mode
## Description

Scenario of the leak:
* a HTTP client connect to /stream/live with recover parameter
* the client disconnect when receiving the recover info and before the broadcasting starts

In this situation a Redis connection is created and will be kept idle until the platform is stopped.

## Environment

1. OS (where OpenCTI server runs): Linux
2. OpenCTI version: 8.7.17

## Reproducible Steps

See the description

## Expected Output

No Redis connection leak

## Actual Output

Redis connection leak

xfournet:
### Proposed changes

* Ensure the Redis client of the stream processor is closed when the processing loop is terminated
* Take the opportunity to fix the info() method that was not using the request stream name

### Related issues
* #12273 

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/database/redis.ts
+++ b/opencti-platform/opencti-graphql/src/database/redis.ts
@@ -780,14 +780,23 @@ export const createStreamProcessor = <T extends BaseEvent> (
     info: async () => processInfo(),
     running: () => streamListening,
     start: async (start = 'live') => {
-      let fromStart = start;
-      if (isEmptyField(fromStart)) {
-        fromStart = 'live';
+      if (streamListening) {
+        let fromStart = start;
+        if (isEmptyField(fromStart)) {
+          fromStart = 'live';
+        }
+        startEventId = fromStart === 'live' ? '$' : fromStart;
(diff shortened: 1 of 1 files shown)


SouadHadjiat:
how can we reproduce exactly this issue ? I tried opening a live stream, with and without recover mode, I tried also to open a data sharing stream, everytime my connection is closed if I close the page.

xfournet:
> how can we reproduce exactly this issue ? I tried opening a live stream, with and without recover mode, I tried also to open a data sharing stream, everytime my connection is closed if I close the page.

the problem happen only if the connection is closed before the broadcasting starts. Recovery data should be big enough and the connection close happen quickly enough to have a chance to reproduce the problem. Using a browser with a manual close may be challenging to reproduce the issue. The observed case was a python program that was doing the call, so it may have been quick enough to close the connection just after having initialized it.

richard-julien (approved):

PR #12279, opened 2025-09-05
[backend] automatically stale PR after 3 months of inactivity (#12277)

### LINKED ISSUE #12277: Automatically stale PR with no activity
## Description

Create stale PR github action with message to inform the author that the PR will be stale because of inactivity

## Environment

4. Other environment details:

## Reproducible Steps

1. { e.g. Run ... }
2. { e.g. Click ... }
3. { e.g. Error ... }

marieflorescontact:
### Proposed changes

* add github actions to automaticaly stale PR after 3 months of inactivity

### Related issues

* closes: https://github.com/OpenCTI-Platform/opencti/issues/12277

### Checklist

### DIFF
--- a/.github/workflows/auto-stale-pr.yml
+++ b/.github/workflows/auto-stale-pr.yml
@@ -1,0 +1,28 @@
+name: Mark stale pull requests
+
+on: pull_request
+
+jobs:
+  stale:
+
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write
+
+    steps:
+      - uses: actions/stale@v5
+        with:
+          days-before-stale: 90
+          days-before-close: 30
+          stale-pr-message: |
+            Hello, this pull request has been inactive for 3 months.
+            It has been marked as stale with the "no-pr-activity" label and will be automatically closed in 1 month if no activity is detected.
+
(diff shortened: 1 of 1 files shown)


.github/workflows/auto-stale-pr.yml
```diff
@@ -0,0 +1,28 @@
+name: Mark stale pull requests
+
+on: pull_request
```
aHenryJard, line 3: I am not sure on trigger, because it will be trigger on PR opening but then it's never stale.

On exemples I see that it's always a cron:
https://github.com/actions/stale?tab=readme-ov-file#usage

https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#schedule

```suggestion
on:
  schedule:
    - cron: '30 1 * * *'
```

aHenryJard (commented):
Thanks you for this proposal !

This action is going to stale also issues so we need to see with product.

Maybe we could start with  debug-only "[debug-only](https://github.com/actions/stale?tab=readme-ov-file#debug-only)" 

To see what would change (but without actual changes yet)

marieflorescontact:
we don't want anymore to stall PR with github action

PR #12286, opened 2025-09-05
[backend] Fix add/remove pir denormalization on entities (#11149)

Archidoit:
### Proposed changes
- Fix add/remove pir denorm on stix domain objects when a creation/deletion of relationship causes a flag in several PIRs


opencti-platform/opencti-graphql/src/database/engine.js
```diff
@@ -4622,27 +4620,36 @@ export const elIndexElements = async (context, user, indexingType, elements) =>
...
-                last_pir_score_date: params.updated_at,
-              }))
-          ];
-          script += '; ctx._source[\'pir_information\'] = params.pir_information';
+          // remove pir_information concerning the pir and add the new pir_information
+          script += `
+          ; if (ctx._source.containsKey('pir_information') && ctx._source['pir_information'] != null) {
+              ctx._source['pir_information'].removeIf(item -> params.pir_ids.contains(item.pir_id));
+              ctx._source['pir_information'].addAll(params.new_pir_information);
+            } else ctx._source['pir_information'] = params.new_pir_information
```
richard-julien, line 4763: Please put {} for the else too

Archidoit: was done on purpose because a ';' is added via the join ... and }; is incorrect

richard-julien (approved):

PR #12292, opened 2025-09-05
[frontend] Fix connector stopped status color chip on create connector (#7328)

esrevi:
### Proposed changes

* Fix on connector created, the status chip color is now red when it is stopped

### Related issues

* #7328 

### Checklist


opencti-platform/opencti-front/src/private/components/data/Root.jsx
```diff
@@ -145,7 +145,7 @@ const Root = () => {
           element={boundaryWrapper(Connectors)}
         />
         <Route
-          path="/ingestion/connectors/:connectorId"
+          path="/ingestion/connectors/:connectorId/*"
```
esrevi, line 148: there was a warning from react router telling to add this trailing star when we navigate in the details of a connector

opencti-platform/opencti-front/src/utils/Connector.tsx
```diff
@@ -66,47 +63,53 @@ export const useGetConnectorAvailableFilterKeys = (connector: Connector): string
...
-  const { t_i18n } = useFormatter();
-  // eslint-disable-next-line react/display-name
-  return ({
-    manager_current_status,
-    manager_requested_status,
-    active,
-  }: Partial<ConnectorsStatus_data$data['connectors'][0]>) => {
-    if (manager_requested_status) {
-      if ((manager_current_status ?? '').slice(0, 5) !== manager_requested_status.slice(0, 5)) {
+export const computeConnectorStatus = ({
```
esrevi, line 66: this function is refactored to return only an object and not also a react component for better separation of concerns.

opencti-platform/opencti-front/src/private/components/data/connectors/ConnectorsStatus.tsx
```diff
@@ -427,32 +426,36 @@ const ConnectorsStatusComponent: FunctionComponent<ConnectorsStatusComponentProp
...
-                                  setConnectorMessages(connector.messages);
-                                }}
-                                aria-haspopup="true"
-                                color="primary"
-                                size="large"
-                                disabled={!!connector.built_in}
-                              >
-                                <PlaylistRemoveOutlined />
-                              </IconButton>
+                              <span>
```
esrevi, line 429: add span to display tooltip even if button is disabled

opencti-platform/opencti-front/src/utils/Connector.tsx
```diff
@@ -66,47 +63,53 @@ export const useGetConnectorAvailableFilterKeys = (connector: Connector): string
...
+          label: manager_requested_status,
+        };
+      }
+
+      if (manager_requested_status === 'stopped') {
         return {
-          processing: ['starting', 'stopping'].includes(manager_requested_status),
-          render: (
-            <ItemBoolean
-              status={['starting', 'stopping'].includes(manager_requested_status) ? undefined : true}
```
esrevi, line 83: the issue comes from this:
when the connector is created, here are the values we got:
{
managerRequestedStatus: "stopped", managerCurrentStatus: null, active: true
}

When managerRequestedStatus is "stopped”, this line ['starting', 'stopping'].includes(manager_requested_status) ? undefined : true, will return true, so status is true -> green

opencti-platform/opencti-front/src/private/components/data/connectors/ConnectorStatusChip.tsx
```diff
@@ -0,0 +1,34 @@
...
+  } else if (status === 'active') {
+    itemBooleanStatus = true;
+  } else {
+    itemBooleanStatus = false;
+  }
+
+  return (
+    <ItemBoolean
+      status={itemBooleanStatus}
+      label={t_i18n(label)}
```
lndrtrbn: We prefer avoid those kind of dynamic translations because our scripts to automatically detect translations cannot work on this. Instead keep the translations `t_i18n('Active')` and `t_i18n('Inactive')` in the helper function instead of just strings

JeremyCloarec (approved):

PR #11885, opened 2025-08-03
[backend/frontend] Fix entities visibility using regarding Of filter (#11496)

richard-julien:
See #11496

PR #12325, opened 2025-09-08
[backend] Improve global count / hasNextPage management (#11885)

richard-julien:
See #11885
Global count and hasNext page improvement

### DIFF
--- a/opencti-platform/opencti-graphql/src/database/engine.js
+++ b/opencti-platform/opencti-graphql/src/database/engine.js
@@ -3889,7 +3889,12 @@ export const elPaginate = async (context, user, indexName, options = {}) => {
 };
 
 const elRepaginate = async (context, user, indexName, connectionFormat, opts = {}) => {
-  const { first = ES_DEFAULT_PAGINATION, maxSize = undefined, logForMigration = false, callback } = opts;
+  const {
+    first = ES_DEFAULT_PAGINATION,
+    maxSize = undefined,
+    logForMigration = false,
+    callback
+  } = opts;
   let batch = 0;
   let emitSize = 0;
   let globalHitsCount = 0;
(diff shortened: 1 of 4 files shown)


opencti-platform/opencti-graphql/src/database/engine.js
```diff
@@ -3868,6 +3867,7 @@ const elRepaginate = async (context, user, indexName, connectionFormat, opts = {
     const paginateOpts = { ...opts, first, after: searchAfter, connectionFormat: true, withResultMeta: true };
     const { elements: page, filterCount, total, endCursor } = await elPaginate(context, user, indexName, paginateOpts);
     totalFilteredCount += filterCount;
+    globalHitsCount = total - totalFilteredCount;
```
xfournet: does the ES total hit count take into account the number of elements that have skipped due to the search after ?

opencti-platform/opencti-graphql/src/database/engine.js
```diff
@@ -3847,17 +3847,16 @@ const elRepaginate = async (context, user, indexName, connectionFormat, opts = {
   } = opts;
```
xfournet, line 3897: some simplification 
* use pageInfo hasNextPage
* remove useless/redundant code branching
* fix bug when using callback `const missingNumber = maxSize - listing.length` was wrong

```
const elRepaginate = async (context, user, indexName, connectionFormat, opts = {}) => {
  const {
    first = ES_DEFAULT_PAGINATION,
    maxSize = undefined,
    logForMigration = false,
    callback
  } = opts;
  let batch = 0;
  let emitSize = 0;
  let globalHitsCount = 0;
  let totalFilteredCount = 0;
  let hasNextPage = true;
  let continueProcess = true;
  let searchAfter = opts.after;
  const listing = [];
  while (continueProcess && emitSize < maxSize && hasNextPage) {
    // Force options to get connection format and manage search after and metadata
    const paginateOpts = { ...opts, first, after: searchAfter, connectionFormat: true, withResultMeta: true };
    const { elements: page, filterCount, total, endCursor } = await elPaginate(context, user, indexName, paginateOpts);

    // when first === maxSize only one iteration is necessary except in case of post filtering
    if (first === maxSize && batch > 10) {
      logApp.warn('[PERFORMANCE] Expensive post filtering detected', { batch, opts });
    }
    if (logForMigration) {
      logMigration.info('Migrating loading batch...', { batch });
    }

    if (page.edges.length > 0) {
      const edgeToPublish = page.edges.slice(0, maxSize - emitSize);
      const elements = connectionFormat ? edgeToPublish : await asyncMap(edgeToPublish, (edge) => edge.node);
      if (callback) {
        const callbackResult = await callback(elements, total);
        continueProcess = callbackResult === true || callbackResult === undefined;
      } else {
        listing.push(...elements);
      }
      emitSize += elements.length;
    }

    batch += 1;
    hasNextPage = page.pageInfo.hasNextPage;
    searchAfter = endCursor;
    totalFilteredCount += filterCount;
    globalHitsCount = total - totalFilteredCount;
  }
  return { elements: listing, totalCount: globalHitsCount };
};
```

opencti-platform/opencti-graphql/src/database/middleware-loader.ts
```diff
@@ -456,8 +456,8 @@ export const pageEntitiesConnection = async <T extends BasicStoreEntity>(context
   // maxSize MUST be aligned with first in this method.
   // As using elConnection is repaginate, removing maxSize will lead to major api breaking
   const paginateArgs = { ...buildEntityFilters(entityTypes, args), first, maxSize: first };
-  const { elements, totalCount, totalFilteredCount } = await elConnection(context, user, computedIndices, paginateArgs);
-  return buildPaginationFromEdges(args.first, args.after, elements, totalCount, totalFilteredCount);
+  const { elements, totalCount } = await elConnection(context, user, computedIndices, paginateArgs);
```
xfournet: in fact elConnection is not returning a 'connection' format. shouldn't we align elConnection ouput with other '*Connection' so here we would only do a return elConnection(...) ?

opencti-platform/opencti-graphql/src/database/engine.js
richard-julien: On my opinion, its the total count independant of the cursor

opencti-platform/opencti-graphql/src/database/engine.js
richard-julien: Thanks

opencti-platform/opencti-graphql/src/database/middleware-loader.ts
richard-julien: Correct. In fact I think its a refactor side effect has elConnection is directly use in another call an expect a connection format.

opencti-platform/opencti-graphql/src/database/engine.js
xfournet: ok in that case it's all good

opencti-platform/opencti-graphql/src/database/engine.js
xfournet: In the proposed version there was a problem when maxSize was undefined, it was not handled.

To be noted that in the previous version, there was a potential problem since searchAfter was updated only in case page.edges was not empty, that mean in case all the element of a page were post filtered, it would have been looping indefinitively. The proposed version fix this problem.

xfournet (approved):

Issue #12264, opened 2025-09-04
[user email template] Incorrect placeholder when creating a user in CE


romain-filigran:
## Description

In Community Edition, when creating a user, the "Email Template" placeholder is incorrect.

<img width="1499" height="781" alt="Image" src="https://github.com/user-attachments/assets/8e7925fe-d470-4061-8238-26d391d4435c" />

## Environment

OpenCTI version: 6.7.17

PR #12344, opened 2025-09-10
[frontend]  Incorrect email template placeholder when creating a user in CE (#12264)

Gwendoline-FAVRE-FELIX:
### Proposed changes

* Fix the label in CE

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/12264

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/common/form/EmailTemplateField.tsx
+++ b/opencti-platform/opencti-front/src/private/components/common/form/EmailTemplateField.tsx
@@ -120,7 +120,7 @@ const EmailTemplateField = ({ ...props }: EmailTemplateFieldProps) => {
             style={fieldSpacingContainerStyle}
             renderOption={() => null}
             textfieldprops={{
-              label: 'Enterprise Edition',
+              label: 'Email template',
             }}
           />
         </EETooltip>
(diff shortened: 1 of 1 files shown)


opencti-platform/opencti-front/src/private/components/common/form/EmailTemplateField.tsx
```diff
@@ -120,7 +120,7 @@ const EmailTemplateField = ({ ...props }: EmailTemplateFieldProps) => {
             style={fieldSpacingContainerStyle}
             renderOption={() => null}
             textfieldprops={{
-              label: 'Enterprise Edition',
+              label: 'Email template',
```
Archidoit: add t_i18n for translation

Archidoit (changes requested):

opencti-platform/opencti-front/src/private/components/common/form/EmailTemplateField.tsx
CelineSebe: it seems that is not necessary:
image

CelineSebe (approved):

Archidoit:
@CelineSebe yes it's necessary : 
image

your screenshot is under EE, the problem is under CE

Archidoit (approved):

PR #12354, opened 2025-09-10
[frontend] enable vertical scrolling for connector logs display (#7328)

maelv-filigran:
### Proposed changes

* Fix connector logs display to enable proper vertical scrolling
* Set a fixed height for the logs container using viewport calculations (`calc(100vh - 280px)`)
* Add explicit `overflowY: 'auto'` to ensure vertical scrolling is enabled for long log outputs

### Related issues

* #7328

### Checklist

### Further comments

This is a straightforward UI fix that improves the user experience when viewing connector logs. The solution uses viewport-relative height calculations to ensure the log container fits properly within the available space while maintaining scrollability.

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/data/connectors/Connector.tsx
+++ b/opencti-platform/opencti-front/src/private/components/data/connectors/Connector.tsx
@@ -649,11 +649,12 @@ const ConnectorComponent: FunctionComponent<ConnectorComponentProps> = ({ connec
 
   // Component for Logs content
   const ConnectorLogs = () => (
-    <Box sx={{ marginBottom: '20px' }}>
+    <Box sx={{ marginBottom: '20px', height: 'calc(100vh - 280px)' }}>
       <pre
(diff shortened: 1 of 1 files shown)


opencti-platform/opencti-front/src/private/components/data/connectors/Connector.tsx
```diff
@@ -649,11 +649,12 @@ const ConnectorComponent: FunctionComponent<ConnectorComponentProps> = ({ connec
 
   // Component for Logs content
   const ConnectorLogs = () => (
-    <Box sx={{ marginBottom: '20px' }}>
+    <Box sx={{ marginBottom: '20px', height: 'calc(100vh - 280px)' }}>
```
esrevi: where come from 280px value ? :) 
maybe we could set in a const with a name to explain it 

otherwise it's working, lgtm ! 👍

esrevi (approved):

PR #12355, opened 2025-09-10
[backend/frontend] manifest array type (#7328)

esrevi:
### Proposed changes

* Transform an array from the manager contract config to a string and send it to backend
* Update schema validator on backend to handle that case

### Related issues

* #7328 

### Checklist


xfournet (dismissed):
please ignore it, it's just for a test :)

opencti-platform/opencti-front/src/private/components/data/IngestionCatalog/IngestionCatalogConnectorCreation.tsx
```diff
@@ -120,7 +126,17 @@ const IngestionCatalogConnectorCreation = ({
   }: Partial<FormikHelpers<ManagedConnectorValues>>) => {
     const manager_contract_configuration = Object.entries(values)
       .filter(([, value]) => value != null)
-      .map(([key, value]) => ({ key, value: [value.toString()] }));
+      .map(([key, value]) => {
+        let computedValue = value;
+        if (Array.isArray(value)) {
+          computedValue = value.join(',');
```
richard-julien: Here you choose to force value to string if its an array.
So graphql API do not need to accept array
And you force all values to be string

opencti-platform/opencti-front/src/private/components/data/IngestionCatalog/utils/JsonFormArrayRenderer.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+  const { t_i18n } = useFormatter();
+
+  // Convert null to empty array for component state
+  const currentValues = data || [];
+  const [inputValue, setInputValue] = useState('');
+
+  const handleValuesChange = useCallback((event: React.SyntheticEvent, newValues: string[]) => {
+    const cleanValues = newValues
+      .filter((value) => value.trim() !== '')
```
richard-julien: I think we have isEmpty utility function

opencti-platform/opencti-front/src/private/components/data/connectors/ManagedConnectorEdition.tsx
```diff
@@ -75,11 +85,25 @@ const ManagedConnectorEdition = ({
     setSubmitting,
     resetForm,
   }: Partial<FormikHelpers<ManagedConnectorValues>>) => {
+    const manager_contract_configuration = Object.entries(values)
+      .filter(([, value]) => value != null)
+      .map(([key, value]) => {
+        let computedValue = value;
+        if (Array.isArray(value)) {
+          computedValue = value.join(',');
```
richard-julien: Exactly same code than creation? Could be generified?

opencti-platform/opencti-front/src/private/components/data/connectors/ManagedConnectorEdition.tsx
```diff
@@ -101,38 +125,63 @@ const ManagedConnectorEdition = ({
...
+    const requiredProps: JsonSchema = {
+      properties: Object.fromEntries(requiredPropertiesArray),
+      required: contract.config_schema.required,
+    };
+
+    const optionalProps: JsonSchema = {
+      properties: Object.fromEntries(optionalPropertiesArray),
+    };
+
+    const reconciled = reconcileManagedConnectorContractDataWithSchema(
```
richard-julien, line 155: I dont think reconcile is an english word

opencti-platform/opencti-front/src/private/components/data/connectors/utils/reconcileManagedConnectorContractDataWithSchema.ts
```diff
@@ -0,0 +1,73 @@
...
+  managerContractProps.forEach(([key, schema]) => {
+    if (reconciledData[key] === undefined && schema.default !== undefined) {
+      reconciledData[key] = schema.default as ReconciledValue;
+    }
+  });
+
+  return reconciledData;
+};
+
+export default reconcileManagedConnectorContractDataWithSchema;
```
richard-julien, line 37: I dont understand why this file is needed. The form is already validating the content. I dont think this is needed at all. You also have the backend doing the validation

opencti-platform/opencti-graphql/src/modules/catalog/catalog-domain.ts
```diff
@@ -159,6 +158,72 @@ const encryptValue = (publicKey: string, value: string) => {
   return encrypted.toString('base64');
 };
 
+const validateContractConfigurations = (
+  contractConfigurations: any[],
```
richard-julien: contract configuration must be an array of key/value string no?

opencti-platform/opencti-graphql/src/modules/catalog/catalog-domain.ts
```diff
@@ -159,6 +158,72 @@ const encryptValue = (publicKey: string, value: string) => {
...
+  // Build validation object with parsed values
+  const contractObject: any = {};
+
+  contractConfigurations.forEach((config) => {
+    const propSchema = targetConfig.properties[config.key];
+
+    if (propSchema && config.value !== undefined && config.value !== null) {
+      // Skip validation for encrypted passwords
+      if (config.encrypted) {
+        contractObject[config.key] = 'encrypted_placeholder';
```
richard-julien: this function aims to validate the contract from value provided, this replacement doesnt make sens for me

opencti-platform/opencti-graphql/src/modules/catalog/catalog-domain.ts
```diff
@@ -159,6 +158,72 @@ const encryptValue = (publicKey: string, value: string) => {
...
+      } else {
+        // Get the actual string value (handle both array and string format)
+        const stringValue = Array.isArray(config.value) ? config.value[0] : config.value;
+        let parsedValue = stringValue;
+
+        // Parse value based on schema type
+        switch (propSchema.type) {
+          case 'array':
+            // Arrays are sent as comma-separated values from frontend
+            if (stringValue.trim() === '') {
```
richard-julien: We have utils to check empty string

opencti-platform/opencti-graphql/src/modules/catalog/catalog-domain.ts
```diff
@@ -159,6 +158,72 @@ const encryptValue = (publicKey: string, value: string) => {
...
+              // Split by comma and clean up each value
+              parsedValue = stringValue.split(',').map((v: string) => v.trim()).filter((v: string) => v !== '');
+            }
+            break;
+          case 'boolean':
+            parsedValue = stringValue === 'true' || stringValue === true;
+            break;
+          case 'integer':
+            parsedValue = parseInt(stringValue, 10);
+            if (Number.isNaN(parsedValue)) {
```
richard-julien: Doing this is not validation, is a fallback that will succeed the validation, so breaking the contract of the function

opencti-platform/opencti-graphql/src/modules/catalog/catalog-domain.ts
```diff
@@ -159,6 +158,72 @@ const encryptValue = (publicKey: string, value: string) => {
...
+            // Arrays are sent as comma-separated values from frontend
+            if (stringValue.trim() === '') {
+              parsedValue = [];
+            } else {
+              // Split by comma and clean up each value
+              parsedValue = stringValue.split(',').map((v: string) => v.trim()).filter((v: string) => v !== '');
+            }
+            break;
+          case 'boolean':
+            parsedValue = stringValue === 'true' || stringValue === true;
```
richard-julien: Doing this is not a validation. Putting an incorrect value will be validated as false entry

richard-julien (dismissed):

opencti-platform/opencti-front/src/private/components/data/IngestionCatalog/utils/JsonFormArrayRenderer.tsx
natalie-natsu: Yes it's from lodash itself. :-) https://lodash.com/docs/#isEmpty

opencti-platform/opencti-front/src/private/components/data/connectors/ManagedConnectorEdition.tsx
natalie-natsu: Tip: with Jetbrains you can configure spelling to be sure you're using english words without typos.

opencti-platform/opencti-front/src/private/components/data/connectors/utils/reconcileManagedConnectorContractDataWithSchema.ts
esrevi: The issue is the backend returns a string, but the schema from the manifest expects an array.

So for instance, we receive this data { connector_scope: "scope1,scope2,scope3" }, but as we use the shape defined in the manifest as a type array to build the form and to validate, the input field is expecting an array, this is why we need to transform that string back to an array when we want to update a connector.

opencti-platform/opencti-front/src/private/components/data/IngestionCatalog/IngestionCatalogConnectorCreation.tsx
esrevi: to add a connector, we call this mutation: 

```
// mutation
managedConnectorAdd(input: AddManagedConnectorInput): ManagedConnector
// input
input AddManagedConnectorInput {
 ....
  manager_contract_configuration: [ContractConfigInput!]!
}
// input type
input ContractConfigInput {
  key: String!
  value: [String!]
}
```
The schema requires value: [String!] when we add a connector. 
We join array values first to create a single string, then wrap it in an array to match the schema definition.

opencti-platform/opencti-front/src/private/components/data/connectors/ManagedConnectorEdition.tsx
esrevi: actually, reconcile exists, ~~but i can change it to reconciliate~~ :) , the jetbrains spelling does not raise warning

opencti-platform/opencti-front/src/private/components/data/IngestionCatalog/utils/JsonFormArrayRenderer.tsx
esrevi: i replaced with isNotEmptyField

opencti-platform/opencti-front/src/private/components/data/IngestionCatalog/IngestionCatalogConnectorCreation.tsx
esrevi: I think we could align the graphql ContractConfigInput, and not accept [string]! but a string instead
```
input ContractConfigInput {
  key: String!
  value: String!
}
```

opencti-platform/opencti-graphql/src/modules/catalog/catalog-domain.ts
maelv-filigran: I'll fix this

opencti-platform/opencti-graphql/src/modules/catalog/catalog-domain.ts
maelv-filigran: I'll use them, thanks!

opencti-platform/opencti-graphql/src/modules/catalog/catalog-domain.ts
maelv-filigran: you're absolutely right !

opencti-platform/opencti-graphql/src/modules/catalog/catalog-domain.ts
maelv-filigran: same

opencti-platform/opencti-front/src/private/components/data/connectors/utils/reconcileManagedConnectorContractDataWithSchema.ts
esrevi: but you are right, the other types are already handled by the schema, only the string to array is needed

opencti-platform/opencti-graphql/src/generated/graphql.ts
```diff
@@ -14562,8 +14562,8 @@ export type Mutation = {
   malwareAnalysisRelationAdd?: Maybe<StixRefRelationship>;
   malwareAnalysisRelationDelete?: Maybe<MalwareAnalysis>;
   malwareEdit?: Maybe<MalwareEditMutations>;
-  managedConnectorAdd?: Maybe<ManagedConnector>;
-  managedConnectorEdit?: Maybe<ManagedConnector>;
+  managedConnectorAdd?: Maybe<Connector>;
+  managedConnectorEdit?: Maybe<Connector>;
```
esrevi: change return to Connector, because on updating the connector, when we open the update dialog form just after, the api key is sent and because of the key, the form fills the input with the api key. 
To fix that, returning Connector prevents this issue because it doesn't return that key

opencti-platform/opencti-front/src/schema/relay.schema.graphql
```diff
@@ -9190,8 +9190,8 @@ type Mutation {
   roleEdit(id: ID!): RoleEditMutations
   pingConnector(id: ID!, state: String, connectorInfo: ConnectorInfoInput): Connector
   registerConnector(input: RegisterConnectorInput): Connector
-  managedConnectorEdit(input: EditManagedConnectorInput): ManagedConnector
-  managedConnectorAdd(input: AddManagedConnectorInput): ManagedConnector
+  managedConnectorEdit(input: EditManagedConnectorInput): Connector
```
richard-julien: Sorry but this doesnt make sens for me

opencti-platform/opencti-front/src/private/components/data/IngestionCatalog/IngestionCatalogConnectorCreation.tsx
```diff
@@ -205,7 +210,7 @@ const IngestionCatalogConnectorCreation = ({
 
     const reqProperties: JsonSchema = {
       properties: requiredProps,
-      required: connector.config_schema.required.filter((req) => req !== 'CONNECTOR_NAME' && req !== 'name'),
+      required: connector.config_schema.required.filter((req) => req !== 'CONNECTOR_NAME' && req !== 'name' && req !== 'CONNECTOR_ID'), // FIXME: remove filter on CONNECTOR_ID when manifest is ok
```
richard-julien: Iam a bit worried about this, CONNECTOR_NAME and CONNECTOR_ID must not be part of the contract

esrevi: you're right, we can remove that now, i had issues before now the manifest seems good

opencti-platform/opencti-front/src/schema/relay.schema.graphql
esrevi: the original issue is after creating/updating the connector, when a connector has an api key field, the mutation returns in ManagedConnector that data. 
It turns out that relay has issue on handling cache when two different entities such as Connector and ManagedConnector shares the same id, then, when we query the Connector, it was using from the cache the data with that api key and is displayed in the form.

To solve that issue, i refetch the data bypassing the cache when opening the update dialog form, and rollback the return types to ManagedConnector on managedConnectorEdit and managedConnectorAdd

richard-julien: Maybe ManagedConnector must be removed as it doest bring any value anymore

JeremyCloarec (approved):

opencti-platform/opencti-front/src/schema/relay.schema.graphql
esrevi: For now, we have let ManagedConnector and we can discuss about that later :)

PR #12287, opened 2025-09-05
[frontend] Add platform OpenCTI version when registering hub

jpkha:
### Proposed changes

* When registering OpenCTI instance, send platform version information
* Send platform version information with the checked connectivity with the hub to keep the version up to date

### Related issues

* https://github.com/FiligranHQ/xtm-hub/issues/919

### Checklist

- [x] I consider the submitted work as finished
- [ ] I tested the code for its functionality
- [ ] I wrote test cases for the relevant uses case (coverage and e2e)
- [ ] I added/update the relevant documentation (either on github or on notion)
- [x] Where necessary I refactored code to improve the overall quality

### Further comments
In order to test it you need to need to use a XTM Hub platform.
Test need to be done with the master and the current branch in order to test the retrocompatibility of different OpenCTI version with the Hub

Related PR : https://github.com/FiligranHQ/xtm-hub/pull/927

PR #12365, opened 2025-09-10
[backend] Simplify version get (#12287)

richard-julien:
Version must be get from package


jpkha (approved):
Ok 👍 locally test the different scenario.
Better solution, thank you 🥇

PR #12380, opened 2025-09-11
[backend] Rules activation launch background tasks even though rule already activated (#12379)

### LINKED ISSUE #12379: Rules activation launch background tasks even though rule already activated
## Description

Calling ruleSetActivation mutation doesn't take into account if a rule is already activated resulting to launching a background tasks.

## Environment

4. Other environment details:

## Reproducible Steps

1. { e.g. Run ... }
2. { e.g. Click ... }
3. { e.g. Error ... }

SarahBocognano:
### Proposed changes

* check if rule already activated 

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/12379

### DIFF
--- a/opencti-platform/opencti-graphql/src/domain/rules.ts
+++ b/opencti-platform/opencti-graphql/src/domain/rules.ts
@@ -22,7 +22,7 @@ import ReportRefsIndicatorBasedOnRule from '../rules/report-refs-indicator-based
 import ReportRefObservableBasedOnRule from '../rules/report-refs-observable-based-on/ReportRefObservableBasedOnRule';
 import ReportRefsLocationLocatedAtRule from '../rules/report-refs-location-located-at/ReportRefLocationLocatedAtRule';
 import ParentTechniqueUseRule from '../rules/parent-technique-use/ParentTechniqueUseRule';
-import {BUS_TOPICS, DEV_MODE, logApp} from '../config/conf';
+import { BUS_TOPICS, DEV_MODE, logApp } from '../config/conf';
 import type { AuthContext, AuthUser } from '../types/user';
 import { isEmptyField } from '../database/utils';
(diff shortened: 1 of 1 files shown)


opencti-platform/opencti-graphql/src/domain/rules.ts
```diff
@@ -73,16 +73,20 @@ export const getActivatedRules = async (context: AuthContext, user: AuthUser): P
...
   return rules.find((e) => e.id === id);
 };
 
-export const setRuleActivation = async (context: AuthContext, user: AuthUser, ruleId: string, active: boolean): Promise<RuleDefinition | undefined> => {
+export const setRuleActivation = async (context: AuthContext, user: AuthUser, ruleId: string, active: boolean): Promise<RuleRuntime | undefined> => {
   const resolvedRule = await getRule(context, user, ruleId);
   if (isEmptyField(resolvedRule)) {
     throw UnsupportedError(`Cant ${active ? 'enable' : 'disable'} undefined rule ${ruleId}`);
   }
+  if (resolvedRule?.activated === active) {
```
richard-julien, line 86: Thanks for this fix that will do the job.
I think maybe it could be a better solution to replace createEntity by a updateAttribute and react after the update only if something really change on the entity. wdyt?

SouadHadjiat (approved):

opencti-platform/opencti-graphql/src/domain/rules.ts
SouadHadjiat, line 86: no, if active is false, and resolvedRule.activated is false, it will return the rule without changing it, and if resolvedRule.activated is true, it will update the rule to set active to false. is there an issue with the condition ?

xfournet: indeed i misread the code, there is no issue

Issue #12372, opened 2025-09-11
No difference between Update and Update_linked chips in Draft


Archidoit:
## Description
In a draft, in entities list, the chip width is not large enough to see the difference between Update and Update_linked chips.

## Reproducible Steps
Go under a draft with both Update (update of an entity) and update_linked (update of a relationship involving the entity) entities.

<img width="973" height="328" alt="Image" src="https://github.com/user-attachments/assets/c7e71d37-3117-4002-9c16-056fbdc3bdab" />

There is not difference between the chips since we only see 'update_' for the 'update_linked' chip, even if we make the column size larger.

## Proposed solution
Increase a little bit the size of the chip or make a little more difference between the colors.

PR #12393, opened 2025-09-12
[frontend] Draft operation chips (#12372)

Archidoit:
### Proposed changes
Fix width and translation of operation chip in Draft lists

### Related issues
https://github.com/OpenCTI-Platform/opencti/issues/12372

### DIFF
--- a/opencti-platform/opencti-front/src/components/ItemOperations.tsx
+++ b/opencti-platform/opencti-front/src/components/ItemOperations.tsx
@@ -1,20 +1,9 @@
 import Chip from '@mui/material/Chip';
 import React, { FunctionComponent } from 'react';
-import makeStyles from '@mui/styles/makeStyles';
 import { useTheme } from '@mui/styles';
 import type { Theme } from './Theme';
 import { useFormatter } from './i18n';
 import { chipInListBasicStyle } from '../utils/chipStyle';
-
-const useStyles = makeStyles(() => ({
-  chipInList: {
-    ...chipInListBasicStyle,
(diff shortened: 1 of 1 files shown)


JeremyCloarec:
Increasing the width of the chip does work properly 👍 
However, now that the operation is translated, it doesn't work well in some languages:
image
Is it something that was confirmed with @alice-debra that we do want translations on those operations? If so, we'll probably need to update the documentation to reflect these changes: https://docs.opencti.io/latest/usage/draftWorkspaces/?h=draft

Archidoit:
@JeremyCloarec I though it was an oversight, if it's decided that there is no translation, I remove it :)

JeremyCloarec:
Yes I think keeping only the width change is a good thing for now, from what I remember from our conversations, it was intentional to not translate those operations

Archidoit:
Ok, I remove the translations @JeremyCloarec  :)

JeremyCloarec:
Now that it's back to the original update_linked text, it stil doesn't display properly :(
image
But at least we can see update_li... instead of just update. I'm not sure what would be the best wait to display it, can you ping @alice-debra to see if this is enough for visibility and can be merged like this? Or if we need to rework it further

Archidoit:
I'm gonna fix that @JeremyCloarec , 120 instead of 100 for width

image

alice-debra:
> I'm gonna fix that @JeremyCloarec , 120 instead of 100 for width
> 
> image

It looks better like this!

JeremyCloarec (approved):

PR #12416, opened 2025-09-16
[backend] Refactor catalog cache management to fix flaky connector tests (#7328)

maelv-filigran:
### Proposed changes

* Extracted catalog building logic into a dedicated `buildCatalogMap()` function in `catalog-domain.ts` to separate cache building from cache retrieval
* Added `resetCatalogCache()` function to allow tests to clear and reload the catalog cache
* Updated connector composer tests to reset catalog cache before running, ensuring they use the correct test catalog
* Re-enabled all previously skipped tests (31 tests) in `connector-composer-test.ts` that were disabled due to catalog caching issues

### Related issues
* Fixes flaky connector composer tests where `getSupportedContractsByImage()` was using a different catalog than the tests
* #7328 

### Checklist

### Further comments

The root cause was the global `catalogMap` variable being cached and not refreshed when tests set a custom catalog path via environment variables. By extracting the catalog building logic and providing a cache reset mechanism, tests can now reliably use their test catalog without interference from the production catalog cache.

All 31 connector composer tests are now passing consistently without any flaky behavior.

### DIFF
--- a/opencti-platform/opencti-graphql/src/modules/catalog/catalog-domain.ts
+++ b/opencti-platform/opencti-graphql/src/modules/catalog/catalog-domain.ts
@@ -17,6 +17,83 @@ addFormats(ajv, ['password', 'uri', 'duration', 'email', 'date-time', 'date']);
 
 // Cache of catalog to read on disk and parse only once
 let catalogMap: Record<string, CatalogType>;
+
+// Build catalog map from files
+const buildCatalogMap = (): Record<string, CatalogType> => {
+  const newCatalogMap: Record<string, CatalogType> = {};
+
+  const catalogs = CUSTOM_CATALOGS.map((custom) => fs.readFileSync(custom, { encoding: 'utf8', flag: 'r' }));
+  catalogs.push(JSON.stringify(filigranCatalog));
+
+  for (let index = 0; index < catalogs.length; index += 1) {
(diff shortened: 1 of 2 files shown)


opencti-platform/opencti-graphql/src/modules/catalog/catalog-domain.ts
```diff
@@ -17,6 +17,83 @@ addFormats(ajv, ['password', 'uri', 'duration', 'email', 'date-time', 'date']);
...
+        })
+      }
+    };
+  }
+
+  return newCatalogMap;
+};
+
+// Reset catalog cache - for testing purposes only
+export const resetCatalogCache = () => {
```
esrevi, line 93: Maybe we can remove this function and for the test, you could try replacing the call of resetCatalogCache with vi.resetModules() (https://vitest.dev/api/vi.html#vi-resetmodules) 

I don't know if it will work though, but could be worth to try :)

maelv-filigran: This isn’t possible, as it would require a dynamic import, which isn’t supported in the current build configuration.

ex:
 ```typescript
const catalogDomain = await import('../../../src/modules/catalog/catalog-domain');
```

esrevi (approved):

PR #12443, opened 2025-09-17
[frontend] Shared With relationships should not be counted in Knowledge view (#11283)

### LINKED ISSUE #11283: Attack patterns "shared with" Orgs in knowledge view of an org
## Description

When sharing a TTP with an Org (org Sharing), the Attack Patterns are counted in the Knowledge View "Attack Pattern" of an organisation. But not displayed. It should not be the case, they should not be counted. 

## Environment

Testing

## Reproducible Steps

1. Share an attack pattern with an organisation
2. Go in the same organisation that you shared the attack pattern with
3. Look at the counter 

## Expected Output

The counter should not be incremented with this "shared with" (in other words, shared with should be excluded from the count of any knowledge of any entity)

## Actual Output

The counter is incremented (+1), but nothing displayed in the knwoledge view.

Archidoit:
### Proposed changes
In Entities knowledge view, in the right bar, the counters should not count 'shared-with' relations

### Related issues
https://github.com/OpenCTI-Platform/opencti/issues/11283


marieflorescontact (approved):
locally tested ok 👌

PR #12444, opened 2025-09-17
[backend](PIR) Fetch stream events range from event id for PIR (#11149)

Archidoit:
### Proposed changes
stream events range are now fetched and processed by batch of 1000 events every 30s by the pir manager

### DIFF
--- a/opencti-platform/opencti-graphql/src/database/redis.ts
+++ b/opencti-platform/opencti-graphql/src/database/redis.ts
@@ -721,7 +721,7 @@ interface StreamOption {
   bufferTime?: number;
   autoReconnect?: boolean;
   streamName?: string;
-  streamBatchTime?: number
+  streamBatchSize?: number
 }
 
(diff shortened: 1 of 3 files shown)


opencti-platform/opencti-graphql/src/database/redis.ts
```diff
@@ -823,7 +823,7 @@ export const fetchStreamEventsRangeFromEventId = async (
 ) => {
   let effectiveStartEventId = startEventId;
   const [startTimestamp] = startEventId.split('-');
-  const endTimestamp = startTimestamp + (opts.streamBatchTime ?? STREAM_BATCH_TIME);
+  const endTimestamp = (Number(startTimestamp) + (opts.streamBatchTime ?? STREAM_BATCH_TIME)).toString();
```
SouadHadjiat: we need to use count instead of endTimestamp

SouadHadjiat (approved):

PR #12451, opened 2025-09-17
[backend] delete work when deleting task & verify that work exists in worker (#12815)

### LINKED ISSUE #12815: Add ability to cancel the processing of tasks by the worker after the task has been provisioned
## Use case

Currently, when deleting a task that has already been provisioned, the worker will always process the task even if it is deleted in the meantime.

JeremyCloarec:
### Proposed changes

* now also delete associated work when deleting task
* in worker, verify that bundle work_id is still valid. If work no longer exists, ack the message without processing it

client PR: https://github.com/OpenCTI-Platform/client-python/pull/983

### Related issues

* #12815
*

### Checklist


opencti-worker/src/worker.py
```diff
@@ -330,6 +330,15 @@ def data_handler(  # pylint: disable=too-many-statements, too-many-locals
...
             self.api.set_previous_standard_header(previous_standard)
+            # Check if work is still valid
+            if work_id is not None:
+                work = self.api.work.get_work(work_id)
+                # If work no longer exists, bundle can be acked without doing anything
+                if work is None:
+                    cb = functools.partial(self.ack_message, channel, delivery_tag)
+                    connection.add_callback_threadsafe(cb)
+                    return
+
```
JeremyCloarec: This introduces an unreasonable overhead. The solution need to be reworked

opencti-platform/opencti-graphql/src/domain/work.js
```diff
@@ -41,6 +41,11 @@ export const findById = (context, user, workId) => {
   return loadWorkById(context, user, workId);
 };
 
+export const isWorkAlive = async (_context, _user, workId) => {
+  const redisWork = await redisGetWork(workId);
+  return !!redisWork && redisWork.is_initialized === 'true';
```
xfournet: * can use elvis operator: `return redisWork?.is_initialized === 'true'`
* is it really the string 'true' that we are testing ?

opencti-platform/opencti-graphql/src/database/redis.ts
```diff
@@ -911,6 +911,12 @@ export const redisUpdateActionExpectation = async (user: AuthUser, workId: strin
   });
   return workId;
 };
+export const redisInitializeWork = async (workId: string) => {
+  await redisTx(getClientBase(), async (tx) => {
+    await updateObjectRaw(tx, workId, { is_initialized: true });
+  });
+  return workId;
```
xfournet: we don't use the returned value (which is the function parameter)

opencti-worker/src/worker.py
xfournet: what overhead, the additional call to redis ?

JeremyCloarec: I left this comment as a reminder after a call with Julien: in the first version of this PR, we were calling self.api.work.get_work, that lead to an elastic read. Julien made the remark that it was too heavy of an overhead on each message processing in the worker.

This was the reason to moving to a self.api.work.get_is_work_alive call, which now leads to a redis call instead

opencti-platform/opencti-graphql/src/domain/work.js
JeremyCloarec: Yes, I didn't implement any bool parsing in redisGetWork so we're getting the raw string value from redis

opencti-worker/src/worker.py
xfournet: ok so now there is no more overhead concerns ?

xfournet (approved):

opencti-platform/opencti-graphql/src/database/redis.ts
xfournet, line 1: in redisUpdateWorkFigures handle the race condition with already closed work

Issue #12441, opened 2025-09-17
Empty authors should be displayed as '-'


Archidoit:
## Description
If no author in an entities list, we should display '-' instead of nothing

## Reproducible Steps
It's the case in several lists, example: lists of entities in an entity Knowledge tab

<img width="1842" height="592" alt="Image" src="https://github.com/user-attachments/assets/2ee810f4-cc87-4590-8e1a-10440eabd8f6" />

## Goal
UI consistancy

Archidoit:
More generally, in entities overviews and lists: empty fields should be displayed as '-'

PR #12456, opened 2025-09-18
[frontend] Empty fields display in entities lists and overview (#12441)

Archidoit:
### Proposed changes
- In entities list, empty authors should be displayed as '-' instead of nothing
- In entities overview, empty fields should be displayed as '-'
- In Add Notes list panel, fix the display of markings and authors (None was always displayed even if there were markings)

### Related issues
https://github.com/OpenCTI-Platform/opencti/issues/12441


CelineSebe:
Works well :). I found some more (not exhaustive)

image
image
image
image
image
image
image

CelineSebe (approved):

PR #12462, opened 2025-09-18
[backend] improve inputResolveRefs perfs (#12455)

### LINKED ISSUE #12455: Technical issue: inputResolveRefs function can lead to high CPU usage
## Description

matchingConfigs computation can lead to high CPU usage: it's currently a O(n²) computation, we should change it to be only O(n)
Image

## Environment

4. Other environment details:

## Reproducible Steps

1. { e.g. Run ... }
2. { e.g. Click ... }
3. { e.g. Error ... }

JeremyCloarec:
### Proposed changes

* inputResolveRefsFn was CPU inefficient: resolved elements were being identified in O(n²) in matchingConfigs computation. It is now O(n), and it now goes through instancesIds instead of going through the whole fetchingsIds for every resolvedElements
*

### Related issues

* #12455
*

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/database/middleware.js
+++ b/opencti-platform/opencti-graphql/src/database/middleware.js
@@ -803,7 +803,7 @@ const inputResolveRefs = async (context, user, input, type, entitySetting) => {
             const vocabularyId = idVocabulary(i, category);
             const vocabularyElement = { id: vocabularyId, destKey, multiple: isListing };
             if (fetchingIdsMap.has(vocabularyId)) {
-              fetchingIdsMap[vocabularyId].push(vocabularyElement);
+              fetchingIdsMap.get(vocabularyId).push(vocabularyElement);
             } else {
               fetchingIdsMap.set(vocabularyId, [vocabularyElement]);
             }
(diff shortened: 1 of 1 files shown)


opencti-platform/opencti-graphql/src/database/middleware.js
```diff
@@ -790,26 +790,46 @@ const inputResolveRefs = async (context, user, input, type, entitySetting) => {
...
         if (src === INPUT_LABELS) {
-          const elements = R.uniq(id.map((label) => idLabel(label)))
-            .map((lid) => ({ id: lid, destKey, multiple: true }));
-          fetchingIds.push(...elements);
-          expectedIds.push(...elements.map((e) => e.id));
+          R.uniq(id.map((label) => idLabel(label)))
+            .forEach((labelId) => {
+              const labelElement = { id: labelId, destKey, multiple: true };
+              expectedIds.push(labelId);
+              fetchingIdsMap.set(labelId, [labelElement]);
```
xfournet: should we rather test if the fetchingIdsMap aleady contains a value, and complete it if it exists ?

JeremyCloarec: We're not supposed to have multiple label inputs, so the `labelIds` should be safely considered unique, but I can add a check to add another layer of safety 👍

xfournet (approved):

PR #12468, opened 2025-09-18
[frontend] fetch smaller queries in ingestion management screens (#7328)

esrevi:
### Proposed changes

* Add new resolver to fetch only the title and slug to filter the connector
* This permits to not require field manager_contract_definition which could be heavy because of the json

### Related issues

* #7328 
*

### Checklist


natalie-natsu:
Fixes https://github.com/OpenCTI-Platform/opencti/issues/12470

opencti-platform/opencti-graphql/src/database/repository.js
```diff
@@ -44,6 +44,20 @@ export const computeManagerConnectorContract = async (_context, _user, cn) => {
   return contract ? JSON.stringify(contract) : contract;
 };
 
+export const computeManagerConnectorExcerpt = async (_context, _user, cn) => {
+  const contracts = getSupportedContractsByImage();
+  const contract = contracts.get(cn.manager_contract_image);
+
+  if (!contract) {
```
maelv-filigran, line 56: Returning null when no contract is found could hide underlying issues.
Would it be possible to add logging to track cases where an expected contract is not found?

opencti-platform/opencti-front/src/private/components/data/connectors/hooks/useConnectorsStatusFilters.ts
```diff
@@ -48,13 +46,7 @@ const useConnectorsStatusFilters = ({ connectors, managerContractDefinitionMap,
     }
 
     if (filters.slug) {
-      const parsedDefinition = managerContractDefinitionMap.get(connector.id);
-      if (!parsedDefinition?.slug) return false;
-
-      const slugMatch = parsedDefinition.slug
-        .toLowerCase()
-        .includes(filters.slug.toLowerCase());
-      if (!slugMatch) return false;
```
JeremyCloarec, line 57: The .toLowerCase() were removed from the code here, shouldn't they be kept?

esrevi: yes indeed, good catch ^^

opencti-platform/opencti-graphql/src/database/repository.js
esrevi: yes !

JeremyCloarec (approved):

PR #12485, opened 2025-09-19
feat(backend): store large stream event messages to S3 instead of Redis (#12089)

### LINKED ISSUE #12089: perf(redis): a specific large event can lead to excessive Redis usage
## Description

On specific circonstances the redis trimming can be an insufficiant limitation to prevent too much memory usage.

## Reproducible Steps

Create a connector that create a container.
Add in this container 500K elements one by one.

## Expected Output

Redis stable in term of memory

## Actual Output

As the element growing at each add, the event size in average in growing to much.
Because of this we accumulate in the Redis very big messages that lead to a memory outage

JeremyCloarec:
### Proposed changes

* store stream event on S3 instead of redis if event data is too large
* clean S3 stream files on a cron schedule

### Related issues

* #12089
*

### Checklist


aHenryJard (commented):
Please add one event bigger than max_event_length in integration tests (maybe lower max_event_length and send a medium event >  than that)

To be able to validate that all sync work correctly with the pointer resolution.

opencti-platform/opencti-graphql/src/manager/garbageCollectionManager.ts
```diff
@@ -10,6 +15,31 @@ const SCHEDULE_TIME = conf.get('garbage_collection_manager:interval') || 60000;
...
+    nextRedisStreamFilesGarbageTime += 86400000;
+
+    const { firstEventId } = await rawFetchStreamInfo();
+    const timestamp = firstEventId.split('-')[0];
+    const allRedisLargeEventFiles = await loadedFilesListing(context, GARBAGE_COLLECTION_MANAGER_USER, STREAM_FILE_DIRECTORY, { rawFormat: true }) as ListObjectsV2CommandOutput;
+    if (!allRedisLargeEventFiles || !allRedisLargeEventFiles.KeyCount || !allRedisLargeEventFiles.Contents) {
+      return;
+    }
+    for (let i = 0; i < allRedisLargeEventFiles?.KeyCount; i += 1) {
+      const file = allRedisLargeEventFiles?.Contents[0];
```
gabrielebonadiman: I think we want to scan the indexed file, `0` is always gonna point to the first element.
```suggestion
      const file = allRedisLargeEventFiles?.Contents[i];
```

gabrielebonadiman (commented):
Also, is there a way or steps I can follow, to test this ?

opencti-platform/opencti-graphql/config/default.json
```diff
@@ -381,7 +381,8 @@
     "ca": [],
     "port": 6379,
     "host_ip_family": 4,
-    "trimming": 2000000
+    "trimming": 2000000,
+    "max_event_length": 10000000
```
aHenryJard, line 385: Could you please add it in documentation ?

opencti-platform/opencti-graphql/src/database/redis-stream.ts
```diff
@@ -58,24 +61,54 @@ const mapStreamToJS = ([id, data]: any): SseEvent<any> => {
...
+    try {
+      const fileContent = await getFileContent(filePath);
+      if (!fileContent) {
+        logApp.warn('Stream event file could not be found', { id, filePath });
+        return null;
+      }
+      const fileResult = JSON.parse(fileContent);
+      return mapStreamToJS([id, fileResult]);
+    } catch (error) {
+      logApp.warn('Error fetching file stream event, skipping event', { id, filePath, error });
```
aHenryJard, line 103: This will lead to data inconsistency on platform synchronization right ? So I wonder about moving that to error level.

opencti-platform/opencti-graphql/src/database/redis-stream.ts
```diff
@@ -58,24 +61,54 @@ const mapStreamToJS = ([id, data]: any): SseEvent<any> => {
...
   const redisClient = getClientBase();
-  const eventStreamData = mapJSToStream(event);
+  let eventStreamData = mapJSToStream(event);
+  const eventStreamDataBlob = new Blob(eventStreamData);
+  const totalStreamEventSize = eventStreamDataBlob.size;
+  if (streamMaxEventSize > 0 && totalStreamEventSize > streamMaxEventSize) {
+    // Add salt to prevent time collision
+    const randomSalt = Math.floor(Math.random() * 1000);
+    const eventId = streamEventId(null, randomSalt);
+    const filePath = `${STREAM_FILE_DIRECTORY}${eventId}`;
```
aHenryJard, line 80: I tested, having one stream in data > sharing : the payload of both "global" stream under /stream and the one under a dedicated stream like /stream/50c547d6-8de8-4202-a73f-9ff6f428f7c0 ends inside the same "stream.opencti" folder. I wonder if for example we want to cleanup a directory when a data > sharing live stream is deleted if we want them all at the same place. In the other hand the garbage collector will cleanup at some point so it's fine. Juste asking if both storage path have been discussed (woth or without id in the path).

Image

opencti-platform/opencti-graphql/src/manager/garbageCollectionManager.ts
```diff
@@ -10,6 +15,31 @@ const SCHEDULE_TIME = conf.get('garbage_collection_manager:interval') || 60000;
...
+  if (nextRedisStreamFilesGarbageTime < currentTime) {
+    nextRedisStreamFilesGarbageTime += 86400000;
+
+    const { firstEventId } = await rawFetchStreamInfo();
+    const timestamp = firstEventId.split('-')[0];
+    const allRedisLargeEventFiles = await loadedFilesListing(context, GARBAGE_COLLECTION_MANAGER_USER, STREAM_FILE_DIRECTORY, { rawFormat: true }) as ListObjectsV2CommandOutput;
+    if (!allRedisLargeEventFiles || !allRedisLargeEventFiles.KeyCount || !allRedisLargeEventFiles.Contents) {
+      return;
+    }
+    for (let i = 0; i < allRedisLargeEventFiles?.KeyCount; i += 1) {
```
aHenryJard: You may be able to sort by event id, and remove all strictly oldest than the firstEventId timestamp.

opencti-platform/opencti-graphql/src/manager/garbageCollectionManager.ts
```diff
@@ -10,6 +15,31 @@ const SCHEDULE_TIME = conf.get('garbage_collection_manager:interval') || 60000;
...
+      const file = allRedisLargeEventFiles?.Contents[0];
+      if (!file.Key) {
+        continue;
+      }
+      const currentTimestamp = file.Key.slice(STREAM_FILE_DIRECTORY.length).split('-')[0];
+      if (currentTimestamp < timestamp) {
+        await deleteFileFromStorage(file.Key);
```
aHenryJard, line 39: it might be interesting to delete several file at once, is it possible ?

opencti-platform/opencti-graphql/src/manager/garbageCollectionManager.ts
```diff
@@ -10,6 +15,31 @@ const SCHEDULE_TIME = conf.get('garbage_collection_manager:interval') || 60000;
 const BATCH_SIZE = conf.get('garbage_collection_manager:batch_size') || 10000;
 const DELETED_RETENTION_DAYS = conf.get('garbage_collection_manager:deleted_retention_days') || 7;
 
+let nextRedisStreamFilesGarbageTime = new Date().getTime();
+const garbageCollectRedisStreamFiles = async (context: AuthContext) => {
+  const currentTime = new Date().getTime();
+  if (nextRedisStreamFilesGarbageTime < currentTime) {
+    nextRedisStreamFilesGarbageTime += 86400000;
```
aHenryJard: can we use a conf.get instead of a magic hard coded number ? (maybe without ms precision to be more readable)

opencti-platform/opencti-graphql/src/manager/garbageCollectionManager.ts
aHenryJard: Also I wonder if we should use the batch_size somehow, I have no idea on how many files can ends there on minio... more than 10k ? At least it's a limit

opencti-platform/opencti-graphql/src/database/redis-stream.ts
JeremyCloarec: There is only one stream stored on redis currently, the global stream. All other streams are a "filter" on the global stream (the filtering is done in sseMiddleware), and do not have any specific stream stored on redis. So the stream folder on S3 is always going to contain the large events from the "global" stream

opencti-platform/opencti-graphql/src/database/redis-stream.ts
JeremyCloarec: Good point, now that I think about it I think we might even want to not catch the error and let it bubble up. Skipping events silently like this can be dangerous I think, it's better to let the consumer fail and retry this event than to skip it. Wdyt?

opencti-platform/opencti-graphql/src/manager/garbageCollectionManager.ts
JeremyCloarec: I was thinking that since it's only run every 24h by default, batching is not that necessary

opencti-platform/opencti-graphql/tests/03-integration/02-resolvers/administrativeArea-test.js
```diff
@@ -103,39 +105,55 @@ describe('AdministrativeArea resolver standard behavior', () => {
...
+      expect(queryResult.data.administrativeAreaAdd.standard_id).toEqual('location--9904c841-f308-58bf-a39a-6ecd6024d3e0');
+      expect(queryResult.data.administrativeAreaAdd.objectLabel.length).toEqual(3);
+      administrativeAreaInternalId = queryResult.data.administrativeAreaAdd.id; // bc1f31d7-4d9d-4754-89b1-9a7813c7c521
+
+      // The creation stream event must have been offloaded to S3.
+      expect(isEventTooLargeSpy).toHaveBeenCalled();
+      expect(rawUploadSpy).toHaveBeenCalled();
+      const offloadedToStream = rawUploadSpy.mock.calls.some(([key]) => typeof key === 'string' && key.startsWith(redisStream.STREAM_FILE_DIRECTORY));
+      expect(offloadedToStream).toBe(true);
+    } finally {
```
gabrielebonadiman, line 153: We are testing the upload functionality fine, but I would also add the read of the uploaded file. it's working fine on my local, both read and write, so I think it would be a nice add to avoid regressions.

gabrielebonadiman: covered by local_synchronizer.py 👍

gabrielebonadiman (approved):

PR #12500, opened 2025-09-22
[backend] Fix pir relationship stix converter (#11149)

Archidoit:
### Proposed changes
 Fix pir relationship stix converter

to avoid having these error in backend when modifying the author of a relationship having caused a flag : 

image


Archidoit:
done in a separated PR

PR #12502, opened 2025-09-22
[backend] add entity type verification

aHenryJard:
### Proposed changes

* Add entity or relation type verification before delete.

### Related issues

* N/A

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/database/middleware.js
+++ b/opencti-platform/opencti-graphql/src/database/middleware.js
@@ -122,7 +122,7 @@ import {
   RELATION_OBJECT_MARKING,
   STIX_REF_RELATIONSHIP_TYPES
 } from '../schema/stixRefRelationship';
-import { ENTITY_TYPE_SETTINGS, ENTITY_TYPE_STATUS, ENTITY_TYPE_USER } from '../schema/internalObject';
+import { ENTITY_TYPE_RULE, ENTITY_TYPE_SETTINGS, ENTITY_TYPE_STATUS, ENTITY_TYPE_USER } from '../schema/internalObject';
 import { isStixCoreObject } from '../schema/stixCoreObject';
 import { isBasicRelationship } from '../schema/stixRelationship';
 import {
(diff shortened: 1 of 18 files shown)


opencti-platform/opencti-graphql/src/database/middleware.js
```diff
@@ -3445,10 +3445,19 @@ const draftInternalDeleteElement = async (context, user, draftElement) => {
...
 
-export const internalDeleteElementById = async (context, user, id, opts = {}) => {
+export const internalDeleteElementById = async (context, user, id, type, opts = {}) => {
   let lock;
   let event;
   const element = await storeLoadByIdWithRefs(context, user, id, { ...opts, includeDeletedInDraft: true });
+
+  if (element === undefined) {
+    throw AlreadyDeletedError({ id });
+  }
```
lndrtrbn: I'm not sure to understand why we need this check, could we not just move the check on type below the `if(!element) ...` instead?

aHenryJard: I'm not sure it's even needed... I think I added that when some tests were failing but it was not the cause at the end.

aHenryJard: Let me try to remove, thank for the sharp eyes !

lndrtrbn (approved):

opencti-platform/opencti-graphql/tests/02-integration/01-database/middleware-delete-test.ts
```diff
@@ -0,0 +1,41 @@
...
+import { addDecayRule } from '../../../src/modules/decayRule/decayRule-domain';
+import type { DecayRuleAddInput } from '../../../src/generated/graphql';
+import { ENTITY_TYPE_GROUP } from '../../../src/schema/internalObject';
+
+describe('Delete functional errors behaviors', async () => {
+  it('should not be able to delete organization that has members', async () => {
+    await expect(() => deleteElementById(testContext, ADMIN_USER, TEST_ORGANIZATION.id, ENTITY_TYPE_IDENTITY_ORGANIZATION))
+      .rejects.toThrowError('Cannot delete an organization that has members.');
+  });
+  it.skip('should not be able to delete individual associated to user', async () => {
```
marieflorescontact, line 16: di you forget .skip?

aHenryJard: No I just moved this test code to a ts file, the skip was already there here https://github.com/OpenCTI-Platform/opencti/blob/a9935e70bd73c33d39ecb2564d4eb5b7352b3638/opencti-platform/opencti-graphql/tests/02-integration/01-database/middleware-test.js#L1473

marieflorescontact (approved):

PR #12508, opened 2025-09-22
feat(frontend/backend): handle `Unknown` values in elAggregationCount to count unassigned field for WidgetDonut (#11030)

### LINKED ISSUE #11030: fix(dashboards): incorrect data rendering in donut widget
## Description

When using the Donut chart widget to visualize report assignments in OpenCTI, the data shown does not reflect the actual distribution of values across all reports. 

Instead, it seems to filter and only display data for the reports that contain the selected attribute (in this case, assignees). 

As a result, the chart misleadingly shows a 50%/50% distribution when, in reality, only 2 out of 693 reports have an assignee. The remaining 691 are ignored in the visualization.

## Environment

OpenCTI 6.6.11

## Reproducible Steps

1. Create a widget "Number" on all reports
2. Create another widget "Number" on all reports AND assignees is empty. 

Now you have the figures. 

Then 
3. Create a "Donut" widget based on all reports 
4. In step 4.Parameters, Select the Attribute Assignee

## Expected Output

The Donut chart should represent the data relative to the entire dataset (e.g., 693 reports), including the proportion of unassigned reports. Example:
Leila Haddad: (1/693)x100 = 0,15%
Pedro Alvarez: (1/693)x100 = 0,15%
Unassigned/Unknown: (691/693)x100 = 99,7%

## Actual Output

On my platform, the chart displays only the 2 assigned reports, showing a 50%/50% distribution between Leila Haddad and Pedro Alvarez, ignoring the remaining 691 unassigned reports.

## Additional information

This leads to a misleading interpretation of the data and is inconsistent with what is shown in the counter widget.

## Screenshots (optional)

Image

delemaf:
### Proposed changes

* Handle `Unknown` values in elAggregationCount to count unassigned field for WidgetDonut 

image

image

### Related issues

* #11030

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/utils/hooks/useDistributionGraphData.ts
+++ b/opencti-platform/opencti-front/src/utils/hooks/useDistributionGraphData.ts
@@ -107,7 +107,7 @@ const useDistributionGraphData = () => {
    */
   const buildWidgetLabelsOption = (distributionData: DistributionQueryData, groupBy: string) => {
     return distributionData.map((n) => {
-      if (!n) return 'Unknown';
+      if (!n || n.label === 'unknown') return 'Unknown';
       if (isFieldForIdentifier(groupBy)) {
         return getMainRepresentative(n.entity);
       }
--- a/opencti-platform/opencti-front/src/utils/hooks/useDistributionGraphData.ts
+++ b/opencti-platform/opencti-front/src/utils/hooks/useDistributionGraphData.ts
@@ -125,7 +125,7 @@ const useDistributionGraphData = () => {
    */
   const buildWidgetWordCloudOption = (distributionData: DistributionQueryData, groupBy: string) => {
     return distributionData.map((n) => {
-      if (!n) return { text: 'Unknown', value: 0 };
+      if (!n || n.label === 'unknown') return { text: 'Unknown', value: 0 };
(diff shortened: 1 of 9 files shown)


JeremyCloarec (approved):

opencti-platform/opencti-front/src/utils/hooks/useDistributionGraphData.ts
fellowseb: Shouldn't we here too handle the `n.label === 'unknown'` case ?

opencti-platform/opencti-graphql/src/database/middleware.ts
```diff
@@ -781,6 +787,10 @@ const convertAggregateDistributions = async (
   }
   return filteredData
     .map((n) => {
+      // The 'unknown' bucket has no backing entity
+      if (n.label === 'unknown') {
+        return { ...n, entity: null as unknown as BasicStoreEntity };
```
fellowseb: Not really fond of this assert. When fixing the typing locally I now see potential "Cannot read properties of null" exception arising in `publicDashboard-domain.ts` for instance. I think it's better to fix the type and follow the rabbit hole

fellowseb (changes requested):

fellowseb (approved):

PR #12512, opened 2025-09-23
[backend] improve inputResolveRefs time & memory complexity

JeremyCloarec:
### Proposed changes

* reduce memory & time complexity by removing construction of deduplicated id array and using foreach to check for duplication

### Related issues
* #12455

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/database/middleware.js
+++ b/opencti-platform/opencti-graphql/src/database/middleware.js
@@ -813,9 +813,12 @@ const inputResolveRefs = async (context, user, input, type, entitySetting) => {
             }
           });
         } else if (isListing) {
-          R.uniq(id).forEach((i) => {
+          id.forEach((i) => {
             const listingElement = { id: i, destKey, multiple: true };
             if (fetchingIdsMap.has(i)) {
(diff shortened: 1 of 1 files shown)


opencti-platform/opencti-graphql/src/database/middleware.js
```diff
@@ -820,9 +813,12 @@ const inputResolveRefs = async (context, user, input, type, entitySetting) => {
             }
           });
         } else if (isListing) {
-          R.uniq(id).forEach((i) => {
+          id.forEach((i) => {
             const listingElement = { id: i, destKey, multiple: true };
             if (fetchingIdsMap.has(i)) {
+              if (fetchingIdsMap.get(i).includes((e) => e.destKey === destKey)) {
+                return;
+              }
```
xfournet, line 821: i need more explanation to understand what we want to achieve here

JeremyCloarec: We were using R.uniq(id) before to deduplicate the ids, but we now check for duplication within the foreach instead. 

This is a minimal perf gain since this only saves up one loop through the id array (by removing R.uniq) but this is still a line that came up sometimes in our saas telemetry so I figured it was still something worth improving

xfournet (approved):

PR #12542, opened 2025-09-25
[Backend] Improved RelayState verification

MTorbay-Filigran:
### Proposed changes

* sanitized RelayState verification
*

### Related issues

*
*

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/http/httpPlatform.js
+++ b/opencti-platform/opencti-graphql/src/http/httpPlatform.js
@@ -458,17 +458,30 @@ const createApp = async (app, schema) => {
   // -- Passport callback
   // -- Default limit is '100kb' based on https://expressjs.com/en/resources/middleware/body-parser.html
   const urlencodedParser = AUTH_PAYLOAD_BODY_SIZE ? bodyParser.urlencoded({ extended: true, limit: AUTH_PAYLOAD_BODY_SIZE }) : bodyParser.urlencoded({ extended: true });
-  app.all(`${basePath}/auth/:provider/callback`, urlencodedParser, async (req, res, next) => {
-    if (req.body.RelayState) {
-      const refererUrl = new URL(req.body.RelayState, getBaseUrl());
-      if (refererUrl.pathname !== req.body.RelayState) {
-        logApp.error('Error auth provider callback : RelayState has been altered', { relayState: req.body.RelayState });
-        setCookieError(res, 'Invalid authentication, please ask your administrator');
-        res.redirect('/');
-        return;
-      }
+  const sanitizeReferer = (refererToSanitize, baseUrl) => {
+    if (!refererToSanitize) return '/';
+    const base = new URL(baseUrl);
+    const refererUrl = new URL(refererToSanitize, base);
+
+    const isSameOrigin = refererUrl.origin === base.origin;
+    const isUrlRelative = refererToSanitize.startsWith('/');
+
+    if (isSameOrigin || isUrlRelative) {
+      return refererUrl.pathname + refererUrl.search + refererUrl.hash;
     }
-    const referer = req.body.RelayState ?? req.session.referer;
+    logApp.info('Error auth provider callback : url has been altered', { url: refererToSanitize });
+    return '/';
+  };
+
+  app.all(`${basePath}/auth/:provider/callback`, urlencodedParser, async (req, res, next) => {
+    const baseUrl = getBaseUrl();
+    let referer = '/';
+
+    if (req.body.RelayState) {
+      referer = sanitizeReferer(req.body.RelayState, baseUrl);
+    } else if (req.session.referer) {
+      referer = sanitizeReferer(req.session.referer, baseUrl);
+    }
     const { provider } = req.params;
     const callbackLogin = () => new Promise((accept, reject) => {
       passport.authenticate(provider, {}, (err, user) => {
--- a/opencti-platform/opencti-graphql/src/http/httpPlatform.js
+++ b/opencti-platform/opencti-graphql/src/http/httpPlatform.js
@@ -37,13 +37,12 @@ import { getChatbotProxy } from './httpChatbotProxy';
 export function sanitizeReferer(refererToSanitize) {
   if (!refererToSanitize) return '/';
   const base = getBaseUrl();
-  const refererUrl = new URL(refererToSanitize, base);
-  const isSameOrigin = refererUrl.origin === base;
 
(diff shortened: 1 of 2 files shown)


marieflorescontact (approved):
locally tested ok (with a base path) 👌

opencti-platform/opencti-graphql/src/http/httpPlatform.js
```diff
@@ -459,6 +459,15 @@ const createApp = async (app, schema) => {
   // -- Default limit is '100kb' based on https://expressjs.com/en/resources/middleware/body-parser.html
   const urlencodedParser = AUTH_PAYLOAD_BODY_SIZE ? bodyParser.urlencoded({ extended: true, limit: AUTH_PAYLOAD_BODY_SIZE }) : bodyParser.urlencoded({ extended: true });
   app.all(`${basePath}/auth/:provider/callback`, urlencodedParser, async (req, res, next) => {
+    if (req.body.RelayState) {
+      const refererUrl = new URL(req.body.RelayState, getBaseUrl());
+      if (refererUrl.pathname !== req.body.RelayState) {
+        logApp.error('Error auth provider callback : RelayState has been altered', { relayState: req.body.RelayState });
```
aHenryJard:
```suggestion
        logApp.info('Error auth provider callback : RelayState has been altered', { relayState: req.body.RelayState });
```

opencti-platform/opencti-graphql/src/http/httpPlatform.js
aHenryJard: check that referer startWith === "getBaseUrl()" or startWith getBaseUrl()/ 
if not redirect to '/'

opencti-platform/opencti-graphql/src/http/httpPlatform.js
```diff
@@ -459,6 +459,15 @@ const createApp = async (app, schema) => {
...
+        setCookieError(res, 'Invalid authentication, please ask your administrator');
+        res.redirect('/');
+        return;
```
aHenryJard: To be removed (it's all in finally)

aHenryJard (commented):
Updated with comment

opencti-platform/opencti-graphql/src/http/httpPlatform.js
MTorbay-Filigran: If we delete all these lines, then nothing happens when there is a mismatch between refererUrl.pathname and req.body.RelayState except a logApp.info, is that enough ?

opencti-platform/opencti-graphql/src/http/httpPlatform.js
```diff
@@ -458,8 +458,30 @@ const createApp = async (app, schema) => {
...
+    return '/';
+  };
+
   app.all(`${basePath}/auth/:provider/callback`, urlencodedParser, async (req, res, next) => {
-    const referer = req.body.RelayState ?? req.session.referer;
+    const baseUrl = getBaseUrl();
+    let referer = '/';
+
+    if (req.body.RelayState) {
+      referer = sanitizeReferer(req.body.RelayState, baseUrl);
```
aHenryJard: I'm not sure we need to have baseUrl as argument here since there is `getBaseUrl()` function.

opencti-platform/opencti-graphql/src/http/httpPlatform.js
```diff
@@ -458,8 +458,30 @@ const createApp = async (app, schema) => {
   // -- Passport callback
   // -- Default limit is '100kb' based on https://expressjs.com/en/resources/middleware/body-parser.html
   const urlencodedParser = AUTH_PAYLOAD_BODY_SIZE ? bodyParser.urlencoded({ extended: true, limit: AUTH_PAYLOAD_BODY_SIZE }) : bodyParser.urlencoded({ extended: true });
+  const sanitizeReferer = (refererToSanitize, baseUrl) => {
```
aHenryJard: You could move that up, export and cover with domain level tests.

opencti-platform/opencti-graphql/src/http/httpPlatform.js
xfournet: * there is many place where we default to '/' (here, when initializing referer, and in the sanitize method)
* to avoid risk of breaking this sanitization later when updating the code here, it would be better to process the sanitization near the redirect

we can remove line 476-481 and here replace with
```
 } finally {
      const sanitizedReferer = sanitizeReferer(req.body.RelayState ?? req.session.referer ?? '/');
      res.redirect(sanitizedReferer);
}
```

opencti-platform/opencti-graphql/src/http/httpPlatform.js
```diff
@@ -34,6 +34,20 @@ import { createAuthenticatedContext } from './httpAuthenticatedContext';
...
 
+export function sanitizeReferer(refererToSanitize) {
+  if (!refererToSanitize) return '/';
+  const base = getBaseUrl();
+
+  const isSameOrigin = refererToSanitize.startsWith(`${base}`);
+  const isUrlRelative = refererToSanitize.startsWith('/');
+
+  if (isSameOrigin || isUrlRelative) {
+    return refererToSanitize;
```
xfournet: security boot seems to don't like returning the value directly despite the check. May be we could extract the redirection from the constructed 

we should probably extract the path from an URL object

```
export function sanitizeReferer(refererToSanitize) {
    const base = getBaseUrl();
    const resolvedUrl =  new URL(refererToSanitize, base).toString();
   if(resolvedUrl === base || resolvedUrl.startsWith(`${base}/`) {
       // same domain URL accept the redirection
      return resolvedUrl;
   } else {
      logApp.info('Error auth provider callback : url has been altered', { url: refererToSanitize });
      return '/';
   }   
```

opencti-platform/opencti-graphql/src/http/httpPlatform.js
```diff
@@ -34,6 +34,20 @@ import { createAuthenticatedContext } from './httpAuthenticatedContext';
 import { setCookieError } from './httpUtils';
 import { getChatbotProxy } from './httpChatbotProxy';
 
+export function sanitizeReferer(refererToSanitize) {
+  if (!refererToSanitize) return '/';
+  const base = getBaseUrl();
+
+  const isSameOrigin = refererToSanitize.startsWith(`${base}`);
+  const isUrlRelative = refererToSanitize.startsWith('/');
```
xfournet: in case of subdomain deployment it could be used to redirect on another subdomain

xfournet (approved):

aHenryJard (approved):
Tested locally with SAML and OpenID, all good to me

PR #12554, opened 2025-09-25
[frontend] Fix default dashboard access warning (#11556)

### LINKED ISSUE #11556: Dashboard Sharing Error Alert Not Working as Intended
## Description

When custom dashboards are shared with organizations, the platform will alert you that the configuration is not set up correctly with an Exclamation warning icon. In previous versions of the platform (last one we were on was 6.5.7), this worked as intended to alert us that the configuration needed to be looked at.

However in the current version we are using (6.6.13), this alert is showing even when everything is configured as normal and the user is able to see the dashboard as their primary homepage view. 

## Environment

1. OS (where OpenCTI server runs): { GCP }
2. OpenCTI version: { OpenCTI 6.6.13 }
3. OpenCTI client: { Frontend }
4. Other environment details:

## Reproducible Steps

1. Create a custom dashboard.
2. In the Manage Access Restrictions, share with the Organization of choice
3. Under Security > Organizations > Select the organization of choice and using the Pencil icon, add the Custom dashboard as the Default
4. See the warning signal on the Organization page "More Information".
5. Set up a test account for the user and see that there is no error, the user sees the dashboard.

## Expected Output

We expect this warning message to appear only when there is a misconfiguration to investigate so that we can focus on finding current problems with user configuration rather than double checking that everything performs as intended.

## Actual Output

There is a loss of time setting up test accounts to verify that users have access to the information we intend for them to see.

Image

MTorbay-Filigran:
…roups and organisations

### Proposed changes

* The id of organizations or groups did not match the corresponding authorizedMembers id because there is a _N added at the end. Instead of using the id I changed to use the member_id.
*

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/11556
*

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/settings/groups/Group.tsx
+++ b/opencti-platform/opencti-front/src/private/components/settings/groups/Group.tsx
@@ -139,7 +139,11 @@ const Group = ({ groupData }: { groupData: Group_group$key }) => {
   );
   const canAccessDashboard = (
     group.default_dashboard?.authorizedMembers || []
-  ).some(({ id }) => ['ALL', group.id].includes(id));
+  ).some(({ id }) => {
+    if (id.startsWith('ALL')) return true;
+    const baseId = id.substring(0, id.lastIndexOf('_'));
+    return baseId === group.id;
+  });
 
   return (
     <div className={classes.container} data-testid="group-details-page">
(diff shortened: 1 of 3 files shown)


maelv-filigran:
A few concerns I’d love to discuss to make this even stronger:

Hardcoded Assumption Risk
Could we make this more resilient, like validating the suffix format or documenting the assumption?

Missing Context
The PR description mentions, “Not knowing what this adding is for.” This makes me nervous, if we don’t fully understand why the suffix exists, we might unintentionally break something downstream. 
Have a look a getAuthorizedMembers in authorizedMembers.ts

maelv-filigran (approved):

PR #12559, opened 2025-09-26
[backend] Add impersonate User to Context (#12181)

### LINKED ISSUE #12181: Unable to share public dashboards: GraphQL query requires authentication
## Environment

OpenCTI version: 6.7.15

## Bug

Hello,

It is no longer possible to share public dashboards on OpenCTI.

The lists return the following error:

`An unknown error occurred. Please provide a support package to your administrator or OpenCTI maintainers.`

When analyzing the page requests, we can see that the GraphQL query `useVocabularyCategoryQuery ` now requires authentication.

**Request**:
```json
{"id":"useVocabularyCategoryQuery","query":"query useVocabularyCategoryQuery {\n  vocabularyCategories {\n    key\n    description\n    entity_types\n    fields {\n      key\n      required\n      multiple\n    }\n  }\n}\n","variables":{}}
```

**Response**:
```json
{
    "errors": [
        {
            "message": "You must be logged in to do this.",
            "locations": [
                {
                    "line": 2,
                    "column": 3
                }
            ],
            "path": [
                "vocabularyCategories"
            ],
            "extensions": {
                "code": "AUTH_REQUIRED",
                "data": {
                    "http_status": 401,
                    "genre": "TECHNICAL"
                }
            },
            "name": "AUTH_REQUIRED"
        }
    ],
    "data": null
}
```
The dashboard is correctly configured with the maximum TLP.

Could you please fix this bug that prevents sharing dashboards publicly without an OpenCTI account?

## Screenshots

Image

marieflorescontact:
### Proposed changes

* Add sharedData to Auth user to allow impersonate User

### Related issues

* closes: https://github.com/OpenCTI-Platform/opencti/issues/12181?reload=1?reload=1*

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/http/httpAuthenticatedContext.js
+++ b/opencti-platform/opencti-graphql/src/http/httpAuthenticatedContext.js
@@ -13,25 +13,25 @@ import { batchEntitySettingsByType } from '../modules/entitySetting/entitySettin
 import { batchIsSubAttackPattern } from '../domain/attackPattern';
 import { executionContext, isUserInPlatformOrganization, SYSTEM_USER } from '../utils/access';
 
-export const computeLoaders = (executeContext, user) => {
+export const computeLoaders = (executeContext) => {
   // Generic loaders
   return {
-    relsBatchLoader: batchLoader(batchInternalRels, executeContext, user),
-    creatorsBatchLoader: batchLoader(batchCreators, executeContext, user),
-    creatorBatchLoader: batchLoader(batchCreator, executeContext, user),
-    idsBatchLoader: batchLoader(elBatchIds, executeContext, user),
-    idsBatchLoaderWithCount: batchLoader(elBatchIdsWithRelCount, executeContext, user),
-    markingsBatchLoader: batchLoader(batchMarkingDefinitions, executeContext, user),
+    relsBatchLoader: batchLoader(batchInternalRels, executeContext),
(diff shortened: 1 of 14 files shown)


opencti-platform/opencti-graphql/src/http/httpAuthenticatedContext.js
```diff
@@ -67,6 +67,8 @@ export const createAuthenticatedContext = async (req, res, contextName) => {
   // endregion
   // Return with batch loaders
   executeContext.changeDraftContext = (draftId) => { executeContext.draft_context = draftId; };
-  executeContext.batch = computeLoaders(executeContext, executeContext.user);
+  executeContext.batch = computeLoaders(executeContext);
+  executeContext.sharedData = { impersonatedUser: executeContext.user };
```
lndrtrbn, line 71: I do not understand the goal of this `sharedData` object. Could you give insight of its purpose please?

xfournet: This allows the context to be mutable, making changes made by one resolver visible to another.
Such behavior isn’t possible by directly modifying the root level of the context, since it is shallow-cloned before being passed to a resolver.
An alternative would be to modify it in the same way as changeDraftContext, which always affects the original context and might look cleaner.

However, these approaches introduce security issues, so we need to find another solution.

marieflorescontact: Draft PR 🙂💐

PR #12565, opened 2025-09-26
feat(frontend/backend): handle `app:validation_mode` in LaunchImportDialog (#11603)

### LINKED ISSUE #11603: fix(ingestion): the validation mode chosen during file import is not taken into account
## Description

When importing a file, it is possible to choose the validation mode (Workbench or Draft). However, if in the configuration of my file import connector (CONNECTOR_VALIDATE_BEFORE_IMPORT), or my platform configuration if it is for CSV mapper (app:validation_mode), I define that I want nothing, I will get nothing. 

This is problematic from a user experience perspective. Without knowing this configuration, we are unable to understand why it behaves this way.

Image

ps: feel free to caterogize it as "feature" if you think it is more appropriate.

## Environment

6.7.3

## Reproducible Steps

1. Import a CSV file
2. Launch an import using the CSV mapper
-> You can select a "Validation mode" 

## Expected Output

If I select "Workbench," I expect to see a workbench.

We should:
- Either stop putting the configuration in the platform configuration (for CSV mappers), or the connector configuration (for file import connectors), and let the user choose each time (Workbench|Draft|Nothing)
- Or, when you have a specific configuration, not let the user choose anything else.

delemaf:
### Proposed changes

*  We want to disable the validation_mode selection when `app:validation_mode` is specified in config file

### Related issues

* #11603 

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/lang/front/en.json
+++ b/opencti-platform/opencti-front/lang/front/en.json
@@ -4017,6 +4017,7 @@
   "Validate this workbench": "Validate this workbench",
   "Validated": "Validated",
   "Validation mode": "Validation mode",
+  "Validation mode has already been setted in platform configuration": "Validation mode has already been setted in platform configuration",
   "Validation progress": "Validation progress",
   "Validator group membership:": "Validator group membership:",
   "Validator membership:": "Validator membership:",
--- a/opencti-platform/opencti-front/lang/front/de.json
+++ b/opencti-platform/opencti-front/lang/front/de.json
@@ -4017,6 +4017,7 @@
   "Validate this workbench": "Diese Workbench validieren",
   "Validated": "Validiert",
   "Validation mode": "Validierungsmodus",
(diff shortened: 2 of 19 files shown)


opencti-platform/opencti-front/lang/front/en.json
```diff
@@ -4017,6 +4017,7 @@
   "Validate this workbench": "Validate this workbench",
   "Validated": "Validated",
   "Validation mode": "Validation mode",
+  "Validation mode has already been setted in platform configuration": "Validation mode has already been setted in platform configuration",
```
JeremyCloarec, line 4020: small typo, it's `set` and not `setted`

JeremyCloarec:
I think there is a misunderstanding from the initial reported issue. The app:validation_mode is only relevant for the import csv connector, and not relevant for the other import file connectors: we would need to know the CONNECTOR_VALIDATE_BEFORE_IMPORT configuration of the connector.
We'll need to sync on this to know how we want to tackle the issue

PR #12579, opened 2025-09-29
[backend] Cannot delete an element with a background task  (#12573)

### LINKED ISSUE #12573: Cannot delete an element with a background task
## Description

Cannot delete an element with a background task

## Reproducible Steps

1. Go to reports
2. Check the box of one report
3. Delete

=> 

Image

ValentinBouzinFiligran:
### Proposed changes

* can delete SCO type on delete element by id

### Related issues

* #12573 

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/database/middleware.js
+++ b/opencti-platform/opencti-graphql/src/database/middleware.js
@@ -3455,10 +3455,12 @@ export const internalDeleteElementById = async (context, user, id, type, opts = 
     throw AlreadyDeletedError({ id });
   }
 
-  const isAbstractAndNotInternalObject = element.parent_types.includes(type) && !element.parent_types.includes(ABSTRACT_INTERNAL_OBJECT);
+  if (!(element.entity_type === type || element.relationship_type === type)) {
+    const isAbstractAndNotInternalObject = element.parent_types.includes(type) && !element.parent_types.includes(ABSTRACT_INTERNAL_OBJECT);
 
-  if (!(element.entity_type === type || element.relationship_type === type || isAbstractAndNotInternalObject)) {
-    throw FunctionalError('Cant find element type for deletion', { id, type });
+    if (element.parent_types && !isAbstractAndNotInternalObject) {
+      throw FunctionalError('Cant find element type for deletion', { id, type });
+    }
   }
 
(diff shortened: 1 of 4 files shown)


ValentinBouzinFiligran:
if (!(element.entity_type === type || element.relationship_type === type )) {

I can see that the entity_type is "Report" and the type is "Stix-Core-Object"
both are ok but it does not match the condition

image

CelineSebe:
Works well !

CelineSebe (dismissed):

aHenryJard (dismissed):
Please wait there is also some migration to fix
1650287551439-remove_sector_locations.js for example; can you check migration files please ?

opencti-platform/opencti-graphql/src/database/middleware.js
```diff
@@ -3454,7 +3454,7 @@ export const internalDeleteElementById = async (context, user, id, type, opts =
     throw AlreadyDeletedError({ id });
   }
 
-  if (!(element.entity_type === type || element.relationship_type === type)) {
+  if (!(element.entity_type === type || element.relationship_type === type || type === ABSTRACT_STIX_RELATIONSHIP)) {
```
richard-julien: I dont understand this addition, can you explain?

aHenryJard (dismissed):
Not sure to understand CI failure; Looks good to me

richard-julien:
@ValentinBouzinFiligran CI tests failure must be investigated.
Something is definitely broken in deletion

> should create and delete retention rules

opencti-platform/opencti-graphql/src/database/middleware.js
ValentinBouzinFiligran: this condition has changed

aHenryJard (approved):
Good for me

opencti-platform/opencti-graphql/src/database/middleware.js
```diff
@@ -3455,7 +3456,11 @@ export const internalDeleteElementById = async (context, user, id, type, opts =
   }
 
   if (!(element.entity_type === type || element.relationship_type === type)) {
```
SouadHadjiat: maybe add type in the opts l3452 in storeLoadByIdWithRefs ? so that we load the element with the right type

Issue #12569, opened 2025-09-26
PIR Manager is displayed as disabled on all instances


SamuelHassine:
## Description

PIR Manager is displayed as disabled on all instances

<img width="1050" height="266" alt="Image" src="https://github.com/user-attachments/assets/ace95a9c-1afa-413c-a836-eb0ca6584f6f" />

All instances impacted (prerelease, testing, prod, etc.)

PR #12582, opened 2025-09-29
[backend] Pir manager status display (#12569)

Archidoit:
### Proposed changes
PIR manager status display in Settings > Parameters should be ENABLED if : EE and enabled in config

### Related issues
https://github.com/OpenCTI-Platform/opencti/issues/12569


SarahBocognano (approved):
Tested ✅

Issue #12570, opened 2025-09-26
When you put your license key and you've buttons to disable in settings double scroll bars displayed on right


SamuelHassine:
## Description

When you put your license key and you've buttons to disable, double scroll bars displayed on right

<img width="3838" height="2054" alt="Image" src="https://github.com/user-attachments/assets/229eeb4f-faa3-4f8e-b11a-9d8490ab099f" />

PR #12589, opened 2025-09-30
[frontend] Fix double scrollbar on settings page (#12570)

SarahBocognano:
### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/12570


JeremyCloarec:
We no longer have two scrollbars, but the remaining scrollbar doesn't work properly: we can't scroll to the bottom of the page with it:
image

Not sure how to fix though :/

Gwendoline-FAVRE-FELIX (approved):
It works.

Issue #12246, opened 2025-09-03
Can't create support package while in draft mode


leitosama:
## Description

If you got a error while you were working in Draft mode you couldn't create a support package.

## Environment

1. OS (where OpenCTI server runs): Fedora (Linux)
2. OpenCTI version: OpenCTI 6.7.16
3. OpenCTI client: frontend
4. Other environment details:

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Get a error while working in Draft mode ( #12245 )
2. Click on the "support package" link
3. Click on the "Generate support package" button

## Expected Output

Platform started generating support package.

## Actual Output

"Cannot index unsupported element in draft context"
```
{"category":"APP","cause":{"attributes":{"genre":"BUSINESS","http_status":500},"code":"UNSUPPORTED_ERROR","message":"Cannot index unsupported element in draft context","name":"UNSUPPORTED_ERROR","stack":"GraphQLError: Cannot index unsupported element in draft context\n    at error (/opt/opencti/build/back.js:1844:2273)\n    at UnsupportedError (/opt/opencti/build/back.js:1844:4480)\n    at validateElementsToIndex (/opt/opencti/build/back.js:2062:908)\n    at elIndexElements (/opt/opencti/build/back.js:2062:1040)\n    at indexCreatedElement (/opt/opencti/build/back.js:2636:87035)\n    at createEntityRaw (/opt/opencti/build/back.js:2636:166725)\n    at async createEntity (/opt/opencti/build/back.js:2636:167441)\n    at async createInternalObject (/opt/opencti/build/back.js:2636:593132)\n    at async prepareNewSupportPackage (/opt/opencti/build/back.js:3709:53868)\n    at async addSupportPackage (/opt/opencti/build/back.js:3709:54316)"},"level":"error","message":"Cannot index unsupported element in draft context","operation":"SupportPackagesMutation","size":60,"source":"backend","time":25,"timestamp":"2025-09-03T10:42:10.679Z","type":"WRITE_ERROR","user_id":"88ec0c6a-13ce-5e39-b486-354fe4a7084f","version":"6.7.16"}
```

## Additional information

Logs from test instance: [3c9b3533-8bbd-42d1-a844-aef1d8137af7.zip](https://github.com/user-attachments/files/22116458/3c9b3533-8bbd-42d1-a844-aef1d8137af7.zip)

SarahBocognano:
[...]

PR #12595, opened 2025-09-30
[frontend] Can't create support package while in draft mode (#12246)

SarahBocognano:
### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/12246

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/Error.jsx
+++ b/opencti-platform/opencti-front/src/private/components/Error.jsx
@@ -30,7 +30,7 @@ export const SimpleError = () => {
             '',
             {
               id: 'An unknown error occurred. Please provide a support package to your administrator or OpenCTI maintainers',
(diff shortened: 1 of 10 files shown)


opencti-platform/opencti-front/src/private/components/Error.jsx
```diff
@@ -27,7 +30,7 @@ export const SimpleError = () => {
             '',
             {
               id: 'An unknown error occurred. Please provide a support package to your administrator or OpenCTI maintainers',
-              values: { link_support_package: <Link to="/dashboard/settings/experience">{t_i18n('support package')}</Link> },
+              values: { link_support_package: disabledInDraft ? t_i18n('support package') : <Link to="/dashboard/settings/experience">{t_i18n('support package')}</Link> },
```
JeremyCloarec: Maybe the message when in draft could be something like: 'First move out, then generate a support package'? Otherwise the user could be a bit lost I think

JeremyCloarec (approved):

Issue #12188, opened 2025-08-29
Draft operation filter visible in lists where it should not


Archidoit:
## Description
In the markings list, there is a filter 'Draft operation', which should not be there.

<img width="1865" height="551" alt="Image" src="https://github.com/user-attachments/assets/db727cb4-6d07-445b-b9c0-8f9cc6858455" />

--> same in investigation list and custom dashboards list

SarahBocognano:
Does the available filters in a view are related to the view of the entity ? For example if I go to a report, are the available filters are only for a Report entity type or it's all filters available in the platform ?
@Archidoit @nino-filigran

Archidoit:
The filters available in an entities list depend on the entity type. @SarahBocognano

PR #12597, opened 2025-09-30
[backend] Draft operation filter visible in lists where it should not (#12188)

SarahBocognano:
### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/12188


Archidoit:
The Draft operation filter is still visible in investigations, custom dashboards, public dashboards and playbooks.

image

Archidoit (approved):

PR #12600, opened 2025-10-01
[backend/frontend] Prevent browser crash with invalid position coordinates (#11841)

### LINKED ISSUE #11841: One of the location -> position crashes my browser.
## Description

One of the location -> position crashes my browser.

## Environment

Chrome

## Reproducible Steps

https://testing.octi.staging.filigran.io/dashboard/locations/positions/3dc908ec-c98d-483c-be5f-535d865a9fa8

## Screenshots (optional)

https://github.com/user-attachments/assets/73f181c6-59ff-4e39-9d0f-2d7fe5d46795

maelv-filigran:
### Proposed changes

* Add coordinate validation in Position component to prevent invalid values from crashing the map
* Add defensive validation in LocationMiniMap to handle edge cases  
* Add Yup validation in PositionEditionOverview to prevent invalid coordinate input
* Add backend validation to reject invalid coordinates at API level

### Related issues

* #11841

### Checklist

### Further comments

This fix implements multi-layer validation to prevent browser crashes caused by invalid position coordinates. The issue was occurring when positions had coordinates outside valid ranges (latitude: -90 to 90, longitude: -180 to 180). The map component (Leaflet) would crash trying to render these invalid coordinates.

#### Steps to reproduce the bug (before fix):
1. On testing environment, create a location with valid coordinates (e.g., [1,1])
2. Navigate to the location page and edit the coordinates to invalid values (e.g., [1,1000])
3. Refresh the page: the browser crashes with 100% CPU usage

#### Solution:
The fix validates coordinates at multiple levels:
1. **Display level** - fallback to default coordinates (Paris) if invalid
2. **Input level** - prevent users from entering invalid values
3. **Backend level** - reject invalid data at API level

This ensures existing invalid data won't crash the browser and prevents new invalid data from being created.

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/locations/positions/Position.tsx
+++ b/opencti-platform/opencti-front/src/private/components/locations/positions/Position.tsx
@@ -23,6 +23,9 @@ const useStyles = makeStyles(() => ({
   },
 }));
 
+// Default coordinates (Paris) to use when position has invalid or missing coordinates
+const DEFAULT_CENTER_COORDINATES: [number, number] = [48.8566969, 2.3514616];
+
 interface PositionComponentProps {
   position: Position_position$data;
 }
@@ -39,6 +42,23 @@ const PositionComponent: FunctionComponent<PositionComponentProps> = ({
       relationship_type: ['located-at'],
     },
   );
+
+  // Validate coordinates to prevent map crash with invalid values
+  const isValidLatitude = (lat: number | null | undefined): boolean => {
+    return lat !== null && lat !== undefined && lat >= -90 && lat <= 90;
+  };
+
+  const isValidLongitude = (lng: number | null | undefined): boolean => {
+    return lng !== null && lng !== undefined && lng >= -180 && lng <= 180;
+  };
+
+  const getValidatedCenter = (): [number, number] => {
+    if (isValidLatitude(position.latitude) && isValidLongitude(position.longitude)) {
+      return [position.latitude as number, position.longitude as number];
+    }
+    // Return default coordinates if invalid or missing
+    return DEFAULT_CENTER_COORDINATES;
+  };
   return (
     <div data-testid="position-details-page">
       <Grid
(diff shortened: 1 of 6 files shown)


opencti-platform/opencti-front/src/private/components/locations/positions/Position.tsx
```diff
@@ -39,6 +42,23 @@ const PositionComponent: FunctionComponent<PositionComponentProps> = ({
       relationship_type: ['located-at'],
     },
   );
+
+  // Validate coordinates to prevent map crash with invalid values
+  const isValidLatitude = (lat: number | null | undefined): boolean => {
```
esrevi: I think it could be great to extract these functions in a utils file, such as position.utils.ts for example, and use them in LocationMiniMap and Position components :)

opencti-platform/opencti-front/src/private/components/locations/positions/Position.tsx
```diff
@@ -39,6 +42,23 @@ const PositionComponent: FunctionComponent<PositionComponentProps> = ({
...
+    return lat !== null && lat !== undefined && lat >= -90 && lat <= 90;
+  };
+
+  const isValidLongitude = (lng: number | null | undefined): boolean => {
+    return lng !== null && lng !== undefined && lng >= -180 && lng <= 180;
+  };
+
+  const getValidatedCenter = (): [number, number] => {
```
esrevi: prefer pass position in param instead of using the state directly, because it could have issues if position change for example, or do it in a useEffect with position as dependency

opencti-platform/opencti-front/src/private/components/locations/positions/Position.tsx
maelv-filigran: got it!

opencti-platform/opencti-front/src/private/components/locations/positions/Position.tsx
maelv-filigran: you're right :)

esrevi (approved):
tested locally, lgtm

PR #12610, opened 2025-10-01
Update Yarn version and configuration (#12609)

### LINKED ISSUE #12609: Update Yarn version and configuration
## Description

Yarn version need to be updated and the configuration has to be updated to conform with current best practice and harden the security

xfournet:
### Proposed changes

* use corepack
* update to v4.10.3
* use global cache
* configuration hardening: whitelist dependencies that are allowed to execute scripts , prevent to install fresh dependencies releases, enforce yarn release integrity

### Related issues
* Fix #12609

### Checklist


.drone.yml
```diff
@@ -33,6 +33,8 @@ steps:
       - pip install -r requirements.txt
       - pip install -e .[dev,doc]
       - cd "$DRONE_WORKSPACE/opencti-platform/opencti-graphql"
+      - npm install -g corepack
+      - cp $DRONE_WORKSPACE/.yarnrc.yml .
```
aHenryJard, line 37: Why do we need to cp the .yarnrc.yml, does it means that locally we will need to cp this file before yarn install too ?

aHenryJard (commented):
You may need to update this page https://docs.opencti.io/latest/development/platform/#backend-api

xfournet:
> You may need to update this page https://docs.opencti.io/latest/development/platform/#backend-api

Documentation PR is available: https://github.com/OpenCTI-Platform/docs/pull/358

.drone.yml
xfournet: The drone CI steps are doing crazy thing with the repo, and and by the end the .yarnrc.yml file is no longer present in the parent directory of the opencti-graphql folder. I ended up using this solution. Not very clean, but the issue root cause comes from the overall approach used in these steps .
Locally there is no problem and no manual copy is required.

aHenryJard: ah ok I see, yes we had some concurrency issues on build.... I think we should build a docker image of opencti to run all opencti required for tests on drone instead of building it 10 time. But that's another story.

aHenryJard (approved):
Tested locally and run OpenCTI for development without any issue. Looks good to me.

Issue #12572, opened 2025-09-27
When using the connector manager, connector names are not correct


SamuelHassine:
## Description

When using the connector manager, connector names are not correct:

1. Not respecting the default connector name with the right brand
2. Frontend logic on the sanitizing

<img width="1345" height="868" alt="Image" src="https://github.com/user-attachments/assets/4da18c75-eb83-443e-8a87-f47b55e7fcc8" />

PR #12614, opened 2025-10-02
[frontend] revamp connector managed details and naming in the list (#12572)

esrevi:
### Proposed changes

* In the list, show display name of the managed connector
* In the details: 
* display connector in basic information with link to the catalog connector
* display the connector identifier name with copy/paste action
* migrate identifier field, it becomes the container_name, and the container_name is now name

### Related issues

* #12572 

### Checklist


opencti-platform/opencti-graphql/src/migrations/1759507412340-update-connector-name.js
```diff
@@ -0,0 +1,56 @@
...
+          internal_id: connector.internal_id,
+          name: connector.name,
+          title: connector.title,
+        });
+
+        const internalId = connector.internal_id;
+
+        const containerNameSuffix = internalId.substring(0, 8);
+        const newName = connector.title || connector.manager_contract_excerpt?.title || connector.name;
+
```
esrevi: maybe we should add a check if no name at all ?

if (!newName) throw new error('No name found for connector ${connector.id}`)

opencti-platform/opencti-graphql/src/migrations/1759507412340-update-connector-name.js
```diff
@@ -0,0 +1,56 @@
...
+  const context = executionContext('migration');
+
+  const connectorsList = await connectors(context, SYSTEM_USER);
+
+  const managedConnectors = connectorsList.filter((connector) => connector.is_managed);
+  logMigration.info(`${message} > found ${managedConnectors.length} managed connectors to update`);
+
+  let successCount = 0;
+
+  await Promise.all(
```
esrevi: I know we wrote at first with promise.all, but now, I think we should change Promise.all to Promise.allSettled, because, the first failure will stop all, instead of allSettled that will catch the error, but will keep process the other promises.

Thus, we could do like this:
```
const results = Promise.allSettled(
   managedConnected.map(async (connector) =>  {
   // No need try/catch, allSettled catches everything
     logMigration.info(`${message} > processing connector ${connector.id}`, {
          internal_id: connector.internal_id,
          name: connector.name,
          title: connector.title,
        });
        ....
        await patchAttribute(context, SYSTEM_USER, connector.id, ENTITY_TYPE_CONNECTOR, patch);
        return connector.id
   }
)

const connectorSuccess = results.filter(r => r.status === 'fullfilled')
logMigration.info(`${message} > completed - ${connectorSuccess.length} connectors updated`);

const connectorFailures = results.filter(r => r.stattus === 'rejected')

if (connectorFailures.length > 0) {
logMigration.info(`${message} > failures - ${connectorFailures.length} connectors failed to update`);
}

```

wdyt ?

opencti-platform/opencti-graphql/src/migrations/1759507412340-update-connector-name.js
maelv-filigran: good idea !

PR #12620, opened 2025-10-02
[backend] add stix filtering for "pattern" key (#10459)

### LINKED ISSUE #10459: Pattern matching for indicator in playbook or inbound/outbound stream
## Use case

Be able to look for specific pattern in the indicator.
For example look for a specific domain from an email address before processing it

## Current Workaround

No workaround

## Proposed Solution

Add a pattern filter (contains/starts with/equal/... operator) in the playbook and stream/taxii/csv

labo-flg:
### Proposed changes

* Add pattern in filters possible
* Add tests using pattern

### Related issues

* #10459


opencti-platform/opencti-graphql/tests/01-unit/database/stix-filtering-testers-test.ts
```diff
@@ -381,27 +381,72 @@ describe('Stix filter testers', () => {
     });
   });
 
-  describe('by Pattern (key=pattern_type)', () => {
+  describe('by Pattern Type (key=pattern_type)', () => {
+    const stixWithPatternType = stixIndicators[0];
+    const stixWithoutPatternType = stixReports[0];
```
lndrtrbn, line 386: ~This filter on report can never work because there is no pattern attribute on it. I think the test would be more representative by using an other indicator that does not match the filter~

lndrtrbn: My bad, you already have the different cases

lndrtrbn (approved):

PR #12624, opened 2025-10-02
[worker] Cleanup codebase (#12623)

### LINKED ISSUE #12623: Cleanup worker code
## Description

Many inconsistencies and historical updates have lead to a difficult to read and maintain codebase on worker. It will make it harder to evolve.

xfournet:
### Proposed changes

* See each step in the commit history

### Related issues
* Fix #12623 

### Checklist

### DIFF
--- a/opencti-worker/src/worker.py
+++ b/opencti-worker/src/worker.py
@@ -611,13 +611,6 @@ class Worker:  # pylint: disable=too-few-public-methods, too-many-instance-attri
             custom_headers=self.opencti_api_custom_headers,
         )
         self.worker_logger = self.api.logger_class("worker")
-        # Initialize variables
-        self.connectors: List[Any] = []
-        self.queues: List[Any] = []
-        self.execution_pool = ThreadPoolExecutor(max_workers=self.opencti_pool_size)
-        self.listen_api_execution_pool = ThreadPoolExecutor(
-            max_workers=self.listen_pool_size
-        )
 
     def stop(self) -> None:
         for thread in self.listen_api_threads:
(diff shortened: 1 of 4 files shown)


opencti-worker/src/worker.py
```diff
@@ -634,27 +255,39 @@ def __post_init__(self) -> None:
...
-            self.listen_api_threads[thread].join()
-        for thread in self.consumer_threads:
-            self.consumer_threads[thread].stop()
-            self.consumer_threads[thread].join()
+        # Initiate stop for all consumers
+        for consumer in self.consumers.values():
+            consumer.stop(False)
+        # Wait for all consumers to stop
+        for consumer in self.consumers.values():
+            consumer.stop(True)
```
JeremyCloarec: I don't get exactly why we call the stop method twice. Is it to first stop all the consumers without waiting for them to stop synchronously, and then call stop(True) to check that every consumer is stopped synchronously?

If this is t he case, I think it could be useful to create two different method to make it easier to understand. Something like async_stop and blocking_stop, wdyt?

opencti-worker/src/worker.py
```diff
@@ -58,475 +46,104 @@
...
     execution_pool: ThreadPoolExecutor
-    connector: Dict[str, Any] = field(hash=False)
-    config: Dict[str, Any] = field(hash=False)
-    listen_api_ssl_verify: bool
-    listen_api_http_proxy: str
-    listen_api_https_proxy: str
-    log_level: str = "info"
-    json_logging: bool = True
-    _is_interrupted: bool = False
+    handle_message: Callable[[str], Literal["ack", "nack", "requeue"]]
```
JeremyCloarec: I think splitting this class into an other separate file could be great also, wdyt?

opencti-worker/src/worker.py
xfournet: Yes, it's a change from the previous implementation, which stopped and waited for each thread one by one.
Now, all threads are signaled to stop at once, and then we wait for them to finish.

xfournet: > If this is t he case, I think it could be useful to create two different method to make it easier to understand. Something like async_stop and blocking_stop, wdyt?

i can split in two methods, i'm thinking about the names

opencti-worker/src/worker.py
xfournet: indeed it would be interresting, i will test it

opencti-worker/src/worker.py
xfournet: i see in Python 3 we can force to specify the argument name=value parameter call style, so it could help to understand the intent of the parameter. In that case we could also be more explicit on the argument name, eg wait -> wait_for_completion

```python
    def stop(self, *, wait_for_completion: bool) -> None:
        self.should_stop = True
        if wait_for_completion:
            self.thread.join()
```

That would look likes this when using it:

```python
        # Initiate stop for all consumers
        for consumer in self.consumers.values():
            consumer.stop(wait_for_completion=False)
        # Wait for all consumers to stop
        for consumer in self.consumers.values():
            consumer.stop(wait_for_completion=True)
```

If we still prefer to split the method
* async_stop i'm not sure about 'async' in the name, it looks confusing since there is no callback from this call. I would propose request_stop()
* blocking_stop : i would change it to have clearer intent eg wait_for_completion() ?

what is your preference ? or other propositions in regards of this feedback ?

opencti-worker/src/worker.py
xfournet: Done

opencti-worker/src/worker.py
JeremyCloarec: I like your proposition for the names of the split functions 👍 Let's go for request_stop() and wait_for_completion()

xfournet: done

JeremyCloarec (approved):

PR #12628, opened 2025-10-02
[backend] Introduce doYield method (#12627)

### LINKED ISSUE #12627: [backend] Introduce doYield utility to prevent event loop blocking
## Description

There is a lot of duplicated code for handling long blocking loops and preventing them from blocking. This should be refactored to make it easier to reuse wherever needed.

xfournet:
### Proposed changes

* Implement a `doYield` function to be used in cases where a long task could block the event loop.
* Replace current pattern with this method

### Related issues
* Fix #12627

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/database/engine.js
+++ b/opencti-platform/opencti-graphql/src/database/engine.js
@@ -24,7 +24,6 @@ import {
   isEmptyField,
   isInferredIndex,
   isNotEmptyField,
-  MAX_EVENT_LOOP_PROCESSING_TIME,
   offsetToCursor,
   pascalize,
   READ_DATA_INDICES,
@@ -222,6 +221,7 @@ import { getPirWithAccessCheck } from '../modules/pir/pir-checkPirAccess';
 import { asyncFilter, asyncMap, uniqAsyncMap } from '../utils/data-processing';
 import { ENTITY_TYPE_PIR } from '../modules/pir/pir-types';
 import { isMetricsName } from '../modules/metrics/metrics-utils';
+import { doYield } from '../utils/eventloop-utils';
 
 const ELK_ENGINE = 'elk';
 const OPENSEARCH_ENGINE = 'opensearch';
@@ -1575,8 +1575,8 @@ const elDataConverter = (esHit) => {
 export const elConvertHitsToMap = async (elements, opts) => {
   const { mapWithAllIds = false } = opts;
   const convertedHitsMap = {};
-  let startProcessingTime = new Date().getTime();
   for (let n = 0; n < elements.length; n += 1) {
+    await doYield();
     const element = elements[n];
     convertedHitsMap[element.internal_id] = element;
     if (mapWithAllIds) {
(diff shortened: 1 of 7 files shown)


opencti-platform/opencti-graphql/src/database/engine.js
```diff
@@ -4496,15 +4461,22 @@ const createDeleteOperationElement = async (context, user, mainElement, deletedE
...
     if (Array.isArray(value)) { // Array of Date, objects, string or number
       const preparedArray = [];
-      let innerProcessingTime = new Date().getTime();
-      let extendLoopSplit = 0;
+      let yieldCount = 0;
       for (let valueIndex = 0; valueIndex < value.length; valueIndex += 1) {
+        if (await doYield()) {
+          // If we extend the preparation 5 times, log a warn
+          // It will help to understand what kind of key have so many elements
+          if (yieldCount === 5) {
```
aHenryJard, line 4507: I wonder about having a configuration for this number to be able to tune it, are you confident about being 5 the right value ?

aHenryJard (commented):
Maybe target release/current to validate in "production context" before targeting master ?

xfournet:
> Maybe target release/current to validate in "production context" before targeting master ?

indeed it was my intention, i created the branch from release/current, but didn't notice that release/current has no ahead commit from master currently so i need to specify it in github. It should be ok now

opencti-platform/opencti-graphql/src/database/engine.js
xfournet: i just take the value that was hardcoded at line 4526 in the previous version

opencti-platform/opencti-graphql/src/utils/eventloop-utils.ts
```diff
@@ -0,0 +1,17 @@
+const MAX_EVENT_LOOP_PROCESSING_TIME = 50;
+
+let nextYield: number | undefined;
+export const doYield = async () => {
+  if (nextYield !== undefined) {
+    if (Date.now() > nextYield) {
+      nextYield = undefined;
+      await new Promise((resolve) => {
+        setTimeout(resolve, 0);
```
SouadHadjiat: what is the difference between setImmediate that was used before and setTimeout used here ?
also won't we have issues with a nextYield globally defined ? what if the function calling doYield is executed in less than 50ms, then the next one is executed some time after, it will be considered as a long process ?

xfournet: Both approaches work here, but setImmediate is probably clearer in intent, so I can switch back to it.

The yield budget is now tracked globally, not per function call, so it maximizes the chance of freeing the event loop once the 50 ms blocking duration is exceeded.

There is one edge case where two widely separated calls to doYield could cause the second one to yield unnecessarily. I’m adding an extra condition to prevent that.

xfournet: I revert to using setImmediate (using the promise-based version). I also won’t add extra checks to prevent unnecessary yields, since that could cause missed yields, which would be a bigger issue. Unnecessary yields only happen when load is low, so it isn’t an issue.

aHenryJard (approved):
Tested locally, looks good to me. But I think you need to rebase.

xfournet:
> Tested locally, looks good to me. But I think you need to rebase.

yes i need to rebase indeed
is it ok for you to target master rather than current/release ?

PR #12632, opened 2025-10-03
[backend/worker] handle too large items (#12816)

### LINKED ISSUE #12816: Add a way to send objects from bundles with too many refs to a dead letter queue
## Use case

Some objects cause problems on ingestion when they are linked to too many other objects in their input. 
To solve this, we need to add a way to send objects from bundles with too many refs to a dead letter queue

JeremyCloarec:
### Proposed changes

* add new internal queue `too-large-bundle`
* add `objects_max_refs` config to worker.py
* in worker, if item has too many dependencies, it is sent to the `too-large-bundle` queue instead of being ingested

client PR: https://github.com/OpenCTI-Platform/client-python/pull/982
doc PR: https://github.com/OpenCTI-Platform/docs/pull/360

### Related issues

* #12816
*

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/database/rabbitmq.js
+++ b/opencti-platform/opencti-graphql/src/database/rabbitmq.js
@@ -264,10 +264,11 @@ const CONNECTOR_QUEUE_PLAYBOOK = { id: 'playbook', name: 'Internal playbook mana
 const CONNECTOR_QUEUE_SYNC = { id: 'sync', name: 'Internal sync manager', type: 'internal', scope: 'sync' };
 /** @deprecated [>=6.3 & <6.6]. Remove and add migration to remove the queues. */
 const DEPRECATED_INTERNAL_QUEUES = [CONNECTOR_QUEUE_PLAYBOOK, CONNECTOR_QUEUE_SYNC];
+const CONNECTOR_QUEUE_BUNDLES_TOO_LARGE = { id: 'too-large-bundle', name: 'Bundle too large for ingestion', type: 'internal', scope: 'dead letter' };
 // endregion
 export const getInternalQueues = () => {
   const backgroundTaskConnectorQueues = getInternalBackgroundTaskQueues();
-  return [...DEPRECATED_INTERNAL_QUEUES, ...backgroundTaskConnectorQueues];
+  return [CONNECTOR_QUEUE_BUNDLES_TOO_LARGE, ...DEPRECATED_INTERNAL_QUEUES, ...backgroundTaskConnectorQueues];
 };
 
 export const initializeInternalQueues = async () => {
--- a/opencti-platform/opencti-graphql/src/database/rabbitmq.js
+++ b/opencti-platform/opencti-graphql/src/database/rabbitmq.js
@@ -210,6 +210,7 @@ export const connectorConfig = (id, listen_callback_uri = undefined) => ({
   listen_routing: listenRouting(id),
   listen_exchange: CONNECTOR_EXCHANGE,
   listen_callback_uri,
+  dead_letter_routing: listenRouting(CONNECTOR_QUEUE_BUNDLES_TOO_LARGE.id),
 });
 
 export const listenRouting = (connectorId) => `${RABBIT_QUEUE_PREFIX}listen_routing_${connectorId}`;
--- a/opencti-platform/opencti-graphql/tests/02-integration/01-database/rabbitmq-test.js
+++ b/opencti-platform/opencti-graphql/tests/02-integration/01-database/rabbitmq-test.js
@@ -20,7 +20,7 @@ describe('Rabbit connector management', () => {
   });
   it('should connector queues available', async () => {
     const data = await metrics(testContext, ADMIN_USER);
-    expect(data.queues.length).toEqual(18);
+    expect(data.queues.length).toEqual(20);
     const aggregationMap = new Map(data.queues.map((q) => [q.name, q]));
     expect(aggregationMap.get(`${RABBIT_QUEUE_PREFIX}listen_${testConnectorId}`)).toBeDefined();
     expect(aggregationMap.get(`${RABBIT_QUEUE_PREFIX}push_${testConnectorId}`)).toBeDefined();
--- a/opencti-worker/src/config.yml.sample
+++ b/opencti-worker/src/config.yml.sample
@@ -7,4 +7,4 @@ opencti:
 worker:
   log_level: 'info'
   telemetry_enabled: false
-  objects_max_deps: 0
+  objects_max_refs: 0
--- a/opencti-worker/src/config.yml.sample
+++ b/opencti-worker/src/config.yml.sample
@@ -7,3 +7,4 @@ opencti:
 worker:
   log_level: 'info'
   telemetry_enabled: false
+  objects_max_deps: 0
--- a/opencti-worker/src/push_handler.py
+++ b/opencti-worker/src/push_handler.py
@@ -22,6 +22,7 @@ class PushHandler:  # pylint: disable=too-many-instance-attributes
     push_exchange: str
     listen_exchange: str
     push_routing: str
+    dead_letter_routing: str
     pika_parameters: pika.ConnectionParameters
     bundles_global_counter: Any
     bundles_processing_time_gauge: Any
@@ -35,10 +36,6 @@ class PushHandler:  # pylint: disable=too-many-instance-attributes
             json_logging=self.json_logging,
             ssl_verify=self.ssl_verify,
         )
-        # Small hack to handle any rabbit prefix that could have been configured
-        self.listen_too_large_routing = self.push_routing.replace(
-            "push_routing", "listen_routing"
-        ).replace(self.connector_id, "too-large-bundle")
 
     def send_bundle_to_specific_queue(
         self,
@@ -111,7 +108,7 @@ class PushHandler:  # pylint: disable=too-many-instance-attributes
                     )
                     if len(too_large_items_bundles) > 0:
                         with pika.BlockingConnection(
-                                self.pika_parameters
+                            self.pika_parameters
                         ) as push_pika_connection:
                             with push_pika_connection.channel() as push_channel:
                                 try:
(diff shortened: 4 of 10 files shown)


opencti-worker/src/config.yml.sample
```diff
@@ -6,4 +6,5 @@ opencti:
 
 worker:
   log_level: 'info'
-  telemetry_enabled: false
\ No newline at end of file
+  telemetry_enabled: false
+  objects_max_deps: 0
```
SouadHadjiat: this config needs to be updated on our prod (and added in documentation)

opencti-platform/opencti-graphql/src/database/rabbitmq.js
```diff
@@ -264,10 +264,11 @@ const CONNECTOR_QUEUE_PLAYBOOK = { id: 'playbook', name: 'Internal playbook mana
 const CONNECTOR_QUEUE_SYNC = { id: 'sync', name: 'Internal sync manager', type: 'internal', scope: 'sync' };
 /** @deprecated [>=6.3 & <6.6]. Remove and add migration to remove the queues. */
 const DEPRECATED_INTERNAL_QUEUES = [CONNECTOR_QUEUE_PLAYBOOK, CONNECTOR_QUEUE_SYNC];
+const CONNECTOR_QUEUE_BUNDLES_TOO_LARGE = { id: 'too-large-bundle', name: 'Bundle too large for ingestion', type: 'internal', scope: 'dead letter' };
```
xfournet, line 268: too large bundle are all going to the same queue whatever their original queue ?

opencti-worker/src/config.yml.sample
```diff
@@ -6,4 +6,5 @@ opencti:
 
 worker:
   log_level: 'info'
-  telemetry_enabled: false
\ No newline at end of file
+  telemetry_enabled: false
+  objects_max_refs: 0
```
xfournet, line 10: did you manage to find a good default value to be used in Filigran SaaS ?

opencti-worker/src/worker.py
```diff
@@ -596,6 +654,13 @@ def __post_init__(self) -> None:
             config,
             default="",
         )
+        self.stix_object_max_refs = get_config_variable(
```
xfournet: both object_max_refs and objects_max_refs names are used, we should choose one

opencti-platform/opencti-graphql/tests/02-integration/01-database/rabbitmq-test.js
```diff
@@ -20,7 +20,7 @@ describe('Rabbit connector management', () => {
   });
   it('should connector queues available', async () => {
     const data = await metrics(testContext, ADMIN_USER);
-    expect(data.queues.length).toEqual(18);
+    expect(data.queues.length).toEqual(20);
```
xfournet, line 23: we added one queue, why is it increasing from 18 to 20 queues ?

opencti-worker/src/worker.py
```diff
@@ -279,6 +280,13 @@ def __post_init__(self) -> None:
         self.current_bundle_id: [str, None] = None
         self.current_bundle_seq: int = 0
 
+        # Small hack to handle any rabbit prefix that could have been configured
+        self.listen_too_large_routing = (
```
xfournet: look a bit too hacky for me :) couldn't we send the dead letter queue name in the connector config we received from graphql ?

opencti-worker/src/worker.py
```diff
@@ -346,9 +376,31 @@ def data_handler(  # pylint: disable=too-many-statements, too-many-locals
...
+                            self.pika_parameters
+                        )
+                        push_channel = push_pika_connection.channel()
+                        try:
+                            push_channel.confirm_delivery()
+                        except Exception as err:  # pylint: disable=broad-except
+                            self.worker_logger.warning(str(err))
+                        for too_large_item_bundle in too_large_items_bundles:
+                            self.worker_logger.warning(
+                                "Detected a bundle too large, sending it to dead letter queue...",
```
xfournet: could we log more information about the rejected bundle ?

opencti-worker/src/push_handler.py
```diff
@@ -30,10 +35,36 @@ def __post_init__(self) -> None:
             json_logging=self.json_logging,
             ssl_verify=self.ssl_verify,
         )
+        # Small hack to handle any rabbit prefix that could have been configured
+        self.listen_too_large_routing = self.push_routing.replace(
+            "push_routing", "listen_routing"
+        ).replace(self.connector_id, "too-large-bundle")
```
xfournet: look a bit too hacky for me :) couldn't we send the dead letter queue name in the connector config we received from graphql ?

(reposted after rebase)

opencti-worker/src/push_handler.py
```diff
@@ -73,43 +104,73 @@ def handle_message(
...
+                    imported_items, too_large_items_bundles = (
+                        self.api.stix2.import_bundle_from_json(
+                            raw_content, update, types, work_id, self.objects_max_refs
+                        )
                     )
+                    if len(too_large_items_bundles) > 0:
+                        push_pika_connection = pika.BlockingConnection(
+                            self.pika_parameters
+                        )
+                        push_channel = push_pika_connection.channel()
```
xfournet: can use automatic resource management as it has been done at line 135

opencti-worker/src/push_handler.py
```diff
@@ -73,43 +104,73 @@ def handle_message(
...
                                 )
                             )
                             # Add expectations to the work
                             if work_id is not None:
                                 self.api.work.add_expectations(work_id, expectations)
+                                # For each bundle too large, send it to the too large queue
+                            for too_large_elements_bundle in too_large_elements_bundles:
+                                self.logger.warning(
+                                    "Detected a bundle too large, sending it to dead letter queue...",
+                                )
```
xfournet: could we log more information about the rejected bundle ?
or maybe it's already covered by python client lots ?

opencti-platform/opencti-graphql/src/database/rabbitmq.js
JeremyCloarec: yes that's the current design, all bundles considered too large are going to this too-large-bundle queue.
But thinking about it that could cause a problem since we no longer know what was the original queue. I'll maybe add an info in the bundles sent to this queue about the original queue

opencti-worker/src/config.yml.sample
JeremyCloarec: Not yet, haven't tested enough to have a clear guideline on the value to set

opencti-worker/src/push_handler.py
```diff
@@ -73,43 +104,73 @@ def handle_message(
...
+                            push_channel.confirm_delivery()
+                        except Exception as err:  # pylint: disable=broad-except
+                            self.logger.warning(str(err))
+                        for too_large_item_bundle in too_large_items_bundles:
+                            self.logger.warning(
+                                "Detected a bundle too large, sending it to dead letter queue...",
+                            )
+                            self.send_bundle_to_specific_queue(
+                                push_channel,
+                                self.listen_exchange,
```
xfournet: previously this was the push_exchange that was used

opencti-worker/src/worker.py
JeremyCloarec: you're right, I aligned everything to the objects_max_ref naming

opencti-worker/src/worker.py
```diff
@@ -231,19 +240,22 @@ def start(self) -> None:
                             self.opencti_url,
                             self.opencti_token,
                             self.opencti_ssl_verify,
+                            connector["id"],
                             connector_config["push_exchange"],
+                            connector_config["listen_exchange"],
```
xfournet, line 245: i didn't expect to use listen_exchange in the push handler

opencti-worker/src/worker.py
JeremyCloarec: my idea was that the bundle will be available in the queue to get all of the information we need, so we didn't need to log much here. 
I'm not sure what info would be useful to add in this log that we won't get in the bundle

opencti-worker/src/worker.py
JeremyCloarec: Yes I'll try to find a way to do something less hacky. I'll push something to move it to the ConnectorConfig

opencti-platform/opencti-graphql/src/database/rabbitmq.js
xfournet: yes, we need this origin metadata

opencti-platform/opencti-graphql/tests/02-integration/01-database/rabbitmq-test.js
JeremyCloarec: since we added a new internal connector to create the queue (too-large-bundle), this results in two queues being created in registerConnectorQueues: the push queue and the listen queue.
We currently don't have a way to only initialize one queue in our backend, that's why I reused the existing connector workflow.

opencti-worker/src/worker.py
xfournet: i don't know if there is some kind of id (from the bundle itself and/or rabbitmq message id) that would be useful

opencti-platform/opencti-graphql/src/database/rabbitmq.js
JeremyCloarec: I added an attribute to the bundles sent to this queue: "original_connector_id"

opencti-worker/src/push_handler.py
JeremyCloarec: yes we're still using the push_exchange when sending the split bundles, but we want to send the large bundles to the listen queue and not the push queue here (since our push queues are supposed to be the queues processed by the worker, and the listen_queue in this case will serve as a dead letter queue not consumed by anyone)

opencti-worker/src/push_handler.py
```diff
@@ -98,18 +157,12 @@ def handle_message(
...
-                                    routing_key=self.push_routing,
-                                    body=json.dumps(data),
-                                    properties=pika.BasicProperties(
-                                        delivery_mode=2,
-                                        content_encoding="utf-8",  # make message persistent
-                                    ),
+                                self.send_bundle_to_specific_queue(
+                                    push_channel,
+                                    self.push_exchange,
+                                    self.push_routing,
```
xfournet, line 163: at first sight i was thinking that, in case of too large bundle, we could avoid a useless roundtrip between rabbitmq and worker by calling compute_object_refs_number and send bundle to push_exchance/push_routing or listen_exchange/dead_letter_routing accordingly.

however i see that we are calling a report expectation in import stix, so i imagine it's because of that we cannot directly send it dead letter queue ?

xfournet (approved):

opencti-worker/src/push_handler.py
JeremyCloarec: yes that was my train of thought also: I tried doing what you said at first, but it was a bit complicated to handle the expectations properly.
I figured that the overhead of the roundtrip rabbitmq <-> worker wasn't problematic considering the gained simplicity in the code: this way, it is now only the responsibility of the import_item in opencti_stix2 to handle the computation of bundle complexity

PR #12637, opened 2025-10-05
Update Renovate configuration to enable lock file maintenance and group dev dependencies updates (#12636)

### LINKED ISSUE #12636: Improve Renovate configuration
## Description

Update Renovate configuration to enable lock file maintenance and group dev dependencies updates

xfournet:
### Proposed changes

* Enable lock file maintenance on weekly schedule to ensure transitive dependencies update
* Group minor/patch dev dependencies weekly
* Update AWS S3 SDK weekly because there is too many frequent updates

Automerge is not updated for now in order to check how it behave. Then it could be activated.

### Related issues
* Fix #12636

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,7 +1,8 @@
 {
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
   "extends": [
-    "config:recommended"
+    "config:recommended",
+    "docker:enableMajor"
   ],
   "labels": [
     "dependencies",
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -11,19 +11,39 @@
...
+    {
+      "extends": "monorepo:aws-sdk-js-v3",
+      "description": "Update aws-sdk-js-v3 only once a week",
+      "schedule": "on sunday after 1am and before 2am"
+    },
+    {
+      "groupName": "devDependencies (non-major)",
+      "groupSlug": "dev-dependencies-non-major",
+      "description": "Batch auto merge and batch on sunday morning, non-major update of dev dependencies",
+      "schedule": "on sunday after 3am and before 4am",
```
aHenryJard: I'm not convinced that schedule above works I think I saw renovate doing PRs out of "allowed" times. So I checked in documentation it's written that cron schedule is preferred.

https://docs.renovatebot.com/key-concepts/scheduling/#recommended-cron-syntax

I think that cronlike have better chances to works.

xfournet: indeed it's documented that the syntax is deprecated and will be removed in a future version, i will change it.
I'm not sure about the PR creation/update out of allowed ranges, it was the intent of updateNotScheduled=false but it's not very clear in the documentation why and when this can happen, and not clear if we can prevent it. We will see :)

aHenryJard (approved):

PR #12638, opened 2025-10-06
Update dev dependencies and refresh lockfile

xfournet:
### Proposed changes

* Update all straightforward dev and minor/patch dependencies
* Refresh all lockfiles


Archidoit (dismissed):

lndrtrbn (commented):
Could it be possible to split frontend and backend in 2 PRs? In case we need to rollback if we miss an issue

xfournet:
> Could it be possible to split frontend and backend in 2 PRs? In case we need to rollback if we miss an issue

yes, in fact i would like to preserve all the commit in order to keep the maximum of flexibility in case rollback would be necessary
but i need to discuss of theses updates first (on monday) so i keep as draft for the moment

lndrtrbn (commented):
image

Issue while trying to view a PDF in tab Content of a Report

PR #12644, opened 2025-10-06
[backend] remove LIVE_CATALOGS feature flag (#12314)

### LINKED ISSUE #12314: [Backend] Revert temporary live catalog mode feature flag
### Proposed changes

* Remove the temporary `LIVE_CATALOGS` feature flag hack from catalog-domain.ts
* Restore the original catalog loading behavior with standard caching
* Clean up the temporary code that was added for local development testing

### Related issues

* PR https://github.com/OpenCTI-Platform/opencti/pull/12312
* Removes temporary development hack that is no longer needed
* Proper catalog management implementation has been completed (or alternative solution found)

### Further comments

Removes the temporary live catalog hack that was introduced for local development and connector testing purposes. As indicated in the original implementation, this was always intended to be a temporary solution.

**Changes made:**
- Removed lines 22-85 containing the `LIVE_CATALOGS` feature flag check and live loading logic
- Removed the `isFeatureEnabled` import from line 7
- Code returns to its exact original state from `release/current`

**Impact:**
- No functional changes for production environments (feature was only for development)
- Developers using `LIVE_CATALOGS` flag should switch to the new catalog management solution
- The catalog system returns to its standard cached behavior

**Verification:**
The code after this revert is identical to the original implementation, ensuring no unintended side effects or regressions.

maelv-filigran:
### Proposed changes

* Removed the temporary `LIVE_CATALOGS` feature flag from catalog-domain.ts

### Related issues

* #12314

### Checklist

### Further comments

This feature was initially implemented as a temporary hack for the connector testing sandbox. While it was planned for removal, the functionality proved valuable for maintaining consistent catalog behavior and reproducible test conditions. 

This refactor removes the feature flag dependency while preserving the essential test functionality through a cleaner internal mechanism. The test mode now explicitly controls catalog cache bypassing, making the code more maintainable and the intent clearer.

No breaking changes - all existing tests continue to work as before.

### DIFF
--- a/opencti-platform/opencti-graphql/src/modules/catalog/catalog-domain.ts
+++ b/opencti-platform/opencti-graphql/src/modules/catalog/catalog-domain.ts
@@ -8,7 +8,7 @@ import { isEmptyField } from '../../database/utils';
 import { UnsupportedError } from '../../config/errors';
 import { idGenFromData } from '../../schema/identifier';
 import filigranCatalog from '../../__generated__/opencti-manifest.json';
-import conf, { isFeatureEnabled, logApp } from '../../config/conf';
+import conf, { logApp } from '../../config/conf';
 import type { ConnectorContractConfiguration, ContractConfigInput } from '../../generated/graphql';
 
 const CUSTOM_CATALOGS: string[] = conf.get('app:custom_catalogs') ?? [];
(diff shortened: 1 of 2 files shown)


opencti-platform/opencti-graphql/src/modules/catalog/catalog-domain.ts
```diff
@@ -91,19 +93,19 @@ const buildCatalogMap = (): Record<string, CatalogType> => {
   return newCatalogMap;
 };
 
-// Reset catalog cache - for testing purposes only
-export const resetCatalogCache = () => {
+// Enable test catalog mode - clears cache and enables live catalog loading for tests
+export const enableTestCatalogMode = () => {
   catalogMap = undefined as any;
+  contractsByImageCache = undefined;
+  testMode = true;
```
esrevi: i m wondering about the naming, even if it is not on purpose now, this could be used for other cases than testing imo.

I would suggest to rename testMode by "bypassCatalogCache" and enableTestCatalogMode by "enableCustomCatalogs"

wdyt?

maelv-filigran: you're right, I'll make the changes!

esrevi (approved):
lgtm

Issue #12362, opened 2025-09-10
[PIR] Use PIR Intelligence in Playbooks


alice-debra:
## Use case

When using PIRs feature, the platform identifies and marks relevant threats with flags and scores. However, this flagging information remains siloed within the PIR module and is not accessible to other platform modules, particularly our Playbook engine.
We need to make this information available to enable automation based on PIR-flagged entities. For example, I want to be able to automatically react to any new Indicators of Compromise that are associated with a threat flagged in a particular PIR with a score > 70.

## Current Workaround

N/A

## Proposed Solution

Make PIR intelligence flagged information and events available in playbook.

PR #12649, opened 2025-10-06
[backend/frontend] PIR in Playbooks - Detect entity enters/leaves a PIR (#12362)

lndrtrbn:
### Proposed changes

* Add a new Playbook component to manage PIR scenario
  * A select field to choose which PIR I want to listen
  * A toggle to listen for entities entering my PIR
  * A toggle to listen for entities leaving my PIR
  * Classic filters field to apply filters on the entity entering/leaving the PIR
* Modify Playbook component notifier to make a correct message for entities entering/leaving a PIR
* Modify `cacheManager` to fetch IDs of PIR at init (same logic than for filters with IDs)
* Split some code of `playbookManager` to improve readability and testability
* Move PIR component and Notifier component in their own files to improve readability and testability
* Unit tests

### Related issues

* #12362 

### Checklist


opencti-platform/opencti-graphql/src/manager/playbookManager.ts
```diff
@@ -256,37 +261,88 @@ const playbookStreamHandler = async (streamEvents: Array<SseEvent<StreamDataEven
...
+                    nextStep,
+                    // Data
+                    previousStepBundle: null,
+                    bundle,
+                  });
+                }
+              }
+            }
+            if (instance && instance.component_id === 'PLAYBOOK_INTERNAL_DATA_STREAM_PIR') {
+              if (scope === 'internal') {
```
SouadHadjiat: what is the difference between these two conditions on scope internal or external ?

lndrtrbn: In the POC we have created a new playbook component to listen on PIR events from the stream. Right now the playbook manager was listening only on external events, we modify this to also listen internal events, by so we have added a condition to not change the actual behavior of the playbook component that listen to knowledge.

In the new playbook component that listen to PIR events only internal events are of interest so we also added this condition.

opencti-platform/opencti-front/src/utils/filters/filtersUtils.tsx
```diff
@@ -129,6 +129,7 @@ export const stixFilters = [
   'takedown_types',
   'note_types',
   'incident_type',
+  'inPir',
```
Archidoit: do we want this filter in stream and triggers ? or only playbooks? If only for playbooks, add 'inPir' only in PlaybookAddComponent in case the selected component is the ListenKnowledge component

SarahBocognano (approved):
It's perfect

Issue #11911, opened 2025-08-05
[notifier] Convert markdown to HTML in platform mailer notifier


romain-filigran:
## Use case

In the current implementation of our platform's mailer notifier, the content is generated in HTML format. However, the description field of an entity is formatted in Markdown by default. We need to introduce a solution for converting Markdown-formatted content to HTML to ensure consistency and proper rendering in the mail notifications.

## Current Workaround

N/A

PR #12658, opened 2025-10-07
[backend] Convert markdown to HTML function for Platform mailer notifier (#11911)

Archidoit:
### Proposed changes
- Add a function usable in platform email notifiers, that converts markdown to HTML.
Enable to have a proper display of description in emails from platform notifiers.
- Replace showdown (not maintained library) by marked

### Related issues
#11911


maelv-filigran (approved):

opencti-platform/opencti-graphql/src/utils/NotificationTool.ts
```diff
@@ -0,0 +1,14 @@
+import showdown from 'showdown';
+
+export default class NotificationTool {
+  public static markdownToHtml(markdownText: string): string {
+    const converter = new showdown.Converter();
```
lndrtrbn: showdown is not maintained anymore. Maybe could be nice to see if there is an alternative?

Archidoit: indeed, we'll have a look, thanks !

Archidoit: done, we know use 'marked'

Archidoit:
exemple of how to use the function: 
<%-octi.markdownToHtml(data[0].instance.description)%>

PR #12667, opened 2025-10-08
[backend] Be more permissive on taxii push content type (#11432)

### LINKED ISSUE #11432: [TAXII] Server throws if content type header contains whitespaces
## Description

When calling POST ${basePath}/taxii2/root/collections/:id/objects` to push objects into a TAXII server, it expects to have a content-type header looks like: 'application/taxii+json;version=2.1'.
But ''application/taxii+json; version=2.1' (with whitespace) according to RFC7231 is also acceptable and should be valid, but currently the server throws 500 internal server error.

## Environment

6.6.18

## Reproducible Steps

1. Create a TAXII collection in ingestion -> TAXII Push
2. Push data to the collection (API call) with content-type header contains whitespace.

## Expected Output

200 status code response.

## Actual Output

500 status code response.

## Additional information

From short investigation, issue is here:
https://github.com/OpenCTI-Platform/opencti/blob/7f3b6d4dc9a5e5fae8c0a0d040c32c38d804c812/opencti-platform/opencti-graphql/src/http/httpTaxii.js#L22

~lines 79-84
const JsonTaxiiMiddleware = express.json({
  type: (req) => {
    return req.headers['content-type'] === TAXII_VERSION;
  },
  limit: nconf.get('app:max_payload_body_size') || '50mb'
});

TAXII_VERSION is defined in line 22:
const TAXII_VERSION = 'application/taxii+json;version=2.1';

aHenryJard:
### Proposed changes

* Allow white space in content type as it's allowed by the HTTP norm
* Add another allowed content type to be more compatible, for example with current implementation of https://github.com/OpenCTI-Platform/connectors/tree/master/stream/taxii-post
* Add test coverage in follow up PR => https://github.com/OpenCTI-Platform/opencti/pull/12690

### Related issues

* closes #11432 

### Checklist


opencti-platform/opencti-graphql/tests/07-worker-required/httpTaxii-test.ts
```diff
@@ -0,0 +1,98 @@
...
+              score: 50
+            }
+          },
+          value: 'patate@truc.fr'
+        }
+      ]
+    };
+
+    // First check that url config is fine.
+    // const axiosHealthResponse = await axios.get(`${getBaseUrl()}/health?health_access_key=cihealthkey`);
```
aHenryJard: Keeping this in comment because I want to uncomment in a follow up PR

opencti-platform/opencti-graphql/tests/utils/globalSetup.ts
```diff
@@ -42,6 +54,7 @@ const initializePlatform = async () => {
   await initializeInternalQueues();
   await initializeBucket();
   await initializeSchema();
+  // await testInitializeMigration(context);
```
aHenryJard: Keeping this as commented because I want to put it back in a follow up PR

opencti-platform/opencti-graphql/tests/07-worker-required/httpTaxii-test.ts
```diff
@@ -0,0 +1,98 @@
...
+      Authorization: `Bearer ${ADMIN_API_TOKEN}`
+    };
+    axiosResponsePost = await axios.post(`${getBaseUrl()}/taxii2/root/collections/${taxiiPushIngestion.id}/objects`, bundleObject, { headers: headersToTest3 });
+    expect(axiosResponsePost.status, 'With correct content type including space and authentication should works fine').toBe(200);
+
+    const headersToTest4 = {
+      'Content-Type': 'application/vnd.oasis.stix+json; version=2.1',
+      Authorization: `Bearer ${ADMIN_API_TOKEN}`
+    };
+    axiosResponsePost = await axios.post(`${getBaseUrl()}/taxii2/root/collections/${taxiiPushIngestion.id}/objects`, bundleObject, { headers: headersToTest4 });
```
SouadHadjiat: shouldn't it impact our counters ? since it will create this email address in the db, right ?

SouadHadjiat: I guess the bundle is pushed in the queue but it's not consumed by the worker ? not enough time ?

opencti-platform/opencti-graphql/src/http/httpTaxii.js
```diff
@@ -78,7 +81,8 @@ const extractUserAndCollection = async (req, res, id) => {
 
 const JsonTaxiiMiddleware = express.json({
   type: (req) => {
-    return req.headers['content-type'] === TAXII_VERSION;
+    const contentTypeFromRequest = parseContentType(req);
```
esrevi: parseContentType throws a TypeError if the string is missing or invalid, is this intentional? Otherwise, we could wrap in try-catch to handle that gracefully

opencti-platform/opencti-graphql/tests/07-worker-required/httpTaxii-test.ts
SouadHadjiat: I think we should add a waitInSec(75) (a bit more than 1min) to be sure the worker runs and process the bundle, and it should add 1 email address (1 create event in the stream)

SouadHadjiat: otherwise raw-test might fail if worker has ingested the bundle in time (another flaky case)

opencti-platform/opencti-graphql/src/http/httpTaxii.js
aHenryJard: ah thanks good catch

opencti-platform/opencti-graphql/tests/07-worker-required/httpTaxii-test.ts
aHenryJard: Damn you are totally right. Ok instead of adding yet one more sleep and make api-test longer, I prefer to move it after sync tests then.

opencti-platform/opencti-graphql/tests/07-worker-required/httpTaxii-test.ts
```diff
@@ -0,0 +1,98 @@
...
+    // First check that url config is fine.
+    // const axiosHealthResponse = await axios.get(`${getBaseUrl()}/health?health_access_key=cihealthkey`);
+    // expect(axiosHealthResponse.status, '200');
+
+    await expect(async () => {
+      await axios.post(`${getBaseUrl()}/taxii2/root/collections/${taxiiPushIngestion.id}/objects`, bundleObject, { });
+    }).rejects.toThrowError('Request failed with status code 401');
+
+    /*
+    UNCOMMENT THIS CODE when fixing taxii2 push server behavior on no content type or bad content type
```
aHenryJard: See https://github.com/OpenCTI-Platform/opencti/issues/12668

opencti-platform/opencti-graphql/tests/07-worker-required/README.md
```diff
@@ -0,0 +1,5 @@
+# Workers required test
+
+This folder is for tests that ingest data using queue and workers and can make the sync tests flaky.
+
+Please keep as many tests as possible in 01/02 folders.
```
JeremyCloarec: Couldn't we instead keep the worker requireds tests in 02-integration, but add a sleep 1 minutes in the before of 03-stream tests to ensure that the worker had time to register to all the queues/process all messages? This way we ensure that the ingestion through workers is properly sent to the stream.

Also, there are already some tests that require the worker to run in some integration tests, like in organization-sharing-test
Image
I think it could be a bit confusing to have worker required tests both in 02 and 07 wdyt?

aHenryJard: yes I know there is other test, they can be move afterwards. It's also a pain to discover that when you run them locally so I think it's also good to be explicit.
I'm really not sure it's a good test pattern to wait for worker for tests that are run on PR. It's more a pattern for end to end tests run outside of PRs to me.

Also that's my point, workers processing tests requires sleep and waiting time so it's good candidate to be run in parallel in a separate step.

aHenryJard: notice the WORKER REQUIRED that I have added in capital case :)

JeremyCloarec (approved):

PR #12673, opened 2025-10-09
[backend] fix some mismatch between graphql API and entities from schema(#11800)

### LINKED ISSUE #11800: GraphQL internal server error
## Description

Querying `objectAssignee` and `spec_version` on a TAG with markings causes an internal server error

Image

```
query Query($id: String!) {
  threatActorGroup(id: $id) {
    id
    objectAssignee {
      id
    }
    objectMarking {
      spec_version
    }
  }
}
```

```
{
  "errors": [
    {
      "message": "Cannot read properties of null (reading 'databaseName')",
      "locations": [
        {
          "line": 4,
          "column": 5
        }
      ],
      "path": [
        "threatActorGroup",
        "objectAssignee"
      ],
      "extensions": {
      },
    },
    {
      "message": "Cannot return null for non-nullable field MarkingDefinition.spec_version.",
      "locations": [
        {
          "line": 8,
          "column": 7
        }
      ],
      "path": [
        "threatActorGroup",
        "objectMarking",
        0,
        "spec_version"
      ],
      "extensions": {
      },
    }
  ],
  "data": {
    "threatActorGroup": {
      "id": "4b0d54c7-abe3-4d8b-bb9a-28e910e51cd4",
      "objectAssignee": null,
      "objectMarking": null
    }
  }
}
```

## Environment

2. OpenCTI version: 6.7.7
4. Other environment details:

## Reproducible Steps

1. { e.g. Run ... }
2. { e.g. Click ... }
3. { e.g. Error ... }

JeremyCloarec:
### Proposed changes

* fix retrocompatiblity for spec_version on stixMetaObject 
* add explicit error when loading of refs not in entity schema
* keep parent_types typing coherent between BasicObject interface and all other types
* fix Notifier typing on static notifier

### Related issues

* #11800 
*

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/config/schema/opencti.graphql
+++ b/opencti-platform/opencti-graphql/config/schema/opencti.graphql
@@ -1443,7 +1443,7 @@ type Settings implements InternalObject & BasicObject & ThemeSettings & IntlSett
   id: ID!
   standard_id: String!
   entity_type: String!
-  parent_types: [String!]!
+  parent_types: [String]!
   metrics:[Metric]
   # Settings
   platform_enterprise_edition: PlatformEE!
@@ -1800,7 +1800,7 @@ type MeUser implements BasicObject & InternalObject {
   id: ID! # internal_id
   standard_id: String!
   entity_type: String!
-  parent_types: [String!]!
+  parent_types: [String]!
   metrics:[Metric]
   user_email: String!
   name: String!
--- a/opencti-platform/opencti-graphql/config/schema/opencti.graphql
+++ b/opencti-platform/opencti-graphql/config/schema/opencti.graphql
@@ -12638,7 +12638,7 @@ type SSHKey implements BasicObject & StixObject & StixCoreObject & StixCyberObse
   id: ID! # internal_id
   standard_id: String!
   entity_type: String!
-  parent_types: [String]!
+  parent_types: [String!]!
   metrics:[Metric]
   # StixObject
   representative: Representative!
--- a/opencti-platform/opencti-graphql/config/schema/opencti.graphql
+++ b/opencti-platform/opencti-graphql/config/schema/opencti.graphql
@@ -1272,7 +1272,7 @@ interface BasicObject {
   id: ID! # internal_id
   standard_id: String!
   entity_type: String!
-  parent_types: [String]!
+  parent_types: [String!]!
   metrics:[Metric]
 }
 
@@ -1443,7 +1443,7 @@ type Settings implements InternalObject & BasicObject & ThemeSettings & IntlSett
   id: ID!
   standard_id: String!
   entity_type: String!
-  parent_types: [String]!
+  parent_types: [String!]!
   metrics:[Metric]
   # Settings
   platform_enterprise_edition: PlatformEE!
@@ -1554,7 +1554,7 @@ type Group implements InternalObject & BasicObject {
   standard_id: String!
   entity_type: String!
   auto_integration_assignation: [String]!
-  parent_types: [String]!
+  parent_types: [String!]!
   metrics:[Metric]
   # Group
   name: String!
@@ -1736,7 +1736,7 @@ type User implements BasicObject & InternalObject {
   id: ID! # internal_id
   standard_id: String!
   entity_type: String!
-  parent_types: [String]!
+  parent_types: [String!]!
   metrics:[Metric]
   # User
   user_email: String!
@@ -1800,7 +1800,7 @@ type MeUser implements BasicObject & InternalObject {
   id: ID! # internal_id
   standard_id: String!
   entity_type: String!
-  parent_types: [String]!
+  parent_types: [String!]!
   metrics:[Metric]
   user_email: String!
   name: String!
@@ -1930,7 +1930,7 @@ type Role implements BasicObject & InternalObject {
   id: ID! # internal_id
   standard_id: String!
   entity_type: String!
-  parent_types: [String]!
+  parent_types: [String!]!
   metrics:[Metric]
   # Role
   name: String!
@@ -1962,7 +1962,7 @@ type Capability implements BasicObject & InternalObject {
   id: ID! # internal_id
   standard_id: String!
   entity_type: String!
-  parent_types: [String]!
+  parent_types: [String!]!
   metrics:[Metric]
   # Capability
   name: String!
@@ -2202,7 +2202,7 @@ type Connector implements BasicObject & InternalObject {
   id: ID! # internal_id
   standard_id: String!
   entity_type: String!
-  parent_types: [String]!
+  parent_types: [String!]!
   metrics:[Metric]
   # Connector
   name: String!
@@ -2250,7 +2250,7 @@ type ManagedConnector implements BasicObject & InternalObject {
   id: ID! # internal_id
   standard_id: String!
   entity_type: String!
-  parent_types: [String]!
+  parent_types: [String!]!
   metrics:[Metric]
   # Connector
   name: String!
@@ -2280,7 +2280,7 @@ type ConnectorManager implements BasicObject & InternalObject {
   id: ID! # internal_id
   standard_id: String!
   entity_type: String!
-  parent_types: [String]!
+  parent_types: [String!]!
   metrics:[Metric]
   # Manager
   name: String!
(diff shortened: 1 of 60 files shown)


opencti-platform/opencti-graphql/config/schema/opencti.graphql
```diff
@@ -1431,7 +1431,7 @@ type Settings implements InternalObject & BasicObject & ThemeSettings & IntlSett
   id: ID!
   standard_id: String!
   entity_type: String!
-  parent_types: [String!]!
+  parent_types: [String]!
```
lndrtrbn: Does it means we can have undefined values in the array? That feels very odd

opencti-platform/opencti-graphql/src/resolvers/stix.js
```diff
@@ -24,6 +24,9 @@ const internalLoadThroughDenormalized = (context, user, element, inputName) => {
   }
   // If not, reload through denormalized relationships
   const ref = schemaRelationsRefDefinition.getRelationRef(element.entity_type, inputName);
+  if (!ref) {
```
lndrtrbn, line 29: Is it easy to add a test for this case?

opencti-platform/opencti-graphql/config/schema/opencti.graphql
JeremyCloarec: I agree that the better typing would be [String!]!, but since the type used in BasicObject is [String]!, I aligned everything on this.
I was afraid that changing everything to [String!]! would break some things (in the case of some old/corrupted data that contained null values in their parent_types array)

opencti-platform/opencti-graphql/src/resolvers/stix.js
JeremyCloarec: It's not too complicated no, I'll just add objectAssignee to some of our fragments in our tests

opencti-platform/opencti-graphql/config/schema/opencti.graphql
xfournet: with current schema that forbid undefined values in the array, what is the behavior of GraphQL if an undefined element is present in the array ?

JeremyCloarec: we get this kind of error when that is the case:
image
An INTERNAL_SERVER_ERROR is raised, and no data is returned (or more precisely, the data is null)
image

xfournet: so after the change it can fail for API consumer that is not expected to receive an undefined value.
could we rather filter the list to remove undefined values in it ?

JeremyCloarec: Hmmm, that's not easy to do in a clean way, since parent_types retrieved by the API is the data from elastic, we don't have any resolver in between. We could do it by introducing a post filtering on parent_types in elDataConverter in engine.js, but I think that's overkill in this case.

I think the risk of breaking changes is quite small, since parent_types having undefined values should never happen unless for some rare cases of corrupted data, and since parent_types is probably not used.

opencti-platform/opencti-graphql/src/modules/notifier/notifier-statics.ts
```diff
@@ -169,6 +173,9 @@ export const STATIC_NOTIFIERS: Array<BasicStoreEntityNotifier> = [
   // @ts-ignore
   {
     id: STATIC_NOTIFIER_EMAIL,
+    standard_id: `notifier--${STATIC_NOTIFIER_UI}`,
```
samczsun: This should be `STATIC_NOTIFIER_EMAIL`?

opencti-platform/opencti-graphql/config/schema/opencti.graphql
xfournet: should we ?

- prevent the insertion of null/undefined values in parent_types
- either do a migration or add on the fly cleanup of parent_types
- align BasicObject.parent_types to [String!]!

opencti-platform/opencti-graphql/src/resolvers/stix.js
xfournet, line 29: objectAssignee on threadactorgroup is supported by the graphql schema & resolvers (both at read and at creation/update), but not in the schema in DB. We should rather update the schema in DB

opencti-platform/opencti-graphql/src/domain/stixCoreObject.js
```diff
@@ -177,7 +177,7 @@ export const batchInternalRels = async (context, user, elements) => {
   const relTypes = new Set();
   for (let index = 0; index < elements.length; index += 1) {
     const { element, definition } = elements[index];
-    if (isNotEmptyField(element[definition.databaseName])) {
+    if (isNotEmptyField(element[definition?.databaseName])) {
```
xfournet: this should be handled by aligning the schema in DB with the graphql schema

opencti-platform/opencti-graphql/src/resolvers/stix.js
JeremyCloarec: But in this case, functionnaly speaking, I'd say the schema is right & graphql is wrong. Assignees are only relevant on containers to be able to assign people there, and it wouldn't make sense to have assignees on threatactorgroups
image
I have no idea why the assignees were added to the `StixDomainObject` interface on graphql though. The proper schema typing would have been to set it on `Container`. But I can't do this change without risking breaking some things for consumers of the API.
This is way I made the changes to at least not have errors when querying objectAssignees on the graphql API on entities not having this ref in the schema

opencti-platform/opencti-graphql/src/domain/stixCoreObject.js
JeremyCloarec: See comment above for not aligning graphql schema with DB schema

opencti-platform/opencti-graphql/src/modules/internal/csvMapper/csvMapper.graphql
```diff
@@ -65,7 +65,7 @@ type CsvMapper implements InternalObject & BasicObject {
   id: ID!
   entity_type: String! @auth
   standard_id: String! @auth
-  parent_types: [String!]!
+  parent_types: [String!]! @auth
   metrics:[Metric] @auth
```
xfournet, line 69: is @auth really necessary on CsvMapper fields ?

opencti-platform/opencti-graphql/src/modules/notifier/notifier.graphql
```diff
@@ -23,15 +23,15 @@ type Notifier implements InternalObject & BasicObject {
...
+    parent_types: [String!]!
     metrics:[Metric]
     created: DateTime
     modified: DateTime
     name: String!
     description: String
     notifier_connector: NotifierConnector!
     notifier_connector_id: String!
-    notifier_configuration: String!
+    notifier_configuration: String
```
xfournet: to be checked

opencti-platform/opencti-graphql/src/resolvers/stix.js
xfournet: -> don't mask the error, but fail properly by indicating API/Db schema mismatch

opencti-platform/opencti-graphql/src/domain/stixCoreObject.js
xfournet: -> don't mask the error, but fail properly by indicating API/Db schema mismatch

opencti-platform/opencti-graphql/src/modules/internal/csvMapper/csvMapper.graphql
xfournet: to be cleaned with potentially other @auth directives in another PR

opencti-platform/opencti-graphql/src/modules/notifier/notifier.graphql
xfournet: fix #11800

xfournet (approved):

opencti-platform/opencti-graphql/src/modules/notifier/notifier.graphql
xfournet: to be rollback and and an empty JSON object config for build UI notifier

PR #12690, opened 2025-10-10
[backend] Fix taxii server error management (#12668)(#8946)

### LINKED ISSUE #8946: OpenCTI Taxii server return HTTP 200 OK for bad credentials instead of HTTP 403
## Description

OpenCTI Taxii server return HTTP 200 OK for bad credentials instead of HTTP 403.

The consequence is that using OpenCTI as taxii client with another OpenCTI => there is no error, so no error log to understand the issue.

## Environment

2. OpenCTI version: { e.g. OpenCTI 1.0.2 } 6.3.11
4. Other environment details:

## Reproducible Steps

1. Create a taxii server on one instance that is not public
2. Call the url with a Curl with wrong data for auth

## Expected Output

Expect to have an HTTP error 403 forbidden

## Actual Output

HTTP 200 OK and OpenCTI login page as HTML in the response body

## Additional information

Please check also other Taxii server error (like 500).

### LINKED ISSUE #12668: Taxii push feed not reponding with HTTP Error when content type is incorrect
## Description

When calling the HTTP endpoint for a Taxii push feed, if content-type is not set or not supported, the server is not responding until there is a timeout.

Should answer an HTTP error code instead (400 or 500).

Found when fixing https://github.com/OpenCTI-Platform/opencti/issues/11432

## Environment

2. OpenCTI version: 6.8.3
4. Other environment details:

## Reproducible Steps

- Run commented test from https://github.com/OpenCTI-Platform/opencti/pull/12667
- Or configure a taxii push feed and with python run =>
```
import requests

url = 'http://localhost:4000/taxii2/root/collections/fcf33d90-xxxx-xxxx-xxxx-xxxxxxxxxxxx/objects/'
myjson = {'somekey': 'somevalue'}


myjson2 = {
    "type": "bundle",
    "spec_version": "2.1",
    "id": "bundle--3d9ba8f5-ae82-4341-81ee-78a6e146d1cd",
    "objects": []
}

#headers = {
#            "Content-Type": "application/vnd.oasis.stix+json; version=2.1",
#            "Accept": "application/vnd.oasis.taxii+json; version=2.1",
#        }
headers = {}
headers["Authorization"] = "Bearer xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx"

print(myjson2)
print(headers)

x = requests.post(url, json = myjson2, headers = headers)

#print the response text (the content of the requested file):

print(x.text)

```

## Additional information

It's all in httpTaxii.js ; probably something to fix inside JsonTaxiiMiddleware
https://github.com/OpenCTI-Platform/opencti/blob/366d8989b5c1501ec80b00e7f0090caccfa2b954/opencti-platform/opencti-graphql/src/http/httpTaxii.js#L79-L84

aHenryJard:
### Proposed changes

* Fix unconsistent HTTP codes and add coverage on Taxii server GET and Push.

#### Tests follow up

* extracting test part from https://github.com/OpenCTI-Platform/opencti/pull/12667 to allow discussion time without stress
* add a new test from a recent issue
* add a new test folder because we don't want to wait for worker to process message in this test (will make sync test flaky)
* Keep the "deprecated" folder for tests at the end
* update the dev test setup to add new folders

### Related issues

* add test coverage to #11432 
* add test coverage to  #12983
* Closes #12668
* Closes #8946

### Checklist


opencti-platform/opencti-graphql/tests/08-httpTaxii/httpTaxii-test.ts
```diff
@@ -0,0 +1,154 @@
...
+        }
+      ]
+    };
+
+    await expect(async () => {
+      await axios.post(`${getBaseUrl()}/taxii2/root/collections/${taxiiPushIngestionId}/objects`, bundleObject, { });
+    }).rejects.toThrowError('Request failed with status code 401');
+
+    /*
+    UNCOMMENT THIS CODE when fixing taxii2 push server behavior on no content type or bad content type
```
xfournet: isn't already fixed ?

opencti-platform/opencti-graphql/tests/08-httpTaxii/httpTaxii-test.ts
```diff
@@ -0,0 +1,154 @@
...
+
+    let axiosResponsePost = await axios.post(`${getBaseUrl()}/taxii2/root/collections/${taxiiPushIngestionId}/objects`, bundleObject, { headers: headersToTest2 });
+    expect(axiosResponsePost.status, 'With correct content type and authentication should works fine').toBe(200);
+    // We do not check entity from bundleObject in database since it requires the worker to process it.
+
+    const headersToTest3 = {
+      'Content-Type': 'application/taxii+json; version=2.1',
+      Authorization: `Bearer ${ADMIN_API_TOKEN}`
+    };
+    axiosResponsePost = await axios.post(`${getBaseUrl()}/taxii2/root/collections/${taxiiPushIngestionId}/objects`, bundleObject, { headers: headersToTest3 });
```
xfournet: we should use separate variable for the different response.
But seems that in fact we need to use test 'it' that looks more appropriate to separate the different tests that are done in a single test.

opencti-platform/opencti-graphql/tests/08-httpTaxii/httpTaxii-test.ts
```diff
@@ -0,0 +1,154 @@
...
+    const headers = {
+      Authorization: `Bearer ${ADMIN_API_TOKEN}`
+    };
+    const taxiiRootResponse = await axios.get(`${getBaseUrl()}/taxii2/root/`, { headers });
+    const { data, status } = taxiiRootResponse;
+    expect(status, 'With correct content type and authentication should works fine').toBe(200);
+    expect(data.versions, 'With correct content type and authentication should works fine').toStrictEqual(['application/taxii+json;version=2.1']);
+  });
+
+  // This one is not working yet, to be investigated
```
xfournet: comment to be removed or there is still a problem ?

opencti-platform/opencti-graphql/tests/08-httpTaxii/httpTaxii-test.ts
```diff
@@ -0,0 +1,154 @@
...
+  it.skip('should taxii collection works', async () => {
+    const headers = {
+      Authorization: `Bearer ${ADMIN_API_TOKEN}`
+    };
+    const taxiiCollectionResponse = await axios.get(`${getBaseUrl()}/taxii2/root/collections/${taxiiCollectionId}/objects/`, { headers });
+    // console.log('taxiiCollectionResponse:', { taxiiCollectionResponse });
+    const { status } = taxiiCollectionResponse;
+    // console.log('DATA collection:', { data });
```
xfournet: commented console.log to be removed ?

opencti-platform/opencti-graphql/vitest.config.dev.ts
```diff
@@ -1,5 +1,5 @@
 /* eslint-disable import/no-extraneous-dependencies */
 import { buildTestConfig } from './vitest.config.test';
 
-export default buildTestConfig(['tests/(01|02|03|05|06)-*/**/*-test.{js,mjs,cjs,ts,mts,cts,jsx,tsx}']);
-// export default buildTestConfig(['tests/(02)-*/**/(loader|filterGroup|grouping|workspace)*-test.{js,mjs,cjs,ts,mts,cts,jsx,tsx}']);
+export default buildTestConfig(['tests/(01|02|03|04|05|06|07|08)-*/**/*-test.{js,mjs,cjs,ts,mts,cts,jsx,tsx}']);
```
xfournet: is it related to the added tests in this PR or opportunity to fix the config that was not good ?

opencti-platform/opencti-graphql/config/test.json
```diff
@@ -1,6 +1,8 @@
 {
   "app": {
     "port": 4010,
+    "base_url": "http://localhost:4010/",
```
xfournet: why using localhost here while 127.0.0.1 is used for other existing config like in sync_raw_start_remote_uri ?
should we align all on localhost or either 127.0.0.1 ?

opencti-platform/opencti-graphql/tests/08-httpTaxii/httpTaxii-test.ts
```diff
@@ -0,0 +1,154 @@
+import { describe, expect, it } from 'vitest';
+import axios from 'axios';
```
xfournet: fetch is available in node since v18. At least for new code, we should use fetch instead of using legacy way like axios

opencti-platform/opencti-graphql/tests/02-integration/08-httpServer/httpHealth-test.ts
```diff
@@ -0,0 +1,13 @@
+import { describe, expect, it } from 'vitest';
+import axios from 'axios';
+import { getBaseUrl } from '../../../src/config/conf';
+
+describe.skip('Should /health be available', () => {
+  // skipped because there is an issue with migration init.
+  // Please see testInitializeMigration in globalSetup
```
xfournet: since it's not working and not directly related to the taxii push test, should we remove it (with globalSetup changes) from this PR ?

opencti-platform/opencti-graphql/tests/08-httpTaxii/httpTaxii-test.ts
```diff
@@ -0,0 +1,154 @@
...
+      ]
+    };
+    const taxiiPushIngestion = await addIngestion(testContext, ADMIN_USER, ingestionAdd);
+    expect(taxiiPushIngestion.id).toBeDefined();
+    expect(taxiiPushIngestion.name).toBe('Taxii push test ingestion');
+    // Start it
+    const moveToRunning: EditInput[] = [{ key: 'ingestion_running', value: ['true'] }];
+    await ingestionEditField(testContext, ADMIN_USER, taxiiPushIngestion.id, moveToRunning);
+
+    taxiiPushIngestionId = taxiiPushIngestion.id;
```
xfournet: it should be handled with a beforeAll

opencti-platform/opencti-graphql/tests/08-httpTaxii/httpTaxii-test.ts
```diff
@@ -0,0 +1,154 @@
...
+    expect(status, 'With correct content type and authentication should works fine').toBe(200);
+    expect(data.versions, 'With correct content type and authentication should works fine').toStrictEqual(['application/taxii+json;version=2.1']);
+  });
+
+  // This one is not working yet, to be investigated
+  it.skip('should taxii collection works', async () => {
+    const headers = {
+      Authorization: `Bearer ${ADMIN_API_TOKEN}`
+    };
+    const taxiiCollectionResponse = await axios.get(`${getBaseUrl()}/taxii2/root/collections/${taxiiCollectionId}/objects/`, { headers });
```
xfournet: in other test with are using /objects and here /objects/
should test both systematically to ensure consistent behavior ? an utility test function could help to make that

we could also test the returned header, eg Content-Type

opencti-platform/opencti-graphql/tests/08-httpTaxii/httpTaxii-test.ts
```diff
@@ -0,0 +1,154 @@
...
+      include_inferences: true,
+      score_to_confidence: false,
+      filters: '{\'mode\':\'and\',\'filters\':[{\'key\':[\'entity_type\'],\'operator\':\'eq\',\'values\':[\'Indicator\'],\'mode\':\'or\'}],\'filterGroups\':[]}'
+    };
+    const taxiiCollection = await createTaxiiCollection(testContext, ADMIN_USER, dataSharingTaxii);
+    expect(taxiiCollection.id).toBeDefined();
+    taxiiCollectionId = taxiiCollection.id;
+    expect(taxiiCollection.name).toBe('Testing coverage on Taxii collection');
+
+    await waitInSec(10);
```
xfournet: we should poll the collection availability rather introducing a long delay that increase test duration

opencti-platform/opencti-graphql/tests/08-httpTaxii/httpTaxii-test.ts
aHenryJard: no the server is hanging on that case... we need to investigate that.

opencti-platform/opencti-graphql/vitest.config.dev.ts
aHenryJard: we never run yarn test:dev locally, we always run yarn test:dev:resume so when there is some counter issue and you need to run something in 04 it's missing here.
Both answer then: 08 is required now and also I too the opportunity to change it.

opencti-platform/opencti-graphql/tests/02-integration/08-httpServer/httpHealth-test.ts
aHenryJard: yes overall I hesitated between remove what is not working yet and not. I want to work on it later but it's not helping for this review that's true.... Ok maybe to remove them all. I was a bit surprise to not being able to call /health in tests to be honest. But I cannot run after every rabbit at once.

opencti-platform/opencti-graphql/config/test.json
aHenryJard: all config changes were for the /health as far as I remeber, I will remove.

opencti-platform/opencti-graphql/tests/08-httpTaxii/httpTaxii-test.ts
aHenryJard: We are using axios in opencti code that is why I took it. I'm fine by using fetch instead.

opencti-platform/opencti-graphql/tests/08-httpTaxii/httpTaxii-test.ts
aHenryJard: https://github.com/OpenCTI-Platform/opencti/issues/12668

opencti-platform/opencti-graphql/config/test.json
aHenryJard: ah no I remember now, yes right

opencti-platform/opencti-graphql/tests/08-httpTaxii/httpTaxii-test.ts
aHenryJard: done.

opencti-platform/opencti-graphql/tests/08-httpTaxii/httpTaxii-test.ts
aHenryJard: I don't understand why I can't make this test work for now... do you prefer that I remove it ?

opencti-platform/opencti-graphql/tests/08-httpTaxii/httpTaxii-test.ts
aHenryJard: done.

opencti-platform/opencti-graphql/tests/08-httpTaxii/httpTaxii-test.ts
aHenryJard: done.

opencti-platform/opencti-graphql/tests/08-httpTaxii/httpTaxii-test.ts
aHenryJard: done.

opencti-platform/opencti-graphql/vitest.config.dev.ts
```diff
@@ -1,5 +1,5 @@
 /* eslint-disable import/no-extraneous-dependencies */
 import { buildTestConfig } from './vitest.config.test';
 
-export default buildTestConfig(['tests/(01|02|03|05|06)-*/**/*-test.{js,mjs,cjs,ts,mts,cts,jsx,tsx}']);
-// export default buildTestConfig(['tests/(02)-*/**/(loader|filterGroup|grouping|workspace)*-test.{js,mjs,cjs,ts,mts,cts,jsx,tsx}']);
+export default buildTestConfig(['tests/(01|02|03|04|05|06|07)-*/**/*-test.{js,mjs,cjs,ts,mts,cts,jsx,tsx}']);
```
xfournet, line 4: 04 is requiring connectors python files ?

opencti-platform/opencti-graphql/tests/06-httpTaxii/httpTaxii-test.ts
```diff
@@ -0,0 +1,157 @@
...
+      body: JSON.stringify(bundleObject),
+    });
+    const data3: any = await postResponse3.json();
+    expect(postResponse3.status, 'With correct content type including space and authentication should works fine').toBe(200);
+    expect(data3.status).toBe('pending');
+  });
+
+  it('should taxii post be refused without authenticated user', async () => {
+    await expect(async () => {
+      await axios.post(`${getBaseUrl()}/taxii2/root/collections/${taxiiPushIngestionId}/objects`, bundleObject, {});
```
xfournet: there is still axios here

opencti-platform/opencti-graphql/tests/06-httpTaxii/httpTaxii-test.ts
```diff
@@ -0,0 +1,157 @@
+import { afterAll, beforeAll, describe, expect, it, vi } from 'vitest';
```
xfournet: vi is not used

opencti-platform/opencti-graphql/tests/08-httpTaxii/httpTaxii-test.ts
xfournet: my comment was unclear, i wanted rather to use it.each to avoid repetition, eg 

```ts
  it.each([
    { contentType: 'application/taxii+json;version=2.1', name: 'TAXII content type and authentication header' },
    { contentType: 'application/taxii+json; version=2.1', name: 'TAXII content type including space and authentication header' },
    { contentType: 'application/vnd.oasis.stix+json; version=2.1', name: 'STIX content type including space and authentication header' },
  ])('should taxii post standard behavior works correctly with $name', async ({ contentType }) => {
    // Testing the content type with no space inside
    const postResponse = await fetch(`${getBaseUrl()}/taxii2/root/collections/${taxiiPushIngestionId}/objects`, {
      method: 'POST',
      headers: {
        'Content-Type': contentType,
        Authorization: `Bearer ${ADMIN_API_TOKEN}`
      },
      body: JSON.stringify(bundleObject),
    });
    const data: any = await postResponse.json();
    expect(postResponse.status).toBe(200);
    expect(data.status).toBe('pending');
    // We do not check entity from bundleObject in database since it requires the worker to process it.
  });
```

opencti-platform/opencti-graphql/tests/02-integration/08-httpServer/httpHealth-test.ts
xfournet: should we add this pb to #12668 or create a separate issue ?

opencti-platform/opencti-graphql/tests/08-httpTaxii/httpTaxii-test.ts
xfournet: should we test the returned content type ?
should we test both with /objects/ and /objects ?

opencti-platform/opencti-graphql/tests/06-httpTaxii/httpTaxii-test.ts
aHenryJard: oh yes 🤦‍♀️

opencti-platform/opencti-graphql/vitest.config.dev.ts
aHenryJard: since nobody runs yarn test:dev (unless on 01) I think it's fine to put all the number here.
Some tests requires worker, some other connectors git along, some other another opencti instance... for now it's always a discovery on what is required exactly and it not exactly match with the numbers.

I can revert to get this PR merged but when you want to run sync test locally you need to change that file and not commit it.

xfournet (approved):

opencti-platform/opencti-graphql/src/http/httpTaxii.js
```diff
@@ -225,13 +229,19 @@ const initTaxiiApi = (app) => {
     }
   });
   app.post(`${basePath}/taxii2/root/collections/:id/objects/`, JsonTaxiiMiddleware, async (req, res) => {
-    const { id } = req.params;
-    const { objects = [] } = req.body;
     try {
+    // Authentication is checked in this method, keep it first but inside try block.
+      const context = await checkAuthenticationFromRequest(req, res);
+      if (!isValidTaxiiPostContentType(req)) {
```
aHenryJard, line 235: Need to be at 2 places because JsonTaxiiMiddleware is parsing json or not with the type option.
Actually request continue but not with a json content.

At least it's my current understanding of https://expressjs.com/en/5x/api.html#express.json

xfournet (approved):

Issue #12616, opened 2025-10-02
Clicking on a relationship between a Relationship & Entity results in a blank page (null)


fruitcakej:
## Description
When I open a relationship between a relationship and an entity from ‘Data>Relationships’, the page does not work and redirects to ‘/dashboard/data/relationships/null/ etc

## Environment
1. 6.6 & after. Also confirmed on 6.8.2

## Reproducible Steps
See Screenshot

## Screenshots (optional)

<img width="1731" height="477" alt="Image" src="https://github.com/user-attachments/assets/5189e5d9-9c56-4d69-87ba-8811e35784ce" />

nino-filigran:
The issue is that the we redirect to the overview of the entity located at the left of the relation (the "From"). But given it's a relation, entity overview does not exists and therefore, the route is not valid.
If you create a relation "entity -> related to -> relation", you will not have this problem, since the from is an entity.
An additional problematic case would be to link to relations between them, but this case does not really sense business wise, as far as I'm aware off.

As a result, the fix should be to redirect to the "To" entity (the right part of the relation), when a relation is created from a relation to an entity (relation -> related to -> entity)

PR #12691, opened 2025-10-10
[frontend] redirection for relationships involving a relationship (#12616)

Archidoit:
### Proposed changes
Fix redirection in Data>Relationship for relationships involving a relationship as source

### Related issues
#12616


opencti-platform/opencti-front/src/utils/Entity.ts
```diff
@@ -162,11 +162,11 @@ export const computeLink = (node: {
       node.from.id
     }/knowledge/sightings/${node.id}`;
   } else if (node.relationship_type) {
-    if (node.from) { // 'from' not restricted
+    if (node.from && !node.from.parent_types?.includes('stix-core-relationship')) { // 'from' not restricted and not a relationship
```
JeremyCloarec: Since the above fragments might not always contain parent_types, we could use isRelationship from useSchema hook, since it only needs the entity_type, wdyt?

JeremyCloarec: In ItemEntityType.tsx, we also have an other way to check if it's a relationship from entity type:
Image
I'd say the hook solution is cleaner, but it looks like it's not used anywhere so I'm not sure if it's still working properly

lndrtrbn: `computeLink` is neither a hook or component, you cannot use a hook inside it. So to use the hook we could use an argument `isRelationsip` and call the hook everywhere the `computeLink` is called but that can be tedious

lndrtrbn: I think the better option is to transform `computeLink` into a hook and call `useSchema` inside as you suggest

Archidoit: thanks for your comments :) it's done !

opencti-platform/opencti-front/src/components/dashboard/WidgetListRelationships.tsx
```diff
@@ -50,7 +52,7 @@ const WidgetListRelationships = ({
         data={data.map(({ node }) => node)}
         globalCount={data.length}
         variant={DataTableVariant.widget}
-        useComputeLink={getRedirectionLink}
+        getComputeLink={getRedirectionLink}
```
Archidoit, line 55: renamed to avoid confusion

ValentinBouzinFiligran (approved):

PR #12695, opened 2025-10-10
[backend] improve task work status tracking and activeTasks retrieval 

### LINKED ISSUE #12521: StixCoreObjectActiveBackgroundTasksQuery timeout on huge reports
## Description

StixCoreObjectActiveBackgroundTasksQuery can be too long or too huge for big report (with 200k entities inside) => timeout and display "please download support package"

Developer console log on browser:
```
graphql:1   Failed to load resource: the server responded with a status of 504 ()
[NEW] Explain Console errors by using Copilot in Edge: click
         
         to explain an error. 
        Learn more
        Don't show again
front-4ESEXN3P.js:8  RRNLRequestError: Relay request for `StixCoreObjectActiveBackgroundTasksQuery` failed by the following reasons:

<html>
<head><title>504 Gateway Time-out</title></head>
<body>
<center><h1>504 Gateway Time-out</h1></center>
<hr><center>nginx</center>
```

## Environment

2. OpenCTI version: 6.7.20
4. Other environment details:

## Reproducible Steps

1. { e.g. Run ... }
2. { e.g. Click ... }
3. { e.g. Error ... }

JeremyCloarec:
### Proposed changes

* fix elFindByIds potentially failing when using null values
* add work_completed attribute to task to keep track of work completion
* modify stixCoreBackgroundActiveOperations to not use any post filtering but instead use work_completed value in filter to retrieve relevant tasks

### Related issues

*
*

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/lang/back/de.json
+++ b/opencti-platform/opencti-front/lang/back/de.json
@@ -956,6 +956,7 @@
   "Window title": "Fenstertitel",
   "Windows registry values": "Windows-Registrierungswerte",
   "Work": "Arbeit",
+  "Work completion status": "Status des Arbeitsabschlusses",
   "Work status": "Arbeitsstatus",
   "Workbench labels": "Workbench-Etiketten",
   "Workflow activated": "Workflow aktiviert",
--- a/opencti-platform/opencti-front/lang/back/en.json
+++ b/opencti-platform/opencti-front/lang/back/en.json
@@ -956,6 +956,7 @@
   "Window title": "Window title",
   "Windows registry values": "Windows registry values",
   "Work": "Work",
+  "Work completion status": "Work completion status",
   "Work status": "Work status",
   "Workbench labels": "Workbench labels",
   "Workflow activated": "Workflow activated",
--- a/opencti-platform/opencti-graphql/src/domain/stixCoreObject.js
+++ b/opencti-platform/opencti-graphql/src/domain/stixCoreObject.js
@@ -5,7 +5,6 @@ import {
   createRelationRaw,
   deleteElementById,
   distributionEntities,
-  stixLoadById,
   storeLoadByIdWithRefs,
   timeSeriesEntities
 } from '../database/middleware';
@@ -66,7 +65,6 @@ import {
   isEmptyField,
   isNotEmptyField,
   READ_ENTITIES_INDICES,
-  READ_INDEX_HISTORY,
   READ_INDEX_INFERRED_ENTITIES,
   READ_INDEX_INTERNAL_OBJECTS,
   UPDATE_OPERATION_ADD,
--- a/opencti-platform/opencti-graphql/src/domain/stixCoreObject.js
+++ b/opencti-platform/opencti-graphql/src/domain/stixCoreObject.js
@@ -10,9 +10,9 @@ import {
   timeSeriesEntities
 } from '../database/middleware';
 import {
+  fullEntitiesList,
   internalFindByIds,
   internalLoadById,
-  fullEntitiesList,
   pageEntitiesConnection,
   pageRegardingEntitiesConnection,
   storeLoadById,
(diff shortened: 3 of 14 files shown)


opencti-platform/opencti-front/lang/back/de.json
```diff
@@ -956,6 +956,7 @@
   "Window title": "Fenstertitel",
   "Windows registry values": "Windows-Registrierungswerte",
   "Work": "Arbeit",
+  "Work complete status": "Status \"Arbeit abgeschlossen",
```
xfournet: the " in the sentence is weird ?

opencti-platform/opencti-front/lang/back/en.json
```diff
@@ -956,6 +956,7 @@
   "Window title": "Window title",
   "Windows registry values": "Windows registry values",
   "Work": "Work",
+  "Work complete status": "Work complete status",
```
xfournet: "Work complete status" doesn't seems to be correct in english. -> "Work completion status" ?

opencti-platform/opencti-graphql/src/domain/stixCoreObject.js
```diff
@@ -137,37 +127,19 @@ const extractStixCoreObjectTypesFromArgs = (args) => {
...
-    let isConcerned = false;
-    if (backTask.type === 'QUERY') {
-      isConcerned = false;
-    }
-    if (backTask.type === 'LIST') {
-      isConcerned = (backTask.task_ids ?? []).includes(stixElement.extensions[STIX_EXT_OCTI].id);
-    }
-    if (isConcerned) {
-      activeTasks.push(backTask);
-    }
```
xfournet, line 168: How this filter is done in the new code ?

opencti-platform/opencti-graphql/src/domain/work.js
```diff
@@ -253,6 +254,17 @@ export const reportExpectation = async (context, user, workId, errorData) => {
...
+        const associatedTaskId = currentWork.background_task_id;
+        const associatedTask = await internalLoadById(context, user, associatedTaskId, { type: ENTITY_TYPE_BACKGROUND_TASK });
+        if (associatedTask) {
+          const sourceScriptUpdateWork = 'ctx._source["work_completed"] = "true"';
+          await elUpdate(associatedTask._index, associatedTaskId, { script: { source: sourceScriptUpdateWork, lang: 'painless' } });
+        } else {
+          logApp.warn('The task associated to work cannot be found in database, task work status cannot be updated.', { associatedTaskId });
+        }
+      }
       await elUpdate(currentWork._index, workId, { script: { source: sourceScript, lang: 'painless', params } });
```
xfournet: ease of read: elUpdate call and related script construction are separated by unrelated code. they should be grouped, either before or after the task update code.
is the order of the updates in ES important ?

opencti-platform/opencti-graphql/src/domain/work.js
```diff
@@ -253,6 +254,17 @@ export const reportExpectation = async (context, user, workId, errorData) => {
     // Update elastic
     const currentWork = await loadWorkById(context, user, workId);
     if (currentWork) {
+      // If work is associated to a task, we also need to update work to completed on the task
+      if (isComplete && currentWork.background_task_id) {
```
xfournet: in case of error (errorData is not undefined) the logic of updating the task is different than the one for the task.
it can then make some difference in stixCoreBackgroundActiveOperations call.
Is it expected ?

xfournet (changes requested):

opencti-platform/opencti-graphql/src/domain/stixCoreObject.js
JeremyCloarec: We now filter directly the task_ids on elastic, with the `{ key: ['task_ids'], values: [id] }` filter

opencti-platform/opencti-graphql/src/domain/work.js
JeremyCloarec: The order is not that important, but I'll change it so that the task is updated after the work, I think updating the work should have priority over the task

opencti-platform/opencti-graphql/src/domain/work.js
JeremyCloarec: For updating the associated task, we do not need to update anything in case of an errorData being there: the only thing we're tracking on the task is wether or not the work is complete (wether is was completing on an error or on a success)

xfournet: you can forget this comment, indeed the logic is the same for work and task, didn't notice the isComplete when building the work update script

opencti-platform/opencti-graphql/src/domain/work.js
```diff
@@ -254,6 +271,10 @@ export const reportExpectation = async (context, user, workId, errorData) => {
     const currentWork = await loadWorkById(context, user, workId);
     if (currentWork) {
       await elUpdate(currentWork._index, workId, { script: { source: sourceScript, lang: 'painless', params } });
+      // If work is associated to a task, we also need to update work to completed on the task
+      if (isComplete) {
+        await updateWorkTaskToComplete(context, user, currentWork);
```
richard-julien, line 276: Why using a specific function instead of adding the attribute directly in the elUpdate?

JeremyCloarec: You mean in the elUpdate just above? The elUpdate above is used to update the work and not the task.
Or did you mean that you would have preferred to inline the updateWorkTaskToComplete method directly in this method?

xfournet (approved):

opencti-platform/opencti-graphql/src/domain/work.js
richard-julien: Oh my bad, you're right its updating the task, forget my comment :)

Issue #12704, opened 2025-10-12
SSO/OIDC authentication broken after upgrading from 6.7.19 to 6.8.4


jercherng:
## Description

**Issue:** During OIDC authentication callbacks, `req.body` is `undefined`, causing:
- JavaScript tries to access `undefined.RelayState`
- This throws `TypeError: Cannot read properties of undefined`
- Authentication fails and users cannot log in via SSO
- User is greeted with 504 Gateway Time-out

## Environment

1. Linux
2. OpenCTI version: 6.7.19 to 6.8.4
3. OpenCTI client: backend (graphql)
4. Other environment details: SSO via OIDC

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Login via SSO
2. Complete SSO flow
3. Stuck without redirecting

## Expected Output

1. Redirect to dashboard `<insert-public-domain-here>/dashboard`
2. No RelayState error
3. No localhost:4000 redirect
4. No 504 timeout

## Actual Output
1. Get Error Message:
    ```
    {
        "category": "APP",
        "level": "error",
        "message": "[OPENCTI] Engine unhandled rejection",
        "reason": "TypeError: Cannot read properties of undefined (reading 'RelayState')\n    at /opt/opencti/build/back.js:4201:12609\n    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)",
        "source": "backend",
        "timestamp": "2025-10-11T22:31:25.263Z",
        "version": "6.8.4"
    }
    ```
3. Front-end responds with 504 Gateway Time-out
4. Refreshing the page actually takes you to the dashboard (authentication successful)

## Additional Information
---

The error occurs in the authentication callback route at line 493 of the source:

```javascript
// In httpPlatform.js line 493
const referer = req.body.RelayState ?? req.session.referer;
```

The code assumes `req.body` always exists, but OIDC callbacks may not have a body.

OIDC authentication flow:
1. User clicks "Login with SSO"
2. Browser redirects to identity provider
3. Identity provider authenticates user
4. Identity provider redirects back to `/auth/oic/callback`
5. **Problem:** Callback data location varies by provider:
   - SAML: Uses `req.body.RelayState`
   - OIDC: Often uses `req.query` or session storage
   - Some providers: Don't send state at all

Cursor-assisted RCA.

marieflorescontact:
may be related to this PR: https://github.com/OpenCTI-Platform/opencti/pull/12542

[...]

PR #12705, opened 2025-10-12
[backend] Add optionals chaining to safeguard against body-parser v2.0 return `undefined`

jercherng:
### Proposed changes
* Add optionals chaining to safely traverse the object chain without throwing errors 
* Pass `req` parameter to `getBaseUrl(...)` as a fallback for when `APP__BASE_URL` is not set.

### Related issues
* #12704 

Analysis:
Introduced in expressjs/body-parser 2.0:
> req.body is no longer always initialized to `{}`
> * it is left undefined unless a body is parsed

_Source: https://github.com/expressjs/body-parser/blob/master/HISTORY.md#200-beta1--2021-12-17_

Related Commit: https://github.com/OpenCTI-Platform/opencti/pull/10650
When renovate bot updated body-parser, it inadvertently caused the logic in [httpPlatform.js:L493](https://github.com/OpenCTI-Platform/opencti/blob/master/opencti-platform/opencti-graphql/src/http/httpPlatform.js#L493) to break. 

### Checklist

🤖 Co-Authored-By: Cursor <noreply@anthropic.com> 🤖

### DIFF
--- a/opencti-platform/opencti-graphql/src/http/httpPlatform.js
+++ b/opencti-platform/opencti-graphql/src/http/httpPlatform.js
@@ -34,8 +34,8 @@ import { createAuthenticatedContext } from './httpAuthenticatedContext';
 import { setCookieError } from './httpUtils';
 import { getChatbotProxy } from './httpChatbotProxy';
 
-export const sanitizeReferer = (refererToSanitize) => {
-  const base = getBaseUrl();
+export const sanitizeReferer = (refererToSanitize, req = null) => {
+  const base = getBaseUrl(req);
   if (!refererToSanitize) return base;
 
   const resolvedUrl = new URL(refererToSanitize, base).toString();
@@ -491,7 +491,7 @@ const createApp = async (app, schema) => {
       setCookieError(res, 'Invalid authentication, please ask your administrator');
     } finally {
       const referer = req.body.RelayState ?? req.session.referer;
-      const sanitizedReferer = sanitizeReferer(referer);
+      const sanitizedReferer = sanitizeReferer(referer, req);
       res.redirect(sanitizedReferer);
     }
   });
--- a/opencti-platform/opencti-graphql/src/http/httpPlatform.js
+++ b/opencti-platform/opencti-graphql/src/http/httpPlatform.js
@@ -490,7 +490,7 @@ const createApp = async (app, schema) => {
       logApp.error('Error auth provider callback', { cause: e, provider });
       setCookieError(res, 'Invalid authentication, please ask your administrator');
     } finally {
-      const referer = req.body.RelayState ?? req.session.referer;
+      const referer = req.body?.RelayState ?? req.session?.referer ?? null;
       const sanitizedReferer = sanitizeReferer(referer, req);
       res.redirect(sanitizedReferer);
     }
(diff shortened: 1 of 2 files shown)


jercherng:
> Thank you for your contribution, but we need you to sign your commits. Please see https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

~sorry.. thought I signed my commits already (one of the first things I did) but it is still showing as unverified. I will try to troubleshoot in the A.M.~ 

Fixed unsigned commit issue.

opencti-platform/opencti-graphql/tests/01-unit/http/httpPlatform-test.ts
```diff
@@ -81,4 +83,87 @@ describe('httpPlatform: sanitizeReferer function', () => {
...
+    it('should NOT throw TypeError when req.body is undefined', () => {
+      const req: any = { body: undefined };
+      
+      // The fix: req.body?.RelayState prevents crash
+      expect(() => {
+        const value = req.body?.RelayState;
+        return value;
+      }).not.toThrow();
+    });
+  });
```
labo-flg: this test is basically testing normal javascript behavior, we can remove it IMHO.

opencti-platform/opencti-graphql/tests/01-unit/http/httpPlatform-test.ts
```diff
@@ -81,4 +83,87 @@ describe('httpPlatform: sanitizeReferer function', () => {
...
+        protocol: 'https',
+        hostname: 'opencti.example.com',
+        headers: { host: 'opencti.example.com' }
+      };
+
+      // Test the fixed fallback chain
+      const referer = req.body?.RelayState ?? req.session?.referer ?? null;
+      expect(referer).toBe('/dashboard');
+    });
+  });
```
labo-flg: same here. Unless we call our own code to test, this is testing javascript normal behavior.

opencti-platform/opencti-graphql/src/http/httpPlatform.js
```diff
@@ -34,8 +34,8 @@ import { createAuthenticatedContext } from './httpAuthenticatedContext';
 import { setCookieError } from './httpUtils';
 import { getChatbotProxy } from './httpChatbotProxy';
 
-export const sanitizeReferer = (refererToSanitize) => {
-  const base = getBaseUrl();
+export const sanitizeReferer = (refererToSanitize, req = null) => {
+  const base = getBaseUrl(req);
```
lndrtrbn: Thanks for your contribution @jercherng. The fix on `referer` line is good for us.
Nonetheless we were wondering why you need to add the `req` here as parameter of `getBaseUrl`?

I think it is not necessary to fix the issue and we would prefer to have the minimal work in the PR with the only goal to fix the crash on body. Thanks you!

opencti-platform/opencti-graphql/tests/01-unit/http/httpPlatform-test.ts
jercherng: done, removed from `httpPlatform-test.ts`.

opencti-platform/opencti-graphql/tests/01-unit/http/httpPlatform-test.ts
jercherng: done, removed from `httpPlatform-test.ts`.

opencti-platform/opencti-graphql/src/http/httpPlatform.js
jercherng: agreed--PRs should be atomic. removed passing around `req` parameters.

marieflorescontact:
Since this bug was critical, it has been fixed by Filigran Team 🙂
https://github.com/OpenCTI-Platform/opencti/pull/12743
thank you for your contribution

PR #12728, opened 2025-10-13
[frontend] add warning when creating a widget using inferred relations (#10462)

### LINKED ISSUE #10462: Widget “Polar Area” incorrect in "Knowledge graph" perspective
## Description

When I use the Polar Area visualization in combination with the "Knowledge graph" perspective, I have wrong numbers.

## Environment

OCTI 6.5.10

## Reproducible Steps

1. Create a Polar Area widget with a "Knowledge graph" perspective.
2. Play with the filters. Example on testing:

Image

Image

-> Around 12K results

3. Verify the results in the "Data > Relationships" window. Example on testing: 

Image

-> 2,05K results

## Expected Output

I want to get the right number of results on the widget.

## Actual Output

In my example, the widget display around 12K results when I have only 2K objects in my platform.

## Additional information

The widget seems to work with Entities perspective. Example on testing:

Image

Image

Image

OctaveLaventure:
### Proposed changes

- Alert component can use all severity levels
- add warning message on widgets creation informing the user that the displayed data will take inferred relations into account
- improve widget visualization typing

### Related issues

* #10462

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/components/AlertInfo.tsx
+++ b/opencti-platform/opencti-front/src/components/AlertInfo.tsx
@@ -1,28 +1,28 @@
-import Alert from "@mui/material/Alert";
-import type { AlertProps } from "@mui/material/Alert";
-import Typography from "@mui/material/Typography";
-import React, { CSSProperties, ReactNode } from "react";
-import { useTheme } from "@mui/styles";
-import type { Theme } from "./Theme";
+import MuiAlert from '@mui/material/Alert';
+import type { AlertProps } from '@mui/material/Alert';
+import Typography from '@mui/material/Typography';
+import React, { CSSProperties, ReactNode } from 'react';
+import { useTheme } from '@mui/styles';
+import type { Theme } from './Theme';
 
 type AlertInfoProps = {
-  content: string | ReactNode;
+  content: ReactNode;
   style?: CSSProperties;
-  severity?: AlertProps["severity"];
+  severity?: AlertProps['severity'];
 };
 
-const AlertInfo = ({ content, style, severity = "info" }: AlertInfoProps) => {
+const AlertInfo = ({ content, style, severity = 'info' }: AlertInfoProps) => {
   const theme = useTheme<Theme>();
 
   return (
     <div style={style}>
-      <Alert
+      <MuiAlert
         severity={severity}
         variant="outlined"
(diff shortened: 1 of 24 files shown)


MTorbay-Filigran (dismissed):

opencti-platform/opencti-front/src/components/AlertInfo.tsx
```diff
@@ -1,28 +1,28 @@
...
 type AlertInfoProps = {
-  content: string | ReactNode
-  style?: CSSProperties
+  content: ReactNode;
+  style?: CSSProperties;
+  severity?: AlertProps['severity'];
 };
 
-const AlertInfo = ({ content, style }: AlertInfoProps) => {
+const AlertInfo = ({ content, style, severity = 'info' }: AlertInfoProps) => {
```
lndrtrbn: I think we should rename it `Alert` now that it is not only info anymore

opencti-platform/opencti-front/src/utils/widget/widgetUtils.tsx
```diff
@@ -199,7 +201,10 @@ const widgetVisualizationTypes = [
     isEntities: true,
     isAudits: true,
   },
-];
+] as const;
+
+export type WidgetVisualizationTypes =
```
lndrtrbn, line 206: I'm not sure to understand the gain to make a type for the list of possible widgets as we have to use a lot of hard-casts after that. Could you explain a bit?

OctaveLaventure: I think we gain more clarity, the cases where an empty string value is possible are more explicit for example which reduces the risk of forgetting to handle it, it avoids typos in hardcoded widget names, provides autocomplete in our IDE and highlights issues such as impossible or unnecessary comparisons, even through the type cast.
I agree that ideally we'd like to avoid these casts in the future, and the fact that they are uncomfortable kind of nudges us toward fixing the functions that use them during refactorings instead of continuing to use a string type.
I glanced over the estimated work it would require to type it all the way up, and I think fixing everything here would sadly be too big of a refacto for such a small PR.

opencti-platform/opencti-front/src/components/AlertInfo.tsx
OctaveLaventure: done

lndrtrbn (approved):

Issue #10964, opened 2025-05-07
Narrative views buttons inconsistancy


Archidoit:
## Description
When selecting a narrative view, the selection color is different according to the button selected

## Reproducible Steps
- Go to the narratives list
- Select the 'lines view', the button background is greener:

![Image](https://github.com/user-attachments/assets/c403ea14-5625-4c3b-9100-71a03228a931)

- Select the 'subnarratives view', the subnarratives button image is green, and the 'lines view' icon is not the same as in the previous configuration:

![Image](https://github.com/user-attachments/assets/5a9484a2-8c5c-4729-9774-44429edaf715)

PR #12729, opened 2025-10-13
[frontend] Narrative views buttons inconsistancy (#10964)

SarahBocognano:
### Proposed changes

* I took the liberty to change the icon for subentities view as I think it was awful and not coherent to what we use in the platform.

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/10964

### DIFF
--- a/opencti-platform/opencti-front/src/components/ViewSwitchingButtons.tsx
+++ b/opencti-platform/opencti-front/src/components/ViewSwitchingButtons.tsx
@@ -1,11 +1,10 @@
 import React, { FunctionComponent } from 'react';
 import ToggleButton from '@mui/material/ToggleButton';
 import Tooltip from '@mui/material/Tooltip';
-import { LibraryBooksOutlined, ViewModuleOutlined } from '@mui/icons-material';
+import { LibraryBooksOutlined, ViewListOutlined, ViewModuleOutlined } from '@mui/icons-material';
 import { FormatListGroup, Group, RelationManyToMany, VectorPolygon } from 'mdi-material-ui';
 import { ToggleButtonGroup } from '@mui/material';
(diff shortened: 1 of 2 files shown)


Archidoit:
If you change the subentities and list icon, I thing you should also change it there (malware > Knowledge > Attack pattern) : 
image

Archidoit:
and there (Malware > analyses)  : 
image

Archidoit:
By the way, can you fix the same inconsistancy in the Analyses tab of stix domain objects , the 'list' icon changes when we switch between list view and graph view : 

image


image

SarahBocognano:
Oh ok so wait I need to ask product what icon do they want in the first place. Because originally it was like this : 
Capture d’écran 2025-10-13 à 15 05 36


I though it was weird to have those two icon next to each other, don't seems to be from the same "icon base" 
What do you want @alice-debra ?

alice-debra:
@SarahBocognano I'm having trouble finding the screenshots that show the changes made. Could you please clarify their location?

SarahBocognano:
Those are my changes : 

Capture d’écran 2025-10-13 à 15 21 32
Capture d’écran 2025-10-13 à 15 21 44

SarahBocognano:
I can either go back to the previous sub item icon with changing the list icon to match the ones in Cathia screenshots.
Or change it as I did

alice-debra:
That doesn't look like a big change, and it's cleaner. Have you changed it everywhere?

SarahBocognano:
Not yet, but I will if you give me the go

alice-debra:
Capture d’écran 2025-10-13 à 15 40 51
Did you change this one too?

SarahBocognano:
Not yet no. originally the one you present me wasn't here. It was the clean list like in my screenshot

CelineSebe (approved):

PR #12731, opened 2025-10-13
[worker] add connector_type filtering in worker 

### LINKED ISSUE #12490: Option to dedicate workers to specific queues
## Use case

Analysts may observe significant delays (sometimes hours) in ingestion after validating Drafts. We need to optimize Draft/Workbench ingestion priorization. 

Currently:

Draft validations are handled like playbooks, background tasks, and feeds. After approval, bundles are sent to ingestion through the push_draft queue. Drafts are therefore managed through queues like any other ingestion. However, there is no way to force a worker to listen to only one queue.

## Workaround

Nothing

## Proposed Solution

Add an option (e.g. a category) in the worker configuration to allow to dedicate workers to specific queues.

This would ensure that Drafts (or other critical tasks) can be ingested without being impacted by other ongoing ingestion processes.

No

JeremyCloarec:
### Proposed changes

* add ability to filter connectors to handle in worker: this could enable deploying workers only used for draft, or other workers only used for tasks etc...
*
Linked to https://github.com/OpenCTI-Platform/client-python/pull/991
Linked to https://github.com/OpenCTI-Platform/docs/pull/362
### Related issues

*
*

### Checklist

### DIFF
--- a/opencti-worker/src/worker.py
+++ b/opencti-worker/src/worker.py
@@ -596,6 +596,12 @@ class Worker:  # pylint: disable=too-few-public-methods, too-many-instance-attri
             config,
             default="",
         )
+        self.targeted_connector_types = get_config_variable(
+            "WORKER_TARGETED_CONNECTOR_TYPES",
+            ["worker", "targeted_connector_types"],
+            config,
+            default=None,
+        )
         # Telemetry
         self.telemetry_enabled = get_config_variable(
             "WORKER_TELEMETRY_ENABLED",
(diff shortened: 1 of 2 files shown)


opencti-worker/src/worker.py
```diff
@@ -596,6 +596,12 @@ def __post_init__(self) -> None:
             config,
             default="",
         )
+        self.targeted_connector_types = get_config_variable(
```
xfournet, line 599: we could rather use a separate thread pool for internal queues in order to keep unspecialized workers

Issue #11916, opened 2025-08-06
Filters: when chosing lower or equal, the selection behaves in an odd way


nino-filigran:
## Description
When I select a date field to filter on, using the operator lower than/equal, the date I chose is always one day after. However, in the filter value, the correct data seems to be passed.

## Environment

Testing

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Go in report
2. Add any date filter (modification date/original creation date...)
3. select the operator: lower than/equal
4. Select for instance the 07th of July
5. Click again on the filter 

## Expected Output

The value displayed in the filter to be the 07th of july & after clicking on the filter, to also see 07th of july

## Actual Output

The value displayed in the filter to be the 07th of july & after clicking on the filter, to also see 08th of july

## Additional information
In the below screenshot, I've selected the 11th of july, as shown in the filter value above. But the calendar displays 12th

<img width="433" height="427" alt="Image" src="https://github.com/user-attachments/assets/25d8ff05-2e59-4368-a99e-620ff0b0d206" />

JeremyCloarec:
Linked to https://github.com/OpenCTI-Platform/opencti/issues/11921

PR #12740, opened 2025-10-13
[frontend] date filters value display in date selector for lte and gt operators (#11916)

Archidoit:
### Proposed changes
- Align date filter values display between filter icon button and date selection calendar
- Frontend utils function to convert a filter date value in a date value displayed in frontend, depending on the operator.

### Related issues
#11916

### DIFF
--- a/opencti-platform/opencti-front/src/utils/Time.js
+++ b/opencti-platform/opencti-front/src/utils/Time.js
@@ -1,4 +1,5 @@
 import moment from 'moment-timezone';
+import { subDays } from 'date-fns';
 import { isNone } from '../components/i18n';
 
 const defaultDateFormat = 'YYYY-MM-DD';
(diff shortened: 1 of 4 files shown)


opencti-platform/opencti-front/src/utils/Time.js
```diff
@@ -199,3 +200,17 @@ export const formatUptime = (uptimeInSeconds, t_i18n) => {
...
 };
+
+/**
+ * Convert a date value stored in filters to a date value displayed in frontend according to the operator
+ *
+ * @param dateValue - The date value stored in filters
+ * @param operator - The associated filter operator
+ * @returns The date value to be displayed
+ */
+export const dateFiltersValueForDisplay = (dateFilterValue, filterOperator) => {
```
OctaveLaventure, line 211: **1. Testing**
Maybe add a little test for this function ?

OctaveLaventure, line 211: **2. Question**
Initially I thought "we should type this function", but I see this is a JS file.

I guess there is a conflict between the rule "new code should be in TS" and "this file is too big for me to migrate it now", which makes some new code be added in old JS in cases like this. Do I understand correctly ?

Maybe we could for example add new code in a `Time.ts` file, even if it duplicates files a bit, at least it would be typed and lay down the path for later migrations. What do you think?

Also, where/when is the best place to discuss this kind of processes?

OctaveLaventure (approved):

opencti-platform/opencti-front/src/utils/Time.js
Archidoit: The rule is are the following:
New files should be in ts / tsx
If you add a minor change to a js(x) file: no type
If you need to completely rework a file : convert to ts(x)

Archidoit: We don't do such files duplication for types.

Archidoit: It can be something to suggest during our monthly dev meetings :)

Archidoit: done :)

Issue #9146, opened 2024-11-26
[CI] Enable codecov metrics for opencti-front


aHenryJard:
## Use case

Codecov is watching code coverage only for `opencti-graphql` package, we are starting to write frontend tests and e2e tests, it would be great to know the coverage and moreover the coverage addition or regression on PRs.

I think that basically we have to
- enable coverage on frontend e2e and have an output in a file
- give this output file to codecov

If easy includes frontend unit tests too.

## If the feature request is approved, would you be willing to submit a PR?

Yes / No (Help can be provided if you need assistance submitting a PR)

labo-flg:
@OctaveLaventure @aHenryJard  is there any more steps to implement for this task ?

OctaveLaventure:
@labo-flg In my opinion no. The initial issue mentioned E2E tests though, and only unit tests are sending coverage on the front. I think coverage on E2E tests is unnecessary, but I'm open to dicussion. Feel free to keep it going or to close the issue @aHenryJard 🙂

aHenryJard:
Let's say it's done :) thanks !

PR #12742, opened 2025-10-13
[CI] set up codecov coverage reporting on opencti-front

OctaveLaventure:
### Proposed changes

- add `yarn test:coverage` command to opencti-front
- send coverage reports to codecov (on Drone CI for now to keep coverage-related tasks together)
- leverage flags to split coverage reports between `opencti-front`, `opencti-graphql` and global `opencti` coverage
- [x] Note: remove `when:` clause before merge in `.drone.yml`

### Related issues

- https://github.com/OpenCTI-Platform/opencti/issues/9146

### Checklist

### Further comments

I didn't add E2E tests to the coverage for now, I think we can discuss the relevance of E2E tests coverage in the CI workgroup.


.drone.yml
```diff
@@ -172,6 +175,23 @@ steps:
     depends_on:
       - frontend-tests
 
+  - name: frontent-coverage
```
aHenryJard: Could you use the coverage step to send both backend and frontend files to codecov ? And then remove the "failure: ignore" also.

OctaveLaventure: - I think we would lose the possibility to use flags to differenciate coverage between `opencti-graphql` and `opencti-front` to gain ~10 seconds of CI time, not sure it's a good trade.
- ✔️ `failure: ignore` removed in both steps

aHenryJard: you can't remove failure: ignore on api-test if you keep 2 steps; because if api-test step is skipped, there is no codecov file and it's failing but the overall build should still succeed.

I'm fine with keeping 2 steps, I don't have a strong opinion on that part.

OctaveLaventure: Done.
On a side note I also tried using an additional step for the skip as I suggested by DM but with the current limitations of drone it doesn't work, either api-test is triggered or we end up with a failed build, and no way to customize the codecov plugin to run conditionally, so let's just roll with what we have now.

aHenryJard (approved):
Good for me

Issue #12796, opened 2025-10-14
Blank space in Observed Data, Entities/Observables tab


Archidoit:
## Description
Go under an Observed Data, under the Entities or the Observables tab. There is a blank space on the right : 

<img width="1914" height="568" alt="Image" src="https://github.com/user-attachments/assets/34dbd08f-0f4a-4716-bb5d-71b4da5256ac" />

nino-filigran:
duplicate of: https://github.com/OpenCTI-Platform/opencti/issues/12088

Issue #12088, opened 2025-08-20
Misalignement in observed data tabs (entities / observables)


richard-julien:
## Description

Right panel still taking place but not visible in entities / observables tabs

<img width="1798" height="358" alt="Image" src="https://github.com/user-attachments/assets/a3ed3633-7d65-49f2-b591-870a93f4605d" />

PR #12797, opened 2025-10-14
[frontend] Observed Data right padding for Entities/Observables tabs (#12088)

Archidoit:
### Proposed changes
- Remove blank space on the Right of the 'entites' and 'observables' tab for Observed Data
- refacto ObservedData Root file to tsx

### Related issues
#12796 

#12088


opencti-platform/opencti-front/src/private/components/events/observed_data/Root.tsx
```diff
@@ -0,0 +1,230 @@
...
+
+  useSubscription<RootObservedDataSubscription>(subConfig);
+
+  const {
+    observedData,
+    connectorsForExport,
+    connectorsForImport,
+  } = usePreloadedQuery<RootObservedDataQuery>(observedDataQuery, queryRef);
+
+  return (
```
esrevi, line 94: Small suggestion:  an early return here would make this a bit cleaner imo:

```tsx
if (!observedData) {
  return <ErrorNotFound />;
}

return (
    <div>
      <Breadcrumbs elements={[
        { label: t_i18n('Events') },
        { label: t_i18n('Observed datas'), link: '/dashboard/events/observed_data' },
        { label: observedData.name, current: true },
      ]}
      /> .....
     </div>
)
```

This reduces the nesting and you can drop some unecessary fragments too I think

esrevi (approved):
lgtm !

opencti-platform/opencti-front/src/private/components/events/observed_data/Root.tsx
Archidoit: done :) thanks

PR #12803, opened 2025-10-14
[backend] buildDataRestrictions perf improve on markings condition 

JeremyCloarec:
### Proposed changes

* simplified the logic of marking condition in buildDataRestrictions: we now get the list of markings that the user doesn't have access to, and we use this list in a terms query in a must_not
*

### Related issues

*
*

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/database/engine.js
+++ b/opencti-platform/opencti-graphql/src/database/engine.js
@@ -509,32 +509,23 @@ export const buildDataRestrictions = async (context, user, opts = {}) => {
     } else {
       const allMarkings = await getEntitiesListFromCache(context, SYSTEM_USER, ENTITY_TYPE_MARKING_DEFINITION);
       // Markings should be grouped by types for restriction
-      const userGroupedMarkings = R.groupBy((m) => m.definition_type, user.allowed_marking);
-      const allGroupedMarkings = R.groupBy((m) => m.definition_type, allMarkings);
-      const markingGroups = Object.keys(allGroupedMarkings);
       const mustNotHaveOneOf = [];
-      for (let index = 0; index < markingGroups.length; index += 1) {
-        const markingGroup = markingGroups[index];
-        const markingsForGroup = allGroupedMarkings[markingGroup].map((i) => i.internal_id);
-        const userMarkingsForGroup = (userGroupedMarkings[markingGroup] || []).map((i) => i.internal_id);
-        // Get all markings the user has no access for this group
-        const res = markingsForGroup.filter((m) => !userMarkingsForGroup.includes(m));
-        if (res.length > 0) {
-          mustNotHaveOneOf.push(res);
+      const userMarkingsIds = new Set(user.allowed_marking.map((m) => m.internal_id));
+      for (let index = 0; index < allMarkings.length; index += 1) {
+        const marking = allMarkings[index];
+        const markingId = marking.internal_id;
+        if (!userMarkingsIds.has(markingId)) {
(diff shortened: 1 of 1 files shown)


opencti-platform/opencti-graphql/src/database/engine.js
```diff
@@ -509,32 +509,21 @@ export const buildDataRestrictions = async (context, user, opts = {}) => {
     } else {
       const allMarkings = await getEntitiesListFromCache(context, SYSTEM_USER, ENTITY_TYPE_MARKING_DEFINITION);
       // Markings should be grouped by types for restriction
```
SouadHadjiat: so you don't group by definition_type anymore ?

opencti-platform/opencti-graphql/src/database/engine.js
```diff
@@ -509,32 +509,21 @@ export const buildDataRestrictions = async (context, user, opts = {}) => {
...
-        mustNotMarkingTerms.push({
-          bool: {
-            should,
-            minimum_should_match: 1,
-          },
-        });
-      }
+      const mustNotMarkingTerms = [{
+        terms: {
+          [buildRefRelationSearchKey(RELATION_OBJECT_MARKING)]: mustNotHaveOneOf
```
SouadHadjiat, line 524: definitely a good idea to use terms query instead of match

opencti-platform/opencti-graphql/src/database/engine.js
JeremyCloarec: unless I missed something, I think it was completely useless

opencti-platform/opencti-graphql/src/database/engine.js
```diff
@@ -507,34 +507,23 @@ export const buildDataRestrictions = async (context, user, opts = {}) => {
       // If user have no marking, he can only access to data with no markings.
       must_not.push({ exists: { field: buildRefRelationKey(RELATION_OBJECT_MARKING) } });
     } else {
+      // Compute all markings that the user doesnt have access to
       const allMarkings = await getEntitiesListFromCache(context, SYSTEM_USER, ENTITY_TYPE_MARKING_DEFINITION);
```
xfournet, line 511: i wonder why we removing the non allowed marking (that seems complex) rather than allowing the allowed marking ?

JeremyCloarec: That's because the user needs to have access to all markings that are on en entity, not only some of them. 
As an example, let's take an entity that has TLP:RED and PAP:RED marking. If the user has access to TLP:RED but only PAP:AMBER, the user shouldn't have access to the entity: he has access to the TLP marking, but not the PAP one

xfournet: does user.allowed_marking contains all the 'infered' marking level, eg if a user has access to TLP:RED he can also access to TLP:WHITE. so there is no case that allowed_marking contains TLP:RED but not TLP:WHITE ?

xfournet: look likes it's handled in computeAvailableMarkings right ?

xfournet (approved):

opencti-platform/opencti-graphql/src/database/engine.js
JeremyCloarec: yes you're correct, the lower level markings are added in computeAvailableMarkings so the markings in the users contains every markings the user has access to (and not only the configured allowed markings)

PR #12809, opened 2025-10-15
[frontend] change StixCyberObservableNestedEntitiesTable ref link (#12612)

### LINKED ISSUE #12612: Clicking on a Nested object in a Network Traffic observable results in an empty page.
## Description
If you are in a NetworkTraffic object and click on a nested object in the ui. The knowledge tab will be activated but the tab will be empty.
The resulting url is http://<octi-platform>/dashboard/observations/observables/{observable_id}/knowledge/relations/{relation_id}}

You get on the same URL if you go to the knowledge tab yourself by clicking on the tab, and then clicking on the nested object.

## Environment
1. 6.8.2

## Additional information
It looks like the path from NetworkTraffic observable → Nested object is broken.

JeremyCloarec:
### Proposed changes

* clicking on a nested ref in observables now redirects to the targeted entity instead of trying to target the relation that can't be displayed in a page
*

### Related issues

* #12612
*

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/observations/stix_cyber_observables/StixCyberObservableNestedEntitiesTable.tsx
+++ b/opencti-platform/opencti-front/src/private/components/observations/stix_cyber_observables/StixCyberObservableNestedEntitiesTable.tsx
@@ -19,7 +19,7 @@ import ItemIcon from '../../../../components/ItemIcon';
 import { StixCyberObservableNestedEntitiesTable_node$data } from './__generated__/StixCyberObservableNestedEntitiesTable_node.graphql';
 import { useBuildEntityTypeBasedFilterContext } from '../../../../utils/filters/filtersUtils';
 import stopEvent from '../../../../utils/domEvent';
-import { computeLink } from '../../../../utils/Entity';
+import { computeLink, ComputeLinkNode } from '../../../../utils/Entity';
 
 const LOCAL_STORAGE_KEY = 'StixCyberObservableNestedEntitiesTable';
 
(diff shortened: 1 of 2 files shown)


opencti-platform/opencti-front/src/private/components/observations/stix_cyber_observables/StixCyberObservableNestedEntitiesTable.tsx
```diff
@@ -338,6 +339,15 @@ const StixCyberObservableNestedEntitiesTable: React.FC<StixCyberObservableNested
     },
   };
 
+  const getRedirectionLink = (stixObject: StixCyberObservableNestedEntitiesTable_node$data) => {
+    const targetObject = stixObject.from?.id === stixCyberObservableId ? stixObject.to : stixObject.from;
+    if (targetObject) {
+      const reconstructedTarget = { ...targetObject, id: targetObject.id ?? '', entity_type: targetObject.entity_type ?? '' };
```
lndrtrbn: I do not understand why we need to do this. How the URL would work if there is no id or entity type?

JeremyCloarec: The issue is that the id & entity_type in the data are typed as potentially undefined in the fragment, since id & entity_type are retrieved in ... on BasicObject. So it's typed as potentially undefined, but in reality it will always be defined since the entities we're retrieving are always inherited from BasicObject.

It might not be the best solution, but it was the first idea I had to ensure that the object we're using in computeLink has the correct typing

JeremyCloarec: I pushed a new commit with a more proper fix for the typing

lndrtrbn: I was thinking of using a if looking like:
`if (targetObject && targetObject.id && targetObject.entity_type)`
but I am not sure it fixes your issue

lndrtrbn (approved):

PR #12813, opened 2025-10-15
[backend] add telemetry on human user login (#12807)(#12916)

### LINKED ISSUE #12807: Add telemetry on Login
## Use case

Ability to know the amount of users who logged in the platform (using login form) to check usage of trial license.

The goal is to count approximatively "human" user, not account for connector or other technical account.
API call with token should not be count.
Both user login with SSO or Local strategy should be count.

### LINKED ISSUE #12916: Telemetry: count user that are service account in a separate gauge
## Use case

usersCount: count all user whatever the status or the type service account or not.

We need to count service account in one side and other user in another count.

aHenryJard:
### Proposed changes

* Add a telemetry point when a user login into the platform
* Split in telemetry the count of user that are not service account from service account

### Related issues

* closes #12807 
* closes #12916

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/manager/telemetryManager.ts
+++ b/opencti-platform/opencti-graphql/src/manager/telemetryManager.ts
@@ -63,7 +63,7 @@ export const TELEMETRY_FORM_INTAKE_CREATED = 'formIntakeCreatedCount';
 export const TELEMETRY_FORM_INTAKE_UPDATED = 'formIntakeUpdatedCount';
 export const TELEMETRY_FORM_INTAKE_DELETED = 'formIntakeDeletedCount';
 export const TELEMETRY_FORM_INTAKE_SUBMITTED = 'formIntakeSubmittedCount';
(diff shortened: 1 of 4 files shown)


opencti-platform/opencti-graphql/src/manager/telemetryManager.ts
```diff
@@ -63,6 +63,7 @@ export const TELEMETRY_FORM_INTAKE_CREATED = 'formIntakeCreatedCount';
 export const TELEMETRY_FORM_INTAKE_UPDATED = 'formIntakeUpdatedCount';
 export const TELEMETRY_FORM_INTAKE_DELETED = 'formIntakeDeletedCount';
 export const TELEMETRY_FORM_INTAKE_SUBMITTED = 'formIntakeSubmittedCount';
+export const TELEMETRY_USER_LOGIN = 'formIntakeSubmittedCount';
```
aHenryJard:
```suggestion
export const TELEMETRY_USER_LOGIN = 'userLoginCount';
```

delemaf (approved):

Issue #7279, opened 2024-06-09
In search / bulk search, have a dropdown to be able to modify / delete and checkboxes for massive operations


SamuelHassine:
## Use case

In search / bulk search, have a dropdown to be able to modify / delete and checkboxes for massive operations

Lhorus6:
Note: 

If we want to enable massive enrichment, here is an example of implementation (interesting point: concatenated view of the results of the different enrichments): https://github.com/stanfrbd/cyberbro?tab=readme-ov-file

PR #12817, opened 2025-10-15
[frontend] Bulk Search refacto: Separated data table for known/unknown entities (#7279)

Archidoit:
### Proposed changes
In Bulk Search
- code refacto
- turn list into data tables
- separate known and unknown entities in 2 different tabs
- add filters and columns ordering for known entities
- display the number of unknown entities

### Related issues
#7279


maelv-filigran:
you can use `git mv` to preserve the git history of the file when converting jsx to tsx ;)

Archidoit:
> you can use `git mv` to preserve the git history of the file when converting jsx to tsx ;)

Okay thanks, but the jsx and tsx files are almost completely different in this case :)

maelv-filigran:
The old JSX version used RxJS with a 500ms debounce:
```javascript
const SEARCH$ = new Subject().pipe(debounce(() => timer(500)));
```
This debounce feature was lost in the refactoring. 
I think this is an important point that should be corrected before merging this PR.

### because:

1. __Performance__ - Each keystroke immediately triggers:

   - Text parsing/transformation (split, map, flat, join)
   - Recalculation of `values`
   - Rendering of SearchBulk and SearchBulkUnknownEntities components
   - Potentially GraphQL queries (with up to 5,000 elements)

2. __Unnecessary network requests__ - If the user types quickly or pastes a large amount of text, this can generate many unnecessary intermediate GraphQL queries

3. __Degraded user experience__ - The interface can become slow/jerky with large lists

maelv-filigran (approved):

PR #12819, opened 2025-10-15
[backend] remove pagination in elFindByIds (#12810)

### LINKED ISSUE #12810: Remove pagination in elFindByIds
## Description

Currently pagination is used to handle large batch of ids to load.
Instead of pagination, we should rather 'paginate' the ids batch before doing elastic queries, so we save a lot of data send to elastic and we can avoid the pagination step

JeremyCloarec:
### Proposed changes

* remove pagination in elFindByIds. Instead, use the already in place system of splitting the query in multiple sub queries with a maximum of ids lower than ES_MAX_PAGINATION.
*

### Related issues

* #12810 
*

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/database/engine.js
+++ b/opencti-platform/opencti-graphql/src/database/engine.js
@@ -1750,7 +1750,7 @@ export const elFindByIds = async (context, user, ids, opts = {}) => {
   const computedIndices = getIndicesToQuery(context, user, queryIndices);
   const hits = [];
   // Leave room in split size compared to max pagination to minimize data loss risk in case of duplicated ids in database
-  const splitSize = Math.max(ES_MAX_PAGINATION - 100, ES_DEFAULT_PAGINATION);
+  const splitSize = Math.max(ES_MAX_PAGINATION / 2, ES_DEFAULT_PAGINATION);
   const groupIds = R.splitEvery(splitSize, processIds);
   for (let index = 0; index < groupIds.length; index += 1) {
     const mustTerms = [];
@@ -1826,7 +1826,10 @@ export const elFindByIds = async (context, user, ids, opts = {}) => {
     });
     const elements = data.hits.hits;
     if (elements.length > workingIds.length) {
-      logApp.warn('Search query returned more elements than expected', { ids: workingIds });
+      logApp.info('Search query returned more elements than expected', { ids: workingIds });
+      if (elements.length >= ES_MAX_PAGINATION) {
+        throw DatabaseError('Ids loading returned more elements than paging allowed for, some elements could not be loaded', { ids: workingIds });
+      }
     }
     if (elements.length > 0) {
       const convertedHits = await elConvertHits(elements);
--- a/opencti-platform/opencti-graphql/src/database/engine.js
+++ b/opencti-platform/opencti-graphql/src/database/engine.js
@@ -1732,7 +1732,7 @@ const findElementsDuplicateIds = (elements) => {
   const elementIds = new Set();
   const checkCurrentIds = (id) => {
     if (!id) return;
-    if (elementIds.has(id)) {
+    if (elementIds.has(id) && !duplicatedIds.includes(id)) {
       duplicatedIds.push(id);
     } else {
       elementIds.add(id);
--- a/opencti-platform/opencti-graphql/src/database/engine.js
+++ b/opencti-platform/opencti-graphql/src/database/engine.js
@@ -1739,8 +1739,6 @@ export const elFindByIds = async (context, user, ids, opts = {}) => {
     mapWithAllIds = false,
     type = null,
     relCount = false,
-    orderBy = null,
-    orderMode = 'asc',
   } = opts;
   const idsArray = Array.isArray(ids) ? ids : [ids];
   const types = (Array.isArray(type) || isEmptyField(type)) ? type : [type];
@@ -1807,9 +1805,6 @@ export const elFindByIds = async (context, user, ids, opts = {}) => {
       body.script_fields = {
         script_field_denormalization_count: REL_COUNT_SCRIPT_FIELD
       };
-    }
-    if (isNotEmptyField(orderBy)) {
-      body.sort = [{ [orderBy]: orderMode }];
     }
     const _source = { excludes: [] };
     if (withoutRels) _source.excludes.push(`${REL_INDEX_PREFIX}*`);
--- a/opencti-platform/opencti-graphql/src/database/engine.js
+++ b/opencti-platform/opencti-graphql/src/database/engine.js
@@ -1826,9 +1826,20 @@ export const elFindByIds = async (context, user, ids, opts = {}) => {
     });
     const elements = data.hits.hits;
     if (elements.length > workingIds.length) {
-      logApp.info('Search query returned more elements than expected', { ids: workingIds });
+      const duplicatedIds = [];
+      for (let i = 0; i < workingIds.length; i += 1) {
+        const workingId = workingIds[i];
+        const associatedElement = elements.filter((e) => e[internalId.name] === workingId
+            || e[standardId.name] === workingId
+            || e[xOpenctiStixIds.name]?.includes(workingId)
+            || e[iAliasedIds.name]?.includes(workingId));
+        if (associatedElement.length > 1) {
+          duplicatedIds.push(workingId);
+        }
+      }
+      logApp.info('Search query returned more elements than expected', { resultCount: elements.length, queryCount: workingIds.length, duplicatedIds });
       if (elements.length >= ES_MAX_PAGINATION) {
-        throw DatabaseError('Ids loading returned more elements than paging allowed for, some elements could not be loaded', { ids: workingIds });
+        throw DatabaseError('Ids loading returned more elements than paging allowed for, some elements could not be loaded', { resultCount: elements.length, queryCount: workingIds.length, duplicatedIds });
       }
     }
     if (elements.length > 0) {
--- a/opencti-platform/opencti-graphql/src/database/engine.js
+++ b/opencti-platform/opencti-graphql/src/database/engine.js
@@ -1730,10 +1730,18 @@ const REL_COUNT_SCRIPT_FIELD = {
 // elFindByIds is not defined to use ordering or sorting (ordering is forced by creation date)
 // It's a way to load a bunch of ids and use in list or map
 export const elFindByIds = async (context, user, ids, opts = {}) => {
-  const { indices, baseData = false, baseFields = [] } = opts;
-  const { withoutRels = true, toMap = false, mapWithAllIds = false, type = null } = opts;
-  const { relCount = false } = opts;
-  const { orderBy = null, orderMode = 'asc' } = opts;
+  const {
+    indices,
+    baseData = false,
+    baseFields = [],
+    withoutRels = true,
+    toMap = false,
+    mapWithAllIds = false,
+    type = null,
+    relCount = false,
+    orderBy = null,
+    orderMode = 'asc',
+  } = opts;
   const idsArray = Array.isArray(ids) ? ids : [ids];
   const types = (Array.isArray(type) || isEmptyField(type)) ? type : [type];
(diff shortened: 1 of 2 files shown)


opencti-platform/opencti-graphql/src/database/engine.js
```diff
@@ -1733,7 +1732,7 @@ export const elFindByIds = async (context, user, ids, opts = {}) => {
   const { indices, baseData = false, baseFields = [] } = opts;
   const { withoutRels = true, toMap = false, mapWithAllIds = false, type = null } = opts;
   const { relCount = false } = opts;
-  const { orderBy = 'created_at', orderMode = 'asc' } = opts;
+  const { orderBy = null, orderMode = 'asc' } = opts;
```
xfournet: take the opportunity to make it clearer ;

```
  const {
    indices,
    baseData = false,
    baseFields = [],
    withoutRels = true,
    toMap = false,
    mapWithAllIds = false,
    type = null,
    relCount = false,
    orderBy = null,
    orderMode = 'asc',
  } = opts;
```

opencti-platform/opencti-graphql/src/database/engine.js
```diff
@@ -1743,7 +1742,9 @@ export const elFindByIds = async (context, user, ids, opts = {}) => {
   const queryIndices = computeQueryIndices(indices, types);
   const computedIndices = getIndicesToQuery(context, user, queryIndices);
   const hits = [];
-  const groupIds = R.splitEvery(MAX_TERMS_SPLIT, processIds);
+  // Leave room in split size compared to max pagination to minimize data loss risk in case of duplicated ids in database
```
xfournet, line 1774: how duplicated ids in DB can happen ?

opencti-platform/opencti-graphql/src/database/engine.js
xfournet: why default orderBy is changed to null ?

i also wonder if this ordering is important and need to be enforced:
* the method comment suggest it's not intended to be ordered ?
* we don't provide a global ordering. this was already the case due to the 64k limit, but even more now with the max pagination size

-> do we need to remove caller choosen order support ?

opencti-platform/opencti-graphql/src/database/engine.js
JeremyCloarec: it should never happen theoretically, but we've encountered some cases on multiple platforms. We already have a warning log in place when this occurs (`logApp.warn('Search query returned more elements than expected', { ids: workingIds });`)
The most common case was when an entity had a standard_id that was also contained in the x_opencti_stix_ids of an other entity, due to a bug in our deduplication buisness logic. But there were some other cases also

opencti-platform/opencti-graphql/src/database/engine.js
JeremyCloarec: The change to default orderBy to null was made to remove sorting from the elastic query by default, and only set sorting when explicitely asked by the caller.

But as you said, ordering was broken when asking above 64k ids before, and will be broken here when asking above 5k ids... I'll look at which callers exactly ask for ordering, and see if it can be done an other way or if it can be completely removed 👍

JeremyCloarec: I looked at all methods using elFindByIds, there is only one spot where we are currently sending an orderBy in options, that's in `taskList` in `taskManager.js`

opencti-platform/opencti-graphql/src/manager/taskManager.js
```diff
@@ -118,15 +117,12 @@ export const taskQuery = async (context, user, task, callback) => {
...
 export const taskList = async (context, user, task, callback) => {
-  const { task_position, task_ids, scope, task_order_mode } = task;
+  const { task_position, task_ids } = task;
   // task_position is no longer used, but we still handle it to properly process task that were processing before task migrated to worker
   const isUndefinedPosition = R.isNil(task_position) || R.isEmpty(task_position);
   const startIndex = isUndefinedPosition ? 0 : task_ids.findIndex((id) => task_position === id) + 1;
   const ids = task_ids.slice(startIndex);
   const options = {
-    orderMode: task_order_mode || 'desc',
-    // processing elements in descending order makes possible restoring from trash elements with dependencies
```
xfournet, line 128: how is it handled now since there is no more order ?

JeremyCloarec: unless I missed something, it was a leftover from a previous version of the code and it's no longer of any use. We're always getting the full list of ids anyway and there is no guarantee that the processing in the worker is done in the exact same order as it is retrieved here

opencti-platform/opencti-graphql/src/database/engine.js
```diff
@@ -1799,47 +1805,31 @@ export const elFindByIds = async (context, user, ids, opts = {}) => {
...
+      query.docvalue_fields = REL_DEFAULT_FETCH;
+    }
+    logApp.debug('[SEARCH] elInternalLoadById', { query });
+    const searchType = `${ids} (${types ? types.join(', ') : 'Any'})`;
+    const data = await elRawSearch(context, user, searchType, query).catch((err) => {
+      throw DatabaseError('Find direct ids fail', { cause: err, query, searchType });
+    });
+    const elements = data.hits.hits;
+    if (elements.length > workingIds.length) {
+      logApp.warn('Search query returned more elements than expected', { ids: workingIds });
```
xfournet: this log will happen when there are duplicated id ?
should we also produce another log test if ES_MAX_PAGINATION  is reached that might indicate some results are discarded ?

JeremyCloarec: Yes, this I kept the log & log condition the same as before: this log happens when there are more results returned than the amount of ids asked for.

I don't think the case where ES_MAX_PAGINATION is reached needs to be handled: it will be handled by the previous condition. Or do you mean keeping this log as a warning, and adding an error log in the case of ES_MAX_PAGINATION reached since it might imply some missed results and thus higher criticity?

xfournet: yes it would be another log
Regarding log level, I would put the current log into info and the new one in info or warning. Both don't have any actionnable counter measure, so it's more for diagnostic

xfournet: or we could use the total_hits to check if we should redo the query to retrieve all the elements

opencti-platform/opencti-graphql/src/database/engine.js
```diff
@@ -1743,7 +1748,9 @@ export const elFindByIds = async (context, user, ids, opts = {}) => {
   const queryIndices = computeQueryIndices(indices, types);
   const computedIndices = getIndicesToQuery(context, user, queryIndices);
   const hits = [];
-  const groupIds = R.splitEvery(MAX_TERMS_SPLIT, processIds);
+  // Leave room in split size compared to max pagination to minimize data loss risk in case of duplicated ids in database
+  const splitSize = Math.max(ES_MAX_PAGINATION - 100, ES_DEFAULT_PAGINATION);
```
xfournet: instead of ES_MAX_PAGINATION - 100 use ES_MAX_PAGINATION / 2 ?

opencti-platform/opencti-graphql/src/database/engine.js
xfournet: or fail like in elLoadById

JeremyCloarec: I'll fail like in elLoadById in case we reach ES_MAX_PAGINATION: now that we split with ES_MAX_PAGINATION / 2, reaching ES_MAX_PAGINATION means there is something wrong going on and it justifies throwing an error I think

opencti-platform/opencti-graphql/src/database/engine.js
```diff
@@ -1799,48 +1805,35 @@ export const elFindByIds = async (context, user, ids, opts = {}) => {
...
+    const searchType = `${ids} (${types ? types.join(', ') : 'Any'})`;
+    const data = await elRawSearch(context, user, searchType, query).catch((err) => {
+      throw DatabaseError('Find direct ids fail', { cause: err, query, searchType });
+    });
+    const elements = data.hits.hits;
+    if (elements.length > workingIds.length) {
+      logApp.info('Search query returned more elements than expected', { ids: workingIds });
+      if (elements.length >= ES_MAX_PAGINATION) {
+        throw DatabaseError('Ids loading returned more elements than paging allowed for, some elements could not be loaded', { ids: workingIds });
       }
```
xfournet: we could add  workingIds.length elements.length and ES_MAX_PAGINATION  in log and exceptions details.
i wouldn't add the full list of workingIds, maybe we could return only the list of ids that have some duplicates ?

opencti-platform/opencti-graphql/src/database/engine.js
```diff
@@ -1800,47 +1806,45 @@ export const elFindByIds = async (context, user, ids, opts = {}) => {
...
+        const associatedElement = elements.filter((e) => e[internalId.name] === workingId
+            || e[standardId.name] === workingId
+            || e[xOpenctiStixIds.name]?.includes(workingId)
+            || e[iAliasedIds.name]?.includes(workingId));
+        if (associatedElement.length > 1) {
+          duplicatedIds.push(workingId);
         }
-      } else {
-        hasNextPage = false;
       }
```
xfournet: it's O(n²)  since workingIds and elements have comparable size, and elements.filter is O(n)

JeremyCloarec: you're right, I coded the straightforward way to find the duplicates, since I figured it wasn't that much of a problem since n is always lower than ES_MAX_PAGINATION, but it's not too complicated to make it O(n). I'll push the change

opencti-platform/opencti-graphql/src/database/engine.js
```diff
@@ -1727,6 +1727,28 @@ const REL_COUNT_SCRIPT_FIELD = {
   }
 };
 
+const findElementsDuplicateIds = (elements) => {
+  const duplicatedIds = [];
+  const elementIds = new Set();
+  const checkCurrentIds = (id) => {
+    if (!id) return;
+    if (elementIds.has(id)) {
+      duplicatedIds.push(id);
```
xfournet: we could have duplicates in duplicatedIds ?

opencti-platform/opencti-graphql/src/database/engine.js
```diff
@@ -1727,6 +1727,28 @@ const REL_COUNT_SCRIPT_FIELD = {
...
+      elementIds.add(id);
+    }
+  };
+  for (let i = 0; i < elements.length; i += 1) {
+    const element = elements[i];
+    const { [internalId.name]: internal_id, [standardId.name]: standard_id, [xOpenctiStixIds.name]: otherStixIds, [iAliasedIds.name]: aliasIds } = element;
+    checkCurrentIds(internal_id);
+    checkCurrentIds(standard_id);
+    otherStixIds?.map((id) => checkCurrentIds(id));
+    aliasIds?.map((id) => checkCurrentIds(id));
```
xfournet, line 1747: should we first deduplicate internal_id standard_id, otherStixIds and aliasIds, so if we have duplicates in it, this will not generate a false duplicate ?

JeremyCloarec: I think this would still be a good information to get in the log, since having duplicates in a single element should never happen. So if it does, getting the id in the log is a good thing I think

opencti-platform/opencti-graphql/src/database/engine.js
```diff
@@ -1728,13 +1727,41 @@ const REL_COUNT_SCRIPT_FIELD = {
   }
 };
 
+const findElementsDuplicateIds = (elements) => {
+  const duplicatedIds = [];
+  const elementIds = new Set();
+  const checkCurrentIds = (id) => {
+    if (!id) return;
+    if (elementIds.has(id) && !duplicatedIds.includes(id)) {
```
xfournet: we could use a set to avoid includes on the array

xfournet (approved):

PR #12823, opened 2025-10-16
[backend] Fix child lock memory issue with callbacks (#12249)

### LINKED ISSUE #12249: Child lock memory issue with callbacks
## Description

Since child locking (enabled by default), when calling lockResources, abort and lock callbacks are set (added to a map in memory) but are not deleted if the lock is already taken.

## Environment

2. OpenCTI version: { e.g. OpenCTI 1.0.2 } latest
4. Other environment details:

## Reproducible Steps

Issue found by adding logs in childLockResources for lockProcess.callbacks and running multiple backends in cluster mode. lockProcess.callbacks map keeps increasing.

SouadHadjiat:
### Proposed changes

* clear callbacks when receiving error on lock

### Related issues

* #12249

### Checklist

### Further comments

To test locally:
- Add a log to see the lockProcess.callbacks map (you can log the size or the keys) at the begining of childLockResources function
- Start the backend as usual -> you should see at some point the log, everytime we try to take a lock (by managers), and the callbacks map is stable (it grows when managers start, then it stays the same)
- Start a second backend on cluster mode ("Backend start cluster (second platform)" configuration) -> you should see the map growing and growing, as the managers try to take the lock and fail (since the first platform has the locks)

### DIFF
--- a/opencti-platform/opencti-graphql/src/lock/master-lock.js
+++ b/opencti-platform/opencti-graphql/src/lock/master-lock.js
@@ -40,15 +40,19 @@ export const initLockFork = () => {
   }
 };
 
+const removeAllCallbacksForOperation = (operation) => {
+  lockProcess.callbacks.delete(`${operation}-lock`);
+  lockProcess.callbacks.delete(`${operation}-unlock`);
+  lockProcess.callbacks.delete(`${operation}-abort`);
+};
+
 // Unlock definition
 const childUnlockResources = async (operation) => {
   return new Promise((resolve, reject) => {
     // Set up the unlock callback
     lockProcess.callbacks.set(`${operation}-unlock`, (msg) => {
       // Cleanup the callback map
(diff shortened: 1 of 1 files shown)


opencti-platform/opencti-graphql/src/lock/master-lock.js
```diff
@@ -40,15 +40,19 @@ export const initLockFork = () => {
   }
 };
 
+const removeAllCallbacksForOperation = (operation) => {
+  lockProcess.callbacks.delete(`${operation}-lock`);
+  lockProcess.callbacks.delete(`${operation}-unlock`);
+  lockProcess.callbacks.delete(`${operation}-abort`);
```
OctaveLaventure, line 46: **1. Suggestion:**
We may want to have the `lock`, `unlock`, `abort` in constants to avoid typos

OctaveLaventure (approved):

JeremyCloarec (approved):

PR #12832, opened 2025-10-16
[CI] Feature branch deployment fix

efaure:
### Proposed changes

Feature branch build was broken due to node version. 
In order to avoid specific case in feature branch like this, this PR harmonize how build is made for feature branch and in Circle CI. 

A different docker file is provided for feature branch as the one used for circle ci required to build opencti front before launching the docker build.

The scope of this PR is to fix feature branch. 
Another PR will be done with Docker file improvement for both feature branch and Circle CI (the target is to merge them.

### DIFF
--- a/.github/workflows/test-feature-branch.yml
+++ b/.github/workflows/test-feature-branch.yml
@@ -19,7 +19,7 @@ on:
 
 jobs:
   worker:
-    name: worker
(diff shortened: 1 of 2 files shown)


OctaveLaventure (approved):
🚀

PR #12850, opened 2025-10-17
[backend] Add ability to send all in cron component 

marieflorescontact:
### Proposed changes

* Add `includeAll` parameter in cron component

### Related issues

*
*

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/database/middleware-loader.ts
+++ b/opencti-platform/opencti-graphql/src/database/middleware-loader.ts
@@ -327,7 +327,7 @@ export const buildAggregationRelationFilter = <T extends BasicStoreCommon>(relat
 };
 
 // entities
-export const buildEntityFilters = <T extends BasicStoreCommon>(entityTypes: string | Array<string> | undefined, args: EntityFilters<T> = {}) => {
+export const buildEntityFilters = <T extends BasicStoreCommon>(entityTypes: string | Array<string> | undefined | null, args: EntityFilters<T> = {}) => {
   const types = !entityTypes || isEmptyField(entityTypes) ? [ABSTRACT_STIX_CORE_OBJECT] : entityTypes;
   return buildRelationsFilter(types, args);
 };
@@ -355,7 +355,7 @@ export const fullEntitiesThoughAggregationConnection = async (context: AuthConte
   return buildPagination(0, null, nodeElements, nodeElements.length);
 };
 
-export const fullEntitiesList = async <T extends BasicStoreEntity>(context: AuthContext, user: AuthUser, entityTypes: Array<string>,
+export const fullEntitiesList = async <T extends BasicStoreEntity>(context: AuthContext, user: AuthUser, entityTypes: Array<string> | null,
   args: EntityOptions<T> = {}): Promise<Array<T>> => {
   const { indices } = args;
   const computedIndices = computeQueryIndices(indices, entityTypes);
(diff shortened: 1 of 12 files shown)


opencti-platform/opencti-graphql/src/modules/playbook/playbook-components.ts
```diff
@@ -217,6 +218,7 @@ const PLAYBOOK_INTERNAL_DATA_CRON_SCHEMA: JSONSchemaType<CronConfiguration> = {
     period: { type: 'string', default: 'hour' },
     triggerTime: { type: 'string' },
     onlyLast: { type: 'boolean', $ref: 'Only last modified entities after the last run', default: false },
+    includeAll: { type: 'boolean', $ref: 'Include all entities in a single bundle', default: false },
```
OctaveLaventure, line 221: **1. Question**
Does this need new translations ? It looks like other `$ref` (for example in `onlyLast`) have some

opencti-platform/opencti-graphql/src/manager/playbookManager.ts
```diff
@@ -451,29 +451,25 @@ const initPlaybookManager = () => {
...
-            const opts = { ...queryOptions, first: PLAYBOOK_CRON_MAX_SIZE };
-            const result = await elPaginate(context, RETENTION_MANAGER_USER, READ_STIX_INDICES, opts);
-            const elements = result.edges;
-            logApp.info(`[OPENCTI-MODULE] Running playbook ${instance.name} on ${elements.length} elements`);
-            for (let index = 0; index < elements.length; index += 1) {
-              const { node } = elements[index];
-              const data = await stixLoadById(context, RETENTION_MANAGER_USER, node.internal_id);
-              if (data) {
+            if (cronConfiguration.includeAll) {
+              const opts = { ...queryOptions, };
```
OctaveLaventure, line 455: **2. Nitpick**
The trailing comma `,` won't break anything but I'm surprised it wasn't linted

opencti-platform/opencti-graphql/src/database/middleware-loader.ts
```diff
@@ -327,7 +327,7 @@ export const buildAggregationRelationFilter = <T extends BasicStoreCommon>(relat
 };
 
 // entities
-export const buildEntityFilters = <T extends BasicStoreCommon>(entityTypes: string | Array<string> | undefined, args: EntityFilters<T> = {}) => {
+export const buildEntityFilters = <T extends BasicStoreCommon>(entityTypes: string | Array<string> | undefined | null, args: EntityFilters<T> = {}) => {
```
OctaveLaventure, line 330: **3. Question**
Is this change needed? Removing it doesn't seem to break the build, is it only to be more explicit? If so, should we extend this addition to some of the variables passed as this argument?

opencti-platform/opencti-graphql/src/modules/playbook/playbook-components.ts
JeremyCloarec: good catch, it's indeed missing translations, thanks 👍

opencti-platform/opencti-graphql/src/database/middleware-loader.ts
JeremyCloarec: At first we were using the `fullEntitiesList` method, where we did need to specify both in `fullEntitiesList` & `buildEntityFilters` that `entityTypes` can be null.

We switched to using the `stixLoadByFilters` function in our code afterwards, so it's no longer needed, but we kept the change since, as you said, it makes it more explicit & shows that a null value is valid. There are probably also other spots where we could modify it as you said, but we didn't take the time to do so in this PR

OctaveLaventure (approved):

Issue #12739, opened 2025-10-13
fix(knowledge-view): inconsistency between entities and relationships view for Campaign Knowledge 'related entities' for some observables


Archidoit:
## Description
Under the 'related entities' section of a Campaign Knowledge tab, some observables are not displayed under the entities view, but their relationship with the entity is displayed under the relationship view.

## Reproducible Steps
Go under a Campaign knowledge tab.

Add a relationship with a software under the 'observables' section.

Go under the 'related entities' section.

The relationship with the software is displayed under the relationship view, but not under the entities view.

<img width="1708" height="869" alt="Image" src="https://github.com/user-attachments/assets/8112085d-e653-4097-9d51-dafb8fed8b72" />

<img width="1727" height="893" alt="Image" src="https://github.com/user-attachments/assets/2437ce6d-786e-4b72-b5af-4e11c47ad680" />

alice-debra:
Hello @Archidoit,  
I wasn't able to reproduce the issue. Could you please send me a link?

Archidoit:
@alice-debra  here is the link : https://testing.octi.staging.filigran.io/dashboard/threats/campaigns/76379e2d-1a3e-444e-969f-3e8e7c48b29e/knowledge/related?searchTerm=&sortBy=created&orderAsc=false&view=relationships

Archidoit:
Do we want all the related-to relations in the 'related entities' section ? or only some entity types ? @alice-debra 
i.e. which view is correct:
- the entities view without the Software,
- or the Relationships view with the Software (and more generally all the related-to relations)

PR #12854, opened 2025-10-20
[frontend] predefined source/target types in Knowledge Related entities (#12739)

Archidoit:
### Proposed changes
- In Knowledge tab of an entity > Related entities section : relationships and entities view should display the same relations

### Related issues
#12739


CelineSebe:
The new relation with Software is not display in related entities (neither in the entities nor in the relationships), but the counter has changed
image

Issue #12489, opened 2025-09-20
Deleting a filter set in widget creation always removes the last set


Lhorus6:
## Description

When creating a widget, if multiple filter sets are defined (during step 3), deleting a set always removes the last one, even if another set is explicitly selected for deletion.

## Environment

OCTI 6.7.20

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Create a widget "vertical bar" and "Entities" perspective, and go to step 3 (filters).
2. Add multiple filter sets.

<img width="1360" height="967" alt="Image" src="https://github.com/user-attachments/assets/c3feb5ef-a230-4b0c-a18b-2c2a0aa06eea" />

3. Delete the first filter set.

<img width="1390" height="905" alt="Image" src="https://github.com/user-attachments/assets/db847b87-b9f4-497d-b640-ce2e8d01bbbd" />

4. Observe that the last filter set is deleted instead.

<img width="1362" height="685" alt="Image" src="https://github.com/user-attachments/assets/049ebc5b-8420-478a-968f-63df9c1af1b2" />

## Expected Output

The selected filter set should be deleted.

## Actual Output

The last filter set is deleted, regardless of which one is selected.

PR #12878, opened 2025-10-21
[frontend] Delete the specific group of filters in multi filter widget (#12489)

esrevi:
### Proposed changes

* Add selectionId in WidgetDataSelection to provide a stable unique identifier
* This make React rendering and updates more predictable and solves the issue of not deleting the selected group of filters in the widget such in VerticalBar Widget
* Remove makeStyles and refactor a little the component

### Related issues

* #12489 
*

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/widgets/WidgetCreationDataSelection.tsx
+++ b/opencti-platform/opencti-front/src/private/components/widgets/WidgetCreationDataSelection.tsx
@@ -5,7 +5,7 @@ import Tooltip from '@mui/material/Tooltip';
 import { InformationOutline } from 'mdi-material-ui';
 import WidgetFilters from '@components/widgets/WidgetFilters';
 import Button from '@mui/material/Button';
-import React from 'react';
+import React, { useState } from 'react';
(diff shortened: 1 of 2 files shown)


opencti-platform/opencti-front/src/private/components/widgets/WidgetCreationDataSelection.tsx
```diff
@@ -75,6 +75,7 @@ const WidgetCreationDataSelection = () => {
     setDataSelection([
       ...dataSelection,
       {
+        selectionId: uuidv4(),
```
Archidoit: Adding an arbitrary id to a data selection only for frontend display purpose is not a good idea I think, since data selection is sent to widget queries

esrevi: ok i understand your point, i have updated to handle a list of item ids that won't change the data selected sent to the backend

Archidoit (approved):
Working well, nice workaround to handle keys !! :)

Issue #6669, opened 2024-04-11
Ability to remove labels using automation/playbook


nino-filigran:
## Use case

Currently, we do not support remove data with playbooks that are written in DB.

## Proposed Solution

Either: 
- add synchro when ingestion (like synchro of platforms), full synch. 
- in a playbook, describe some atomic operation that needs to be performed. as a result, we do not send a bundle but a list of action, like the background task.
- Other solution to think about: a kind of mix between bundle ingestion & list of operations, where you have the initial point, the operation to be performed, and the expected output. This might be the preffered option. 

As a result, we would need to tie the background tasks with this behavior.

CyberKaizen:
This is badly needed for when you need to normalize labeling coming from 3rd party intel sources.

romain-filigran:
Depend of https://github.com/OpenCTI-Platform/opencti/issues/10274

jasonkasih:
We require this feature as well, to automate the 'bulk-edit and remove labels' step. Will save precious time. thank you.

fruitcakej:
Hi five Sarah!
:-)

PR #12880, opened 2025-10-21
[backend][chunk 1] Remove labels from playbook - manipulate knowledge component patch (#6669)

SarahBocognano:
### Proposed changes

* Change update method by replacing the upsert with a patch operation in manipulate knowledge component

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/6669

### DIFF
--- a/opencti-platform/opencti-front/lang/front/en.json
+++ b/opencti-platform/opencti-front/lang/front/en.json
@@ -3785,6 +3785,7 @@
   "The password has been updated": "The password has been updated",
   "The public dashboard is disabled": "The public dashboard is disabled",
   "The relations attached to selected entities will be copied to the merged entity.": "The relations attached to selected entities will be copied to the merged entity.",
+  "The remove operation will now permanently delete the selected attribute from the entity, including attributes that already exist in the platform (not just those added during playbook execution). This action cannot be undone through the playbook.": "The remove operation will now permanently delete the selected attribute from the entity, including attributes that already exist in the platform (not just those added during playbook execution). This action cannot be undone through the playbook.",
   "The reset code you entered is invalid or has expired. You can request a new code after a delay of 30 seconds.": "The reset code you entered is invalid or has expired. You can request a new code after a delay of 30 seconds.",
   "The retention policy will be applied on global files (files contained in": "The retention policy will be applied on global files (files contained in",
   "The retention policy will be applied on global workbenches (workbenches contained in": "The retention policy will be applied on global workbenches (workbenches contained in",
--- a/opencti-platform/opencti-front/src/private/components/data/playbooks/PlaybookAddComponents.jsx
+++ b/opencti-platform/opencti-front/src/private/components/data/playbooks/PlaybookAddComponents.jsx
@@ -791,7 +791,7 @@ const PlaybookAddComponentsContent = ({
                               )}
                               {(actionsInputs[i]?.op === 'remove') && (
                                 <Alert severity="warning" style={{ marginBottom: 20 }}>
-                                  {t_i18n('The remove operation will now permanently delete the selected attribute from the entity, including attributes that already exist in the platform (not just those added during playbook execution). This action cannot be undone through the playbook.')}
+                                  {t_i18n('Remove operation will only apply on field values added in the context of this playbook such as enrichment or other knowledge manipulations but not if values are already written in the platform.')}
                                 </Alert>
                               )}
                               <div key={i} className={classes.step}>
--- a/opencti-platform/opencti-front/src/private/components/data/playbooks/PlaybookAddComponents.jsx
+++ b/opencti-platform/opencti-front/src/private/components/data/playbooks/PlaybookAddComponents.jsx
@@ -791,7 +791,7 @@ const PlaybookAddComponentsContent = ({
                               )}
                               {(actionsInputs[i]?.op === 'remove') && (
                                 <Alert severity="warning" style={{ marginBottom: 20 }}>
-                                  {t_i18n('Remove operation will only apply on field values added in the context of this playbook such as enrichment or other knowledge manipulations but not if values are already written in the platform.')}
+                                  {t_i18n('The remove operation will now permanently delete the selected attribute from the entity, including attributes that already exist in the platform (not just those added during playbook execution). This action cannot be undone through the playbook.')}
                                 </Alert>
                               )}
                               <div key={i} className={classes.step}>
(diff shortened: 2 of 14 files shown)


opencti-platform/opencti-graphql/src/modules/playbook/playbook-components.ts
```diff
@@ -978,27 +976,63 @@ const PLAYBOOK_UPDATE_KNOWLEDGE_COMPONENT: PlaybookComponent<UpdateConfiguration
...
             if (multiple) {
               const currentValues = jsonpatch.getValueByPointer(bundle, path) ?? [];
-              const actionValues = action.value.map((o) => {
+              const actionPatchValues = action.value.map((o) => {
                 // If value is an id, must be converted to standard_id has we work on stix bundle
                 if (cacheIds.has(o.patch_value)) return (cacheIds.get(o.patch_value) as BasicStoreCommon).standard_id;
                 // Else, just return the value
                 return convertValue(attributeType, o.patch_value);
               });
+              const actionValues = action.value.map((o) => {
```
lndrtrbn, line 986: For my understanding, could you explain the difference between `actionpatchValues` and `actionValues` please?

opencti-platform/opencti-graphql/src/modules/playbook/playbook-components.ts
```diff
@@ -978,27 +976,63 @@ const PLAYBOOK_UPDATE_KNOWLEDGE_COMPONENT: PlaybookComponent<UpdateConfiguration
...
+                // If value is an id, must be converted to standard_id has we work on stix bundle
+                if (cacheIds.has(o.value)) return (cacheIds.get(o.value) as BasicStoreCommon).standard_id;
+                // Else, just return the value
+                return convertValue(attributeType, o.value);
+              });
               if (action.op === EditOperation.Add) {
-                return { op: EditOperation.Replace, path, value: R.uniq([...currentValues, ...actionValues]) };
+                return {
+                  op: action.op,
```
lndrtrbn, line 994: I do not understand what is action operation and action patch operation. Could you explain me please?

SouadHadjiat:
We can't merge this PR, either we merge everything (patch operation for all components, not just manipulate knowledge) or we need a feature flag

opencti-platform/opencti-graphql/src/modules/playbook/playbook-components.ts
SarahBocognano: The difference between both of those is just the format of the value. The problem we encountered here is that we receive a actionValue as a "label" for exemple "John Doe" for an Author, and for the patch to happen we need a patch value as an "id" for exemple "identity--58030383838393..." for John Doe.

opencti-platform/opencti-graphql/src/modules/playbook/playbook-utils.ts
```diff
@@ -66,3 +67,16 @@ export const convertMembersToUsers = async (
...
+export const applyOperationFieldPatch = (element: StixObject, patchObject: {
+  key: string;
+  value: any[];
+  operation: 'add' | 'replace' | 'remove'
+}[]) => {
+  if (isFeatureEnabled('FIELD_PATCH_IN_PLAYBOOKS')) {
+    // eslint-disable-next-line no-param-reassign
+    element.extensions[STIX_EXT_OCTI].opencti_operation = 'patch';
+    // eslint-disable-next-line no-param-reassign
+    element.extensions[STIX_EXT_OCTI].opencti_field_patch = patchObject;
```
SouadHadjiat: maybe add the patch object to the possibly existing opencti_field_patch instead of replacing it ? like we have seen together ?

opencti-platform/opencti-graphql/src/modules/playbook/playbook-utils.ts
```diff
@@ -66,3 +67,19 @@ export const convertMembersToUsers = async (
...
+  value: any[];
+  operation: 'add' | 'replace' | 'remove'
+}[]) => {
+  if (isFeatureEnabled('FIELD_PATCH_IN_PLAYBOOKS')) {
+    // eslint-disable-next-line no-param-reassign
+    element.extensions[STIX_EXT_OCTI].opencti_operation = 'patch';
+    // eslint-disable-next-line no-param-reassign
+    element.extensions[STIX_EXT_OCTI].opencti_field_patch = [
+      ...(element.extensions[STIX_EXT_OCTI].opencti_field_patch ?? []),
+      patchObject
```
SouadHadjiat: should be ...patchObject

opencti-platform/opencti-front/src/private/components/data/playbooks/PlaybookAddComponents.jsx
```diff
@@ -791,7 +791,7 @@ const PlaybookAddComponentsContent = ({
                               )}
                               {(actionsInputs[i]?.op === 'remove') && (
                                 <Alert severity="warning" style={{ marginBottom: 20 }}>
-                                  {t_i18n('Remove operation will only apply on field values added in the context of this playbook such as enrichment or other knowledge manipulations but not if values are already written in the platform.')}
+                                  {t_i18n('The remove operation will now permanently delete the selected attribute from the entity, including attributes that already exist in the platform (not just those added during playbook execution). This action cannot be undone through the playbook.')}
```
SouadHadjiat: I don't think we should change this part, since we are still in feature flag

SouadHadjiat: We could do it at the end of the feature, it's not necessary for first chunk.

opencti-platform/opencti-front/lang/front/en.json
```diff
@@ -3785,6 +3785,7 @@
   "The password has been updated": "The password has been updated",
   "The public dashboard is disabled": "The public dashboard is disabled",
   "The relations attached to selected entities will be copied to the merged entity.": "The relations attached to selected entities will be copied to the merged entity.",
+  "The remove operation will now permanently delete the selected attribute from the entity, including attributes that already exist in the platform (not just those added during playbook execution). This action cannot be undone through the playbook.": "The remove operation will now permanently delete the selected attribute from the entity, including attributes that already exist in the platform (not just those added during playbook execution). This action cannot be undone through the playbook.",
```
SouadHadjiat, line 3788: so this translation can be removed

SarahBocognano: Well I was thinking of leaving it there as it will be added again after. But if you want I can remove it yes

Issue #12684, opened 2025-10-10
Can't link entities to relationships


Archidoit:
## Description
It's not possible to link an entity to a relationship from the relationship overview: page crashes.

## Reproducible Steps
- Go under an entity knowledge tab, and select a relationship
- In the relationship overview, click on the '+' button next to 'Linked entities'

<img width="1723" height="651" alt="Image" src="https://github.com/user-attachments/assets/2f7c7404-a8e9-4d83-955a-809075f43e0d" />

- An error occurs

<img width="1031" height="214" alt="Image" src="https://github.com/user-attachments/assets/bf09efbc-6c2d-48ee-b036-1eabdccc3353" />

<img width="532" height="176" alt="Image" src="https://github.com/user-attachments/assets/d6ff6126-9c61-4e06-92ec-2f50567c77ad" />

alice-debra:
I can't access the relationship. Is it linked?
https://github.com/user-attachments/assets/4e1006e2-9f48-4cb2-89e7-fc11ab702f66

Archidoit:
> I can't access the relationship. Is it linked? https://github.com/user-attachments/assets/4e1006e2-9f48-4cb2-89e7-fc11ab702f66

No @alice-debra , it's not linked. There is already an other issue for this problem: https://github.com/OpenCTI-Platform/opencti/issues/12616

PR #12892, opened 2025-10-22
[frontend] add provider for safe usage of computeLink

### LINKED ISSUE #12887: Additional issues with Public Dashboard and v6.8.7
## Description

There is an issue with Public Dashboard in the List (distribution) widget. In the Custom Dashboard view it works, but when viewing it as a Public Dashboard the widget throws the error [Please provide a support package to your administrator or OpenCTI maintainers].

Web browser Developer Tools says this:

front-ULRIAZ2G.js:1  Error: Invalid user context !
    at useAuth (front-ULRIAZ2G.js:1:2238562)
    at useSchema (front-ULRIAZ2G.js:1:11503920)
    at useComputeLink (front-ULRIAZ2G.js:1:11504550)
    at WidgetDistributionList (front-ULRIAZ2G.js:1:12321207)
    at renderWithHooks (front-ULRIAZ2G.js:1:72582)
    at updateFunctionComponent (front-ULRIAZ2G.js:1:104836)
    at beginWork (front-ULRIAZ2G.js:1:116136)
    at performUnitOfWork (front-ULRIAZ2G.js:1:153243)
    at workLoopSync (front-ULRIAZ2G.js:1:152172)
    at renderRootSync (front-ULRIAZ2G.js:1:151972)
defaultOnCaughtError @ front-ULRIAZ2G.js:1

The purpose of Public Dashboard would be that it doesn't require any user or authentication...

## Environment

1. OS (where OpenCTI server runs): Ubuntu 24.04.3
2. OpenCTI version: 6.8.7

## Reproducible Steps

1. Create a List (distribution) widget on a Custom Dashboard
2. Make it Public
3. Go to the public dahsboard
4. View the error

OctaveLaventure:
### Proposed changes

* add provider for safe usage of `computeLink` in components called on both private and public paths and that use links generation

### Related issues

* https://github.com/orgs/OpenCTI-Platform/projects/8/views/50?pane=issue&itemId=134936493&issue=OpenCTI-Platform%7Copencti%7C12887

### Checklist

### Context

-  There was previously a blank page when creating a clickable link to a relationship with another relationship as their `from` (issue: https://github.com/OpenCTI-Platform/opencti/issues/12616 )
-  This was fixed in https://github.com/OpenCTI-Platform/opencti/issues/12684, but the solution requires the `isRelationship` function from `useSchema` > `useAuth`, which breaks public dashboards using widgets dependent on `computeLink`


opencti-platform/opencti-front/src/utils/hooks/useIsPublicRoute.ts
```diff
@@ -0,0 +1,21 @@
...
+ * ```tsx
+ * const isPublic = useIsPublicRoute();
+ * if (isPublic) {
+ *   // Handle public route logic
+ * }
+ * ```
+ */
+const useIsPublicRoute = (): boolean => {
+  const location = useLocation();
+  return location.pathname.startsWith('/public');
```
xfournet: could be safer by testing on '/public/'

OctaveLaventure: Done

opencti-platform/opencti-front/src/utils/hooks/WithPublicSafeComputeLink.tsx
xfournet, line 1: i fear it need to be handled case by case for all potentially public component.
could we find a solution that address the problem without need to wrap the component ?

OctaveLaventure:
Updated the solution to avoid using render props by using a react context and a more global provider (thanks @esrevi for the great idea!)

opencti-platform/opencti-front/src/utils/hooks/useIsPublicRoute.ts
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * @example
+ * ```tsx
+ * const isPublic = useIsPublicRoute();
+ * if (isPublic) {
+ *   // Handle public route logic
+ * }
+ * ```
+ */
+const useIsPublicRoute = (): boolean => {
```
esrevi: do we need still need this hook ? if not used, I think we could remove it

OctaveLaventure: Indeed, I deleted it, thanks!

esrevi (approved):
lgtm

opencti-platform/opencti-front/src/utils/hooks/useComputeLink.tsx
```diff
@@ -41,4 +43,38 @@ const useComputeLink = () => {
...
+  const computeLink = useComputeLinkFn();
+  return (
+    <ComputeLinkContext.Provider value={computeLink as ComputeLinkFn}>
+      {children}
+    </ComputeLinkContext.Provider>
+  );
+};
+
+export const ComputeLinkProvider = ({ children }: { children: ReactNode }) => {
+  const location = useLocation();
```
xfournet, line 68: useLocation triggers a rerender on every navigation, which could cause unnecessary rerendering.

Since we only instantiate ComputeLinkProvider in two places, we could simply pass a prop to indicate whether it's a public or private one.

xfournet:
The context-based approach looks much cleaner

opencti-platform/opencti-front/src/utils/hooks/useComputeLink.tsx
xfournet: (sorry for the after merge comment, it was left in pending comments...)

OctaveLaventure: Handled in https://github.com/OpenCTI-Platform/opencti/pull/12914

xfournet: thanks!

PR #12893, opened 2025-10-22
[frontend/backend] Bulk Search: Unknown entities API (#7279)

Archidoit:
### Proposed changes
In Bulk Search context
- API for unknown entities instead of post filtering logic in frontend
- possibility to order unknown entities by value

### Related issues
#7279

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/SearchBulk.tsx
+++ b/opencti-platform/opencti-front/src/private/components/SearchBulk.tsx
@@ -131,7 +131,7 @@ interface SearchBulkProps {
 const SearchBulk = ({ inputValues, dataColumns }: SearchBulkProps) => {
   const buildSearchBulkFilters = (values: string[], filters: FilterGroup) => {
     return values.length > 0
-      ? addFilter(filters, allEntitiesKeyList as unknown as string, values) // TODO INVALID TYPE
+      ? addFilter(filters, allEntitiesKeyList as unknown as string, values)
       : filters;
   };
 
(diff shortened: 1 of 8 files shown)


maelv-filigran:
##  Great Work on Your Backend Optimization! 

Solid move pushing this logic backend-side. You're solving the right problem here.

### What You Did Really Well

- **Smart architectural decision**: Moving heavy filtering to the backend is spot on
- **Clean implementation**: Your code is readable and follows the project patterns
- **Good test coverage**: Your tests clearly demonstrate the expected behavior
- **UI improvements**: Making columns sortable is a nice touch


### Two things on performance

First, `globalSearchPaginated` with no limits? That's gonna bite you. Picture this: production database, 2 million entities. Your function loads everything into memory. Boom.

Quick fix - slap a `first: 10000` parameter on there. Not perfect, but prevents the worst case.

Second thing - you're sitting on a goldmine with our Elasticsearch layer. Check out `elPaginate` and `elList` in `database/engine.js`. Instead of fetching everything then filtering, you could query ES directly with a terms filter. We already do this for hash searches, peek at `database/middleware.js` line with `hashes.*`.

### About that hash matching

```javascript
Object.values(stixObject.hashes).filter((h) => !!h).some((h) => h === value)
```

What if `hashes` is undefined? Or contains nested objects? I'd make this more defensive.

Also noticed: representative matching is case-insensitive, hash matching isn't. Intentional?

### TypeScript angle

Here's my take, this is new code, critical search functionality. Why not TypeScript from the start?

###  Potential Path

You're clearly capable of tackling more complex optimizations!
I propose a path:

__Level 1__: Current PR ✅\
__Level 2__: Add that pagination limit\
__Level 3__: Rewrite with `elPaginate` and terms filters\
__Level 4__: TypeScript migration\
__Level 5__: Direct ES aggregations


Merge this: it works and fixes the issue. 
But create a follow-up ticket for the ES optimization. That's where the 10-100x perf gain lives.

Tell me if this makes sense or if I'm overthinking the perf angle.
Your call on the follow-up work ;)

maelv-filigran (approved):

PR #12894, opened 2025-10-22
[backend] fix access for users outside platform org (#12926)

### LINKED ISSUE #12818: Creator filter shows no values for users outside the main organization when organization segregation is enabled
## Description

When the organization segregation feature is enabled, users who do not belong to the platform’s main organization cannot use the Creator filter properly.

The dropdown list of possible creators appears empty, meaning no selectable values are available.

## Environment

OCTI 6.8.5

## Reproducible Steps

1. Add "Org A" as the platform’s main organization
2. Log in with a user that does not belong to "Org A"
3. Try to use the Creator filter on any entity list (e.g. Reports).

Observe that the filter values are empty.

Image

## Expected Output

The user should be able to see and select creators from their accessible scope

## Actual Output

The filter is empty, no creator values are displayed.

### LINKED ISSUE #12926: Users can see others of others orgs when assigning/participating/authorize members
## Description

We have done: https://github.com/OpenCTI-Platform/opencti/issues/11964.

But it seems that a user can still see other users not pertaining to their own org, when trying to list assignees. 
## Environment

Local

## Reproducible Steps

1. Have a platform setup with OrgA=platform Org
2. Create two other orgs (Org B and C) 
3. Create one user per org (OrgBuser1 and OrgCuser1)
4. With a user OrgBuser1 create a report.
5. Click on assignee
6. Look for OrgCuser1 in the list

## Expected Output

You should not finf OrgCuser1 in the list

## Actual Output

You can find OrgCuser1 in the list. However if you select the user, it will appear as "Restricted". 

## Additional information

When filtering on Assignee/Participant, the problem is the same.

JeremyCloarec:
### Proposed changes

* add org filter for users outside platform org on members query
* add post filtering on user orgs for users outside platform org on creators/participants/assignees query

### Related issues

* #12818
* https://github.com/OpenCTI-Platform/opencti/issues/12926

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/database/engine.js
+++ b/opencti-platform/opencti-graphql/src/database/engine.js
@@ -3733,7 +3733,7 @@ export const elAggregationNestedTermsWithFilter = async (context, user, indexNam
 };
 
 export const elAggregationsList = async (context, user, indexName, aggregations, opts = {}) => {
-  const { types = [], resolveToRepresentative = true } = opts;
+  const { types = [], resolveToRepresentative = true, postResolveFilter } = opts;
   const queryAggs = {};
   aggregations.forEach((agg) => {
     queryAggs[agg.name] = {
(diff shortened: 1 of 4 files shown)


lndrtrbn (commented):
Tested locally and works fine

SouadHadjiat (approved):

PR #12915, opened 2025-10-24
[frontend/backend] Bulk Search: Bulk actions on known entities (#7279)

Archidoit:
### Proposed changes
- Add bulk actions on known entities in Bulk Search
- Fix Unknown entities api: no duplicates
- Special filter key for bulk search keywords in backend
- Improve post filtering of unknown entities (example: if an alias is found, don't put the value in unknown values)

### Related issues
#7279

### DIFF
--- a/opencti-platform/opencti-graphql/src/domain/stixCoreObject.js
+++ b/opencti-platform/opencti-graphql/src/domain/stixCoreObject.js
@@ -196,7 +196,7 @@ export const findUnknownStixCoreObjects = async (context, user, args) => {
         return Array.isArray(stixObjectValue)
           ? stixObjectValue.includes(value)
           : stixObjectValue === value;
-      })
+      });
     }
     return representativeMatch;
   };
--- a/opencti-platform/opencti-graphql/src/domain/stixCoreObject.js
+++ b/opencti-platform/opencti-graphql/src/domain/stixCoreObject.js
@@ -190,14 +190,13 @@ export const findUnknownStixCoreObjects = async (context, user, args) => {
         if (hashMatch) return hashMatch;
       }
       // try to find in attributes of bulk search filter
-      for (let i = 0; i < BULK_SEARCH_KEYWORDS_FILTER_KEYS.length; i += 1) {
-        const key = BULK_SEARCH_KEYWORDS_FILTER_KEYS[i];
+      return BULK_SEARCH_KEYWORDS_FILTER_KEYS.some((key) => {
         const stixObjectValue = stixObject[key];
-        if (stixObjectValue) {
-          const keyMatch = Array.isArray(stixObjectValue) ? stixObjectValue.includes(value) : stixObjectValue === value;
-          if (keyMatch) return true;
-        }
-      }
+        if (!stixObjectValue) return false;
+        return Array.isArray(stixObjectValue)
+          ? stixObjectValue.includes(value)
+          : stixObjectValue === value;
+      })
     }
     return representativeMatch;
   };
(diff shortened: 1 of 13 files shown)


opencti-platform/opencti-graphql/src/domain/stixCoreObject.js
```diff
@@ -165,7 +165,19 @@ export const globalSearchPaginated = async (context, user, args) => {
 };
 
 export const findUnknownStixCoreObjects = async (context, user, args) => {
-  const { values, filters, orderBy, orderMode } = args;
+  const { values: inputValues, orderBy, orderMode } = args;
+  if (inputValues.length === 0) {
+    return [];
+  }
+  const values = R.uniq(inputValues);
```
maelv-filigran: Isn't using Ramda considered bad practice ?
If so, would it be better to use something like `const values = [...new Set(inputValues)];` instead?

Archidoit: thanks ! i'm gonna fix this

opencti-platform/opencti-graphql/src/domain/stixCoreObject.js
```diff
@@ -177,14 +189,25 @@ export const findUnknownStixCoreObjects = async (context, user, args) => {
         const hashMatch = Object.values(stixObject.hashes).filter((h) => !!h).some((h) => h === value);
         if (hashMatch) return hashMatch;
       }
+      // try to find in attributes of bulk search filter
+      for (let i = 0; i < BULK_SEARCH_KEYWORDS_FILTER_KEYS.length; i += 1) {
```
esrevi: if i'm correct, we're looking for at least one matching key, even if the index is not yet equal to the last index item.

Thus, I think we could simplify this function using "some" instead of looping, it returns true as one match is found and stops looping.
```
return BULK_SEARCH_KEYWORDS_FILTER_KEYS.some((key) => {
  const stixObjectValue = stixObject[key];
  
  if (!stixObjectValue) return false;
  
  return Array.isArray(stixObjectValue) 
    ? stixObjectValue.includes(value) 
    : stixObjectValue === value;
})
```

opencti-platform/opencti-graphql/src/domain/stixCoreObject.js
```diff
@@ -177,14 +189,25 @@ export const findUnknownStixCoreObjects = async (context, user, args) => {
...
+        const key = BULK_SEARCH_KEYWORDS_FILTER_KEYS[i];
+        const stixObjectValue = stixObject[key];
+        if (stixObjectValue) {
+          const keyMatch = Array.isArray(stixObjectValue) ? stixObjectValue.includes(value) : stixObjectValue === value;
```
SouadHadjiat: this should be case insensitive, is it ? do we have unit tests for some of these cases ?

opencti-platform/opencti-graphql/src/domain/stixCoreObject.js
Archidoit: thanks! indeed that's better this way :) I'm gonna fix this

opencti-platform/opencti-graphql/src/domain/stixCoreObject.js
Archidoit: I added a test on case insensitivity :)

Archidoit: thanks :)

esrevi (approved):
lgtm ! :)

Issue #10, opened 2019-04-17
Implement the knowledge graph of a report


manonviallet:
## Problem to Solve

Report knowledge creation must be available in order to create relations between entities and modeling the knowledge of the report. 

## Current Workaround

None.

## Proposed Solution

Create the knowledge graph and all the features such as adding new entities, create and manage relations, move the node and edges.

## Additional Information

None.

PR #12924, opened 2025-10-24
Improve docker build

efaure:
### Proposed changes

Improve Docker build to simplify it and reduce OpenCTI image (potential change from 1.7 GB to 1 GB). 

 - Remove the step `graphql-deps-builder` which seems redundant with `graphql-builder`  + add immutability and yarn cache clean to `graphql-builder`  https://github.com/OpenCTI-Platform/opencti/pull/12924/commits/47fac29f1350321817f6e2e254785a62647ad09c   
 - Small refacto : reduce number of layer and clean apk cache (no major impact) https://github.com/OpenCTI-Platform/opencti/pull/12924/commits/d15e94bf339a91ee83cd447961a42b6ba3d2d603
 - Improve apk management in final image: 
   - Separate the layer for apk dependencies that will stay in the image (no size impact as it will stay). This will improve build cache management and build speed
   - Add dependencies that can be removed after build in `virtual` and remove them after python build
   - Remove some dependencies that seems not useful for the app image: g++, make, cargo

The final goal is to merge this Dockerfile with Circle CI docker file. 

The other difference with Circle Ci docker file is to build opencti-front inside the docker file (instead of before, as currently done in Circle CI)

### Related issues

*
*

### Checklist

### DIFF
--- a/opencti-platform/Dockerfile_featurebranch
+++ b/opencti-platform/Dockerfile_featurebranch
@@ -28,6 +28,7 @@ RUN set -ex; \
 
 FROM base AS app
 
+COPY opencti-graphql/src/python/requirements.txt /opt/opencti/src/python/requirements.txt
 RUN set -ex; \
     apk add --no-cache git tini gcc g++ make musl-dev cargo python3 python3-dev postfix postfix-pcre; \
     rm -f /usr/lib/python3.11/EXTERNALLY-MANAGED; \
@@ -35,11 +36,12 @@ RUN set -ex; \
     python3 -m ensurepip; \
     rm -rv /usr/lib/python*/ensurepip; \
     pip3 install --no-cache-dir --upgrade pip setuptools wheel; \
-    ln -sf python3 /usr/bin/python;
+    ln -sf python3 /usr/bin/python; \
+    pip3 install --no-cache-dir --requirement /opt/opencti/src/python/requirements.txt; \
+    apk del git gcc musl-dev; \
+    rm -rf /var/cache/apk/*
+
 WORKDIR /opt/opencti
-COPY opencti-graphql/src/python/requirements.txt ./src/python/requirements.txt
-RUN pip3 install --no-cache-dir --requirement ./src/python/requirements.txt
-RUN apk del git gcc musl-dev
 
 COPY --from=graphql-builder /opt/opencti-build/opencti-graphql/node_modules ./node_modules
 COPY --from=graphql-builder /opt/opencti-build/opencti-graphql/build ./build
--- a/opencti-platform/Dockerfile_featurebranch
+++ b/opencti-platform/Dockerfile_featurebranch
@@ -1,33 +1,28 @@
 FROM node:22-alpine AS base
 
-FROM base AS graphql-builder
+FROM base AS builder
 
 ARG TAG_VERSION
 ENV TAG_VERSION=${TAG_VERSION}
 
 # Get graphql dependencies for layer cache optimization
 WORKDIR /opt/opencti-build/opencti-graphql
-COPY opencti-graphql/package.json opencti-graphql/yarn.lock ../.yarnrc.yml ./
+RUN apk add --no-cache g++ make musl-dev python3 python3-dev postfix-pcre
 COPY opencti-graphql/patch ./patch
-RUN set -ex; \
-    apk add --no-cache git tini gcc g++ make musl-dev cargo python3 python3-dev postfix postfix-pcre \
-    && rm -f /usr/lib/python3.11/EXTERNALLY-MANAGED \
-    && rm -f /usr/lib/python3.12/EXTERNALLY-MANAGED \
-    && npm install -g node-gyp \
-    && npm install -g corepack \
-    && yarn install  --immutable
+RUN npm install -g node-gyp
+RUN npm install -g corepack
+COPY opencti-graphql/package.json opencti-graphql/yarn.lock ../.yarnrc.yml ./
+RUN yarn install --immutable
 
 # Build graphql
 COPY opencti-graphql /opt/opencti-build/opencti-graphql
-WORKDIR /opt/opencti-build/opencti-graphql
 RUN yarn build:prod
 
 # Build front
 WORKDIR /opt/opencti-build/opencti-front
 COPY opencti-front ../.yarnrc.yml /opt/opencti-build/opencti-front
-RUN set -ex; \
-  yarn install; \
-  yarn build
+RUN yarn install
+RUN yarn build
 
 FROM base AS app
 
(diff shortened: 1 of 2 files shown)


efaure:
@xfournet @aHenryJard 
I'm interested to have your PoV on this change. 

I made it on the docker file dedicated to feature branch, but my goal is to merge it with the Circle CI docker file. 

I spitted commit and added some explanation in description. 
Main point of attention / question : 
 - Do we need cargo in the final image ?? If we remove cargo, g++ and make we can reduce the image size from 1.7 GB to 1 GB
 - I removed the first step of multi stage used only to build node js, because I don't see what is the added value. Please let me know if I miss something and this step is useful ? 
 - The remaining difference with Circle CI docker file is that currently in Circle CI, the front end is not build inside the dockerfile but in the CircleCI step `build_frontend`. I think it's safe and clean to move the build of the front in the docker build  https://github.com/OpenCTI-Platform/opencti/blob/f1a130e04c16d7366a64b3de08866a98fc13ad09/.circleci/config.yml#L6 

If all those changes seems okay to you I will report them to other Dockerfile (and test the result)

xfournet:
Hi @efaure i have also WIP on Docker images. I will contact you

opencti-platform/Dockerfile_featurebranch
```diff
@@ -31,7 +31,7 @@ FROM base AS app
 COPY opencti-graphql/src/python/requirements.txt /opt/opencti/src/python/requirements.txt
 RUN apk add --no-cache tini python3 python3-dev postfix postfix-pcre
 RUN set -ex; \
-    apk add --no-cache --virtual build-dep git gcc musl-dev g++ make cargo; \
+    apk add --no-cache --virtual build-dep git gcc musl-dev; \
```
xfournet: git could be removed too ?

efaure: Indeed. Fixed in https://github.com/OpenCTI-Platform/opencti/pull/12924/commits/afd58d370eb7a52d751aea55d67a6570c2ecaffa

efaure:
@xfournet  I tested this last version on a feature branch and it's seems to work without issue. 
The new size of the image is 800 M (before it was 1,7 GB). 
I would like to merge this PR and wait a little bit from feedback of feature branch users before updating the build done in Circle CI for production images (and merging this feature branch docker file with the circle ci one)

opencti-platform/Dockerfile_featurebranch
```diff
@@ -27,33 +15,37 @@ RUN set -ex; \
...
 RUN set -ex; \
-  yarn install && yarn cache clean --all; \
-  yarn build && yarn cache clean --all;
+  yarn install; \
+  yarn build
 
 FROM base AS app
 
+COPY opencti-graphql/src/python/requirements.txt /opt/opencti/src/python/requirements.txt
+RUN apk add --no-cache tini python3 python3-dev postfix postfix-pcre
```
xfournet: is postfix required ?

opencti-platform/Dockerfile_featurebranch
```diff
@@ -1,22 +1,10 @@
...
-RUN set -ex; \
-    apk add --no-cache git tini gcc g++ make musl-dev cargo python3 python3-dev postfix postfix-pcre \
-    && rm -f /usr/lib/python3.11/EXTERNALLY-MANAGED \
-    && rm -f /usr/lib/python3.12/EXTERNALLY-MANAGED \
-    && npm install -g node-gyp \
-    && npm install -g corepack \
-    && yarn install --immutable && yarn cache clean --all
-
-
 FROM base AS graphql-builder
```
xfournet: could be renamed to builder since it's building both backend and frontend

opencti-platform/Dockerfile_featurebranch
xfournet: are git, postfix, tini and cargo required ?
gcc and g++ to be tested also ?

xfournet: except for yarnrc.yml theses copies are useless since it's already done later on line 21

i propose to change the order:
1/ RUN apk add and node-gyp & corepack installs (so it's stable in the cache since not dependent of files in the build context)
2/ copy . in /opt/opencti-build
3/ build back (eg yarn install & yarn build;prod)
4/ build front (eg yarn install & yarn build)

opencti-platform/Dockerfile_featurebranch
```diff
@@ -27,33 +15,37 @@ RUN set -ex; \
...
+    apk add --no-cache --virtual build-dep gcc musl-dev; \
     rm -f /usr/lib/python3.11/EXTERNALLY-MANAGED; \
     rm -f /usr/lib/python3.12/EXTERNALLY-MANAGED; \
     python3 -m ensurepip; \
     rm -rv /usr/lib/python*/ensurepip; \
     pip3 install --no-cache-dir --upgrade pip setuptools wheel; \
-    ln -sf python3 /usr/bin/python;
+    ln -sf python3 /usr/bin/python; \
+    pip3 install --no-cache-dir --requirement /opt/opencti/src/python/requirements.txt; \
+    apk del build-dep
```
xfournet: i would also split in two :
first everything that is not dependent of the file context (to get stable cache)
then copy and rest of the build

opencti-platform/Dockerfile_featurebranch
xfournet: i think only python files are required here

xfournet: scripts are already present in the /build

opencti-platform/Dockerfile_featurebranch
efaure: Maybe not. I deployed a feature branch without issue and didn't face any error.

opencti-platform/Dockerfile_featurebranch
efaure: After some test: 
The build fail if we remove `g++` 
> #10 17.36 ➤ YN0009: │ node-calls-python@patch:node-calls-python@npm%3A1.11.1#./patch/node-calls-python.patch::version=1.11.1&hash=13bcc5&locator=opencti-graphql%40workspace%3A. couldn't be built successfully (exit code 1, logs can be found here: /tmp/xfs-fbbd7827/build.log)

The build is successful when removing git, postfix, tini, cargo and gcc

efaure: 1.  I agree, it's better to run apk, node-gyp and corepack on separate layer for cache improvement.

2. Copying dependencies files before yarn build allow a cache optim. Dependencies will probably not change as often as code, so all the install part can be cached.  I think it won't have a negative impact on build time and for the build step we don't care about layer size. Wdyt to keep it ?

opencti-platform/Dockerfile_featurebranch
efaure: Indeed :+1: 
I thought it would require to keep the build dep and increase the layer size, but the pip install for requirement don't seems to need build-dep.

efaure:
@xfournet I made some update following your feedback. 
The new version of the image is deployed as a feature branch on https://feat-oob-improve-doc.octi.staging.filigran.io and seems to work fine after a quick check.

New changes in the resulting image: 
 - Remove install of useless packages during builder docker build  : git, postfix, tini, cargo and gcc
 - Remove install of useless packages during app docker build: postfix
 - Copy of only src/python instead of the whole src folder
 - Do not copy opencti-graphql/scripts
 - Some change in layer organization 

I'll ask someone with a better functional knowledge of OpenCTI to help me test this version.

xfournet (approved):

efaure:
@xfournet 
I tested this last version on a feature branch. It seems to work fine.
As discussed I made some test: 
 - Removing make from builder step leads to build failure, so I kept it.
 - I removed musl-dev from node builder step and python build, it seems to have no impact.
 - I confirmed that TAG_VERSION is well used for the connector manifest generation
 - I tested to send an email on the feature branch to ensure there is no side effect following removal of postfix

Cosmetic: I also removed the usage of "virtual build-dep" for the python build step. As we are now installing only gcc, it's not anymore useful to group dependencies under this virtual build-dep package.

PR #12929, opened 2025-10-25
[backend/frontend] Improving regardingOf filtering to support not_eq

richard-julien:
Improving regardingOf filtering to support not_eq


Archidoit:
'or local mode between relationship type is not working : 

image

image

opencti-platform/opencti-graphql/src/database/engine.js
```diff
@@ -3080,16 +3092,16 @@ const completeSpecialFilterKeys = async (context, user, inputFilters) => {
...
         } else {
           const keys = isEmptyField(types)
             ? buildRefRelationKey('*', '*')
             : types.flatMap((t) => [buildRefRelationKey(t, ID_INTERNAL), buildRefRelationKey(t, ID_INFERRED)]);
-          regardingFilters.push({ key: keys, operator, values: ids });
+          regardingFilters.push({ key: keys, operator: filter.operator, values: ids });
         }
         finalFilterGroups.push({
-          mode: filter.mode ?? FilterMode.Or,
+          mode: FilterMode.And, // type and id are grouped to one filter
```
Archidoit: with this change, we don't take into account the filter.mode anymore... The 'or' is always turned to an 'and'

Archidoit (changes requested):

Archidoit:
With 'not_eq', the 'or' local mode is useless (not eq (X or Y).
By default, we rather use a 'AND' between values for not_eq filters. Here we can't switch the mode :

image

Archidoit:
The 'or' global mode is switch to a 'and' : 

image


Whereas there are countries in my platform : 

image

richard-julien:
Close for https://github.com/OpenCTI-Platform/opencti/pull/12963

PR #12956, opened 2025-10-27
[worker] implement thread pool selector for push handlers in worker 

JeremyCloarec:
### Proposed changes

* to keep total workers count low, but keep realtime connector priority handling, a new thread pool selector is added. The selector enables realtime work to run on default pool if no default work is ongoing, or enables default work to run on realtime pool if no realtime work is ongoing
*

### Related issues

*
*

### Checklist

### DIFF
--- a/opencti-worker/src/message_queue_consumer.py
+++ b/opencti-worker/src/message_queue_consumer.py
@@ -1,9 +1,7 @@
 import functools
-from concurrent.futures import ThreadPoolExecutor
 from dataclasses import dataclass, field
 from threading import Thread
 from typing import Any, Callable, Literal
-from thread_pool_selector import ThreadPoolSelector
 
 import pika
 
@@ -17,7 +15,6 @@ class MessageQueueConsumer:  # pylint: disable=too-many-instance-attributes
     submit_fn: Callable[[Callable[[int, str], None], int, str], Any]
     handle_message: Callable[[str], Literal["ack", "nack", "requeue"]]
     should_stop: bool = field(default=False, init=False)
-    is_realtime: bool = False
 
     def __post_init__(self) -> None:
         self.pika_connection = pika.BlockingConnection(self.pika_parameters)
@@ -81,9 +78,7 @@ class MessageQueueConsumer:  # pylint: disable=too-many-instance-attributes
                     },
                 )
                 task_future = self.submit_fn(
-                    self.consume_message,
-                    method.delivery_tag,
-                    body
+                    self.consume_message, method.delivery_tag, body
                 )
                 while task_future.running():  # Loop while the thread is processing
                     self.pika_connection.sleep(0.05)
--- a/opencti-worker/src/message_queue_consumer.py
+++ b/opencti-worker/src/message_queue_consumer.py
@@ -1,6 +1,7 @@
 import functools
 from dataclasses import dataclass, field
 from threading import Thread
+from concurrent.futures import Future
 from typing import Any, Callable, Literal
 
 import pika
@@ -12,7 +13,7 @@ class MessageQueueConsumer:  # pylint: disable=too-many-instance-attributes
     consumer_type: Literal["listen", "push"]
     queue_name: str
     pika_parameters: pika.ConnectionParameters
-    submit_fn: Callable[[Callable[[int, str], None]], Any]
+    submit_fn: Callable[[Callable[[], None]], Future[None]]
     handle_message: Callable[[str], Literal["ack", "nack", "requeue"]]
     should_stop: bool = field(default=False, init=False)
 
--- a/opencti-worker/src/message_queue_consumer.py
+++ b/opencti-worker/src/message_queue_consumer.py
@@ -17,6 +17,7 @@ class MessageQueueConsumer:  # pylint: disable=too-many-instance-attributes
     execution_pool: ThreadPoolExecutor | ThreadPoolSelector
     handle_message: Callable[[str], Literal["ack", "nack", "requeue"]]
     should_stop: bool = field(default=False, init=False)
+    is_realtime: bool = False
 
     def __post_init__(self) -> None:
         self.pika_connection = pika.BlockingConnection(self.pika_parameters)
@@ -79,11 +80,19 @@ class MessageQueueConsumer:  # pylint: disable=too-many-instance-attributes
                         "tag": method.delivery_tag,
                     },
                 )
-                task_future = self.execution_pool.submit(
-                    self.consume_message,
-                    method.delivery_tag,
-                    body,
-                )
+                if type(self.execution_pool) is ThreadPoolSelector:
+                    task_future = self.execution_pool.submit(
+                        self.is_realtime,
+                        self.consume_message,
+                        method.delivery_tag,
+                        body,
+                    )
+                else:
+                    task_future = self.execution_pool.submit(
+                        self.consume_message,
+                        method.delivery_tag,
+                        body,
+                    )
                 while task_future.running():  # Loop while the thread is processing
                     self.pika_connection.sleep(0.05)
                 self.logger.info("Message processed, thread terminated")
--- a/opencti-worker/src/message_queue_consumer.py
+++ b/opencti-worker/src/message_queue_consumer.py
@@ -12,7 +12,7 @@ class MessageQueueConsumer:  # pylint: disable=too-many-instance-attributes
     consumer_type: Literal["listen", "push"]
     queue_name: str
     pika_parameters: pika.ConnectionParameters
-    submit_fn: Callable[[Callable[[int, str], None], int, str], Any]
+    submit_fn: Callable[[Callable[[int, str], None]], Any]
     handle_message: Callable[[str], Literal["ack", "nack", "requeue"]]
     should_stop: bool = field(default=False, init=False)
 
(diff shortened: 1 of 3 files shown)


opencti-worker/src/message_queue_consumer.py
```diff
@@ -13,9 +14,10 @@ class MessageQueueConsumer:  # pylint: disable=too-many-instance-attributes
     consumer_type: Literal["listen", "push"]
     queue_name: str
     pika_parameters: pika.ConnectionParameters
-    execution_pool: ThreadPoolExecutor
+    execution_pool: ThreadPoolExecutor | ThreadPoolSelector
```
xfournet: execution_pool and is_realtime could be abstracted, replaced by a single 'submitTask' arg that can be directly used by messagequeueconsumer

opencti-worker/src/worker.py
```diff
@@ -264,16 +271,17 @@ def start(self) -> None:
                             bundles_processing_time_gauge,
                             self.objects_max_refs,
                         )
-                        execution_pool = push_execution_pool
+                        is_realtime = False
                         if is_priority_connector(connector["connector_priority_group"]):
-                            execution_pool = realtime_push_execution_pool
+                            is_realtime = True
```
xfournet: can be simplified
`is_realtime = is_priority_connector(connector["connector_priority_group"])`

opencti-worker/src/thread_pool_selector.py
xfournet, line 1: If there are 100 default messages already submitted and waiting to be processed by the thread pool, a new real-time message will have to wait until those 100 messages are processed before it can be handled.

JeremyCloarec: Hmm, how so? If there are 100 default messages, there are at most 3 of them being handled by the relatime threadpool. The other 97 ones should be handled by the default threadpool. Then when a new realtime message is handled in the submit, the message will be sent to the realtime threadpool and not the default threadpool, and will have at most to wait for one of the default message handled by the realtime threadpool to finish

The selector's goal it to utilize the other threapool (the realtime threadpool for default messages and default threadpool for realtime messages) when it is currently idle, but it makes sure not to overload it and leave it ready to be used for it's main usage.

opencti-worker/src/message_queue_consumer.py
```diff
@@ -13,9 +14,10 @@ class MessageQueueConsumer:  # pylint: disable=too-many-instance-attributes
     consumer_type: Literal["listen", "push"]
     queue_name: str
     pika_parameters: pika.ConnectionParameters
-    execution_pool: ThreadPoolExecutor
+    submit_fn: Callable[[Callable[[int, str], None], int, str], Any]
     handle_message: Callable[[str], Literal["ack", "nack", "requeue"]]
     should_stop: bool = field(default=False, init=False)
+    is_realtime: bool = False
```
xfournet: it can be removed now

opencti-worker/src/thread_pool_selector.py
```diff
@@ -0,0 +1,56 @@
...
+        )
+        self.realtime_active_threads.add(task_future)
+        task_future.add_done_callback(self.realtime_active_threads.remove)
+        return task_future
+
+    def submit(self, is_realtime, consume_message_fn, delivery_tag, body):
+        if is_realtime:
+            if (
+                len(self.default_active_threads) <= self.default_pool_size
+                and len(self.realtime_active_threads) > self.realtime_pool_size
```
xfournet: can this condition happen ? active threads count should normally never exceed the pool_size

opencti-worker/src/thread_pool_selector.py
```diff
@@ -0,0 +1,56 @@
...
+                    consume_message_fn, delivery_tag, body
+                )
+            else:
+                return self.submit_to_realtime_pool(
+                    consume_message_fn, delivery_tag, body
+                )
+        else:
+            if (
+                len(self.realtime_active_threads) <= self.realtime_pool_size
+                and len(self.default_active_threads) > self.default_pool_size
```
xfournet: cannot happen ?

opencti-worker/src/thread_pool_selector.py
xfournet: i would extract the two variables (updated in regard of the comment related to pool size comparison) to make the code more readable:

```python
    def submit(self, is_realtime, consume_message_fn, delivery_tag, body):
        is_default_pool_full = len(self.default_active_threads) >= self.default_pool_size
        is_realtime_pool_full = len(self.realtime_active_threads) >= self.realtime_pool_size
        if is_realtime:
            if is_realtime_pool_full and not is_default_pool_full:
                return self.submit_to_default_pool(
                    consume_message_fn, delivery_tag, body
                )
            else:
                return self.submit_to_realtime_pool(
                    consume_message_fn, delivery_tag, body
                )
        else:
            if is_default_pool_full and not is_realtime_pool_full:
                return self.submit_to_realtime_pool(
                    consume_message_fn, delivery_tag, body
                )
            else:
                return self.submit_to_default_pool(
                    consume_message_fn, delivery_tag, body
                )
```

opencti-worker/src/thread_pool_selector.py
```diff
@@ -0,0 +1,56 @@
...
+            consume_message_fn, delivery_tag, body
+        )
+        self.realtime_active_threads.add(task_future)
+        task_future.add_done_callback(self.realtime_active_threads.remove)
+        return task_future
+
+    def submit(self, is_realtime, consume_message_fn, delivery_tag, body):
+        if is_realtime:
+            if (
+                len(self.default_active_threads) <= self.default_pool_size
```
xfournet: we should not try to submit a job to the default pool if it is full ? (pool size equality case)

opencti-worker/src/thread_pool_selector.py
```diff
@@ -0,0 +1,56 @@
...
+                return self.submit_to_default_pool(
+                    consume_message_fn, delivery_tag, body
+                )
+            else:
+                return self.submit_to_realtime_pool(
+                    consume_message_fn, delivery_tag, body
+                )
+        else:
+            if (
+                len(self.realtime_active_threads) <= self.realtime_pool_size
```
xfournet: we should not try to submit a job to the realtime pool if it is full (case size are equals)

opencti-worker/src/thread_pool_selector.py
xfournet: Indeed I hadn’t understood correctly.

There might be some out-of-order edge cases. For example, if only real-time bundles are coming in, they could overload the processing queue and start piling up in the real-time executor. Occasionally, however, a new real-time bundle might arrive just as the default thread pool becomes available to process it immediately. I’m wondering whether this could confuse end users who experience delays for most of their actions, yet notice that some requests are executed instantly while earlier ones are still pending.

opencti-worker/src/thread_pool_selector.py
```diff
@@ -0,0 +1,56 @@
...
+@dataclass(unsafe_hash=True)
+class ThreadPoolSelector:  # pylint: disable=too-many-instance-attributes
+    default_pool_size: int
+    default_execution_pool: ThreadPoolExecutor
+    realtime_pool_size: int
+    realtime_execution_pool: ThreadPoolExecutor
+
+    def __post_init__(self) -> None:
+        self.default_active_threads = set()
+        self.realtime_active_threads = set()
```
xfournet: set are no thread safe, they must be locked to mutate them safely.

also we could use int rather than set ? (still need to lock them to update them)

opencti-worker/src/thread_pool_selector.py
```diff
@@ -0,0 +1,56 @@
...
+
+    def submit_to_realtime_pool(self, consume_message_fn, delivery_tag, body):
+        task_future = self.realtime_execution_pool.submit(
+            consume_message_fn, delivery_tag, body
+        )
+        self.realtime_active_threads.add(task_future)
+        task_future.add_done_callback(self.realtime_active_threads.remove)
+        return task_future
+
+    def submit(self, is_realtime, consume_message_fn, delivery_tag, body):
```
xfournet: we could make it abstract from the consume_message signature:
`def submit(self, is_realtime, fn: Callable[..., Any], *args: Any, **kwargs: Any):`

opencti-worker/src/thread_pool_selector.py
```diff
@@ -0,0 +1,65 @@
...
+
+    def __post_init__(self) -> None:
+        self.default_active_threads_count = 0
+        self.realtime_active_threads_count = 0
+        self.count_lock = threading.Lock()
+
+    def decrement_default_count(self, _future):
+        self.count_lock.acquire()
+        self.default_active_threads_count -= 1
+        self.count_lock.release()
```
xfournet: locks management can be simplified:

```
with self.lock:
    self.default_active_threads_count -= 1
```

JeremyCloarec: thanks, I didn't know that

xfournet:
@JeremyCloarec there was some typing warning on IDEA, i pushed this commit 2157ec44bade138f598019a98784a5efa95d4667 to fix it, let me know if it's ok for you

xfournet (approved):

Issue #12969, opened 2025-10-28
Public dashboard are broken


nino-filigran:
## Description

Try to open a public dashboard: it fails. 

## Environment

Testing 

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. open a public dashboard

## Expected Output

It opens without issues
## Actual Output

Error

<img width="1888" height="973" alt="Image" src="https://github.com/user-attachments/assets/f2229647-a26f-44b3-a97b-1c1f6cec063e" />

PR #12989, opened 2025-10-29
Revert "[frontend] Can't create support package while in draft mode (#12246)"

lndrtrbn:
This reverts commit 18b8b6ec2672c8b46bb48c28e6af8f2ab7f38bca.

### Proposed changes

* We cannot call `useDraftContext` in component Error. This hook need an authenticated user, which is not compatible with public dashboards.

### Related issues

* #12969

### Checklist


labo-flg:
I have reopened #12246 to solve the issue with another approach then.

labo-flg (approved):

Issue #12988, opened 2025-10-29
[XTM HUB] Email sent to admins on demo everyday (connectivity lost)


jbanety:
## Description

On demo instances that are reset daily, the XTM Hub connectivity check fails regularly because the registration state becomes invalid after each reset. This triggers automated emails to administrators every 24 hours when connectivity is lost, causing email spam.

While the connectivity check itself is useful to display the connection status in the UI, the automated emails are problematic for demo environments.

## Environment

1. OS (where OpenCTI server runs): Any
2. OpenCTI version: Current master branch
3. OpenCTI client: Backend (hub registration manager)
4. Other environment details: Affects instances with daily resets

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Register an OpenCTI instance to the XTM Hub
2. Reset the instance database (simulating a demo reset)
3. Wait for the hub registration manager to run (executes every hour)
4. The connectivity check fails and updates the status to "lost_connectivity"
5. After 24 hours without successful connectivity, an email is sent to all administrators
6. This repeats every 24 hours as long as connectivity remains lost

## Expected Output

On demo instances with environment variable configured:
- The Hub connectivity check should still run normally
- The connection status should still be updated and visible in the UI
- **No emails should be sent** to administrators about connectivity loss

## Actual Output

Currently:
- The Hub connectivity check runs and updates the status correctly
- Emails are sent to administrators every 24 hours when connectivity is lost
- There is no way to disable these emails via configuration

## Proposed Solution

Add an environment variable to control email notifications for Hub connectivity issues:

- New environment variable: `OPENCTI__XTM__XTMHUB_CONNECTIVITY_EMAIL_ENABLED` (default: `true`)
- When set to `false`, the connectivity check still runs and updates the UI status, but no emails are sent
- This allows demo instances to disable emails while keeping the functionality intact for production instances

**Files to modify:**
1. `config/default.json` - Add `xtmhub_connectivity_email_enabled: true` in the `xtm` section
2. `src/domain/xtm-hub.ts` - Check this configuration before sending emails in the `checkXTMHubConnectivity()` function

## Additional information

The current implementation already has a `xtm_hub_should_send_connectivity_email` setting stored in the database, but this gets reset along with the demo instances. An environment variable provides persistent configuration that survives database resets.

**Benefits:**
- Minimal change (only 2 files)
- Backward compatible (default behavior unchanged)
- Hub connectivity check continues to work normally
- Status remains visible in the UI
[...]

PR #12990, opened 2025-10-29
[backend] xtm-hub: add configuration to disable connectivity email notifications (#12988)

jbanety:
### Proposed changes

* Add environment variable `xtmhub_connectivity_email_enabled` to control Hub connectivity lost email notifications
* Modify `checkXTMHubConnectivity()` to check configuration before sending emails
* Add unit test to verify emails are not sent when configuration is disabled
* Keep connectivity check and UI status display active regardless of email configuration

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/12988
* https://github.com/FiligranHQ/xtm-hub/issues/1175

### Checklist

### Further comments

**Context:**
On demo instances that are reset daily, the Hub connectivity check fails regularly because the registration state becomes invalid after each reset. This triggers automated emails to administrators every 24 hours, causing email spam.

**Solution:**
This PR adds a new environment variable `OPENCTI__XTM__XTMHUB_CONNECTIVITY_EMAIL_ENABLED` (default: `true`) to control email notifications while keeping all other functionality intact:

✅ **What continues to work:**
- Hub connectivity check runs every hour (via `hub_registration_manager`)
- Connection status is updated in the database
- Status remains visible in the UI
- Connectivity restoration triggers re-enable email notifications

❌ **What can be disabled:**
- Email notifications to administrators when connectivity is lost

**Usage for demo instances:**
Set environment variable: `OPENCTI__XTM__XTMHUB_CONNECTIVITY_EMAIL_ENABLED=false`

**Files changed:**
- `config/default.json` - Added configuration with default value `true`
- `src/domain/xtm-hub.ts` - Added configuration check before sending emails
- `tests/01-unit/domain/xtm-hub-test.ts` - Added test case for disabled emails scenario

**Backward compatibility:**
The default value is `true`, preserving the current behavior for all existing instances. Only instances that explicitly set the variable to `false` will have emails disabled.

**Testing:**
Added unit test `should not send connectivity email when email sending is disabled via configuration` that verifies:
- When configuration is set to `false`
- And all other conditions for sending email are met (24h passed, connectivity lost, settings flag true)
- Then no email is sent and no settings update occurs

### DIFF
--- a/opencti-platform/opencti-graphql/src/domain/xtm-hub.ts
+++ b/opencti-platform/opencti-graphql/src/domain/xtm-hub.ts
@@ -5,7 +5,7 @@ import { ENTITY_TYPE_SETTINGS } from '../schema/internalObject';
 import { xtmHubClient } from '../modules/xtm/hub/xtm-hub-client';
 import { type AutoRegisterInput, XtmHubRegistrationStatus } from '../generated/graphql';
 import { updateAttribute } from '../database/middleware';
-import { BUS_TOPICS, PLATFORM_VERSION } from '../config/conf';
+import { booleanConf, BUS_TOPICS, PLATFORM_VERSION } from '../config/conf';
 import { HUB_REGISTRATION_MANAGER_USER } from '../utils/access';
 import { getSettings, settingsEditField } from './settings';
 import { notify } from '../database/redis';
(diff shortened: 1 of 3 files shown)


opencti-platform/opencti-graphql/src/domain/xtm-hub.ts
```diff
@@ -5,47 +5,77 @@ import { ENTITY_TYPE_SETTINGS } from '../schema/internalObject';
...
 import { utcDate } from '../utils/format';
 import { sendAdministratorsLostConnectivityEmail } from '../modules/xtm/hub/xtm-hub-email';
 
-export const checkXTMHubConnectivity = async (context: AuthContext, user: AuthUser): Promise<{
-  status: XtmHubRegistrationStatus
+export const checkXTMHubConnectivity = async (
+  context: AuthContext,
+  user: AuthUser,
+): Promise<{
+  status: XtmHubRegistrationStatus;
```
SouadHadjiat: why all these formatting changes ? do you have a different formatter ? It makes it difficult to read the PR and see the changes

jbanety: sorry I don't know what happened. It's fixed

lndrtrbn (approved):

Issue #12981, opened 2025-10-28
Knowledge graph on report with more that 500 elements sometimes stuck on first elements


aHenryJard:
## Description

Not all the time but sometimes the graph knowledge is stuck on a number < 500.

From current investigation:
- Only the first query GraphContainerKnowledgeObjectsQuery run, the follow up GraphContainerKnowledgeObjectsRefetchQuery is not trigger
- instead of 500 elements, there is a bit less (like 493)
- on graphQL playground console there is a wrong globalCount
```
"objects": {
        "pageInfo": {
          "endCursor": 
          "hasNextPage": false,
          "globalCount": 493
        },
```

instead of 
```
"objects": {
        "pageInfo": {
          "endCursor": 
          "hasNextPage": true,
          "globalCount": 1205
        },
```

When doing the same query but with "count": 1500 => the graph is correctly display and globalCount is correct.

One other issue example:

<img width="1170" height="840" alt="Image" src="https://github.com/user-attachments/assets/fbc183e2-21d6-4fe6-bbf7-3c3c562d1756" />

## Environment

1. OS (where OpenCTI server runs): { e.g. Mac OS 10, Windows 10, Ubuntu 16.4, etc. }
2. OpenCTI version: 6.8.8
3. OpenCTI client: { e.g. frontend or python }
4. Other environment details:

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Given a report named "My report"
2. Include "My report" inside another report "bug report"
3. after that add more than 500 elements in this report
4. Show graph of "bug report"
=> there is always one element missing in the count (see screenshot above)

JeremyCloarec:
Seems to be related to the change to the hasNextPage computation in buildPaginationFromEdges here https://github.com/OpenCTI-Platform/opencti/pull/12325/files

JeremyCloarec:
Might be fixed with https://github.com/OpenCTI-Platform/opencti/pull/12991

JeremyCloarec:
https://github.com/OpenCTI-Platform/opencti/pull/12991 has been merged, it might fix this issue. We'll need to test again when the fix is available to see if it is resolved

Lhorus6:
The buggy reports on the instance have been tested and the bug is no longer present. We can close this issue :)

PR #12991, opened 2025-10-29
[backend] fix elPaginate hasNextPage compute when elements are post filtered (#11885)

JeremyCloarec:
### Proposed changes

* reverted change on hasNextPage compute from https://github.com/OpenCTI-Platform/opencti/pull/12325
*

### Related issues

* #11885
*

### Checklist

### Further comments

The hasNextPage value was incorrect as soon as an element was post filtered: to check if there is a nextPage, we verified that the postFilteredElements count was equal to the query size. But if an element is post filtered, this condition can never be met

### DIFF
--- a/opencti-platform/opencti-graphql/src/database/engine.js
+++ b/opencti-platform/opencti-graphql/src/database/engine.js
@@ -3865,10 +3865,10 @@ const buildRegardingOfFilter = async (context, user, elements, filters) => {
   return undefined;
 };
 
-const buildSearchResult = (elements, first, searchAfter, globalCount, connectionFormat) => {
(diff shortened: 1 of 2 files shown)


lndrtrbn:
Could this fix also fix known issue on graph where there are more entities but hasNextPage is false so the front does not load the rest?

JeremyCloarec:
> Could this fix also fix known issue on graph where there are more entities but hasNextPage is false so the front does not load the rest?

Like this issue you mean https://github.com/OpenCTI-Platform/opencti/issues/12981? Yes I think it could, but I'm not 100% sure. There are probably still a lot of bug caused by the post filtering with the regardingOf filter, so there might still be another issue behind this one

xfournet (approved):

Issue #12994, opened 2025-10-29
Unable to access "opinion" details


romain-filigran:
## Description

On a report, for example, it's possible to add a review. Once the review has been submitted, it's possible to click on the radar icon to display all reviews related to the entity and to click on one of them to view the details.
However, clicking on one of them redirects the user to a page that displays an error/exception.

## Environment

OpenCTI version: 6.8.8

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Express your opinion by clicking on the "Like" icon.
2. Once you've done that, click on the radar icon to view all opinions.
3. Click on the opinion you just submitted.

## Actual Output

<img width="2560" height="580" alt="Image" src="https://github.com/user-attachments/assets/807ad67b-c697-497d-b8fe-4dacd177cae8" />

PR #12995, opened 2025-10-29
[frontend] prevent crashing when DeleteComponent undefined in ContainerHeader (#12994)

JeremyCloarec:
### Proposed changes

* check if DeleteComponent is defined before using it in ContainerHeader
*

### Related issues

* #12994
*

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/common/containers/ContainerHeader.jsx
+++ b/opencti-platform/opencti-front/src/private/components/common/containers/ContainerHeader.jsx
@@ -788,7 +788,9 @@ const ContainerHeader = (props) => {
                     </Box>
                   )}
                 </PopoverMenu>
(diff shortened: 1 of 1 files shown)


Kedae:
Wouldn't it be better to add a deleteComponent for Opinion and add a clear "throw error" in the ContainerHeader if there is no DeleteComponent ?

JeremyCloarec:
> Wouldn't it be better to add a deleteComponent for Opinion and add a clear "throw error" in the ContainerHeader if there is no DeleteComponent ?

It would be better to add it in the Opinion component, but I went for a quick fix in this PR. There are a lot more things that we need to change on the Opinion component, we should open a followup issue to tackle everything (change edition button, fix container styling etc...)

lndrtrbn (approved):

PR #12997, opened 2025-10-29
[frontend] Can't create support package while in draft mode (#12246)

SarahBocognano:
### Proposed changes

* Change the method to explain user that it's not possible to generate a support package while in draft. It was checked before in the Error component with the useDraftContext, but the useDraftContext using the useAuth it created a conflict for public dashboards.
I've decided to disable the button to generate a support package instead with an alert to explain why.

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/12246

### DIFF
--- a/opencti-platform/opencti-front/lang/front/fr.json
+++ b/opencti-platform/opencti-front/lang/front/fr.json
@@ -4379,6 +4379,7 @@
   "You can configure your platform to consume OpenCTI Streams. A list of public and commercial native feeds is available in the": "Vous pouvez configurer votre plateforme pour consommer les flux OpenCTI. Une liste de flux natifs publics et commerciaux est disponible dans la section",
   "You can enable/disable default values for marking in the customization of each entity type.": "Vous pouvez activer/désactiver les valeurs par défaut pour le marquage dans la personnalisation de chaque type d'entité.",
   "You can now set a new password for your account.": "Vous pouvez maintenant définir un nouveau mot de passe pour votre compte.",
+  "You cannot generate a support package while in draft mode. Make sure to be out of draft to generate one.": "Vous ne pouvez pas générer de dossier de soutien lorsque vous êtes en mode brouillon. Assurez-vous de ne pas être en mode brouillon pour en générer un.",
   "You cannot have more than 500 e-mail addresses": "Vous ne pouvez pas avoir plus de 500 adresses électroniques",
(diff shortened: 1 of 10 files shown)


opencti-platform/opencti-front/lang/front/fr.json
```diff
@@ -4371,6 +4371,7 @@
   "You can configure your platform to consume OpenCTI Streams. A list of public and commercial native feeds is available in the": "Vous pouvez configurer votre plateforme pour consommer les flux OpenCTI. Une liste de flux natifs publics et commerciaux est disponible dans la section",
   "You can enable/disable default values for marking in the customization of each entity type.": "Vous pouvez activer/désactiver les valeurs par défaut pour le marquage dans la personnalisation de chaque type d'entité.",
   "You can now set a new password for your account.": "Vous pouvez maintenant définir un nouveau mot de passe pour votre compte.",
+  "You cannot generate a support package while in draft mode. Make sure to be out of draft to generate one.": "Vous ne pouvez pas générer de dossier de soutien lorsque vous êtes en mode brouillon. Assurez-vous de ne pas être en mode brouillon pour en générer un.",
```
lndrtrbn, line 4382:
```suggestion
  "You cannot generate a support package while in draft mode. Make sure to be out of draft to generate one.": "Vous ne pouvez pas générer de dossier de support lorsque vous êtes en mode brouillon. Assurez-vous de ne pas être en mode brouillon pour en générer un.",
```

opencti-platform/opencti-front/src/private/components/settings/support/SupportPackages.tsx
```diff
@@ -123,25 +126,34 @@ const SupportPackages = () => {
...
+          >
+          {t_i18n('You cannot generate a support package while in draft mode. Make sure to be out of draft to generate one.')}
+        </Alert> : (
+          <Alert
+            severity="warning"
+            variant="outlined"
+            style={{ position: 'relative', marginTop: 20, marginBottom: 20 }}
+          >
+            {t_i18n('We are doing our best to remove any sensitive information from support packages but we encourage you to check the content before sharing a support package depending on your security policy.')}
+          </Alert>
```
lndrtrbn: As the only thing changing is the text, you can avoid duplicating the Alert component and just wrap the text to render in the condition

JeremyCloarec (approved):

PR #13010, opened 2025-10-30
[backend] fix manual & auto enrichment (#13008)

### LINKED ISSUE #13008: Manual enrichment fails with error 'No bundle resolver for type X"
## Description

Manual enrichment is broken
## Environment

Testing

## Reproducible Steps

1. try to manually enrich an observable

## Expected Output

it works
## Actual Output

it fails

JeremyCloarec:
### Proposed changes

* fix an issue in askElementEnrichmentForConnectors with the introduction of stixBundleByIdStringify
* => now only compute stix_objects if required by the connector
* => now return null instead of throwing an error in stixBundleByIdStringify if no bundleResolver is defined for entity type
* fix an issue in filterConnectorsForElementEnrichment where autoTrigger could be computed to true even if connectors are not configured with auto_update = true
* => changed autoTrigger to match previous behavior

### Related issues

* #13008
*

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/database/middleware.js
+++ b/opencti-platform/opencti-graphql/src/database/middleware.js
@@ -544,10 +544,10 @@ export const stixLoadByIds = async (context, user, ids, opts = {}) => {
 };
 export const stixBundleByIdStringify = async (context, user, type, id) => {
   const resolver = modules.get(type)?.bundleResolver;
-  if (resolver) {
-    return await resolver(context, user, id);
+  if (!resolver) {
+    return null;
   }
-  return null;
+  return await resolver(context, user, id);
 };
 
 export const stixLoadByIdStringify = async (context, user, id) => {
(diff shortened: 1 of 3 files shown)


opencti-platform/opencti-graphql/src/database/middleware.js
```diff
@@ -544,8 +544,10 @@ export const stixLoadByIds = async (context, user, ids, opts = {}) => {
 };
 export const stixBundleByIdStringify = async (context, user, type, id) => {
   const resolver = modules.get(type)?.bundleResolver;
-  if (resolver) return await resolver(context, user, id);
-  throw UnsupportedError('No bundle resolver for type', { type });
+  if (resolver) {
```
esrevi: I think inverting as early return pattern is easier for readability and keeps the happy path at the main indentation level :)

```suggestion
  if (!resolver) {
    return null
  }
  
  return await resolver(context, user, id);
```

SouadHadjiat (approved):
tested locally ✅

PR #13012, opened 2025-10-30
[backend] change SSO absolute redirection (#13011)

### LINKED ISSUE #13011: SSO absolute redirection prevent login
## Description

Instead of finish login, the app force redirection to the APP__BASE which when browser is accessing website with a different baseURL.

## Environment

OpenCTI version: From 6.8.3

aHenryJard:
### Proposed changes

* Keep relative URL in SSO callback

### Related issues

* closes #13011

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/http/httpPlatform.js
+++ b/opencti-platform/opencti-graphql/src/http/httpPlatform.js
@@ -40,7 +40,11 @@ export const sanitizeReferer = (refererToSanitize) => {
   const resolvedUrl = new URL(refererToSanitize, base).toString();
   if (resolvedUrl === base || resolvedUrl.startsWith(`${base}/`)) {
     // same domain URL accept the redirection
-    return refererToSanitize;
+    if (refererToSanitize.startsWith('/')) {
+      // in case of relative URL, keep relative.
+      return refererToSanitize;
+    }
+    return resolvedUrl;
   }
   logApp.info('Error auth provider callback : url has been altered', { url: refererToSanitize });
   return '/';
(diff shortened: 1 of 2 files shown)


lndrtrbn (approved):

opencti-platform/opencti-graphql/src/http/httpPlatform.js
```diff
@@ -35,16 +35,19 @@ import { setCookieError } from './httpUtils';
...
 
 export const sanitizeReferer = (refererToSanitize) => {
+  if (!refererToSanitize) return '/';
   const base = getBaseUrl();
-  if (!refererToSanitize) return base;
-
   const resolvedUrl = new URL(refererToSanitize, base).toString();
   if (resolvedUrl === base || resolvedUrl.startsWith(`${base}/`)) {
     // same domain URL accept the redirection
+    if (refererToSanitize.startsWith('/')) {
```
xfournet, line 43: i'm not sure why is it needed ?

Anyway, to test a relative URL you should also test that it is not a protocol relative URL, eg not starting with '//' since protocol relative URL permit to specify a different host

aHenryJard: without this an simple IP can be inferred (and is not redirect to baseURL or '/' )  this use case => https://github.com/OpenCTI-Platform/opencti/blob/694a38704f341897383db283704a42eef91588c3/opencti-platform/opencti-graphql/tests/01-unit/http/httpPlatform-test.ts#L80

xfournet: if Location is 22.0.0.1 it's not a valid URL, so it's considered as a relative path, eg redirecting to https://demo.octi.filigran.io/22.0.0.1

PR #13018, opened 2025-10-30
[CI] Fix octi_config extra var in awx call

efaure:
### Proposed changes

*  Fix octi_config extra var in awx call 
* Wait for the awx job to end


efaure:
@OctaveLaventure  This PR is a small fix to be able to manage configuration of env var for feature branch. 
It's already tested https://github.com/OpenCTI-Platform/opencti/actions/runs/18947795730
I also added a parameter to wait for the awx job to end. It should also retrieve the status of the awx job and fail the github action workflow. Without it, it was difficult to know if the deployment was successful. 

@Kroustille When this PR will be merge, you can rebase and deploy your feature branche.

OctaveLaventure (approved):

Issue #12703, opened 2025-10-10
fix(graph): grouping correlation unfix button changes the "show" filter


fruitcakej:
## Description
Found an odd bug when looking at the correlation view of a grouping. Not all the times but sometimes when clicking the unfix nodes button, it’ll swap the “Show all correlated entries” view to “Show only correlated observables and indicators” without any indicator that it has done that (the “Show only correlated observables and indicators” button still stays highlighted).

## Environment
6.8.3

## Reproducible Steps
If you follow the order of the attached screencast you can recreate

## Screenshots (optional)

![Image](https://github.com/user-attachments/assets/b407929e-98d3-44b8-803c-d07b5d1b44b5)

nino-filigran:
Repro steps are: 
- Go in correlation view (from grouping overview click on correlation or go in knowledge view & select correlation view)
- Click on Show all correlated entities
- Click on unfix the nodes

Expected result: 
- all correlated entities are displayed but with nodes unfixed.

Current result: 
- the view switches from all entities displayed but only correlated IOC and Observables are shown (and the CTA to show the view mode in the bottom nav bar is showing that all entities are displayed)

PR #13021, opened 2025-10-31
fix(frontend): correlation unfix button changes the "show" filter (#12703)

Gwendoline-FAVRE-FELIX:
### Proposed changes

* Remove setRawObjects in rebuildGraphData

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/12703


opencti-platform/opencti-front/src/components/graph/utils/useGraphInteractions.ts
```diff
@@ -383,7 +383,6 @@ const useGraphInteractions = () => {
         o.entity_type === 'Indicator' || o.parent_types.includes('Stix-Cyber-Observable')
       ))
       : objects;
-    setRawObjects(filteredObjects);
```
lndrtrbn, line 386: I have the feeling that removing this line will add new unexpected behavior because of a desync with raw objects. I do not think the bug should be fixed this way.

Issue #12922, opened 2025-10-24
Notes header is not aligned


labo-flg:
## Description

In an entity details, we can create notes.
The notes panel has a header, but buttons seem misaligned.

First, the vertical alignment is off.

Then, I think we shall not mix frameless (burger) button with outlined (enrichment) button.
Also, I would rather have the burger button as the last one, on the far right.
We can align with what we have in the entity header.

<img width="1735" height="276" alt="Image" src="https://github.com/user-attachments/assets/1ddd54fa-d523-4523-8276-da3a74505f5a" />

## Environment
6.8.8

alice-debra:
@esrevi Can we close this bug?

esrevi:
Yes we can ! :)

PR #13032, opened 2025-10-31
[frontend] notes header alignment (#12922)

esrevi:
### Proposed changes
* Enhance note card behaviour
* Invert vert dots button with enrichment button
* Update vert dots icon to match the style of the action buttons on each page
* Fix scroll window when expanding the accordion write a note 
* Scroll window when clicking on "more fields" so we can see the extra fields
* On created note, scroll up to the note created because when there are some notes, it was looking like it wasn't created but as the notes are ordered desc by date, we don't see the new one.
* Refactor code with grid2, remove deprecated makestyles, split components

https://github.com/user-attachments/assets/b40e64f7-ef37-4b8c-895d-44bdea571567

### Related issues

* #12922 

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/analyses/notes/StixCoreObjectOrStixCoreRelationshipNoteCard.tsx
+++ b/opencti-platform/opencti-front/src/private/components/analyses/notes/StixCoreObjectOrStixCoreRelationshipNoteCard.tsx
@@ -17,6 +17,7 @@ import makeStyles from '@mui/styles/makeStyles';
 import { useTheme } from '@mui/styles';
 import Chip from '@mui/material/Chip';
 import DialogTitle from '@mui/material/DialogTitle';
+import { Stack } from '@mui/material';
 import { useFormatter } from '../../../../components/i18n';
 import { noteMutationRelationDelete } from './AddNotesLines';
 import NotePopover from './NotePopover';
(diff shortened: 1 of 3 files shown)


opencti-platform/opencti-front/src/private/components/analyses/notes/StixCoreObjectOrStixCoreRelationshipNoteCard.tsx
```diff
@@ -161,127 +139,122 @@ StixCoreObjectOrStixCoreRelationshipNoteCardComponentProps
...
-            >
-              <ItemMarkings
-                variant="inList"
-                markingDefinitions={note.objectMarking ?? []}
-                limit={1}
-              />
-            </div>
-          </div>
+          <Stack direction="row" alignItems="center" justifyContent="space-between" spacing={1} sx={{ marginTop: 0.5 }}>
+            <Stack direction="row" spacing={1}>
```
CelineSebe, line 147: Could you tell me what's the goal of replacing this div by Stack ?

CelineSebe:
Works well :)

CelineSebe (approved):

opencti-platform/opencti-front/src/private/components/analyses/notes/StixCoreObjectOrStixCoreRelationshipNoteCard.tsx
esrevi: Sure, since we're using MUI, it automatically uses MUI spacing system and supports responsive props without media queries, and make the code more maintainable imo because we replace the style css with props directly, it avoids having <div style={{ display: 'flex', gap: '16px' }} /> for example, we will have <Stack spacing={2} /> (better readability in my opinion)

PR #13038, opened 2025-11-03
[backend] Update GraphQL telemetry log regarding unnamed queries

xfournet:
### Proposed changes

* Log indicate an harmless condition, no need to be in ERROR level -> INFO
* Log message should be in lower case

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/graphql/telemetryPlugin.js
+++ b/opencti-platform/opencti-graphql/src/graphql/telemetryPlugin.js
@@ -36,7 +36,7 @@ export default {
         };
 
         if (!sendContext.operationName) {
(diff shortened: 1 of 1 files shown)


opencti-platform/opencti-graphql/src/graphql/telemetryPlugin.js
```diff
@@ -36,7 +36,7 @@ export default {
         };
 
         if (!sendContext.operationName) {
-          logApp.error('TELEMETRY PLUGIN UNDEFINED OPERATION', { query: stripIgnoredCharacters(sendContext.request?.query ?? 'undefined') });
+          logApp.info('[TELEMETRY] GraphQL operation is not named', { query: stripIgnoredCharacters(sendContext.request?.query ?? 'undefined') });
```
OctaveLaventure: we can use "unnamed" instead of "not named", it is more formal and standard for technical documentation

OctaveLaventure (approved):

PR #13065, opened 2025-11-07
Issue/12362 update

MTorbay-Filigran:
### Proposed changes

* This PR allows playbooks to listen to updates of flagged entities in a Pir 
* We changed the default message for notifier when an update is made on an entity to match the event message instead of a create message
* We made the listen Pir Utils to listen to external events so that it can get an update of an entity and not only the pir relationship, and from that update, we added logic to know if the updated entity is in one of the selected Pir in the playbook
* When it is the case then the entity is sent in the playbookExecutor 

### Related issues

* https://github.com/orgs/OpenCTI-Platform/projects/8/views/51?pane=issue&itemId=128324166&issue=OpenCTI-Platform%7Copencti%7C12362
*

### Checklist


MTorbay-Filigran:
Closed because already opened

PR #10870, opened 2025-04-28
Create Relationship FAB Replacement for Threats

Bonsai8863:
### Proposed changes

* Added dedicated button "Create Relationship" in entities headers for Threats.
* Added context provider so "Create Relationship" drawer is context-aware. Only shows Malwares when on the Malwares tab, for example.
* Replaced the create entity/observable speeddial and FAB with dedicated buttons in the drawer's header.
* Hid Create Relationship FAB in Threats knowledge.

### Related issues

*
*

### Checklist

- [x] I consider the submitted work as finished
- [x] I tested the code for its functionality
- [x] I wrote test cases for the relevant uses case (coverage and e2e)
- [ ] I added/update the relevant documentation (either on github or on notion)
- [x] Where necessary I refactored code to improve the overall quality

### Further comments

I have only made these modifications to Threats (TAIs, TAGs, Intrusion Sets, and Campaigns). ~~I also left the original create relationship FAB for now as a way to sanity check the new button.~~ I hid the original create relationship FAB for just Threat knowledge. I didn't replace much of the existing code so to not break as many things. I can easily extend this code to the remaining entity types and remove the create relationship FAB if this looks good. For now, it'd be easier to make changes and bug fixes if I only made the code applicable to a small group of entities.

PR #13076, opened 2025-11-07
[frontend] temporary fix for hidden controlled dial button

### LINKED ISSUE #13007: Missing “+” button to create object on the fly when adding a relationship
## Description

It’s no longer possible to create new entities “on the fly” when adding a relationship to an existing entity.

Previously, a “+” button was available in the bottom-right corner of the side panel to create a new relation directly from the relationship creation view. This button has now disappeared.

Image

## Environment

OCTI 6.8.8

## Reproducible Steps

1. Go to an Intrusion Set entity.
2. Open the Knowledge tab, then go to the Malware section.
3. Click the “+” button in the bottom-right corner to add a relationship.
-> In the side panel, if the malware you want to relate does not exist yet, there is no option to create it. The “+” button that used to allow creating a new entity is not here anymore.

## Expected Output

The “+” button should be visible in the side panel, allowing users to create a new entity (e.g., Malware, Tool, Campaign, etc.) when it does not already exist.

## Actual Output

The “+” button is missing, preventing users from creating new entities directly during the relationship creation process.

OctaveLaventure:
This bug is already fixed in PR https://github.com/OpenCTI-Platform/opencti/pull/10870 which is merged in release/current, so this is just to fix this bug for clients until the next release

### Proposed changes

* fix the `Create <entity>` button on the Create Relationship view. I kept the fix super minimal as this is quite urgent and to minimize conflicts when we will merge the next release, as this bug has already been fixed in https://github.com/OpenCTI-Platform/opencti/pull/10870 with a pretty big refacto, so this will reduce the conflicts scope

### Related issues

* closes https://github.com/OpenCTI-Platform/opencti/issues/13007

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/common/stix_core_relationships/StixCoreRelationshipCreationFromEntity.tsx
+++ b/opencti-platform/opencti-front/src/private/components/common/stix_core_relationships/StixCoreRelationshipCreationFromEntity.tsx
@@ -846,6 +846,8 @@ const StixCoreRelationshipCreationFromEntity: FunctionComponent<StixCoreRelation
               speeddial={undefined}
               handleClose={undefined}
               onCompleted={undefined}
+              // TODO: remove with release 6.9
+              controlledDialStyles={{ position: 'fixed', top: 16, right: 16 }}
             />
           )}
           {targetEntities.length === 0 && isOnlySCOs && (
@@ -906,6 +908,8 @@ const StixCoreRelationshipCreationFromEntity: FunctionComponent<StixCoreRelation
                 isFromBulkRelation={undefined}
                 defaultMarkingDefinitions={undefined}
                 stixDomainObjectTypes={undefined}
+                // TODO: remove with release 6.9
+                controlledDialStyles={{ position: 'fixed', top: 16, right: 16 }}
               />
               <StixCyberObservableCreation
                 display={open}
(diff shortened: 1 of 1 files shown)


opencti-platform/opencti-front/src/private/components/common/stix_core_relationships/StixCoreRelationshipCreationFromEntity.tsx
```diff
@@ -846,6 +846,8 @@ const StixCoreRelationshipCreationFromEntity: FunctionComponent<StixCoreRelation
               speeddial={undefined}
               handleClose={undefined}
               onCompleted={undefined}
+              // TODO: remove after next release
```
Archidoit:
```suggestion
              // TODO: remove after 6.9
```

opencti-platform/opencti-front/src/private/components/common/stix_core_relationships/StixCoreRelationshipCreationFromEntity.tsx
```diff
@@ -906,6 +908,8 @@ const StixCoreRelationshipCreationFromEntity: FunctionComponent<StixCoreRelation
                 isFromBulkRelation={undefined}
                 defaultMarkingDefinitions={undefined}
                 stixDomainObjectTypes={undefined}
+                // TODO: remove after next release
```
Archidoit: same

Archidoit (approved):

PR #13090, opened 2025-11-10
[frontend] Allow immediate deletion of managed connectors after stop request (#12474)

### LINKED ISSUE #12474: [composer] when a connector instance is stopped, do not wait ping timeout to be able to delete it
## Description
When stopping a connector instance, the composer won't allow us to delete it until it has detected that the process is dead.
This means waiting a long timeout.

## Environment

6.8.0

## Reproducible Steps

start an instance, stop it, and wait for the delete action to be available.

## Expected Output

Once stopped we should be able to delete the instance

maelv-filigran:
### Proposed changes

* Modified the deletion button logic in `ConnectorPopover.tsx` to enable deletion of managed connectors as soon as stop has been requested
* Previously, users had to wait for the composer to detect the process was dead before deletion was allowed

### Related issues

* #12474 

### Checklist

### Further comments

This is a UX improvement that eliminates unnecessary waiting time. The change is minimal and focused:
- Built-in connectors: deletion remains disabled (unchanged)
- Managed connectors: deletion enabled immediately when stop is requested
- Non-managed connectors: original behavior preserved

The backend/composer will handle cleanup of stopping/stopped containers appropriately.

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/data/connectors/ConnectorPopover.tsx
+++ b/opencti-platform/opencti-front/src/private/components/data/connectors/ConnectorPopover.tsx
@@ -162,7 +162,22 @@ const ConnectorPopover = ({ connector, onRefreshData }: ConnectorPopoverProps) =
           <MenuItem onClick={handleOpenResetState}>{t_i18n('Reset the connector state')}</MenuItem>
         )}
         <MenuItem onClick={handleOpenClearWorks}>{t_i18n('Clear all works')}</MenuItem>
-        <MenuItem onClick={handleOpenDelete} disabled={!!connector.active || !!connector.built_in}>{t_i18n('Delete')}</MenuItem>
+        <MenuItem
+          onClick={handleOpenDelete}
+          disabled={(() => {
+            if (connector.built_in) return true;
+            if (connector.is_managed) {
+              // Allow deletion once stop has been requested or connector is stopped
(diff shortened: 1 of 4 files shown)


opencti-platform/opencti-front/src/private/components/data/connectors/ConnectorPopover.tsx
```diff
@@ -162,7 +162,22 @@ const ConnectorPopover = ({ connector, onRefreshData }: ConnectorPopoverProps) =
           <MenuItem onClick={handleOpenResetState}>{t_i18n('Reset the connector state')}</MenuItem>
         )}
         <MenuItem onClick={handleOpenClearWorks}>{t_i18n('Clear all works')}</MenuItem>
-        <MenuItem onClick={handleOpenDelete} disabled={!!connector.active || !!connector.built_in}>{t_i18n('Delete')}</MenuItem>
+        <MenuItem
+          onClick={handleOpenDelete}
+          disabled={(() => {
+            if (connector.built_in) return true;
```
esrevi: I think it could be great to extract this function into a helper file for example and call this function canDeleteConnector(connector) and add a test 

because, we also need to call this function in ConnectorStatus.tsx component, there's a delete button too that should use the same domain rule I think (you can find it on line 380 of that file) :)

esrevi (approved):
lgtm

PR #13094, opened 2025-11-10
[backend] entities and relations counters contants in backend tests

Archidoit:
### Proposed changes
Create constants for entities and relations counters in backend tests to make tests counters easier to maintain


opencti-platform/opencti-graphql/tests/utils/entityCountHelper.ts
```diff
@@ -1,11 +1,46 @@
 export const entitiesCounter: Record<string, number> = {};
-entitiesCounter.capability = 51;
-entitiesCounter.entitySetting = 45;
-entitiesCounter.label = 13;
-entitiesCounter.statusTemplate = 8;
-entitiesCounter.externalReference = 7;
+entitiesCounter.Capability = 51;
```
esrevi: I think it would be better to define the constants like this:

```
export const entitiesCounter = {
  Credential: 1,
  DecayRule: 4,
  ...
} as const
```

Benefits: the IDE will have better autocompletion and as const, will make this object immutable :)

opencti-platform/opencti-graphql/tests/utils/entityCountHelper.ts
```diff
@@ -1,11 +1,46 @@
...
+entitiesCounter.Sector = 3;
+entitiesCounter.Status = 7;
+entitiesCounter.StatusTemplate = 8;
+entitiesCounter.StixFile = 1;
+entitiesCounter.Vocabulary = 354;
 
 export const entitiesCounterTotal = Object.values(entitiesCounter).reduce((acc, value) => acc + value, 0);
 
 export const relationsCounter: Record<string, number> = {};
+relationsCounter['accesses-to'] = 28;
```
esrevi: should the attributes be aligned like the entitiesCounter in PascalCase ?

Archidoit: entity types are pascal cases (exemple : Report), whereas relationshp types are not (ex: located-at)

opencti-platform/opencti-graphql/tests/utils/entityCountHelper.ts
Archidoit: gonna fix this! thanks !!!

esrevi (approved):
lgtm !

Issue #13096, opened 2025-11-12
When creating a PIR, the dialog does not have the default behavior


SamuelHassine:
## Description

When creating a PIR, the dialog does not have the default behavior

<img width="1225" height="794" alt="Image" src="https://github.com/user-attachments/assets/5167ae65-9947-4228-adb8-4643afe500c9" />

You cannot click outside to close the dialog like others, just on cancel.

PR #13102, opened 2025-11-12
[frontend] Fix closing PIR creation form (13096)

Archidoit:
### Proposed changes
When clicking outside the PIR creation form, the form should close.

image

### Related issues
#13096


opencti-platform/opencti-front/src/private/components/pir/pir_form/PirCreationForm.tsx
```diff
@@ -27,11 +28,11 @@ import { PirCreationFormGeneralSettingsRedisStreamQuery } from './__generated__/
 import Loader, { LoaderVariant } from '../../../../components/Loader';
 
 interface PirCreationFormProps {
-  onCancel: () => void
+  onClose: () => void
```
lndrtrbn: Why renaming to onClose here? It's the callback of the Cancel button so I think it should stay onCancel as name

Archidoit: I fix this :)

lndrtrbn (approved):

Issue #3360, opened 2023-05-20
In massive search, when something is not found, be able to quick add


SamuelHassine:
## Use case

In massive search, when something is not found, be able to quick add with a button.

Jipegien:
What do you call "massive search" ?

SamuelHassine:
Bulk search:

![image](https://github.com/OpenCTI-Platform/opencti/assets/1334279/3b0f4b3a-543f-4a2e-b211-bfe462cab6d9)

Jipegien:
- Simple "+" button inline with the something not found.
- Create form needs to have a way (form's list field / first step with a list of all entities as when addind Observables to Cases)
- when create form is validated, we need to stay in the bulk result page and update the line with the entity that have been created.

PR #13104, opened 2025-11-12
[frontend] Bulk Search: Unknown entities bulk creation (#3360)

Archidoit:
### Proposed changes
in Bulk Search, in the Unknown entities tab
- add toolbar with an icon to bulk create the entities
- when clicking on the icon, a pop up opens to create the selected unknown entities via the already existing workflow to import data

### Related issues
#7279
#3360

### DIFF
--- a/opencti-platform/opencti-front/src/components/dataGrid/DataTableWithoutFragment.tsx
+++ b/opencti-platform/opencti-front/src/components/dataGrid/DataTableWithoutFragment.tsx
@@ -77,9 +77,9 @@ const DataTableWithoutFragment = (props: OCTIDataTableProps & {
       resolvePath={(a) => a}
       initialValues={{}}
       disableLineSelection={!taskScope}
-      dataTableToolBarComponent={taskScope
+      dataTableToolBarComponent={taskScope && Array.isArray(data) && data[0].has('id')
         ? <DataTableWithoutFragmentInternalToolbar
-            dataIds={(data as { id: string, value: string, entity_type: string }[]).map((d) => d.id)}
+            dataIds={data.map((d) => d.id)}
             taskScope={taskScope}
           />
         : undefined
--- a/opencti-platform/opencti-front/src/components/dataGrid/DataTableWithoutFragment.tsx
+++ b/opencti-platform/opencti-front/src/components/dataGrid/DataTableWithoutFragment.tsx
@@ -1,6 +1,10 @@
 import React from 'react';
+import DataTableToolBar from '@components/data/DataTableToolBar';
+import { useTheme } from '@mui/styles';
 import type { DataTableProps } from './dataTableTypes';
 import DataTableComponent from './components/DataTableComponent';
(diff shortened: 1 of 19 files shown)


opencti-platform/opencti-front/src/private/components/common/files/import_files/ImportFilesFreeText.tsx
```diff
@@ -10,7 +10,13 @@ import { now } from '../../../../../utils/Time';
 
 type FileFreeTextType = { content: string };
 
-const ImportFilesFreeText = ({ onSumbit, onClose }: { onSumbit: (file: File) => void; onClose: () => void }) => {
+interface ImportFilesFreeTextProps {
+  onSumbit: (file: File) => void;
```
esrevi:
```suggestion
  onSubmit: (file: File) => void;
```

opencti-platform/opencti-front/src/components/dataGrid/DataTableWithoutFragment.tsx
```diff
@@ -34,7 +76,14 @@ const DataTableWithoutFragment = (props: OCTIDataTableProps) => {
       dataQueryArgs={(line: never) => line}
       resolvePath={(a) => a}
       initialValues={{}}
-      disableLineSelection={true}
+      disableLineSelection={!taskScope}
+      dataTableToolBarComponent={taskScope
+        ? <DataTableWithoutFragmentInternalToolbar
+            dataIds={(data as { id: string, value: string, entity_type: string }[]).map((d) => d.id)}
```
esrevi: I guess data is always an array, but since data is typed as unknown, should we typeguard data first ? like checking if data is an array and id is present in every element of data ? 
or change the type of data with a generic instead of unknown ? but it may be a bigger refacto ^^

opencti-platform/opencti-front/src/private/components/common/files/import_files/ImportFilesFreeText.tsx
Archidoit: fixed :)

opencti-platform/opencti-front/src/components/dataGrid/DataTableWithoutFragment.tsx
Archidoit: data is object[]

But if taskScope is specified, we are in the case of 'unknown entities' and we know it has the format: { id: string, value: string, entity_type: string }[] <===> object & { id: string }[]

Archidoit: I tried an improvement

Archidoit: data is of type object[] | null | undefined | object


So I will let data as unknown since it can be anything, and let the special types in case taskScope is defined

esrevi (approved):
lgtm ! :)

PR #13109, opened 2025-11-13
[backend/frontend] Integration Manager warning banner not reappearing when inactive (#12895) 

### LINKED ISSUE #12895: XTM-Composer/Connector catalog `active` status expire
## Description

Currently once XTM-Composer is seen by OpenCTI, there will not be any warning that XTM composer is not connected anymore.

The `active` status could expire after some time without XTM-Composer checking in.

## Environment

1. OS (where OpenCTI server runs): 
2. OpenCTI version: 6.8.7
4. Other environment details:

## Reproducible Steps

1. Deploy OCTI
2. Observe a warning telling that the XTM-Composer is not available
3. Deploy XTM-Composer
4. Warning is gone
5. Remove XTM-Composer
6. Warning is not shown.

## Expected Output

Display a warning that XTM-Composer is not available

## Actual Output

No warning.

maelv-filigran:
### Proposed changes

* Updated frontend logic to check `hasActiveManagers` instead of `hasRegisteredManagers` 
* Warning banner now correctly reappears when Integration Manager stops responding (~5-6 min after XTM-Composer stops)
* Added `hasActiveManagers` computed property to ConnectorManagerStatusContext
* Updated ConnectorDeploymentBanner and related components to use active status
* Form fields now disabled when no active managers available

### Related issues

* #12895

### Checklist

### Further comments

**Root Cause**: The frontend was checking if any connector managers existed in the database (`hasRegisteredManagers`) instead of checking if they were actively responding (`hasActiveManagers`).

**Solution**: Frontend-only fix. The backend already correctly computes the `active` status via `sinceNowInMinutes(cm.updated_at) < 5` and the frontend already polls for it every 30 seconds. We simply needed to use the correct boolean flag.

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/data/IngestionCatalog/IngestionCatalogConnector.tsx
+++ b/opencti-platform/opencti-front/src/private/components/data/IngestionCatalog/IngestionCatalogConnector.tsx
@@ -37,7 +37,7 @@ const ingestionCatalogConnectorQuery = graphql`
 
 interface IngestionCatalogConnectorComponentProps {
   queryRef: PreloadedQuery<IngestionCatalogConnectorQuery>;
-  onClickDeploy: (connector: IngestionConnector, catalogId: string, hasRegisteredManagers: boolean, deploymentCount: number) => void;
+  onClickDeploy: (connector: IngestionConnector, catalogId: string, hasRegisteredManagers: boolean, hasActiveManagers: boolean, deploymentCount: number) => void;
   openConfig?: boolean;
 }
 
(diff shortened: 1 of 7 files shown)


opencti-platform/opencti-front/src/private/components/data/IngestionCatalog/IngestionCatalogConnector.tsx
```diff
@@ -37,7 +37,7 @@ const ingestionCatalogConnectorQuery = graphql`
 
 interface IngestionCatalogConnectorComponentProps {
   queryRef: PreloadedQuery<IngestionCatalogConnectorQuery>;
-  onClickDeploy: (connector: IngestionConnector, catalogId: string, hasRegisteredManagers: boolean, deploymentCount: number) => void;
+  onClickDeploy: (connector: IngestionConnector, catalogId: string, hasRegisteredManagers: boolean, hasActiveManagers: boolean, deploymentCount: number) => void;
```
esrevi: I think we should remove hasRegisteredManagers arg since we don't use it anymore

maelv-filigran: fixed! thanks :)

esrevi (approved):
lgtm ! :)

PR #13129, opened 2025-11-13
[backend] add telemetry on SSO usages (#13101)

### LINKED ISSUE #13101: Add telemetry to understand SSO type deployed per platform
## Use case

Because we want to work on SSO and the possibility of defining it within UI, we would like to understand what SSO type platforms are using.
We should have 4 distinct telemetry points:
- Local
- LDAP
- OpenID
- SAML

aHenryJard:
### Proposed changes

- Add a telemetry gauge on all SSO strategy

#### Refactoring
* Split providers.js into 2 files to avoid cyclic dependencies (telemetry - user -provider)
* Add some types
* Removed unused counter on onboarding

### Related issues

* closes #13101 

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/config/providers.js
+++ b/opencti-platform/opencti-graphql/src/config/providers.js
@@ -105,16 +105,16 @@ const configRemapping = (config) => {
 
 // Providers definition
 export const INTERNAL_SECURITY_PROVIDER = '__internal_security_local_provider__';
-const STRATEGY_LOCAL = 'LocalStrategy';
+export const STRATEGY_LOCAL = 'LocalStrategy';
 export const STRATEGY_CERT = 'ClientCertStrategy';
-const STRATEGY_HEADER = 'HeaderStrategy';
-const STRATEGY_LDAP = 'LdapStrategy';
-const STRATEGY_OPENID = 'OpenIDConnectStrategy';
-const STRATEGY_FACEBOOK = 'FacebookStrategy';
-const STRATEGY_SAML = 'SamlStrategy';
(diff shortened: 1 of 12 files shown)


opencti-platform/opencti-graphql/src/config/providers.js
```diff
@@ -620,11 +604,8 @@ for (let i = 0; i < providerKeys.length; i += 1) {
         });
     });
     passport.use('local', adminLocalStrategy);
-    providers.push({ name: INTERNAL_SECURITY_PROVIDER, type: AUTH_FORM, strategy, provider: 'local' });
+    PROVIDERS.push({ name: INTERNAL_SECURITY_PROVIDER, type: AuthType.AUTH_FORM, strategy, provider: 'local' });
   }
 }
 
-export const PROVIDERS = providers;
```
aHenryJard, line 627: I have no idea why this assignation was needed.

JeremyCloarec (commented):
Overall the refactoring looks good to me, great job 👍 
I had a question though before approving. I'm not familiar enough with providers, but could it be possible to have multiple providers of the same strategy? In which case, wouldn't it be useful to send the number of activated providers per strategy

JeremyCloarec (approved):

aHenryJard:
> Overall the refactoring looks good to me, great job 👍 I had a question though before approving. I'm not familiar enough with providers, but could it be possible to have multiple providers of the same strategy? In which case, wouldn't it be useful to send the number of activated providers per strategy

So answer from product: we keep a boolean it's more easy to understand usage.

PR #13164, opened 2025-11-14
[backend] engine.ts

JeremyCloarec:
### Proposed changes

* engine.ts

### Related issues

*
*

### Checklist

### Further comments

It's far from perfect, but a first step towards making our engine layer more robust and maintainable.
A lot of things are not properly typed or typed as `any`, and still need improvements.
Next steps would be:
* split engine concerns into seperate files: raw engine layer, engine query builder & filter handling, platform init & mapping etc...
* improve types in current engine.ts: create a type for es body & filters etc...
* move remaining js files to ts: data-builder, data-initialization etc...

### DIFF
--- a/opencti-platform/opencti-graphql/src/database/draft-engine.ts
+++ b/opencti-platform/opencti-graphql/src/database/draft-engine.ts
@@ -279,8 +279,8 @@ export const resolveDraftUpdateFiles = async (context: AuthContext, user: AuthUs
     const loadedFileValues = [];
     for (let i = 0; i < fileIds.length; i += 1) {
       const currentFileId = fileIds[i];
-      const currentFile = await loadFile(context, user, currentFileId);
-      const currentFileContent = toBase64(await getFileContent(currentFileId));
+      const currentFile = await loadFile(context, user, currentFileId as string);
+      const currentFileContent = toBase64(await getFileContent(currentFileId as string));
       if (currentFile) {
         const currentFileObject = {
           name: currentFile.name,
--- a/opencti-platform/opencti-graphql/src/database/draft-engine.ts
+++ b/opencti-platform/opencti-graphql/src/database/draft-engine.ts
@@ -32,8 +32,11 @@ import { buildRefRelationKey } from '../schema/general';
 import { getFileContent } from './raw-file-storage';
 import { loadFile } from './file-storage';
 import { EditOperation } from '../generated/graphql';
+import type { AuthContext, AuthUser } from '../types/user';
+import type { BasicStoreCommon, BasicStoreRelation, InternalEditInput, StoreRelation } from '../types/store';
+import type { S3File } from './file-storage-helper';
 
-const completeDeleteElementsFromDraft = async (context, user, elements) => {
+const completeDeleteElementsFromDraft = async (context: AuthContext, user: AuthUser, elements: BasicStoreCommon[]): Promise<void> => {
   const draftContext = getDraftContext(context, user);
   if (!draftContext) { return; }
   await elDeleteInstances(elements);
@@ -41,11 +44,11 @@ const completeDeleteElementsFromDraft = async (context, user, elements) => {
   await elRemoveDraftIdFromElements(context, user, draftContext, elementsIds);
 };
 
-const isCreateOrDraftDelete = (draftOp) => {
+const isCreateOrDraftDelete = (draftOp: string): boolean => {
   return draftOp === DRAFT_OPERATION_CREATE || draftOp === DRAFT_OPERATION_DELETE || draftOp === DRAFT_OPERATION_DELETE_LINKED;
 };
 
-const elRemoveCreateElementFromDraft = async (context, user, element) => {
+const elRemoveCreateElementFromDraft = async (context: AuthContext, user: AuthUser, element: BasicStoreCommon): Promise<void> => {
   if (element.draft_change?.draft_operation !== DRAFT_OPERATION_CREATE) {
     return;
   }
@@ -53,7 +56,7 @@ const elRemoveCreateElementFromDraft = async (context, user, element) => {
   // We get all relations that were created in draft that target this element (should be all of the relations in this case, since element itself was created in draft)
   const draftCreatedRelations = relations.filter((f) => f.draft_change && f.draft_change.draft_operation === DRAFT_OPERATION_CREATE);
   // Add element to relations to get the impacts from if current element is itself a relation
-  const relationToRemove = draftCreatedRelations.concat(isBasicRelationship(element.entity_type) ? [element] : []);
+  const relationToRemove = draftCreatedRelations.concat(isBasicRelationship(element.entity_type) ? [element as BasicStoreRelation] : []);
   const draftRelationsElementsImpact = await computeDeleteElementsImpacts(relationToRemove, [element.internal_id], relationsToRemoveMap);
 
   // Clean up all denormalized rel impact of relations deletion, then delete all relations
@@ -62,15 +65,17 @@ const elRemoveCreateElementFromDraft = async (context, user, element) => {
   await elDeleteInstances([element, ...draftCreatedRelations]);
 };
 
-const elRemoveUpdateElementFromDraft = async (context, user, element) => {
+const elRemoveUpdateElementFromDraft = async (context: AuthContext, user: AuthUser, element: BasicStoreCommon): Promise<void> => {
   if (element.draft_change?.draft_operation !== DRAFT_OPERATION_UPDATE) {
     return;
   }
 
   // apply reverse field patch
   const elementWithRefs = await storeLoadByIdWithRefs(context, user, element.internal_id);
-  const reverseUpdateFieldPatch = buildReverseUpdateFieldPatch(element.draft_change.draft_updates_patch);
-  await updateAttributeFromLoadedWithRefs(context, user, elementWithRefs, reverseUpdateFieldPatch);
+  if (element.draft_change.draft_updates_patch) {
+    const reverseUpdateFieldPatch = buildReverseUpdateFieldPatch(element.draft_change.draft_updates_patch);
+    await updateAttributeFromLoadedWithRefs(context, user, elementWithRefs, reverseUpdateFieldPatch);
+  }
   // TODO: clean up UPDATE_LINKED impacted elements that no longer need to be in draft => how to know that an update_linked element can be safely removed?
 
   // verify if element can be entirely removed from draft or if it needs to be kept as update_linked
@@ -86,7 +91,11 @@ const elRemoveUpdateElementFromDraft = async (context, user, element) => {
   }
 };
 
-const removeDraftDeleteLinkedRelations = async (context, user, deleteLinkedRelations) => {
+const removeDraftDeleteLinkedRelations = async (
+  context: AuthContext,
+  user: AuthUser,
+  deleteLinkedRelations: { rel: any, dep: any }[]
+): Promise<void> => {
   // Reapply denormalized refs on elements impacted with deleteLinked rel removal
   const elementsToUpdate = deleteLinkedRelations.map((deleteLinkedRelToRemove) => {
     const { rel, dep } = deleteLinkedRelToRemove;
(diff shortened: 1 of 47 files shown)


marieflorescontact (commented):
looks good functionnaly, I can't see anything broken. 
I haven't look at the code yet 😅

MTorbay-Filigran (commented):
Great job! I mainly reviewed the typing, and everything looks good to me.
I noticed that return values for some functions are explicitly typed while others are not, is there a specific reason for that?

For functional testing, do you have a few use cases in mind that we could try to make sure nothing is broken?

JeremyCloarec:
> Great job! I mainly reviewed the typing, and everything looks good to me. I noticed that return values for some functions are explicitly typed while others are not, is there a specific reason for that?
> 
> For functional testing, do you have a few use cases in mind that we could try to make sure nothing is broken?

The fact that some functions are not explicitly typed was just to save some time to be honest 😅 Ideally every return values should be explicitly typed to improve code readability. But that's something that we'll be able to improve in the future now.

As for functional testing, I'd say the things that you can test is making sure that complex filters are still all working properly. Engine is used in almost all queries, so it's already pretty well covered with our test suite, but it could be a great idea to test edge cases manually: making sure that pagination works properly when moving to the next page/going back to the next page, deleting a large item, complex filters etc...

opencti-platform/opencti-graphql/src/database/draft-engine.ts
```diff
@@ -62,15 +64,17 @@ const elRemoveCreateElementFromDraft = async (context, user, element) => {
...
+const elRemoveUpdateElementFromDraft = async (context: AuthContext, user: AuthUser, element: BasicStoreCommon): Promise<void> => {
   if (element.draft_change?.draft_operation !== DRAFT_OPERATION_UPDATE) {
     return;
   }
 
   // apply reverse field patch
   const elementWithRefs = await storeLoadByIdWithRefs(context, user, element.internal_id);
-  const reverseUpdateFieldPatch = buildReverseUpdateFieldPatch(element.draft_change.draft_updates_patch);
-  await updateAttributeFromLoadedWithRefs(context, user, elementWithRefs, reverseUpdateFieldPatch);
+  if (element.draft_change.draft_updates_patch) {
```
SarahBocognano, line 74: Why put a condition here ? Was it missing ?

opencti-platform/opencti-graphql/src/database/draft-engine.ts
```diff
@@ -237,12 +248,12 @@ export const elDeleteDraftContextFromUsers = async (context, user, draftId) => {
         }
       },
     },
-  }).catch((err) => {
+  }).catch((err: any) => {
```
SarahBocognano, line 251: Not possible to have a typed error here ? (curious)

opencti-platform/opencti-graphql/src/modules/draftWorkspace/draftWorkspace-domain.ts
```diff
@@ -58,7 +58,7 @@ export const getObjectsCount = async (context: AuthContext, user: AuthUser, draf
     convertEntityTypeLabel: true,
   };
   const draftContext = { ...context, draft_context: draft.id };
-  const distributionResult = await elAggregationCount(draftContext, context.user, READ_INDEX_DRAFT_OBJECTS, opts);
```
SarahBocognano, line 61: Out of curiosity: why remove the context ?

opencti-platform/opencti-graphql/src/types/store.d.ts
```diff
@@ -66,7 +66,8 @@ interface Representative {
 interface InternalEditInput {
   key: string;
   operation?: EditOperation | null;
-  value: (string | object | null)[];
+  value: any[];
```
SarahBocognano: why changing the value type by any[] ?

opencti-platform/opencti-graphql/src/database/draft-engine.ts
JeremyCloarec: It wasn't really missing, but since I typed the buildReverseUpdateFieldPatch input as string, I needed to make sure that element.draft_change.draft_updates_patch wasn't undefined/null here. This also prevents calling updateAttributeFromLoadedWithRefs when it's not needed

opencti-platform/opencti-graphql/src/database/draft-engine.ts
JeremyCloarec: It's probably possible, but a bit hard to do. We'd need to know what kind of error deleteByQuery can raise, both for opensearch and elastic clients, and I didn't take the time to dive into it. But it'd be more clean to do so you're right :)

opencti-platform/opencti-graphql/src/modules/draftWorkspace/draftWorkspace-domain.ts
JeremyCloarec: Mostly to solve a TS issue: the user in AuthContext is typed as possibly undefined, so using context.user was throwing a TS issue since elAggregationCount doesn't accept an undefined user.
I don't know why we were using context.user instead of user directly here though

opencti-platform/opencti-graphql/src/types/store.d.ts
JeremyCloarec: I was getting a lot of TS errors with the (string | object | null)[] typing, so this was the shortcut I took to make it work.
But that's a good remark, I pushed a new commit to revert to a better typing than any[]! And I applied the needed changes to handle properly the typing

MTorbay-Filigran (approved):
I tested functionnaly with the use cases you gave me and found nothing wrong so I validate :)

opencti-platform/opencti-graphql/src/database/engine.ts
```diff
@@ -2517,62 +2552,29 @@ export const buildLocalMustFilter = (validFilter) => {
...
   if (arrayKeys.length > 1) {
     throw UnsupportedError('A filter with these multiple keys is not supported', { keys: arrayKeys });
   }
   // at this point arrayKeys === ['entity_type']
 
   // we'll build these new filters or filterGroup, depending on the situation
-  let newFilter;
-  let newFilterGroup;
+  let newFilter: Filter | undefined;
+  let newFilterGroup: FilterGroup = emptyFilterGroup;
```
Archidoit:
```suggestion
  let newFilterGroup: FilterGroup | undefined;
```
avoid adding a useless empty filter group in the array filters.filterGroups....

opencti-platform/opencti-graphql/src/database/engine.ts
```diff
@@ -2592,78 +2594,708 @@ const adaptFilterToEntityTypeFilterKey = (filter) => {
...
+  if (arrayKeys[0] !== IDS_FILTER || arrayKeys.length > 1) {
+    throw UnsupportedError('A filter with these multiple keys is not supported', { keys: arrayKeys });
+  }
+  if (mode === 'and') {
+    throw UnsupportedError('Unsupported filter: \'And\' operator between values of a filter with key = \'ids\' is not supported');
+  }
+  // at this point arrayKey === ['ids'], and mode is always 'or'
+
+  // we'll build these new filters or filterGroup, depending on the situation
+  let newFilterGroup: FilterGroup = emptyFilterGroup;
```
Archidoit: same

opencti-platform/opencti-graphql/src/database/engine.ts
```diff
@@ -2592,78 +2594,708 @@ const adaptFilterToEntityTypeFilterKey = (filter) => {
...
-  // at this point arrayKey === ['ids'], and mode is always 'or'
-
-  // we'll build these new filters or filterGroup, depending on the situation
-  let newFilterGroup;
-
-  const idsArray = [...ids_attributes]; // the keys to handle additionally
-
-  if (operator === 'nil' || operator === 'not_nil') { // nil and not_nil operators must have a single key
-    const filters = idsArray.map((idKey) => { return { ...filter, key: idKey }; });
+  let newFilterGroup: FilterGroup = emptyFilterGroup;
```
Archidoit: same for every adaptXXFilter functions

opencti-platform/opencti-graphql/src/database/engine.ts
```diff
@@ -2809,36 +3441,37 @@ const adaptFilterToPirFilterKeys = async (context, user, filterKey, filter) => {
...
     key: ['metrics'],
-    mode: 'and',
+    mode: FilterMode.And,
     nested: [
-      { key: 'name', values: [filter.key], operator: 'eq' },
+      { key: 'name', values: [filter.key], operator: FilterOperator.Eq },
       { key: 'value', values: filter.values, operator: filter.operator, mode: filter.mode },
-    ]
+    ],
+    values: []
```
Archidoit, line 3453: values is useless for a filter with a 'nested' attribute. I guess it's a question of typing...

opencti-platform/opencti-graphql/src/database/engine.ts
```diff
@@ -3066,14 +3640,14 @@ const completeSpecialFilterKeys = async (context, user, inputFilters) => {
...
-          keys.forEach((relKey) => {
+            : types.map((t: any) => buildRefRelationKey(t, '*'));
+          keys.forEach((relKey: any) => {
             regardingFilters.push({ key: [relKey], operator: filter.operator, values: ['EXISTS'] });
           });
         } else {
           const keys = isEmptyField(types)
             ? buildRefRelationKey('*', '*')
-            : types.flatMap((t) => [buildRefRelationKey(t, ID_INTERNAL), buildRefRelationKey(t, ID_INFERRED)]);
+            : types.flatMap((t: any) => [buildRefRelationKey(t, ID_INTERNAL), buildRefRelationKey(t, ID_INFERRED)]);
```
Archidoit:
```suggestion
            : types.flatMap((t: string) => [buildRefRelationKey(t, ID_INTERNAL), buildRefRelationKey(t, ID_INFERRED)]);
```

opencti-platform/opencti-graphql/src/database/engine.ts
```diff
@@ -3066,14 +3640,14 @@ const completeSpecialFilterKeys = async (context, user, inputFilters) => {
         const mode = (filter.operator === 'eq' || isEmptyField(filter.operator)) ? FilterMode.Or : FilterMode.And;
         if (isEmptyField(ids)) {
           const keys = isEmptyField(types) ? buildRefRelationKey('*', '*')
-            : types.map((t) => buildRefRelationKey(t, '*'));
-          keys.forEach((relKey) => {
+            : types.map((t: any) => buildRefRelationKey(t, '*'));
```
Archidoit:
```suggestion
            : types.map((t: string) => buildRefRelationKey(t, '*'));
```

opencti-platform/opencti-graphql/src/database/engine.ts
```diff
@@ -3066,14 +3640,14 @@ const completeSpecialFilterKeys = async (context, user, inputFilters) => {
...
+          keys.forEach((relKey: any) => {
```
Archidoit:
```suggestion
          keys.forEach((relKey: string) => {
```

opencti-platform/opencti-graphql/src/database/engine.ts
```diff
@@ -3262,174 +3836,66 @@ const completeSpecialFilterKeys = async (context, user, inputFilters) => {
       if (arrayKeys.length > 1) {
         throw UnsupportedError('A filter with these multiple keys is not supported', { keys: arrayKeys });
       }
-      const definition = schemaAttributesDefinition.getAttributeByName(key[0]);
+      const definition: any = schemaAttributesDefinition.getAttributeByName(key[0]);
```
Archidoit:
```suggestion
      const definition: AttributeDefinition = schemaAttributesDefinition.getAttributeByName(key[0]);
```

opencti-platform/opencti-graphql/src/database/engine.ts
JeremyCloarec: Yes, the idea was to stick to the Filter type that always requires a values attribute

opencti-platform/opencti-graphql/src/database/engine.ts
JeremyCloarec: I changed it to type it as ComplexAttribute instead, since AttributeDefinition doesn't always contain a format

Archidoit: nice ! perfect

opencti-platform/opencti-graphql/src/database/engine.ts
Archidoit: okay I understand

Issue #7066, opened 2024-05-22
Implement some Decay Exclusion Rules


Kradma:
## Use case

Store detection rules (Threat Hunting related) without an expiration date. These rules are stored as "sigma" objects under the Indicators section.

## Current Workaround

Currently, all indicators are subject to decay models which enforce a lifecycle on the IoCs. For threat hunting rules like sigma and Yara, this is not appropriate since they are designed to be permanent. The lack of filtering options for these types of indicators results in their unintended expiration.

## Proposed Solution

- Add a new rule type called "Decay Exclusion Rule"
- Enable the creation of mutliple exclusion rules
- Enable the same filters than on Decay rules (Author, Creator, Pattern type, indicator type, valid from (empty), valid until (emtpy), label, marking)
- Ensure that if an ioc matches both an exclusion rule & a decay rule, IOC will not get decayed.

## If the feature request is approved, would you be willing to submit a PR?

No

romain-filigran:
Hi @Kradma, thanks for your suggestion! You're right—we need to incorporate more filtering capabilities when defining decay rules, especially for creating specific rules for Sigma or YARA indicator patterns. We’re planning to improve this in the coming months.

You can track this issue: https://github.com/OpenCTI-Platform/opencti/issues/7775

nino-filigran:
This feature will be implemented together with the above mentioned issue. I've modified your issue description to make it clear regarding what will be implemented.

richard-julien:
For me, this one should be cover by https://github.com/OpenCTI-Platform/opencti/issues/7775. @nino-filigran @romain-filigran 
Globally its:
- Create a filter -> targetting a rule
- Create a filter -> targetting "no rule"

PR #13168, opened 2025-11-14
[frontend/backend] Decay Exclusion Rule (#7066)

ValentinBouzinFiligran:
### Proposed changes

* new Decay Exclusion Rule backend module

### Related issues

* #7066

### Checklist


opencti-platform/opencti-graphql/src/modules/indicator/indicator-domain.ts
```diff
@@ -272,8 +273,13 @@ export const addIndicator = async (context: AuthContext, user: AuthUser, indicat
   checkScore(indicatorBaseScore);
 
   const isDecayActivated: boolean = await isDecayEnabled();
+
+  const activeDecayExclusionRuleList = await getActiveDecayExclusionRule(context, user);
```
aHenryJard: Not sure but I think that exclusion are enabled only if decay is enabled. Maybe to be checked with product but I don't see what we can do on indicator excluded when decay is not enabled.

opencti-platform/opencti-graphql/src/modules/indicator/indicator-domain.ts
```diff
@@ -287,7 +293,17 @@ export const addIndicator = async (context: AuthContext, user: AuthUser, indicat
     R.assoc('revoked', revoked),
   )(indicator);
   let finalIndicatorToCreate;
-  if (isDecayActivated && !revoked) {
+  if (!revoked && hasExclusionRuleMatching && exclusionRule) {
```
aHenryJard: another good question for product : do we also check exclusions on revoked indicator. We have seen some time indicator being unrevoked.

opencti-platform/opencti-graphql/src/modules/decayRule/exclusions/decayExclusionRule-domain.ts
```diff
@@ -0,0 +1,108 @@
...
+};
+
+type CheckDecayExclusionRulesTypes = {
+  hasExclusionRuleMatching: boolean,
+  exclusionRule: DecayExclusionRuleModel | null,
+};
+
+export const checkDecayExclusionRules = async (
+  context: AuthContext,
+  user: AuthUser,
```
aHenryJard, line 45: it seems that you don't need context & user here, I don't think you need async either

opencti-platform/opencti-graphql/tests/01-unit/domain/decayExclusionRule-test.ts
```diff
@@ -0,0 +1,76 @@
...
+        expect(hasExclusionRuleMatching).toBe(true);
+      });
+
+      it('should have exclusionRule id to be test1 id ', () => {
+        expect(exclusionRule?.id).toBe('test1 id');
+      });
+    });
+
+    describe('If there is only an inactive rule on the list', async () => {
+      const { exclusionRule, hasExclusionRuleMatching, } = await checkDecayExclusionRules(testContext, ADMIN_USER, observableType, [decayExclusionRuleInactiveModel]);
```
aHenryJard: not a blocker, but it's a bit anoying to not have test data inside the test, I keep scrolling up and down to understand. 

I propose to move them inside tests.

```suggestion
      const { exclusionRule, hasExclusionRuleMatching, } = await checkDecayExclusionRules(testContext, ADMIN_USER, 'IPV4', [{
  id: 'test2 id',
  name: 'test2 name',
  description: 'test2 description',
  created_at: new Date(),
  decay_exclusion_observable_types: ['IPV6', 'Report', 'test'],
  active: false,
}]);
```

opencti-platform/opencti-graphql/src/modules/decayRule/exclusions/decayExclusionRule-domain.ts
```diff
@@ -0,0 +1,110 @@
...
+};
+export const findDecayExclusionRulePaginated = (context: AuthContext, user: AuthUser, args: QueryDecayExclusionRulesArgs) => {
+  if (!isDecayExclusionRuleEnabled) throw UnsupportedError('Feature not yet available');
+  return pageEntitiesConnection<BasicStoreEntityDecayExclusionRule>(context, user, [ENTITY_TYPE_DECAY_EXCLUSION_RULE], args);
+};
+
+export const getActiveDecayExclusionRule = async (context: AuthContext, user: AuthUser) => {
+  if (!isDecayExclusionRuleEnabled) return [];
+  const decayExclusionRuleEdges = await findDecayExclusionRulePaginated(context, user, {});
+  return decayExclusionRuleEdges.edges.map(({ node }) => node).filter((rule) => rule.active);
```
JeremyCloarec: Is moving the decay exclusion rules to the cache planned for a later chunk?

JeremyCloarec: Also, right now this doesn't work if there are more rules than the default page size, since you are using pageEntitiesConnection. You can use fullEntitiesList instead to get all the possible decay exclusion rules

opencti-platform/opencti-graphql/src/modules/decayRule/exclusions/decayExclusionRule-domain.ts
```diff
@@ -0,0 +1,110 @@
...
+};
+
+export const checkDecayExclusionRules = async (
+  context: AuthContext,
+  user: AuthUser,
+  resolvedIndicator: ResolvedDecayExclusionRule,
+  activeDecayExclusionRuleList: DecayExclusionRuleModel[]
```
JeremyCloarec: You can use the StoreEntityDecayExclusionRule type instead of DecayExclusionRuleModel

opencti-platform/opencti-graphql/src/modules/decayRule/exclusions/decayExclusionRule-domain.ts
```diff
@@ -0,0 +1,110 @@
...
+): Promise<DecayExclusionRuleModel | null> => {
+  if (!isDecayExclusionRuleEnabled) return null;
+
+  const formattedIndicator = {
+    ...resolvedIndicator,
+    object_marking_refs: (resolvedIndicator[INPUT_MARKINGS] ?? []).map((marking: MarkingDefinition) => marking.standard_id),
+    created_by_ref: resolvedIndicator[INPUT_CREATED_BY]?.standard_id ?? '',
+    labels: (resolvedIndicator[INPUT_LABELS] ?? []).map((label: Label) => label.id),
```
JeremyCloarec: Shouldn't we get the internal_id on all of those instead? I think the decay_exclusion_filters filters contain an internal_id on all values no?

opencti-platform/opencti-graphql/src/modules/decayRule/exclusions/decayExclusionRule.graphql
```diff
@@ -0,0 +1,54 @@
...
+  # DecayExclusionRule
+  name: String!
+  description: String
+  decay_exclusion_filters: String!
+  active: Boolean!
+}
+
+enum DecayExclusionRuleOrdering {
+  name
+  order
```
JeremyCloarec: I think this order orderingi's a leftover from DecayRules

opencti-platform/opencti-graphql/src/modules/decayRule/exclusions/decayExclusionRule.graphql
```diff
@@ -0,0 +1,54 @@
...
+    filters: FilterGroup
+    search: String
+  ): DecayExclusionRuleConnection @auth(for: [SETTINGS_SETCUSTOMIZATION])
+}
+
+input DecayExclusionRuleAddInput {
+  name: String! @constraint(minLength: 2, format: "not-blank")
+  description: String
+  decay_exclusion_filters: String!
+  active: Boolean!
```
JeremyCloarec, line 47: It's a minor detail, but couldn't these inputs be marked as non required ? The backend could default those to empty string for the filter and inactive of the active flag

opencti-platform/opencti-graphql/src/modules/decayRule/exclusions/decayExclusionRule.ts
```diff
@@ -0,0 +1,37 @@
...
+  },
+  identifier: {
+    definition: {
+      [ENTITY_TYPE_DECAY_EXCLUSION_RULE]: () => uuidv4()
+    },
+  },
+  attributes: [
+    { name: 'name', label: 'Name', type: 'string', format: 'short', mandatoryType: 'internal', editDefault: false, multiple: false, upsert: false, isFilterable: true },
+    { name: 'description', label: 'Description', type: 'string', format: 'text', mandatoryType: 'customizable', editDefault: false, multiple: false, upsert: false, isFilterable: true },
+    { name: 'active', label: 'Status', type: 'boolean', mandatoryType: 'no', editDefault: false, multiple: false, upsert: false, isFilterable: true },
```
JeremyCloarec, line 23: active is tagged as non mandatory here whereas it is marked as mandatory on GraphQL. This shuld be changed to be mandatory = true

opencti-platform/opencti-graphql/src/modules/decayRule/exclusions/decayExclusionRule.ts
```diff
@@ -0,0 +1,37 @@
...
+    { name: 'decay_exclusion_filters', label: 'Indicator filters', type: 'string', format: 'text', mandatoryType: 'no', editDefault: false, multiple: false, upsert: false, isFilterable: true },
```
JeremyCloarec: Same as above. Also, isFilterable should be set to false

opencti-platform/opencti-graphql/src/modules/indicator/indicator-domain.ts
```diff
@@ -272,22 +276,46 @@ export const addIndicator = async (context: AuthContext, user: AuthUser, indicat
   checkScore(indicatorBaseScore);
 
   const isDecayActivated: boolean = await isDecayEnabled();
+
+  let exclusionRule = null;
+  if (isDecayExclusionRuleEnabled) {
+    const activeDecayExclusionRuleList = await getActiveDecayExclusionRule(context, user);
+    const entitySetting = await getEntitySettingFromCache(context, ENTITY_TYPE_INDICATOR);
+    const resolvedIndicator = await inputResolveRefs(context, user, indicator, ENTITY_TYPE_INDICATOR, entitySetting);
```
JeremyCloarec: resolvedIndicator should be reused afterwards to prevent having to resolve twice the inputs.
To do this, indicatorToCreate should use resolvedIndicator instead of indicator

opencti-platform/opencti-graphql/src/modules/indicator/indicator-domain.ts
```diff
@@ -272,22 +276,46 @@ export const addIndicator = async (context: AuthContext, user: AuthUser, indicat
...
+
+  if (isDecayActivated && exclusionRule && isDecayExclusionRuleEnabled) {
+    finalIndicatorToCreate = {
+      ...indicatorToCreate,
+      decay_exclusion_applied_rule: {
+        decay_exclusion_id: exclusionRule.id,
+        decay_exclusion_name: exclusionRule.name,
+        decay_exclusion_created_at: exclusionRule.created_at,
+        decay_exclusion_filters: exclusionRule.decay_exclusion_filters,
+      }
```
JeremyCloarec, line 317: What's the reasoning behind storing the whole decay rule in a flat object on ES? I don't see it used right now on the frontend side, is it going to be used in a later chunk to display the rule explaining the decay exclusion? 
Also, why not store the decay exclusion rule id instead and resolve it dynamically?

opencti-platform/opencti-graphql/src/modules/decayRule/exclusions/decayExclusionRule.graphql
```diff
@@ -0,0 +1,53 @@
...
+  # DecayExclusionRule
+  name: String!
+  description: String
+  decay_exclusion_filters: String!
+  active: Boolean
+}
+
+enum DecayExclusionRuleOrdering {
+  name
+  active
```
JeremyCloarec, line 17: You need to keep the _score ordering since decayExclusionRules accept a search

JeremyCloarec (approved):

PR #13204, opened 2025-11-17
build: add oxlint

delemaf:
### Proposed changes

* Add Oxlint alongside ESlint
* Fix file with eslint tags

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/13355

### Checklist


opencti-platform/opencti-front/.oxlintrc.json.bak
```diff
@@ -0,0 +1,26 @@
...
+    "no-unused-vars": [
+      "error",
+      {
+        "argsIgnorePattern": "^_",
+        "varsIgnorePattern": "^_",
+        "caughtErrorsIgnorePattern": "^_"
+      }
+    ]
+  }
+}
```
OctaveLaventure, line 26: **1. Cleanup**
do we need to keep this file?

opencti-platform/opencti-front/.oxlintrc.json
```diff
@@ -0,0 +1,267 @@
...
+      },
+      "env": {
+        "browser": true,
+        "jest": true
+      }
+    },
+    {
+      "files": ["*.json", "*.mjs"],
+      "rules": {
+        "sort-keys": "off",
```
OctaveLaventure, line 243: **2. Config**
I think we can disable it everywhere

OctaveLaventure (approved):
Amazing! Works well on my end too
We should add documentation regarding the reasons behind the change (and why we chose it over other options such as Biome) and the strategy to deal with the large amount of silent warnings going forward, and link it in the PR and maybe as a comment in the oxlint configs for future reference

PR #13205, opened 2025-11-17
[frontend] Refactor some playbooks components (#12362)

lndrtrbn:
### Proposed changes

* Refactor how playbook components are combined with each others
* Preparatory work for refactoring component PlaybookAddComponents.jsx later

### Related issues

* #12362

### Checklist


opencti-platform/opencti-front/src/private/components/data/playbooks/playbookFlow/PlaybookFlow.tsx
```diff
@@ -0,0 +1,138 @@
+/*
```
esrevi, line 1: for my understanding, is it required to put that license header in the files that will be used in EE? switching to 2026, do we need to update the date for all files with that header?

opencti-platform/opencti-front/src/private/components/data/playbooks/playbookFlow/PlaybookFlowSelectComponent.tsx
```diff
@@ -0,0 +1,72 @@
...
+  const { t_i18n } = useFormatter();
+
+  const isSelectedNodeEntryPoint = selectedNode?.data?.component?.is_entry_point ?? false;
+  const entryComponents = components.flatMap((component) => {
+    if (!component || component.is_entry_point !== isSelectedNodeEntryPoint) return [];
+    return component;
+  });
+
+  return (
+    <div>
```
esrevi: is this div necessary?

opencti-platform/opencti-front/src/private/components/data/playbooks/hooks/useManipulateComponents.jsx
MTorbay-Filigran, line 1: Next improvement would be to type this hook and add some tests but maybe in another PR

MTorbay-Filigran (approved):

opencti-platform/opencti-front/src/private/components/data/playbooks/playbookFlow/PlaybookFlow.tsx
```diff
@@ -0,0 +1,138 @@
...
+
+interface PlaybookFlowProps {
+  dataPlaybook: PlaybookFlow_playbook$key;
+  dataPlaybookComponents: PlaybookFlow_playbookComponents$key;
+}
+
+const PlaybookFlow = ({ dataPlaybookComponents, dataPlaybook }: PlaybookFlowProps) => {
+  const playbook = useFragment(playbookFragment, dataPlaybook);
+  const { playbookComponents } = useFragment(playbookComponentsFragment, dataPlaybookComponents);
+  const definition = JSON.parse(playbook.playbook_definition ?? '{}');
```
esrevi: I think there could be some crash if for some reason, playbook.playbook_definition is malformed, imo, it could be good to wrap with try/catch no ?

Also if playbook.playbook_definition is empty string, JSON.parse will throw error, is it intended ? otherwise change '??' with '||' to handle empty string too

opencti-platform/opencti-front/src/private/components/data/playbooks/playbookFlow/PlaybookFlow.tsx
```diff
@@ -0,0 +1,138 @@
...
+    setSelectedNode,
+    selectedEdge,
+    setSelectedEdge,
+    onConfigAdd,
+    onConfigReplace,
+    deleteNode,
+  } = useManipulateComponents(playbook);
+
+  const initialNodes = computeNodes(
+    definition.nodes,
```
esrevi: if defintion fails, does computeNodes handles definition.nodes undefined? or should we ensure with definition?.nodes ?? [] ?

opencti-platform/opencti-front/src/private/components/data/playbooks/playbookFlow/PlaybookFlow.tsx
lndrtrbn: yes we will

opencti-platform/opencti-front/src/private/components/data/playbooks/playbookFlow/PlaybookFlow.tsx
lndrtrbn: It's copy paste from old code so we should be pretty good with crashes but I agree with you that parsing strings can be a bit shaky, we do that in multiple features, for example dashboard configurations.

For the try/catch, I think the approach right now is to let the page crashes because if the playbook_definition is malformed it means we have a bug and need to be fixed.

I'll make the change from ?? to ||

opencti-platform/opencti-front/src/private/components/data/playbooks/playbookFlow/PlaybookFlowSelectComponent.tsx
lndrtrbn: Maybe not, it's copy pasted code, I'll test it

opencti-platform/opencti-front/src/private/components/data/playbooks/playbookFlow/PlaybookFlow.tsx
lndrtrbn: It crashes, but right now we had no issues with this code (again copy pasted).
This will be managed when the file containing computeNodes will be moved to TS, in other PR

PR #13224, opened 2025-11-18
[backend] Fix notifier issue with special characters (#13216)

### LINKED ISSUE #13216: Notifer issue with special character
## Description

If a csutomer notifier is supposed to send a notification that contains the description of an entity and the description contains a "/", the notification is not sent.
## Environment

Testing

## Reproducible Steps

1.Create the following custom notifier 
{
"title": "<%= data[0].instance.name %>",
"description": "<%= data[0].instance.description%>",
"id":"<%= notification_content[0].events[0].instance_id %>",
}
2. trigger a notification for an entity that has a "/" in its description 

## Expected Output

notification received
## Actual Output

notification not received
## Additional information

After investigation, it seems that special characters are not well supported. The fix should consist in using JSON.stringify rather than doing escaping.

maelv-filigran:
### Proposed changes

* Replace custom string escaping logic with JSON.stringify in webhook notifications
* Use EJS escape option to properly handle special characters including forward slashes
* Remove unnecessary try-catch wrapper in internalProcessNotification
* Add test case for forward slash handling

### Related issues
* #13216 

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/manager/publisherManager.ts
+++ b/opencti-platform/opencti-graphql/src/manager/publisherManager.ts
@@ -202,28 +202,15 @@ export async function handleSimplifiedEmailNotification(
 }
 
 export async function handleWebhookNotification(configurationString: string | undefined, templateData: object) {
-  function escapeLineBreak(str: string):string {
-    return str
-      .replace(/\n/g, '\\n');
-  }
-
-  function escapeStringValuesForJSON(obj: any): any {
-    if (typeof obj === 'string') {
-      return escapeLineBreak(obj);
-    } if (Array.isArray(obj)) {
-      return obj.map(escapeStringValuesForJSON);
-    } if (typeof obj === 'object' && obj !== null) {
-      return Object.fromEntries(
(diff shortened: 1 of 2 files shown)


JeremyCloarec (approved):

opencti-platform/opencti-graphql/src/manager/publisherManager.ts
```diff
@@ -251,41 +238,37 @@ export const internalProcessNotification = async (
...
-        break;
-      case NOTIFIER_CONNECTOR_WEBHOOK:
-        await handleWebhookNotification(notifierConfigurationString, assembledTemplateData);
-        break;
-      default:
-        // TODO: Handle other notifier scenarios
-        break;
-    }
-  } catch (error) {
-    return { error: (error as Error).message };
```
xfournet, line 288: this has not been reported to testNotifier in opencti-platform/opencti-graphql/src/modules/notifier/notifier-domain.ts which was the only caller to rely on return { error }. so it need to be catch in testNotifier now.
I would also removed the // eslint-disable-next-line consistent-return  and explicitly return undefined when everything is ok

PR #13225, opened 2025-11-18
[backend] Introduce option to restrict the capacity to change the platform sender email (#13214)

### LINKED ISSUE #13214: [backend] Introduce option to force platform sender email
## Use case

In a normal case, the from of an email should be "User \<user[at]email.com\>" where the email part must be strictly equals to the account email configured to connect to the smtp system.

Its authorized to change it can be detected as a bad behavior.
In some circonstance it can be changed but we need a way to force it by configuration.

## Current Workaround

No

## Proposed Solution

Introduce a new configuration to force a sender email

```
  "smtp": {
    "forced_sender_email": "user[at]email.com"
  }
```

In this mode, the setting `Sender email address` in the UI will be disabled and forced to the smtp configuration, and so every mail that will be sent by the platform will only used the `smtp:forced_sender_email`

richard-julien:
See #13214

### DIFF
--- a/opencti-platform/opencti-graphql/config/default.json
+++ b/opencti-platform/opencti-graphql/config/default.json
@@ -414,6 +414,7 @@
     "queue_type": "classic"
   },
   "smtp": {
+    "allow_email_rewrite": false,
     "hostname": "localhost",
     "use_ssl": false,
     "reject_unauthorized": false,
--- a/opencti-platform/opencti-graphql/config/default.json
+++ b/opencti-platform/opencti-graphql/config/default.json
@@ -414,13 +414,13 @@
     "queue_type": "classic"
   },
   "smtp": {
-    "allow_email_rewrite": false,
     "hostname": "localhost",
     "use_ssl": false,
     "reject_unauthorized": false,
     "port": 25,
     "username": "",
     "password": "",
+    "forced_sender_email": "",
     "email_max_cc_size": 500
   },
   "providers": {
--- a/opencti-platform/opencti-graphql/config/test.json
+++ b/opencti-platform/opencti-graphql/config/test.json
@@ -109,6 +109,7 @@
     }
   },
   "smtp": {
+    "allow_email_rewrite": true,
     "enabled": false
   }
 }
--- a/opencti-platform/opencti-graphql/config/test.json
+++ b/opencti-platform/opencti-graphql/config/test.json
@@ -109,7 +109,6 @@
     }
   },
   "smtp": {
-    "allow_email_rewrite": true,
     "enabled": false
   }
 }
--- a/opencti-platform/opencti-graphql/src/database/smtp.js
+++ b/opencti-platform/opencti-graphql/src/database/smtp.js
@@ -1,7 +1,11 @@
 import nodemailer from 'nodemailer';
 import conf, { booleanConf, logApp } from '../config/conf';
 import { meterManager } from '../config/tracing';
+import { getEntityFromCache } from './cache';
+import { ENTITY_TYPE_SETTINGS } from '../schema/internalObject';
+import { executionContext, SYSTEM_USER } from '../utils/access';
 
+export const ALLOW_EMAIL_REWRITE = booleanConf('smtp:allow_email_rewrite', false);
 const USE_SSL = booleanConf('smtp:use_ssl', false);
 const REJECT_UNAUTHORIZED = booleanConf('smtp:reject_unauthorized', false);
 const SMTP_ENABLE = booleanConf('smtp:enabled', true);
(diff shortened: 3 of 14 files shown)


opencti-platform/opencti-graphql/config/default.json
```diff
@@ -414,6 +414,7 @@
     "queue_type": "classic"
   },
   "smtp": {
+    "allow_email_rewrite": false,
```
xfournet: conf key need to be more clear. I propose  allow_from_email_override

opencti-platform/opencti-graphql/src/manager/publisherManager.ts
```diff
@@ -184,7 +183,7 @@ export async function handleSimplifiedEmailNotification(
   const renderedEmail = ejs.render(SIMPLIFIED_EMAIL_TEMPLATE, finalTemplateData);
 
   const emailPayload = {
-    from: `${settings.platform_title} <${settings.platform_email}>`,
+    from: await smtpComputeFrom(settings.platform_title),
```
xfournet: since settings.platform_title is the default, it could be removed and so setting is no more needed in this method

opencti-platform/opencti-graphql/src/modules/notifier/notifier-domain.ts
```diff
@@ -200,6 +200,7 @@ export const testNotifier = async (context: AuthContext, user: AuthUser, notifie
     MOCK_NOTIFICATIONS[notifier.notifier_test_id],
     [{ created: (new Date()).toISOString() }] as unknown as BasicStoreEntityTrigger[],
     usersFromCache,
-  );
-  return result?.error;
+  ).catch((error) => {
```
xfournet, line 203: it's fixing  https://github.com/OpenCTI-Platform/opencti/pull/13224/files#r2543859135 ;)

opencti-platform/opencti-graphql/src/database/smtp.js
```diff
@@ -27,6 +31,18 @@ if (conf.get('smtp:username') && conf.get('smtp:username').length > 0) {
 
 export const transporter = nodemailer.createTransport(smtpOptions);
 
+export const smtpConfiguredEmail = (settings) => {
+  return ALLOW_EMAIL_REWRITE ? settings.platform_email : conf.get('smtp:username');
```
xfournet: SMTP username is not always an email address (eg SMTP server connected on LDAP)
We should add a new conf to specify the 'forced' email address, that could replace the allow_email_rewrite.
eg smtp.enforced_from_email : if set we use it else the setting 

->
```
export const smtpConfiguredEmail = (settings) =>conf.get('smtp.enforced_from_email') ?? settings.platform_email;
```

opencti-platform/opencti-graphql/config/test.json
```diff
@@ -109,6 +109,7 @@
     }
   },
   "smtp": {
+    "allow_email_rewrite": true,
```
xfournet: no more used, to be removed

opencti-platform/opencti-graphql/src/database/smtp.js
```diff
@@ -27,6 +33,18 @@ if (conf.get('smtp:username') && conf.get('smtp:username').length > 0) {
 
 export const transporter = nodemailer.createTransport(smtpOptions);
 
+export const smtpConfiguredEmail = (settings) => {
+  return isEmptyField(SMTP_FORCED_EMAIL) ? settings.platform_email : SMTP_FORCED_EMAIL;
```
xfournet: can use ALLOW_EMAIL_REWRITE  that is already equals to `isEmptyField(SMTP_FORCED_EMAIL)`

opencti-platform/opencti-graphql/config/test.json
richard-julien: done

opencti-platform/opencti-graphql/src/database/smtp.js
richard-julien: done

opencti-platform/opencti-graphql/src/database/smtp.js
richard-julien: approach changed / issue rewritten

opencti-platform/opencti-graphql/src/manager/publisherManager.ts
richard-julien: done

xfournet: settings are no more used, you can remove the settings from the method parameters

xfournet (approved):

opencti-platform/opencti-graphql/src/manager/publisherManager.ts
richard-julien: done

Issue #11073, opened 2025-05-22
Duplicate reliability in organization


alice-debra:
Keep the right one
<img width="1141" alt="Image" src="https://github.com/user-attachments/assets/89dfa6f3-5973-495e-8fb6-51e338e8c722" />

romain-filigran:
@alice-debra: Maybe we should talk about it before to do something. I'm not sure it's a bug, as the two “reliability” fields may represent different things in this case.

alice-debra:
Both must have a different key on the right, this is the reliability of the author who created this entity.

Archidoit:
Solution: Same display as for reports 
- reliability displayed in the right ('basic information'): reliability if there is one, reliability of author if no reliability
- no reliability displayed in the left ('details')

PR #13228, opened 2025-11-19
[frontend] Remove Reliability in Organization details (#11073)

Archidoit:
### Proposed changes
- Remove 'reliability' in organization details, since it's already displayed in 'basic information'

image

### Related issues
#11073

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/entities/organizations/OrganizationDetails.tsx
+++ b/opencti-platform/opencti-front/src/private/components/entities/organizations/OrganizationDetails.tsx
@@ -1,24 +1,42 @@
 import React, { FunctionComponent } from 'react';
-import { createFragmentContainer, graphql } from 'react-relay';
+import { graphql, useFragment } from 'react-relay';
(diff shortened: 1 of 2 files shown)


opencti-platform/opencti-front/src/private/components/entities/organizations/OrganizationDetails.tsx
```diff
@@ -1,59 +1,72 @@
...
+      id
+      value
+      color
+    }
+  }
+`;
 
 interface OrganizationDetailsComponentProps {
-  organization: OrganizationDetails_organization$data;
+  organizationData: Organization_organization$data;
```
lndrtrbn: You can replace the type by `OrganizationDetails_organization$key` directly. And by so you would be able to remove the type you add on `useFragment`, it should be found automatically

Archidoit: thanks! it's done :)

lndrtrbn (approved):

PR #13237, opened 2025-11-19
[client] fix usage of mapping_cache in draft (#13226)

### LINKED ISSUE #13226: CSV import in Draft: import of label is inconsitent
## Description

Importing data with CSV import & create a draft highlight the fact that labels are not consistlenty applied to entities. 

## Environment

Pre-rleease (and prod, bug intiated by user)

## Reproducible Steps

1. Import following csv mapper: [CSV-Mapper.json](https://github.com/user-attachments/files/23622733/CSV-Mapper.json)
2. Click on step by step import and import this file & create a draft
3. Go in the draft
4. Look at the containers label
5. DO NOT Validate the draft
6. Reiterate the operation  (reimport the file in a new drat)
7. Go in the newly create draft
8. Notice the difference of labels on containers with the preivous draft created

## Expected Output

Label application should be consistent. 
## Actual Output

Label application on entities is inconsistent and should not be erratic. 

## Additional information

the issue does not happen on workbenches: the label is consistently applied on various workbenches, even when ou import multiple times the same file.
it only happens on Drafts: the first attempt often contains the right data, but the second import of the same file (without validating the initial draft) does not have all the labels... So  I guess the issue lies in the drafts & not in the parsing of the CSV. 

I have the file, please contact me to get it

JeremyCloarec:
### Proposed changes

* mapping_cache was being used independently of draft context in opencti_stix2. This means that it could try to use ids of label/entities created in an other draft, that have a different internal_id. 
* we now suffix all items inserted in mapping_cache with current draft_id
*

### Related issues

* #13226 
*

### Checklist

### DIFF
--- a/client-python/pycti/api/opencti_api_client.py
+++ b/client-python/pycti/api/opencti_api_client.py
@@ -258,6 +258,9 @@ class OpenCTIApiClient:
         self.user = User(self)
         self.settings = Settings(self)
 
+        # Keep track of draft context
+        self.draft_id = ""
+
         # Check if openCTI is available
         if perform_health_check and not self.health_check():
             raise ValueError(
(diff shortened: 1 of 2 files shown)


OctaveLaventure (approved):
LGTM but didn't test locally, we can peer review if you want. Approving for fast merge if needed.

PR #13247, opened 2025-11-20
[client] refactor recursive import retries to iterative (#13253)

### LINKED ISSUE #13253: Handle worker maximum recursion depth exceeded
## Use case

In some cases, worker can fail with maximum recursion depth exceeded.
Process retry without recursion:
* Transform long tail recursive to iterative in Python Client
* Check how max number of try are ensured (should have been blocked at 100 with MAX_PROCESSING_COUNT)
* NB: should we have a look to the unorderer processing of bundle split with rejection & retry in the client

JeremyCloarec:
### Proposed changes

* long tail recursive retries could lead to maximum recursion depth exceeded errors. We now retry iteratively instead
*

### Related issues

* #13253
*

### Checklist

### DIFF
--- a/client-python/pycti/utils/opencti_stix2.py
+++ b/client-python/pycti/utils/opencti_stix2.py
@@ -2861,9 +2861,7 @@ class OpenCTIStix2:
                 if "observed_data_refs" in item:
                     for observed_data_ref in item["observed_data_refs"]:
                         for to_id in to_ids:
-                            self.import_sighting(
-                                item, observed_data_ref, to_id, update
-                            )
+                            self.import_sighting(item, observed_data_ref, to_id, update)
             # endregion
         elif item["type"] == "label":
             stix_ids = self.opencti.get_attribute_in_extension("stix_ids", item)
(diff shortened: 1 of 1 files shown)


xfournet:
In recursive version, timeout case was missing a log. Could be added at the same time of this refactor ?

client-python/pycti/utils/opencti_stix2.py
```diff
@@ -2823,242 +2823,228 @@ def import_item(
...
+    def import_item_with_retries(
+        self,
+        item,
+        update: bool = False,
+        types: List = None,
+        work_id: str = None,
+    ):
+        processing_count = 0
+        worker_logger = self.opencti.logger_class("worker")
+        while processing_count <= MAX_PROCESSING_COUNT:
```
xfournet, line 2958: look like if processing_count > MAX_PROCESSING_COUNT then the message will be silently discarded ?

JeremyCloarec: Yes I kept the behavior the same as before, but I can change it to keep retrying indefinitely instead if we want it

xfournet (approved):

PR #13250, opened 2025-11-20
[backend/frontend] Adapt to introduce LTS validation in LTS platform

### LINKED ISSUE #13575: [backend/frontend] Introduce LTS validation in LTS platform
## Use case

Preparation / Introduce LTS validation in LTS platform.
This work aims to facilitate the upcoming LTS version of OpenCTI

richard-julien:
# Definition
This PR change the behavior of a platform where the version end by -lts.
In this mode, a LTS license must be provided to have any usage of the platform.
After loggin, the only screen available will be the license setup.
If the license is LTS, it will be accepted, if not rejected.

If the license is setup by config or env and NOT a LTS, the UI will be blocked and a specific message to change the license will be provided.

# How to test
Change the version in your package.json to prefix the version with -lts
For example from 6.8.13 to 6.8.13-lts
Starts your platform.

Ask a test LTS license to @richard-julien


opencti-platform/opencti-graphql/config/schema/opencti.graphql
```diff
@@ -1461,6 +1461,7 @@ type Settings implements InternalObject & BasicObject & ThemeSettings & IntlSett
   platform_title: String
   platform_favicon: String
   platform_email: String
+  platform_type: String!
```
aHenryJard: Is it possible to use an enum instead of a String for the platform type ?

opencti-platform/opencti-graphql/src/resolvers/settings.js
```diff
@@ -38,6 +39,7 @@ const settingsResolvers = {
     relationships: (_, __, context) => elAggregationCount(context, context.user, READ_DATA_INDICES, { types: ['stix-relationship'], field: 'entity_type' }),
   },
   Settings: {
+    platform_type: () => (PLATFORM_VERSION.endsWith('lts') ? 'lts' : 'standard'),
```
aHenryJard: so trial and nfl (or i don't remember the name for Charity licenses) are all standard ? Or it is possible to have trial on lts too ?

richard-julien: Its not the type of the license but the type of the platform, so only LTS or STANDARD

opencti-platform/opencti-graphql/config/schema/opencti.graphql
richard-julien: Fixed

opencti-platform/opencti-graphql/src/http/httpAuthenticatedContext.js
```diff
@@ -60,6 +61,8 @@ export const createAuthenticatedContext = async (req, res, contextName) => {
         executeContext.user_otp_validated = req.session?.user.otp_validated ?? false;
       }
       executeContext.user_inside_platform_organization = isUserInPlatformOrganization(user, settings);
+      const licenseInfo = getEnterpriseEditionInfo(settings);
+      executeContext.blocked_for_lts_validation = PLATFORM_VERSION.endsWith('lts') && !licenseInfo.license_validated;
```
xfournet: there is many place where we are doing PLATFORM_VERSION.endsWith('lts') -> to be replaced by a isLTSVersion in config/conf.js so we can adapt this test if needed

opencti-platform/opencti-graphql/src/graphql/authDirective.js
```diff
@@ -59,6 +60,10 @@ export const authDirectiveBuilder = (directiveName) => {
                   throw OtpRequired();
                 }
               }
+              // LTS version must be validated
+              if (info.fieldName !== 'setupEnterpriseLicense' && info.fieldName !== 'settings' && blocked_for_lts_validation) {
```
xfournet: isn't possible to add this 'allowUnlicencedLTS' info in the @auth tag instead of checking fieldName?

opencti-platform/opencti-graphql/src/http/httpAuthenticatedContext.js
richard-julien: done

opencti-platform/opencti-graphql/src/graphql/authDirective.js
richard-julien: I try a approach with an annotation that can be added with the @auth.
Tell me what you think. If you think its great I will open a new PR to replace the field check with a new annotation allowUnprotectedOTP for the OTP part.

xfournet: looks good! ok to also add allowUnprotectedOTP

xfournet (approved):

opencti-platform/opencti-graphql/src/graphql/authDirective.js
richard-julien: Open in https://github.com/OpenCTI-Platform/opencti/pull/13587

opencti-platform/opencti-graphql/src/http/httpAuthenticatedContext.js
```diff
@@ -60,6 +61,8 @@ export const createAuthenticatedContext = async (req, res, contextName) => {
         executeContext.user_otp_validated = req.session?.user.otp_validated ?? false;
       }
       executeContext.user_inside_platform_organization = isUserInPlatformOrganization(user, settings);
+      const licenseInfo = getEnterpriseEditionInfo(settings);
```
xfournet, line 64: @richard-julien 
now it's deployed, it's visible on pyroscope, and should be changed to use a cache

xfournet: see #13824

PR #13257, opened 2025-11-20
[backend] add stream type managers state tracking & state recovery (#13254)

### LINKED ISSUE #13254: OpenCTI manager live listening after restart lead to missing events
## Use case

When the platform restart where is a time between the availability of the API and the manager starts.
If the platform restart is not done gracefully it could be longer than in normal situation.

API is available before the stream is consumed and so we have activity that happen before the "real" listening of the stream. Manager that can be impacted:

    NotificationManager
    PublisherManager
    PlaybookManager

Maybe we always need a state to ensure that the listening starts at the right level.

JeremyCloarec:
### Proposed changes

* stream type managers could potentially miss some events, either at startup or in case of a node failure. To prevent this, we now keep track of the state of the stream events handling by the managers, so that this state can be reused at manager startup to continue where the last run left off
*

### Related issues

* #13254 
*

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/manager/fileIndexManager.ts
+++ b/opencti-platform/opencti-graphql/src/manager/fileIndexManager.ts
@@ -21,6 +21,7 @@ import { EVENT_TYPE_UPDATE, isEmptyField, waitInSec } from '../database/utils';
 import conf, { ENABLED_FILE_INDEX_MANAGER, logApp } from '../config/conf';
 import { createStreamProcessor } from '../database/stream/stream-handler';
 import { type StreamProcessor } from '../database/stream/stream-utils';
+import { redisGetManagerEventState, redisSetManagerEventState } from '../database/redis';
 import { lockResources } from '../lock/master-lock';
 import { executionContext, SYSTEM_USER } from '../utils/access';
 import { getEntityFromCache } from '../database/cache';
@@ -42,6 +43,7 @@ const FILE_INDEX_MANAGER_KEY = conf.get('file_index_manager:lock_key');
 const SCHEDULE_TIME = conf.get('file_index_manager:interval') || 60000; // 1 minute
 const STREAM_SCHEDULE_TIME = 10000;
 const FILE_INDEX_MANAGER_STREAM_KEY = conf.get('file_index_manager:stream_lock_key');
+const FILE_INDEX_MANAGER_NAME = 'file_index_manager';
 
 interface FileToIndexObject {
   id: string;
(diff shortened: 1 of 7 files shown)


JeremyCloarec:
A reminder: I noticed that the rule manager is currently saving it's state on elastic instead of redis. We could change this behavior in this PR also

opencti-platform/opencti-graphql/tests/utils/rule-utils.js
```diff
@@ -73,10 +74,10 @@ export const changeRule = async (ruleId, active) => {
   let stableCount = 1;
   while (stableCount < 3) {
     const innerInfo = await fetchStreamInfo();
-    const ruleManager = await internalLoadById(testContext, SYSTEM_USER, 'rule_engine_settings');
+    const ruleManagerInfo = await getManagerInfo(testContext, SYSTEM_USER);
     await wait(2000);
     const lastEventDate = new Date(parseInt(innerInfo.lastEventId.split('-').at(0), 10));
-    const managerEventDate = new Date(parseInt(ruleManager.lastEventId.split('-').at(0), 10));
+    const managerEventDate = new Date(parseInt(ruleManagerInfo.lastEventId.split('-').at(0), 10));
```
xfournet: ruleManagerInfo.lastEventId can be undefined

opencti-platform/opencti-graphql/src/manager/fileIndexManager.ts
```diff
@@ -186,7 +187,8 @@ const initFileIndexManager = () => {
         running = true;
         logApp.info('[OPENCTI-MODULE] Running file index manager stream handler');
         streamProcessor = createStreamProcessor(SYSTEM_USER, 'File index manager', handleStreamEvents, { bufferTime: 5000 });
-        await streamProcessor.start('live');
+        const lastEventState = await redisGetManagerEventState(FILE_INDEX_MANAGER_NAME);
+        await streamProcessor.start(lastEventState ?? 'live');
```
xfournet, line 193: since we are now saving the state, for the first startup we don't have value, should we start with '0-0' instead of 'live' ?
but may be this could be a dangerous in case the state is removed
(same for other places with same pattern)

opencti-platform/opencti-graphql/src/manager/ruleManager.ts
```diff
@@ -294,11 +304,13 @@ const initRuleManager = () => {
       running = true;
       const ruleManager = await getInitRuleManager();
       const { lastEventId } = ruleManager;
```
xfournet, line 298: should remove the previous from elastic and put it in redis, so we don't retain old state values in elastic which can be confusing during investigation

opencti-platform/opencti-graphql/src/manager/fileIndexManager.ts
JeremyCloarec: I don't think we can default to '0-0' instead of 'live', this would mean that after the PR is merged, all managers will process events that they have already processed

opencti-platform/opencti-graphql/src/manager/ruleManager.ts
JeremyCloarec: you're right, I'll add the deletion of the value also

xfournet (approved):

JeremyCloarec:
Waiting on https://github.com/OpenCTI-Platform/opencti/pull/13272 to be merged to also add PIR manager state tracking to redis
EDIT: will be done in a separate PR

PR #13272, opened 2025-11-21
[backend] refactor & split concerns for stream

JeremyCloarec:
### Proposed changes

* added a stream client abstraction layer with stream-handler.ts and stream-utils.ts. stream-handler.ts handles the stream buisness logic independently from the stream client layer
* added a stream client interface to formalize stream client contract: all stream clients now have to implement the RawStreamClient interface defined in stream-utils.ts
* moved raw redis stream logic from redis.ts to redis-stream.ts. redis-streams.ts implements a valid RawStreamClient that can be used by stream-handler.ts

### Related issues

*
*

### Checklist


maelv-filigran:
### ~~1. Complete lack of testing for critical refactoring~~

~~__Why this is a blocker__:~~

~~- We are refactoring the __core of the event system__ (creation, update, delete, merge of all entities)~~
~~- No unit tests on the new `RawStreamClient` abstraction~~
~~- No integration tests to verify that the old behavior is preserved~~
~~- Streaming is critical: loss of events = data corruption, inconsistency between systems.~~

~~__Impact__:~~

~~- High risk of silent regression in production~~
~~- Impossible to validate that the `RawStreamClient` contract is respected~~

### ~~2. Potential race condition in `rawCreateStreamProcessor`~~

~~__Problem__:~~

~~- `streamListening` is modified in `shutdown()` while `processStep()` is reading it~~
~~- No mutual exclusion mechanism~~
~~- Concurrent calls to `start()` or `shutdown()` are unprotected~~

### 3.  Silent error handling in `rawCreateStreamProcessor`

__Multiple issues__:

1. __No distinction between recoverable and fatal errors__:

   - Temporary Redis network error → retry OK
   - Parsing error in `streamResult` → fatal, infinite loop
   - Exception in user `callback` → propagated or swallowed?

2. __Unsafe destructuring__:

   ```typescript
   const [, results] = streamResult[0];  // If streamResult is malformed → crash
   ```

3. __`processStreamResult` can throw__ but is handled the same as a Redis error

4. __No circuit breaker__: if Redis is down, we wait 5 seconds indefinitely. Maybe exponential backoff retry ?

### 4. Redis client created/destroyed on each call in `rawFetchStreamEventsRangeFromEventId`

__Problem__:

- Creating a Redis connection is expensive (TCP handshake, auth, select db)
- If called in a loop (paging, sync), performance is degraded
- The “client-per-request” pattern is not optimal for Redis

###  5. `RawStreamClient` interface not completely abstract
__Issues__:

1. __`rawEvent: string[]`__ exposes the Redis XADD format:

```typescript
   // Redis format: [‘field1’, ‘value1’, ‘field2’, ‘value2’]
   ```

   Another implementation (Kafka) should marshal to string[] and then unmarshal

2. __`AuthContext` and `AuthUser` in the interface__:

   - Coupling with the OpenCTI business domain
   - A pure stream client should not know about `AuthContext`
   - Only `stream-handler.ts` should handle this mapping

3. __`initializeStreams` optional__: weak contract

File: opencti-platform/opencti-graphql/src/manager/pirManager.ts [hunk 1]
```diff
@@ -165,29 +166,22 @@ const processStreamEventsForPir = (context:AuthContext, pir: BasicStoreEntityPir
...
-        redisClient,
-        pir.lastEventId,
-        processStreamEventsForPir(context, pir),
-        { streamBatchSize: PIR_MANAGER_STREAM_BATCH_SIZE }
-      );
+  const context = executionContext(PIR_MANAGER_CONTEXT);
+  const allPirs = await getEntitiesListFromCache<BasicStoreEntityPir>(context, PIR_MANAGER_USER, ENTITY_TYPE_PIR);
+
+  // Loop through all Pirs by group
+  await BluePromise.map(allPirs, async (pir) => {
```
xfournet: could we use Promise.all instead of BluePromise.map ?

opencti-platform/opencti-graphql/src/manager/pirManager.ts
```diff
@@ -165,29 +166,22 @@ const processStreamEventsForPir = (context:AuthContext, pir: BasicStoreEntityPir
  * Handler called every {PIR_MANAGER_INTERVAL} and studying a range of stream events.
  */
 const pirManagerHandler = async () => {
-  const redisClient = await createRedisClient(PIR_MANAGER_LABEL, false);
```
xfournet, line 168: i'm not sure to understand why we didn't pass the client anymore, and have a specific logic in redis-stream to handle the PIR case

that's said i'm sure also to understand why initialy we need a separate client for PIR ?

opencti-platform/opencti-graphql/src/manager/pirManager.ts
```diff
@@ -165,29 +166,22 @@ const processStreamEventsForPir = (context:AuthContext, pir: BasicStoreEntityPir
...
-      if (lastEventId !== pir.lastEventId) {
-        await updatePir(context, PIR_MANAGER_USER, pir.id, [{ key: 'lastEventId', value: [lastEventId] }], { auditLogEnabled: false });
-      }
-    }, { concurrency: PIR_MANAGER_MAX_CONCURRENCY });
-  } finally {
-    // close redis client connexion
-    redisClient.disconnect();
-  }
+    if (lastEventId !== pir.lastEventId) {
+      await updatePir(context, PIR_MANAGER_USER, pir.id, [{ key: 'lastEventId', value: [lastEventId] }], { auditLogEnabled: false });
```
xfournet, line 180: BTW here the 'state' is saved at the end of the batch, so a element of a batch can be redone if it has been stopped in the middle. not sure if it can be problem ?

The state is currently stored in Elasticsearch, which isn’t an issue because we only update it after a full batch execution. However, if we move toward more granular state updates, it would likely be better to store the state in Redis rather than Elasticsearch.

opencti-platform/opencti-graphql/src/database/redis-stream.ts
```diff
@@ -0,0 +1,241 @@
...
+        effectiveStartEventId = lastStreamResultId;
+      }
+    } else {
+      await processStreamResult([], callback, opts.withInternal);
+    }
+  } catch (err) {
+    logApp.error('Redis stream consume fail', { cause: err });
+    if (opts.autoReconnect) {
+      await waitInSec(2);
+    }
```
xfournet: autoreconnect is always false in usages. and if true, i don't understand why a wait is needed ?

opencti-platform/opencti-graphql/src/manager/pirManager.ts
JeremyCloarec: Will be handled in #13257 after this PR is merged

xfournet (approved):

xfournet:
I'm surprised that code coverage is not very high, but probably it is not taking account end to end tests ?

JeremyCloarec:
> I'm surprised that code coverage is not very high, but probably it is not taking account end to end tests ?

I'm not sure either, but I think that it's because these method are not tested directly but indirectly, through the managers running on the test instance

PR #13273, opened 2025-11-21
[frontend/backend] Dynamic PIR filters format refacto to handle several PIR ids (#13344)

Archidoit:
### Proposed changes
Modify the format of the dynamic pir filters (ie 'pir_score' and 'last_pir_score_date')
- to be able to later generalize those filters to several Pirs
- to be consistant with other filters having a given filter key instead of a key depending on an id

### Old vs new format

Former format:
image

New format:
image

### Related issues
#13344

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/pir/pir_knowledge/PirKnowledge.tsx
+++ b/opencti-platform/opencti-front/src/private/components/pir/pir_knowledge/PirKnowledge.tsx
@@ -34,12 +34,12 @@ interface PirKnowledgeProps {
 const PirKnowledge = ({ data }: PirKnowledgeProps) => {
   const pir = useFragment(knowledgeFragment, data);
   const pirId = pir.id;
-  const LOCAL_STORAGE_KEY = `PirSourcesFlaggedList-${pirId}`;
+  const LOCAL_STORAGE_KEY = `Pir-SourcesFlaggedList-${pirId}`;
 
   const initialValues = {
     filters: {
       ...emptyFilterGroup,
-      filters: useGetDefaultFilterObject([`pir_score.${pirId}`, `last_pir_score_date.${pirId}`], ['Stix-Domain-Object']),
+      filters: useGetDefaultFilterObject(['pir_score', 'last_pir_score_date'], ['Stix-Domain-Object']),
     },
     searchTerm: '',
     sortBy: 'pir_score',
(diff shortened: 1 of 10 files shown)


opencti-platform/opencti-front/src/private/components/pir/pir_knowledge/PirKnowledge.tsx
```diff
@@ -34,12 +34,12 @@ interface PirKnowledgeProps {
 const PirKnowledge = ({ data }: PirKnowledgeProps) => {
   const pir = useFragment(knowledgeFragment, data);
   const pirId = pir.id;
-  const LOCAL_STORAGE_KEY = `PirSourcesFlaggedList-${pirId}`;
+  const LOCAL_STORAGE_KEY = `Pir-SourcesFlaggedList-${pirId}`;
```
Archidoit, line 37: local storage key changed to clean the local storage and thus avoid having pir filters in the old format

opencti-platform/opencti-graphql/src/utils/filtering/filtering-completeSpecialFilterKeys.ts
```diff
@@ -524,22 +524,21 @@ const adaptFilterToIsInferredFilterKey = (filter: Filter) => {
...
+    throw FunctionalError('This filter should be related to at least 1 Pir', { filter });
   }
-  const pirKey = splittedKey[0];
-  const pirId = splittedKey[1];
   // check the user has access to the PIR
-  await getPirWithAccessCheck(context, user, pirId);
-  // push the nested pir_score filter associated to the given PIR ID
+  await Promise.all(pirIds.map((pirId) => getPirWithAccessCheck(context, user, pirId)));
+  // push the nested pir filter associated to the given PIR IDs
+  const subKey = filterKey === PIR_SCORE_FILTER ? 'score' : 'date';
```
OctaveLaventure: Is there any way we can avoid having 'score'/'date' magic strings here? maybe a constant somewhere or at least a union type to name this behavior

OctaveLaventure (approved):

opencti-platform/opencti-graphql/src/utils/filtering/filtering-completeSpecialFilterKeys.ts
Archidoit: fixed :) I created 2 constants for these 2 subfilters keys :)

opencti-platform/opencti-front/src/private/components/pir/pir_knowledge/PirKnowledge.tsx
lndrtrbn: So after the merge users will lose their filters on PIR table

lndrtrbn (approved):

opencti-platform/opencti-front/src/private/components/pir/pir_knowledge/PirKnowledge.tsx
Archidoit: Yes exactly !

PR #13298, opened 2025-11-24
[backend] Support upsert on modified (#13296)

### LINKED ISSUE #13296: Upsert modified field
## Use case

Upsert modified field

SamuelHassine:
### Proposed changes

* [backend] Support upsert on modified

### Description

The "modified" field is now upserted when it is after the existing one, a stream event is emmitted. 

### Related issues

* Closes #13296


aHenryJard (approved):

PR #13308, opened 2025-11-24
[backend] fix @me value conversion for stix filtering (#13040)

### LINKED ISSUE #13040: Group Triggers Using "Assignees = @me" Notify All Group Members, Not Just the Assigned Member
## Description
When a trigger is created under a Group and filters use "Entity type =Report" and "Assignees = @me," when the Report is assigned to only one member of the Group, all members will receive a notification. The expected behavior is that only the assignee will receive a notification.

## Environment
6.8.10

## Reproducible Steps

1. Create a Trigger under a Group with multiple members, using the filters "Entity type =Report" and "Assignees = @me."
2. Navigate to a report and assign the Report to only one member of the group.
3. Observe how all members of the Group receive a notification.

## Expected Output
Only the member who is assigned to the Report receives a notification. 

## Actual Output
All members of the Group receive a notification.

Archidoit:
### Proposed changes
- Backend filters checking little code improvements
- Fix @me filter value conversion for stix filtering (avoid filters values reassignement: create a new value that is returned)
- add backend tests for @me filter value in stix filtering
- improve replaceEnrichValuesInFilters by removing the params reassignement
- add tests for replaceEnrichValuesInFilters with labels values replacement

### Related issues
#13040

### DIFF
--- a/opencti-platform/opencti-graphql/tests/02-integration/01-database/stix-filtering-test.ts
+++ b/opencti-platform/opencti-graphql/tests/02-integration/01-database/stix-filtering-test.ts
@@ -140,7 +140,7 @@ describe('Stix Filtering', () => {
     expect(await isStixMatchFilterGroup_MockableForUnitTests(testContext, ADMIN_USER, stixIndicator, filterGroup, MOCK_RESOLUTION_MAP)).toEqual(true);
   });
 
-  it('matches stix object with filters with @me value in filters', async () => {
+  it('matches stix object with filters containing a @me value (@me value representing the current user)', async () => {
     const filterGroup = {
       mode: 'and',
       filters: [{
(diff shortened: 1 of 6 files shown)


opencti-platform/opencti-graphql/tests/02-integration/01-database/stix-filtering-test.ts
```diff
@@ -139,6 +140,30 @@ describe('Stix Filtering', () => {
     expect(await isStixMatchFilterGroup_MockableForUnitTests(testContext, ADMIN_USER, stixIndicator, filterGroup, MOCK_RESOLUTION_MAP)).toEqual(true);
   });
 
+  it('matches stix object with filters with @me value in filters', async () => {
```
MTorbay-Filigran: Could it be more explicit what the @me is made to match, maybe such as : 

it('matches stix object when the @me filter targets the current user'

MTorbay-Filigran (commented):
Working functionaly, all good for me, I just added a comment on the test title to be clearer about what does the filter

opencti-platform/opencti-graphql/tests/02-integration/01-database/stix-filtering-test.ts
Archidoit: fixed :)

MTorbay-Filigran (approved):

PR #13327, opened 2025-11-25
[frontend/backend] PIR in Playbooks - detect entities linked to entities of interest (#12362)

lndrtrbn:
### Proposed changes

* Some refactoring on code of previous chunks
* Listen to relationship creation in Playbook component "Listen PIR events"
* Trigger playbook if a relationship creation concerned an entity with a flagged entity
* Update tests

### Related issues

* #12362

### Checklist


opencti-platform/opencti-graphql/tests/01-unit/manager/playbookManager/listenPirEventsUtils-test.ts
```diff
@@ -130,53 +145,182 @@ describe('listenPirEventsUtils', () => {
...
+      expect(middlewareLoader.storeLoadById).not.toHaveBeenCalled();
+    });
+  });
+
+  describe('stixIdOfLinkedEntity', () => {
+    it('should return null if it is a stix relation but not a relationship creation', async () => {
+      vi.spyOn(playbookUtils, 'isEventCreateRelationship').mockReturnValue(false);
+      vi.spyOn(stixRelationship, 'isStixRelation').mockReturnValue(true);
+
+      const result = await stixIdOfLinkedEntity(
```
SouadHadjiat: why not test on a stix relationship event instead of mocking isEventCreateRelationship & stixRelationship ? 
I think it would be more comprehensive to have a unit test on stixIdOfLinkedEntity with a "real" stix relationship creation event.

opencti-platform/opencti-graphql/src/manager/playbookManager/listenPirEventsUtils.ts
```diff
@@ -77,9 +75,25 @@ export const isEventMatchesPir = async (
...
+ * @returns True if the event matches selected PIR.
+ */
+export const isUpdateEventMatchPir = async (
+  context: AuthContext,
+  eventData: StreamDataEvent,
+  config: PirStreamConfiguration,
+  pirList?: { value: string }[]
+) => {
+  if (isEventUpdateOnEntity(eventData) && config.update) {
     const entityPirList = await listOfPirInEntity(context, eventData.data.id);
```
SouadHadjiat: why do we need to fetch the entity to get the pir_information, isn't it already part of the stream ? pir_information is in stix-converter, and if it's an update on the entity we should see it in the stream

opencti-platform/opencti-front/src/private/components/data/playbooks/PlaybookAddComponents.jsx
```diff
@@ -838,15 +838,22 @@ const PlaybookAddComponentsContent = ({
                     );
                   }
                   if (v.type === 'boolean') {
+                    let helperText = '';
+                    if (v.$ref === 'An entity is linked to an entity from a selected PIR') {
```
SouadHadjiat: could we test on the key 'create_rel' instead of the message which might change ?

opencti-platform/opencti-graphql/src/manager/playbookManager/listenPirEventsUtils.ts
SouadHadjiat: otherwise, if we need to fetch PIRs, we can use the cache (all PIRs are stored in cache)

opencti-platform/opencti-graphql/src/manager/playbookManager/listenPirEventsUtils.ts
```diff
@@ -88,11 +102,40 @@ export const isEventMatchesPir = async (
...
+ * @param eventData The event.
+ * @param pirList List of selected PIR.
+ * @returns the id of the linked entity if the event matches selected PIR.
+ */
+export const stixIdOfLinkedEntity = (
+  eventData: StreamDataEvent,
+  config: PirStreamConfiguration,
+  pirList?: { value: string }[],
+) => {
+  if (isEventCreateRelationship(eventData) && isStixRelation(eventData.data) && config.create_rel) {
```
SouadHadjiat, line 113: isStixRelation is unecessary here since it's already tested in isEventCreateRelationship

opencti-platform/opencti-graphql/src/manager/playbookManager/listenPirEventsUtils.ts
SouadHadjiat: from my debug, pir_information already contains pirs internal ids, and pirList also has internal ids of pirs, so we should be able to directly compare them without listing anything from the database.

opencti-platform/opencti-graphql/src/manager/playbookManager/listenPirEventsUtils.ts
lndrtrbn: it is necessary to let TypeScript infer the type correctly, otherwise we need to make a hard cast (... as StixRelation)

opencti-platform/opencti-graphql/src/manager/playbookManager/listenPirEventsUtils.ts
lndrtrbn: changed thanks

opencti-platform/opencti-front/src/private/components/data/playbooks/PlaybookAddComponents.jsx
lndrtrbn: indeed way better

SarahBocognano (approved):
Tested and pair reviewed ✅

opencti-platform/opencti-graphql/tests/01-unit/manager/playbookManager/playbookManagerUtils-test.ts
```diff
@@ -98,4 +104,58 @@ describe('playbookManagerUtils', () => {
...
+        type: StreamDataEventTypeEnum.CREATE,
+        scope: 'external',
+      } as unknown as StreamDataEvent;
+
+      const result = isEventCreateRelationship(streamEventMock);
+
+      expect(result).toBe(true);
+    });
+
+    it('should return false if alla conditions are not met: not a sti relationship', () => {
```
SarahBocognano: Fix the message alla/met/sti xD

opencti-platform/opencti-graphql/src/manager/playbookManager/listenPirEventsUtils.ts
```diff
@@ -38,34 +37,26 @@ export const buildPirFilters = (pirList: { value: string }[]) => {
...
  */
-export const listOfPirInEntity = async (context: AuthContext, entityId: string) => {
-  const entityFromId = await storeLoadById<BasicStoreEntity>(
-    context,
-    AUTOMATION_MANAGER_USER,
-    entityId,
-    ABSTRACT_STIX_CORE_OBJECT
-  );
-  return entityFromId.pir_information?.map((pir) => pir.pir_id) || [];
+export const listOfPirInEntity = async (entity: StixCoreObject) => {
```
SouadHadjiat: not an async function anymore. 
Is the naming listOfPirInEntity still coherent with the function ?

opencti-platform/opencti-graphql/tests/01-unit/manager/playbookManager/listenPirEventsUtils-test.ts
```diff
@@ -130,53 +145,178 @@ describe('listenPirEventsUtils', () => {
...
+      expect(result).toBeNull();
+    });
+
+    it('should return expected id if there is no pirList and at least one source ref pir id', async () => {
+      vi.spyOn(playbookUtils, 'isEventCreateRelationship').mockReturnValue(true);
+      vi.spyOn(stixRelationship, 'isStixRelation').mockReturnValue(true);
+
+      const randomEventInPirWithPirOnlyInSource = { ...randomEventInPir };
+
+      const result = await stixIdOfLinkedEntity(
```
SouadHadjiat: I find it weird to have this test based on randomEventInPir.data which is not a stix relationship but this test returns a result because the mocked functions work. But really, the input used in this test should not return this result.

SouadHadjiat: It would be more accurate to have a more representative input, that would really return the result that we expect.

SouadHadjiat: and also this test depend on the implementation of other functions like isEventCreateRelationship, which is not unit tested, and which could change, but here we mock the result, so it won't break the test even though the result change. This test should be able to respond to "if I have this input, I should get this output".

opencti-platform/opencti-graphql/tests/01-unit/manager/playbookManager/listenPirEventsUtils-test.ts
SouadHadjiat: also this function is not async, there is no need for await

opencti-platform/opencti-graphql/tests/01-unit/manager/playbookManager/playbookManagerUtils-test.ts
lndrtrbn: done, "met" is not a typo

opencti-platform/opencti-graphql/tests/01-unit/manager/playbookManager/listenPirEventsUtils-test.ts
```diff
@@ -130,53 +136,168 @@ describe('listenPirEventsUtils', () => {
...
+  describe('stixIdOfLinkedEntity', () => {
+    it('should return null if it is a stix relation but not a relationship creation', async () => {
+      const randomRelationshiEventNotCreation = {
+        ...randomEventNotInPir, data: {
+          ...randomEventNotInPir.data, 
+          type:'delete'
+        }
+      } as unknown as SseEvent<StreamDataEvent>;
+
+      const result = await stixIdOfLinkedEntity(
```
SouadHadjiat: there is no need for await here since stixIdOfLinkedEntity is not async (and the test shouldn't be async too)

PR #13335, opened 2025-11-26
[backend] Introduce new default capabilties in connector role (#13334)

### LINKED ISSUE #13334: Incorrect default "connectors" role capabilities
## Description

Following the capability splitting described in this issue: https://github.com/OpenCTI-Platform/opencti/issues/8102, the default "connectors" role does not have the correct capabilities enabled by default.

In the "Access to admin functionalities" section, the role should have the following capabilities.
- Manage marking definitions
- Manage labels
- Manage vocabularies
- Manage kill chain phases

Currently, it only has the following: 
- Manage marking definitions
- Manage labels

This causes a problem for many connectors that need to manage kill chain phases and vocabulary.

## Environment

OpenCTI version: 6.8.13

richard-julien:
See #13334


romain-filigran (approved):
All good. Thank you

Issue #13275, opened 2025-11-21
Playbook trigger does not fire when using creators = "@me"


snitram59:
## Description

When configuring a playbook trigger in OpenCTI, the trigger does not execute if the filter includes:
Entity type: IPv4 address
Creators: @me
When I create an IPv4 address myself (as the logged-in user), the playbook is never triggered.

## Environment

1. OS (where OpenCTI server runs): { e.g. Mac OS 10, Windows 10, Ubuntu 16.4, etc. }
2. OpenCTI version: { e.g. OpenCTI 1.0.2 }
3. OpenCTI client: { e.g. frontend or python }
4. Other environment details:

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. { e.g. Run ... }
2. { e.g. Click ... }
3. { e.g. Error ... }

snitram59:
**Environment**

OCTI 6.8.13

**Actual behavior**

No trigger is fired.
If I remove the creators = "@ me" filter, the trigger works normally.

**Steps to reproduce**

1.Create a new playbook.

2. Add a trigger:

3. Listen to create events

4. Add filters:
Entity type = IPv4 address
Creators = @ me

5.Save the playbook.

6.Create a new IPv4 address as the same logged-in user.

➡️ The playbook does not trigger.

nino-filigran:
I think the me filter has been introduced by mistake in the playbook. Indeed, filtering on "me" on a playbook does not really make sense, because the playbook does know who is "me". @RomainGUIGNARD what's your thoughts on that?

Archidoit:
Note that there is the same issue with a trigger or a stream with the @me value in the filters

Explanations: @me value is not converted for stix filtering
--> Do we want to support @me value in playbooks (@me value will be the one of the playbook creator)? or only in triggers and streams (@me value is the value of the user for which the trigger/stream is running)?

romain-filigran:
Agree with you @nino-filigran. As playbook are executed with a "technical account", it's doesn't make sense to have this "@me" filter in playbook. To be removed from playbook components.

@snitram59 : If your goal is to process entities you created in a playbook, you must explicitly set your "user account" as the creator and not "@me".

@Archidoit : Do you think possible to remove this "@me" filter value in playbook ?

Archidoit:
@romain-filigran Okay! So I will remove the @me value possibility in playbook (UI). What do we do for existing playbooks having this value? simply remove the filter ?

[...]

PR #13341, opened 2025-11-26
[frontend/backend] @me value should not be authorized in playbooks components with stix filtering (#13275)

Archidoit:
### Proposed changes
- UI: In Playbook components with stix filtering
       - the @me filter value should not be listed in the possible values of user-value filters (exemple: creator filter)
       - for former playbooks with a @me value in filters, the @me value is displayed as deleted
- Backend: For playbook components with stix filtering, remove filters with @me value if they are some
- Add backend tests for playbook-utils.ts

### Related issues
#13275

### DIFF
--- a/opencti-platform/opencti-front/src/components/FilterIconButton.tsx
+++ b/opencti-platform/opencti-front/src/components/FilterIconButton.tsx
@@ -61,6 +61,7 @@ const FilterIconButtonWithRepresentativesQuery: FunctionComponent<FilterIconButt
     filterValuesContentQuery,
     {
       filters: removeIdFromFilterGroupObject(filters) as unknown as GqlFilterGroup,
+      isMeValueForbidden: searchContext?.elementType === 'Playbook-Stix-Component',
     },
   );
   return (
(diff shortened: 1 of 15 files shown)


SarahBocognano:
Capture d’écran 2025-12-01 à 11 38 41

The @me filter is still appearing if I created this filter before checkout to your branch.
The playbook doesn't seems to trigger as well 

It's not showing anymore in the filter list though ✅

Archidoit:
> Capture d’écran 2025-12-01 à 11 38 41
> The @me filter is still appearing if I created this filter before checkout to your branch. The playbook doesn't seems to trigger as well
> 
> It's not showing anymore in the filter list though ✅

It has been decided that for a former playbook with a @me value, the @me value is displayed as 'deleted' and the filter acts as if the user filter value doesn't exist : 
image

SarahBocognano:
It's good for me. The creators is indeed displayed as "deleted", and for what we talk about, it is normal that the playbooks doesn't trigger as there is this deleted value in the filters.

SarahBocognano (approved):

PR #13346, opened 2025-11-26
[backend] fix filterMembersWithUsersOrgs (#13338)

### LINKED ISSUE #13338: User segregation: entities not accessible since user is "null" in knowledge history
## Description

In context of segregation:
- Org A has a user that has done some changes on an entity.
Via authorize members, this entity access is granted to an other org B.
- User of org B tries to access the entity. 

The screen crashes, since the user of Org B that tries to access the entity cannot see the user of Org A in history, since the user value is "null". It should not be null, but rather should be filled with the user, but displayed as restricted for the user of OrgB. 
Users of the same Org than the user of Org A should see the name of the user who did the change. 

## Environment

Production
## Reproducible Steps

1. Have a paltform with Org Segregation
2. Define two Orgs, Org A and Org B.
3. Create a report with a userA of Org A. Change the status of report. 
4. With a playbook, enable authorize member on the report, and give users of OrgA read Access & user of OrgB read access
5. Login with user B (user of ORgB)
6. Try to access the report

## Expected Output

Report should be accessible, and in the history we should see "restricted" has replaced {newStatusValue} in Status

## Actual Output

Crash

Image

JeremyCloarec:
### Proposed changes

* fix filterMembersWithUsersOrgs logic introduced in https://github.com/OpenCTI-Platform/opencti/pull/12894
* before changes, we were correctly returning the member if it was a member in the same orgs as the user. After the change, this default return was removed, causing all members in the same orgs to be filtered out
*

### Related issues

* #13338 
*

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/utils/access.ts
+++ b/opencti-platform/opencti-graphql/src/utils/access.ts
@@ -929,7 +929,9 @@ export const filterMembersWithUsersOrgs = async (
       } else {
         const memberOrgIds = member[RELATION_PARTICIPATE_TO] ?? [];
         const sameOrg = memberOrgIds.some((id) => userOrgIds.includes(id));
-        if (!sameOrg) {
+        if (sameOrg) {
+          resultMembers.push(member);
+        } else {
           if (filterMode === FilterMembersMode.RESTRICT) {
             const restrictedMember = {
               ...member,
@@ -942,8 +944,6 @@ export const filterMembersWithUsersOrgs = async (
             };
             resultMembers.push(restrictedMember);
           }
-        } else {
-          resultMembers.push(member);
         }
       }
     }
--- a/opencti-platform/opencti-graphql/src/utils/access.ts
+++ b/opencti-platform/opencti-graphql/src/utils/access.ts
@@ -926,21 +926,24 @@ export const filterMembersWithUsersOrgs = async (
       const member = members[i];
       if (member.id === user.id || INTERNAL_USERS[member.id] || member.user_service_account) {
         resultMembers.push(member);
-      }
-      const memberOrgIds = member[RELATION_PARTICIPATE_TO] ?? [];
-      const sameOrg = memberOrgIds.some((id) => userOrgIds.includes(id));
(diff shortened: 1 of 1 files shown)


opencti-platform/opencti-graphql/src/utils/access.ts
```diff
@@ -926,21 +926,24 @@ export const filterMembersWithUsersOrgs = async (
...
-              main: RESTRICTED_USER.name,
-              secondary: RESTRICTED_USER.name
-            }
-          };
-          resultMembers.push(restrictedMember);
+      } else {
+        const memberOrgIds = member[RELATION_PARTICIPATE_TO] ?? [];
+        const sameOrg = memberOrgIds.some((id) => userOrgIds.includes(id));
+        if (!sameOrg) {
+          if (filterMode === FilterMembersMode.RESTRICT) {
```
Kedae, line 935: Not sure to get that part. If we are in exclude mode we will still have an empty array eventually ?

JeremyCloarec: Yes, we're using the exclude mode when fetching the possible creators/participants/assignees values in filters. In this case, we want to not receive any restricted member instead of showing them as Restricted

opencti-platform/opencti-graphql/src/utils/access.ts
```diff
@@ -926,21 +926,24 @@ export const filterMembersWithUsersOrgs = async (
...
-            representative: {
-              main: RESTRICTED_USER.name,
-              secondary: RESTRICTED_USER.name
-            }
-          };
-          resultMembers.push(restrictedMember);
+      } else {
+        const memberOrgIds = member[RELATION_PARTICIPATE_TO] ?? [];
+        const sameOrg = memberOrgIds.some((id) => userOrgIds.includes(id));
+        if (!sameOrg) {
```
OctaveLaventure: **1. Refactoring**
avoid `if not` for clarity, I think you can reverse the if/else

OctaveLaventure (approved):

PR #13358, opened 2025-11-27
[Backend] Moved a unit test to integration

MTorbay-Filigran:
### Proposed changes

* Quick PR to be able to launch the unit test locally without failing dued to configuration : here test was failing when user is not on an Etreprise Edition so I moved it to the integration test folder
*

### Related issues

*
*

### Checklist


Archidoit (approved):

opencti-platform/opencti-graphql/tests/02-integration/01-database/sorting-test.ts
```diff
@@ -0,0 +1,12 @@
+import { describe, expect, it, vi } from 'vitest';
+import { buildElasticSortingForAttributeCriteria } from '../../../src/utils/sorting';
+import { SYSTEM_USER } from '../../../src/utils/access';
+import { testContext } from '../../utils/testQuery';
+
+describe('Sorting utilities', () => {
```
Archidoit, line 6: since sorting is not a resolver, I suggest not to create a sorting-test file in resolvers test... Maybe add this test to the pir-test.ts file, or in 01-database/sorting-test.ts

Archidoit: I think the best solution is to create a new file sorting-test, but in 01-database, instead of in 02-resolver

PR #13383, opened 2025-11-28
[backend] Fix pagination hasNextPage when post-filtering removes hits (#13087)

### LINKED ISSUE #13087: fix(filters): pagination not always working
## Description
In the attached screenshot you can see there are 1.52k results in total, the first page shows 100 results, if I then click the small arrow to the right to go the next page it remains empty. 

None of the following pages work as well. We have observed this behavior on multiple views (reports/indicators/all entities). It is inconsistent, and does sometimes work. Clearing the cache seems to solve it sometimes.

## Environment
1. 6.8.8+
2. Firefox 142.0 & 144.0.2

## Expected Output
The expected behavior is that the next 100 entities will appear.

## Actual Output
Nothing appears

## Additional information
While I could not replicate this on staging testing (6.8.10), I was able to on the CSM platform when paginating the second page, on reports (GIF below).
I am concerned that this is sporadic. When it does not work, the browser dev tools (via Network 'Response') reports hasNextPage='False' when clearly there are more pages to render.  Could it be a timeout response from Elastic?? There seems to be a time element aspect to this..

## Screenshots (optional)

Image

amiraifori:
## Summary

This PR fixes a backend pagination issue where some list views (e.g. reports, indicators, all entities) can show a large total result count (for example ~1.52k) but navigating to the next page occasionally returns an empty page. When this happens, the GraphQL response exposes `pageInfo.hasNextPage = false` even though more results clearly exist.

The root cause is in the generic pagination helper that computes `hasNextPage` based only on how many hits are returned in a single batch, without fully taking into account post-filtering and the global hit count from the search engine.

---

## Root cause

Pagination for entity and relation lists goes through:

- `elPaginate` → `buildSearchResult` → `buildPagination` → `buildPaginationFromEdges`
- `pageEntitiesConnection` / `pageRelationsConnection` (and higher-level resolvers) use this flow.

In `src/database/utils.ts`, `buildPaginationFromEdges` was previously setting:

```ts
const hasNextPage = (edges.length + filteredCount) === limit;
```

However:

- `elPaginate` enables `track_total_hits: true` and retrieves:
  - `globalCount = data.hits.total.value`
  - `elements = elConvertHits(data.hits.hits)`
- It then applies additional post-filtering (e.g. `buildRegardingOfFilter`, ACL logic), resulting in:
  - `filteredElements`
  - `filterCount = elements.length - filteredElements.length`

When a page of ES hits is partly or fully filtered out, it is possible that:

- `edges.length + filteredCount < limit`  
  (because many hits were filtered in this batch), **but**
- `globalCount` is still high and there are more matching entities after the current cursor.

In those cases the previous logic sets `hasNextPage = false` too early, which:

- Stops the repagination loop in `elRepaginate`,
- Returns `pageInfo.hasNextPage = false` from GraphQL,
- Leads the frontend to show an empty next page despite the total count indicating many more results.

This matches the sporadic behavior described in issue **#13087**.

---

## Implementation details

**File changed**

- `opencti-platform/opencti-graphql/src/database/utils.ts`

**Function**

- `buildPaginationFromEdges<T>(
    limit: number | undefined,
    searchAfter: string | undefined | null,
    edges: BasicNodeEdge<T>[],
    globalCount: number,
    filteredCount = 0
  ): BasicConnection<T>`

**Previous logic**

```ts
// Because of stateless approach its difficult to know if its finish
// this test could lead to an extra round trip sometimes
const hasNextPage = (edges.length + filteredCount) === limit;
```

**New logic**

```ts
// Because of stateless approach its difficult to know if its finish
// this test could lead to an extra round trip sometimes
//
// Previous implementation was relying only on (edges.length + filteredCount) === limit
// which can incorrectly report no next page when a whole batch of hits is
// filtered out, even if more data exists after the current cursor.
let hasNextPage = false;
if (limit && limit > 0) {
  const pageSize = edges.length + filteredCount;
  if (pageSize >= limit) {
    // We received at least the requested page size (including filtered out hits),
    // so it is safe to assume there might be more data.
    hasNextPage = true;
  } else if (
    pageSize < limit &&
    filteredCount > 0 &&
    globalCount > pageSize &&
    searchAfter !== undefined &&
    searchAfter !== null
  ) {
    // We received fewer usable edges than the requested limit, but some hits
    // were filtered out and the global count indicates there are still more
    // results in the index. In this situation, we allow another round-trip
    // so that subsequent pages can surface remaining matching entities.
    hasNextPage = true;
  }
}

const hasPreviousPage = searchAfter !== undefined && searchAfter !== null;
const startCursor = edges.length > 0 ? edges[0].cursor : '';
const endCursor = edges.length > 0 ? edges[edges.length - 1].cursor : '';
const pageInfo = {
  startCursor,
  endCursor,
  hasNextPage,
  hasPreviousPage,
  globalCount,
};
return { edges, pageInfo };
```

**Behavioral impact**

- In the common case where ES returns at least `limit` relevant hits for a page (after post-filtering), behavior is unchanged: `hasNextPage` remains `true`.
- In edge cases where:
  - a page is strongly filtered (`filteredCount > 0`),
  - the resulting `pageSize` is below `limit`,
  - but `globalCount` still indicates additional hits and we are not on the first page (`searchAfter` is set),
  the backend now **allows an extra page fetch** instead of stopping pagination prematurely.

This is consistent with the existing comment acknowledging that a stateless cursor-based approach can sometimes cost an extra round trip.

---

## Scope and risk

- The change is confined to a **single helper** (`buildPaginationFromEdges`) used by:
  - `elPaginate` / `buildSearchResult`,
  - `elConnection` / `pageEntitiesConnection` / `pageRelationsConnection` and their consumers.
- It does **not** change API shape or schema:
  - Same `pageInfo` fields,
  - Same `edges` structure.
- The main effect is to prevent `hasNextPage` from becoming `false` in situations where:
  - `globalCount` and the user-visible total clearly indicate more data,
  - but the current batch was heavily filtered.

---

## Testing

**Local**

- Attempted to run backend tests with:

  ```bash
  yarn test opencti-graphql
  ```

Tests not run locally due to Corepack/Yarn 4 requirement; relying on project CI.

---

## Related issue

- Fixes: #13087

### DIFF
--- a/opencti-platform/opencti-graphql/src/database/utils.ts
+++ b/opencti-platform/opencti-graphql/src/database/utils.ts
@@ -269,25 +269,11 @@ export const buildPaginationFromEdges = <T>(
   // Previous implementation was relying only on (edges.length + filteredCount) === limit
   // which can incorrectly report no next page when a whole batch of hits is
   // filtered out, even if more data exists after the current cursor.
-  let hasNextPage = false;
-  if (limit && limit > 0) {
-    const pageSize = edges.length + filteredCount;
-    if (pageSize >= limit) {
-      // We received at least the requested page size (including filtered out hits),
-      // so it is safe to assume there might be more data.
-      hasNextPage = true;
-    } else if (pageSize < limit
-      && filteredCount > 0
-      && globalCount > pageSize
-      && searchAfter !== undefined
-      && searchAfter !== null) {
-      // We received fewer usable edges than the requested limit, but some hits
-      // were filtered out and the global count indicates there are still more
-      // results in the index. In this situation, we allow another round-trip
-      // so that subsequent pages can surface remaining matching entities.
-      hasNextPage = true;
-    }
-  }
+  const pageSize = edges.length + filteredCount;
+  // pageSize corresponds to the number of raw hits returned by the search engine
+  // (including hits later removed by post-filtering).
+  // If the engine returned a full batch, another page might exist.
+  const hasNextPage = !!limit && limit > 0 && pageSize > 0 && pageSize >= limit;
   // For same reason its difficult to know if a previous page exists.
   // Considering for now that if user specific an offset, it should exists a previous page.
   const hasPreviousPage = searchAfter !== undefined && searchAfter !== null;
--- a/opencti-platform/opencti-graphql/src/database/utils.ts
+++ b/opencti-platform/opencti-graphql/src/database/utils.ts
@@ -265,7 +265,29 @@ export const buildPaginationFromEdges = <T>(
 ): BasicConnection<T> => {
   // Because of stateless approach its difficult to know if its finish
   // this test could lead to an extra round trip sometimes
-  const hasNextPage = (edges.length + filteredCount) === limit;
+  //
+  // Previous implementation was relying only on (edges.length + filteredCount) === limit
+  // which can incorrectly report no next page when a whole batch of hits is
+  // filtered out, even if more data exists after the current cursor.
+  let hasNextPage = false;
+  if (limit && limit > 0) {
+    const pageSize = edges.length + filteredCount;
+    if (pageSize >= limit) {
+      // We received at least the requested page size (including filtered out hits),
+      // so it is safe to assume there might be more data.
+      hasNextPage = true;
+    } else if (pageSize < limit
+      && filteredCount > 0
+      && globalCount > pageSize
+      && searchAfter !== undefined
+      && searchAfter !== null) {
+      // We received fewer usable edges than the requested limit, but some hits
+      // were filtered out and the global count indicates there are still more
+      // results in the index. In this situation, we allow another round-trip
+      // so that subsequent pages can surface remaining matching entities.
+      hasNextPage = true;
+    }
+  }
   // For same reason its difficult to know if a previous page exists.
   // Considering for now that if user specific an offset, it should exists a previous page.
   const hasPreviousPage = searchAfter !== undefined && searchAfter !== null;
--- a/opencti-platform/opencti-graphql/tests/01-unit/database/utils-test.ts
+++ b/opencti-platform/opencti-graphql/tests/01-unit/database/utils-test.ts
@@ -1,5 +1,9 @@
 import { describe, expect, it } from 'vitest';
-import { extractObjectsPirsFromInputs, extractObjectsRestrictionsFromInputs } from '../../../src/database/utils';
+import {
+  buildPaginationFromEdges,
+  extractObjectsPirsFromInputs,
+  extractObjectsRestrictionsFromInputs,
+} from '../../../src/database/utils';
 import { ENTITY_TYPE_CONTAINER_REPORT, ENTITY_TYPE_MALWARE } from '../../../src/schema/stixDomainObject';
 import { EditOperation } from '../../../src/generated/graphql';
 
@@ -178,5 +182,19 @@ describe('Function extractObjectsPirsFromInputs()', () => {
   it('should return empty array if not a container', () => {
     const { pir_ids } = extractObjectsPirsFromInputs(pirInputs, ENTITY_TYPE_MALWARE);
     expect(pir_ids).toEqual([]);
+  });
+});
+
+describe('buildPaginationFromEdges()', () => {
+  it('should set hasNextPage=true when raw page size (edges + filteredCount) reaches limit', () => {
+    const edges = Array.from({ length: 7 }, (_, i) => ({ node: { id: i }, cursor: `c${i}` } as any));
+    const result = buildPaginationFromEdges(10, null, edges, 100, 3);
+    expect(result.pageInfo.hasNextPage).toEqual(true);
+  });
+
+  it('should set hasNextPage=false when raw page size is below limit', () => {
+    const edges = Array.from({ length: 7 }, (_, i) => ({ node: { id: i }, cursor: `c${i}` } as any));
+    const result = buildPaginationFromEdges(10, null, edges, 100, 0);
+    expect(result.pageInfo.hasNextPage).toEqual(false);
   });
 });


amiraifori:
Hi @maelv-filigran @dcode , i trust you are well, could you please review the PR and guide me

amiraifori:
@xfournet , thank you for self-requesting a review to my PR would go a long way getting a revieww

amiraifori:
Good day @xfournet , @marieflorescontact 
Still hoping this PR ges reviewed when you are ready.

opencti-platform/opencti-graphql/src/database/utils.ts
```diff
@@ -265,7 +265,29 @@ export const buildPaginationFromEdges = <T>(
...
+      && globalCount > pageSize
+      && searchAfter !== undefined
+      && searchAfter !== null) {
+      // We received fewer usable edges than the requested limit, but some hits
+      // were filtered out and the global count indicates there are still more
+      // results in the index. In this situation, we allow another round-trip
+      // so that subsequent pages can surface remaining matching entities.
+      hasNextPage = true;
+    }
+  }
```
xfournet: Hi @amiraifori, 
thanks for the PR and sorry for the delay.

I reviewed it, but I can't understand what is fixed. Do you have a repro case ?

I have also doubt on the `globalCount > pageSize` test since, except for the first page when the result set don't need pagination, this condition will be always true ?

You mention

> edges.length + filteredCount < limit
> (because many hits were filtered in this batch), but
> filterCount = elements.length - filteredElements.length

since `filteredElements = edges.length` it gives `edges.length + filteredCount = elements.length` which correspond to the raw result before post filtering, so we are testing `elements.lengtht < limit` which is what we intend to do.

At the end since:
* `pageSize >= limit` and `pageSize < limit && filteredCount > 0` are complementary
* we can consider that `globalCount > pageSize` is always true
* searchAfter is probably set when pageSize >= limit
then we can compact this test to `const hasNextPage = !!searchAfter`, which may be the solution but i would prefer to understand the problem precisely before moving on it.

amiraifori: Hi @xfournet, thanks for the review and the detailed feedback.

You’re right about the logic: given how `filteredCount` is computed.

I updated the implementation to make this explicit and simpler:
- define `pageSize = edges.length + filteredCount` (raw hits returned by ES for this request, before post-filtering)
- set `hasNextPage = pageSize >= limit`

This is the intended behavior: if Elasticsearch returned a full batch, there may still be results after the cursor, even if post-filtering reduced `edges.length`.

I also added unit coverage for the `filteredCount` scenario.

Repro scenario (conceptual): any query using `regardingOf` post-filtering where ES returns `limit` hits, but some are removed by the post-filter, can otherwise incorrectly mark `hasNextPage=false` and prevent reaching remaining valid hits beyond the cursor.

opencti-platform/opencti-graphql/src/database/utils.ts
```diff
@@ -261,11 +261,22 @@ export const buildPaginationFromEdges = <T>(
...
+  // which can incorrectly report no next page when a whole batch of hits is
+  // filtered out, even if more data exists after the current cursor.
+  let hasNextPage = false;
+  if (limit && limit > 0) {
+    const pageSize = edges.length + filteredCount;
+    // pageSize corresponds to the number of raw hits returned by the search engine
+    // (including hits later removed by post-filtering).
+    // If the engine returned a full batch, another page might exist.
+    hasNextPage = pageSize >= limit;
+  }
```
xfournet: ok this can be summarized to
```
const pageSize = edges.length + filteredCount;
const hasNextPage = pageSize > 0 && pageSize >= limit
```

the main difference is that we are testing the case pageSize is strictly greater than limit. Did you identify such cases ?

amiraifori: Hi @xfournet,

Agreed - I updated the code to the compact form:
const pageSize = edges.length + filteredCount;
const hasNextPage = pageSize > 0 && pageSize >= limit (with the existing limit>0 guard).

About pageSize > limit: in the main [elPaginate](cci:1://file:///c:/Users/brass/OneDrive/Desktop/Work/App/opencti/opencti-platform/opencti-graphql/src/database/engine.ts:3071:0-3127:2) path, ES returns at most `limit` hits so pageSize shouldn’t exceed limit.

However [buildPaginationFromEdges](cci:1://file:///c:/Users/brass/OneDrive/Desktop/Work/App/opencti/opencti-platform/opencti-graphql/src/database/utils.ts:258:0-289:2) is also used by [elConnection()](cci:1://file:///c:/Users/brass/OneDrive/Desktop/Work/App/opencti/opencti-platform/opencti-graphql/src/database/engine.ts:3188:0-3196:2) after [elRepaginate()](cci:1://file:///c:/Users/brass/OneDrive/Desktop/Work/App/opencti/opencti-platform/opencti-graphql/src/database/engine.ts:3133:0-3186:2) where `elements` is an accumulated list across multiple internal paginate calls (it’s not guaranteed to be capped to `first`). In that case `edges.length` can be > `limit`, so the `>=` keeps behavior correct for those callers too. I didn’t find a case where we need to test strictly `>` specifically; `>=` is sufficient and still covers the original bug (post-filtering reducing edges while raw batch reached the requested size).

amiraifori:
Hello @xfournet 
Hope you are doing alright, just a quick reminder I would appreciate it if my PR is reviewed.

amiraifori:
Hello @xfournet 
Hope you are doing alright, just a quick reminder I would appreciate it if my PR is reviewed.

xfournet:
Hi @amiraifori,

thanks for the update. I still need a way to reproduce the problem, so i can understand which condition is not correctly tested currently. Do you have a way to reproduce it ?

amiraifori:
> Hi @amiraifori,
> 
> thanks for the update. I still need a way to reproduce the problem, so i can understand which condition is not correctly tested currently. Do you have a way to reproduce it ?

Hi @xfournet,

I found a reliable repro: it happens on lists using the `regardingOf` filter, because the backend applies a post-filter after Elasticsearch.

Code path:
- [elPaginate()](cci:1://file:///c:/Users/brass/OneDrive/Desktop/Work/App/opencti/opencti-platform/opencti-graphql/src/database/engine.ts:3071:0-3127:2) (src/database/engine.ts) runs [buildRegardingOfFilter()](cci:1://file:///c:/Users/brass/OneDrive/Desktop/Work/App/opencti/opencti-platform/opencti-graphql/src/database/engine.ts:4273:0-4361:2) when filters contain `regardingOf` / `dynamicRegardingOf`
- It then removes some hits after ES (`filterCount = elements.length - filteredElements.length`)
- The previous `hasNextPage` logic relied on the returned edges length and could stop too early when many hits in a batch were filtered out, even though more hits exist after the cursor.

Repro (example using PIR knowledge entities):
1) Create a PIR and link >25 entities to it using `in-pir`.
2) Make part of those linked entities not visible to the current user (e.g., via markings/org restrictions) so that post-filtering removes some ES hits.
3) Go to PIR -> Knowledge -> Entities (this view explicitly sends `filters: { key: "regardingOf", values: [{key:"id", values:[pirId]}, {key:"relationship_type", values:["in-pir"]}] }`).
4) On master, the first page can return fewer edges than `first`, and `pageInfo.hasNextPage` can become false even though more valid results exist after the cursor.

The fix adjusts [buildPaginationFromEdges](cci:1://file:///c:/Users/brass/OneDrive/Desktop/Work/App/opencti/opencti-platform/opencti-graphql/src/database/utils.ts:258:0-289:2) to compute `hasNextPage` using the raw page size (returned edges + filteredCount), so we don’t prematurely stop pagination when post-filtering is active.

JeremyCloarec:
Hi @amiraifori,

Unfortunately, we're not reproducing the issue with the given repo case (test both in 6.9.8 and 6.9.12).
Which version did you reproduce the bug on? Also, would it be possible to create a repro case with an integration test to have a consistent repro?

PR #13399, opened 2025-12-01
[frontend] convert playbook.js to ts #12362

MTorbay-Filigran:
### Proposed changes

* convert playbook.js to ts
*

### Related issues

*
*

### Checklist


opencti-platform/opencti-front/src/private/components/data/playbooks/playbookFlow/PlaybookFlow.tsx
lndrtrbn, line 1: I think you can still type things to avoid using hard cast

opencti-platform/opencti-front/src/private/components/data/playbooks/utils/playbook.ts
```diff
@@ -0,0 +1,179 @@
+import { Edge } from 'reactflow';
+interface PortDefinition {
+  id: string
+  type: string
+}
+interface PlaybookComponent {
```
lndrtrbn: The fact that you need to set readonly on each attribute makes me think that data came from a query and so, can you reuse existing generated type?

SarahBocognano: Agree with you on this. Probably related to PlaybookFlow_playbookComponents$data

opencti-platform/opencti-front/src/private/components/data/playbooks/utils/playbook.ts
```diff
@@ -0,0 +1,179 @@
...
+  id: string,
+  name: string,
+  position: { x: number, y: number },
+  component_id: string,
+  configuration: string // json
+}
+
+export const computeNodes = (
+  playbookNodes: ComputeNodeDefinition[],
+  playbookComponents:readonly (PlaybookComponent | null | undefined)[],
```
SarahBocognano: maybe we can avoid the readonly here as well with the right generated type from query

opencti-platform/opencti-front/src/private/components/data/playbooks/playbookFlow/PlaybookFlow.tsx
MTorbay-Filigran: I would have to type useManipulateComponents, I can do it in another PR

lndrtrbn (approved):

PR #13405, opened 2025-12-01
[backend] standalone connector migration composer (#13202)

### LINKED ISSUE #13202: Add API endpoint to migrate standalone connectors to composer-managed deployment
## Use case

OpenCTI previously supported connector deployment in "standalone" mode. Recently, we introduced the ability to manage connector deployment through our composer via the UI interface. As a result, we need to migrate existing standalone connectors to composer-managed deployments.

## Current Workaround

Currently, there is no automated way to migrate standalone connectors to the new composer-managed deployment model. Manual migration is error-prone and risks losing critical connector state and configuration.

## Proposed Solution

Introduce a new API endpoint in OpenCTI to facilitate the migration of standalone connectors to composer-managed deployments. This API should handle the redeployment process while preserving essential connector information.

The API must transfer and preserve the following information during redeployment:

- Connector ID: Maintain the unique identifier of the connector
- Associated User: Preserve the user account linked to the deployment
- Connector State: Retain the current ingestion state to ensure continuity of data processing

esrevi:
addresses #13202 

### Proposed changes

* Add a query to assess the migration, checks if standalone connector is elligible for migration to composer
```
query assessConnectorMigration {
  connectorMigrationAssessment(
    connectorId: "nist-cve-id-4",  
    contractSlug: "nist-nvd-cve", 
    configuration: [
      { key: "CVE_BASE_URL", value: "http://test" },
      { key: "CONNECTOR_LOG_LEVEL", value: "error" }
    ]
  ) {
    connector_id
  	contract_slug
    mapped {
      key
      value
      type
      required
    }
    missing {
      key
      type
      description
      enum
      default
    }
  }
}
```

* Add a mutation to migrate the standalone connector to managed connector

Here's an example of the migration mutation: 
```
mutation migrate {
  connectorMigrateToManaged(input: {
    connectorId: "mitre-standalone-id14",
    contractSlug: "mitre",
    resetConnectorState: false,
    convertUserToServiceAccount: true,
    configuration: [
    	{
      	key: "CONNECTOR_LOG_LEVEL", value: "debug"
      }
    ]
  }){
    connector {
      id
      name
      standard_id
      manager_contract_configuration {key value}
      connector_user {
        user_service_account
      }
    }   
  }
}
```
- Depending on the connector catalog's schema, some keys could be required such as xxx_api_key, if not set in the mutation, the function will fail. 
- Configuration could also be empty, if no required keys are found
- The config keys not override uses the keys defined from the connector's catalog schema properties, it will also use the default set in the manifest

### Related issues

* #13202 

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/config/schema/opencti.graphql
+++ b/opencti-platform/opencti-graphql/config/schema/opencti.graphql
@@ -2198,6 +2198,77 @@ enum ConnectorPriorityGroup {
   DEFAULT
 }
 
+# CONNECTOR MIGRATION INPUT TYPES
+input ConnectorMigrationAssessmentInput {
+  connectorId: ID!
+  contractSlug: String!
+  configuration: [ContractConfigInput!]
+}
+
+input MigrateConnectorToManagedInput {
+  connectorId: ID!
+  contractSlug: String!
+  configuration: [ContractConfigInput!]
+  preserveState: Boolean
+}
+
+# CONNECTOR MIGRATION ASSESSMENT TYPES
+type MigrationAssessmentSummary {
+  total_source_keys: Int!
+  mapped_keys: Int!
+  ignored_keys: Int!
+  missing_mandatory_keys: Int!
+  can_migrate: Boolean!
+  assessment_date: DateTime!
+}
+
+type MappedConfigKey {
+  key: String!
+  sourceKey: String!
+  value: String!
+  type: String!
+  required: Boolean!
+}
+
+type IgnoredConfigKey {
+  key: String!
+  value: String!
+  reason: String!
+}
+
+type MissingConfigKey {
+  key: String!
+  type: String!
+  description: String!
+  format: String
+  enum: [String!]
+}
+
+type MigrationAssessment {
+  connector_id: ID!
+  connector_name: String!
+  contract_slug: String!
+  contract_title: String!
+  summary: MigrationAssessmentSummary!
+  mapped: [MappedConfigKey!]!
+  ignored: [IgnoredConfigKey!]!
+  missing: [MissingConfigKey!]!
+}
+
+# CONNECTOR MIGRATION RESULT TYPES
+type MigrationWarning {
+  field: String
+  message: String!
+}
+
+type MigrateConnectorToManagedResult {
+  connector: Connector!
+  warnings: [MigrationWarning!]!
+  state_preserved: Boolean!
+  dryRun: Boolean!
+  skipped: Boolean
+}
+
 type Connector implements BasicObject & InternalObject {
   id: ID! # internal_id
   standard_id: String!
@@ -13699,6 +13770,7 @@ type Query {
   connectorsForImport: [Connector] @auth(for: [KNOWLEDGE])
   connectorsForAnalysis: [Connector] @auth(for: [KNOWLEDGE])
   connectorsForNotification: [Connector] @auth(for: [SETTINGS_SETACCESSES, SETTINGS_SETCUSTOMIZATION])
+  connectorMigrationAssessment(connectorId: ID!, contractSlug: String!, configuration: [ContractConfigInput!]): MigrationAssessment! @auth(for: [MODULES])
   work(id: ID!): Work @auth(for: [MODULES])
   isWorkAlive(id: ID!): Boolean @auth(for: [MODULES])
   works(
@@ -15369,6 +15441,7 @@ type Mutation {
   updateConnectorLogs(input: LogsConnectorStatusInput!): ID! @auth(for: [CONNECTORAPI])
   updateConnectorHealth(input: HealthConnectorStatusInput!): ID! @auth(for: [CONNECTORAPI])
   updateConnectorTrigger(id: ID!, input: [EditInput]!): Connector @auth(for: [MODULES_MODMANAGE])
+  connectorMigrateToManaged(input: MigrateConnectorToManagedInput!, dryRun: Boolean): MigrateConnectorToManagedResult! @auth(for: [CONNECTORAPI])
   feedAdd(input: FeedAddInput!): Feed @auth(for: [TAXIIAPI_SETCOLLECTIONS])
   feedDelete(id: ID!): ID! @auth(for: [TAXIIAPI_SETCOLLECTIONS])
   feedEdit(id: ID!, input: FeedAddInput!): Feed! @auth(for: [TAXIIAPI_SETCOLLECTIONS])
--- a/opencti-platform/opencti-graphql/config/schema/opencti.graphql
+++ b/opencti-platform/opencti-graphql/config/schema/opencti.graphql
@@ -13757,7 +13757,7 @@ type Query {
   connectorsForImport: [Connector] @auth(for: [KNOWLEDGE])
   connectorsForAnalysis: [Connector] @auth(for: [KNOWLEDGE])
   connectorsForNotification: [Connector] @auth(for: [SETTINGS_SETACCESSES, SETTINGS_SETCUSTOMIZATION])
-  connectorMigrationAssessment(connectorId: ID!, containerImage: String!, configuration: [ContractConfigInput!]): MigrationAssessment! @auth(for: [MODULES])
+  connectorMigrationAssessment(connectorId: ID!, containerImage: String!, configuration: [ContractConfigInput!]): MigrationAssessment! @auth(for: [MODULES_MODMANAGE])
   work(id: ID!): Work @auth(for: [MODULES])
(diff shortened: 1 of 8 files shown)


sbocahu:
Instead of contractSlug, could we use the connector image name (for example, opencti/connector-cve) ?

esrevi:
sure, I can replace the contractSlug and use instead the container_image

opencti-platform/opencti-graphql/config/schema/opencti.graphql
```diff
@@ -2198,6 +2198,67 @@ enum ConnectorPriorityGroup {
...
+  connector_id: ID!
+  connector_name: String!
+  contract_image: String!
+  summary: MigrationAssessmentSummary!
+  mapped: [MappedConfigKey!]!
+  ignored: [IgnoredConfigKey!]!
+  missing: [MissingConfigKey!]!
+}
+type MigrateConnectorToManagedResult {
+  connector: Connector!
```
JeremyCloarec: connector type should be typed as ManagedConnector! instead I think.
Also, was there a reason to create the MigrateConnectorToManagedResult and not return the connector directly in the return of connectorMigrateToManaged?

opencti-platform/opencti-graphql/src/resolvers/connector.js
```diff
@@ -186,6 +190,19 @@ const connectorResolvers = {
...
+      return migrateConnectorToManaged(
+        context,
+        context.user,
+        connectorId,
+        containerImage,
+        configuration,
+        convertUserToServiceAccount,
+        resetConnectorState,
+      );
+    },
```
JeremyCloarec, line 205: dryRun parameter from the API is never used, should it be removed from the API or should it be implemented in migrateConnectorToManaged? I don't know what is expected

opencti-platform/opencti-graphql/src/resolvers/connector.js
```diff
@@ -186,6 +190,19 @@ const connectorResolvers = {
     synchronizerStart: (_, { id }, context) => patchSync(context, context.user, id, { running: true }),
     synchronizerStop: (_, { id }, context) => patchSync(context, context.user, id, { running: false }),
     synchronizerTest: (_, { input }, context) => testSync(context, context.user, input),
+
+    connectorMigrateToManaged: (_, { input }, context) => {
+      const { connectorId, containerImage, configuration, resetConnectorState, convertUserToServiceAccount = false } = input;
```
JeremyCloarec: convertUserToServiceAccount is also defaulted in migrateConnectorToManaged, the defaulting here could be removed I think

opencti-platform/opencti-graphql/src/domain/connector-migration.ts
```diff
@@ -0,0 +1,408 @@
...
+  reason: string;
+};
+
+export const getConnector = async (context: AuthContext, user: AuthUser, id: string) => {
+  const connectorFound = connector(context, user, id);
+
+  if (!connectorFound) {
+    throw FunctionalError('No connector found with the specified ID', { id });
+  }
+};
```
JeremyCloarec: it looks like this method is no longer used

opencti-platform/opencti-graphql/src/domain/connector-migration.ts
```diff
@@ -0,0 +1,408 @@
...
+
+  let contract;
+  try {
+    contract = JSON.parse(contractData.contract);
+  } catch {
+    throw FunctionalError('Cannot parse contract found');
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  const { logo, description, short_description, ...contractDefinition } = contract;
```
JeremyCloarec: I don't get why we need this, can't we use contract directly instead of contractDefinition?

opencti-platform/opencti-graphql/src/domain/connector-migration.ts
```diff
@@ -0,0 +1,408 @@
...
+  } else {
+    Object.entries(configuration).forEach(([key, value]) => {
+      configMap.set(key.toUpperCase(), value);
+    });
+  }
+
+  return configMap;
+};
+
+const autoMapConnectorFields = (connectorOjb: any): Map<string, string> => {
```
JeremyCloarec: small detail, but I think connectorObj should be able to be typed as Connector from generated/graphql here

opencti-platform/opencti-graphql/src/domain/connector-migration.ts
```diff
@@ -0,0 +1,408 @@
...
+  const autoMapped = new Map<string, string>();
+
+  if (connectorOjb.name) {
+    autoMapped.set('CONNECTOR_NAME', connectorOjb.name);
+  }
+
+  if (connectorOjb.connector_scope) {
+    const scopeValue = Array.isArray(connectorOjb.connector_scope)
+      ? connectorOjb.connector_scope.join(',')
+      : connectorOjb.connector_scope;
```
JeremyCloarec: I think connector_scope is always an array here no?

opencti-platform/opencti-graphql/src/domain/connector-migration.ts
```diff
@@ -0,0 +1,408 @@
...
+
+export const getConnector = async (context: AuthContext, user: AuthUser, id: string) => {
+  const connectorFound = connector(context, user, id);
+
+  if (!connectorFound) {
+    throw FunctionalError('No connector found with the specified ID', { id });
+  }
+};
+
+const buildConfigMap = (configuration: ConfigInput[] | Record<string, string>): Map<string, string> => {
```
JeremyCloarec: You should be able to only type configuration as ConfigInput[]: the Record type is only used in assessConnectorMigration for the default value of configuration. The default value of configuration should be changed to [] there

opencti-platform/opencti-graphql/src/domain/connector-migration.ts
```diff
@@ -0,0 +1,408 @@
...
+  }
+
+  return autoMapped;
+};
+
+const categorizeKeys = (
+  schemaProperties: any,
+  requiredKeys: string[],
+  configMap: Map<string, string>,
+  autoMappedKeys: Map<string, string>,
```
JeremyCloarec: since configMap already contains the autoMappedKeys, can't the autoMappedKeys parameter be removed?

opencti-platform/opencti-graphql/src/resolvers/connector.js
esrevi: yes, i remove that param, since we can assess with the other resolver 👍

opencti-platform/opencti-graphql/src/domain/connector-migration.ts
esrevi: can't remember the reason today, so i removed that spread indeed :p

opencti-platform/opencti-graphql/src/domain/connector-migration.ts
esrevi: nice catch, thx !

opencti-platform/opencti-graphql/src/domain/connector-migration.ts
```diff
@@ -0,0 +1,408 @@
...
+    mapped,
+    ignored,
+    missing,
+    message: configuration === null
+      ? 'No configuration provided. You must provide configuration manually with exact schema keys.'
+      : null,
+  };
+};
+
+const hasServiceAccountProperty = (user: any): user is AuthUser & { service_account?: boolean } => {
```
JeremyCloarec: I don't get the return type, I think it should be a bool instead

opencti-platform/opencti-graphql/src/domain/connector-migration.ts
```diff
@@ -0,0 +1,408 @@
...
+  const filteredConfigurations = configurations.filter(
+    (config) => !RUNTIME_PROVIDED_FIELDS.includes(config.key),
+  );
+
+  const existingUser = await storeLoadById(
+    context,
+    user,
+    existingConnector.connector_user_id,
+    ENTITY_TYPE_USER,
+  );
```
JeremyCloarec: user could be retrieved in cache instead to save a call to elastic

opencti-platform/opencti-graphql/src/domain/connector-migration.ts
```diff
@@ -0,0 +1,408 @@
...
+
+  // If existing user is not a service account, transform it to service account
+  if (
+    hasServiceAccountProperty(existingUser)
```
JeremyCloarec: i don't get what this condition is checking exactly? shouldn't the isServiceAccountUser condition be enough?

JeremyCloarec (approved):

Issue #13397, opened 2025-12-01
[ci] GitHub Action: Use client-python version from the branch instead of released one


efaure:
## Description

When a PR modify client-python, api tests are not run using the modified version but the client-python release defined in `opencti-platform/opencti-graphql/src/python/requirements.txt`
Example : https://github.com/OpenCTI-Platform/opencti/actions/runs/19759554020/job/56618249921?pr=11628. 

## Environment

GitHub action 

## Expected Output

All test from github action should be run with the client-python sources from the branch  (always or only when it's updated ? )

PR #13414, opened 2025-12-01
[ci] GitHub Action: Use client-python version from the branch instead of released one

efaure:
### Proposed changes

* Use a dedicated Dockerfile for CI in order to build and use local client python version
* Also use this image to launch test

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/13397
*

### DIFF
--- a/.github/workflows/ci-docker-build.yml
+++ b/.github/workflows/ci-docker-build.yml
@@ -21,6 +21,16 @@ on:
         type: string
         required: false
         default: opencti-platform
+      dockerfile_worker:
+        description: "Dockerfile"
+        type: string
+        required: false
+        default: Dockerfile_featurebranch
+      build_context_worker:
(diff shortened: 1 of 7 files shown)


efaure:
@aHenryJard  In order to test this change, I made a commit with a change of log in client python and confirmed that I see the updated log in api integration test and workers log.
I reverted it in the last commit https://github.com/OpenCTI-Platform/opencti/pull/13414/commits/b3d57d9e2568615552754a91a45e9c102b151fee

Do you see a was have a better testing of the usage of client python ? 

Maybe we can also merge it "as is" as it's not harmful, and check that tests in  https://github.com/OpenCTI-Platform/opencti/pull/11628 are successful after rebasing

aHenryJard (approved):
Looks good to me

Issue #13320, opened 2025-11-25
Import Form does not prevent submitting badly formatted values


Jermain-N:
## Description

The input form does not generate an error when a wrongly formatted value is entered for an observablme (e.g entering a text-based string into a IP observable does not generate a visible error message, nor does it prevent a user from submitting the form).

## Environment

1. OS (where OpenCTI server runs):
2. OpenCTI version: 6.8.13
3. OpenCTI client: frontend
4. Other environment details:  https://testing.octi.staging.filigran.io

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Create a form with an IPv4 entry
2. Import using the form
3. Enter a text value (e.g. "test")
4. Submit the form

## Expected Output

Expecting an error message to prevent the form from being submitted.

## Actual Output

The form accepts the bad values and goes through the rest of the ingestion process:

<img width="1875" height="858" alt="Image" src="https://github.com/user-attachments/assets/48a87de9-f3cb-498a-82db-d05f04ab5293" />

Then loads the connectors page...?

<img width="1895" height="886" alt="Image" src="https://github.com/user-attachments/assets/b785d4fe-f69e-4a30-bd49-e9f6a06dad2a" />

SarahBocognano:
Can you upload a video of the use case ? I don't understand how to reproduce

PR #13421, opened 2025-12-01
[backend] Import Form does not prevent submitting badly formatted values (#13320)

SarahBocognano:
### Proposed changes

* Add a backend check before creation for observables entities as they are the one with syntax complexity

To test this feature we need to : 
- create a form using form intake (data -> ingestion -> form intakes)
- create an ipv4 form (a simple one with only value is good enough)
- now import data with header icon
- select "import using a form"
- select the one previously created for ipv4
- try to enter a false value such as "plop" -> you should have an error in display and the creation of the entity is stoped 

** please tell me if you think of any other special field to check for this case **

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/13320

### DIFF
--- a/opencti-platform/opencti-graphql/src/modules/form/form-domain.ts
+++ b/opencti-platform/opencti-graphql/src/modules/form/form-domain.ts
@@ -406,20 +406,38 @@ export const formSubmit = async (
   } catch (error) {
     throw FunctionalError('Cannot read values', { error });
   }
+  const schema: FormSchemaDefinition = JSON.parse(form.form_schema);
 
-  if (isStixCyberObservable(form.main_entity_type)) {
-    const observableSyntaxResult = checkObservableSyntax(form.main_entity_type, values);
-    if (observableSyntaxResult !== true) {
-      throw FunctionalError('Observable is not correctly formatted',
-        {
-          type: form.main_entity_type,
-          input: values,
-          doc_code: 'INCORRECT_OBSERVABLE_FORMAT'
-        });
-    }
+  // check if any addionalEntities are present in the form
+  const additionalEntities = schema.additionalEntities ?? [];
+
+  // we need to check here the syntax of observables to be sure the format is respected
+  const observableInputs = [];
+  observableInputs.push({
+    type: form.main_entity_type,
+    value: values.value,
+  });
+  for (const additionalEntity of additionalEntities) {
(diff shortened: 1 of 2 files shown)


opencti-platform/opencti-graphql/src/modules/form/form-domain.ts
```diff
@@ -405,6 +407,18 @@ export const formSubmit = async (
     throw FunctionalError('Cannot read values', { error });
   }
 
+  if (isStixCyberObservable(form.main_entity_type)) {
```
JeremyCloarec: This only checks the main entity of the form, I think we probably also want to check all additional entities

SarahBocognano: Indeed haha. I knew I needed an extra pair of eyes

opencti-platform/opencti-graphql/src/modules/form/form-domain.ts
```diff
@@ -404,8 +406,38 @@ export const formSubmit = async (
...
   const schema: FormSchemaDefinition = JSON.parse(form.form_schema);
+
+  // check if any addionalEntities are present in the form
+  const additionalEntities = schema.additionalEntities ?? [];
+
+  // we need to check here the syntax of observables to be sure the format is respected
+  const observableInputs = [];
+  observableInputs.push({
+    type: form.main_entity_type,
+    value: values.value,
```
JeremyCloarec: depending on the observable type, the checkObservableSyntax doesn't only check the value of the observable, it also checks the number/hashes.
I think that instead of only getting the value from values, you can spread the whole values into the observable

opencti-platform/opencti-graphql/src/modules/form/form-domain.ts
```diff
@@ -404,8 +406,38 @@ export const formSubmit = async (
...
+  });
+  for (const additionalEntity of additionalEntities) {
+    observableInputs.push({
+      type: additionalEntity.entityType,
+      value: values[`additional_${additionalEntity.id}`].value, // stock this way in DB
```
JeremyCloarec: Same as above, you can take the whole object in values[`additional_${additionalEntity.id}`] in the input, not only the value

JeremyCloarec (approved):

PR #13444, opened 2025-12-03
[worker/client] add ability to specify a provider in API client user agent

### LINKED ISSUE #13625: Add ability to specify a provider in API client user agent and specify one in worker
## Use case

Add ability to specify a provider in API client user agent and specify one in worker

JeremyCloarec:
### Proposed changes

*  add ability to specify a provider in API client user agent
* worker now registers in API client as "worker" provider. It's user agent will now be "pycti/worker/${pycti_version}"

### Related issues

*
*

### Checklist

### DIFF
--- a/client-python/pycti/api/opencti_api_client.py
+++ b/client-python/pycti/api/opencti_api_client.py
@@ -9,6 +9,7 @@ import shutil
 import signal
 import tempfile
 import threading
+import re
 from typing import Dict, Tuple, Union
 
 import magic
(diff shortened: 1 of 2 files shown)


client-python/pycti/api/opencti_api_client.py
```diff
@@ -91,9 +91,13 @@
 _PROXY_SIGNAL_HANDLERS_REGISTERED = False
 
 
-def build_request_headers(token: str, custom_headers: str, app_logger):
+def build_request_headers(token: str, custom_headers: str, app_logger, provider: str):
```
xfournet, line 95: we should rather cumulate the information in the user agent. eg
* no additional parameter -> `pycti/6.8.15` (as today)
* with additional parameter `client/1.2.3` ->   `pycti/6.8.15 client/1.2.3`

xfournet: we need to either validate the additional parameter or split in two part (name and version, but still need to be validated ?)

JeremyCloarec: I added validation, worker currently doesn't have any version number associated though, so it defaults to a hardcoded value of 1.0

opencti-worker/src/worker.py
```diff
@@ -186,7 +186,7 @@ def __post_init__(self) -> None:
             perform_health_check=False,  # No need to prevent worker start if API is not available yet
             custom_headers=self.opencti_api_custom_headers,
             requests_timeout=self.opencti_api_requests_timeout,
-            provider="worker/1.0"
+            provider="worker/1.0",
```
xfournet: could we use "from pycti import __version__" like in listen_handler.py

xfournet (approved):

PR #13446, opened 2025-12-03
[worker/client] add ability to customize API requests timeout in OpenCTIApiClient init and in worker config

JeremyCloarec:
### Proposed changes

* Some heavy requests can sometime take longer to process than the default 5minutes timeout in the request timeout (rule apply operations for example). There is currently no way to configure and change this default 5 minutes timeout. The goal of the change here is to make this timeout configurable in the worker
*

### Related issues

*
*

### Checklist


client-python/pycti/api/opencti_api_client.py
xfournet, line 613: need to be changed too

xfournet (approved):

Issue #10981, opened 2025-05-13
Authorize members for Draft


nino-filigran:
## Use case

Currently, every user can be able to approve/edit/view any draft created in the platform. This cause some issues, since drafts give more manipulation capabilities than workbench.

## Proposed Solution

The goal of the feature is to be able to apply the principle of authorized members on draft: 
- By default, only the creator of the draft should have access to the draft
- This could be managed as a configuration (like any entity) to be able to list by default authorized members. 
- When adding the approval workflow, we need a way to be able to add approvers automatically into Authotrize members.

PR #13459, opened 2025-12-03
[frontend] fix check bypass include capa with bypassref and bypassfields (#10981)

frapuks:
### Proposed changes

* don't check if the name of the user capa `includes` the name of the capa checked for the `bypass` capa, because the string 'BYPASS' is includes in 'KNOWLEDGE_KNUPDATE_KNBYPASSREFERENCE', and the 'BYPASS' check is already executed before this condition.
* Can be tested by check if a user with 'KNOWLEDGE_KNUPDATE_KNBYPASSREFERENCE' can see the entry `Data/restriction` in the left menu

### Related issues

* #10981

### Checklist


frapuks:
fix included into chunk 2

Issue #11719, opened 2025-07-18
Edit Severity in playbooks on applies on Severity of Cases


nino-filigran:
## Description

Edit Severity in playbooks on applies on Severity of Case incident but not on incidents. 
## Environment

Testing 

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Create a playbook that listen to incident create
2. Add a step manipulate knowledge, select "Add" or "Replace", with the field "Severity" and select a severity
3. Add send for ingestion 

## Expected Output

Incident is updated.
We should have both values (incident & case incident severity as values selectable too).

## Actual Output

Incident is not updated. 

## Additional information

The reason is that only the case_incident severity is retrieved and applied.

PR #13462, opened 2025-12-03
[backend/frontend] Edit Severity in playbooks on applies on Severity of Cases (#11719)

SarahBocognano:
### Proposed changes

* Revamp the openVocabField to allow to have several types of vocab categories for 'cases_severity_ov' and 'incident_severity_ov' to be taken into account in a generic severity action field on playbooks

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/11719

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/common/form/OpenVocabField.tsx
+++ b/opencti-platform/opencti-front/src/private/components/common/form/OpenVocabField.tsx
@@ -124,7 +124,7 @@ const OpenVocabFieldComponent: FunctionComponent<OpenVocabProps> = ({
         name={name}
         required={required}
         onFocus={onFocus}
-        onChange={(n: string, v: VocabFieldOption & VocabFieldOption[]) => {
+        onChange={(n: string, v: VocabFieldOption | VocabFieldOption[]) => {
           internalOnChange?.(n, v);
           internalOnSubmit?.(n, v);
         }}
--- a/opencti-platform/opencti-front/src/private/components/common/form/OpenVocabField.tsx
+++ b/opencti-platform/opencti-front/src/private/components/common/form/OpenVocabField.tsx
@@ -11,13 +11,18 @@ import useQueryLoading from '../../../../utils/hooks/useQueryLoading';
 import { RenderOption } from '../../../../components/list_lines';
 import { FieldOption } from '../../../../utils/field';
 
+interface VocabFieldOption extends FieldOption {
+  description?: string;
+  category: string;
+}
+
 interface OpenVocabProps {
-  type: string;
+  type: string | string[];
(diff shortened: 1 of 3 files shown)


lndrtrbn (approved):

opencti-platform/opencti-front/src/private/components/common/form/OpenVocabField.tsx
```diff
@@ -75,23 +82,25 @@ Omit<OpenVocabProps, 'type'>
...
       value,
       label: value,
       description,
-    }));
-  let internalOnChange: ((n: string, v: FieldOption | FieldOption[]) => void) | undefined;
-  let internalOnSubmit: ((n: string, v: FieldOption | FieldOption[]) => void) | undefined;
+      category: category.key,
+    }))
+    .sort((a, b) => a.category.localeCompare(b.category));
+  let internalOnChange: ((n: string, v: VocabFieldOption | VocabFieldOption[]) => void) | undefined;
```
esrevi, line 92: I think we could simplify this code to make it easier to maintain and understand :) 

We could create a function such as to factorize the transformation logic
```
const extractValues = (v: VocabFieldOption | VocabFieldOption[]) => {
  return Array.isArray(v) 
    ? v.map((item) => item?.value ?? item)
    : v?.value ?? v;
```

Then call directly in 
```
onChange={(n: string, v: VocabFieldOption | VocabFieldOption[]) => {
   const values = extractValues(v)
   onChange?.(n, values)
   onSubmit?.(n, values)
  }}
```

opencti-platform/opencti-front/src/private/components/common/form/OpenVocabField.tsx
```diff
@@ -115,7 +124,7 @@ Omit<OpenVocabProps, 'type'>
         name={name}
         required={required}
         onFocus={onFocus}
-        onChange={(n: string, v: FieldOption & FieldOption[]) => {
+        onChange={(n: string, v: VocabFieldOption & VocabFieldOption[]) => {
```
esrevi: should this be VocabFieldOption | VocabFieldOption[] instead ?

opencti-platform/opencti-front/src/private/components/common/form/OpenVocabField.tsx
lndrtrbn: The component needs a global refactoring, there are multiple topics that need to be addressed in this file. I think it should be done in a separate PR

esrevi: ok for me for doing that in another PR ^^

esrevi (approved):

Issue #13347, opened 2025-11-26
Prevent leaks of Users, Organizations and Groups


nino-filigran:
## Use case

There are various places in the app where users can list Users, Organizations and Groups : filters, assignee, participants, authors, access restrictions…
Depending on your business use-case, you might want to control precisely which of these entities are visible or not for the end users.
In some cases, we need strict control : users from an organization should never be able to see Users from another organization.
Our problem is that today, the behavior is 1/ inconsistent across the app, and 2/ cannot prevent leaks at all.

## Proposed Solution
Make the app behavior consistent: all places where we need to list “Members” (i.e. users, orga and groups), we should tie to uder paticipation in the organisation (if a user participate in a org, user can see members of the org). There are some variations to this behavior, but overal this would be the new behavior.

- Users will only see users pertaining to their own orgs.
    - In org segregation mode, this will be enforced and cannot be modified.
    - In non org segregation, this will be applied by default at migration
        - but admin can disable the functionality to allow all users to see all users regardless of participation to an org.
- In details, it means that you can only see users attached to your organisation. You cannot see users of other childs organizations that participate in the same parent organization than you. Visibility is based on “attachement” to an org. However, If a user is participating to a parent org (not via inferred relation), then this user must be visible to child org users.

## If the feature request is approved, would you be willing to submit a PR?

Yes / No (Help can be provided if you need assistance submitting a PR)

PR #13478, opened 2025-12-04
[backend] Support inferred subfilter in regardingOf filter (#13347)

Archidoit:
### Proposed changes
- Support is_inferred subfilter in regardingOf filter in backend

is is_inferred subfilter is specified to true/false, look only for the inferred/not inferred relationships when doing a regardingOf filter

Example: a filter to fetch the entitites having a not inferred 'targets' relationship : 

{"key": "regardingOf",
      "values": [
        { "key": "relationship_type", "values": ["targets"] },
        { "key": "is_inferred", "values": ["false"]}
        ]
}

- rename regardingOf subfilter keys constants

- create separated function for adaptFilterToRegardingOfFilterKey in engine.ts

### Related issues
Necessary preparation for:
#13347


opencti-platform/opencti-graphql/tests/utils/syncCountHelper.ts
```diff
@@ -26,7 +26,7 @@ testCreatedCounter['external-reference'] = 17;
 testCreatedCounter.feedback = 1;
 testCreatedCounter.file = 10;
 testCreatedCounter.grouping = 1;
-testCreatedCounter.identity = 40;
+testCreatedCounter.identity = 43;
```
Archidoit, line 29: the 3 created organizations in users-visibility-test.ts

Gwendoline-FAVRE-FELIX (approved):

PR #13487, opened 2025-12-04
[backend] Fix RSS parsing crash when items lack title elements (#10830)

### LINKED ISSUE #10830: Issues parsing RSS feed when <ITEM> tags contain no <Title> element
## Description
When setting up an RSS feed using: 
https://infosec.exchange/users/monitorsg.rss
Nothing is ingested into the platform, no activity shown for this feed.

## Environment
1. SaaS 6.6.7

1. Set up a RSS feed using above URL

## Expected Output
RSS feed to be ingested

## Actual Output
Nothing is ingested

## Additional information
If the title is not provided in each item element, maybe, as discussed internally, the title tag could be populated from the link tag

maelv-filigran:
### Proposed changes

* Fixed RSS feed parser crash when `<title>` element is missing from RSS/Atom items
* Added fallback logic: uses link URL when title is absent, or "Untitled" as last resort
* Updated test expectations to reflect that all 44 items are now correctly processed (previously 43 due to crash)

### Related issues

Fixes #10830

### Checklist

### Further comments

**Root Cause:**
The RSS parser was directly accessing `entry.title._` and `item.title._` without checking if the `title` property exists. When RSS items lack a `<title>` tag entirely, the XML parser returns `undefined`, causing a `TypeError: Cannot read property '_' of undefined`. This crash prevented the entire feed from being processed.

**Solution:**
Implemented safe property access using optional chaining (`?.`) with a fallback cascade:
1. Use the title if present
2. Fall back to the link URL if title is missing (as suggested in the issue)
3. Use "Untitled" as final fallback

**Testing:**
The existing test data (`rss-wago.xml`) already contained an item without a title (line 148), which was causing the test to expect 43 items instead of 44. The fix now correctly processes all 44 items, including the one without a title.

### DIFF
--- a/opencti-platform/opencti-graphql/src/manager/ingestionManager.ts
+++ b/opencti-platform/opencti-graphql/src/manager/ingestionManager.ts
@@ -173,14 +173,15 @@ interface DataItem {
 
 const rssItemV1Convert = (turndownService: TurndownService, feed: RssElement, entry: RssItem): DataItem => {
   const { updated } = feed;
-  // Safe access to title with fallback to link if title is missing
-  const link = isNotEmptyField(entry.link) ? (entry.link as { href: string }).href?.trim() : '';
+  // Safe access to link with fallback to empty string
+  const link = entry.link?.href?.trim() ?? '';
+  // Use link as fallback if title is missing, or 'Untitled' as last resort
   const title = entry.title?._ || link || 'Untitled';
   
   return {
     title,
     description: turndownService.turndown(entry.summary?._ ?? ''),
-    link,
+    link: link || undefined,
     content: turndownService.turndown(entry.content?._ ?? ''),
     labels: [], // No label in rss v1
     pubDate: utcDate(sanitizeForMomentParsing(entry.updated?._ ?? updated?._ ?? FROM_START_STR)),
@@ -189,14 +190,15 @@ const rssItemV1Convert = (turndownService: TurndownService, feed: RssElement, en
 
 const rssItemV2Convert = (turndownService: TurndownService, channel: RssElement, item: RssItem): DataItem => {
   const { pubDate } = channel;
-  // Safe access to title with fallback to link if title is missing
-  const link = isNotEmptyField(item.link) ? ((item.link as { _: string })._ ?? '').trim() : '';
+  // Safe access to link with fallback to empty string
+  const link = item.link?._?.trim() ?? '';
+  // Use link as fallback if title is missing, or 'Untitled' as last resort
   const title = item.title?._ || link || 'Untitled';
   
   return {
     title,
     description: turndownService.turndown(item.description?._ ?? ''),
-    link,
+    link: link || undefined,
     content: turndownService.turndown(item['content:encoded']?._ ?? item.content?._ ?? ''),
     labels: R.uniq(asArray(item.category).filter((c) => isNotEmptyField(c)).map((c) => (c as { _: string })._.trim())),
     pubDate: utcDate(sanitizeForMomentParsing(item.pubDate?._ ?? pubDate?._ ?? FROM_START_STR)),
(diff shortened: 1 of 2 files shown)


opencti-platform/opencti-graphql/src/manager/ingestionManager.ts
```diff
@@ -185,10 +189,14 @@ const rssItemV1Convert = (turndownService: TurndownService, feed: RssElement, en
 
 const rssItemV2Convert = (turndownService: TurndownService, channel: RssElement, item: RssItem): DataItem => {
   const { pubDate } = channel;
+  // Safe access to title with fallback to link if title is missing
+  const link = isNotEmptyField(item.link) ? ((item.link as { _: string })._ ?? '').trim() : '';
+  const title = item.title?._ || link || 'Untitled';
```
OctaveLaventure, line 196: **1. Translation**
I guess it would make sense to translate `'Untitled'`, what do you think?

opencti-platform/opencti-graphql/src/manager/ingestionManager.ts
```diff
@@ -173,10 +173,14 @@ interface DataItem {
 
 const rssItemV1Convert = (turndownService: TurndownService, feed: RssElement, entry: RssItem): DataItem => {
   const { updated } = feed;
+  // Safe access to title with fallback to link if title is missing
+  const link = isNotEmptyField(entry.link) ? (entry.link as { href: string }).href?.trim() : '';
```
OctaveLaventure: **2. Typing**
- I think the cast is unnecessary, or if we really want to keep it to be explicit then we don't need the `?` after `href` as it seems inconsistent with the `href: string` casted type
- do we want to return an empty string `''` when no link is provided? We could use `undefined` as the function's return type seems to allow it, or make it allow `null` to be explicit

OctaveLaventure: **3. Refactor**
wouldn't `const link = entry.link?.href?.trim() ?? ''` achieve the same result? It would avoid adding more dependency to Ramda

opencti-platform/opencti-graphql/src/manager/ingestionManager.ts
```diff
@@ -185,10 +189,14 @@ const rssItemV1Convert = (turndownService: TurndownService, feed: RssElement, en
 
 const rssItemV2Convert = (turndownService: TurndownService, channel: RssElement, item: RssItem): DataItem => {
   const { pubDate } = channel;
+  // Safe access to title with fallback to link if title is missing
+  const link = isNotEmptyField(item.link) ? ((item.link as { _: string })._ ?? '').trim() : '';
```
OctaveLaventure: **4. Refactor**
same as commment `3.`: `const link = item.link?._?.trim() ?? ''`

opencti-platform/opencti-graphql/src/manager/ingestionManager.ts
maelv-filigran: This is backend code, we do not use i18n here, even if you suggestion total make sens, it's not possible

opencti-platform/opencti-graphql/src/manager/ingestionManager.ts
maelv-filigran: Fixed! Removed the unnecessary cast and the inconsistent `?` operator. Now using `entry.link?.href?.trim() ?? ''` which is cleaner and type-safe.

maelv-filigran: one! Replaced with `entry.link?.href?.trim() ?? ''` much cleaner and removes the unnecessary Ramda dependency for this simple check.

opencti-platform/opencti-graphql/src/manager/ingestionManager.ts
maelv-filigran: Applied! Changed to `item.link?._?.trim() ?? ''` for consistency with the v1 converter.

opencti-platform/opencti-graphql/src/manager/ingestionManager.ts
OctaveLaventure: Oh sorry I assumed we did as there was translation files in `back/xx.json` but it's also in the front app 😅

OctaveLaventure (approved):

PR #13489, opened 2025-12-04
[backend] Fix IPv4 validation to reject leading zeros (#12494)

### LINKED ISSUE #12494: Malformed IPs with leading zeros can be created
## Description

It is currently possible to create IP addresses with a leading zero in one of the octets, example
- 01.1.1.1
- 1.1.01.1

Such IPs are invalid and should not be accepted by the platform.

## Environment

OCTI 6.7.20

## Reproducible Steps

1. Create a new observable of type IPv4.
2. Enter 01.1.1.1 as the value.
3. Save the observable.
4. The malformed IP is successfully created.

Image

maelv-filigran:
### Proposed changes

* Fixed IPv4 address validation regex to reject leading zeros in octets (e.g., `01.1.1.1`, `192.168.001.1`)
* Updated validation logic to ensure each octet is properly validated as 0-255 without leading zeros (except for the value `0` itself)
* Added comprehensive unit tests covering valid IPs, invalid leading zeros, out-of-range values, and CIDR notation
* Maintained backward compatibility for all valid IPv4 addresses and CIDR ranges

### Related issues

Fixes #12494

### Checklist

### Further comments

The previous regex pattern `/^(?:[0-9]{1,3}\.){3}[0-9]{1,3}(?:\/([0-9]|[1-2][0-9]|3[0-2]))?$/` accepted any 1-3 digit sequences without validating:
- Range (allowing values > 255 like `999.999.999.999`)
- Leading zeros (allowing `01.1.1.1`, `192.168.001.1`, etc.)

The new regex properly validates each octet using alternation:
- `25[0-5]` matches 250-255
- `2[0-4][0-9]` matches 200-249
- `1[0-9]{2}` matches 100-199
- `[1-9]?[0-9]` matches 0-99 without leading zeros

This ensures strict RFC compliance while maintaining CIDR notation support (`/0-32`).


JeremyCloarec (commented):
New regex looks good to me, but I'm wondering if it wouldn't be better to rely on an external library for this kind of checks? Like ip-address for exemple

maelv-filigran:
@JeremyCloarec I agree with you.  Plus ip-address can do a lot more (e.g., ipv6, etc.). But I think that for this bugfix, the regular expression is sufficient. To justify adding a dependency to an external library, we would need to utilize more of its potential . So we're leaving the bug fix and entering the feature domain. 
One line change or an new dependency ?
What do you think?

JeremyCloarec (approved):
I agree with the reasoning, let's tackle the move to an external lib for sanitization in an other ticket/feature 👍

PR #13493, opened 2025-12-04
[backend] Handle base_path setting in httpPlatform redirection (#13450)

### LINKED ISSUE #13450: Incorrect redirect of SAML callback when app.base_path is configured
## Description

On SSO callback, the redirect URL is sanitize and in case of missing referer is redirected to '/'.

But in case of opencti configured with a base_path user should be redirected to "/base_path" instead.

## Environment

4. Other environment details:

## Reproducible Steps

1. { e.g. Run ... }
2. { e.g. Click ... }
3. { e.g. Error ... }

ParamConstructor:
This should correct the behavior of the redirect that was altered via (https://github.com/OpenCTI-Platform/opencti/commit/a8f648b3a7c6bb632e098bf87f397a35fb8bca88) and did not account for systems that had app:base_path setup for deployment behind a reverse proxy

Also accounts for an oddity where in some reverse proxy situations - - may actually be the `string` '**_undefined_**' versus the javascript `value` of `undefined`.

### Proposed changes

* Modify checking in function to account for app:base_path being setup/configured
* Handle odd edge case of `string` '**_undefined_**' versus the javascript `value` of `undefined`.

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/13450

### Checklist

- [X] I tested the code for its functionality (via SAML and normal logins with and without base_path defined)

### Further comments
None

### DIFF
--- a/opencti-platform/opencti-graphql/src/http/httpPlatform.js
+++ b/opencti-platform/opencti-graphql/src/http/httpPlatform.js
@@ -37,7 +37,14 @@ import { getChatbotProxy } from './httpChatbotProxy';
 import { isStrategyActivated, StrategyType } from '../config/providers-configuration';
 
 export const sanitizeReferer = (refererToSanitize) => {
-  if (!refererToSanitize) return '/';
+  // NOTE: basePath will be configured, if the site is hosted behind a reverseProxy otherwise '/' should be accurate
+  // Ternary Operator (?): Defaults if basePath is undefined, null, "" (empty string), 0, etc (falsy values).
+  // basePath is trimmed in '../config/conf.js' to prevent a user from setting it to something like '       '
+  // NOTE: Do NOT use Nullish Coalescing (??): Would only default if basePath is undefined or null. 
+  // It might be set to an empty string and would fail to set properly in base2return var in next line 
+  const base2return = basePath ? basePath : '/';
+  // In some odd configurations refererToSanitize will be the string('undefined') versus value(undefined)
+  if (!refererToSanitize || refererToSanitize === 'undefined') return base2return;
   const base = getBaseUrl();
   const resolvedUrl = new URL(refererToSanitize, base).toString();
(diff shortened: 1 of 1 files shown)


opencti-platform/opencti-graphql/src/http/httpPlatform.js
```diff
@@ -37,7 +37,14 @@ import { getChatbotProxy } from './httpChatbotProxy';
 import { isStrategyActivated, StrategyType } from '../config/providers-configuration';
 
 export const sanitizeReferer = (refererToSanitize) => {
-  if (!refererToSanitize) return '/';
+  // NOTE: basePath will be configured, if the site is hosted behind a reverseProxy otherwise '/' should be accurate
+  // Ternary Operator (?): Defaults if basePath is undefined, null, "" (empty string), 0, etc (falsy values).
+  // basePath is trimmed in '../config/conf.js' to prevent a user from setting it to something like '       '
+  // NOTE: Do NOT use Nullish Coalescing (??): Would only default if basePath is undefined or null. 
+  // It might be set to an empty string and would fail to set properly in base2return var in next line 
+  const base2return = basePath ? basePath : '/';
```
lndrtrbn, line 45: Indeed `??` would allow empty string but `||` not. It will have the same behavior as the ternary

lndrtrbn, line 45: @aHenryJard do you remember if we have special cases where basePath was not adapted? I cannot remember all the discussions about this function

aHenryJard: I think we had plenty of discussion on baseUrl function, basePath is a bit more simplier. It looks good at first look the way it is use here. It will requires some testing on several authentication strategy to be sure.

lndrtrbn (approved):

PR #13496, opened 2025-12-04
[frontend] Simplify OpenVocabField component

lndrtrbn:
### Proposed changes

* Small refactor of component `OpenVocabField`

### Related issues

*

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/common/form/OpenVocabField.tsx
+++ b/opencti-platform/opencti-front/src/private/components/common/form/OpenVocabField.tsx
@@ -1,37 +1,13 @@
-import React, { FunctionComponent } from 'react';
+import { HTMLAttributes, Suspense } from 'react';
 import { Field } from 'formik';
-import MenuItem from '@mui/material/MenuItem';
+import { MenuItem, Tooltip } from '@mui/material';
 import { graphql, PreloadedQuery, usePreloadedQuery } from 'react-relay';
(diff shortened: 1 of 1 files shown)


opencti-platform/opencti-front/src/private/components/common/form/OpenVocabField.tsx
```diff
@@ -74,155 +74,107 @@ const OpenVocabFieldComponent: FunctionComponent<OpenVocabProps> = ({
...
-          ) : undefined,
-        }}
-      />
-    );
-  }
+
+  const internalOnChange = (v: VocabFieldValue) => {
+    const values = Array.isArray(v) 
+      ? v.map((item) => item?.value ?? item)
+      : v?.value ?? v;
```
esrevi, line 95: I wonder if we fallback on v, v is VocabFieldValue and not a string or string[], could that happen ? because onChange and onSubmit expect string | string[] type for value arg

lndrtrbn: I suppose it cannot but it was there before and I was a bit afraid to break an edge case so I let it for now

esrevi: Ok, understood !

esrevi (approved):
lgtm !

PR #13504, opened 2025-12-05
Temporary remove the rolling restart for prerelease-ce, as currently our security rules prevents deployment from circleci

victorhery-filigran:
Firewall prevents deploying to prerelease-ce from circleci, which block the CI.

Removing the steps to see quietly how to solve this correctly


efaure (approved):
Maybe you can also remove the add_ssh_keys to avoid it ending up as a strange legacy... Not mandatory to merge

victorhery-filigran:
> Maybe you can also remove the add_ssh_keys to avoid it ending up as a strange legacy... Not mandatory to merge

I thought about it, and was unsure, if we both think it 's better, done it's removed :-D

PR #13536, opened 2025-12-05
[ci] Release - Wait for connector manifest on the tag before OpenCTI build

efaure:
### Proposed changes

Fix of the release process following client-python migration to OpenCTI rep., 
Now the release process: 
 - start by tagging OpenCTI repository in order to build client python release needed by connector
 - after 10 minute (approximate duration of client python release)  tag connector repository to launch release build on it

The issue is that opencti build step after client python needs to retrieve the connector manifest on the tag of connector repo. 

The goal of this PR is to add in OpenCTI release build in circle CI a step (`wait_for_connector_release`) to wait for the tag on connector manifest before building OpenCTI to avoid workflow failure. 

This step timeout after 15 minutes if the manifest is not available on the tag.


efaure:
@labo-flg This PR should ensure that during release we wait for connector's manifest to be available on the tag

aHenryJard (approved):
Look good, thanks

Issue #13386, opened 2025-11-28
OCTI Redirection issue when deploying a connector


EllynBsc:
## Description

Related to https://github.com/FiligranHQ/xtm-hub/issues/1319

When we’re not logged in to OpenCTI and attempt to use the one-click deploy feature, we are redirected to:
https://filigran.octi.filigran.io/dashboard/xtm-hub/deploy-connector/hygiene?openConfig=true.
However, after logging in, we are not returned to the original URL we were trying to access.

labo-flg:
It also happens for any other URL actually.
If you are not logged in, you are redirected to login screen but then you are not redirected back to the original URL you requested.

esrevi:
Hello @EllynBsc , @labo-flg,

I click on a link with that URL on **local** env : http://localhost:3000/dashboard/xtm-hub/deploy-connector/hygiene?openConfig=true
I log through the login/pwd form, all goes well, the URL remains the same and the app navigates to that page.

On **testing** env, same URL, I am redirected to login page, I can only log with the SSO button, and indeed, it redirects to /dashboard instead of the original URL.

Could you confirm me that the issue is on login via a provider ?

EllynBsc:
Yes for me it's happening on testing or dev environment @esrevi  (cc @jpkha)

alice-debra:
Hello @esrevi, 
This one is done, but not close do we have more work to do?

EllynBsc:
cc @jpkha @jbanety what's the status of this one ? we can close it right ?

esrevi:
Hi @alice-debra , this has been fixed, it could be closed now I think !

PR #13545, opened 2025-12-05
[backend] redirect after sso login (#13386)

esrevi:
The issue is when authenticating with Saml provider, the user is redirected to base url thus the dashboard in the app.

### Proposed changes

* Fix SAML authentication redirect by passing RelayState via additionalParams in passport.authenticate(). This preserves the original URL through the SAML flow, allowing post-login redirection. 
* Should fix redirection with search params for SAML and OpenId/Oauth providers

Auth with SAML provider and redirect on auth to the url with search params

https://github.com/user-attachments/assets/7b54ff59-839f-4c30-893d-006dfab4e73f

Auth with OpenId provider and redirect on auth to the url with search params

https://github.com/user-attachments/assets/a03ce3a2-2a7f-466b-b5ba-851362b0ca90

### Related issues

* #13386 

### Checklist

### DIFF
--- a/opencti-platform/opencti-dev/docker-compose.yml
+++ b/opencti-platform/opencti-dev/docker-compose.yml
@@ -165,8 +165,8 @@ services:
       container_name: opencti-dev-keycloak
       command: start-dev
       environment:
-        KEYCLOAK_ADMIN: "admin"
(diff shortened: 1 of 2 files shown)


aHenryJard (approved):
Tested locally with OpenID, SAML, and with base_path, looks good to me.

PR #13546, opened 2025-12-05
[backend] move sharing and Auth members changes to activity logs (#8237)

### LINKED ISSUE #8237: Protect visilibility of protected attributes in history
# Problem
Currently in Org segregation, if a an entity is shared to an org, this organization will have access to the Entity & its history.
If the entity is shared once more to another org, even if the org will appear as "restricted" in the shared orgs, the history can show the name in clear. 
This can lead to leakage of information.

Outside the Org segregation, if an entity is applied some Authorized Members, authorized Members will appear in the logs. Once a user has access to the entity in read only, if the Authorized Members are modified, users that can see the entity will see users in the logs (or even organizations). This leaks information.

# Solution
Sharing/Unsharing to an org & apply authorize members/editing authorize members/removing authorize members actions will not be logged in the **knowledge history** of the entity, but will be logged in the Activity log.

This way, only users with Manage Activity will be able to see sharing logs and no information will be leaked.

marieflorescontact:
### Proposed changes

* we don't stor any more Auth members or OrgSharing changes 
* history message is rebuild from changes
* use `publiUserAction` to push in activity logs if Auth members or OrgSharing changes 

### Related issues

* closes: https://github.com/OpenCTI-Platform/opencti/issues/8237

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/database/middleware.js
+++ b/opencti-platform/opencti-graphql/src/database/middleware.js
@@ -242,6 +242,7 @@ import { modules } from '../schema/module';
 import { doYield } from '../utils/eventloop-utils';
 import { ENTITY_TYPE_SECURITY_COVERAGE, RELATION_COVERED } from '../modules/securityCoverage/securityCoverage-types';
 import { findById as findDraftById } from '../modules/draftWorkspace/draftWorkspace-domain';
+import {publishUserAction} from '../listener/UserActionListener';
 
 // region global variables
 const MAX_BATCH_SIZE = nconf.get('elasticsearch:batch_loader_max_size') ?? 300;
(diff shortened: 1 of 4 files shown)


opencti-platform/opencti-graphql/src/database/middleware.js
```diff
@@ -3100,7 +3117,7 @@ const createEntityRaw = async (context, user, rawInput, type, opts = {}) => {
   const input = { ...rawInput };
   const { confidenceLevelToApply } = controlCreateInputWithUserConfidence(user, input, type);
   input.confidence = confidenceLevelToApply; // confidence of new entity will be capped to user's confidence
-  // authorized_members renaming
+  // authori  zed_members renaming
```
aHenryJard: why an additional space here ?

opencti-platform/opencti-graphql/src/database/middleware.js
```diff
@@ -2471,6 +2473,20 @@ export const updateAttributeMetaResolved = async (context, user, initial, inputs
       } : undefined;
       const relatedRestrictions = extractObjectsRestrictionsFromInputs(updatedInputs, initial.entity_type);
       const { pir_ids } = extractObjectsPirsFromInputs(updatedInputs, initial.entity_type);
+      const isSecurityUpdate = () => {
+        return updatedInputs.some((input) => input.key === objectOrganization.name || input.key === authorizedMembers.name);
```
SouadHadjiat, line 2516: What about when there are other keys along with objectOrganizations or authorizedMembers ? like for example, you change both description and orga sharing ?

marieflorescontact: yes you are right @SouadHadjiat  ,
In the case of an update with multiple operations, for example if a playbook updates the description AND shares a Report with an organization, this solution poses a problem: we will have the following message 
 ````
"message":"adds `identity` in `Label` - `ANSSI` in `Shared with`"
 ````
 and the logs of the 2 operations will go into the Activity

PR #13574, opened 2025-12-08
[frontend] (bulk search) add number of known entities in tab title (#3360)

Archidoit:
### Proposed changes
Bulk Search : Add number of known entities in 'known entities' tab title

image

### Related issues
#3360


lndrtrbn (commented):
You cannot use a simple calculation `values.length - numberOfUnknownEntities`?

Archidoit:
@lndrtrbn No because we can have several known values for the same input value (exemple: for the input value 'test', we can have a report, a malware and a city, each of them having 'test' for representative).

lndrtrbn (approved):

PR #13587, opened 2025-12-08
[backend] Refactor OTP to use a directive to allow specific API

richard-julien:
Simple refactor to use a directive instead of direct naming in the authDirective.js

### DIFF
--- a/opencti-platform/opencti-front/src/schema/relay.schema.graphql
+++ b/opencti-platform/opencti-front/src/schema/relay.schema.graphql
@@ -2,6 +2,8 @@ directive @auth(for: [Capabilities] = [], and: Boolean = false) on OBJECT | FIEL
 
 directive @public on OBJECT | FIELD_DEFINITION
 
+directive @allowUnprotectedOTP on OBJECT | FIELD_DEFINITION
+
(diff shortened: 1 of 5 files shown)


opencti-platform/opencti-graphql/src/graphql/authDirective.js
```diff
@@ -1,13 +1,19 @@
-/* eslint-disable no-underscore-dangle,no-param-reassign */
 import { getDirective, MapperKind, mapSchema } from '@graphql-tools/utils';
 import { filter, includes, map } from 'ramda';
-// eslint-disable-next-line import/extensions
+// noinspection JSFileReferences
 import { defaultFieldResolver } from 'graphql/index.js';
```
xfournet: you can replace
import { defaultFieldResolver } from 'graphql/index.js'
by
import { defaultFieldResolver } from 'graphql'

and avoid this // noinspect comment

xfournet: and the same in opencti-platform/opencti-graphql/src/types/graphql.d.ts

xfournet (approved):

PR #13597, opened 2025-12-09
[frontend] Playbooks: fix bug with field array (#12362)

lndrtrbn:
### Proposed changes

* Fix bug using field array (for example notifier in component send notification)

The bug was that when choosing one, the id was displayed instead of title and when selecting multiple there was a crash

### Related issues

* #12362

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/data/playbooks/playbookFlow/playbookFlowFields/PlaybookFlowFieldArray.tsx
+++ b/opencti-platform/opencti-front/src/private/components/data/playbooks/playbookFlow/playbookFlowFields/PlaybookFlowFieldArray.tsx
@@ -14,7 +14,7 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 */
 
 import { MenuItem, Tooltip, TooltipProps } from '@mui/material';
-import { Field, useFormikContext } from 'formik';
+import { Field } from 'formik';
 import AutocompleteField from '../../../../../../components/AutocompleteField';
 import { fieldSpacingContainerStyle } from '../../../../../../utils/field';
 import useEntityTranslation from '../../../../../../utils/hooks/useEntityTranslation';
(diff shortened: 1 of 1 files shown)


opencti-platform/opencti-front/src/private/components/data/playbooks/playbookFlow/playbookFlowFields/PlaybookFlowFieldArray.tsx
```diff
@@ -38,14 +38,10 @@ const PlaybookFlowFieldArray = ({
...
+  const fieldOptions = options.map(o => o.const);
 
-  const setOneOfValue = (value: Option) => {
-    setFieldValue(name, value.const);
-  };
-
-  const setMultipleValue = (values: Option[]) => {
-    setFieldValue(name, values.map((o) => (o.const)));
+  const findOption = (value: string) => {
+    return options.find(o => o.const === value);
```
OctaveLaventure: **1. Question**
Are we sure we'll always find an option matching the value ? If not maybe a fallback return could be needed

opencti-platform/opencti-front/src/private/components/data/playbooks/playbookFlow/playbookFlowFields/PlaybookFlowFieldArray.tsx
```diff
@@ -38,14 +38,10 @@ const PlaybookFlowFieldArray = ({
   multiple = false,
 }: PlaybookFlowFieldArrayProps) => {
   const { translateEntityType } = useEntityTranslation();
-  const { setFieldValue } = useFormikContext();
+  const fieldOptions = options.map(o => o.const);
```
OctaveLaventure, line 41: I don't know how big the options array is, but if it's big we could use a useMemo to avoid recomputing at every render (not worth it if it's a small array as useMemo has its own perf footprint)

OctaveLaventure (approved):
Nice! 👍🏻 
Reviewed code only, if you want to demo the fix I'm available

opencti-platform/opencti-front/src/private/components/data/playbooks/playbookFlow/playbookFlowFields/PlaybookFlowFieldArray.tsx
lndrtrbn: No need to use a memo, kind of small, most of the time around 10 elements

PR #13608, opened 2025-12-10
[backend] make export timeout configurable via platform settings (#13594)

### LINKED ISSUE #13594: feat(export): make data export timeout configurable
## Use case

The data export feature currently has a hardcoded timeout of 20 minutes. This timeout should be made configurable through platform settings to accommodate various export scenarios, particularly for very large data exports that may require more time to complete.

Move the export timeout configuration to a configurable setting that can be adjusted without code changes.

Image

hsachdeva9:
### Proposed changes
* Added `platform_export_timeout` field to Settings GraphQL schema to make export timeout configurable
* Modified `workToExportFile` function in `work.js` to read timeout value from platform settings instead of hardcoded 20 minutes
* Updated function signature to async and added context parameter for settings access
* Updated all callers of `workToExportFile` to properly handle async nature and pass context
* Implemented fallback to default 20 minutes if setting is not configured, maintaining backward compatibility

### Related issues
Fixes #13594

### Checklist

### Further comments

**Implementation Details:**

The export timeout was previously hardcoded at 20 minutes in the `workToExportFile` function. This change makes it configurable through the platform settings UI without requiring code changes or redeployment.

**Changes made:**
1. **GraphQL Schema** (`config/schema/opencti.graphql`): Added `platform_export_timeout: Int` field to Settings type
2. **Backend Logic** (`src/domain/work.js`): 
   - Function now fetches settings using `getEntityFromCache` 
   - Reads `platform_export_timeout` with fallback to 20 minutes
   - Made function async and added context parameter

**Why this approach:**
- Follows existing OpenCTI patterns for platform settings (similar to `platform_session_timeout`)
- Minimal code changes required
- Backward compatible - defaults to original 20-minute timeout
- No breaking changes to API or behavior

**Testing:**
- Verified the new field appears in platform settings UI
- Tested export functionality with default value (20 minutes)
- Tested with custom timeout values (30, 45 minutes)
- Confirmed backward compatibility when setting is not configured

### DIFF
--- a/opencti-platform/opencti-graphql/config/schema/opencti.graphql
+++ b/opencti-platform/opencti-graphql/config/schema/opencti.graphql
@@ -1507,6 +1507,7 @@ type Settings implements InternalObject & BasicObject & ThemeSettings & IntlSett
   platform_banner_text: String
   platform_banner_level: String
   platform_session_idle_timeout: Int
+  platform_export_timeout: Int
   platform_session_timeout: Int
   platform_whitemark: Boolean
   platform_demo: Boolean
--- a/opencti-platform/opencti-graphql/src/domain/work.js
+++ b/opencti-platform/opencti-graphql/src/domain/work.js
@@ -15,10 +15,16 @@ import { RELATION_OBJECT_MARKING } from '../schema/stixRefRelationship';
 import { DRAFT_VALIDATION_CONNECTOR, DRAFT_VALIDATION_CONNECTOR_ID } from '../modules/draftWorkspace/draftWorkspace-connector';
 import { logApp } from '../config/conf';
 import { internalLoadById } from '../database/middleware-loader';
+import { getEntityFromCache } from '../database/cache';
+import { ENTITY_TYPE_SETTINGS } from '../schema/internalObject';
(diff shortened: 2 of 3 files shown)


aHenryJard (commented):
Could you please use environment variable instead of settings. Because using settings from database will requires to have some UI to be able to change it and it's not needed.

In json configuration it could be :
```
"app": {
    "work": {
      "export_timeout": 20
    },
```

Which corresponds to environment variable `APP__WORK__EXPORT_TIMEOUT`

You can see as example this one:
https://github.com/OpenCTI-Platform/opencti/blob/23e924ec6ddc093eba26b82fc36a1ab0491db4fd/opencti-platform/opencti-graphql/src/http/httpRollingFeed.ts#L18

nino-filigran:
@hsachdeva9 In two weeks, without any response, we will close your PR. You will be able to reopen it if needed. Thank you for your contribution.

marieflorescontact:
Hello @hsachdeva9 ,
As previously mentioned, we're closing this issue due to inactivty. Feel free to reopen it if you need it.

Issue #13361, opened 2025-11-27
[ci] Migrate pycti example test to GitHub action


efaure:
## Use case

Migrate the pycti example step : 
 - from drone https://github.com/OpenCTI-Platform/opencti/blob/4e45975d68269b9c8798d973e8e9797dfce0bb63/.drone.yml#L283
 - to GitHub Action https://github.com/OpenCTI-Platform/opencti/blob/master/.github/workflows/ci-test-client-python.yml

PR #13638, opened 2025-12-11
[ci] Move pycti-examples from Drone to Github actions (#13361)

Archidoit:
### Proposed changes
Move pycti-examples from Drone to Github actions

### Related issues
#13361


aHenryJard (approved):
Looks good to me and CI is green so all good. Thanks !

Issue #13637, opened 2025-12-11
A user with manage creds but no org can login in platforms with Org segregation


nino-filigran:
## Description

A user with manage creds but no org can login in platforms with Org segregation.
This should not be possible and can cause issues down the llne.
## Environment

Testing 

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Define a main org platform
2. Ensure that you have a user with a role having only the capa the capa "manage credentials"
3. Grant this role to the user
4. Log in with this user
## Expected Output

The user should not be able to log in 

## Actual Output

The user is able to log in

lndrtrbn:
The issue is not linked to capability Manager Credentials directly.
Right now the behavior is that if you have any capability of Settings you can log in.

Expected: can login if you have Bypass and that's the only capability allowing you to login.

PR #13644, opened 2025-12-11
[backend] Only allow bypass user to login without organisations (#13637)

delemaf:
### Proposed changes

* Only allow bypass bypass user to login without organisations

Testing:
* Try to log with a user that have **no org** and capa **Manage credentials** you shouldn't be able to login
* Try to log with a user that have **no org** and capa **Bypass** you should be able to login

### Related issues

* #13637 

### Checklist


marieflorescontact (approved):
locally tested ok 👌

PR #13658, opened 2025-12-12
[frontend/backend] SSH key open vocab in Observable forms (#13656)

### LINKED ISSUE #13656: SSH key: key type does not fetch the key type open vocab
## Description

Key type is an open vocab.
When creating a SSH key, the key field should e a dropdown with all values for key type.
## Environment

Testing

## Reproducible Steps

1. Create a new SSH key
2. In the creation form; go in the filed key type

## Expected Output

It should be drop down that reads from the open vocab key type
## Actual Output

It's a string field, not restricting to the values set as open vocab
## Additional information

By default, the open vocab shuld contain the values: 
- RSA
- DSA
- ECDSA
- EdDSA

Archidoit:
### Proposed changes
- SSH key type in observable creation and edition forms should be a dropdown with the key_type_ov vocabulary values
- StixCyberObservableDetails to tsx

### Related issues
#13656


JeremyCloarec (approved):

opencti-platform/opencti-front/src/private/components/observations/stix_cyber_observables/StixCyberObservableDetails.tsx
```diff
@@ -0,0 +1,562 @@
...
+
+const reorderMediaContentObservablesAttributes = (data: { key: string, value: string }[]) => {
+  const desiredOrder = ['content', 'title', 'media_category', 'url', 'publication_date'];
+
+  return desiredOrder
+    .map((key) => data.find((item) => item.key === key))
+    .filter(Boolean) as { key: string, value: string }[];
+};
+
+const useTransformValue = (value: string | boolean | string[] | number, key: string) => {
```
SarahBocognano: As it is a very generic hook, what do you think if we move this one into a general hooks folder ?

opencti-platform/opencti-front/src/private/components/observations/stix_cyber_observables/StixCyberObservableDetails.tsx
```diff
@@ -0,0 +1,562 @@
...
+
+  return result;
+};
+
+interface DownloadFileButtonMenuProps {
+  fileSize: number | null | undefined,
+  encodedFilePath: string | null,
+}
+
+const DownloadFileButtonMenu = ({
```
SarahBocognano, line 307: Same here, maybe we can move it to a generic component folder ? To make the component less big

SarahBocognano (commented):
I have a question regarding the default values of key type vocab. I don't have them when launching the platform, should we add them at the init as default values ? @nino-filigran

SarahBocognano:
Other than my questions and remarks, everything is great :D ✅

nino-filigran:
Yes @SarahBocognano  & @Archidoit : if we could add the following key types also by default in platforms, it would be great.

RSA
DSA
ECDSA
EdDSA

Archidoit:
> Yes @SarahBocognano & @Archidoit : if we could add the following key types also by default in platforms, it would be great.
> 
> RSA DSA ECDSA EdDSA

@nino-filigran It's strange, there is already a migration for that... It has not been played well ? (migration name: add-vocabulary-ssh-key-type)

written by @CelineSebe do you have more info ?

CelineSebe:
> > Yes @SarahBocognano & @Archidoit : if we could add the following key types also by default in platforms, it would be great.
> > RSA DSA ECDSA EdDSA
> 
> @nino-filigran It's strange, there is already a migration for that... It has not been played well ? (migration name: add-vocabulary-ssh-key-type)
> 
> written by @CelineSebe do you have more info ?

Was there a problem with this migration? The key types are not added when I look at testing in Vocabularies (except 'rsa').
We have an issue with the key_type too @nino-filigran ?
image
image

Archidoit:
> > > Yes @SarahBocognano & @Archidoit : if we could add the following key types also by default in platforms, it would be great.
> > > RSA DSA ECDSA EdDSA
> > 
> > 
> > @nino-filigran It's strange, there is already a migration for that... It has not been played well ? (migration name: add-vocabulary-ssh-key-type)
> > written by @CelineSebe do you have more info ?
> 
> Was there a problem with this migration? The key types are not added when I look at testing in Vocabularies (except 'rsa'). We have an issue with the key_type too @nino-filigran ? image image

If I upgrade the migration number and the migration is re-run, the vocabularies are well created... maybe a problem with the migration number in our instances?

CelineSebe:
> If I upgrade the migration number and the migration is re-run, the vocabularies are well created... maybe a problem with the migration number in our instances?

Yes, it's a migration number issue, for sure!

nino-filigran:
LEt's go with (for sake of clarity) : 
image

opencti-platform/opencti-front/src/private/components/observations/stix_cyber_observables/StixCyberObservableDetails.tsx
Archidoit: I'm gonna do this :)

PR #13662, opened 2025-12-12
[ci] Docker build in GitHub action - change client-python local build

efaure:
### Proposed changes

Usage of local client python version has been added in  https://github.com/OpenCTI-Platform/opencti/issues/13397 
But this build strategy rely on using the locally checked out version of client python, requiring to make the client python install in the docker build. 

For image that will be deployed in testing environment, it's better to have images closest to the release images version. Building python locally will introduce a difference. It's better for those images to update the requirements.txt in order to use client-python from a specific commit instead of local one. 

For harmonization purpose, it's relevant to also update testing image build in the same way.

For image 

### Related issues

*
*

### Checklist

### DIFF
--- a/.github/workflows/ci-docker-build.yml
+++ b/.github/workflows/ci-docker-build.yml
@@ -11,11 +11,16 @@ on:
         description: "Reference of the branch or commit sha"
         type: string
         default: ${{ github.sha }}
-      dockerfile:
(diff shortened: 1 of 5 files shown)


efaure:
Outdated with https://github.com/OpenCTI-Platform/opencti/pull/13698

PR #13702, opened 2025-12-15
[frontend/backend] add report `content` attribute to fintel (#12530)

### LINKED ISSUE #12530: [Fintel] ‘Content’ inclusion into FIntel template
## Use case

We need to generate “FINTEL” reports, including the report’s content as part of them.
However, when creating a template, it is currently not possible to reference the ‘content’, and only the description of the report can be used.

delemaf:
### Proposed changes

* Add report `content` attribute to fintel 

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/12530

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/utils/outcome_template/engine/stix_core_objects/StixCoreObjectsAttributesQuery.tsx
+++ b/opencti-platform/opencti-front/src/utils/outcome_template/engine/stix_core_objects/StixCoreObjectsAttributesQuery.tsx
@@ -83,6 +83,7 @@ const stixCoreObjectsAttributesQuery = graphql`
                 description
                 report_types
                 x_opencti_reliability
+                report_content: content
             }
             ... on Grouping {
                 name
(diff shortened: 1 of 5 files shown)


opencti-platform/opencti-front/src/utils/outcome_template/engine/stix_core_objects/StixCoreObjectsAttributesQuery.tsx
```diff
@@ -83,6 +83,7 @@ const stixCoreObjectsAttributesQuery = graphql`
                 description
                 report_types
                 x_opencti_reliability
+                report_content: content
```
Archidoit, line 86: why this renaming? Cannot it cause confusion ?

Archidoit:
It's working well :)

Archidoit (approved):

opencti-platform/opencti-front/src/utils/outcome_template/engine/stix_core_objects/StixCoreObjectsAttributesQuery.tsx
delemaf: There was a graphql conflict with Note content which is `String!` but Report content is `String` (there's also is this kind of renaming in other query with containers alongside Note)

PR #13703, opened 2025-12-15
[backend] Allow to customize URL and headers for manifest fetch during build (#13069)

### LINKED ISSUE #13069: Connector manifest download from GitHub prevent build in air-gapped CI
## Description

During build, the connector manifest is downloaded from github. That's prevent to build in air-gapped CI.

## Environment

OpenCTI version: since 6.8.0

## Reproducible Steps

1. ensure to be offline
2. go to `opencti-platform/opencti-graphql`  directory
3. execute `yarn build:prod` 

## Expected Output

Build success

## Actual Output

Build failure

xfournet:
### Proposed changes

*  Allow to customize URL and headers for manifest fetch during build by using env var: `OPENCTI_MANIFEST_URL` & `OPENCTI_MANIFEST_HEADERS`

### Related issues
* Fix #13069

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/script/get-connectors-manifest.js
+++ b/opencti-platform/opencti-graphql/script/get-connectors-manifest.js
@@ -3,6 +3,10 @@ import path from 'path';
 import { fileURLToPath } from 'url';
 
 const getSchemaURL = () => {
+  if (process.env.OPENCTI_MANIFEST_URL) {
+    return process.env.OPENCTI_MANIFEST_URL;
+  }
+
   if (process.env.TAG_VERSION) {
     return `https://raw.githubusercontent.com/OpenCTI-Platform/connectors/refs/tags/${process.env.TAG_VERSION}/manifest.json`;
   }
@@ -10,6 +14,11 @@ const getSchemaURL = () => {
   return 'https://raw.githubusercontent.com/OpenCTI-Platform/connectors/refs/heads/master/manifest.json';
 };
 // use tags instead of master if env from circle TAG is set
+
+const getFetchHeaders = () => {
+  const headersEnv = process.env.OPENCTI_MANIFEST_HEADERS;
+  return headersEnv ? JSON.parse(headersEnv) : {};
+};
 
 const OUTPUT_DIR = '../src/__generated__';
 const OUTPUT_FILE = 'opencti-manifest.json';
@@ -23,7 +32,9 @@ async function getConnectorManifest() {
   try {
     // fetch file
     console.info(`➡️ Fetching manifest from: ${schemaUrl}`);
-    const res = await fetch(schemaUrl);
+    const res = await fetch(schemaUrl, {
+      headers: getFetchHeaders(),
+    });
     if (!res.ok) throw new Error(`HTTP ${res.status}`);
     const data = await res.text();
 


opencti-platform/opencti-graphql/script/get-connectors-manifest.js
```diff
@@ -3,6 +3,10 @@ import path from 'path';
 import { fileURLToPath } from 'url';
 
 const getSchemaURL = () => {
+  if (process.env.OPENCTI_MANIFEST_URL) {
+    return process.env.OPENCTI_MANIFEST_URL;
+  }
```
OctaveLaventure, line 8: **1. Documentation**
We should document this behavior in the docs so users don't get confused. More specifically, the impact of `TAG_VERSION` being ignored here if `OPENCTI_MANIFEST_URL` is set

opencti-platform/opencti-graphql/script/get-connectors-manifest.js
```diff
@@ -23,7 +32,9 @@ async function getConnectorManifest() {
   try {
     // fetch file
     console.info(`➡️ Fetching manifest from: ${schemaUrl}`);
-    const res = await fetch(schemaUrl);
+    const res = await fetch(schemaUrl, {
+      headers: getFetchHeaders(),
```
OctaveLaventure, line 36: **2. Question**
Is the behavior the same with an empty object as with no `headers` key ?

OctaveLaventure (commented):
Approving to not block it but consider the comments

OctaveLaventure (approved):

opencti-platform/opencti-graphql/script/get-connectors-manifest.js
```diff
@@ -11,6 +15,11 @@ const getSchemaURL = () => {
 };
 // use tags instead of master if env from circle TAG is set
 
+const getFetchHeaders = () => {
+  const headersEnv = process.env.OPENCTI_MANIFEST_HEADERS;
+  return headersEnv ? JSON.parse(headersEnv) : {};
```
esrevi, line 20: I think we should enforce the validation here because JSON.parse() can successfully parse values that aren't valid header objects such as array for example

opencti-platform/opencti-graphql/script/get-connectors-manifest.js
xfournet: TAG_VERSION is currently not documented, and more generally except the overall build behavior we don't document the build process in details.
As we will probably remove the embedded manifest in OpenCTI later, i don't think it worth to document it.
It's planned to communicate that only to the customer that required it (air gapped custom build)

opencti-platform/opencti-graphql/script/get-connectors-manifest.js
xfournet: the spec says that this are the headers to be added to the request, so being empty or not having the key is supposed to be the same.

https://developer.mozilla.org/en-US/docs/Web/API/RequestInit#headers

opencti-platform/opencti-graphql/script/get-connectors-manifest.js
xfournet: fetch is already providing headers validation, so setting an array (or anything that is not valid) will fail properly. Given it's build only features it's not needed to provide more information

xfournet:
> Approving to not block it but consider the comments

@OctaveLaventure  @esrevi just answered your comment and merged, let me know in case their is concern about that

PR #13711, opened 2025-12-16
[worker] add missing user agent on push handler

JeremyCloarec:
### Proposed changes

* add missing user agent on push handler. We missed it when adding user agent to worker
*

### Related issues

*
*

### Checklist

### DIFF
--- a/opencti-worker/src/push_handler.py
+++ b/opencti-worker/src/push_handler.py
@@ -7,7 +7,7 @@ from typing import Any, Dict, Union, Literal
 import pika
 from pika.adapters.blocking_connection import BlockingChannel
 
-from pycti import OpenCTIApiClient, OpenCTIStix2Splitter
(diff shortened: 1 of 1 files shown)


opencti-worker/src/push_handler.py
```diff
@@ -35,6 +35,7 @@ def __post_init__(self) -> None:
             log_level=self.log_level,
             json_logging=self.json_logging,
             ssl_verify=self.ssl_verify,
+            provider="pushHandler/" + __version__,
```
xfournet: why not using 'worker/' ?

JeremyCloarec: I thought it could be great to have a different user agent on the client used in worker compared to the one used in push_handler. But I can use worker/ on both if you think that's overkill

xfournet (approved):

PR #13712, opened 2025-12-16
[backend] Ability to run a given migration from the API (#13713)

Archidoit:
### Proposed changes
New API to run a given migration 

### Related issues
#13713

### DIFF
--- a/opencti-platform/opencti-graphql/src/database/migration.js
+++ b/opencti-platform/opencti-graphql/src/database/migration.js
@@ -2,7 +2,7 @@ import * as R from 'ramda';
 import { MigrationSet } from 'migrate';
 import Migration from 'migrate/lib/migration';
 import { logApp, logMigration, PLATFORM_VERSION } from '../config/conf';
-import { DatabaseError } from '../config/errors';
+import { DatabaseError, FunctionalError } from '../config/errors';
 import { RELATION_MIGRATES } from '../schema/internalRelationship';
 import { ENTITY_TYPE_MIGRATION_REFERENCE, ENTITY_TYPE_MIGRATION_STATUS } from '../schema/internalObject';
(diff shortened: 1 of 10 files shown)


opencti-platform/opencti-graphql/src/modules/migration/migration.graphql
```diff
@@ -0,0 +1,3 @@
+type Query {
+  runMigration(migrationName: String!): String! @auth(for: [BYPASS])
```
SouadHadjiat: I think it should be a mutation

Archidoit: indeed! done :)

opencti-platform/opencti-graphql/src/database/migration.js
```diff
@@ -18,21 +20,37 @@ const normalizeMigrationName = (rawName) => {
...
+  const title = normalizeMigrationName(name);
+  const [time] = title.split('-');
+  const timestamp = parseInt(time, 10);
+  return { title, timestamp, up: migration.up, down: migration.down };
+};
+
+export const retrieveMigration = async (migrationFileName) => {
+  try {
+    const migration = await import(`../migrations/${migrationFileName}.js`);
+    return buildMigrationObject(migration, `${migrationFileName}.js`);
```
marieflorescontact, line 33: are we sure to have only .js file? (no ts?)

marieflorescontact (commented):
locally tested ok 👌

opencti-platform/opencti-graphql/src/database/migration.js
Archidoit: all the migrations are in .js, it was the former code to consider only .js, we only did a refacto here

marieflorescontact (approved):

PR #13714, opened 2025-12-16
build(backend): remove esbuild-plugin-import-glob (#13769)

Archidoit:
### Proposed changes
Remove esbuild-plugin-import-glob and global import from migration.js

Necessary for vite v4

### Related issues
#13769

### DIFF
--- a/opencti-platform/opencti-graphql/src/database/migration.js
+++ b/opencti-platform/opencti-graphql/src/database/migration.js
@@ -28,7 +28,7 @@ const retrieveMigrations = async () => {
     migrationsFilenames.map((file) => import(`../migrations/${file}`)));
 
   const knexMigrations = migrations.map((migration, i) => ({
-    name: migrationsFilenames[i].substring('../migrations/'.length),
+    name: migrationsFilenames[i],
     migration,
   }));
   return knexMigrations.map(({ name, migration }) => {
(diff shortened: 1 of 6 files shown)


opencti-platform/opencti-graphql/src/database/migration.js
```diff
@@ -18,7 +20,13 @@ const normalizeMigrationName = (rawName) => {
   return rawName;
 };
 
-const retrieveMigrations = () => {
+const retrieveMigrations = async () => {
+  const migrationsFilenames = (await fs.readdir(MIGRATION_DIRECTORY_PATH))
```
xfournet, line 24: this cannot works because of bundling. After bundling sources are no more available as individual files.

To remove the global import we should rather create an index.ts file in src/migrations directory like that

```ts
import * as migration_1608047073975_fix_missing_deletion from './1608047073975-fix_missing_deletion';
....
import * as migration_1765807866114_add_vocabulary_ssh_key_type from './1765807866114-add-vocabulary-ssh-key-type';

export default {
  migration_1608047073975_fix_missing_deletion,
   ....
  migration_1765807866114_add_vocabulary_ssh_key_type,
};
```

and then we could be able to replace the glob import by `import migrations from '../migrations'`  and adapt code accordingly

Archidoit: Oh I wanted to avoid this solution because I'm afraid people will forget to add their new migration file name each time they create a migration, or will forget to change the file name if they rename a migration...

Archidoit: In local the solution we did was working... Do you know how to test correctly in local to reproduce the bundling ?

Archidoit: > Oh I wanted to avoid this solution because I'm afraid people will forget to add their new migration file name each time they create a migration, or will forget to change the file name if they rename a migration...

We can use eslint to detect that

PR #13741, opened 2025-12-17
[frontend] AutocompleteField to Typescript (#13845)

lndrtrbn:
### Proposed changes

* AutocompleteField to Typescript
* New component Field to benefits from Typescript

### Related issues

* #13845

### Checklist


opencti-platform/opencti-front/src/utils/field.tsx
```diff
@@ -0,0 +1,32 @@
...
+  value: string;
+  label: string;
+  color?: string;
+  type?: string;
+  standard_id?: string;
+}
+
+// TODO move this interface inside file KillChainPhasesField
+// when it has been transformed it TypeScript.
+export interface KillChainPhaseFieldOption extends FieldOption {
```
aHenryJard, line 17: So is this todo still accurate ? Either this code should be moved or the todo removed as far as I understand it.

lndrtrbn: The component KillChainPhasesField has not been updated to TS here, so the code and comment are still relevant. This component has been reworked in an other branch: https://github.com/OpenCTI-Platform/opencti/pull/13749 still in draft

aHenryJard: ah ok I understood the other way around.

aHenryJard (approved):
Tested locally for country on threat actor, case templates and creator, all good for me.

PR #13777, opened 2025-12-19
fix(backend): gate AI exclusively by settings.platform_ai_enabled (#13568)

### LINKED ISSUE #13568: fix(ask-ai): only settings.platform_ai_enabled should be used to enable/disable AI in the platform
## Description

Only  settings.platform_ai_enabled should be used to enable/disable AI in the platform, the environment variable should not be use anymore.

https://github.com/OpenCTI-Platform/opencti/blob/c1eb4d4658528214004450033ffac813d50b23c9/opencti-platform/opencti-graphql/src/database/ai-llm.ts#L18

This has been migrated on 
https://github.com/OpenCTI-Platform/opencti/blob/master/opencti-platform/opencti-graphql/src/migrations/1756815510612-add-ai-in-platform-settings.js

amiraifori:
Fixes #13568

- Remove legacy ai:enabled gating from AI LLM initialization (ai-llm.ts)
- Add runtime guard in ai-domain.ts to block AI operations when settings.platform_ai_enabled is false
- Add integration coverage ensuring AI mutations are blocked when platform AI is disabled

Note: CI/integration tests were not run locally due to a Windows test bootstrap crash referencing detectors/platform/node/machine-id/execAsync (unrelated to this change).

### DIFF
--- a/opencti-platform/opencti-graphql/src/database/ai-llm.ts
+++ b/opencti-platform/opencti-graphql/src/database/ai-llm.ts
@@ -5,7 +5,7 @@ import { Mistral } from '@mistralai/mistralai';
 import type { ChatCompletionStreamRequest } from '@mistralai/mistralai/models/components';
 import { AuthenticationError, AzureOpenAI, OpenAI } from 'openai';
 import conf, { BUS_TOPICS, logApp } from '../config/conf';
-import { UnknownError, UnsupportedError } from '../config/errors';
+import { FunctionalError, UnknownError, UnsupportedError } from '../config/errors';
 import { OutputSchema } from '../modules/ai/ai-nlq-schema';
 import type { Output } from '../modules/ai/ai-nlq-schema';
 import { AI_BUS } from '../modules/ai/ai-types';
@@ -14,8 +14,8 @@ import { truncate } from '../utils/format';
 import { notify } from './redis';
 import { isEmptyField } from './utils';
 import { addNlqQueryCount } from '../manager/telemetryManager';
+import { AI_DISABLED_ERROR_MESSAGE } from '../utils/ai/aiConstants';
 
-const AI_ENABLED = conf.get('ai:enabled');
 const AI_TYPE = conf.get('ai:type');
 const AI_ENDPOINT = conf.get('ai:endpoint');
 const AI_TOKEN = conf.get('ai:token');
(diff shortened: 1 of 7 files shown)


opencti-platform/opencti-graphql/src/database/ai-llm.ts
```diff
@@ -27,7 +26,7 @@ const AI_AZURE_DEPLOYMENT = conf.get('ai:ai_azure_deployment');
 
 let client: Mistral | OpenAI | AzureOpenAI | null = null;
 let nlqChat: ChatOpenAI | ChatMistralAI | AzureChatOpenAI | null = null;
-if (AI_ENABLED && AI_TOKEN) {
+if (AI_TOKEN) {
```
aHenryJard: I think we need to keep the AI_ENABLED but get the value from settings instead of conf.get('ai:enabled');

amiraifori: Thanks for the review!
 
Agreed on keeping an AI enablement gate in ai-llm. I updated the implementation to keep `AI_ENABLED` in [ai-llm.ts], but it is no longer sourced from `conf.get('ai:enabled')`.
Instead, `AI_ENABLED` is now runtime-controlled from [settings.platform_ai_enabled] via a small exported setter [setAiEnabled].

amiraifori:
Hello @aHenryJard 
Hope you are doing alright, would appreciate it if my PR is reviewed.

amiraifori:
Hello @aHenryJard 
Hope you are doing alright, just a quick reminder I would appreciate it if my PR is reviewed.

amiraifori:
Hello @aHenryJard
Hope you are doing alright, just a quick reminder I would appreciate it if my PR is reviewed.

opencti-platform/opencti-graphql/src/database/ai-llm.ts
```diff
@@ -99,12 +111,29 @@ if (AI_ENABLED && AI_TOKEN) {
...
   }
+};
+
+export const setAiEnabled = (enabled: boolean) => {
+  AI_ENABLED = enabled;
+  if (!AI_ENABLED) {
+    resetClients();
+    return;
+  }
+  initClients();
```
aHenryJard: I think this will init client at each AI call instead of once as it was before.

Would be better to init client only if client is null, and reset only if client is not null and AI is switched to disabled.

Could you try that way please ?

aHenryJard (commented):
Would be better to init client only if client is null, and reset only if client is not null and AI is switched to disabled.

I was thinking on init/reset on settings edit field but it will not work on clusters. So the best way that I see to have it dynamically taken is the one you have implemented, but we should find a way to init only once and not at each AI calls.

opencti-platform/opencti-graphql/src/database/ai-llm.ts
amiraifori: I have committed the code to preserve the “dynamic” behaviour but avoid repeatedly reinitialising the clients.

amiraifori:
> Would be better to init client only if client is null, and reset only if client is not null and AI is switched to disabled.
> 
> I was thinking on init/reset on settings edit field but it will not work on clusters. So the best way that I see to have it dynamically taken is the one you have implemented, but we should find a way to init only once and not at each AI calls.

I have committed the updated code to match the request.

amiraifori:
Hello @aHenryJard 
Hope you are doing alright, just a quick reminder I would appreciate it if my PR is reviewed.

amiraifori:
Hello @aHenryJard , @SamuelHassine 
Hope you are doing alright, just a quick reminder I would appreciate it if my PR is reviewed.

amiraifori:
Thanks! Both points are already addressed on the branch:
- duplicate import removed in ai-domain.ts
- AI disabled guard uses `FunctionalError('AI is disabled in platform settings')` consistently

nino-filigran:
@amiraifori Thanks again for you contribution. Can you please rebase your PR, since we delivered a new major version in the meantime?

aHenryJard:
Hello, all AI configuration is going to be removed in OpenCTI in favor of xtm-one, I don't think it's worthy to continue this pull request so I'm closing.

Issue #13767, opened 2025-12-18
Add nx to handle yarn in opencti sub repo


delemaf:
Because life should be simple

yarn build ⇒ it’s building everything needed

yarn dev ⇒ starting what is required

PR #13779, opened 2025-12-19
[oob] add nx to octi/opencti repo (#13767)

delemaf:
### Proposed changes

- Add `nx` as a task orchestrator for the `opencti` monorepo
- Define `nx` projects and targets for existing workspaces (_front_, _graphql_, _worker_, _client-python_)
- Expose unified root commands :
```bash
yarn deps   # install dependencies across projects
yarn build  # build all projects
yarn lint   # lint all projects
yarn test   # run all tests and front e2e 
yarn dev    # start backend, frontend and a worker
```

### Related issues

* OOB 🥷  https://github.com/OpenCTI-Platform/opencti/issues/13767

### Checklist

### Further comments
Nx is introduced in a **non-invasive** way: it only orchestrates existing commands and does not change build logic, runtime behavior, or enable caching. This improves task coordination and maintainability without impacting current workflows.


delemaf:
https://github.com/OpenCTI-Platform/opencti/pull/13256

OctaveLaventure (approved):

client-python/project.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "client-python",
+  "$schema": "../node_modules/nx/schemas/project-schema.json",
+  "sourceRoot": "client-python",
+  "projectType": "application",
+  "types": ["python"],
```
fellowseb: I don't see a definition for a `types` field in the Nx project.json schema 🤔 https://github.com/nrwl/nx/blob/master/packages/nx/schemas/project-schema.json

(Same comment for all project.json files)

client-python/project.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "client-python",
+  "$schema": "../node_modules/nx/schemas/project-schema.json",
+  "sourceRoot": "client-python",
+  "projectType": "application",
```
fellowseb: Shouldn't this be `library` ?

fellowseb (approved):
Looking forward to this. It would actually unlock for instance the capability to introduce additional quality/linting-related stuff like Knip 👍

client-python/project.json
delemaf: You're right 👌

client-python/project.json
delemaf: You're right, tags would have been more appropriate but we don't use them yet, I'll remove this 👍

OctaveLaventure (approved):

PR #13323, opened 2025-11-25
[backend] Playbook upsert operations POC (#6669)

SouadHadjiat:
### Proposed changes

*
*

### Related issues

*
*

### Checklist

- [ ] I consider the submitted work as finished
- [ ] I tested the code for its functionality
- [ ] I wrote test cases for the relevant uses case (coverage and e2e)
- [ ] I added/update the relevant documentation (either on github or on notion)
- [ ] Where necessary I refactored code to improve the overall quality

PR #13784, opened 2025-12-19
[client/backend] implement upsert operation for playbooks (#6669)

### LINKED ISSUE #6669: Ability to remove labels using automation/playbook

SarahBocognano:
### Proposed changes
 
* Add the upsert operation API to all entities on input and client

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/6669

### Linked to
* https://github.com/OpenCTI-Platform/opencti/pull/13323

### Checklist


opencti-platform/opencti-graphql/src/types/stix-2-1-common.d.ts
```diff
@@ -70,8 +70,10 @@ interface StixOpenctiExtension {
   labels_ids: string[];
   created_by_ref_id: string;
   converter_csv?: string | undefined;
-  opencti_operation?: string;
   metrics: Array<StixMetric>;
+  opencti_operation?: string;
+  opencti_field_patch?: EditInput[];
```
xfournet, line 76: to be removed ?

xfournet: can be kept since it was something alread existing but not present in typescript

opencti-platform/opencti-graphql/src/modules/playbook/playbook-components.ts
```diff
@@ -932,9 +957,26 @@ export const PLAYBOOK_ACCESS_RESTRICTIONS_COMPONENT: PlaybookComponent<AccessRes
           entityType: internalType,
           busTopicKey: ABSTRACT_STIX_DOMAIN_OBJECT,
         };
-        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-        // @ts-expect-error
-        await editAuthorizedMembers(context, AUTOMATION_MANAGER_USER, args);
+        const restrictedMembers = await buildRestrictedMembers(context, AUTOMATION_MANAGER_USER, args) as AuthorizedMember[];
```
xfournet, line 963: add test for add and remove of restricted members

SouadHadjiat (approved):

Issue #13785, opened 2025-12-19
Starts_with error in webhooks notifiers


Archidoit:
## Description
Error for webhook notifiers if empty field 

## Reproducible Steps
- Under Settings > Customization, create a notifier of type 'Webhook' with the following template : 

<%
  var inst = (data && data[0] && data[0].instance) ? data[0].instance : {};
%>{
  "source": "OpenCTI",
  "event": "New OpenCTI Report",
  "id": "<%=notification.id%>3",
  "timestamp":"<%= ((+new Date()) / 1000) | 0 %>",
  "reports": [
    {
        "description": "<%= inst.description || '' %>"
    }
  ]
}

<img width="939" height="837" alt="Image" src="https://github.com/user-attachments/assets/429ba59b-f0f4-4ea9-978c-6982d44acb4b" />

An error occurs 

<img width="444" height="245" alt="Image" src="https://github.com/user-attachments/assets/01ec5e09-8caa-43b1-8da7-8bea9c1c75a9" />

## Comment
The error comes from the code that recreate the escape function for json stringification, in case the value is undefined

PR #13789, opened 2025-12-19
[backend] Fix ejs escape option for undefined values (#13785)

Archidoit:
### Proposed changes
Fix safe ejs escape option for undefined values

### Related issues
#13785

### DIFF
--- a/opencti-platform/opencti-graphql/src/utils/safeEjs.worker.ts
+++ b/opencti-platform/opencti-graphql/src/utils/safeEjs.worker.ts
@@ -21,7 +21,7 @@ export const customEscapeFunction = (value: any): string => {
     return '';
   }
   const result = JSON.stringify(value);
-  return typeof result === 'string' && result.startsWith('"') && result.endsWith('"')
+  return result && result.startsWith('"') && result.endsWith('"')
     ? result.slice(1, -1)
     : result;
 };
--- a/opencti-platform/opencti-graphql/src/utils/safeEjs.worker.ts
+++ b/opencti-platform/opencti-graphql/src/utils/safeEjs.worker.ts
@@ -16,6 +16,15 @@ export interface WorkerReply {
   error?: string;
 }
 
+export const customEscapeFunction = (value: any): string => {
+  if (typeof value === 'string') {
+    return value.startsWith('"') && value.endsWith('"')
+      ? value.slice(1, -1)
+      : value;
+  }
+  return value;
+};
+
 // Main worker execution
 const executeWorker = async () => {
   const { template, data, options, useJsonEscape } = workerData as WorkerRequest;
@@ -24,12 +33,7 @@ const executeWorker = async () => {
   const safeEjsOptions = { ...options };
   if (useJsonEscape) {
     // Recreate the escape function for JSON stringification
-    safeEjsOptions.escape = (value: any) => {
-      const result = JSON.stringify(value);
-      return typeof result === 'string' && result.startsWith('"') && result.endsWith('"')
-        ? result.slice(1, -1)
-        : result;
-    };
+    safeEjsOptions.escape = customEscapeFunction;
   }
 
   // Use the core logic from safeEjs (await in case it returns a Promise)
--- a/opencti-platform/opencti-graphql/src/utils/safeEjs.worker.ts
+++ b/opencti-platform/opencti-graphql/src/utils/safeEjs.worker.ts
@@ -26,7 +26,9 @@ const executeWorker = async () => {
     // Recreate the escape function for JSON stringification
     safeEjsOptions.escape = (value: any) => {
       const result = JSON.stringify(value);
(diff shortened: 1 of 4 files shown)


opencti-platform/opencti-graphql/src/utils/safeEjs.worker.ts
```diff
@@ -16,6 +16,15 @@ export interface WorkerReply {
...
 
+export const customEscapeFunction = (value: any): string => {
+  if (typeof value === 'string') {
+    return value.startsWith('"') && value.endsWith('"')
+      ? value.slice(1, -1)
+      : value;
+  }
+  return value;
+};
+
```
JeremyCloarec, line 28: The function is missing the JSON.stringify(value)

xfournet: yes we should stringify all types of values, not only strings.

Archidoit: fixed :)

xfournet: in the original code, it's the result to be check if it's undefined, probably just adding a result !== undefined && .... should be enough

opencti-platform/opencti-graphql/tests/01-unit/utils/safeEjs-test.ts
```diff
@@ -327,9 +329,11 @@ describe('check safeRender on real files', () => {
     async ({ name }) => {
       const templateFile = `${testFilename.substring(0, testFilename.lastIndexOf('.'))}.${name}`;
       const template = await fs.readFile(templateFile, 'utf8');
-      const safeRendered = safeRender(template, data, { useNotificationTool: true });
+      const jsonFilesTemplates = ['template-5', 'template-6', 'template-7'];
+      const escape = jsonFilesTemplates.some((f) => name.includes(f)) ? customEscapeFunction : undefined;
```
xfournet: you may test if if `name` ends with `.json` instead of building a list a json files

Archidoit: oh indeed, it's silly how I coded this! Thanks

opencti-platform/opencti-graphql/src/utils/safeEjs.worker.ts
```diff
@@ -16,6 +16,16 @@ export interface WorkerReply {
   error?: string;
 }
 
+export const customEscapeFunction = (value: any): string => {
+  if (Array.isArray(value) && value.length === 0) {
```
xfournet, line 20: why handling this case specifically ? was it done in this way before safe EJS ?

opencti-platform/opencti-graphql/tests/01-unit/utils/safeEjs-test.ts
```diff
@@ -327,9 +329,11 @@ describe('check safeRender on real files', () => {
     async ({ name }) => {
       const templateFile = `${testFilename.substring(0, testFilename.lastIndexOf('.'))}.${name}`;
       const template = await fs.readFile(templateFile, 'utf8');
-      const safeRendered = safeRender(template, data, { useNotificationTool: true });
+      const jsonFilesTestCases = fileTestCases.filter((fileName) => fileName.endsWith('.json'));
+      const escape = jsonFilesTestCases.some((f) => name.includes(f)) ? customEscapeFunction : undefined;
```
xfournet: you don't need to build the intermediate list, just test the name directly ;)

opencti-platform/opencti-graphql/src/utils/safeEjs.worker.ts
Archidoit: It was to fix an error happening on test template-6. The error was : 
image

opencti-platform/opencti-graphql/tests/01-unit/utils/safeEjs-test.ts
Archidoit: Indeed :) Fixed! thanks :)

opencti-platform/opencti-graphql/src/utils/safeEjs.worker.ts
xfournet: i'm not sure to understand the expected behavior, let's talk about it

Archidoit: It's to align with the classical behavior. To Do: check the former behavior with empty arrays before our ejs improvements

Archidoit: @xfournet  The former behavior was the one with nothing instead of empty arrays : 
image
 So I guess I can keep the part of code I added for arrays

xfournet (approved):

Issue #13837, opened 2025-12-26
Knowledge Graph widget includes Has Label meta-relations


Lhorus6:
## Description

When creating a widget with the Knowledge Graph perspective, it appears that the widget also includes "Has Label" relations. These are meta-relations between objects and labels and are not STIX relationships. Is it expected ?

If so, currently, there is no way to filter them using the "Relationship Type" filter. The value "Has Label" does not appear in the filter, so it cannot be excluded or included explicitly.

This causes issues, especially when trying to list relationships but excluding certain types. In such cases, "Has Label" relations appear, impacting the expected results of the widget.

## Environment

OCTI 6.9.3

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Create an Intrusion set "Test rel" with few relations and few labels
2. Create a List widget with a Knowledge Graph perspective and those filters:

<img width="1201" height="651" alt="Image" src="https://github.com/user-attachments/assets/5170f5da-fa7b-4829-874e-6cb7801cc203" />

3. Observe that "Has Label" meta-relations appear in the widget.

<img width="3130" height="652" alt="Image" src="https://github.com/user-attachments/assets/362915bc-2df3-4fd9-b5da-7cc237b5a4b2" />

NB: There is no way to filter on "Has Label" with the "Relationship Type" filter:

<img width="378" height="372" alt="Image" src="https://github.com/user-attachments/assets/0a5b8c81-5b97-4244-bbf4-a4180a1b838e" />

## Expected Output

- Either the "Has Label" meta-relations are not taken into account, or 
- Ability to include or exclude "Has Label" meta-relations via the "Relationship Type" filter, just like other relationship types.

Archidoit:
In relationships widgets, the relationships displayed are : 
- stix core relationships
- sightings
- the 'contains' relationships between a container and an entity
- the 'has label' relationships

Indded, a solution would be to add the 'has label' value in the relationship type and entity type filters.

nino-filigran:
Yes, removing it via filter would be the best way forward: we should have "has label" relation proposed as a value of the filter.

PR #13859, opened 2025-12-31
[frontend] Add has-label in relationship type filter values for knowledge widgets (#13837)

Archidoit:
### Proposed changes
Add the 'has label' relationship type in the values of the relationship type filter for knowledge widget : 

image

### Related issues
#13837


CelineSebe:
Tested locally. Works well :)

CelineSebe (approved):

PR #13865, opened 2026-01-01
[backend/frontend] JSON mapper: Support multiple identifiers definition for based_on attributes (#13863)

### LINKED ISSUE #13863: [JSON mapper] Support multiple identifiers definition for based_on attributes
## Use case

For based_on attributes like report **contains**, multiple identifiers need to be supported.
Current identifier is single and support composite building only.

To be able to create a report containing all different needed elements, targetted reprensentations need to be selected according to correct identifier.

richard-julien:
# Feature: JSON Mapper Complex Identifiers

Closes #13863 

## 🎯 Objective
This update refines the JSON Mapper to support explicit binding between representation IDs and their corresponding JSON path identifiers. Previously, the system separated identifiers and representations into two disconnected lists, which created ambiguity when multiple representations were selected. Now, each representation can have its own specific identifier JSON path.

### Backend (Core Engine)
*   **GraphQL Schema (`jsonMapper.graphql`)**:
    *   Introduced `JsonIdentifier` type: `{ identifier: String, representation: String }`.
    *   Updated `JsonAttributeBasedOn` to potentially use `[JsonIdentifier]` for strict typing of complex identifiers.
*   **Types (`jsonMapper-types.ts`)**:
    *   Updated `AttributeBasedOn` interface to allow `identifier` to be either:
        *   `string` (Legacy support for simple comma-separated identifiers).
        *   `AttributeBasedOnIdentifierComplex[]` (New structure: `[{ identifier, representation }]`).
*   **Parser Logic (`json-mapper.ts`)**:
    *   Refactored `handleBasedOnAttribute` to support key-based identifier resolution.
    *   Implemented logic to iterate through selected `representations` and look up the specific identifier path defined for each one in the `identifier` array.
    *   Maintained backward compatibility: If `identifier` is a simple string, it parses it using the legacy logic (splitting by comma).

### Frontend (UI)
*   **Form Logic (`JsonMapperRepresentationAttributeRefForm.tsx`)**:
    *   **Representation Selection**: Users select representations using a multi-select combobox.
    *   **Dynamic Identifier Inputs**: The UI now iterates over the selected representations and renders a dedicated "JSON Path" text field for each one.
    *   **Synchronization**: Logic added to automatically sync the `identifier` state with the selected `representations`. When a representation is added, a new blank identifier object is created. When removed, its identifier is cleaned up.
*   **Data Transformation (`AttributeUtils.ts`)**:
    *   Updated data mapping functions (`jsonMapperAttributeToFormData`, `formDataToJsonMapperAttribute`) to correctly serialize/deserialize the new complex identifier structure between the frontend form state and the backend API.
    *   Ensures that saving the form sends the correct `[{ identifier, representation }]` structure to the server.

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/data/jsonMapper/representations/attributes/Attribute.d.ts
+++ b/opencti-platform/opencti-front/src/private/components/data/jsonMapper/representations/attributes/Attribute.d.ts
@@ -14,6 +14,11 @@ export interface JsonAttrPathConfiguration {
   timezone: string | null | undefined;
 }
 
+export type BasedOnType = {
+  identifier?: { identifier: string; representation: string | null }[] | null;
+  representations?: string[] | undefined | null;
+} | null;
+
 export interface JsonMapperRepresentationAttributeFormData {
   key: string;
   mode: string; // 'simple' | ' complex' | 'base'
@@ -25,9 +30,6 @@ export interface JsonMapperRepresentationAttributeFormData {
   complex_path?: {
     formula: string;
   };
-  based_on?: {
-    identifier?: { identifier: string; representation: string | null }[] | null;
-    representations?: string[] | undefined | null;
-  } | null;
+  based_on?: BasedOnType;
   default_values?: DefaultValues;
 }
--- a/opencti-platform/opencti-front/src/private/components/data/jsonMapper/representations/attributes/Attribute.d.ts
+++ b/opencti-platform/opencti-front/src/private/components/data/jsonMapper/representations/attributes/Attribute.d.ts
@@ -26,7 +26,7 @@ export interface JsonMapperRepresentationAttributeFormData {
     formula: string;
   };
   based_on?: {
-    identifier?: string | null;
+    identifier?: { identifier: string; representation: string | null }[] | null;
     representations?: string[] | undefined | null;
   } | null;
   default_values?: DefaultValues;
--- a/opencti-platform/opencti-front/src/private/components/data/jsonMapper/representations/attributes/Attribute.d.ts
+++ b/opencti-platform/opencti-front/src/private/components/data/jsonMapper/representations/attributes/Attribute.d.ts
@@ -14,7 +14,7 @@ export interface JsonAttrPathConfiguration {
   timezone: string | null | undefined;
 }
 
-export type BasedOnType = {
+export type BasedOn = {
   identifier?: { identifier: string; representation: string | null }[] | null;
   representations?: string[] | undefined | null;
 } | null;
@@ -30,6 +30,6 @@ export interface JsonMapperRepresentationAttributeFormData {
   complex_path?: {
     formula: string;
   };
-  based_on?: BasedOnType;
+  based_on?: BasedOn;
   default_values?: DefaultValues;
 }
--- a/opencti-platform/opencti-front/src/private/components/data/jsonMapper/representations/attributes/JsonMapperRepresentationAttributeRefForm.tsx
+++ b/opencti-platform/opencti-front/src/private/components/data/jsonMapper/representations/attributes/JsonMapperRepresentationAttributeRefForm.tsx
@@ -155,10 +155,10 @@ const JsonMapperRepresentationAttributeRefForm: FunctionComponent<
     value: JsonMapperRepresentationFormData[],
   ) => {
     const selectedIds = value.map((v) => v.id);
-    const textIdentifiers = (field.value?.based_on?.identifier ?? []).filter((i) => typeof i !== 'string');
+    const nonStrIdentifiers = (field.value?.based_on?.identifier ?? []).filter((i) => typeof i !== 'string');
 
     // Add new identifiers for newly selected representations
-    const newIdentifiers = [...textIdentifiers];
+    const newIdentifiers = [...nonStrIdentifiers];
     selectedIds.forEach((id) => {
       if (!newIdentifiers.find((i) => i.representation === id)) {
         newIdentifiers.push({ identifier: '', representation: id });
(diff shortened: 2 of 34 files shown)


richard-julien:
Need to switch the approach to have identifier really related to a specific entity selection to prevent identifier mixin.
In the case with contains. Malware + Campaign we need to have
- Identifier for Malware
- Identifier for Campaign

opencti-platform/opencti-front/src/private/components/data/jsonMapper/representations/attributes/JsonMapperRepresentationAttributeRefForm.tsx
```diff
@@ -198,79 +211,75 @@ const JsonMapperRepresentationAttributeRefForm: FunctionComponent<
...
+            );
+          }
+
+          return selectedRepresentations.map((rep) => {
+            const identifiers = field.value?.based_on?.identifier ?? [];
+            const identifierObj = identifiers.find((i) => i.representation === rep.id);
+            const identifierValue = identifierObj?.identifier ?? '';
+            return (
+              <React.Fragment key={rep.id}>
+                <div>Identifier ({representationLabel(entity_representations.indexOf(rep), rep, t_i18n)})</div>
```
lndrtrbn: Is it expected behavior that `Identifier` here is not translated?

opencti-platform/opencti-graphql/src/modules/internal/jsonMapper/jsonMapper-resolvers.ts
```diff
@@ -1,18 +1,3 @@
...
-licensed under the OpenCTI Enterprise Edition License (the "License");
-you may not use this file except in compliance with the License.
-You may obtain a copy of the License at
-
-https://github.com/OpenCTI-Platform/opencti/blob/master/LICENSE
-
-Unless required by applicable law or agreed to in writing, software
-distributed under the License is distributed on an "AS IS" BASIS,
-WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-*/
```
lndrtrbn, line 14: Is it expected to remove copyrights on this file but not on the other files of the module?

opencti-platform/opencti-graphql/src/modules/internal/jsonMapper/jsonMapper-utils.ts
```diff
@@ -47,6 +48,19 @@ export const parseJsonMapper = (mapper: any): JsonMapperParsed => {
...
+    representations.forEach((rep) => {
+      rep.attributes.forEach((attr) => {
+        if (attr.mode === 'base' && !Array.isArray(attr.based_on.identifier)) {
+          attr.based_on.identifier = (attr.based_on.representations ?? []).map((r) => {
+            return { identifier: attr.based_on.identifier, representation: r } as AttributeBasedOnIdentifierComplex;
+          });
+        }
+      });
+    });
+  }
```
lndrtrbn, line 47: Is it for retro-compat with old format?

opencti-platform/opencti-front/src/private/components/data/jsonMapper/representations/attributes/JsonMapperRepresentationAttributeRefForm.tsx
richard-julien: Just forgot i think

opencti-platform/opencti-graphql/src/modules/internal/jsonMapper/jsonMapper-resolvers.ts
richard-julien: We can remove it. json mapper is CE

opencti-platform/opencti-graphql/src/modules/internal/jsonMapper/jsonMapper-utils.ts
richard-julien: Yes

lndrtrbn:
~Form validation seems broken because we can now add multiple representations in an attribute not defined as `multiple`~

EDIT: The issue was probably between chair and keyboard

nino-filigran:
@richard-julien I checked the PR and I did not see any regression: quick question though, I've noticed it's not possible to add relations in the "contains" section. I end up in having containers with multiple entities, but without relations. 
Do you think this should be added in context of this PR (it was already a limitation before your PR, not a behavior introduced by it).

opencti-platform/opencti-front/src/private/components/data/jsonMapper/representations/attributes/Attribute.d.ts
```diff
@@ -14,6 +14,11 @@ export interface JsonAttrPathConfiguration {
   timezone: string | null | undefined;
 }
 
+export type BasedOnType = {
```
fellowseb: ultra nit: adding the `Type` suffix seems useless for a type

opencti-platform/opencti-front/src/private/components/data/jsonMapper/representations/attributes/JsonMapperRepresentationAttributeRefForm.tsx
```diff
@@ -155,36 +150,54 @@ const JsonMapperRepresentationAttributeRefForm: FunctionComponent<
     }
   }, [errors]);
 
-  const onIdentifierValueChange = async (val: SelectChangeEvent) => {
+  const handleRepresentationChange = async (
+    event: React.SyntheticEvent,
+    value: JsonMapperRepresentationFormData[],
+  ) => {
+    const selectedIds = value.map((v) => v.id);
+    const textIdentifiers = (field.value?.based_on?.identifier ?? []).filter((i) => typeof i !== 'string');
```
fellowseb: nit: maybe `textIdentifiers` is not the best choice of name given we exclude `string` values ?

fellowseb: The type `BasedOnType` doesn't seem to indicate the possibility that `based_on.identifier` could be an array of strings. Consider updating the type.

fellowseb (approved):

opencti-platform/opencti-front/src/private/components/data/jsonMapper/representations/attributes/JsonMapperRepresentationAttributeRefForm.tsx
lndrtrbn: In fact the strings we can have are from the config format before this refactoring. But the objective now is to not have strings, I think this code is to stay retro-compatible

fellowseb (approved):

Issue #13571, opened 2025-12-08
ci: publish rolling and prerelease images and deploy to staging


efaure:
## Use case

Currently docker builds done in Circle CI for rolling, prerelease and release images use old Dockerfile that can be improved. 

Docker build has already been improved for feature branch: https://github.com/OpenCTI-Platform/opencti/pull/12924

See if it's possible to use the same Dockerfile for fips images

## Scope of the issue

 - Build docker images on commit on master and release current and publish them to docker registry
   - rolling and prerelease images
   - alpine and fips images
 - Build packages classic (debian) and musl (alpine)
 - Deploy on staging environments:
   - On commit on master, deploy rolling image on testing, dev and  prerelease CE (GCP) 
   - On commit on prerelease, deploy prerelease on prerelease aws

PR #13871, opened 2026-01-02
ci: publish rolling and prerelease images and deploy to staging

efaure:
### Proposed changes

 - Build docker images on commit on master and release current and publish them to docker registry
   - rolling and prerelease images
   - alpine and fips images
 - Use the same dockerfile as the one used for feature branch for alpine and fips images
 - Build packages classic (debian) and musl (alpine)
 - Deploy on staging environments:
   - On commit on master, deploy rolling image on testing, dev and  prerelease CE (GCP) 
   - On commit on prerelease, deploy prerelease on prerelease aws

### TO DO 

Some configuration are done in order to ease test for this PR but should be modified before merge: 

 - change destination registry to push docker hub public repository (opencti) instead of internal (filigran) => To be done when release migration is also ready
 - action trigger on push on master and release instead of current branch => To be done when release migration is also ready
 - rename Dockerfile_featurebranch to `Dockerfile` when the switching from Circle CI to GitHub action. 
 - ~~remove build based on current branch (issue/13571-build-images) to build image test-13571.~~ Done: 9f5edab1fb3ad69c82a481607bf4a42580e07601 
 - ~~remove deployment on cicd-test.~~ Done: t 9f5edab1fb3ad69c82a481607bf4a42580e07601

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/13571


efaure:
@aHenryJard  I would be happy to have your review on this. 

This PR is just a preparation step, but I would like to have a review on it before moving forward. 
There is no trigger, so it won't start publishing / deploying anything on master merge. 

The scope of this one is to build of images, packages an deployment to staging for rollout (master) and prerelease (release/current) images. I plan to do an other PR for "release" workflow. 

I would like to have your opinion on one point: 
 - I separated the "publish and deploy" main workflow from the "OpenCTI CI" existing one. It's also possible to merge those 2 main workflow. Do you think it will ease the view for dev to have everything in the same workflow, or that it will be confusing to have too many steps in it ? If unsure, I can test to merge to see the result. 

Some changes may be added later (after confirmation we want to do it):
 - Publication of artifact to jfrog (the code is already here, just need to be uncommented and activated, and maybe to create a new jfrog token)
 - Currently Circle CI "build, publish, deploy" flow start without waiting for CI test. For now I kept it with the same flow but we may change it to wait  for the end of test before publication and deployment.

efaure:
@xfournet 
I'm interested to have your opinion on the Dockerfile changes in this PR https://github.com/OpenCTI-Platform/opencti/blob/cefa4f86a6e4e443570169fe623b95710f319109/opencti-platform/Dockerfile_featurebranch

I updated the Dockerfile we previously worked on in order to use only one Dockerfile for all usage. 
It's still named "Dockerfile_feautrebranch" as I didn't want to change the one currently used by circle CI.

A small explanation of the Dockerfile: 
 - Switch to use fips image. 
   The default is to use "alpine" image but with a`--build-arg BASE_TYPE=fips` it will use the fips as a base.
   Depending of the base type, node option and some packages to install are different. I added this as Env var in each base image.
- Build for images used in CI
  To run tests in images, we need to keep dependencies, so I added (in a previous PR) a `testing` target. 
  By default, a docker build won't build it (as it's not used in the final image) but with `--target testing` option it will build only the testing image.

.github/workflows/cd-publish-and-deploy.yml
```diff
@@ -0,0 +1,79 @@
...
+
+      - name: Deploy on testing
+        if: ${{ github.ref_name == 'master' }}
+        run: |
+          awx job_templates launch 'Rollout restart an instance resources' --wait -f human \
+            --inventory eu-west-staging \
+            --extra_vars '{"customer":"testing","solution": "opencti" }'
+
+      - name: Deploy on dev
+        if: ${{ github.ref_name == 'master' }}
```
aHenryJard: dev is based on release/current today

.github/workflows/cd-publish-package.yml
```diff
@@ -0,0 +1,137 @@
...
+    - name: Build OpenCTI on debian
+      run: |
+        docker run --name build-package \
+          -v ${{ github.workspace }}:/home/workspace \
+          nikolaik/python-nodejs:python3.11-nodejs22 \
+          sh -c "set -eux
+           cd /home/workspace
+           apt-get update --allow-insecure-repositories --allow-unauthenticated && apt-get install -y build-essential libffi-dev curl g++ make python3 python3-dev
+           npm install -g node-gyp           
+           cd opencti-platform/opencti-graphql
```
aHenryJard: I don't see a yarn install & build inside opencti-front folder so I don't understand for now when the react app is build/minified in the process.

efaure: You're right ! I wanted to migrate the step as it is in https://github.com/OpenCTI-Platform/opencti/blob/master/.circleci/config.yml#L176
But I missed than this step is using attaching the workspace already persiste in the `build_frontend` step. I'll update it.

labo-flg (approved):

opencti-worker/Dockerfile
xfournet: Some deps may be useless ?

```suggestion
RUN apk --no-cache add git libmagic \
    && pip3 install --no-cache-dir --requirement requirements.txt \
    && apk del git
```

or

```suggestion
RUN apk --no-cache add libmagic \
    && pip3 install --no-cache-dir --requirement requirements.txt
```

opencti-platform/Dockerfile_featurebranch
```diff
@@ -1,5 +1,20 @@
...
+### Create base image fips or alpine
+##############################
+
+ARG BASE_TYPE=alpine
+ARG BASE_IMAGE_ALPINE=node:22-alpine
+ARG BASE_IMAGE_FIPS=filigran/python-nodejs-fips:latest
+
+FROM ${BASE_IMAGE_ALPINE} AS base-alpine
+ENV EXTRA_NODE_OPTIONS=""
+ENV EXTRA_PACKAGES="python3 python3-dev g++"
```
xfournet: is g++ needed ?

.github/workflows/cd-publish-package.yml
```diff
@@ -0,0 +1,121 @@
...
+      run: find . -name requirements.txt -exec sed 's|^pycti==.*$|pycti @ git+https://github.com/OpenCTI-Platform/opencti@${{ github.ref_name }}#subdirectory=client-python|' -i {} \;
+
+    - name: Build OpenCTI on debian
+      run: |
+        docker run --name build-package \
+          -v ${{ github.workspace }}:/home/workspace \
+          nikolaik/python-nodejs:python3.11-nodejs22 \
+          sh -c "set -eux
+           apt-get update --allow-insecure-repositories --allow-unauthenticated && apt-get install -y build-essential libffi-dev curl g++ make python3 python3-dev
+           npm install -g node-gyp
```
xfournet: not needed ?

.github/workflows/cd-publish-package.yml
```diff
@@ -0,0 +1,121 @@
...
+      if: ${{ inputs.is_client_python_local == true }}
+      run: find . -name requirements.txt -exec sed 's|^pycti==.*$|pycti @ git+https://github.com/OpenCTI-Platform/opencti@${{ github.ref_name }}#subdirectory=client-python|' -i {} \;
+
+    - name: Build OpenCTI on debian
+      run: |
+        docker run --name build-package \
+          -v ${{ github.workspace }}:/home/workspace \
+          nikolaik/python-nodejs:python3.11-nodejs22 \
+          sh -c "set -eux
+           apt-get update --allow-insecure-repositories --allow-unauthenticated && apt-get install -y build-essential libffi-dev curl g++ make python3 python3-dev
```
xfournet: do you test if every of theses packages are needed ?

.github/workflows/cd-publish-package.yml
```diff
@@ -0,0 +1,121 @@
...
+           yarn install
+           yarn build
+        "
+
+    - name: Prepare OpenCTI debian package
+      run: |
+        set -eux
+        sudo apt-get update -qq && sudo apt install curl
+        mkdir release
+        rm -Rf ./opencti-platform/opencti-graphql/.yarn/cache
```
xfournet: useless, there is no more files here

.github/workflows/cd-publish-package.yml
```diff
@@ -0,0 +1,121 @@
...
+      if: ${{ inputs.is_client_python_local == true }}
+      run: find . -name requirements.txt -exec sed 's|^pycti==.*$|pycti @ git+https://github.com/OpenCTI-Platform/opencti@${{ github.ref_name }}#subdirectory=client-python|' -i {} \;
+
+    - name: Build OpenCTI on alpine musl
+      run: |
+        docker run --name build-package \
+          -v ${{ github.workspace }}:/home/workspace \
+          nikolaik/python-nodejs:python3.11-nodejs22-alpine \
+          sh -c "set -eux
+           apk update && apk upgrade && apk --no-cache add git tini gcc g++ make musl-dev python3 python3-dev postfix postfix-pcre build-base libmagic libffi-dev curl
```
xfournet: to be removed:
* certainly : postfix postfix-pcre
* probably : tini curl
* maybe other to be tested

.github/workflows/cd-publish-package.yml
```diff
@@ -0,0 +1,121 @@
...
+           yarn install
+           yarn build
+        "
+
+    - name: Prepare OpenCTI musl package
+      run: |
+        set -eux
+        sudo apt-get update -qq && sudo apt install curl
+        mkdir release
+        rm -Rf ./opencti-platform/opencti-graphql/.yarn/cache
```
xfournet: can be removed

.github/workflows/cd-publish-package.yml
efaure: Removed  `postfix postfix-pcre tini curl`  and also  `build-base libmagic libffi-dev musl-dev gcc`. 
Fixed in 64d3da3d03817a9e4eb019058d898a18c6454b43

.github/workflows/cd-publish-package.yml
efaure: I took those step "as is" from Circle CI build steps without challenging them.
Now, I tested, and indeed `build-essential libffi-dev curl `are not needed. Fixed in 64d3da3d03817a9e4eb019058d898a18c6454b43
Git is needed to build packages for master version (git used in requirements.txt). (readded in e40581373416a4f5cb1f0612db05812343af63c9)

.github/workflows/cd-publish-package.yml
efaure: It seems so... Removed in 64d3da3d03817a9e4eb019058d898a18c6454b43

opencti-platform/Dockerfile_featurebranch
efaure: Indeed, it's not needed here. It's already present in packages used for build, and not needed in app image. Fixed in 2620452f59680ff83860e34c37f39d968d4c3808

opencti-worker/Dockerfile
efaure: Indeed. I removed build-base and libffi-dev. Fixed in  6f9bb45701150449b4ffeb3c9c6e5b1a8ce2d2a9
Git is needed when to build requirements when pycti is retrieved from master. (readded in c99afc17cb5499fba1949a29fa283034674846ea)

aHenryJard:
Done with https://github.com/OpenCTI-Platform/opencti/pull/17072

PR #13883, opened 2026-01-05
[backend] refresh label/marking in resolved filter cache when it is updated (#13880)

### LINKED ISSUE #13880: Playbooks and live triggers are not working after modifying a label
## Description

Playbooks and live triggers are not working after modifying a label

Tested both on testing and locally.

1. Create a live trigger Entity type = Report and label = test listening all events
2. Create a new report with label = test
=> Notification emmitted

3. Modify the label in taxonomy, and put a completely different value value like "pop".
4. Create a new report with label = pop
=> No notification

Same with playbooks, may also be the same with live streams etc. Hence critical.

JeremyCloarec:
### Proposed changes

* refresh label in resolved filter cache when label is updated
* refresh markings in resolved filter cache when marking is updated

### Related issues

* #13880 
*

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/database/cache.ts
+++ b/opencti-platform/opencti-graphql/src/database/cache.ts
@@ -14,15 +14,17 @@ import { type BasicStoreEntityPublicDashboard, ENTITY_TYPE_PUBLIC_DASHBOARD } fr
 import { wait } from './utils';
 import { ENTITY_TYPE_PIR } from '../modules/pir/pir-types';
 import { ENTITY_TYPE_DECAY_EXCLUSION_RULE } from '../modules/decayRule/exclusions/decayExclusionRule-types';
+import { ENTITY_TYPE_LABEL } from '../schema/stixMetaObject';
 
 const STORE_ENTITIES_LINKS: Record<string, string[]> = {
-  // Resolved Filters in cache must be reset depending on connector/stream/triggers/playbooks/Pir modifications
+  // Resolved Filters in cache must be reset depending on connector/stream/triggers/playbooks/Pir/label modifications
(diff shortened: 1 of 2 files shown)


SamuelHassine:
@JeremyCloarec are we sure we don't have the same issue with marking definitions?

JeremyCloarec:
> @JeremyCloarec are we sure we don't have the same issue with marking definitions?

You're right, updating a marking description modifies it's standard id, so it needs to be updated in filters cache also. 
I'm pushing an other commit to add markings handling

opencti-platform/opencti-graphql/src/database/cache.ts
```diff
@@ -14,15 +14,18 @@ import { type BasicStoreEntityPublicDashboard, ENTITY_TYPE_PUBLIC_DASHBOARD } fr
 import { wait } from './utils';
 import { ENTITY_TYPE_PIR } from '../modules/pir/pir-types';
 import { ENTITY_TYPE_DECAY_EXCLUSION_RULE } from '../modules/decayRule/exclusions/decayExclusionRule-types';
+import { ENTITY_TYPE_LABEL, ENTITY_TYPE_MARKING_DEFINITION } from '../schema/stixMetaObject';
 
 const STORE_ENTITIES_LINKS: Record<string, string[]> = {
-  // Resolved Filters in cache must be reset depending on connector/stream/triggers/playbooks/Pir modifications
+  // Resolved Filters in cache must be reset depending on connector/stream/triggers/playbooks/Pir/label modifications
```
Archidoit, line 20:
```suggestion
  // Resolved Filters in cache must be reset depending on connector/stream/triggers/playbooks/Pir/label/markings modifications
```

Archidoit (approved):

Issue #13246, opened 2025-11-20
Relationships representative not displayed in Indicator Knowledge


Archidoit:
## Description
In an indicator knowledge tab, nothing is displayed in the 'name' column for relationships.

## Reproducible Steps
- Go to a relationship and add an indicator in 'linked entities'

<img width="1806" height="813" alt="Image" src="https://github.com/user-attachments/assets/ab64a883-9f99-4c3e-b0ae-c6feacfa3cf5" />

- Go to the Indicator knowledge tab. Nothing is displayed in the 'name' column

<img width="1837" height="452" alt="Image" src="https://github.com/user-attachments/assets/9be0117f-3ce2-4910-8894-35f7287b2efd" />

## Expected Output
Display the relationship representative in the name column

Take exemple on how to display relationship representative in Data>Entities (see a relationship between an entity and a relationship) : 

<img width="1821" height="337" alt="Image" src="https://github.com/user-attachments/assets/f0bf949d-8dac-44d1-ad7e-b5988a9b8181" />

## Actual Output
Nothing displayed in the 'name' column

nino-filigran:
Agreed, we should see the representative of the relation in the name.

PR #13887, opened 2026-01-05
[frontend] Relationships representative not displayed in Indicator Knowledge (#13246)

Gwendoline-FAVRE-FELIX:
### Proposed changes

* Fix the display of relationship names in the indicator relations list by using the relationship representative.

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/13246

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/observations/indicators/IndicatorEntityLine.jsx
+++ b/opencti-platform/opencti-front/src/private/components/observations/indicators/IndicatorEntityLine.jsx
@@ -63,6 +63,18 @@ class IndicatorEntityLineComponent extends Component {
     const element = node.to?.id === entityId ? node.from : node.to;
     const restricted = isEmptyField(element);
     const link = `${entityLink}/relations/${node.id}`;
+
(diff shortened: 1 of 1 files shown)


opencti-platform/opencti-front/src/private/components/observations/indicators/IndicatorEntityLine.jsx
```diff
@@ -63,6 +63,18 @@ class IndicatorEntityLineComponent extends Component {
     const element = node.to?.id === entityId ? node.from : node.to;
     const restricted = isEmptyField(element);
     const link = `${entityLink}/relations/${node.id}`;
+
+    const isRelationship = element.__typename === 'StixCoreRelationship';
```
JeremyCloarec: We usually either use isRelationship from useSchema, or this kind of condition: const isRelationship = t_i18n(`relationship_${entityType}`) !== `relationship_${entityType}`;
Was there a reason to use your condition instead of the condition above?

JeremyCloarec (approved):

PR #14034, opened 2026-01-15
[backend/client] add ruleApplyAsync method (#11626)

### LINKED ISSUE #11626: Change rule activation in background tasks

JeremyCloarec:
### Proposed changes

* add ruleApplyAsync method to handle long running ruleApply calls
*

### Related issues

* #11626 
*

### Checklist

### DIFF
--- a/client-python/pycti/api/opencti_api_work.py
+++ b/client-python/pycti/api/opencti_api_work.py
@@ -243,7 +243,10 @@ class OpenCTIApiWork:
         :return: empty string if error, None otherwise
         :rtype: str or None
         """
+        state = self.get_work(work_id=work_id)
         status = ""
+        if len(state) > 0:
+            status = state["status"]
         while status != "complete":
             state = self.get_work(work_id=work_id)
             if len(state) > 0:
--- a/client-python/pycti/api/opencti_api_work.py
+++ b/client-python/pycti/api/opencti_api_work.py
@@ -243,10 +243,7 @@ class OpenCTIApiWork:
         :return: empty string if error, None otherwise
         :rtype: str or None
         """
-        state = self.get_work(work_id=work_id)
         status = ""
-        if len(state) > 0:
-            status = state["status"]
         while status != "complete":
             state = self.get_work(work_id=work_id)
             if len(state) > 0:
@@ -257,7 +254,8 @@ class OpenCTIApiWork:
                         "Unexpected connector error", {"state_errors": state["errors"]}
                     )
                     return ""
-
+            if status == "complete":
+                return
             time.sleep(1)
 
     def get_work(self, work_id: str) -> Dict:
--- a/client-python/pycti/entities/opencti_stix_core_object.py
+++ b/client-python/pycti/entities/opencti_stix_core_object.py
@@ -1,6 +1,5 @@
 # coding: utf-8
 import json
-from time import sleep
 
 
 class StixCoreObject:
--- a/client-python/pycti/entities/opencti_stix_core_object.py
+++ b/client-python/pycti/entities/opencti_stix_core_object.py
@@ -1,6 +1,5 @@
 # coding: utf-8
 import json
-import time
 
 
 class StixCoreObject:
--- a/client-python/pycti/entities/opencti_stix_core_object.py
+++ b/client-python/pycti/entities/opencti_stix_core_object.py
@@ -1899,7 +1899,8 @@ class StixCoreObject:
         if element_id is not None and rule_id is not None and execution_id is not None:
             while not rule_apply_complete:
                 self.opencti.app_logger.info(
-                    "Apply rule async stix_core_object", {"id": element_id, "execution_id": execution_id}
+                    "Apply rule async stix_core_object",
+                    {"id": element_id, "execution_id": execution_id},
                 )
                 query = """
                         mutation StixCoreApplyRule($elementId: ID!, $ruleId: ID!, $executionId: ID!) {
--- a/client-python/pycti/entities/opencti_stix_core_object.py
+++ b/client-python/pycti/entities/opencti_stix_core_object.py
@@ -1,7 +1,6 @@
 # coding: utf-8
 import json
 import time
-import uuid
 
 
 class StixCoreObject:
--- a/client-python/pycti/entities/opencti_stix_core_object.py
+++ b/client-python/pycti/entities/opencti_stix_core_object.py
@@ -1899,7 +1899,7 @@ class StixCoreObject:
         if element_id is not None and rule_id is not None and execution_id is not None:
             while not rule_apply_complete:
                 self.opencti.app_logger.info(
-                    "Apply rule async stix_core_object", {"id": element_id}
+                    "Apply rule async stix_core_object", {"id": element_id, "execution_id": execution_id}
                 )
                 query = """
                         mutation StixCoreApplyRule($elementId: ID!, $ruleId: ID!, $executionId: ID!) {
@@ -1915,9 +1915,9 @@ class StixCoreObject:
                     },
                 )
                 rule_apply_complete = result["data"]["ruleApplyAsync"]
-                self.opencti.app_logger.info(
-                    "Apply rule async stix_core_object complete", {"id": element_id}
-                )
+            self.opencti.app_logger.info(
+                "Apply rule async stix_core_object complete", {"id": element_id}
+            )
         else:
             self.opencti.app_logger.error(
                 "[stix_core_object] Cannot apply rule, missing parameters: id"
--- a/client-python/pycti/entities/opencti_stix_core_object.py
+++ b/client-python/pycti/entities/opencti_stix_core_object.py
@@ -1,5 +1,6 @@
 # coding: utf-8
 import json
+from time import sleep
 
 
 class StixCoreObject:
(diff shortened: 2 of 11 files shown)


opencti-platform/opencti-graphql/src/database/redis.ts
```diff
@@ -506,7 +506,19 @@ export const redisInitializeWork = async (workId: string) => {
   });
 };
 // endregion
-
+// region async calls tracking
+const ASYNC_CALL_TTL = 3600;
+const ASYNC_CALL_KEY_PREFIX = 'async_status_';
+export const redisInitializeAsyncCall = async (asyncCallId: string) => {
+  await getClientBase().setex(ASYNC_CALL_KEY_PREFIX + asyncCallId, ASYNC_CALL_TTL, '0');
```
xfournet: setex is documented as deprecated in Redis (https://redis.io/docs/latest/commands/setex/) we should use set with ex option instead

opencti-platform/opencti-graphql/src/database/redis.ts
```diff
@@ -506,7 +506,19 @@ export const redisInitializeWork = async (workId: string) => {
   });
 };
 // endregion
-
+// region async calls tracking
+const ASYNC_CALL_TTL = 3600;
+const ASYNC_CALL_KEY_PREFIX = 'async_status_';
```
xfournet: in Redis we usually separate the 'domain' from the id with `:` -> `async_status:`

opencti-platform/opencti-graphql/src/manager/ruleManager.ts
```diff
@@ -370,6 +373,60 @@ export const ruleApply = async (context: AuthContext, user: AuthUser, elementId:
...
+  if (!rule) {
+    throw FunctionalError('Cant find rule to scan', { id: ruleId });
+  }
+  const dataCanonicalize = jsonCanonicalize({ elementId, ruleId, executionId }) as string;
+  const ruleApplyId = uuidv5(dataCanonicalize, OPENCTI_NAMESPACE);
+  let lock;
+  let result = false;
+  try {
+    // Try to get the lock in redis
+    lock = await lockResources(ruleApplyId, { draftId: getDraftContext(context, user) });
```
xfournet: is draftId necessary here ?

opencti-platform/opencti-graphql/src/manager/ruleManager.ts
```diff
@@ -370,6 +373,60 @@ export const ruleApply = async (context: AuthContext, user: AuthUser, elementId:
...
+    const currentAsyncCall = await redisGetAsyncCall(ruleApplyId);
+    // If a call is already processing, return the call status
+    if (currentAsyncCall) {
+      result = currentAsyncCall !== '0';
```
xfournet: if call is in progress, we should also wait some time before returning the result

opencti-platform/opencti-graphql/src/manager/ruleManager.ts
```diff
@@ -370,6 +373,60 @@ export const ruleApply = async (context: AuthContext, user: AuthUser, elementId:
...
+        await redisFinishAsyncCall(ruleApplyId);
+        result = true;
+      } catch {
+        // If timeout promise is the first to finish, we do not await the rule apply but instead we return the ongoing work
+        // The work will be moved to complete when the rule apply is finished
+        ruleApplyPromise.catch((err) => {
+          logApp.error('[OPENCTI] Error during rule apply', { cause: err });
+        }).finally(async () => {
+          await redisFinishAsyncCall(ruleApplyId);
+        });
```
xfournet: mixing promise style and try/catch/finally is a bit complicated, you could stick to try/catch/finally everywhere

opencti-platform/opencti-graphql/src/manager/ruleManager.ts
```diff
@@ -370,6 +373,60 @@ export const ruleApply = async (context: AuthContext, user: AuthUser, elementId:
...
+        setTimeout(() => reject(new Error('Timeout')), 10000),
+      );
+      const ruleApplyPromise = ruleApply(context, user, elementId, ruleId);
+      try {
+        await Promise.race([ruleApplyPromise, timeoutPromise]);
+        // If rule apply promise is the first to finish, we move the work to complete here
+        await redisFinishAsyncCall(ruleApplyId);
+        result = true;
+      } catch {
+        // If timeout promise is the first to finish, we do not await the rule apply but instead we return the ongoing work
```
xfournet: ruleApplyPromise can fail too. instead of relying on throwing an error for timeout, we could resolve the timeout on a symbol that can be tested as the result of await Promise.race

client-python/pycti/api/opencti_api_work.py
```diff
@@ -243,7 +243,10 @@ def wait_for_work_to_finish(self, work_id: str):
         :return: empty string if error, None otherwise
         :rtype: str or None
         """
+        state = self.get_work(work_id=work_id)
         status = ""
+        if len(state) > 0:
+            status = state["status"]
```
xfournet: is it to avoid the uneeded sleep in case status is already completed ?

opencti-platform/opencti-graphql/src/manager/ruleManager.ts
JeremyCloarec: currently no, since rules are only applied outside of drafts, draftId will always be undefined. 
But I think we have a feature request to have rules running in draft, so setting the draftId here was done to already handle that use case if necessary

opencti-platform/opencti-graphql/src/manager/ruleManager.ts
JeremyCloarec: I don't fully get what you mean. Do you want the ruleApplyAsync function to always wait the same amount of time after the first call ? Can't we instead return directly and let the caller decide on the wait time between each call to ruleApplyAsync to check if it is done?

client-python/pycti/api/opencti_api_work.py
JeremyCloarec: exactly, we no longer need it since we don't use works anymore, but I felt like it was an improvement worth keeping

opencti-platform/opencti-front/src/schema/relay.schema.graphql
```diff
@@ -9951,6 +9951,7 @@ type Mutation {
   ruleSetActivation(id: ID!, enable: Boolean!): Rule!
   ruleManagerClean(eventId: ID): RuleManager!
   ruleApply(elementId: ID!, ruleId: ID!): Boolean
```
xfournet, line 9953: i wonder if we could merge ruleApply and ruleApplyAsync by specifing an optional asyncExecutionId: ID parameter ?

client-python/pycti/api/opencti_api_work.py
xfournet: ok, i would rather test the status != complete just before the wait to reduce code duplication

client-python/pycti/entities/opencti_stix_core_object.py
```diff
@@ -1873,6 +1874,48 @@ def rule_apply(self, **kwargs):
...
+                        "ruleId": rule_id,
+                        "executionId": execution_id,
+                    },
+                )
+                rule_apply_complete = result["data"]["ruleApplyAsync"]
+                self.opencti.app_logger.info(
+                    "Apply rule async stix_core_object complete", {"id": element_id}
+                )
+                if not rule_apply_complete:
+                    time.sleep(10)
```
xfournet: we should rather have an active wait in the backend

opencti-platform/opencti-graphql/src/database/redis.ts
```diff
@@ -506,7 +506,29 @@ export const redisInitializeWork = async (workId: string) => {
   });
 };
 // endregion
-
+// region async calls tracking
+const ASYNC_CALL_TTL = 3600;
+const ASYNC_CALL_KEY_PREFIX = 'async_status:';
```
xfournet: we can encapsulate the constant with this method
```suggestion
const asyncCallKey = (id: string) => `async_status:${id}`;
```

opencti-platform/opencti-graphql/src/manager/ruleManager.ts
```diff
@@ -370,6 +373,60 @@ export const ruleApply = async (context: AuthContext, user: AuthUser, elementId:
   return executeRuleApply(context, user, rule, elementId);
 };
 
+export const ruleApplyAsync = async (context: AuthContext, user: AuthUser, elementId: string, ruleId: string, executionId: string) => {
```
xfournet: In case the node executing the rule apply fail, it won't be able to resume before 1h (in worst case)

The challenges are:
* being able to detect a 'dead' ruleApply due to node failure
* being able to make active wait, even from a node that is not executing this ruleApply

I propose to rely on redis status + redis lock + local map of async task promise:
* keep the lock during all the ruleApply execution
* keep the lock alive with a // async task

Draft algo : when receiving a request
1. check the local promise map -> if present go to step 3.
2. try to get the lock with timeout
2.1.  if lock cannot be acquired during the timeout, task is still in progress, then return false
2.2.  if lock can be acquired, check status
2.2.1. status = done, task is terminated, return true
2.2.2. no status or status = in progress (that mean dead task) then create the async call promise
3. wait the call promise with race timeout, if timeout win the race then return false, else true

JeremyCloarec: I pushed a version of the draft algo. I'm not sure we need the local map of async tasks though, can't we just always try to get the lock?

xfournet (approved):

Issue #14041, opened 2026-01-15
Prevent JSON Mapper from crashing the app when generating a high amount of relationships


OctaveLaventure:
The back-end can run out of RAM when using a JSON mapper configuration and data generating thousands of relationships with nested representations

PR #14042, opened 2026-01-15
[backend] Optimize JSON mapper memory usage and performance

OctaveLaventure:
### Proposed changes

- Refactor `jsonMappingExecution` for better readability
- Implement early STIX conversion to reduce memory pressure during large dataset processing
- Optimize relationship resolution with O(1) map lookups instead of array filtering

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/14041#issue-3818455091

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/modules/ingestion/ingestion-json-domain.ts
+++ b/opencti-platform/opencti-graphql/src/modules/ingestion/ingestion-json-domain.ts
@@ -14,8 +14,6 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 */
 
 import * as JSONPath from 'jsonpath-plus';
-import { v4 as uuidv4 } from 'uuid';
-import type { StixBundle } from '../../types/stix-2-1-common';
 import type { AuthContext, AuthUser } from '../../types/user';
 import { fullEntitiesList, pageEntitiesConnection, storeLoadById } from '../../database/middleware-loader';
 import { type BasicStoreEntityIngestionJson, type DataParam, ENTITY_TYPE_INGESTION_JSON } from './ingestion-types';
@@ -145,18 +143,12 @@ export const executeJsonQuery = async (context: AuthContext, ingestion: BasicSto
   };
   const platformUsers = await getEntitiesMapFromCache<AuthUser>(context, SYSTEM_USER, ENTITY_TYPE_USER);
   const ingestionUser = ingestion.user_id ? platformUsers.get(ingestion.user_id) : null;
-  const objects = await jsonMappingExecution(context, ingestionUser || SYSTEM_USER, requestData, jsonMapperParsed);
-  const bundle: StixBundle = {
-    id: `bundle--${uuidv4()}`,
-    spec_version: '2.1',
-    type: 'bundle',
-    objects,
-  };
+  let objects = await jsonMappingExecution(context, ingestionUser || SYSTEM_USER, requestData, jsonMapperParsed);
   let nextExecutionState = buildQueryObject(ingestion.query_attributes, { ...requestData, ...responseHeaders }, false);
   // region Try to paginate with next page style
   if (ingestion.pagination_with_sub_page && isNotEmptyField(ingestion.pagination_with_sub_page_attribute_path)) {
     let url = getValueFromPath(ingestion.pagination_with_sub_page_attribute_path, requestData);
-    while (isNotEmptyField(url) && (maxResults === 0 || (bundle.objects ?? []).length < maxResults)) {
+    while (isNotEmptyField(url) && (maxResults === 0 || objects.length < maxResults)) {
       logApp.info(`> Sub query: ${url}`);
       await wait(100); // Wait 100 ms between 2 calls
       const { data: paginationData } = await httpClient.call({
@@ -168,7 +160,7 @@ export const executeJsonQuery = async (context: AuthContext, ingestion: BasicSto
       nextExecutionState = { ...nextExecutionState, ...paginationVariables };
       const paginationObjects = await jsonMappingExecution(context, ingestionUser || SYSTEM_USER, paginationData, jsonMapperParsed);
       if (paginationObjects.length > 0) {
-        bundle.objects = bundle.objects.concat(paginationObjects);
+        objects = objects.concat(paginationObjects);
       }
       url = getValueFromPath(ingestion.pagination_with_sub_page_attribute_path, paginationData);
     }
@@ -176,9 +168,9 @@ export const executeJsonQuery = async (context: AuthContext, ingestion: BasicSto
   // endregion
   // In case of limitation, ensure to not return too many elements
   if (maxResults > 0) {
-    bundle.objects = bundle.objects.slice(0, maxResults);
+    objects = objects.slice(0, maxResults);
   }
-  return { bundle, variables, nextExecutionState };
+  return { objects, variables, nextExecutionState };
 };
 
 export const findById = async (context: AuthContext, user: AuthUser, ingestionId: string, removeCredentials = false) => {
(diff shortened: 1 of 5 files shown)


opencti-platform/opencti-graphql/src/parser/json-mapper.ts
```diff
@@ -341,10 +333,50 @@ const computeOrderedRepresentations = (representations: JsonMapperRepresentation
...
+const processAndStoreInput = (
+  representation: JsonMapperRepresentation,
+  input: any,
+  results: Map<string, Map<string, Record<string, InputType>>>,
+  seenIds: Set<string>,
+  finalObjects: StixObject[],
+) => {
+  let stixObject: StixObject | null = null;
+  try {
+    stixObject = convertStoreToStix_2_1(input as unknown as StoreCommon);
```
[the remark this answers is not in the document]
OctaveLaventure: out of scope, it was already there

opencti-platform/opencti-graphql/src/parser/json-mapper.ts
```diff
@@ -341,10 +333,50 @@ const computeOrderedRepresentations = (representations: JsonMapperRepresentation
...
+    delete input.__tos;
+  } else {
+    if (results.has(representation.id)) {
+      const current = results.get(representation.id);
+      if (!current?.has(input.__identifier)) {
+        current?.set(input.__identifier, input);
+      }
+    } else {
+      results.set(representation.id, new Map([[input.__identifier, input]]));
+    }
```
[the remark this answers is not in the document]
xfournet:
```suggestion
    const current = results.get(representation.id);
    if (current) {
      if (!current.has(input.__identifier)) {
        current.set(input.__identifier, input);
      }
    } else {
      results.set(representation.id, new Map([[input.__identifier, input]]));
    }
```

opencti-platform/opencti-graphql/src/parser/json-mapper.ts
```diff
@@ -267,16 +266,21 @@ const handleBasedOnAttribute = async (
...
+    let entities: Record<string, InputType>[];
     if (attribute.based_on.identifier) {
       const mappedIdentifiers = extractTargetIdentifierFromJson(base, record, attribute.based_on.identifier, definition);
       entities = attribute.based_on.representations
-        .map((id) => otherEntities.get(id)).flat()
-        .filter((e) => e !== undefined && mappedIdentifiers.includes(e.__identifier as string)) as Record<string, InputType>[];
+        .flatMap((id) => {
+          const representationsMap = otherEntities.get(id);
+          if (!representationsMap) return [];
+          return mappedIdentifiers.map((ident) => representationsMap.get(ident)).filter((e) => e !== undefined);
```
xfournet: can avoid this map+filter too ?

```suggestion
          return mappedIdentifiers.flatMap((ident) => {
            const e = reps.get(ident);
            return e ? [e] : [];
          });
```

opencti-platform/opencti-graphql/src/parser/json-mapper.ts
```diff
@@ -433,26 +465,22 @@ const jsonMappingExecution = async (context: AuthContext, user: AuthUser, data:
...
-  const bundleBuilder = new BundleBuilder();
-  bundleBuilder.addObjects(stixObjects);
-  return bundleBuilder.build();
+
+  const bundle: StixBundle = {
+    id: `bundle--${uuidv4()}`,
+    spec_version: STIX_SPEC_VERSION,
+    type: 'bundle',
+    objects: finalObjects,
+  };
```
xfournet: we could even return only finalObjects since every callers doesn't use other informations

opencti-platform/opencti-graphql/src/parser/json-mapper.ts
```diff
@@ -341,10 +333,50 @@ const computeOrderedRepresentations = (representations: JsonMapperRepresentation
...
+  results: Map<string, Map<string, Record<string, InputType>>>,
+  seenIds: Set<string>,
+  finalObjects: StixObject[],
+) => {
+  let stixObject: StixObject | null = null;
+  try {
+    stixObject = convertStoreToStix_2_1(input as unknown as StoreCommon);
+  } catch (e) {
+    logApp.error('JSON mapper convert error', { cause: e });
+  }
```
xfournet: is it expected to continue the method execution if the conversion failed ?

opencti-platform/opencti-graphql/src/parser/json-mapper.ts
```diff
@@ -341,10 +333,50 @@ const computeOrderedRepresentations = (representations: JsonMapperRepresentation
...
 };
 
+const processAndStoreInput = (
+  representation: JsonMapperRepresentation,
+  input: any,
+  results: Map<string, Map<string, Record<string, InputType>>>,
+  seenIds: Set<string>,
+  finalObjects: StixObject[],
+) => {
+  let stixObject: StixObject | null = null;
```
xfournet: I prefer to use undefined rather than null when there is no specific semantic (we can have a chat about this opinion if you are interested)

```suggestion
  let stixObject: StixObject | undefined = undefined ;
```

opencti-platform/opencti-graphql/src/parser/json-mapper.ts
```diff
@@ -341,10 +333,50 @@ const computeOrderedRepresentations = (representations: JsonMapperRepresentation
   return [baseEntities, basedOnEntities, relationships];
 };
 
+const processAndStoreInput = (
```
xfournet: you could define it inside the jsonMappingExecution after the declaration of results, finalObjects & seenIds so you can remove them from the parameters list

opencti-platform/opencti-graphql/src/parser/json-mapper.ts
```diff
@@ -341,10 +333,50 @@ const computeOrderedRepresentations = (representations: JsonMapperRepresentation
...
+  if (stixObject && !seenIds.has(stixObject.id)) {
+    seenIds.add(stixObject.id);
+    finalObjects.push(stixObject);
+  }
+
+  if (representation.type === JsonMapperRepresentationType.Relationship) {
+    delete input.from;
+    delete input.to;
+    delete input.__froms;
+    delete input.__tos;
```
xfournet: i struggle to understand why relationship is handled differently than in the previous implementation.
* not storing in results anymore: i suppose it was not necessary in the previous implementation and has been removed to save memory ?
* removing from/to/__froms/_tos : could you add a comment to clarify it

OctaveLaventure: not storing: actually we should, the relationships can be referenced later on in some cases such as "sighting-of"
removing from/to: we don't need them later on, just the relationship's ID so we can free that memory

opencti-platform/opencti-graphql/src/modules/ingestion/ingestion-json-domain.ts
```diff
@@ -143,7 +145,13 @@ export const executeJsonQuery = async (context: AuthContext, ingestion: BasicSto
...
   const platformUsers = await getEntitiesMapFromCache<AuthUser>(context, SYSTEM_USER, ENTITY_TYPE_USER);
   const ingestionUser = ingestion.user_id ? platformUsers.get(ingestion.user_id) : null;
-  const bundle = await jsonMappingExecution(context, ingestionUser || SYSTEM_USER, requestData, jsonMapperParsed);
+  const objects = await jsonMappingExecution(context, ingestionUser || SYSTEM_USER, requestData, jsonMapperParsed);
+  const bundle: StixBundle = {
+    id: `bundle--${uuidv4()}`,
+    spec_version: '2.1',
+    type: 'bundle',
+    objects,
+  };
```
xfournet: it's created a bit too early, following the executeJsonQuery calls, bundle need to be created just before pushBundleToConnectorQueue https://github.com/OpenCTI-Platform/opencti/blob/master/opencti-platform/opencti-graphql/src/manager/ingestionManager.ts#L648

xfournet (approved):
Tested:
* ✅ out of memory scenario is working now (slowly but this scenario is not functionally valid)
* ✅ functionally valid mapping continue to work as before the PR

Code coverage missed lines review:
* ✅ on existing untested code impacted by minor refactor
* ✅ edge case on convertStoreToStix_2_1 error handling

PR #14104, opened 2026-01-19
[backend] stabilize email mime part type and windows registry value type standard_id (#13501)

### LINKED ISSUE #13501: Standard STIX IDs are being Modified
## Description
Standard STIX identifiers are modified with each update of observables, especially for "Email message - MIME Part" and "Windows Registry Key Value".

## Environment
6.8.13

## Expected Output

Standard STIX IDs do not get modified

## Actual Output

Standard STIX IDs are modified

JeremyCloarec:
### Proposed changes

* stabilize email mime part type and windows registry value type standard_id by specifying attributes to use for standard_id computation
*

### Related issues

* #13501
*

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/schema/identifier.js
+++ b/opencti-platform/opencti-graphql/src/schema/identifier.js
@@ -138,8 +138,8 @@ const stixBaseCyberObservableContribution = {
     [C.ENTITY_ICCID]: [{ src: 'value' }],
     [C.ENTITY_IMSI]: [{ src: 'value' }],
     // Types embedded
-    [C.ENTITY_EMAIL_MIME_PART_TYPE]: [], // ALL
-    [C.ENTITY_WINDOWS_REGISTRY_VALUE_TYPE]: [], // ALL
+    [C.ENTITY_EMAIL_MIME_PART_TYPE]: [{ src: 'body' }, { src: 'content_type' }, { src: 'data_content_dispositiontype' }],
+    [C.ENTITY_WINDOWS_REGISTRY_VALUE_TYPE]: [{ src: 'name' }, { src: 'data' }, { src: 'data_type' }],
   },
   resolvers: {
     from(from) {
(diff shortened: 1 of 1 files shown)


opencti-platform/opencti-graphql/src/schema/identifier.js
```diff
@@ -132,8 +132,8 @@ const stixBaseCyberObservableContribution = {
     [C.ENTITY_SSH_KEY]: [{ src: 'fingerprint_sha256' }],
     [C.ENTITY_PERSONA]: [{ src: 'persona_name' }, { src: 'persona_type' }],
     // Types embedded
-    [C.ENTITY_EMAIL_MIME_PART_TYPE]: [], // ALL
-    [C.ENTITY_WINDOWS_REGISTRY_VALUE_TYPE]: [], // ALL
+    [C.ENTITY_EMAIL_MIME_PART_TYPE]: [{ src: 'body' }, { src: 'content_type' }, { src: 'data_content_dispositiontype' }],
```
SouadHadjiat: I guess this has been validated ?
I can't manage to test it correctly, the edition form always displays empty values, even if I change them. It's as if the body and content type are never fetch from graphql

JeremyCloarec: I didn't test it with email mime part type, but you're right that our frontend looks to be completely broken for this kind of observable. It's not retrieving any values

JeremyCloarec: I gues we can open an other bug to tackle the issue

opencti-platform/opencti-graphql/src/schema/identifier.js
```diff
@@ -134,8 +132,8 @@ const stixBaseCyberObservableContribution = {
...
```
ludovic: I think you want to use "content_disposition" and not "data_content_dispositiontype" ?

ludovic (commented):
Shouldn't you plan a migration to recalculate the existing ids?

JeremyCloarec:
> Shouldn't you plan a migration to recalculate the existing ids?

I didn't find it that necessary, since the correct IDs will be recomputed on the next ingestion anyway. Going for a migration might be overkill on this one

ludovic (approved):

Issue #13303, opened 2025-11-24
[EPIC] UI Improvements


romain-filigran:
## Use case

This parent issue tracks all UI improvements planned for the next iteration, including the introduction of design system components.

PR #14110, opened 2026-01-19
[frontend] Files Search in Data Table (#13303)

Archidoit:
### Proposed changes
In Files indexing Search results:
- 'Files search' should begin with an upper case letter
- fix alignement in columns
- move result in Data Table

### Related issues
#13303


opencti-platform/opencti-front/src/private/components/search/SearchContainerQuery.tsx
```diff
@@ -60,7 +60,7 @@ const SearchContainer: FunctionComponent<SearchRootComponentProps> = ({ children
             value="files"
             label={(
               <Badge badgeContent={filesCount} color="primary">
-                <div style={{ padding: '0px 12px', display: 'flex' }}>{t_i18n('Files search')}<EEChip /></div>
+                <div style={{ padding: '0px 12px', display: 'flex', textTransform: 'none' }}>{t_i18n('Files search')}<EEChip /></div>
```
esrevi: could we breakline the props ? :)

opencti-platform/opencti-front/src/private/components/search/SearchIndexedFiles.tsx
```diff
@@ -142,36 +219,51 @@ const SearchIndexedFilesComponent = () => {
...
+                  </Tooltip>
+                );
+              }
+            }}
+          />
+        )}
       </>
     );
   };
+
```
esrevi, line 266: i can't comment the lines below directly, but on line 274, we can remove position: relative imo

and on line 285, we can change 
style={{ marginLeft: 20 }} with sx={{ marginLeft: 1 }}

Because we re using 8px system space in design system

opencti-platform/opencti-front/src/private/components/search/SearchContainerQuery.tsx
Archidoit: yes ! done

opencti-platform/opencti-front/src/private/components/search/SearchIndexedFiles.tsx
Archidoit: Thanks for the precisions :) done

esrevi (approved):
cool refactoring with datatable :)

Issue #13460, opened 2025-12-03
Standardize button styles and shapes


sandyghs:
Unify button styles, sizes and shapes to ensure visual consistency and strengthen interaction patterns.

PR #14122, opened 2026-01-20
[frontend] Some improvements on new Button design (#13460)

lndrtrbn:
### Proposed changes

* Style of disabled button fixed
* Some improvement on External Reference page
* Init design system palette in theme

### Related issues

* #13460

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/components/common/button/Button.utils.ts
+++ b/opencti-platform/opencti-front/src/components/common/button/Button.utils.ts
@@ -15,7 +15,7 @@ export const getColorDefinitions = (theme: Theme): Record<ButtonColorKey, ColorD
       main: theme.palette.error.main ?? '',
       hover: theme.palette.error.dark ?? '',
       focus: theme.palette.text.primary ?? '',
-      text: theme.palette.error.main ?? '',
+      text: theme.palette.text.primary ?? '',
       border: theme.palette.error.main ?? '',
     },
     ai: {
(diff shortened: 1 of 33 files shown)


opencti-platform/opencti-front/src/components/common/button/Button.utils.ts
```diff
@@ -1,21 +1,22 @@
...
 
 export const getColorDefinitions = (theme: Theme): Record<ButtonColorKey, ColorDefinition> => {
   return {
     default: {
-      main: theme.palette.primary.main,
-      hover: theme.palette.primary.dark,
-      focus: theme.palette.text.primary,
-      text: theme.palette.primary.contrastText,
-      border: theme.palette.primary.main,
+      main: theme.palette.primary.main ?? '',
```
esrevi, line 8: i think this case won't happen but if we have color like color={''} , what happen?

lndrtrbn: I don't know, indeed it won't happen it's for typescript compliance until we have fully moved to design system colors

opencti-platform/opencti-front/src/private/components/data/connectors/ConnectorWorkLine.tsx
```diff
@@ -152,7 +152,7 @@ const ConnectorWorkLine: FunctionComponent<
         <Button
           sx={{ position: 'absolute', right: 16, top: 16 }}
           variant="secondary"
-          color={(workErrors ?? []).length === 0 ? 'success' : 'warning'}
+          color={(workErrors ?? []).length === 0 ? 'success' : 'error'}
```
esrevi: should this be ?
color={(workErrors ?? []).length === 0 ? 'success' : undefined}
intent={(workErrors ?? []).length >= 0 ? 'destructive': undefined}

esrevi (approved):

Issue #13452, opened 2025-12-03
Left Navigation Redesign


sandyghs:
Update the left navigation to refine hierarchy, improve readability and ensure consistency.

PR #14123, opened 2026-01-20
[frontend] Preserve case for markings in tooltips and autocomplete chips (#13452)

Gwendoline-FAVRE-FELIX:
### Proposed changes

* Prevents text-transform from altering marking values in Tag tooltips.
* Renders ObjectMarking autocomplete chips with Tag to preserve original casing.

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/13452

### DIFF
--- a/opencti-platform/opencti-front/src/components/AutocompleteField.tsx
+++ b/opencti-platform/opencti-front/src/components/AutocompleteField.tsx
@@ -143,7 +143,7 @@ const AutocompleteField = <
                 {...getTagProps({ index })}
                 applyLabelTextTransform={applyLabelTextTransform}
                 key={value}
-                label={truncate(label, 50)}
+                label={label}
               />
             );
           })
--- a/opencti-platform/opencti-front/src/components/AutocompleteField.tsx
+++ b/opencti-platform/opencti-front/src/components/AutocompleteField.tsx
@@ -1,13 +1,15 @@
+import React, { ReactNode, useCallback } from 'react';
 import IconButton from '@common/button/IconButton';
 import { Add } from '@mui/icons-material';
 import { Autocomplete, AutocompleteProps, AutocompleteValue, TextField, TextFieldProps } from '@mui/material';
 import { FieldProps, useField } from 'formik';
-import { fieldToAutocomplete } from 'formik-mui';
-import { ReactNode, useCallback } from 'react';
 import { FieldOption } from '../utils/field';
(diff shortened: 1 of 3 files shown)


opencti-platform/opencti-front/src/components/AutocompleteField.tsx
```diff
@@ -125,6 +128,16 @@ const AutocompleteField = <
         noOptionsText={noOptionsText}
         {...fieldProps}
         renderOption={renderOption}
+        renderTags={(values, getTagProps) => (
+          values.map((option, index) => (
+            <Tag
+              {...getTagProps({ index })}
+              applyLabelTextTransform={applyLabelTextTransform}
+              key={(option as FieldOption).value}
```
esrevi: i think casting could be could generate issue if we pass an array of string instead of object to Autocomplete. 

I'd suggest to typeguard that with a function for example
```
  const getOptionData = (option: Value) => {
    return typeof option === 'object' && option !== null
      ? { label: option.label, value: option.value }
      : { label: String(option), value: String(option) };
  };
  ```
  then use in renderTags
   const { label, value } = getOptionData(option);

File: opencti-platform/opencti-front/src/components/common/tag/Tag.tsx [hunk 2]
```diff
@@ -6,7 +6,7 @@ interface TagProps extends Omit<ChipProps, 'color'> {
   label: string;
   color?: string;
   onClick?: (e: React.MouseEvent) => void;
-  onDelete?: () => void;
+  onDelete?: (event?: any) => void;
```
esrevi: prefer avoiding any type

replace with React.MouseEvent

opencti-platform/opencti-front/src/components/AutocompleteField.tsx
```diff
@@ -125,6 +134,19 @@ const AutocompleteField = <
...
         renderOption={renderOption}
+        renderTags={(values, getTagProps) => (
+          values.map((option, index) => {
+            const { label, value } = getOptionData(option);
+            return (
+              <Tag
+                {...getTagProps({ index })}
+                applyLabelTextTransform={applyLabelTextTransform}
+                key={value}
+                label={truncate(label, 50)}
```
esrevi: you can remove truncate i think

esrevi (approved):

PR #14125, opened 2026-01-20
[backend] Add upsert fallback when using auto merging option and confidence don't match (#14124)

### LINKED ISSUE #14124: [backend] Add upsert fallback when using auto merging option and confidence dont match
## Use case

When connector is allowed to do auto merging but the user have not enough confidence level, instead of just failing the merge, we want to fallback to upsert as some information are useful to be kept and rely on upsert logic.

richard-julien:
### 🎯 Target of the Modification
The primary goal of this update is to **resolve a blocking issue where low-confidence users could not update or enrich existing high-confidence entities** when their action triggered an automatic "merge" operation.

Previously, if a user with low confidence (e.g., `50`) tried to create an entity that already existed (based on STIX ID) and was created by a high-confidence user (e.g., `100`), the system would attempt to **merge** the two. This merge operation would fail with an `INSUFFICIENT_CONFIDENCE_LEVEL` error, completely blocking the user from adding any data.

The fix ensures that instead of failing, the system **falls back to an "Upsert"** (Update/Insert) strategy. This allows the user to contribute data (like new aliases or references) without overwriting high-confidence fields they shouldn't touch.

#### 1. Backend Logic: Auto-Merge Fallback (`middleware.js`)
*   **Problem:** `createEntityRaw` attempts to merge duplicate entities (using specific option update=true). If the user lacks sufficient confidence to modify the existing entity during a merge, the operation throws an error.
*   **Fix:** A `try-catch` block was added around the creation logic.
    *   It listens specifically for the `INSUFFICIENT_CONFIDENCE_LEVEL` error code.
    *   If caught, it logs a warning and **re-executes the operation with `update: false`**.
    *   This forces the system to skip the "merge" attempt and instead treat it as a standard "upsert", allowing non-conflicting additions to proceed.

#### 2. Upsert Logic: Relaxed Confidence for Metadata (`upsert-utils.js`)
*   **Problem:** Strictly enforcing confidence levels prevented users from adding simple non-destructive metadata like **Labels** or **External References** to high-confidence objects.
*   **Fix:** In `generateRefsInputsForUpsert`, the confidence check was bypassed for specific relationship types:
    *   `object-label` (Labels)
    *   `external-reference` (External References)
*   **Result:** A lower-confidence user can now add a label to a high-confidence object, whereas previously this would have been rejected.

#### 3. Testing Infrastructure ([testQuery.ts](cci:7://file:///Users/julienrichard/Documents/filigran/opencti/core-engine/opencti-platform/opencti-graphql/tests/utils/testQuery.ts:0:0-0:0) & [middleware-test.js](cci:7://file:///Users/julienrichard/Documents/filigran/opencti/core-engine/opencti-platform/opencti-graphql/tests/01-unit/database/middleware-test.js:0:0-0:0))
*   **Validation:** A new test case `should multiple threat actors that need merge still upserted with using lower confidence` was added.
    *   It creates a high-confidence Threat Actor.
    *   It attempts to "update" it using a low-confidence user.
    *   It verifies that the operation succeeds (no error) and that the allowed fields are updated while high-confidence fields remain untouched.
*   **Utils:** Updated [buildStandardUser](cci:1://file:///Users/julienrichard/Documents/filigran/opencti/core-engine/opencti-platform/opencti-graphql/tests/utils/testQuery.ts:692:0-728:2) to accept a `maxConfidence` parameter, facilitating easier testing of these confidence-mismatch scenarios.

### 📝 Summary of Impact
| Feature | Before | After |
| :--- | :--- | :--- |
| **Low Confidence Merge** | Failed with `Access Denied` / `Insufficient Confidence`. | **Success**: Falls back to specialized upsert behavior. |
| **Adding Labels** | Blocked if user confidence < entity confidence. | **Allowed**: Labels are treated as additive metadata regardless of confidence. |
| **Adding External Refs** | Blocked if user confidence < entity confidence. | **Allowed**: External refs are treated simply as additive. |

### DIFF
--- a/opencti-platform/opencti-graphql/src/database/middleware.js
+++ b/opencti-platform/opencti-graphql/src/database/middleware.js
@@ -3430,7 +3430,7 @@ const createEntityRaw = async (context, user, rawInput, type, opts = {}) => {
   } catch (e) {
     // In case of insufficient confidence level, don't reject and continue to upsert
     // as upsert have a complex strategy about confidence that doesn't reject everything
-    if (e?.extensions?.data?.doc_code === INSUFFICIENT_CONFIDENCE_LEVEL) {
+    if (rawInput.update !== false && e?.extensions?.data?.doc_code === INSUFFICIENT_CONFIDENCE_LEVEL) {
       logApp.warn('Merging stopped because of user confidence level, applying upsert', { cause: e });
       // Try to execute the method forcing update to false, prevent auto merging.
       return await internalCreateEntityRaw(context, user, { ...rawInput, update: false }, type, opts);
--- a/opencti-platform/opencti-graphql/src/database/middleware.js
+++ b/opencti-platform/opencti-graphql/src/database/middleware.js
@@ -3181,15 +3181,14 @@ export const getExistingEntities = async (context, user, input, type) => {
 const cleanEntityForIdsCollision = (input, type, target, foundEntities) => {
   // We can upsert element except the aliases that are part of other entities
   const key = resolveAliasesField(type).name;
-  const concurrentAliases = R.flatten(R.map((c) => [c[key], c.name], foundEntities));
-  const normedAliases = R.uniq(concurrentAliases.map((c) => normalizeName(c)));
-  const filteredAliases = R.filter((i) => !normedAliases.includes(normalizeName(i)), input[key] || []);
+  const concurrentAliases = foundEntities.flatMap((c) => [c[key], c.name]);
+  const normedAliases = new Set(concurrentAliases.map((c) => normalizeName(c)));
+  const filteredAliases = (input[key] ?? []).filter((i) => !normedAliases.has(normalizeName(i)));
   // We need also to filter eventual STIX IDs present in other entities
-  const concurrentStixIds = R.flatten(R.map((c) => [c.x_opencti_stix_ids, c.standard_id], foundEntities));
-  const normedStixIds = R.uniq(concurrentStixIds);
-  const filteredStixIds = R.filter(
-    (i) => isNotEmptyField(i) && !normedStixIds.includes(i) && i !== target.standard_id,
-    [...(input.x_opencti_stix_ids ?? []), input.stix_id],
+  const concurrentStixIds = foundEntities.flatMap((c) => [c.x_opencti_stix_ids, c.standard_id]);
+  const normedStixIds = new Set(concurrentStixIds);
+  const filteredStixIds = [...(input.x_opencti_stix_ids ?? []), input.stix_id].filter(
+    (i) => isNotEmptyField(i) && !normedStixIds.has(i) && i !== target.standard_id,
   );
   return { ...input, [key]: filteredAliases, x_opencti_stix_ids: filteredStixIds };
 };
--- a/opencti-platform/opencti-graphql/src/database/middleware.js
+++ b/opencti-platform/opencti-graphql/src/database/middleware.js
@@ -2953,7 +2953,7 @@ export const createRelationRaw = async (context, user, rawInput, opts = {}) => {
 
   // when creating stix ref, we must check confidence on from side (this count has modifying this element itself)
   // Through UI we use direct relationship creation ref to add label, external ref ...
-  // That's also true for some connectors that directly used the API. It's a bad practice but still a reality
+  // This is also true for some connectors that directly use the API. It's a bad practice but still a reality
   if (isStixRefRelationship(relationshipType) && shouldCheckConfidenceOnRefRelationship(relationshipType)) {
     controlUserConfidenceAgainstElement(user, from);
   }
@@ -3344,7 +3344,7 @@ const internalCreateEntityRaw = async (context, user, rawInput, type, opts = {})
         await mergeEntities(context, user, target.internal_id, sources.map((s) => s.internal_id), { locks: participantIds });
         return upsertElement(context, user, target, type, resolvedInput, { ...opts, locks: participantIds });
       }
-      // We can't merge, so we need a least to upsert one element.
+      // We can't merge, so we need at least to upsert one element.
       // First, looking for an instance that directly has the provided stix id
       // If we found an entity by the official stix_id, no need to cumulate the id, only upsert information.
       const targetByStixId = R.head(filteredEntities.filter((n) => getInstanceIds(n).includes(resolvedInput.stix_id)));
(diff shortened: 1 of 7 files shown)


opencti-platform/opencti-graphql/src/database/middleware.js
```diff
@@ -3177,7 +3178,23 @@ export const getExistingEntities = async (context, user, input, type) => {
...
+  const filteredAliases = R.filter((i) => !normedAliases.includes(normalizeName(i)), input[key] || []);
+  // We need also to filter eventual STIX IDs present in other entities
+  const concurrentStixIds = R.flatten(R.map((c) => [c.x_opencti_stix_ids, c.standard_id], foundEntities));
+  const normedStixIds = R.uniq(concurrentStixIds);
+  const filteredStixIds = R.filter(
+    (i) => isNotEmptyField(i) && !normedStixIds.includes(i) && i !== target.standard_id,
+    [...(input.x_opencti_stix_ids ?? []), input.stix_id],
+  );
+  return { ...input, [key]: filteredAliases, x_opencti_stix_ids: filteredStixIds };
+};
```
xfournet, line 3194: get rid off Ramda + replace array.includes() by Set.has() (perf)
```suggestion
const cleanEntityForIdsCollision = (input, type, target, foundEntities) => {
  // We can upsert element except the aliases that are part of other entities
  const key = resolveAliasesField(type).name;
  const concurrentAliases = foundEntities.flatMap((c) => [c[key], c.name]);
  const normedAliases = new Set(concurrentAliases.map((c) => normalizeName(c)));
  const filteredAliases = (input[key] ?? []).filter((i) => !normedAliases.has(normalizeName(i)));
  // We need also to filter eventual STIX IDs present in other entities
  const concurrentStixIds = foundEntities.flatMap((c) => [c.x_opencti_stix_ids, c.standard_id]);
  const normedStixIds = new Set(concurrentStixIds);
  const filteredStixIds = [...(input.x_opencti_stix_ids ?? []), input.stix_id].filter(
    (i) => isNotEmptyField(i) && !normedStixIds.has(i) && i !== target.standard_id,
  );
  return { ...input, [key]: filteredAliases, x_opencti_stix_ids: filteredStixIds };
};
```

opencti-platform/opencti-graphql/src/database/middleware.js
```diff
@@ -3314,21 +3331,9 @@ const createEntityRaw = async (context, user, rawInput, type, opts = {}) => {
...
-        // We need also to filter eventual STIX IDs present in other entities
-        const concurrentStixIds = R.flatten(R.map((c) => [c.x_opencti_stix_ids, c.standard_id], concurrentEntities));
-        const normedStixIds = R.uniq(concurrentStixIds);
-        const filteredStixIds = R.filter(
-          (i) => isNotEmptyField(i) && !normedStixIds.includes(i) && i !== existingByStandard.standard_id,
-          [...(resolvedInput.x_opencti_stix_ids ?? []), resolvedInput.stix_id],
-        );
-        const finalEntity = { ...resolvedInput, [key]: filteredAliases, x_opencti_stix_ids: filteredStixIds };
-        return upsertElement(context, user, existingByStandard, type, finalEntity, { ...opts, locks: participantIds });
+        const otherEntities = R.filter((e) => e.standard_id !== standardId, filteredEntities);
```
xfournet:
```suggestion
        const otherEntities = filteredEntities.filter((e) => e.standard_id !== standardId);
```

opencti-platform/opencti-graphql/src/database/middleware.js
```diff
@@ -3418,6 +3423,24 @@ const createEntityRaw = async (context, user, rawInput, type, opts = {}) => {
...
+    if (e?.extensions?.data?.doc_code === INSUFFICIENT_CONFIDENCE_LEVEL) {
+      logApp.warn('Merging stopped because of user confidence level, applying upsert', { cause: e });
+      // Try to execute the method forcing update to false, prevent auto merging.
+      return await internalCreateEntityRaw(context, user, { ...rawInput, update: false }, type, opts);
+    }
+    // For other errors, propagate the error.
+    // noinspection ExceptionCaughtLocallyJS
+    throw e;
+  }
+};
```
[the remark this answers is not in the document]
xfournet:
```suggestion
const createEntityRaw = async (context, user, rawInput, type, opts = {}) => {
  try {
    // Await is mandatory here to correctly catch the promise exception.
    return await internalCreateEntityRaw(context, user, rawInput, type, opts);
  } catch (e) {
    // In case of insufficient confidence level, don't reject and continue to upsert
    // as upsert have a complex strategy about confidence that doesn't reject everything
    if (rawInput.update !== false && e?.extensions?.data?.doc_code === INSUFFICIENT_CONFIDENCE_LEVEL) {
      logApp.warn('Merging stopped because of user confidence level, applying upsert', { cause: e });
      // Try to execute the method forcing update to false, prevent auto merging.
      return await internalCreateEntityRaw(context, user, { ...rawInput, update: false }, type, opts);
    }
    // For other errors, propagate the error.
    // noinspection ExceptionCaughtLocallyJS
    throw e;
  }
};
```

opencti-platform/opencti-graphql/src/database/middleware.js
```diff
@@ -3339,14 +3344,14 @@ const createEntityRaw = async (context, user, rawInput, type, opts = {}) => {
...
-        const target = R.head(filteredEntities);
-        const resolvedStixIds = { ...target, x_opencti_stix_ids: [...target.x_opencti_stix_ids, resolvedInput.stix_id] };
-        return upsertElement(context, user, target, type, resolvedStixIds, { ...opts, locks: participantIds });
-      }
-      // Return the matching STIX IDs in others
-      return { element: R.head(filteredEntities.filter((n) => getInstanceIds(n).includes(resolvedInput.stix_id))), event: null, isCreation: false };
+      // We can't merge, so we need at least to upsert one element.
+      // First, looking for an instance that directly has the provided stix id
+      // If we found an entity by the official stix_id, no need to cumulate the id, only upsert information.
+      const targetByStixId = R.head(filteredEntities.filter((n) => getInstanceIds(n).includes(resolvedInput.stix_id)));
```
xfournet:
```suggestion
      const targetByStixId = filteredEntities.filter((n) => getInstanceIds(n).includes(resolvedInput.stix_id))[0];
```

opencti-platform/opencti-graphql/src/utils/upsert-utils.js
```diff
@@ -373,13 +373,17 @@ const generateRefsInputsForUpsert = (context, user, resolvedElement, _type, upda
...
-              || (isInputWithData && !isCurrentWithData)
-            ) {
-              // If data is provided, different from existing data, and of higher confidence
-              // OR if existing data is empty and data is provided (even if lower confidence, it's better than nothing),
-              // --> apply an add operation
-              inputs.push({ key: inputField, value: inputToCurrentDiff, operation: UPDATE_OPERATION_ADD });
+            } else {
+              const fillEmptyData = isInputWithData && !isCurrentWithData;
+              const hasDataDifferential = isCurrentWithData && isInputWithData && inputToCurrentDiff.length > 0;
+              const isAllowedAddRefWithoutConfidence = relDef.name === objectLabel.name || relDef.name === externalReferences.name;
```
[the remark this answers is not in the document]
richard-julien: name is the identifier

xfournet (approved):

Issue #13627, opened 2025-12-10
[ci] Ensure tests in Github Action also work for fork (safely)


efaure:
## Use case

New Github action CI tests do not work for fork.
Ensure tests can be launched without security risk

## If the feature request is approved, would you be willing to submit a PR?

Yes / No (Help can be provided if you need assistance submitting a PR)

PR #14173, opened 2026-01-23
[ci] Handle license for fork

efaure:
### Proposed changes

* Use temporary EE license from Github vars

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/13627

### DIFF
--- a/.github/workflows/ci-test-api.yml
+++ b/.github/workflows/ci-test-api.yml
@@ -71,7 +71,6 @@ jobs:
           -v ${{ github.workspace }}/test-results:/opt/opencti/test-results \
           --env-file ${{ github.workspace }}/scripts/ci/ci-common.env \
           -e APP__BASE_URL=http://api-tests:4010/ \
-          -e APP__ENTERPRISE_EDITION_LICENSE="${{ secrets.EE_LICENSE }}" \
           -e APP__SYNC_RAW_START_REMOTE_URI=http://opencti-raw-start:4100/graphql \
           -e APP__SYNC_LIVE_START_REMOTE_URI=http://opencti-live-start:4100/graphql \
           -e APP__SYNC_DIRECT_START_REMOTE_URI=http://opencti-direct-start:4100/graphql \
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-test-api.yml
```diff
@@ -71,14 +71,14 @@ jobs:
...
           -e APP__BASE_URL=http://api-tests:4010/ \
-          -e APP__ENTERPRISE_EDITION_LICENSE="${{ secrets.EE_LICENSE }}" \
           -e APP__SYNC_RAW_START_REMOTE_URI=http://opencti-raw-start:4100/graphql \
           -e APP__SYNC_LIVE_START_REMOTE_URI=http://opencti-live-start:4100/graphql \
           -e APP__SYNC_DIRECT_START_REMOTE_URI=http://opencti-direct-start:4100/graphql \
           -e APP__SYNC_RESTORE_START_REMOTE_URI=http://opencti-restore-start:4100/graphql \
           -e PYTHONUNBUFFERED=1  \
           filigran/platform:${{ github.sha }} \
           sh -c 'set -eux
+            wget -O  /tmp/myfile https://filigran.jfrog.io/artifactory/ci-utils/opencti-license/temporary_ee_license.crt  && export APP__ENTERPRISE_EDITION_LICENSE="$(cat /tmp/myfile)" || exit 1
```
xfournet:
```suggestion
export APP__ENTERPRISE_EDITION_LICENSE="$(curl https://filigran.jfrog.io/artifactory/ci-utils/opencti-license/temporary_ee_license.crt)"
```
idem in ci-test-frontend.yml to avoid to create a tmp file

xfournet: temporary_ee_license.crt -> opencti_ee_ci_licence.crt that better represent what it is ?

efaure: @xfournet  Your suggestion is what I made at first but : 
 - curl is not installed in opencti image
 - this won't handle curl (or wget) error. If for some reason the license is not downloaded it won't be easy to see. The command is a bit more complicated, but in case of failure, it will stop immediately. If you see a cleaner way to handle the error, I take it ! We can also check if the var is empty after setting.

xfournet: Yes, we can stick with wget
To handle the variable substitution error you need to separate the variable substitution from the variable export (the command export exit code seems to override the command substitution exit code if they are combined)

eg
```suggestion
APP__ENTERPRISE_EDITION_LICENSE="$(wget -qO- https://filigran.jfrog.io/artifactory/ci-utils/opencti-license/temporary_ee_licen
se.crt)"
export APP__ENTERPRISE_EDITION_LICENSE
```

efaure: Updated :ok_hand:  I tried too hard to have do it on only one line, your version is clearer.

xfournet (approved):

Issue #3923, opened 2023-08-03
[EPIC] Ability to Configure Single Sign On from the GUI rather than just the backend


CyberKaizen:
## Use case

Currently, I've observed multiple users, myself included that face time consuming and large difficulty performing SSO integration. Those facing issues range from DevOps deployment engineers to those fairly new to deploy web apps.

To list a few I've observed and experienced:
- Syntax errors
- Typos
- Case Sensitivity
- Incorrect use of parameters or environment variables
- Depending on the vendor providing SAML, OpenID, LDAP, etc. Configurations can be wildly different than what the official documentation tells users to do.

SSO setup is arguable one of the most important pieces for increasing tool adoption or uptake.

## Current Workaround

No workaround except to hope that end user has followed instructions correctly, even in that event, there maybe some underlying issues causing any potential workaround to not work.

## Proposed Solution

Providing a SSO settings page in the GUI it will help ensure that the SSO integration experience will be consistent for users using Azure, AWS, Okta, etc. This will help with reducing the time spent by OpenCTI devs or support triaging issues/tickets related to SSO.

In the SSO settings interface, the user will be able to select their vendor and SSO strategy.

Based on the response the GUI will return specific fields for user input of the standard information requested in 
https://docs.opencti.io/5.9.X/deployment/authentication/.

These fields will run validation checks on the user input to ensure that the correct values are entered based on the selected vendor, SSO strategy, and existing settings in the OpenCTI platform such as platform organization settings.

These validation checks are necessary to help reduce the issues that today plague SSO integration with OpenCTI such as: typos, case sensitivity, incorrect use of parameters, and syntax errors.

This feature will help ensure that SSO with RBAC works as expected as the devs of OpenCTI intend.

## Additional Information

\
An example from one of our tools:
![image](https://github.com/OpenCTI-Platform/opencti/assets/78124503/255bca34-21a2-4d5e-953a-980af9ad1e68)

## If the feature request is approved, would you be willing to submit a PR?

Yes, in the form of documentation or additional details if requested. If documentation is provided on what portions of the code need to be changed, and/or maybe even a template. I can try working on it from there.

sbocahu:
@Jipegien 👍

CyberKaizen:
@sbocahu 
This is still relevant, even in 2025. 

However, now, there are alot of features around Org Segregation and sharing.

Would it make sense that the SSO GUI settings apply to each Organization, instead of the entire platform?

PR #14175, opened 2026-01-23
[backend/frontend] SSO : local strategy dedicated behaviors (#3923)

marieflorescontact:
### Proposed changes

* Prevent delete or disabled local strategy if it is the only one strategy
* UI/UX: 
* * Remove label button for local strategy edition
* * Remove Group and Org tab for local strategy edition

### Related issues

* #3923

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/settings/sso_definitions/SSODefinitionForm.tsx
+++ b/opencti-platform/opencti-front/src/private/components/settings/sso_definitions/SSODefinitionForm.tsx
@@ -345,8 +345,8 @@ const SSODefinitionForm = ({
               onChange={(event, value) => handleChangeTab(value)}
             >
               <Tab label={t_i18n('SSO Configuration')} />
-              <Tab label={t_i18n('Groups configuration')} />
-              <Tab label={t_i18n('Organizations configuration')} />
+              {selectedStrategy !== 'LocalAuth' && <Tab label={t_i18n('Groups configuration')} />}
+              {selectedStrategy !== 'LocalAuth' && <Tab label={t_i18n('Organizations configuration')} />}
             </Tabs>
           </Box>
           {currentTab === 0 && (
@@ -381,15 +381,17 @@ const SSODefinitionForm = ({
                 label={t_i18n(`Enable ${selectedStrategy} authentication`)}
                 containerstyle={{ marginLeft: 2, marginTop: 20 }}
               />
-              <Field
-                component={TextField}
-                variant="standard"
-                name="label"
-                onSubmit={updateField}
-                label={t_i18n('Login Button Name')}
-                fullWidth
-                style={{ marginTop: 10 }}
-              />
+              {selectedStrategy !== 'LocalAuth' && (
+                <Field
+                  component={TextField}
+                  variant="standard"
+                  name="label"
+                  onSubmit={updateField}
+                  label={t_i18n('Login Button Name')}
+                  fullWidth
+                  style={{ marginTop: 10 }}
+                />
+              )}
               {selectedStrategy === 'SAML' && <SAMLConfig updateField={updateField} />}
               {selectedStrategy === 'OpenID' && <OpenIDConfig updateField={updateField} />}
               {selectedStrategy === 'LDAP' && <LDAPConfig updateField={updateField} />}
@@ -473,8 +475,8 @@ const SSODefinitionForm = ({
               </FieldArray>
             </>
           )}
-          {currentTab === 1 && <SSODefinitionGroupForm updateField={updateField} selectedStrategy={selectedStrategy} isEditionMode={!!onSubmitField} />}
-          {currentTab === 2 && <SSODefinitionOrganizationForm updateField={updateField} selectedStrategy={selectedStrategy} isEditionMode={!!onSubmitField} />}
+          {currentTab === 1 && selectedStrategy !== 'LocalAuth' && <SSODefinitionGroupForm updateField={updateField} selectedStrategy={selectedStrategy} isEditionMode={!!onSubmitField} />}
+          {currentTab === 2 && selectedStrategy !== 'LocalAuth' && <SSODefinitionOrganizationForm updateField={updateField} selectedStrategy={selectedStrategy} isEditionMode={!!onSubmitField} />}
           {!onSubmitField && (
             <div
               style={{
--- a/opencti-platform/opencti-front/src/private/components/settings/sso_definitions/SSODefinitionForm.tsx
+++ b/opencti-platform/opencti-front/src/private/components/settings/sso_definitions/SSODefinitionForm.tsx
@@ -345,8 +345,13 @@ const SSODefinitionForm = ({
               onChange={(event, value) => handleChangeTab(value)}
             >
               <Tab label={t_i18n('SSO Configuration')} />
-              {selectedStrategy !== 'LocalAuth' && <Tab label={t_i18n('Groups configuration')} />}
-              {selectedStrategy !== 'LocalAuth' && <Tab label={t_i18n('Organizations configuration')} />}
+
+              {selectedStrategy !== 'LocalAuth' && (
+                <>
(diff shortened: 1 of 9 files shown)


opencti-platform/opencti-graphql/src/modules/singleSignOn/providers-initialization.js
```diff
@@ -417,33 +412,6 @@ export const initializeEnvAuthenticationProviders = async (context, user) => {
...
-            });
-        });
-        passport.use(LOCAL_STRATEGY_IDENTIFIER, adminLocalStrategy);
-        PROVIDERS.push({
-          name: INTERNAL_SECURITY_PROVIDER,
-          type: AuthType.AUTH_FORM,
-          strategy,
-          provider: LOCAL_STRATEGY_IDENTIFIER,
-        });
-      }
```
aHenryJard, line 706: We need keep this part.

opencti-platform/opencti-front/src/private/components/settings/sso_definitions/SSODefinitionForm.tsx
```diff
@@ -316,8 +316,8 @@ const SSODefinitionForm = ({
               onChange={(event, value) => handleChangeTab(value)}
             >
               <Tab label={t_i18n('SSO Configuration')} />
-              <Tab label={t_i18n('Groups configuration')} />
-              <Tab label={t_i18n('Organizations configuration')} />
+              {selectedStrategy !== 'LocalAuth' && <Tab label={t_i18n('Groups configuration')} />}
+              {selectedStrategy !== 'LocalAuth' && <Tab label={t_i18n('Organizations configuration')} />}
             </Tabs>
```
marieflorescontact, line 355: needs to group condition

opencti-platform/opencti-graphql/config/default.json
```diff
@@ -409,10 +409,5 @@
     "password": "",
     "forced_sender_email": "",
     "email_max_cc_size": 500
-  },
-  "providers": {
-    "local": {
-      "strategy": "LocalStrategy"
-    }
```
SouadHadjiat, line 416: why remove this configuration ?

opencti-platform/opencti-graphql/src/modules/singleSignOn/providers-initialization.js
```diff
@@ -676,6 +679,33 @@ export const initializeEnvAuthenticationProviders = async (context, user) => {
           HEADERS_AUTHENTICATORS.push(headerProvider);
         }
       }
+
+      // In case of disable local strategy, setup protected fallback for the admin user
+      const hasLocal = PROVIDERS.find((p) => p.strategy === EnvStrategyType.STRATEGY_LOCAL);
+      if (!hasLocal && !willLocalBeInDatabase) {
```
SouadHadjiat: what is the difference between hasLocal & willLocalBeInDatabase?

opencti-platform/opencti-graphql/src/modules/singleSignOn/singleSignOn-domain.ts
```diff
@@ -69,6 +69,12 @@ export const findSingleSignOnPaginated = async (context: AuthContext, user: Auth
 export const internalAddSingleSignOn = async (context: AuthContext, user: AuthUser, input: SingleSignOnAddInput, skipRegister: boolean) => {
   const defaultOps = { created_at: now(), updated_at: now() };
   const singleSignOnInput = { ...input, ...defaultOps };
+  // if input est local et que local deja dans en db on ajoute pas
+  const strategies = await findAllSingleSignOn(context, user);
+  const hasLocalStrategy = strategies.some((s) => s.strategy === StrategyType.LocalStrategy);
+  if (hasLocalStrategy && input.strategy === StrategyType.LocalStrategy) {
+    throw FunctionalError('Local Strategy already exists in database');
+  }
```
marieflorescontact, line 90: needs to check only if StrategyType.LocalStrategy
instead of findAll, filter on Strategy

opencti-platform/opencti-graphql/tests/03-integration/10 - Modules/singleSignOn/singleSignOn-resolver-test.ts
```diff
@@ -286,3 +288,43 @@ describe('Single Sign On', () => {
...
+      ],
+    };
+    const createdLocalStrategy = await queryAsAdminWithSuccess({
+      query: SINGLE_SIGN_ON_CREATE,
+      variables: { input: createInput },
+    });
+    expect(createdLocalStrategy).toBeDefined();
+    expect(createdLocalStrategy?.data?.singleSignOnAdd.name).toBe('local');
+    createdLocalStrategyId = createdLocalStrategy?.data?.singleSignOnAdd.id;
+  });
```
marieflorescontact, line 309: add test : try to create 2nd local, should fail

opencti-platform/opencti-graphql/config/default.json
marieflorescontact: because we don't wan't anymore providers configured with env

opencti-platform/opencti-graphql/src/modules/singleSignOn/providers-initialization.js
marieflorescontact: @aHenryJard

opencti-platform/opencti-graphql/config/default.json
marieflorescontact: but as we do want to be able to use it also with env, i presume I need to revert this change

opencti-platform/opencti-graphql/src/modules/singleSignOn/providers-initialization.js
```diff
@@ -676,6 +679,33 @@ export const initializeEnvAuthenticationProviders = async (context, user) => {
...
+            });
+        });
+        passport.use(LOCAL_STRATEGY_IDENTIFIER, adminLocalStrategy);
+        PROVIDERS.push({
+          name: INTERNAL_SECURITY_PROVIDER,
+          type: AuthType.AUTH_FORM,
+          strategy,
+          provider: LOCAL_STRATEGY_IDENTIFIER,
+        });
+      }
```
SouadHadjiat: this block is duplicated in the next if(isForcedEnv), I think we need to keep only one of these blocks

opencti-platform/opencti-graphql/tests/03-integration/10 - Modules/singleSignOn/singleSignOn-resolver-test.ts
marieflorescontact: done

opencti-platform/opencti-graphql/src/modules/singleSignOn/providers-initialization.js
SouadHadjiat: for me this block is not right, since it is executed at each step of the loop on providers strategy, it should be outside of the loop, replacing the next block.

SouadHadjiat: @aHenryJard

opencti-platform/opencti-graphql/src/modules/singleSignOn/providers-initialization.js
aHenryJard: hasLocal means that there is one local strategy in the PROVIDERS list (kind of cache of providers registered) ; when willLocalBeInDatabase means that there is a local strategy in env that will be converted to database sso

opencti-platform/opencti-graphql/src/modules/singleSignOn/providers-initialization.js
aHenryJard: in master code it's inside the for loop (from top of my head), but I agree that it should be outside

SouadHadjiat (approved):

Issue #13645, opened 2025-12-11
feat(ui): drawer redesign


sandyghs:
Revise the layout and hierarchy of drawer.

PR #14197, opened 2026-01-26
[frontend] DatableToolbar drawers

esrevi:
### Proposed changes

* Input select background should be fixed now
image
* Live alerts and digest drawers updated (/dashboard/settings/activity/alerting update and create)
image
* Update Observable drawer
image
* Datatable toolbar drawers (from toolbar in indicators for example, all drawers are updated)
image

### Related issues

* #13645 

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/components/ThemeDark.ts
+++ b/opencti-platform/opencti-front/src/components/ThemeDark.ts
@@ -410,13 +410,12 @@ const ThemeDark = (
       styleOverrides: {
         root: {
           color: text_color,
-          backgroundColor: '#0C1524',
           '& fieldset': {
             border: 'none',
           },
-          '.opencti-drawer-container &': {
-            backgroundColor: 'transparent',
-          },
+        },
+        outlined: {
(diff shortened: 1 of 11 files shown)


opencti-platform/opencti-front/src/components/ThemeDark.ts
```diff
@@ -410,13 +410,12 @@ const ThemeDark = (
       styleOverrides: {
         root: {
           color: text_color,
-          backgroundColor: '#0C1524',
```
lndrtrbn, line 413: You can set transparent here directly

opencti-platform/opencti-front/src/components/ThemeLight.ts
```diff
@@ -412,13 +412,12 @@ const ThemeLight = (
       styleOverrides: {
         root: {
           color: text_color,
-          backgroundColor: '#FFFFFF',
```
lndrtrbn, line 415: You can set transparent here directly

lndrtrbn (approved):

PR #14209, opened 2026-01-27
[frontend] In Processing/Tasks : Label too close from each other (#13303)

Gwendoline-FAVRE-FELIX:
### Proposed changes

* Fix the spacing between tags.

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/13303?reload=1

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/data/tasks/TasksList.jsx
+++ b/opencti-platform/opencti-front/src/private/components/data/tasks/TasksList.jsx
@@ -343,7 +343,10 @@ const TasksList = ({ data, options }) => {
                         {task.actions
                           && R.map(
                             (action) => (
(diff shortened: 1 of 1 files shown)


opencti-platform/opencti-front/src/private/components/data/tasks/TasksList.jsx
```diff
@@ -343,7 +343,10 @@ const TasksList = ({ data, options }) => {
                         {task.actions
                           && R.map(
                             (action) => (
-                              <div key={task.actions.indexOf(action)}>
+                              <div
```
esrevi:
```suggestion
                              <Stack 
                              key={task.actions.indexOf(action)}
                              direction="row"
                              gap={1}
                              >
```

lndrtrbn (approved):

PR #14217, opened 2026-01-27
[frontend] container header 

esrevi:
### Proposed changes

* Update ContainerHeader and StixDomainObjectHeader layout
* New behaviour on subscription, now click on it opens the drawer to subscribe instead of subscribing directly
* Alias tags and Organisation tags are now on another line and should overflow
* Long title now no longer breaks the page layout and should ellipses correctly

### Related issues

* #13303
*

### Checklist


opencti-platform/opencti-front/src/components/common/header/HeaderMainEntityLayout.tsx
```diff
@@ -0,0 +1,116 @@
+import Stack from '@mui/material/Stack';
+import TitleMainEntity from '../typography/TitleMainEntity';
+import { Tooltip } from '@mui/material';
+
+interface HeaderLayoutProps {
+  title: string;
+  hideTitle?: boolean;
+  titleRight?: React.ReactNode;
```
lndrtrbn: Wouldn't be better to explicitly import ReactNode instead of using global React module?

esrevi: no problem !

lndrtrbn (approved):

lndrtrbn (approved):

PR #14219, opened 2026-01-27
[backend/frontend] Make test URL button work for all SSO (#3923)

SarahBocognano:
### Proposed changes

* Add a new mutation and method to check if an URL is reachable and well formatted for SSO config.

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/3923


xfournet:
To be reworked later based on the refactor that has been done

PR #13588, opened 2025-12-08
[backend/frontend] Add new IMEI, ICCID, and IMSI Observables (#3182)

scarletmerlin123:
### Proposed changes

* Added support for three new observable types
  * IMEI
  * ICCID
  * IMSI
* Added relationships 
  * IMEI ↔ ICCID — uses
  * IMSI ↔ IMEI — uses
  * ICCID ↔ IMSI — has
  * IMSI ↔ phone number — uses
  * ICCID ↔ phone number — resolves-to
  * IMEI → MAC address — has

### Related issues
* Related to #3182 
* Connector updates that use changes from this PR [[Import-Document] Add support for extracting phone numbers, IMEI, ICCID, and IMSI (#3182)](https://github.com/OpenCTI-Platform/connectors/pull/5477)

### Checklist

- [x] I consider the submitted work as finished
- [x] I tested the code for its functionality
- [x] I wrote test cases for the relevant uses case (coverage and e2e)
- [ ] I added/update the relevant documentation (either on github or on notion)
- [x] Where necessary I refactored code to improve the overall quality

### Further comments

No further comments

PR #14237, opened 2026-01-28
[backend/frontend] Add new IMEI, ICCID and IMSI Observables (#3182)

### LINKED ISSUE #3182: Add new observable types: IMEI / ICCID / IMSI
## Use case

Add IMEI, ICCID & IMSI observables:

**IMEI = International Mobile Equipment Identity**
It’s a unique serial number for your phone.
Every device has its own IMEI.
It helps identify a phone if it’s lost, stolen, or used on a mobile network.

**ICCID = Integrated Circuit Card Identifier**
This is the unique serial number of your SIM card, printed on the SIM itself.
It tells the network which SIM card you are using.
It does not identify you; it identifies the physical SIM card.

**IMSI = International Mobile Subscriber Identity**
This identifies you as a subscriber in the mobile network.
It’s stored inside the SIM card, not on the phone.
The IMSI tells the mobile network which user you are, so it can connect you to your account, phone number, and services.

In terms of implementation we should:
- implement the 3 observables types, with the following fields:

Value
IMEI: 14 digits + 1 check digit (15 or 16 for legacy digits total), numeric only
Regxp to validate: ^[0-9]{15}$
ICCID: up to 19–20 digits, numeric only; often starts with an industry identifier (e.g. 89)
Regexp to validate: ^[0-9]{18,22}$
IMSI: usually 15 digits (can be 14–15), numeric only; composed of MCC+MNC+MSIN
Regexp to validate: ^[0-9]{14,15}$
Description
Label
Author
Creator
Description
External Ref
Score
- implement the follwoing relations:

IMEI ↔ ICCID — uses
Device uses SIM card (multidirectional)

IMSI ↔ IMEI — uses
Subscriber uses device (multidirectional)

ICCID ↔ IMSI — has
SIM card has subscriber identity (multidirectional).

IMSI ↔ phone number — uses
Subscriber uses this phone number (multidirectional).

 ICCID ↔ phone number — resolves-to
SIM is provisioned with this phone number (shortcut to avoid forcing the modeling ICCD has IMSI uses Phone number, multidirectional).

IMEI → MAC address — has
This device (IMEI) has / is associated with this MAC address(unidirectionnal)

Ensure that we can export them in pycti too

labo-flg:
### Proposed changes

* Added support for three new observable types
  * IMEI
  * ICCID
  * IMSI
* Added relationships 
  * IMEI ↔ ICCID — uses
  * IMSI ↔ IMEI — uses
  * ICCID ↔ IMSI — has
  * IMSI ↔ phone number — uses
  * ICCID ↔ phone number — resolves-to
  * IMEI → MAC address — has

### Related issues
* Closes #3182 

### Checklist

### Further comments

This PR initial state is a mirror of https://github.com/OpenCTI-Platform/opencti/pull/13588 by @scarletmerlin123
We identified some problems that need to be addressed before merging this new feature again.


labo-flg:
thanks @lndrtrbn !

The e2e test you added is a good start. We need to expand the suite and test more observables types.

opencti-platform/opencti-front/src/private/components/observations/stix_cyber_observables/StixCyberObservableCreation.jsx
```diff
@@ -388,6 +397,10 @@ const StixCyberObservableCreation = ({
           });
         }
       }
+      // remove any non-numbers from IMEI on submit
+      if (bulkConf.type === 'IMEI') {
```
labo-flg: This will cause an error at submit for any observable type not having a `bulkConf` set above.

See observables in https://github.com/OpenCTI-Platform/opencti/pull/14237/changes#diff-5c42a99504e22c92ba6bb3e64786571bf0138a653dc92ffffa8158fb120d5104L230

All other types of observables would have `bulkConf` to `undefined`

quick fix would be to simply use `bulkConf?.type`, but actually this code brings some functional questions.
Is this value adaptation really necessary ? User inputs something and it gets modified in their back.

scarletmerlin123: Hi, the reason for this is to handle if a user were to put in an IMEI that is formatted differently. I wanted to add this handling because I've seen some instances where IMEI is reported as 11-222222-333333-4 instead of only numbers for better readability. 

The position of the dashes (or special character or spaces) varies, but the total amount of numbers remains the same. So for consistency with the way it is typically reported I wanted it to be saved into the system with only numbers. I could remove this additional handling and validate so IMEI needs to be numeric if you want.

lndrtrbn: Hello @scarletmerlin123, I see the point of simplify the user life by accepting dashes and removing them later but I think it can be confusing for the user to have a value saved that is different from what they typed and it complicates our technical solution. For a first version maybe we can do something simpler. I will take time tomorrow morning (west Europe timezone) to review the pull request and guide you in a direction where everything works as expected and where we can move quickly to a merge :+1:

client-python/pycti/utils/constants.py
```diff
@@ -757,3 +760,63 @@ class CustomObservableSshKey:
...
+            "object_marking_refs",
+            ListProperty(
+                ReferenceProperty(valid_types="marking-definition", spec_version="2.1")
+            ),
+        ),
+    ],
+    ["value"],
+)
+class CustomObservableIMEI:
+    """IMEI observable."""
```
lndrtrbn: Extend doc

client-python/pycti/utils/constants.py
```diff
@@ -757,3 +760,63 @@ class CustomObservableSshKey:
...
+            "object_marking_refs",
+            ListProperty(
+                ReferenceProperty(valid_types="marking-definition", spec_version="2.1")
+            ),
+        ),
+    ],
+    ["value"],
+)
+class CustomObservableICCID:
+    """ICCID observable."""
```
lndrtrbn: Extend doc

client-python/pycti/utils/constants.py
```diff
@@ -757,3 +760,63 @@ class CustomObservableSshKey:
...
+            "object_marking_refs",
+            ListProperty(
+                ReferenceProperty(valid_types="marking-definition", spec_version="2.1")
+            ),
+        ),
+    ],
+    ["value"],
+)
+class CustomObservableIMSI:
+    """IMSI observable."""
```
lndrtrbn: Extend doc

opencti-platform/opencti-front/src/private/components/observations/stix_cyber_observables/StixCyberObservableCreation.jsx
```diff
@@ -388,6 +397,10 @@ const StixCyberObservableCreation = ({
...
+        adaptedValue.value = adaptedValue.value.replace(/[^0-9]/g, '');
```
lndrtrbn: To remove

opencti-platform/opencti-front/src/private/components/observations/stix_cyber_observables/StixCyberObservableCreation.jsx
```diff
@@ -618,6 +631,30 @@ const StixCyberObservableCreation = ({
                   ['hashes_SHA-256', 'name'],
                   ['hashes_SHA-512', 'name'],
                 ];
+              } else if (status.type === 'IMEI') {
+                const imeiRegex = /(\d{2})([^a-z\d]{1})?(\d{4})([^a-z\d]{1})?(\d{2})([^a-z\d]{1})?(\d{6})([^a-z\d]{1})?(\d{1,2})$/i;
```
lndrtrbn: Simplify with only digits

opencti-platform/opencti-front/src/private/components/observations/stix_cyber_observables/StixCyberObservableCreation.jsx
```diff
@@ -618,6 +631,30 @@ const StixCyberObservableCreation = ({
...
+                extraFieldsToValidate = {
+                  [attribute.value]: Yup.string().required(t_i18n('This field is required')).matches(imeiRegex, t_i18n('IMEI values must be 15 to 16 digits. Special characters are accepted as delimiters.')),
```
lndrtrbn: Update error message accordingly

opencti-platform/opencti-graphql/src/database/stix-2-1-converter.ts
```diff
@@ -1211,7 +1214,57 @@ const convertSSHKeyToStix = (instance: StoreCyberObservable, type: string): SCO.
...
+    description: instance.x_opencti_description,
+    score: instance.x_opencti_score,
+    created_by_ref: instance[INPUT_CREATED_BY]?.standard_id,
+    external_references: buildExternalReferences(instance),
+    extensions: {
+      [STIX_EXT_OCTI]: stixCyberObject.extensions[STIX_EXT_OCTI],
+      [STIX_EXT_OCTI_SCO]: { extension_type: 'new-sco' },
+    },
+  };
+};
```
lndrtrbn, line 1254: @SouadHadjiat cannot it be simplified by the following:

```ts
const convertIMEIToStix = (instance: StoreCyberObservable, type: string): SCO.StixIMEI => {
  assertType(ENTITY_IMEI, type);
  const stixCyberObject = buildStixCyberObservable(instance);
  return {
    ...stixCyberObject,
    value: instance.value,
    // No need for the rest as it's already present in extension
    // create by the function buildStixCyberObservable
  };
};

File: opencti-platform/opencti-graphql/src/database/stix-2-1-converter.ts [hunk 9]
```diff
@@ -1211,7 +1214,57 @@ const convertSSHKeyToStix = (instance: StoreCyberObservable, type: string): SCO.
...
+const convertICCIDToStix = (instance: StoreCyberObservable, type: string): SCO.StixICCID => {
```
lndrtrbn, line 1256: Same remark for simplification

opencti-platform/opencti-graphql/src/database/stix-2-1-converter.ts
```diff
@@ -1211,7 +1214,57 @@ const convertSSHKeyToStix = (instance: StoreCyberObservable, type: string): SCO.
...
+    score: instance.x_opencti_score,
+    created_by_ref: instance[INPUT_CREATED_BY]?.standard_id,
+    external_references: buildExternalReferences(instance),
+    extensions: {
+      [STIX_EXT_OCTI]: stixCyberObject.extensions[STIX_EXT_OCTI],
+      [STIX_EXT_OCTI_SCO]: { extension_type: 'new-sco' },
+    },
+  };
+};
+const convertIMSIToStix = (instance: StoreCyberObservable, type: string): SCO.StixIMSI => {
```
lndrtrbn, line 1274: Same remark for simplification

opencti-platform/opencti-graphql/tests/03-integration/02-resolvers/stixCyberObservable-test.js
```diff
@@ -441,6 +462,227 @@ describe('StixCyberObservable resolver standard behavior', () => {
     expect(queryResult).not.toBeNull();
     expect(queryResult.data.stixCyberObservable).toBeNull();
   });
+  it('should not create invalid ICCID stixCyberObservable', async () => {
```
lndrtrbn: Move tests to separate files

opencti-platform/opencti-graphql/src/types/stix-2-1-sco.d.ts
```diff
@@ -333,6 +333,54 @@ export interface StixPaymentCard extends StixCyberObject {
   };
 }
 
+// Custom object extension - IMEI
+// value
+export interface StixIMEI extends StixCyberObject {
```
lndrtrbn, line 338: @SouadHadjiat cannot it be simplified with:

```ts
export interface StixIMEI extends StixCyberObject {
  value: string;
}
```

as the other props are already defined in extension of `StixCyberObject`

opencti-platform/opencti-graphql/src/types/stix-2-1-sco.d.ts
```diff
@@ -333,6 +333,54 @@ export interface StixPaymentCard extends StixCyberObject {
...
+  external_references: Array<StixInternalExternalReference>;
+  extensions: {
+    [STIX_EXT_OCTI]: StixOpenctiExtension;
+    [STIX_EXT_OCTI_SCO]: { extension_type: 'new-sco' };
+  };
+}
+
+// Custom object extension - ICCID
+// value
+export interface StixICCID extends StixCyberObject {
```
lndrtrbn, line 354: Same remark on simplification

opencti-platform/opencti-graphql/src/types/stix-2-1-sco.d.ts
```diff
@@ -333,6 +333,54 @@ export interface StixPaymentCard extends StixCyberObject {
...
+  external_references: Array<StixInternalExternalReference>;
+  extensions: {
+    [STIX_EXT_OCTI]: StixOpenctiExtension;
+    [STIX_EXT_OCTI_SCO]: { extension_type: 'new-sco' };
+  };
+}
+
+// Custom object extension - IMSI
+// value
+export interface StixIMSI extends StixCyberObject {
```
lndrtrbn, line 370: Same remark on simplification

lndrtrbn (commented):
**1. Global approach**

To manage IMEI, ICCID and IMSI values correctness in frontend: code does not make any implicit transformation, user has to enter a valid value without any dashes. If any other character than number is entered or length is invalid, the input is red with an error message on it, the form cannot be submitted.

> It's what you have done already for ICCID and IMSI, need to do the same for IMEI.


**2. Client Python**

_EDIT: this part is done_

Code looks good to me, just the documentation in `constans.py` can be extended a bit to be similar to other observables.
For example for `IMEI`:

```py
"""IMEI observable.

Represents an International Mobile Equipment Identity which is a phone serial number.

:param value: The IMEI value (required)
:type value: str
:param spec_version: STIX specification version, fixed to "2.1"
:type spec_version: str
:param object_marking_refs: List of marking definition references
:type object_marking_refs: list
"""
```

**3. Frontend**

_EDIT: this part is done_

Most of the changes are good, in the STIX Cyber Observable creation form (`StixCyberObservableCreation.jsx`):

- Simplify the regex that validate the IMEI value, only digits allowed, regex should be the same in frontend and backend,
- Remove the transformation of IMEI value in the submit function.

Add e2e tests that, for each new observable type: create one, create multiple in bulk.

> For e2e tests, I have added a commit that contains an example of observable creation, I will also add a test for bulk creation for example.

When modifications are done, verify if some translations are no more used and remove them.

**4. Backend**

_EDIT: this part is done_

Verify that stix converter is ok in `stix-2-1-converter.ts`. I'll ask confirmation but I think the following is enough (@SouadHadjiat do you confirm?):

```ts
const convertIMEIToStix = (instance: StoreCyberObservable, type: string): SCO.StixIMEI => {
  assertType(ENTITY_IMEI, type);
  const stixCyberObject = buildStixCyberObservable(instance);
  return {
    ...stixCyberObject,
    value: instance.value,
    // No need for the rest as it's already present in extension
    // create by the function buildStixCyberObservable
  };
};
```

If my thoughts are confirmed, make the changes for the 3 observables.

On the same topic of stix conversion in `stix-2-1-sco.d.ts`. If I am correct, classes can be simplified to:

```ts
export interface StixIMEI extends StixCyberObject {
  value: string;
}
```

ABout tests, the test suite on observable CRUD operation is good. It would be nice to put them in separate files instead of all in `stixCyberObservable-test.js` as it is already a huge file.

I suggest having 3 files: `imei-test.ts`, `imsi-test.ts`, `iccid-test.ts` in the same folder as `stixCyberObservable-test.js`.
 
> Note that new files are in TypeScript.

And on each file add some tests on the new relationships that can be created. There is examples in the file `stixCoreRelationship-test.js`.

---

As you can see there is not that much to change, mainly tests to add.

Thank you! You can ping me here when modifications are done.

opencti-platform/opencti-graphql/src/utils/syntax.js
```diff
@@ -146,6 +146,9 @@ export const macAddrChecker = /^([0-9A-Fa-f]{2}[:-]){5}([0-9A-Fa-f]{2})$/;
 export const ipv4Checker = /^(?:(?:25[0-5]|2[0-4][0-9]|1[0-9]{2}|[1-9]?[0-9])\.){3}(?:25[0-5]|2[0-4][0-9]|1[0-9]{2}|[1-9]?[0-9])(?:\/([0-9]|[1-2][0-9]|3[0-2]))?$/;
 export const cpeChecker = /^cpe:\/\/[a-zA-Z0-9_./:-]+|^cpe:\/[a-zA-Z0-9_./:-]+$/;
 export const fintelTemplateVariableNameChecker = /^[A-Za-z0-9_-]+$/;
+export const imeiChecker = /(^[0-9]{15,16})$/;
```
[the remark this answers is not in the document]
lndrtrbn: Length of 16 is for legacy and is intentional

opencti-platform/opencti-graphql/src/database/stix-2-1-converter.ts
JeremyCloarec: no, buildStixCyberObservable builds the attributes in the extensions, but not at the root of the SCO stix object

opencti-platform/opencti-graphql/src/types/stix-2-1-sco.d.ts
lndrtrbn: Nope, IMEI is a custom SCO and so we are not using the extensions

opencti-platform/opencti-front/tests_e2e/observables/emailMessage.spec.ts
fellowseb, line 1: Bonus E2E test :yay: ?

client-python/pycti/utils/constants.py
```diff
@@ -788,3 +791,103 @@ class CustomObservableAIPrompt:
...
+    ],
+    ["value"],
+)
+class CustomObservableICCID:
+    """ICCID observable.
+
+    Represents an unique serial number of a SIM card,
+    printed on the SIM itself.
+
+    Format: up to 19-20 digits, numeric only.
```
fellowseb: Shouldn't this be:

> Format: between 18 and 22 digits, numeric only.

?

opencti-platform/opencti-graphql/config/schema/opencti.graphql
```diff
@@ -13323,6 +13323,399 @@ input AIPromptAddInput {
...
+    orderBy: FileOrdering
+    orderMode: OrderingMode
+    search: String
+    filters: FilterGroup
+  ): FileConnection
+  exportFiles(first: Int): FileConnection
+  editContext: [EditUserContext!]
+  connectors(onlyAlive: Boolean): [Connector]
+  jobs(first: Int): [Work]
+}
```
fellowseb, line 13662: Quite difficult for me to check that this big list of exposed fields is entirely used 😅

fellowseb (approved):
That's a lot of work 👏 .
Will require another PR to update the docs I believe

fellowseb (approved):

PR #14243, opened 2026-01-28
[backend] Fix use of synchronized-upsert header (#14015)

### LINKED ISSUE #14015: Fix use of header for full synchronization

SouadHadjiat:
### Proposed changes

* header synchronized-upsert should only be used by bypass users

### Related issues

* #14015 

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/domain/user.js
+++ b/opencti-platform/opencti-graphql/src/domain/user.js
@@ -1825,7 +1825,7 @@ const internalAuthenticateUser = async (context, req, user) => {
   let authenticatedUser = user;
   const settings = await getEntityFromCache(context, SYSTEM_USER, ENTITY_TYPE_SETTINGS);
   const synchronizedUpsert = req.headers['synchronized-upsert'] === 'true' && isBypassUser(authenticatedUser);
-    const applicantId = req.headers['opencti-applicant-id'];
+  const applicantId = req.headers['opencti-applicant-id'];
   if (applicantId && isBypassUser(authenticatedUser)) {
     const platformUsers = await getEntitiesMapFromCache(context, SYSTEM_USER, ENTITY_TYPE_USER);
     authenticatedUser = platformUsers.get(applicantId) || INTERNAL_USERS[applicantId];
--- a/opencti-platform/opencti-graphql/src/domain/user.js
+++ b/opencti-platform/opencti-graphql/src/domain/user.js
@@ -161,7 +161,7 @@ const roleUsersCacheRefresh = async (context, user, roleId) => {
   await notify(BUS_TOPICS[ENTITY_TYPE_USER].EDIT_TOPIC, users, user);
 };
 
-export const userWithOrigin = (req, user) => {
+export const userWithOrigin = (req, user, originHeaders = {}) => {
   // /!\ This metadata information is used in different ways
   // - In audit logs to identify the user
   // - In stream message to also identifier the user
@@ -183,6 +183,7 @@ export const userWithOrigin = (req, user) => {
     applicant_id: req?.headers['opencti-applicant-id'],
     call_retry_number: req?.headers['opencti-retry-number'],
     playbook_id: req?.headers['opencti-playbook-id'],
+    ...originHeaders,
   };
   return { ...user, origin };
 };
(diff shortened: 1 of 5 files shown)


opencti-platform/opencti-graphql/tests/03-integration/01-database/httpAuthenticatedContext-test.ts
```diff
@@ -0,0 +1,51 @@
...
+      },
+    };
+
+    const executeContext = await createAuthenticatedContext(req, res, 'api-test');
+    expect(executeContext).toBeDefined();
+    expect(executeContext.user).toBeDefined();
+    expect(executeContext.user?.origin).toBeDefined();
+    expect(executeContext.synchronizedUpsert).toBe(true);
+  });
+  // don't know how I can test this yet, since I don't have user_editor api token.
```
aHenryJard: Maybe you can create a new token for this user. Either here or in test query and keep it in a const.

aHenryJard (commented):
Maybe we should add in advance conf that the selected user should be bypass for full syncronization here =>

Image

opencti-platform/opencti-graphql/src/domain/user.js
```diff
@@ -1823,6 +1824,7 @@ export const authenticateUserByUserId = async (context, req, userId) => {
 const internalAuthenticateUser = async (context, req, user) => {
   let authenticatedUser = user;
   const settings = await getEntityFromCache(context, SYSTEM_USER, ENTITY_TYPE_SETTINGS);
+  const synchronizedUpsert = req.headers['synchronized-upsert'] === 'true' && isBypassUser(authenticatedUser);
```
aHenryJard: Do you think we could add an info log when there is a synchronized-upsert and user is not bypass ? I am worry that it will be hard to debug if it occurs and it seems to be an error to use user that is not bypass if this header is present. In the other hand I don't want to give too much information either.

aHenryJard:
It's working as expected locally for me, we just need to find a way to quickly know that the required user is not bypass and it should. Either a warning in UI or a log (or if you have a better idea it's open)

opencti-platform/opencti-graphql/tests/03-integration/01-database/httpAuthenticatedContext-test.ts
fellowseb, line 1: Is `test/03-integration/01-database` the default folder to do integration tests when there's no other place ? (:.

opencti-platform/opencti-graphql/src/domain/user.js
fellowseb: @aHenryJard  What do you mean by "I don't want to give too much information either" ? That you don't want too many logs ? (if so agreed on both, would add the error/warn log)

opencti-platform/opencti-graphql/tests/03-integration/01-database/httpAuthenticatedContext-test.ts
fellowseb: From what I tested we can indeed call the `userAdminTokenAdd` mutation in `testQuery#createUser`:

```
const USER_TOKEN_ADD_MUTATION = `
  mutation UserTokenAdd($userId: ID!) {
    userAdminTokenAdd(userId: $userId, input: {
      duration: UNLIMITED
      name: "TOKEN_FOR_TESTS"
    }){
      plaintext_token
    }
  }
`;
...
      const userCreationResult = await internalAdminQuery(USER_CREATION_MUTATION, {
        email: user.email,
        name: user.email,
        password: user.password,
      });
      const tokenCreationResult = await internalAdminQuery(USER_TOKEN_ADD_MUTATION, {
        userId: userCreationResult.data.userAdd.id,
      });
```
Then mutate test users or create a map next to them.
The annoying part is that all this happens in `globalSetup` so in another process that you won't have access to from tests. Using the `provide` / `inject` vitest api (https://vitest.dev/config/globalsetup.html#globalsetup) would work, but you need to JSON.serialize that map of userId => tokens.

SouadHadjiat:
> It's working as expected locally for me, we just need to find a way to quickly know that the required user is not bypass and it should. Either a warning in UI or a log (or if you have a better idea it's open)

@aHenryJard It's the worker user, that needs to be bypass anyway, it's not the user that is impersonated. I don't see where we should put a warning to say that the worker user needs to be bypass.

opencti-platform/opencti-graphql/src/domain/user.js
SouadHadjiat: this can't happen, because it's the worker user, this check is done before we resolve the applicant id that will be the user that we impersonate.

aHenryJard: seen together actually it's the worker user to it's more simple.

aHenryJard (approved):
Tested locally, it's all working as expected.

SouadHadjiat:
Can I merge this PR ? @aHenryJard

aHenryJard:
> Can I merge this PR ? @aHenryJard

yes go !

SouadHadjiat:
after discussion today, we need to throw an exception if the header is used and user is not bypass

aHenryJard (approved):
Tested again today, all good to me. I tested both with a bypass user token on worker and a not bypass and it has worked as expected.

PR #14309, opened 2026-02-02
[frontend] connector status icon (#13303)

esrevi:
### Proposed changes

* Change itemboolean to icon true/false 
image

### Related issues

* #13303 

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/data/connectors/ConnectorsStatus.tsx
+++ b/opencti-platform/opencti-front/src/private/components/data/connectors/ConnectorsStatus.tsx
@@ -1,44 +1,47 @@
-import React, { FunctionComponent, useCallback, useEffect, useMemo, useState } from 'react';
-import { interval } from 'rxjs';
-import Dialog from '@mui/material/Dialog';
-import DialogContent from '@mui/material/DialogContent';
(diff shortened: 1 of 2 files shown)


opencti-platform/opencti-front/src/private/components/data/connectors/ConnectorsStatus.tsx
```diff
@@ -436,11 +440,15 @@ const ConnectorsStatusContent: FunctionComponent<ConnectorsStatusContentProps> =
...
-                              <ItemBoolean
-                                label={connector.is_managed ? 'TRUE' : 'FALSE'}
-                                status={connector.is_managed}
-                              />
+                            <div
+                              className={classes.bodyItem}
+                              style={{ justifyContent: 'center' }}
+                            >
+                              {
+                                connector.is_managed
```
lndrtrbn: If it's something that will be used more globally in the platform, should we not create a component instead?

esrevi: i added a component BooleanStatusIcon

lndrtrbn (approved):

PR #14321, opened 2026-02-03
[backend/frontend] using cyberark with new SSO implementation (#3923)

aHenryJard:
### Proposed changes

* Move mandatory fields check after cyberark resolution
* Add documentation on docker compose usage for development
* Change cyberark mock to match local OpenID with keycloak
* Update update drawer to allow edit missing mandatory fields in OpenID

### Related issues

* relates to #3923

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/modules/singleSignOn/singleSignOn-migration.ts
+++ b/opencti-platform/opencti-graphql/src/modules/singleSignOn/singleSignOn-migration.ts
@@ -27,6 +27,7 @@ const NO_CONFIGURATION_KEY = ['label', 'disabled'];
 
 const GROUP_MANAGEMENT_KEY = 'groups_management';
 const ORG_MANAGEMENT_KEY = 'organizations_management';
+const CREDENTIALS_PROVIDER_KEY = 'credentials_provider';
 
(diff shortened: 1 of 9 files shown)


opencti-platform/opencti-graphql/src/modules/singleSignOn/singleSignOn-migration.ts
```diff
@@ -173,6 +174,9 @@ const computeConfiguration = (envConfiguration: any, strategy: StrategyType) =>
         }
 
         organizations_management['organizations_mapping'] = organizations_mapping ?? [];
+      } else if (configKey === CREDENTIALS_PROVIDER_KEY) {
+        skipped_configuration.push(configKey);
```
xfournet, line 178: only the key is skipped, does it mean the rest of the configuration will work, but migrated configuration won't be enriched ?

aHenryJard: it means that the other fields will be migrated, and since enrichment is still with env var it will work immediately. They need to keep the cyberark part in providers env config.

xfournet (approved):

PR #14328, opened 2026-02-03
[frontend] Moved buttons and search inputs in the right places in Darwers (#13303)

MTorbay-Filigran:
### Proposed changes

* Moved buttons and search inputs from the header to the top of the lists in Darwers

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/13303
*

### Checklist


opencti-platform/opencti-front/src/private/components/techniques/data_sources/AddDataComponents.tsx
```diff
@@ -49,31 +50,31 @@ const AddDataComponents: FunctionComponent<{
         open={open}
         onClose={handleClose}
         title={t_i18n('Add data components')}
-        header={(
-          <>
+      >
+        <Stack gap={2}>
```
esrevi: maybe it could be worth to have a component layout to address the drawer content such as DrawerBodyLayout for example and it would take a left and right components and we could use like, so it would be more maintainable ? 
```
<DrawerBodyLayout
header={
  left: <SearchInput />
  right: (
  <>
  <DataComponentCreation />
  ...
  <OtherComponent ... />
  </>
)}
>
  {queryRef && ... }
</DrawerBodyLayout
```

wdyt?

esrevi (approved):

PR #14334, opened 2026-02-03
[frontend] fix togglebuttons, buttons in headers (#13303)

esrevi:
### Proposed changes

* Add additionalHeaderToggleButtons to not mix up with single buttons to have more coherent layout
* Update toggle button style

### Related issues

* #13303 

### Checklist


opencti-platform/opencti-front/src/components/dataGrid/DataTableFilters.tsx
```diff
@@ -147,35 +145,53 @@ const DataTableFilters = ({
...
             )}
           </ToggleButtonGroup>
+
+          {
+            additionalHeaderButtons && (
+              <Stack
+                direction="row"
+                gap={1}
+                sx={{
+                  '&:empty': {
```
esrevi, line 185: need that style because, when for example, we set the StixCoreObjectForms here, so the additionalHeaderButtons lenght > 0 but the component itself can return null, so to avoid having an empty div that breaks the layout, check if there are children or display none if none

opencti-platform/opencti-front/src/components/dataGrid/DataTableFilters.tsx
```diff
@@ -138,7 +136,7 @@ const DataTableFilters = ({
             />
           )}
         </div>
-        <div style={{ display: 'flex' }}>
+        <div style={{ display: 'flex', gap: '8px' }}>
```
lndrtrbn: Prefer using theme.spacing instead of hard value

opencti-platform/opencti-front/src/components/CreateEntityControlledDial.tsx
```diff
@@ -34,20 +30,19 @@ const CreateEntityControlledDial: FunctionComponent<CreateEntityControlledDialPr
   const currentAccessRight = useGetCurrentUserAccessRight(draftContext?.currentUserAccessRight);
   const canDisplayButton = !draftContext || currentAccessRight.canEdit;
 
-  return canDisplayButton ? (
+  if (!canDisplayButton) return;
```
lndrtrbn: return null ? More than undefined to be consistent with other React components

lndrtrbn (approved):

PR #14349, opened 2026-02-04
[frontend] tag width and global workflow update (#13303)

esrevi:
### Proposed changes

* Add max width for alias and organization tags to 150px
* Update global workflow with tags and move edition button at the end of the workflow
* Update security coverage icon size

### Related issues

* #13303 

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectSharingList.tsx
+++ b/opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectSharingList.tsx
@@ -85,7 +85,6 @@ const StixCoreObjectSharingList = ({ data, disabled, inContainer, children }: St
       renderTag={(organization) => (
         <Tag
           label={organization.name}
-          disableTooltip
           onDelete={() => removeOrganization(organization.id)}
           disabled={fullyDisabled || disabledOrgs.includes(organization.id)}
           icon={<AccountBalanceOutlined fontSize="small" />}
--- a/opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectSharingList.tsx
+++ b/opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectSharingList.tsx
@@ -89,6 +89,7 @@ const StixCoreObjectSharingList = ({ data, disabled, inContainer, children }: St
           onDelete={() => removeOrganization(organization.id)}
           disabled={fullyDisabled || disabledOrgs.includes(organization.id)}
           icon={<AccountBalanceOutlined fontSize="small" />}
+          maxWidth={150}
         />
       )}
       direction="rtl"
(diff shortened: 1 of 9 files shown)


opencti-platform/opencti-front/src/private/components/settings/sub_types/entity_setting/EntitySettingSettings.tsx
```diff
@@ -137,7 +137,7 @@ const EntitySettingSettings = ({ entitySettingsData }: EntitySettingSettingsProp
                   }
                 />
               )}
-              label={t_i18n('Hide in the platform')}
+              label={<Typography variant="body2">{t_i18n('Hide in the platform')}</Typography>}
```
lndrtrbn: I suppose we want this style for every Switch label we have in the platform? In that case I think it would be better to have our own custom Switch component that define this style. What do you think?

opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectSharingList.tsx
```diff
@@ -89,6 +89,7 @@ const StixCoreObjectSharingList = ({ data, disabled, inContainer, children }: St
           onDelete={() => removeOrganization(organization.id)}
           disabled={fullyDisabled || disabledOrgs.includes(organization.id)}
           icon={<AccountBalanceOutlined fontSize="small" />}
+          maxWidth={150}
```
lndrtrbn, line 91: We need a tooltip now that value can be truncated

esrevi: actually, the Tag component already handles a tooltip

lndrtrbn: That is not working on my side

esrevi: ow ok, let me check :)

opencti-platform/opencti-front/src/private/components/settings/sub_types/entity_setting/EntitySettingSettings.tsx
esrevi: I have seen with Design, i'm reverting that, since the switch for that specific part may need reworking.
I agree that having a custom Switch component could be nice to have consistent toggle switch in the app. Let's do this later when we'll work on that

opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectSharingList.tsx
esrevi: the reason was i set disableTooltip explicitly on Tag, now i remove it to fix that !

lndrtrbn (approved):

PR #14352, opened 2026-02-04
[frontend] UI: Global kill chain knowledge list (#13303)

Archidoit:
### Proposed changes

- improve empty description rendering

- Rework UI in : 
a malware > knowledge tab > bottom of the page > 'global kill chain' tab

## before
image

## after

image

### Related issues
#13303

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/common/stix_domain_objects/StixDomainObjectGlobalKillChain.jsx
+++ b/opencti-platform/opencti-front/src/private/components/common/stix_domain_objects/StixDomainObjectGlobalKillChain.jsx
@@ -2,6 +2,7 @@ import React, { Component } from 'react';
 import * as PropTypes from 'prop-types';
 import * as R from 'ramda';
 import { Link } from 'react-router-dom';
+import { Box, ListItemButton } from '@mui/material';
 import withStyles from '@mui/styles/withStyles';
 import IconButton from '@common/button/IconButton';
 import List from '@mui/material/List';
@@ -11,7 +12,6 @@ import Collapse from '@mui/material/Collapse';
 import { Launch } from 'mdi-material-ui';
 import { ExpandLess, ExpandMore } from '@mui/icons-material';
 import { createRefetchContainer, graphql } from 'react-relay';
-import { ListItemButton } from '@mui/material';
 import ListItem from '@mui/material/ListItem';
 import { yearFormat } from '../../../../utils/Time';
 import inject18n from '../../../../components/i18n';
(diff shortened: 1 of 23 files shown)


opencti-platform/opencti-front/src/private/components/common/stix_domain_objects/StixDomainObjectGlobalKillChain.jsx
```diff
@@ -185,45 +187,41 @@ class StixDomainObjectGlobalKillChainComponent extends Component {
...
-                                  )
+                                  !restricted
+                                    ? (entityToDisplay.entity_type === 'Attack-Pattern'
+                                        ? defaultRender(
+                                            <span>
+                                              <strong>
+                                                {entityToDisplay.x_mitre_id}
+                                              </strong>{'  '}
+                                              - {entityToDisplay.name}
+                                            </span>,
```
esrevi: we always show the "-" separator between x_mitre_id and name, but if either field is empty, we end up with a "-" (like "MitreId1234 - " or " - EntityName)

opencti-platform/opencti-front/src/private/components/common/stix_domain_objects/StixDomainObjectGlobalKillChain.jsx
```diff
@@ -185,45 +187,41 @@ class StixDomainObjectGlobalKillChainComponent extends Component {
...
+                                    ? (
                                         <MarkdownDisplay
                                           content={stixDomainObject.description}
                                           remarkGfmPlugin={true}
                                           commonmark={true}
                                         />
-                                      ) : (
-                                        t('No description of this usage')
                                       )
+                                    : EMPTY_VALUE
```
esrevi: i wonder if we want to display the EMPTY_VALUE here, because we will always have a "-" in secondary text when no description

Archidoit: I'll keep it with EMPTY_VALUe since it's what is done everywhere: 

example here in Malware cards, for the description : 
image

opencti-platform/opencti-front/src/private/components/common/stix_domain_objects/StixDomainObjectGlobalKillChain.jsx
Archidoit: done

Archidoit:
Final render after review: 

image

esrevi (approved):

PR #14364, opened 2026-02-05
[backend/frontend] do not start disabled strategy on init and use Authentication naming (#3923)

aHenryJard:
#### Code

* [backend] Refactored registerStrategy function to only process enabled strategies (except LocalStrategy which has special handling)
* [backend] Add creation of local SSO entity on new platform (reason to add +1 to elastic counters)
* [frontend] Update all frontend to use "Authentication" instead of "SSO"
* [frontend] Change url part for authentication (single_sign_on to authentication)
* [backend] Fix label is identifier for OpenID

#### CI & tests
* Modified test setup to call initializeAuthenticationProviders during global test initialization
* Add enterprise license to rule tests since the elastic counter count everything and now there is an SSO entity with enterprise and not one with CE.
* Update test error message when worker is needed

### Related issues

* relates to #3923 

### Checklist


RBourgeoisGH (approved):
works locally

PR #14367, opened 2026-02-05
[backend/frontend/api/worker] Introduce new token management system

### LINKED ISSUE #13948: Introduce new token management system
## Use case

- each user can have a set to token (instead a single token as today)
- when new user is created, no token is automatically created
- token can be created
    - the following parameters are chosen by the user
        - a name (unique for a single user)
        - an expiration date
            - 30 days by default
            - can be customized, including set to unlimited
    - token value is randomly generated at creation time
        - token format `flgrn_octi_tkn_[A-Za-z0-9]{64}`
            - prefix will be registered in secret/token scan database (such as [[GitHub secret scanning patterns](https://docs.github.com/en/code-security/secret-scanning/introduction/supported-secret-scanning-patterns)](https://docs.github.com/en/code-security/secret-scanning/introduction/supported-secret-scanning-patterns), [[Gitleaks](https://github.com/gitleaks/gitleaks/blob/master/config/gitleaks.toml)](https://github.com/gitleaks/gitleaks/blob/master/config/gitleaks.toml), etc…)
        - token is not stored in plain text, instead we store
            - the SHA256 of the token
            - to permit token identification, an ellipsed version the token is stored. All random characters are ellipsed expect the 4 first one and 4 last one, eg `flgrn_octi_tkn_45dA(...)Zt0s`
        - token value can be accessed once at creation time
            - token create API returns the clear value of the token
            - UI display this value when token has been created
    - an UUID is also generated to facilitate API, UI & DB model
- token can be revoked
    - revocation is immediate, whatever the expiration data that was chosen at creation time
- token can be listed, the table show
    - the name
    - the create date
    - the expiration date
    - a flag that indicate if the token is expired
    - the last usage date of the token
    - the elipsed token value
    - an action menu/button to revoke the token
- database storage
    - currently the token is stored in clear text in Elasticsearch in user object
    - new storage will be
        - token information are stored in Elasticsearch, except last usage date for performance reason
        - token last usage date is stored in Redis (~1s debounce per token is acceptable to limit Redis writes)
    - migration
        - hash the existing clear-text token  in elastic and create the corresponding entries in Redis
        - remove the clear-text token in Elastic
        - migrated token are flagged in the user token list to encourage to revoke theses tokens that were stored in clear-text
- security consideration
    - use a secure CSPRNG when creating the token
    - use constant-time comparison algo when comparing hash values to avoid timing attacks
    - token clear value must never be logged

richard-julien:
## Executive Summary

This changeset replaces the legacy **single cleartext `api_token` per user** with a modern **multi-token system** featuring HMAC-hashed storage, expiration policies, per-token usage tracking, and capability-based access control. Connector-to-platform authentication is upgraded from raw token passthrough to **JWT-based mutual authentication** using platform-derived Ed25519 key pairs. A **cache race condition** in [getEntitiesFromCache](file:///Users/julienrichard/Documents/filigran/opencti/core-engine/opencti-platform/opencti-graphql/src/database/cache.ts#121-153) is also fixed.

---

### 1.1 Data Model Migration

The user entity's `api_token: String` field is replaced with `api_tokens: [ApiToken!]!`, a nested object array.

```mermaid
erDiagram
    User {
        string id
        string user_email
    }
    ApiToken {
        string id
        string name
        string hash
        string masked_token
        datetime created_at
        datetime expires_at
    }
    User ||--o{ ApiToken : "api_tokens"
```

Each `ApiToken` stores an **HMAC hash** (never the plaintext), a user-facing `masked_token` (`****xxxx`), a `name`, and optional `expires_at`.

### 1.2 Token Lifecycle

```mermaid
sequenceDiagram
    participant Client
    participant GraphQL
    participant UserDomain as user-domain.ts
    participant PlatformCrypto
    participant ElasticSearch
    participant Redis

    Client->>GraphQL: userTokenAdd(name, duration)
    GraphQL->>UserDomain: addUserToken(context, user, input)
    UserDomain->>UserDomain: generateSecureToken()
    Note over UserDomain: crypto.randomBytes(48) → base64url<br/>prefix: flgrn_octi_tkn_
    UserDomain->>PlatformCrypto: generateTokenHmac(plaintext)
    PlatformCrypto-->>UserDomain: HMAC-SHA256 hash (base64)
    UserDomain->>ElasticSearch: updateAttribute(ADD, api_tokens, {id, name, hash, masked_token, ...})
    UserDomain-->>Client: { token_id, plaintext_token, masked_token, expires_at }
    Note over Client: plaintext_token shown ONCE, never stored server-side
```

### 1.3 GraphQL Schema Changes

**New types:**
- `ApiToken` — [id](file:///Users/julienrichard/Documents/filigran/opencti/core-engine/opencti-platform/opencti-graphql/src/manager/notificationManager.ts#337-368), `name`, `created_at`, `expires_at`, `masked_token`, `last_used_at`
- `TokenGenerated` — `token_id`, `plaintext_token`, `expires_at`, `masked_token` (returned on create)
- `UserTokenAddInput` — `name: String!`, `duration: TokenDuration!`
- `TokenDuration` enum — `DAYS_30`, `DAYS_60`, `DAYS_90`, `DAYS_365`, `UNLIMITED`

**New mutations:**
| Mutation | Auth | Purpose |
|----------|------|---------|
| `userTokenAdd(input)` | `@auth` (self) | Create token for current user |
| `userTokenRevoke(id)` | `@auth` (self) | Revoke own token |
| `userAdminTokenAdd(userId, input)` | `SETTINGS_SETACCESSES` | Admin creates token for any user |
| `userAdminTokenRevoke(userId, id)` | `SETTINGS_SETACCESSES` | Admin revokes token for any user |
| `tokenJWTConnector` | `SETTINGS_SETACCESSES` | Issue short-lived JWT for managed connectors |

**Removed mutations:**
- `meTokenRenew` — replaced by `userTokenAdd` + `userTokenRevoke`
- `UserEditMutations.tokenRenew` — removed

**Field changes:**
- `User.api_token: String!` → `User.api_tokens: [ApiToken!]!`
- `MeUser.api_token: String!` → `MeUser.api_tokens: [ApiToken!]!`
- `Connector.jwks: String!` — new field exposing connector JWKS public key

**Removed types:**
- `MetricAttributes` and `MetricDefinition` (moved/reordered in relay schema only)

### 1.4 Token Authentication Flow

```mermaid
flowchart TD
    A[Incoming Request] --> B{Authorization header?}
    B -->|Bearer ey...| C[JWT Authentication]
    B -->|Bearer flgrn_octi_tkn_...| D[Token Authentication]
    B -->|Basic base64| E[Basic Auth]
    B -->|No auth| F[Session/Cookie check]

    C --> C1[Verify Ed25519 signature via platformCrypto]
    C1 --> C2[Extract userId from JWT sub]
    C2 --> C3[authenticateUserByUserId]

    D --> D1[HMAC hash the token]
    D1 --> D2[Look up user by hash in cache]
    D2 --> D3{APIACCESS_USETOKEN capability?}
    D3 -->|No| D4[ForbiddenAccess]
    D3 -->|Yes| D5{Token expired?}
    D5 -->|Yes| D6[FunctionalError: Token expired]
    D5 -->|No| D7[updateTokenUsage in Redis]
    D7 --> D8[internalAuthenticateUser]

    E --> E1{APIACCESS_USEBASICAUTH capability?}
    E1 -->|No| E2[ForbiddenAccess]
    E1 -->|Yes| E3[login with username/password]
    E3 --> E4[internalAuthenticateUser]
```

**Key changes in `authenticateUserFromRequest`:**
- Token lookup is now hash-based (`platformUsers.get(hashedToken)`) instead of UUID-based
- JWT tokens (prefixed `ey`) are routed to `authenticateUserByJWT`
- Basic auth now extracted into `authenticateUserByBasicAuth` with capability check
- Token expiration is checked at authentication time
- `updateTokenUsage()` is called on successful token auth

### 1.5 New Capabilities

Two new capabilities are introduced under a new `APIACCESS` parent:

| Capability | Purpose |
|---|---|
| `APIACCESS_USETOKEN` | Required to authenticate via API tokens |
| `APIACCESS_USEBASICAUTH` | Required to authenticate via HTTP Basic Auth |

These are registered in `data-initialization.js` and granted to all existing roles via migration.

---

### 2.1 HMAC Token Hashing (`utils/hash.ts`)

New `generateTokenHmac()` function using `platformCrypto.deriveHmac()`:
- Derivation path: `['authentication', 'token']`
- HMAC-SHA256 with key derived from platform secret
- Result: base64-encoded hash stored in `api_tokens[].hash`
- Lazy singleton pattern for key derivation

### 2.2 Secure Token Generation (`utils/security.ts`)

New file — `generateSecureToken()`:
- 48 bytes of `crypto.randomBytes` → base64url (384 bits of entropy)
- Prefix: `flgrn_octi_tkn_` (easily identifiable)
- Returns `{ token, hash, masked_token }`

### 2.3 JWT Connector Authentication

New Ed25519 key pair derived from platform secret:
- Derivation path: `['connector', 'http']` (connector JWKS)
- Derivation path: `['authentication', 'xtm1']` (user JWT)

Two issuance points:
1. **`issueAuthenticationJWT(user)`** — 1-hour JWT for chatbot proxy (`sub: user.id`)
2. **`issueConnectorJwtToken()`** — 1-hour JWT for managed connectors (`sub: 'connector'`, `iss: 'opencti'`)

### 2.4 HMAC Derivation in `platformCrypto.ts`

New `deriveHmac(derivationPath, version)` method added to the crypto key factory, producing deterministic HMAC keys from the platform secret via HKDF.

---

## 3. Token Usage Tracking (`database/redis/token_usage.ts`)

New Redis-based tracking of when each API token was last used:

- **Redis key pattern:** `token_usage:{tokenId}`
- **L1 cache:** LRU cache (5000 entries) to throttle Redis writes
- **Throttle:** Only writes to Redis if last check was >15 min ago (configurable via `app:token_activity_period`)
- **Batch retrieval:** `getTokensUsage(tokenIds[])` using Redis `MGET`
- **Batch loader:** `batchUserTokens` in `user.js` enriches `ApiToken` with `last_used_at` from Redis

---

### Migration 1: `1769815233918-legacy_token_migration.js`

Converts existing `api_token` (cleartext UUID) to the new `api_tokens[]` format:
1. Iterates all users with `fullEntitiesOrRelationsList`
2. For each user with `api_token`: computes HMAC hash → creates `ApiToken` entry with `name: 'Legacy Token'`, `expires_at: null`
3. Sets `api_token: null`, `api_tokens: [...currentTokens, newToken]`
4. Idempotent: checks if hash already exists before creating

### Migration 2: `1769964742100-add-access-token-capability.js`

1. Creates the `APIACCESS` capability tree (`APIACCESS_USETOKEN`, `APIACCESS_USEBASICAUTH`)
2. Grants both capabilities to **all existing roles** via `has-capability` relations
3. Ensures backward compatibility: all existing users retain API token and basic auth access

---

### `internalObject-registrationAttributes.ts`

New `apiTokens` attribute definition (exported for reuse):

```typescript
export const apiTokens: AttributeDefinition = {
  name: 'api_tokens',
  label: 'Api tokens',
  type: 'object',
  format: 'flat',
  multiple: true,
  upsert: true,
  mappings: [
    { name: 'id', type: 'string', format: 'short', ... },
    { name: 'name', type: 'string', format: 'short', ... },
    { name: 'hash', type: 'string', format: 'short', ... },
    { name: 'masked_token', type: 'string', format: 'short', ... },
    { name: 'created_at', type: 'date', ... },
    { name: 'expires_at', type: 'date', ... },
  ],
};
```

Replaces `{ name: 'api_token', type: 'string', ... }` in user attributes.

---

### `database/middleware.js`

The user cache indexing is updated to index users by token **hashes** instead of cleartext UUIDs:

```diff
- userToIndex[user.api_token] = user;
+ for (const token of (user.api_tokens || [])) {
+   userToIndex[token.hash] = user;
+ }
```

This enables O(1) lookup of users by HMAC-hashed token during authentication.

---

### `providers-initialization.js`

- Removed the `externally_managed` admin configuration branch (which previously allowed deleting the admin user)
- Simplified to always require admin email/password/token in config
- The `findById` and `userDelete` imports are no longer needed

### `domain/user.js` — [addUser](file:///Users/julienrichard/Documents/filigran/opencti/core-engine/opencti-platform/opencti-graphql/src/manager/cacheManager.ts#237-250)

- `api_token: uuid()` is no longer auto-generated during user creation
- `$user.api_token` template variable removed from email templates

### [httpAuthenticatedContext.js](file:///Users/julienrichard/Documents/filigran/opencti/core-engine/opencti-platform/opencti-graphql/src/http/httpAuthenticatedContext.js)

- Applied `settings?.otp_mandatory ?? false` for null safety on settings access

---

### Profile Page (`ProfileOverview.jsx`)

- Removed: `api_token` display, show/hide toggle, token renew button, required headers section
- Added: "Generate Token" button (gated by `APIACCESS_USETOKEN` capability)
- Added: `<TokenCreationDrawer>` and `<TokenList>` components
- Removed: `meTokenRenew` GraphQL mutation
- Removed: `api_token` from `ProfileOverview_me` fragment

### New Components

| Component | Purpose |
|---|---|
| `TokenCreationDrawer.tsx` | Drawer wrapper for token creation form |
| `TokenCreationForm.tsx` | Form with name + duration fields, shows plaintext token once |
| `TokenList.tsx` | Lists user's tokens with masked display + last used + revoke action |
| `TokenResultView.tsx` | One-time display of generated plaintext token with copy functionality |

### Admin User Page (`settings/users/User.tsx`)

- Removed: `api_token` display, show/hide, renew button
- Added: `UserTokenCreationDrawer`, `UserTokenList` for admin token management

### `useGranted.ts`

- New constant: `APIACCESS_USETOKEN = 'APIACCESS_USETOKEN'`

### `ItemCopy.tsx`

Refactored copy component with animation and styling improvements.

---

### `opencti_user.py`

- Removed: `token_properties`, `include_token` parameter from [create()](file:///Users/julienrichard/Documents/filigran/opencti/core-engine/opencti-platform/opencti-graphql/src/http/httpAuthenticatedContext.js#40-79), [me()](file:///Users/julienrichard/Documents/filigran/opencti/core-engine/opencti-platform/opencti-graphql/src/manager/notificationManager.ts#209-239), `list()`
- Removed: `token_renew()` method
- Added: `create_token(user_id, token_name, token_duration)` — calls `userAdminTokenAdd` mutation
- Added: `remove_token(user_id, token_id)` — calls `userAdminTokenRevoke` mutation

### `opencti_connector_helper.py`

- On-the-fly connector users now create/renew tokens via `create_token()` instead of relying on `api_token`
- Token rotation: finds existing connector token by name → removes it → creates new one
- Connector JWKS is fetched and passed to the listen queue config

### `opencti_api_connector.py`

- Connector configuration now includes `jwks` field

### `opencti_api_client.py`

- New method: `token_jwt_connector()` — calls `tokenJWTConnector` mutation

### Worker (`opencti-worker`)

- `ListenHandler` now receives the full `OpenCTIApiClient` instead of raw `connector_token`
- On each message dispatch, calls `api.token_jwt_connector()` to get a fresh JWT
- New `is_token_valid()` method for incoming HTTP callback validation using PyJWT + JWKS

### Dependencies

- Added: `pyjwt[crypto]~=2.11.0` (requirements.txt, setup.cfg, test-requirements.txt)

---

### New Test Files

| File | Type | Coverage |
|---|---|---|
| [tests/01-unit/domain/user-auth-test.ts](file:///Users/julienrichard/Documents/filigran/opencti/core-engine/opencti-platform/opencti-graphql/tests/01-unit/domain/user-auth-test.ts) | Unit | Token HMAC authentication, expiration, capability checks |
| `tests/01-unit/modules/user/user-domain-test.ts` | Unit | `addUserToken`, `revokeUserToken`, with mocked middleware |
| `tests/01-unit/utils/security-test.ts` | Unit | `generateSecureToken` format/uniqueness |
| `TokenCreationForm.test.tsx` | Frontend | Form rendering and submission |
| `TokenList.test.tsx` | Frontend | Token list display and revoke action |
| `TokenResultView.test.tsx` | Frontend | Plaintext token display |
| `UserTokenCreationForm.test.tsx` | Frontend | Admin token creation form |
| `UserTokenList.test.tsx` | Frontend | Admin token list |
| `tests/02-resolvers/user-token-test.ts` | Integration | Full token CRUD lifecycle via GraphQL resolvers |

### Modified Tests

- `tests/01-unit/domain/user-test.ts` — Updated for `api_tokens` field
- `tests/02-resolvers/user-test.ts` — Removed `tokenRenew` tests, updated assertions
- `tests/01-dataCount/elasticSearch-test.js` — Updated entity count expectations
- `tests/utils/testQuery.ts` — Added token creation helpers for test infrastructure

---

## 11. CI / Config

- `.github/workflows/ci-test-frontend.yml` — Minor pipeline update
- `scripts/ci/ci-common.env` / `ci-sync.env` — New env var added
- `opencti-graphql/config/default.json` — Version bump
- `opencti-graphql/config/test.json` — Added `token_activity_period` config
- `opencti-graphql/package.json` — Updated `jose` dependency, added `lru-cache`

---

## Breaking Changes

> [!CAUTION]
> **GraphQL API breaking changes:**
> - `User.api_token` → `User.api_tokens` (type changed from `String!` to `[ApiToken!]!`)
> - `MeUser.api_token` → `MeUser.api_tokens`
> - `meTokenRenew` mutation removed
> - `UserEditMutations.tokenRenew` mutation removed

> [!CAUTION]
> **Python client breaking changes:**
> - `MeUser.api_token` → `MeUser.api_tokens`
> - `token_renew()` method removed, replaced by `create_token()` / `remove_token()`

> [!WARNING]
> **Migration required:**
> - Migration `1769815233918` converts legacy tokens to hashed format — backward compatible but irreversible
> - Migration `1769964742100` creates new capabilities and grants them to all roles


opencti-platform/opencti-graphql/config/schema/opencti.graphql
```diff
@@ -15658,6 +15675,7 @@ type Mutation {
   ######## INTERNAL OBJECT ENTITIES
   frontendErrorLog(message: String!, codeStack: String, componentStack: String): Boolean @auth
   token(input: UserLoginInput): String @public # Use for login
+  tokenJWTConnector: String! @auth(for: [SETTINGS_SETACCESSES])
```
xfournet: naming, tokenJWTConnector -> connectorJWT ? (JWT already mean token)

opencti-worker/src/listen_handler.py
```diff
@@ -14,20 +15,22 @@
...
+    api: OpenCTIApiClient
     callback_uri: str
     listen_api_ssl_verify: bool
     listen_api_http_proxy: str
     listen_api_https_proxy: str
 
     def handle_message(self, body: str) -> Literal["ack", "nack", "requeue"]:
         try:
+            # Round trip with JWT service every 30 minutes with issueConnectorJWT
+            connector_jwt = self.api.token_jwt_connector()
```
xfournet: the token is retrieved for each call instead of every 30 minutes as indicated in comment

opencti-platform/opencti-graphql/tests/03-integration/02-resolvers/threatActorIndividual-test.ts
```diff
@@ -574,7 +574,7 @@ describe('Threat actor individual resolver standard behavior', () => {
       key: 'height',
       value: [],
       object_path: '/height',
-      operation: EditOperation.Remove,
+      operation: EditOperation.Replace,
```
xfournet, line 577: is this change related to api token ? or just the opportunity to fix something else ?

opencti-platform/opencti-graphql/tests/03-integration/02-resolvers/user-test.ts
```diff
@@ -1111,7 +1082,7 @@ describe('meUser specific resolvers', async () => {
         lastname
         language
         theme
-        api_token
+        theme
```
xfournet: theme is duplicated

opencti-platform/opencti-graphql/src/utils/hash.ts
```diff
@@ -18,3 +19,18 @@ export const hashSHA256 = (input: string) => {
 export const compareHashSHA256 = (input: string, hash: string) => {
   return hashSHA256(input) === hash;
 };
+
+/**
+ * Hash a token using hmac algorithm.
+ * @param token the token to hash
+ */
+let hmacDerivationPromise: Promise<{ hmac: (data: string) => string }>;
+const TOKEN_DERIVATION_PATH = ['authentication', 'token'];
```
xfournet: i would make it private to security.ts (or user-domain.ts cf comment on security.ts) since the derivation path is specific to autentication.

opencti-platform/opencti-graphql/src/utils/security.ts
xfournet, line 1: i would not expose as a general 'security' utility since it's rather specific to authentication.
To be moved to modules/user/user-domain ?

opencti-platform/opencti-graphql/src/utils/platformCrypto.ts
```diff
@@ -8,6 +8,7 @@ import { enrichWithRemoteCredentials } from '../config/credentials';
 import { confNameToEnvName } from '../config/conf';
 import { ConfigurationError, UnsupportedError } from '../config/errors';
 
+export const JWT_TOKEN_PREFIX = 'ey';
```
xfournet: it is not the appropriate place for this constant.
Moreover only used in user.js, to be moved here.

opencti-platform/opencti-graphql/src/modules/user/user-domain.ts
```diff
@@ -63,6 +71,208 @@ export const createOnTheFlyUser = async (
...
+    },
+  });
+
+  // Notify for cache invalidation
+  await notify(BUS_TOPICS[ENTITY_TYPE_USER].EDIT_TOPIC, element, user);
+
+  return tokenId;
+};
+
+export const addUserTokenByAdmin = async (context: AuthContext, user: AuthUser, targetUserId: string, input: UserTokenAddInput) => {
```
xfournet: could be factorized with addUserToken, only user selection+capability test differs, all the rest is identical

opencti-platform/opencti-graphql/src/modules/user/user-domain.ts
```diff
@@ -63,6 +71,208 @@ export const createOnTheFlyUser = async (
...
     userInput = { ...userInput, user_confidence_level: { max_confidence: userConfidence, overrides: [] } };
   }
-  const newlyCreatedUser = await addUser(context, user, userInput);
-  return newlyCreatedUser;
+  return await addUser(context, user, userInput);
+};
+
+// -- API Token Logic --
+export const addUserToken = async (context: AuthContext, user: AuthUser, input: UserTokenAddInput) => {
+  if (!isUserHasCapability(user, 'APIACCESS_USETOKEN')) {
```
xfournet: shouldn't be declared and enforced on the schema ?

opencti-platform/opencti-graphql/src/modules/user/user-domain.ts
```diff
@@ -63,6 +71,208 @@ export const createOnTheFlyUser = async (
...
+    },
+  });
+
+  // Notify for cache invalidation
+  await notify(BUS_TOPICS[ENTITY_TYPE_USER].EDIT_TOPIC, element, user);
+
+  return tokenId;
+};
+
+export const revokeUserTokenByAdmin = async (context: AuthContext, user: AuthUser, targetUserId: string, tokenId: string) => {
```
xfournet: could be factorized with revokeUserToken since they are very similar

opencti-platform/opencti-graphql/src/modules/attributes/internalObject-registrationAttributes.ts
```diff
@@ -183,6 +184,25 @@ export const settingsMessages = {
...
+  label: 'API Tokens',
+  type: 'object',
+  format: 'nested',
+  mandatoryType: 'no',
+  editDefault: false,
+  multiple: true,
+  upsert: true,
+  isFilterable: false,
+  mappings: [
+    { name: 'id', label: 'ID', type: 'string', format: 'short', editDefault: false, mandatoryType: 'no', multiple: false, upsert: true, isFilterable: false },
```
xfournet, line 198: what mean format 'short' for a string ?

opencti-platform/opencti-graphql/src/http/httpChatbotProxy.ts
```diff
@@ -33,9 +34,10 @@ export const getChatbotProxy = async (req: Express.Request, res: Express.Respons
       return;
     }
 
+    const jwt = await issueAuthenticationJWT(context.user);
```
xfournet, line 39: in that case it's acceptable to issue a JWT for each call, but need to precise in a comment to avoid this pattern to be used in much intensive scenario.

opencti-platform/opencti-graphql/src/database/repository.js
```diff
@@ -16,9 +16,33 @@ import { SYSTEM_USER } from '../utils/access';
...
 import { injectProxyConfiguration } from '../config/proxy-config';
+import { getPlatformCrypto } from '../utils/platformCrypto';
+import { SignJWT } from 'jose';
 
 export const CONNECTOR_PRIORITY_GROUP_VALUES = Object.values(ConnectorPriorityGroup);
 
+let connectorKeyPairPromise;
+const CONNECTOR_DERIVATION_PATH = ['connector', 'http'];
+export const getConnectorJwks = async () => {
+  const factory = await getPlatformCrypto();
```
xfournet, line 26: no need to retrieve getPlatformCrypto before testing connectorKeyPairPromise (same in every other similar pattern)

opencti-platform/opencti-graphql/src/database/redis/token_usage.ts
```diff
@@ -0,0 +1,54 @@
...
+
+// L1 Cache to avoid hitting Redis for every request
+const l1Cache = new LRUCache<string, DateTime>({
+  max: 5000,
+  // We manage TTL validity manually to ensure consistency with Date.now()
+  // ttl: REDIS_THROTTLE_MS,
+});
+
+export const updateTokenUsage = async (userId: string, tokenId: string) => {
+  const redis = getClientBase();
```
xfournet: not needed before line 31

opencti-platform/opencti-graphql/src/database/redis/token_usage.ts
```diff
@@ -0,0 +1,54 @@
...
+  const key = `${REDIS_TOKEN_USAGE_PREFIX}${tokenId}`;
+  const now = DateTime.now().toUTC();
+  try {
+    // Check L1 Cache first
+    const l1LastChecked = l1Cache.get(key);
+    if (l1LastChecked && now.diff(l1LastChecked, 'minutes').minutes < REDIS_THROTTLE_MIN) {
+      return;
+    }
+    const data: TokenUsage = { user: userId, token: tokenId, last_check: now.toISO() };
```
xfournet: tokenId is already in the key
is userId needed ?
last_check can be just a number rather a date string ?

-> store directly now epoch millis ?

opencti-platform/opencti-graphql/src/config/providers-initialization.js
```diff
@@ -10,45 +10,35 @@ import { Strategy as SamlStrategy } from '@node-saml/passport-saml';
...
 // Admin user initialization
 export const initializeAdminUser = async (context) => {
-  const isExternallyManaged = conf.get('app:admin:externally_managed') === true;
-  if (isExternallyManaged) {
-    logApp.info('[INIT] admin user initialization disabled by configuration');
-    const existingAdmin = await findById(context, SYSTEM_USER, OPENCTI_ADMIN_UUID);
-    if (existingAdmin) {
-      await userDelete(context, SYSTEM_USER, OPENCTI_ADMIN_UUID);
-    }
+  const adminEmail = conf.get('app:admin:email');
```
xfournet: we need to document the removal of app:admin:externally_managed ?
is there an option to disable it by conf ?

client-python/pycti/connector/opencti_connector_helper.py
```diff
@@ -741,6 +744,27 @@ def _data_handler(self, json_data: Dict) -> None:
...
+            unverified_header = jwt.get_unverified_header(token)
+            used_kid = unverified_header["kid"]
+            selected_jwk = jwks.get(used_kid)
+            if selected_jwk is None:
+                self.helper.connector_logger.error(
+                    "Error: Public key not found in JWKS."
+                )
+                return False
+            key = PyJWK.from_dict(selected_jwk)
+            jwt.decode(token, key=key, issuer="opencti", subject="connector")
```
xfournet: checking a hardcoded issuer is not relevant. either we think it's not needed because we are sure the private key used to sign cannot be used by another system (and this won't change over the time) either we need to enforce a specific issuer

client-python/pycti/entities/opencti_user.py
```diff
@@ -766,42 +753,68 @@ def delete_organization(self, **kwargs) -> Optional[Dict]:
...
+        """
+        user_id = kwargs.get("user_id", None)
+        token_name = kwargs.get("token_name", None)
+        token_duration = kwargs.get("token_duration", "UNLIMITED")
+        if user_id is None:
+            self.opencti.admin_logger.error("[opencti_user] Missing parameter: user_id")
+            return None
+
+        self.opencti.admin_logger.info(
+            "Rotating API key for user", {"user_id": user_id}
```
xfournet: it's not a token rotation

client-python/tests/02-integration/entities/test_user.py
```diff
@@ -53,18 +53,3 @@ def test_user_organization(api_client):
     finally:
         user_test.base_class().delete(id=test_user["id"])
         org_test.base_class().delete(id=test_org["id"])
-
-
-def test_user_token_renew(api_client):
```
xfournet, line 58: token creation / remove test to be created ?

client-python/pycti/entities/opencti_user.py
```diff
@@ -766,42 +753,68 @@ def delete_organization(self, **kwargs) -> Optional[Dict]:
             result["data"]["userEdit"]["organizationDelete"]
         )
 
-    def token_renew(self, **kwargs) -> Optional[Dict]:
```
xfournet, line 769: breaking change to be documented ?

opencti-platform/opencti-graphql/src/config/providers-initialization.js
richard-julien: app:admin:externally_managed is not removed, now its change from delete to account lock

opencti-platform/opencti-graphql/tests/03-integration/02-resolvers/threatActorIndividual-test.ts
richard-julien: Yes. Remove action with empty array should not cleanup the field and do nothing, it must be a replace

opencti-platform/opencti-graphql/src/utils/hash.ts
richard-julien: ok

opencti-platform/opencti-graphql/src/utils/security.ts
richard-julien: ok

opencti-platform/opencti-graphql/src/utils/platformCrypto.ts
richard-julien: ok

opencti-platform/opencti-graphql/src/modules/user/user-domain.ts
richard-julien: yes

opencti-platform/opencti-front/lang/back/de.json
```diff
@@ -426,6 +428,7 @@
   "Holder name": "Name des Inhabers",
   "HUB_REGISTRATION_MANAGER": "Hub-Registrierungsmanager",
   "IBAN": "IBAN",
+  "ID": "ID",
```
xfournet, line 431: We should rather reuse "Id" instead of creating "ID"

opencti-platform/opencti-graphql/config/schema/opencti.graphql
```diff
@@ -1395,17 +1402,6 @@ interface IntlSettings {
...
-  attribute: String!
-  name: String!
-  description: String
-}
-
-type MetricDefinition {
-  entity_type: String!
-  metrics: [MetricAttributes!]
-}
-
```
xfournet, line 1407: why is it removed ?

opencti-platform/opencti-graphql/src/modules/attributes/internalObject-registrationAttributes.ts
richard-julien: It means that is will be indexed with keyword to allow exact untokenized search

client-python/pycti/api/opencti_api_client.py
```diff
@@ -1193,6 +1193,16 @@ def get_stix_content(self, id):
         result = self.query(query, {"id": id})
         return json.loads(result["data"]["stix"])
 
+    def token_jwt_connector(self):
```
xfournet: to be renamed to connector_jwt to align with API name

opencti-platform/opencti-graphql/src/modules/user/user-domain.ts
```diff
@@ -63,6 +72,190 @@ export const createOnTheFlyUser = async (
...
+    },
+  });
+  // Notify for cache invalidation
+  await notify(BUS_TOPICS[ENTITY_TYPE_USER].EDIT_TOPIC, element, user);
+  return {
+    token_id: token.id,
+    plaintext_token: token_value,
+    masked_token: token.masked_token,
+    expires_at: token.expires_at,
+  };
```
xfournet, line 139: it could be factorized in addToken (only user action message differ)

opencti-platform/opencti-graphql/src/modules/user/user-domain.ts
```diff
@@ -63,6 +72,190 @@ export const createOnTheFlyUser = async (
...
+      id: user.id,
+      entity_type: ENTITY_TYPE_USER,
+      input: {
+        token_id: tokenId,
+      },
+    },
+  });
+  // Notify for cache invalidation
+  await notify(BUS_TOPICS[ENTITY_TYPE_USER].EDIT_TOPIC, element, user);
+  return tokenId;
```
xfournet, line 178: can be factorized (only user action message differs)

opencti-platform/opencti-graphql/config/schema/opencti.graphql
xfournet: -> it is a duplicate, already present in this file

client-python/pycti/connector/opencti_connector_helper.py
```diff
@@ -762,11 +786,7 @@ async def _http_process_callback(self, request: Request) -> JSONResponse:
         # 01. Check the authentication
         authorization: str = request.headers.get("Authorization", "")
         items = authorization.split() if isinstance(authorization, str) else []
-        if (
-            len(items) != 2
-            or items[0].lower() != "bearer"
-            or items[1] != self.opencti_token
-        ):
+        if self.is_token_valid(items[1], self.connector_jwks) is False:
```
xfournet: we should keep the test on items.lenth and first item = 'bearer'

opencti-platform/opencti-graphql/src/domain/user.js
```diff
@@ -1816,36 +1893,50 @@ export const authenticateUserFromRequest = async (context, req) => {
...
+    existingAdmin = await addUser(context, SYSTEM_USER, userToCreate);
+  }
+  // Create base token if needed
+  const tokenId = 'base_token_' + OPENCTI_ADMIN_UUID;
+  const tokensWithoutBaseOne = (existingAdmin.api_tokens ?? []).filter((t) => t.id !== tokenId);
+  const now = DateTime.now().toUTC().toString();
+  const newToken = {
+    id: tokenId,
+    name: 'Base token',
+    hash: await generateTokenHmac(tokenValue),
```
xfournet, line 1928: should we enforce token value format (prefix, minimal length ?)

opencti-platform/opencti-graphql/src/http/httpChatbotProxy.ts
```diff
@@ -33,9 +34,12 @@ export const getChatbotProxy = async (req: Express.Request, res: Express.Respons
       return;
     }
 
+    // In that case it's acceptable to issue a JWT for each call
+    // Do not use this pattern in other place as is CPU intensive.
+    const jwt = await issueAuthenticationJWT(context.user);
     const vars = {
       OPENCTI_URL: getChatbotUrl(req),
-      OPENCTI_TOKEN: context.user?.api_token,
+      OPENCTI_TOKEN: `Bearer ${jwt}`,
```
xfournet: why adding 'Bearer' prefix here ?

opencti-platform/opencti-graphql/src/domain/user.js
```diff
@@ -1638,13 +1664,70 @@ export const resolveUserById = async (context, id) => {
...
+    // Checking expiration
+    if (matchingToken.expires_at) {
+      const now = new Date();
+      const expiresAt = new Date(matchingToken.expires_at);
+      if (now >= expiresAt) {
+        throw FunctionalError('Token expired');
+      }
+    }
+    // Update token usage
+    if (matchingToken) {
```
xfournet: test can be removed, since we throw at line 1716 if matching token is falsy

opencti-platform/opencti-graphql/src/database/redis/token_usage.ts
```diff
@@ -0,0 +1,54 @@
+import { LRUCache } from 'lru-cache';
+import { getClientBase } from '../redis';
+import { DateTime } from 'luxon';
+import { logApp } from '../../config/conf';
+import nconf from 'nconf';
+
+const REDIS_TOKEN_USAGE_PREFIX = 'token_usage:';
+// Only update activity every 15 minutes by default
+const REDIS_THROTTLE_MIN = nconf.get('app:token_activity_period') ?? 15;
```
xfournet: i still prefer to have 1 minute default so these can be help to diagnose token usage more quickly.
redis won't be affected by this load :)

opencti-platform/opencti-graphql/src/domain/user.js
```diff
@@ -1773,6 +1832,8 @@ export const sessionAuthenticateUser = async (context, req, user, provider) => {
 };
 
 export const HEADERS_AUTHENTICATORS = [];
+
+export const JWT_TOKEN_PREFIX = 'ey';
 // This method can only be used in createAuthenticatedContext
 // If you need to check auth and create context, use directly createAuthenticatedContext method
 export const authenticateUserFromRequest = async (context, req) => {
```
xfournet, line 1833: i understand why the token last usage doesn't seems to work : when i tested it i click on the playground button that appears on the token screen. This open the playground in the same browser so it's using the user session whatever the authorization header i put on it.
It's quite confusing, and i wonder if we should change the logic of this method to take authorization bearer first ?

opencti-platform/opencti-graphql/src/database/redis/token_usage.ts
```diff
@@ -0,0 +1,54 @@
...
+
+export type TokenUsage = { user: string; token: string; last_check: string };
+
+// L1 Cache to avoid hitting Redis for every request
+const l1Cache = new LRUCache<string, DateTime>({
```
xfournet: wouldn't be possible to put the information directly on user cache to avoid to create another separate cache which is handled quite differently ?

opencti-platform/opencti-front/lang/front/fr.json
```diff
@@ -281,6 +284,8 @@
   "And many more features...": "Et bien d'autres fonctionnalités...",
   "API access": "Accès à l'API",
   "API key": "Clé d'API",
+  "API Token management has been moved to the \"Tokens\" tab.": "La gestion des jetons API a été déplacée vers l'onglet \"Jetons\".",
```
xfournet: don't see where is it used ?

xfournet (approved):

Issue #14345, opened 2026-02-04
Bulk edit in Draft view updates unintended entities


Lhorus6:
## Description

The bulk updates in List view in Drafts overview can trigger updates on entities that should not be affected. I.e. using "Select All" within a Draft may apply changes far beyond the entities actually present in that Draft.

## Environment

OCTI 6.9.13

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Create a new Draft.
2. Inside the Draft, create a new Report.
3. Check the "Select All" box
4. Once selected, click on the bulk update button (the pen)
5. Change the Author field to a specific organization (e.g. "Replace > Author > Filigran")
-> The bulk edit updates the Author field on a large number of unrelated entities across the platform:

<img width="3108" height="1078" alt="Image" src="https://github.com/user-attachments/assets/1f0d058d-e0a4-4a48-9332-fd5f064e1e50" />

## Expected Output

Only the entities contained in the Draft (in this case, the newly created Report) should be updated.

## Actual Output

Despite the Draft containing only a single Report, the bulk edit updates the Author field on a large number of unrelated entities across the platform (see screenshot above)

tobybutler405:
Reproduced on 6.9.15 

When reproduced all reports on the system were updated. This can be seen in the screenshot of the task executed. We have not tested this on other entity types. 

<img width="987" height="293" alt="Image" src="https://github.com/user-attachments/assets/255f3e6d-ee3e-40ec-a9c3-b0751b981a15" />

PR #14371, opened 2026-02-05
[frontend] Context filters for draft entities background tasks (#14345)

Archidoit:
### Proposed changes
- Add context filters in draft (for draft entities, relationships and sightings) for background tasks on the entities in the draft
- unit test for useBuildEntityTypeBasedFilterContext

### Related issues
#14345

### DIFF
--- a/opencti-platform/opencti-front/src/utils/filters/filtersUtils.tsx
+++ b/opencti-platform/opencti-front/src/utils/filters/filtersUtils.tsx
@@ -61,6 +61,8 @@ const NOT_CLEANABLE_FILTER_KEYS = [
   'entity_id',
   'ids',
   'bulkSearchKeywords',
+  'draft_ids',
+  'draft_change',
   PIR_SCORE_FILTER,
   LAST_PIR_SCORE_DATE_FILTER,
 ];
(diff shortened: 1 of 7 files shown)


opencti-platform/opencti-front/src/utils/filters/filtersUtils.tsx
```diff
@@ -888,28 +888,46 @@ export const removeIdAndIncorrectKeysFromFilterGroupObject = (filters: FilterGro
...
+}
+
 export const useBuildEntityTypeBasedFilterContext = (
   entityTypeParam: string | string[],
   filters: FilterGroup | undefined,
-  excludedEntityTypeParam?: string | string[] | undefined,
-  entityTypesContext?: string[],
+  args: BuildEntityTypeBasedFilterContextArgs = {},
 ): FilterGroup => {
+  const { excludedEntityTypesParam = null, entityTypesContext = null, draftId = null } = args;
```
lndrtrbn: You set the default value as null but null is not an acceptable value in the interface defined above. Prefer either using undefined or accept null in the interface

Archidoit: fixed :)

lndrtrbn (approved):

PR #14377, opened 2026-02-06
test(frontend): draft background tasks e2e tests (#14345)

Archidoit:
### Proposed changes
Add e2e tests to check a background task on draft entities only target the entities that are in the draft

### Related issues
#14345


Archidoit:
old PR , not planned

PR #14127, opened 2026-01-20
[backend/frontend] add DraftWorkspace type to entity customization (#14046)

delemaf:
### Proposed changes

* add DraftWorkspace type to entity customization (with attribute AM and creators)
* FF on DraftWorkspace
* refacto SubType

### Related issues

* #14046 

### Checklist

- [x] I consider the submitted work as finished
- [x] I tested the code for its functionality
- [ ] I wrote test cases for the relevant uses case (coverage and e2e)
- [ ] I added/update the relevant documentation (either on github or on notion)
- [x] Where necessary I refactored code to improve the overall quality

### Further comments

Fix ES on change branch:
```
POST /opencti_internal_objects-000001/_delete_by_query
{
  "query": {
    "bool": {
      "must": [
        { "term": { "target_type.keyword": "DraftWorkspace" } }
      ]
    }
  }
}
```

Issue #14046, opened 2026-01-16
feat(workflow): add draft to the list of entities that can be customized


nino-filigran:
## Use case

Workflow will be defined per entities.
Given workflow will first be introduced on draft, we need to have within our GUI, a place to manage draft as an entity, to be able later, to define a workflow. 

## Proposed Solution

Add draft in list of entities that can be customized.
At first, the UI will be a bit a specific given that less options applies to draft (enforce reference, mandatory fields...).

## If the feature request is approved, would you be willing to submit a PR?

Yes / No (Help can be provided if you need assistance submitting a PR)

PR #14412, opened 2026-02-10
[frontend] add missing FF handling in DraftCreation (#14046)

JeremyCloarec:
### Proposed changes

* add missing FF handling in DraftCreation, caused by #14127
*

### Related issues

* #14046
*

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/utils/hooks/useEntitySettings.tsx
+++ b/opencti-platform/opencti-front/src/utils/hooks/useEntitySettings.tsx
@@ -8,6 +8,7 @@ import {
 import { entitySettingFragment } from '../../private/components/settings/sub_types/entity_setting/EntitySettingSettings';
 import useAuth from './useAuth';
 import { useFormatter } from '../../components/i18n';
+import useHelper from './useHelper';
 
 export type EntitySetting = EntitySettingSettings_entitySetting$data;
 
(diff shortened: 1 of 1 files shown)


opencti-platform/opencti-front/src/private/components/drafts/DraftCreation.tsx
```diff
@@ -62,10 +63,15 @@ interface DraftAddInput {
...
   const { t_i18n } = useFormatter();
   const { me: owner, settings } = useAuth();
-  const { mandatoryAttributes } = useIsMandatoryAttribute(DRAFTWORKPACE_TYPE);
   const showAllMembersLine = !settings.platform_organization?.id;
   const canEditAuthorizedMembers = useGranted([KNOWLEDGE_KNUPDATE_KNMANAGEAUTHMEMBERS]);
 
+  const { isFeatureEnable } = useHelper();
+  const isDraftWorkflowFeatureEnabled = isFeatureEnable('DRAFT_WORKFLOW');
+  let mandatoryAttributes = ['name'];
+  if (isDraftWorkflowFeatureEnabled) {
```
esrevi: unfortunately, having useIsMandatoryAttribute in a condition breaks the rules of hooks of react (https://react.dev/reference/rules/rules-of-hooks#only-call-hooks-at-the-top-level) 

one solution could be to to conditionally use the result like that I think:

```
const mandatoryAttributes = isDraftWorkflowFeatureEnabled 
  ? draftMandatoryAttributes 
  : ['name'];
```

opencti-platform/opencti-front/src/utils/hooks/useEntitySettings.tsx
```diff
@@ -45,8 +46,13 @@ export const useIsEnforceReference = (id: string): boolean => {
...
 
 export const useIsMandatoryAttribute = (id: string) => {
+  const { isFeatureEnable } = useHelper();
+  const isDraftWorkflowFeatureEnabled = isFeatureEnable('DRAFT_WORKFLOW');
   const entitySettings = useEntitySettings(id).at(0);
   if (!entitySettings) {
+    if (id === 'DraftWorkspace' && !isDraftWorkflowFeatureEnabled) {
+      return { entitySettings, mandatoryAttributes: ['name'] };
+    }
     throw Error(`Invalid type for setting: ${id}`);
```
esrevi, line 56: just noticed that, should it be ```throw new Error()```instead ?

JeremyCloarec: not sure, we have other locations where we also throw Error instead of throw new Error in our frontend codebase

esrevi: it seems both could be accepted :)

esrevi (approved):

Issue #13264, opened 2025-11-21
Allow "Reduce Knowledge" component to filter on any entities in the bundle, not just the triggering entities


romain-filigran:
## Use case

The "Reduce Knowledge" component currently only works if the filtering result matches the entity that triggered the playbook. This creates a significant limitation that prevents useful playbook implementations.

The reduce step fails (takes "unmatched" route) whenever the filtered entities don't include the original triggering entity, even if the bundle contains valid entities matching the filter criteria.

## Current Workaround

N/A

## Proposed Solution

The "Reduce Knowledge" component should be able to filter and reduce on any entities present in the bundle, regardless of which entity triggered the playbook.

tobybutler405:
Reviewed and tested in Product - working as expected.

PR #14421, opened 2026-02-10
[backend] Allow "Reduce Knowledge" component to filter on any entities in the bundle, not just the triggering entities (#13264)

SarahBocognano:
### Proposed changes

* Rollback changes
* Fix the old code as it was not working and filtering properly
* Add tests for reduce knowledge playbook

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/13264

### DIFF
--- a/opencti-platform/opencti-graphql/src/modules/playbook/playbook-components.ts
+++ b/opencti-platform/opencti-graphql/src/modules/playbook/playbook-components.ts
@@ -289,20 +289,19 @@ export const PLAYBOOK_REDUCING_COMPONENT: PlaybookComponent<ReduceConfiguration>
     for (let index = 0; index < bundle.objects.length; index += 1) {
       const bundleElement = bundle.objects[index];
       const isMatch = await isStixMatchFilterGroup(context, SYSTEM_USER, bundleElement, jsonFilters);
-      if (isMatch) {
-        matchedElements.push(bundleElement);
-      } else {
+      if (!isMatch && bundleElement.id !== baseData.id) {
         unmatchedElements.push(bundleElement);
+      } else if (isMatch) {
+        matchedElements.push(bundleElement);
       }
     }
     if (matchedElements.length === 0) {
-      // Add the unmatched objects to another bundle to be able to work with them as well if needed.
-      const unmatchedObjects = uniqBy(prop('id'), [baseData, ...unmatchedElements]);
-      return { output_port: 'unmatch', bundle: mergeDeepRight(bundle, { objects: unmatchedObjects }) };
+      // Only unmatched elements, base element is not included to avoid side effects on other branches
+      const unmatchedObjects = R.uniqBy(R.prop('id'), [...unmatchedElements]);
+      return { output_port: 'unmatch', bundle: { ...bundle, objects: unmatchedObjects } };
     }
-    const mergedObjects = uniqBy(prop('id'), [baseData, ...matchedElements]);
-    const newBundle = mergeDeepRight(bundle, { objects: mergedObjects });
-    return { output_port: 'out', bundle: newBundle };
+    const mergedObjects = R.uniqBy(R.prop('id'), [baseData, ...matchedElements]);
+    return { output_port: 'out', bundle: { ...bundle, objects: mergedObjects } };
   },
 };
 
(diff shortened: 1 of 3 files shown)


MTorbay-Filigran (approved):

opencti-platform/opencti-graphql/tests/03-integration/01-database/playbookComponents/reduce-knowledge-component-test.ts
```diff
@@ -0,0 +1,87 @@
...
+  });
+
+  it('should remove objects that do not match the filter', async () => {
+    const report = createObject(REPORT_ID, 'report');
+    const campaign = createObject(CAMPAIGN_ID, 'campaign');
+    const bundle = createBundle(report, campaign);
+
+    vi.spyOn(filterUtils, 'isStixMatchFilterGroup')
+      .mockResolvedValueOnce(false) // report (base element, always kept regardless)
+      .mockResolvedValueOnce(false); // campaign does not match
```
lndrtrbn: I understand that if some entities matches we also keep the main entity in the bundle, but if nothing matches it feels a bit odd that we still return the main entity, if nothing matches, should the playbook not stopped?

SarahBocognano: There is voluntary no "out" in this component to allow the playbook to continue running. BUT as I'm not sure this is what we want, it's a good remark. I'm pinging product right now.
What do you think @tobybutler405 ? ^^

tobybutler405: Discussion 26 Feb at sync: If nothing matches then the unmatch should leave the bundle untouched (and playbook author can choose to leave this branch unused)

opencti-platform/opencti-graphql/src/modules/playbook/playbook-components.ts
```diff
@@ -331,17 +331,16 @@ const PLAYBOOK_REDUCING_COMPONENT: PlaybookComponent<ReduceConfiguration> = {
...
     const matchedElements = [baseData];
     for (let index = 0; index < bundle.objects.length; index += 1) {
       const bundleElement = bundle.objects[index];
       const isMatch = await isStixMatchFilterGroup(context, SYSTEM_USER, bundleElement, jsonFilters);
       if (isMatch && baseData.id !== bundleElement.id) matchedElements.push(bundleElement);
     }
     const newBundle = { ...bundle, objects: matchedElements };
+    if (matchedElements.length === 1) {
+      return { output_port: 'unmatch', bundle: newBundle };
+    }
```
lndrtrbn, line 297: That means you will go to `unmatch` if only the main entity matches and that's not what we want.

I think you need to have
```
const matchedElements = [];
```
And remove the condition
```
baseData.id !== bundleElement.id
```
And then
```
if (matchedElements.length === 0) {
      return { output_port: 'unmatch', bundle: newBundle };
}
```

opencti-platform/opencti-graphql/tests/03-integration/01-database/playbookComponents/reduce-knowledge-component-test.ts
```diff
@@ -0,0 +1,87 @@
...
+
+    const result = await callExecutor(bundle);
+
+    expect(result.output_port).toBe('unmatch');
+    expect(result.bundle.objects).toHaveLength(1);
+    expect(result.bundle.objects.find((o) => o.id === REPORT_ID)).toBeDefined();
+    expect(result.bundle.objects.find((o) => o.id === CAMPAIGN_ID)).toBeUndefined();
```
lndrtrbn: Also add a test where only the main matches the filter and verify we have `out` port

opencti-platform/opencti-graphql/src/modules/playbook/playbook-components.ts
SarahBocognano: Wait. The last thing I remember was that we wanted to do nothing if it only matches the main entity.

lndrtrbn (approved):

PR #14430, opened 2026-02-11
[frontend] Fix global kill chain icons display (#13303)

Archidoit:
### Proposed changes
In an entity > Knowledge tab > Overview (right panel) > Global kill chain (at the bottom) : fix UI icons margins, alignement and size : 
- markings alignement when name is long
- years Chip height
- margins between chips

Remove unused StixDomainObjectVictimologyRegions files

## Before
image

## After

image

### Related issues
#13303


opencti-platform/opencti-front/src/components/ItemYears.tsx
```diff
@@ -8,15 +7,9 @@ interface ItemYearsProps {
...
-  const chipStyle: CSSProperties = {
-    fontSize: 12,
-    lineHeight: '12px',
-    height: 20,
-    marginRight: 8,
-  };
   return (
     <Tag
-      sx={chipStyle}
+      sx={{ marginRight: 8 }}
```
esrevi: because this is an atomic component, this one should not have layout props such as margin right :) 

instead, handle the gap in the component that uses ItemYears

Archidoit: fixed :)

esrevi (approved):

PR #14432, opened 2026-02-11
[frontend] multiple ui fixes (#13303)

esrevi:
### Proposed changes

* Fix Gradient background when scrolling with fix background attachment
* Update external reference details layout
* Use logo filigran with good resolution to avoid blur issue on Safari in light theme

### Related issues

* #13303 

### Checklist


opencti-platform/opencti-front/src/private/components/nav/LeftBar.jsx
```diff
@@ -831,7 +830,7 @@ const LeftBarComponent = ({ queryRef }) => {
               }
               <img
                 alt="logo"
-                src={isLightTheme ? logoFiligranLight : logoFiligranDark}
+                src={logoFiligranDark}
```
fellowseb: (nit)
Consider renaming the file to remove the "dark" notion given it's now theme-agnostic. (And renaming the variable too)
Also a screenshot in the PR desc can help reviewers see the visual diff (:.

opencti-platform/opencti-front/src/private/components/analyses/external_references/ExternalReferenceDetails.tsx
```diff
@@ -48,41 +47,49 @@ const ExternalReferenceDetailsComponent = ({
...
-          </Grid>
-        </Grid>
+      <Card
+        title={t_i18n('Details')}
+        sx={{
+          display: 'flex',
+          flexDirection: 'column',
+          gap: 2,
+        }}
+      >
```
fellowseb, line 54: (context/knowledge sharing, no changes requested/needed)

I don't know anything about this component but this seems like a "major" change to its layout no ? Do you think it would be a good case for creating a new variant ?

fellowseb (approved):

opencti-platform/opencti-front/src/private/components/analyses/external_references/ExternalReferenceDetails.tsx
esrevi: actually, the Card accepts any component to be rendered inside, however, to handle the layout inside, I add these style props. 
maybe later, we could add some layout components, to make it more homogenous and easy to use ! :)

opencti-platform/opencti-front/src/private/components/nav/LeftBar.jsx
esrevi: yes indeed :p 
about screenshots, it was trivial here but i understand

PR #14434, opened 2026-02-11
[frontend] Filter icon button styles (#13303)

Archidoit:
### Proposed changes
- Refacto of the styling system of filter icon button to give more clarity and readability (rename styleNumber with functional string variant, turn classes into style and sx, remove duplicate styles)
- Fix the size of filters in Pir Details overview, to be consistant with Tag

### Related issues
#13303


opencti-platform/opencti-front/src/components/FilterIconButtonContainer.tsx
```diff
@@ -20,99 +18,20 @@ import { FilterChipPopover, FilterChipsParameter } from './filters/FilterChipPop
...
-    display: 'flex',
-    alignItems: 'center',
-    gap: 4,
-  },
-}));
+export type FilterIconButtonVariant
+  = undefined // default style (variant is undefined) is the one for filters applied in datatables
+    | 'inLine' // for filters in a datatable line
+    | 'inForm' // for filters in a form
+    | 'TagLike'; // for filters with the Tag style
```
lndrtrbn: Use same case sensitive for all.

I would prefer names that represent more the UI diffs than where they are used right now, we could need to use for example `inLine` in an other context than a datatable line and in that case the name is maybe not representative. If the diff is that they are smaller then the variant can be called `small`

lndrtrbn: And you can call it `variant` I think, shorter and more consistent with the rest of components

Archidoit: done

lndrtrbn (approved):

Issue #14443, opened 2026-02-12
Override of confidence level spacing issue


frapuks:
## Description

spacing issue

## Environment

v7

## Reproducible Steps

1. Go into a user
2. Click on update
3. Click on confidence
4. Click on add a specific max conf level for an entity type

## Expected Output

<img width="808" height="503" alt="Image" src="https://github.com/user-attachments/assets/e85c400d-2782-4392-9ba3-b9ff476b0fff" />

## Actual Output

<img width="806" height="457" alt="Image" src="https://github.com/user-attachments/assets/7d32adf0-6698-421c-a6cb-0d4575944ac6" />

frapuks:
issue already reported in notion page, opened for git branch and PR.

tobybutler405:
moved from triage

PR #14445, opened 2026-02-12
[SEP-19] spacing under confidence level title (#14443)

frapuks:
### Proposed changes

* use and adapt the `Label` component

### Related issues

* #14443 

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/components/common/label/Label.tsx
+++ b/opencti-platform/opencti-front/src/components/common/label/Label.tsx
@@ -6,12 +6,14 @@ import { PropsWithSx } from '../../../utils/props';
 
 interface LabelProps extends PropsWithChildren, PropsWithSx {
   action?: ReactNode;
+  sxTitle?: SxProps;
 }
 
 const Label = ({
   children,
   action,
   sx,
+  sxTitle,
 }: LabelProps) => {
   const theme = useTheme<Theme>();
 
@@ -29,6 +31,7 @@ const Label = ({
     fontSize: '12px',
     fontWeight: 400,
     fontFamily: '"IBM Plex Sans", sans-serif',
+    ...sxTitle,
   };
 
   return (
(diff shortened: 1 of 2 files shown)


opencti-platform/opencti-front/src/private/components/settings/users/edition/UserEditionConfidence.tsx
```diff
@@ -182,18 +185,23 @@ const UserEditionConfidence: FunctionComponent<UserEditionConfidenceProps> = ({
...
-                    {t_i18n('Add a specific max confidence level for an entity type')}
-                  </Typography>
-                  <IconButton
-                    color="primary"
-                    aria-label="Add"
-                    onClick={() => arrayHelpers.push({ entity_type: '', max_confidence: defaultOverrideConfidence })}
-                    style={{ marginTop: '5px' }}
-                    disabled={values.overrides.some((o) => o.entity_type === '')}
+                  <Label
+                    sx={{ marginTop: 2, marginBottom: 2 }}
```
esrevi: I think it is better to handle the layout at the from the parent container which is Form so you could remove sx here and in : 
```
<Form style={{ display: 'flex', flexDirection: 'column', gap: theme.spacing(2) }}>
```

opencti-platform/opencti-front/src/components/common/label/Label.tsx
```diff
@@ -6,12 +6,14 @@ import { PropsWithSx } from '../../../utils/props';
...
 interface LabelProps extends PropsWithChildren, PropsWithSx {
   action?: ReactNode;
+  sxTitle?: SxProps;
 }
 
 const Label = ({
   children,
   action,
   sx,
+  sxTitle,
```
esrevi: I think we should not expose low level style to prevent giving too much freedom and be inconsistent later, i'd suggest to remove sxTitle and add a variant for example :
```
interface LabelProps {
  variant: 'default' | 'drawer' 
}

const Label = ({
  children,
  action,
  sx,
variant = 'default'
}} => {
... 

  const titleStyle = variant === 'card'
    ? { color: theme.palette.text.light, textTransform: 'capitalize' }
    : { color: theme.palette.text.primary, textTransform: 'none' };

  const titleSx: SxProps = {
    marginBottom: 0,
    lineHeight: '21px',
    fontSize: '12px',
    fontWeight: 400,
    fontFamily: '"IBM Plex Sans", sans-serif',
    ...titleStyle,
  };
...
}

```

esrevi (approved):

PR #14453, opened 2026-02-12
[backend] RelationToRule inference from/to types of relationships checking (#11824)

Archidoit:
### Proposed changes
- RelationParticipateToRule should not create participate-to relationships between a user and a sector if an organization is part of a sector. This rule should only be applied for organizations being part of organizations, and create participate-to relationships between a user and an organization.

- buildRelationToRelationRule function should eventually take into account source/target type of right/left relationships implied in the rule definition to ensure a complete type checking if the right and left relationship types are not enough to fully identify the concerned relations

### Related issues
#11824

### DIFF
--- a/opencti-platform/opencti-graphql/src/rules/relationToRelationBuilder.ts
+++ b/opencti-platform/opencti-graphql/src/rules/relationToRelationBuilder.ts
@@ -51,12 +51,11 @@ const buildRelationToRelationRule = (ruleDefinition: RuleDefinition, relationTyp
           await createInferredRelation(context, input, ruleContent);
         }
       };
-      const listFromArgs = {
+      await fullRelationsList(context, RULE_MANAGER_USER, leftType, {
         toId: sourceRef,
         fromTypes: leftTypesFrom,
         callback: listFromCallback,
-      };
-      await fullRelationsList(context, RULE_MANAGER_USER, leftType, listFromArgs);
+      });
     }
     // Need to discover on the from and the to if relation also exists
     // (A) -> leftType -> (B)
@@ -85,12 +84,11 @@ const buildRelationToRelationRule = (ruleDefinition: RuleDefinition, relationTyp
           await createInferredRelation(context, input, ruleContent);
         }
       };
-      const listToArgs = {
+      await fullRelationsList(context, RULE_MANAGER_USER, rightType, {
         fromId: targetRef,
         toTypes: rightTypesTo,
         callback: listToCallback,
-      };
-      await fullRelationsList(context, RULE_MANAGER_USER, rightType, listToArgs);
+      });
     }
   };
   // Contract
--- a/opencti-platform/opencti-graphql/src/rules/relationToRelationBuilder.ts
+++ b/opencti-platform/opencti-graphql/src/rules/relationToRelationBuilder.ts
@@ -1,4 +1,3 @@
-/* eslint-disable camelcase */
 import { buildPeriodFromDates, computeRangeIntersection } from '../utils/format';
 import { createInferredRelation, deleteInferredRuleElement } from '../database/middleware';
 import { createRuleContent } from './rules-utils';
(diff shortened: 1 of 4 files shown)


opencti-platform/opencti-graphql/tests/20-rules/relationToRelationBuilder-test.ts
```diff
@@ -0,0 +1,204 @@
...
+    vi.spyOn(middlewareLoader, 'fullRelationsList').mockImplementation(async (_ctx, _user, _type, args) => {
+      if (args && typeof args.callback === 'function') {
+        await args.callback([
+          {
+            internal_id: existingRelationId,
+            confidence: 50,
+            fromId: existingRelationSourceId,
+            start_time: existingRelationStartTime,
+            stop_time: existingRelationStopTime,
+          } as BasicStoreRelation,
```
fellowseb, line 77: Can we use the `satisfies` operator instead to make sure we don't forget mandatory fields added in the future ?

fellowseb: (Same for two other cases below)

opencti-platform/opencti-graphql/src/rules/relationToRelationBuilder.ts
```diff
@@ -82,7 +85,11 @@ const buildRelationToRelationRule = (ruleDefinition: RuleDefinition, relationTyp
           await createInferredRelation(context, input, ruleContent);
         }
       };
-      const listToArgs = { fromId: targetRef, callback: listToCallback };
+      const listToArgs = {
+        fromId: targetRef,
+        toTypes: rightTypesTo,
+        callback: listToCallback,
+      };
       await fullRelationsList(context, RULE_MANAGER_USER, rightType, listToArgs);
```
fellowseb: Consider removing the temporary `listToArgs` variable and using a literal object instead; that would catch props in excess/typos.

opencti-platform/opencti-graphql/tests/20-rules/relationToRelationBuilder-test.ts
```diff
@@ -0,0 +1,204 @@
...
+  const createdRelationId = uuid();
+  const createdRelationSourceId = uuid();
+  const createdRelationTargetId = uuid();
+  const existingRelationId = uuid();
+  const createdRelationStartTime = new Date('2020-01-01').toISOString();
+  const createdRelationStopTime = new Date('2020-12-31').toISOString();
+  const existingRelationStartTime = new Date('2019-01-01');
+  const existingRelationStopTime = new Date('2019-12-31');
+
+  it('should handle creation of a relationship matching the right side of the rule definition and call createInferredRelation', async () => {
```
fellowseb, line 57: (observation, not action required)

Although TBH I don't grasp 100% of what's going on in this part of the code yet, I feel like the PR description was clear (although requires mental gymnastics hehe!), and I'm surprised about not finding tests that match the cases you mentioned there. Does it make sense ? (: I'm mainly talking about the test descriptions, that feel more technical than business/use-case oriented. Maybe it's because those unit tests are not at the same "level" as the remarks in the PR desc ?

Feel free to ignore this (we can take it offline it easier x))

opencti-platform/opencti-graphql/tests/20-rules/relationToRelationBuilder-test.ts
Archidoit: using 'satisfies' instead of 'as' force to fill lots of mandatory fields useless in the test, this will overload the test file... 
image

opencti-platform/opencti-graphql/tests/20-rules/relationToRelationBuilder-test.ts
Archidoit: yes i only wrote a unit test for this function since this function is quite technical (and more since my last modifications) and was not unit tested at all. The new test I worte protects at least the modifications I did. The whole rule scenario is already tested for some of the inference rules (in the 'rule' folder), but not all the rules. I can write a full rule test for the concerned rule (ParticipateToPartRule) if you think it's useful.

opencti-platform/opencti-graphql/src/rules/relationToRelationBuilder.ts
Archidoit: done :)

opencti-platform/opencti-graphql/tests/20-rules/relationToRelationBuilder-test.ts
fellowseb: Plus it already handles non-matching excessive fields so `satisfies Partial<>` wouldn't help. You're right 👍 .

fellowseb (approved):

opencti-platform/opencti-graphql/tests/20-rules/relationToRelationBuilder-test.ts
fellowseb: Oh ok. Up to your judgement (: for the other tests.

PR #14454, opened 2026-02-12
[frontend] SSO GUI - UI improvement part 2

SarahBocognano:
### Proposed changes

* In form, add key between bracket HIGH (goal: have the technical name displayed next to the label so that user understand what’s being edited)
* 16PX between each line of the form

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/3923


opencti-platform/opencti-front/src/private/components/settings/sso_definitions/LDAPConfig.tsx
esrevi: I think it would be better not to set the margin on each field, instead, you could replace the fragment with a Stack i think , maybe it will need to add width 100% to the stack i don't know
```
<Stack gap={2}> 
       <Field
        component={TextField}
        variant="standard"
        name="issuer"
        label={t_i18n('SAML Entity ID/Issuer')}
        onSubmit={updateField}
        required
        fullWidth
/>
  <Field ... />
</Stack>
```

Main advantage is you let the wrapper handle the layout,no need to worry about which item gets a margin and which doesn't

This could apply to all the different forms :)

opencti-platform/opencti-front/src/private/components/settings/sso_definitions/SSODefinitionForm.tsx
```diff
@@ -394,7 +393,7 @@ const SSODefinitionForm = ({
                   onSubmit={updateField}
                   label={t_i18n('Login Button Name')}
                   fullWidth
-                  style={{ marginTop: 10 }}
+                  style={{ marginBottom: 16 }}
                 />
               )}
               {selectedStrategy === 'SAML' && <SAMLConfig updateField={updateField} />}
```
CelineSebe: Image

This changes are missing, I think

SarahBocognano: Done !

CelineSebe: some translations missing:
-groupSearchBase,
-groupSearchFilter, 
-allow_self_signed, 
-logout_remote

xfournet:
Abandonned

PR #14464, opened 2026-02-13
[frontend] add category in context filters in Vocabularies (#14451)

### LINKED ISSUE #14451: Bulk delete removes all vocabularies
## Description
When attempting to delete items from a specific vocabulary category in the settings, the deletion action incorrectly affects all vocabularies instead of just the selected one.

## Environment
6.9.15

## Reproducible Steps

1: Go to settings -> taxonomies -> vocabularies.
2: Select any vocabulary category from the list (e.g., account_type_ov).
3: Click the "select all" option under this vocabulary.
4: Click the delete icon.

Image

_Bulk deletion performed on all items under account_type_ov_

Image

_Task of bulk delete operation. Note the number of entities affected by the operation, 370_

Image

_All items under attack_motivation_ov are deleted_

## Expected Output
Only the items within the selected vocabulary category (e.g., account_type_ov) should be deleted.

## Actual Output
All vocabularies are deleted, including those not under the selected category. This behavior is consistent across all vocabularies.

JeremyCloarec:
### Proposed changes

*  add category in context filters in Vocabularies 
*

### Related issues

* fix #14451
*

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/settings/Vocabularies.tsx
+++ b/opencti-platform/opencti-front/src/private/components/settings/Vocabularies.tsx
@@ -12,12 +12,13 @@ import { useFormatter } from '../../../components/i18n';
 import LabelsVocabulariesMenu from './LabelsVocabulariesMenu';
 import VocabularyCreation from './attributes/VocabularyCreation';
 import useVocabularyCategory, { vocabFragment } from '../../../utils/hooks/useVocabularyCategory';
-import { addFilter, emptyFilterGroup, useBuildEntityTypeBasedFilterContext } from '../../../utils/filters/filtersUtils';
+import { emptyFilterGroup, isFilterGroupNotEmpty } from '../../../utils/filters/filtersUtils';
 import Breadcrumbs from '../../../components/Breadcrumbs';
 import useConnectedDocumentModifier from '../../../utils/hooks/useConnectedDocumentModifier';
 import { DataTableProps } from '../../../components/dataGrid/dataTableTypes';
 import { UsePreloadedPaginationFragment } from '../../../utils/hooks/usePreloadedPaginationFragment';
 import DataTable from '../../../components/dataGrid/DataTable';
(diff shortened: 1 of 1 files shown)


opencti-platform/opencti-front/src/private/components/settings/Vocabularies.tsx
```diff
@@ -104,7 +104,8 @@ const Vocabularies = () => {
   );
 
   const { filters } = viewStorage;
-  const contextFilters = useBuildEntityTypeBasedFilterContext('Vocabulary', filters);
+  const baseContextFIlters = useBuildEntityTypeBasedFilterContext('Vocabulary', filters);
+  const contextFilters = addFilter(baseContextFIlters, 'category', category, 'eq', 'and');
```
fellowseb: So I'm trying to understand the code and apply the fix locally. I see this results in the display of a `filters are not fully displayed` message in the UI:

Image

with the following filter object:

```
The complete Filter object is as follows:
{
  "mode": "and",
  "filters": [
    {
      "key": "category",
      "values": [
        "account_type_ov"
      ],
      "operator": "eq",
      "mode": "and"
    }
  ],
  "filterGroups": [
    {
      "mode": "and",
      "filters": [
        {
          "key": "entity_type",
          "values": [
            "Vocabulary"
          ],
          "operator": "eq",
          "mode": "or"
        }
      ],
      "filterGroups": []
    }
  ]
}
```

Isn't there a way that would make the filter object simpler (like flattening everything) and thus displayable with current logic ?

JeremyCloarec: I pushed a new version with simplified filter structure

fellowseb: Great !

fellowseb (approved):

Archidoit (approved):

Issue #14354, opened 2026-02-04
Cannot export in PDF Report containing embedded images


lndrtrbn:
## Description

If my report contains embedded images in the description, the export in PDF crashes when using Fintel Templates.

## Environment

1. OS (where OpenCTI server runs): 
2. OpenCTI version: 6.9
3. OpenCTI client: frontend
4. Other environment details:

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Create a report
2. Add an embedded image
3. Try to export in PDF usin default fintel template

## Expected Output

The PDF is generated

## Actual Output

Nothing happens, there is an error in console

romain-filigran:
@natalie-natsu @SarahBocognano : Can we consider this problem to be solved/closed?

SarahBocognano:
Yes we can ! @romain-filigran

PR #14502, opened 2026-02-16
[frontend] Cannot export in PDF Report containing embedded images (#14354)

SarahBocognano:
### Proposed changes

* create a function to handle the embedded urls for pdfMake library

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/14354

### DIFF
--- a/opencti-platform/opencti-front/src/utils/htmlToPdf/htmlToPdf.ts
+++ b/opencti-platform/opencti-front/src/utils/htmlToPdf/htmlToPdf.ts
@@ -73,27 +73,27 @@ const normalizeEntityBaseUrl = (url: string) =>
 const entityBaseUrl = `${window.location.origin}${window.location.pathname}`.replace(/\/$/, '');
 const resolvedEntityBaseUrl = normalizeEntityBaseUrl(entityBaseUrl);
 
-const resolvePdfMakeEmbeddedImages = async (
-  pdfMakeObject: TDocumentDefinitions, resolvedEntityBaseUrl: string,
-) => {
-  if (!pdfMakeObject.images) return;
+export const resolvePdfMakeEmbeddedImages = async (
+  images: TDocumentDefinitions['images'],
+  resolvedEntityBaseUrl: string,
+): Promise<Record<string, string | ImageDefinition>> => {
+  if (!images) return {};
 
-  const refs = pdfMakeObject.images as unknown as Record<string, string>;
-  const resolved: Record<string, string | ImageDefinition> | undefined = { ...refs };
+  const entries = await Promise.all(
+    Object.entries(images).map(async ([key, value]) => {
+      const strValue = typeof value === 'string' ? value : null;
+      if (!strValue?.startsWith('embedded/')) return [key, value] as const;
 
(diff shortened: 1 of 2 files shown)


opencti-platform/opencti-front/src/utils/htmlToPdf/htmlToPdf.ts
```diff
@@ -62,6 +62,40 @@ export const htmlToPdf = (fileName: string, content: string) => {
...
+ */
+const normalizeEntityBaseUrl = (url: string) =>
+  url
+    .replace(/\/content\/?$/, '')
+    .replace(/\/$/, '');
+
+const entityBaseUrl = `${window.location.origin}${window.location.pathname}`.replace(/\/$/, '');
+const resolvedEntityBaseUrl = normalizeEntityBaseUrl(entityBaseUrl);
+
+const resolvePdfMakeEmbeddedImages = async (
```
esrevi: I think it could be good if we add some unit tests for that function :)

opencti-platform/opencti-front/src/utils/htmlToPdf/htmlToPdf.ts
```diff
@@ -62,6 +62,40 @@ export const htmlToPdf = (fileName: string, content: string) => {
...
+  pdfMakeObject: TDocumentDefinitions, resolvedEntityBaseUrl: string,
+) => {
+  if (!pdfMakeObject.images) return;
+
+  const refs = pdfMakeObject.images as unknown as Record<string, string>;
```
esrevi: can we use the type from images instead of casting it ?

opencti-platform/opencti-front/src/utils/htmlToPdf/htmlToPdf.ts
```diff
@@ -62,6 +62,40 @@ export const htmlToPdf = (fileName: string, content: string) => {
...
+
+      const fileName = value.slice('embedded/'.length);
+      const url = `${resolvedEntityBaseUrl}/embedded/${encodeURIComponent(fileName)}`;
+
+      const img = await getBase64ImageFromURL(url);
+      resolved[key] = img.startsWith('data:') ? img : `data:image/png;base64,${img}`;
+    }),
+  );
+
+  pdfMakeObject.images = resolved as never;
```
esrevi: I think it would be better no to mutate the pdfMakeObject here, but instead , return the images object, so it is easier to test, to follow the code and separate the interests.

thus in htmlToPdf.ts, we would have something like that, it is more explicit, we can see where images comes from

```
const resolvedImages = entityBaseUrl
  ? await resolveEmbeddedImages(pdfMakeObject.images, resolvedEntityBaseUrl)
  : pdfMakeObject.images;
  
  ...
  
  const docDefinition: TDocumentDefinitions = {
    pageMargins: [20, 30],
    styles: { /* ... * },
    defaultStyle: { /* */},
    ...pdfMakeObject,
    images: resolvedImages,
    ....

```

esrevi (approved):

natalie-natsu (approved):

PR #14511, opened 2026-02-17
[backend] add ability to configure playbook connectors as Realtime priority instead of default (#14512)

### LINKED ISSUE #14512: Add ability to set playbook connectors as Realtime priority instead of default
## Use case

We sometime want to prioritise consuming playbooks queues over other queues to prevent playbook queues from piling up more than they can be consumed

JeremyCloarec:
### Proposed changes

* add ability to configure playbook connectors as Realtime priority instead of default
* defaults to false (i.e. Default priority), but can be set to true in config (i.e. Realtime priority)

### Related issues

* fix #14512
*

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/config/default.json
+++ b/opencti-platform/opencti-graphql/config/default.json
@@ -277,7 +277,8 @@
     "lock_cron_key": "playbook_cron_manager_lock",
     "interval": 60000,
     "cron_max_size": 500,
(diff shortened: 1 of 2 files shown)


romain-filigran:
@JeremyCloarec : Will this also work for manually triggering the connector (like data enrichment)? In other words, if I manually request data enrichment, will the result be processed first?

PR #14525, opened 2026-02-18
[backend] switch express query parser from default simple to extended (#14517)

### LINKED ISSUE #14517: Taxii collection query param match not taken in count
## Description

When calling the url /collection/aaaa/objects?match[id]=stix-id, it returns a collection of ids instead of the stix object we want.

## Environment

1. OS (where OpenCTI server runs): all
2. OpenCTI version: 6.9.17
3. OpenCTI client: frontend
4. Other environment details:

## Reproducible Steps

1. Create a taxii collection, select filter entity type, leave value empty
2. Click on the taxii element, it opens a new tab with the url /taxii2/root/collections/:id/objects/
3. Now, select one id in the list 
4. enter url /taxii2/root/collections/:id/objects?match[id]=objectId

## Expected Output

It should return the object instead of the collection

## Actual Output

it returns the collection

JeremyCloarec:
### Proposed changes

* to handle object params in req.query for taxii objects endpoint (match param is not currently handled properly because of this), we switch express query parser from default simple to extended
*

### Related issues

* fix #14517 
*

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/http/httpTaxii.js
+++ b/opencti-platform/opencti-graphql/src/http/httpTaxii.js
@@ -97,6 +97,7 @@ const JsonTaxiiMiddleware = express.json({
 });
 
 const initTaxiiApi = (app) => {
+  app.set('query parser', 'extended');
   // Discovery api
   app.get(`${basePath}/taxii2/`, async (req, res) => {
     try {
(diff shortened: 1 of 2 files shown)


esrevi (approved):
tested, ok !

PR #14533, opened 2026-02-19
[drafts] Enable org sharing in drafts (#14056)

### LINKED ISSUE #14056: Worfklow: Ability in Draft to manage org sharing
## Use case

Because workflow will be applied on Draft, it is important, if a user makes a mistake, to be able to manage Org sharing on entities in draft so that user can repair its mistake. 

## Proposed Solution

- Ability, via mass operation, to manage org sharing within a draft
- Ability, in a an entity, to perform org sharing within a draft.

OctaveLaventure:
### Proposed changes

* enable org share/unshare mass operation and single entity operation

### Related issues

* Fixes #14056

### Checklist


opencti-platform/opencti-graphql/tests/03-integration/02-resolvers/draft-organization-sharing-test.ts
```diff
@@ -0,0 +1,266 @@
+import gql from 'graphql-tag';
+import { afterAll, beforeAll, describe, expect, it, vi } from 'vitest';
+import * as ee from '../../../src/enterprise-edition/ee';
+import { adminQuery } from '../../utils/testQuery';
+import { awaitUntilCondition } from '../../utils/testQueryHelper';
+
+describe.skip('Draft organization sharing', () => {
```
OctaveLaventure, line 8: Tests are functional but skipped for now as I couldn't make the enterprise edition mocks to work, apparently because of a hoisting issue between the loadings of vitest and the middleware.js file. The only way I could make it work is with a global file mock for ee.ts, but that's not ideal and would probably break other tests, if you have a better idea please help 🙏🏻

opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectSharingList.tsx
```diff
@@ -39,10 +38,8 @@ interface StixCoreObjectSharingListProps {
 }
 
 const StixCoreObjectSharingList = ({ data, disabled, inContainer, children }: StixCoreObjectSharingListProps) => {
-  const draftContext = useDraftContext();
   const { t_i18n } = useFormatter();
-  const disabledInDraft = !!draftContext;
-  const fullyDisabled = disabled || disabledInDraft;
+  const fullyDisabled = disabled;
```
delemaf: Maybe we can directly pass `disabled`

delemaf (approved):
Testing OK, we might need a FF

delemaf (approved):
🚀

Issue #5220, opened 2023-12-14
Add an operator "Only equals to" in filters


Jipegien:
## Use case

As a user, I want to be able to filter a list of Object for which the technical creator is X and only X.

nino-filigran:
this is a duplicate https://github.com/OpenCTI-Platform/opencti/issues/7049

Lhorus6:
Wouldn't this be easy to implement, given Elasticsearch's search capabilities? Example

```
GET your_index/_search
{
  "query": {
    "bool": {
      "filter": [
        {
          "term": {
            "your_field.keyword": "X"
          }
        },
        {
          "script": {
            "script": {
              "source": """
                doc['your_field.keyword'].length == 1 && doc['your_field.keyword'][0] == 'X'
              """,
              "lang": "painless"
            }
          }
        }
      ]
    }
  }
}
```

Lhorus6:
The opposite would be also needed, i.e. "not only"

The underlying need is: Ability to exclude IOCs shared only by a specific source

I want to share a bundle of IOCs, except those that come only from a specific commercial source, for example CrowdStrike.

If an IOC is provided exclusively by CrowdStrike, it should be excluded from the shared bundle. However, if the same IOC is also shared by another source in addition to CrowdStrike, then it should still be included. 
=> So I would like to have "Creator not only CrowdStrike"

PR #14543, opened 2026-02-20
[frontend/backend] 'Only equal to' new filter operator (#5220)

Archidoit:
### Proposed changes
Add 2 new filter operators: 'only_eq_to' and 'not_only_eq_to'
- Frontend: add the new operators for filters that can have several values
- Backend implementation: handling the new operators for stix filtering and dynamic filtering
- Backend tests
- Documentation

### Related issues
#5220

### Screenshot

image


opencti-platform/opencti-graphql/src/database/engine.ts
```diff
@@ -2560,6 +2563,29 @@ export const buildLocalMustFilter = (validFilter: any) => {
...
+                  source: `
+                    def fieldValues = doc['${buildFieldForScriptQuery(headKey)}'];
+                    if (fieldValues == null || fieldValues.length == 0) return false;
+                    def filterValues = params.values;
+                    if (params.mode == 'and') {
+                      return fieldValues.length == filterValues.length && fieldValues.every(v -> filterValues.contains(v));
+                    } else if (params.mode == 'or') {
+                      return fieldValues.length == 1 && filterValues.contains(fieldValues[0]);
+                    }
+                    return false;
```
esrevi, line 2580: is this intentional ? If we pass an unknown mode, it will fail silently, will return no match, or should we throw an error when there is no mode or another mode that we don't handle?

opencti-platform/opencti-graphql/tests/02-dataInjection/01-dataCount/elasticSearch-test.js
```diff
@@ -451,7 +451,7 @@ describe('Elasticsearch pagination', () => {
     expect(entityTypeMap.get('Vocabulary')).toBe(entitiesCounter.Vocabulary);
     expect(entityTypeMap.get('RetentionRule')).toBe(entitiesCounter.RetentionRule);
     expect(data.edges.length).toEqual(entitiesCounterTotal);
-    const filterBaseTypes = R.uniq(R.map((e) => e.node.base_type, data.edges));
+    const filterBaseTypes = R.uniq(data.edges.map((e) => e.node.base_type));
```
esrevi: I think this should be equivalent if we want to remove r.uniq

const filterBaseTypes = [...new Set(data.edges.map((e) => e.node.base_type))];

opencti-platform/opencti-graphql/tests/02-dataInjection/01-dataCount/elasticSearch-test.js
```diff
@@ -898,7 +898,7 @@ describe('Elasticsearch pagination', () => {
     expect(metaByEntityType['kill-chain-phase'].length).toEqual(relationsCounter['kill-chain-phase']);
     expect(metaByEntityType['operating-system'].length).toEqual(relationsCounter['operating-system']);
     expect(data.edges.length).toEqual(totalCountRelations);
-    let filterBaseTypes = R.uniq(R.map((e) => e.node.base_type, data.edges));
+    let filterBaseTypes = R.uniq(data.edges.map((e) => e.node.base_type));
```
esrevi: same here

opencti-platform/opencti-graphql/tests/02-dataInjection/01-dataCount/elasticSearch-test.js
```diff
@@ -927,7 +927,7 @@ describe('Elasticsearch pagination', () => {
     expect(entityTypeMap.get('operating-system')).toBe(relationsCounter['operating-system']);
     expect(entityTypeMap.get('stix-sighting-relationship')).toBe(relationsCounter['stix-sighting-relationship']);
     expect(data.length).toEqual(totalCountRelations);
-    filterBaseTypes = R.uniq(R.map((e) => e.base_type, data));
+    filterBaseTypes = R.uniq(data.map((e) => e.base_type));
```
esrevi: same here

opencti-platform/opencti-graphql/src/database/engine.ts
Archidoit: by default, mode = 'or' if it's not defined : 
image

and if mode is not equal to 'or' or 'and', the script will return false (no match)

opencti-platform/opencti-graphql/tests/02-dataInjection/01-dataCount/elasticSearch-test.js
Archidoit: indeed... it's a bit heavy and less clear, that's why I didn't replace uniq, but I can also remove uniq if you think it's better

alice-debra (dismissed):
Toby has pointed out a spelling error; it should be either 'only equals' or 'only equal to'.

Archidoit:
> Toby has pointed out a spelling error; it should be either 'only equals' or 'only equal to'.

@alice-debra which one do you prefer ? Maybe 'only equals' is shorter

alice-debra:
> > Toby has pointed out a spelling error; it should be either 'only equals' or 'only equal to'.
> 
> @alice-debra which one do you prefer ? Maybe 'only equals' is shorter

"only equal to" checked with @tobybutler405 :)

opencti-platform/opencti-graphql/tests/02-dataInjection/01-dataCount/elasticSearch-test.js
fellowseb: Readability-wise I think both are similar so I would go with removing the ramda dep here and below . Either way, whatever the choice consider updating the notion page (:, https://www.notion.so/filigran/How-to-remove-Ramda-a0a7ddfbf1b2400c883173b16d2c9fac

fellowseb (approved):
Nice !

fellowseb (approved):

PR #14555, opened 2026-02-22
[client] fix: batch_size race conditions (#15187)

### LINKED ISSUE #15187: [BUG] batch_size not enforced as a cap in BatchCallbackWrapper
### Environment

Latest version

### Reproducible Steps

1. Create a livestream
2. Ingest data
2. Start a batch stream connector with delay

### Actual Outcome

Too much events in the batch. In BatchCallbackWrapper, the batch_size parameter acts as a minimum threshold to trigger processing rather than a strict upper limit on the number of events extracted per batch.

### Expected Outcome

batch_size should be a strict cap: each call to _extract_batch_data should return at most  batch_size events, leaving any overflow in the buffer for subsequent processing

### Add any file, screenshots, video

_No response_

Renizmy:
### Proposed changes

Previously, batch_size act as a minimum threshold to trigger background processing rather than a strict limit. A race condition existed between the thread appending messages from the livestream and the timer thread attempting to acquire the lock to clear the buffer. As a result, the actual number of extracted events could significantly exceed the batch_size. This behavior was typically observed under high ingestion throughput with a low batch_size (< 10), or when fetching historical livestream data

Fix https://github.com/OpenCTI-Platform/opencti/issues/15187

Updated _extract_batch_data to extract at most   batch_size events, leaving any newly arrived overflow messages safely in the buffer (state not updated)

### Checklist

### DIFF
--- a/client-python/pycti/connector/opencti_connector_helper.py
+++ b/client-python/pycti/connector/opencti_connector_helper.py
@@ -1345,16 +1345,26 @@ class BatchCallbackWrapper:
                     if should_process:
                         batch_data = self._extract_batch_data(trigger_reason)
 
-            if batch_data is not None:
+            while batch_data is not None:
                 self._execute_batch_callback(batch_data)
+                batch_data = None
+                if self._stop_event.is_set():
+                    break
+                with self._lock:
+                    if self.batch_size and len(self.batch) >= self.batch_size:
+                        batch_data = self._extract_batch_data("size_limit")
 
         # Process remaining messages after loop exits
         batch_data = None
         with self._lock:
             if len(self.batch) > 0:
                 batch_data = self._extract_batch_data("shutdown")
-        if batch_data is not None:
+        while batch_data is not None:
             self._execute_batch_callback(batch_data)
+            batch_data = None
+            with self._lock:
+                if len(self.batch) > 0:
+                    batch_data = self._extract_batch_data("shutdown")
 
     def __call__(self, msg) -> None:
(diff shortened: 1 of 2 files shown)


nino-filigran:
Thank @Renizmy we're going to have a look at your PR!

JeremyCloarec:
Hi @Renizmy,

Thanks for your contribution! Could you expand a bit on the race condition and how that could happen before? Was it in the cases where the thread appending the messages was so high loeaded that it kept acquiring the lock instead of the time_loop?
In any case, the changes look good 👍 Last remaining thing to do, could you create a related issue and connector the PR to it? And change the name of the PR to be comatible with our PR naming checks, i.e. something like '[client] ..... (#issueNumber)'

Renizmy:
Hi @JeremyCloarec , thanks for the review!

The race condition is not really about lock starvation, both threads acquire the lock fine. The issue is that the original _extract_batch_data always extracted all buffered messages (self.batch.copy()) regardless of batch_size. The batch_size was only used as a threshold to trigger processing (len(self.batch) >= self.batch_size), but once triggered, the entire buffer was drained in one run.
This becomes a problem under "high throughput": between the moment the timer thread releases the lock (after extracting) and the next time it checks again, the __call__ thread can append more messages than batch_size. On the next timer iteration. Especially during fetching historical data, the actual batch delivered to the callback could be larger than expected.                                                                                        

I'll create the issue and update the PR name accordingly.

JeremyCloarec (approved):

PR #12984, opened 2025-10-28
[frontend]missing vocabularies translations(#12799)

CelineSebe:
### Proposed changes

* add description vocabularies translations

### Related issues

* #12799 

### Checklist

- [x] I consider the submitted work as finished
- [x] I tested the code for its functionality
- [ ] I wrote test cases for the relevant uses case (coverage and e2e)
- [ ] I added/update the relevant documentation (either on github or on notion)
- [ ] Where necessary I refactored code to improve the overall quality

Issue #12799, opened 2025-10-14
Missing vocabularies translations in settings


CelineSebe:
## Description

We don't have the translations of the vocabulary descriptions.

## Environment

1.everywhere

## Reproducible Steps

Steps to create the smallest reproducible scenario:

1. Go to Settings > Taxonomie > Vocabularies

<img width="1145" height="357" alt="Image" src="https://github.com/user-attachments/assets/c90619b9-988d-4742-84ef-b244e7e041e6" />

alice-debra:
Hello @CelineSebe,
I just noticed that your PR has been merged, but I still don't see the translations on testing.

CelineSebe:
Hello @alice-debra 
It was fixed, but the latest changes for V7 broke the translations again. I'm reopening a PR.

PR #14607, opened 2026-02-25
[frontend] Fix of vocabulary descriptions(#12799)

CelineSebe:
### Proposed changes

* fix translations after datable refactoring (V7)

### Related issues

* #12984 

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/settings/VocabularyCategories.tsx
+++ b/opencti-platform/opencti-front/src/private/components/settings/VocabularyCategories.tsx
@@ -57,10 +57,15 @@ const VocabularyCategories = () => {
       isSortable: true,
     },
     description: {
-      id: 'description',
+      id: t_i18n('description'),
       percentWidth: 60,
       isSortable: false,
-      render: (data: { category: VocabularyDefinition }) => defaultRender(data.category.description),
+      render: (data: { category: VocabularyDefinition }) => {
+        if (data.category.description) {
+          return defaultRender(t_i18n(`${data.category.description}`));
+        }
+        return '-';
+      },
     },
   };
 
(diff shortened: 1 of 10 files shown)


opencti-platform/opencti-front/src/private/components/settings/VocabularyCategories.tsx
```diff
@@ -57,10 +57,15 @@ const VocabularyCategories = () => {
       isSortable: true,
     },
     description: {
-      id: 'description',
+      id: t_i18n('description'),
```
Archidoit: id should not be translated

Archidoit:
```suggestion
      id: 'description',
```

opencti-platform/opencti-front/src/private/components/settings/VocabularyCategories.tsx
```diff
@@ -57,10 +57,15 @@ const VocabularyCategories = () => {
...
       percentWidth: 60,
       isSortable: false,
-      render: (data: { category: VocabularyDefinition }) => defaultRender(data.category.description),
+      render: (data: { category: VocabularyDefinition }) => {
+        if (data.category.description) {
+          return defaultRender(t_i18n(`${data.category.description}`));
```
Archidoit: defaultRender already handle empty values, don't need to use a 'if'

Archidoit (changes requested):

opencti-platform/opencti-front/src/private/components/settings/VocabularyCategories.tsx
CelineSebe: I get a null value without it

Archidoit: You get null because you put the description to translate in a string, that will be empty if no description. Here is the implementation to avoid the 'if' and the call to EMPTY_VALUE : 

Image

CelineSebe: same results:

Image

Archidoit (approved):

Issue #13115, opened 2025-11-13
Entity restoration should restore former creation dates of relationships


fruitcakej:
## Description
We had another instance of a user mistakenly deleting a sector. We restored it, and it worked. However, all its relationships now have a platform creation date = today.
As we use this date to filter our dashboards, we loose all the temporal information because the relationship creation dates were not preserved! 

Finally, the inferred relationships also have the same problem: As the relationships are inferred after entity restoration, their date is today too.

## Environment
1. SaaS 6.8.10

## Expected Output
Upon restoration the original platform creation date should be restored

## Actual Output
Date changes to the date of restoration

## Additional information
Also, the original 'Creator' of the relationship is lost.

PR #14609, opened 2026-02-25
[backend] Entity restoration should restore former creation dates of relationships (#13115)

SarahBocognano:
### Proposed changes

* As the entity is "re created" when using restore of the trash. I used the restore opts attribute to change the created_at attribute with the created attribute (the created one keeps the plateform creation date) on the data-building of the creation of an entity

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/13115


opencti-platform/opencti-graphql/src/database/data-builder.js
JeremyCloarec: I think you should handle it here instead: instead of always setting today, you could do something like 
`restore ? input.created_at : today`.

fellowseb:
Could we update an integration test (for example `deleteOperation-test.ts`) to make sure there's no regression on this change later ?

JeremyCloarec (approved):

PR #14638, opened 2026-02-26
[frontend] Improve graph labels (#14234)

lndrtrbn:
### Proposed changes

*

### Related issues

* #14234

### Checklist


opencti-platform/opencti-front/src/utils/String.test.ts
```diff
@@ -0,0 +1,87 @@
...
+
+  describe('displayEntityTypeForTranslation', () => {
+    it('should translate an entity type in a translatable string', () => {
+      expect(displayEntityTypeForTranslation(undefined)).toEqual(undefined);
+      expect(displayEntityTypeForTranslation('Malware')).toEqual('entity_Malware');
+      expect(displayEntityTypeForTranslation('targets')).toEqual('relationship_targets');
+    });
+  });
+
+  describe('Function: isHtmlSafe()', () => {
```
fellowseb: nit: typo in the name of the function

fellowseb: This is globally fine but I would consider rather not trying to test `dompurify` 's behavior (which is what you're doing here x)). Because it would take a lot of test cases to be complete (https://github.com/cure53/DOMPurify/blob/main/test/fixtures/expect.mjs) and it's actually not our responsibility. IMO it's one of the rare cases where mocking makes sense to distinguish two cases: purify returning an unchanged `input` string or a `sanitized-input`.
Lemme know if it doesn't make sense !

opencti-platform/opencti-front/src/components/graph/utils/useGraphParser.ts
```diff
@@ -175,7 +175,7 @@ const useGraphParser = () => {
       isObservable: !!data.observable_value,
       numberOfConnectedElement: numberOfConnectedElement ?? data.numberOfConnectedElement,
       ...getNodeImg(data),
-      name: getNodeName(data),
+      name: sanitize(getNodeName(data), true),
       label: getNodeLabel(data),
```
fellowseb, line 179: From what I see this prop too could be potentially an XSS target, could we sanitize it too here ?

fellowseb (approved):
Well done !

opencti-platform/opencti-front/src/components/graph/utils/useGraphParser.ts
lndrtrbn: I have tested and `label` doesn't seem affected, only `name`. You had found the opposite?

fellowseb: Not per se, only a potential case. I think this PR is good as is; I'll create a ticket to track all potential issues in our dependencies and investigate more 👍 .

fellowseb (approved):

PR #14655, opened 2026-02-26
[backend] Fix strict_dynamic_mapping_exception exceptions thrown in fileIndexManager (#14665)

### LINKED ISSUE #14665: Backend error: strict_dynamic_mapping_exception, ([publisher] or [rating] within [attachment]), fileIndexManager usecase
# Symptoms / detection

The backend `strict_dynamic_mapping_exception` error was detected with the following specific fields:

- cause_message: `Update indexing fail`

- cause_attributes_cause_message:
```
strict_dynamic_mapping_exception
	Root causes:
		strict_dynamic_mapping_exception: [1:538] mapping set to strict, dynamic introduction of [publisher] within [attachment] is not allowed
```

- cause_attributes_cause_stack:
```
ResponseError: strict_dynamic_mapping_exception
	Root causes:
		strict_dynamic_mapping_exception: [1:538] mapping set to strict, dynamic introduction of [publisher] within [attachment] is not allowed
    at SniffingTransport._request (/opt/opencti/build/back.js:790:194528)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async /opt/opencti/build/back.js:790:197793
    at async SniffingTransport.request (/opt/opencti/build/back.js:790:197613)
    at async _Client.IndexApi [as index] (/opt/opencti/build/back.js:853:172288)
    at async elIndex (/opt/opencti/build/back.js:1829:21943)
    at async elIndexFiles (/opt/opencti/build/back.js:1878:105502)
    at async indexImportedFiles (/opt/opencti/build/back.js:4285:18107)
    at async fileIndexHandler (/opt/opencti/build/back.js:4285:20286)
    at async /opt/opencti/build/back.js:4285:22052
    at async _SetIntervalAsyncTimer.runHandlerAndScheduleTimeout_fn (/opt/opencti/build/back.js:894:3825)
    at async Timeout._onTimeout (/opt/opencti/build/back.js:894:3612)
```
- cause_message: `Simple indexing fail`

Happened in versions 6.9.15 to 6.9.20.

# Analysis / explanation

The general explanation of the strict_dynamic_mapping_exception error in Elasticsearch is that it occurs when you try to index a document with a new field that doesn't match the existing mapping rules in a strict dynamic mapping context. This exception is thrown to prevent unintended schema changes that could lead to data corruption or inconsistent mappings.

This specific instance of `strict_dynamic_mapping_exception` occurs in the `fileIndexManager` use case. Docs [here](https://docs.opencti.io/latest/deployment/managers/?h=manager#file-index-manager).

More investigation needed to understand the root cause.

# Resolution brainstorm

See PR description.

fellowseb:
### Context

The issue arose because of missing index mappings in the attachment sub-document: we use an Elasticsearch pipeline processor for attachments that extracts fields for us. By default this processor extracts all the fields it can: https://www.elastic.co/guide/en/elasticsearch/reference/8.19/attachment.html#attachment-fields.

The problem is that we've created index mapping sfor only a subset of those fields (see `document.ts`). This added to the fact that we enforce `dynamic: strict` behavior on indices, meaning we don't let unknown fields be pushed on an index, resulted in a few exceptions.

### Proposed changes

- This PR specifies which fields to extract when configuring the `attachment` pipeline in ES/OS: those for which we already have a mapping.
We could consider ingesting the other pieces of data but I'm not sure it's useful and the volume is ultra low for now.

- I added an integration test making sur that a PDF with a metadata (`dc:publisher` or `dc:rating`) that could be extracted by the processor,  but that isn't because we now tell it not to, doesn't fail the indexing.

- I ran the test with ES and OS locally. Running with OS required tweaking the dev setup to install the `ingest-attachment` plugin before starting the OS process which requires a custom Dockerfile (https://docs.opensearch.org/latest/install-and-configure/install-opensearch/docker/#working-with-plugins).

### Related issues

* Fix https://github.com/OpenCTI-Platform/opencti/issues/14665

### Checklist

### Further comments

I had to track down how to name the metadata by looking at the ES code (https://github.com/elastic/elasticsearch/blob/main/modules/ingest-attachment/src/main/java/org/elasticsearch/ingest/attachment/AttachmentProcessor.java#L200) and the library itself uses (Apache TIka). Without the `dc:` prefix it wouldn't be seen by the processor.

I used [https://www.embedpdf.com/tools/pdf-metadata-editor](https://www.embedpdf.com/tools/pdf-metadata-editor) to add metadata to the test file and [tika](https://github.com/apache/tika/?tab=readme-ov-file#quick-start) to read them like the ES dependency.


opencti-platform/opencti-graphql/src/types/type-utils.ts
```diff
@@ -3,3 +3,9 @@ import { isNotEmptyField } from '../database/utils';
 export const filterEmpty = <T>(data: T | null | undefined): data is T => {
   return isNotEmptyField(data);
 };
+
+/**
+ * Inverse operation of the built-in Readonly<T> utility type:
+ * makes all records of an object mutable.
+ */
+export type Mutable<T> = { -readonly [P in keyof T]: T[P]; };
```
fellowseb: Not sure I'm in the right file here. Lemme know if there's a better place.

opencti-platform/opencti-graphql/src/modules/internal/document/document.ts
```diff
@@ -1,10 +1,57 @@
...
+  standardId,
+  updatedAt,
+  type MappingDefinition,
+  type BasicStoreAttribute,
+} from '../../../schema/attribute-definition';
 import { ENTITY_TYPE_MARKING_DEFINITION } from '../../../schema/stixMetaObject';
 import { ABSTRACT_STIX_CORE_OBJECT } from '../../../schema/general';
 import { UPLOAD_STATUS_VALUES } from './document-domain';
 
+export const ATTACHMENT_MAPPINGS = [
```
fellowseb: I moved this literal without changing its contents.

aHenryJard:
Can you please create a public issue in OpenCTI repo and link it https://github.com/OpenCTI-Platform/opencti/issues

fellowseb:
> Can you please create a public issue in OpenCTI repo and link it https://github.com/OpenCTI-Platform/opencti/issues

Done. I 'll make sure the correct issue number is in the commit message when squashing too 👍 .

opencti-platform/opencti-graphql/src/modules/internal/document/document.ts
```diff
@@ -1,10 +1,57 @@
...
+
+// Compile-time shenanigans to make sure we don't forget to update
+// ATTACHMENT_MAPPINGS when/if we start extracting new fields
+// via the ES/OS attachment ingest pipeline.
+type AttachmentMappingsWithCheck = Exclude<
+  AttachmentProcessorExtractedProp,
+  typeof ATTACHMENT_MAPPINGS[number]['name']
+> extends never
+  ? MappingDefinition<BasicStoreAttribute>[]
+  : 'Make sure ATTACHMENT_MAPPINGS defines one mapping for each AttachmentProcessorExtractedProp';
```
SouadHadjiat: I don't understand well this part with this type check, could you explain it please? why is it needed?

fellowseb: So basically the issue with this exception came from not having defined a `mapping` for each field extracted from the document.
The solution I suggest is configuring the attachment ingest thing to extract only the fields we want.

I wanted a way to statically make sure those two lists (the mappings and the ingest process config) remain in sync so I came up with this "hack" that uses an intermediate variable `TYPE_CHECKED_ATTACHMENT_MAPPINGS` on which we assert one of two types:
- either its correct type `MappingDefinition<BasicStoreAttribute>[]`, if no mappings are missing (the `never` case)
- either a string literal actually explaining the problem

While this was fun, it's not maybe the most clear way to do that (:, so I pushed another refactor commit to create an `assertType` utility that's a noop at runtime but is useful for compile-time type strict equality checks. Let me know if that's better ?! With that method I can keep the mappings definition in place.

opencti-platform/opencti-graphql/src/types/type-utils.ts
fellowseb: Moved to another file.

opencti-platform/opencti-graphql/src/modules/internal/document/document.ts
fellowseb: New method reverted this.

opencti-platform/opencti-graphql/src/modules/internal/document/document.ts
fellowseb: Removed compile-time check as discussed.

SouadHadjiat (approved):

Issue #14115, opened 2026-01-20
Custom Right-Hand Side Background Image


alice-debra:
Allow platform administrators to upload a custom image that replaces the entire right-hand side panel of the login page, while keeping the Filigran marking in place.

SamuelHassine:
@alice-debra @esrevi, finding the right image can be difficult, I propose also that the user can choose plain color or gradiant to just having full colored panel or gradiant panel corresponding to its theme in addition to an image = so image or plain color or gradiant color.

PR #14667, opened 2026-02-27
[docs/frontend] Update Theme customization documentation (#14115)

Archidoit:
### Proposed changes
Fix documentation about theme screenshots order
Remove useless props enableAskAi
Fix a translation

### Related issues
#14115


opencti-platform/opencti-front/src/private/components/analyses/reports/Root.tsx
```diff
@@ -133,7 +133,6 @@ const RootReport = () => {
                     enableQuickSubscription={true}
                     enableQuickExport={true}
                     enableEnrollPlaybook={true}
-                    enableAskAi={false}
```
fellowseb, line 136: Intentional or misclick ? (:

Archidoit: intentional :) this parameter is not used anymore

Archidoit: I put it in this PR since it's a really small removal :D

fellowseb (approved):

PR #14681, opened 2026-02-28
[backend/frontend] Support relationship-based neighbor resolution in CSV Feed columns (#14680)

### LINKED ISSUE #14680: [CSV Feed] Support relationship-based neighbor resolution in CSV Feed columns
## Description

Currently, CSV Feeds (live URL endpoints serving data at runtime) only support columns based on the direct attributes of the targeted entity type. For example, a CSV Feed based on Indicators can include columns like `indicator value`, `indicator type`, `confidence`, `created`, `updated`, etc.

**The feature request is to support resolving first-degree neighbors (via relationships) and including their attributes as additional columns in the CSV Feed, resolved at runtime when the feed URL is consumed.**

## Use Case

When consuming an Indicator-based CSV Feed, downstream systems (SIEM, SOAR, firewalls) expect a flat, enriched CSV with contextual attribution alongside raw indicators — without additional lookups. For example:

| Indicator Value | STIX ID | Malware Name | Campaign Name | Intrusion Set Name | Indicator Type | Confidence | Created | Updated |
|---|---|---|---|---|---|---|---|---|
| `1.2.3.4` | `indicator--xxx` | `EvilLoader` | `Operation X` | `APT28` | `IPv4-Addr` | `High` | 12/02/2026 | 12/02/2026 |

Here, `Malware Name`, `Campaign Name`, and `Intrusion Set Name` are **not** attributes of the Indicator itself — they come from resolving first-degree neighbors through relationships (e.g., Indicator → `indicates` → Intrusion Set, Indicator → `indicates` → Malware).

## Expected Behavior

In the CSV Feed configuration, users should be able to:

1. **Add neighbor-based columns** — select a relationship type (e.g., `indicates`, `uses`, `attributed-to`) and a target entity type (e.g., `Intrusion Set`, `Malware`, `Campaign`).
2. **Pick which attribute** of the resolved neighbor to display in the column (e.g., `name`, `description`, `aliases`).
3. The CSV Feed engine **resolves these neighbors at runtime** when the feed URL is requested and populates the corresponding columns dynamically.

## Open Design Question

If multiple neighbors match (e.g., an Indicator `indicates` 2 Intrusion Sets), a strategy is needed: first match, comma-separated list, or configurable behavior.

## Additional Context

This is a common pattern in threat intelligence sharing workflows where consumers pull a live CSV Feed URL and expect enriched, flat data ready for ingestion — no post-processing or additional API calls required.

SamuelHassine:
## Summary

Implements #14680 — Adds support for **relationship-based neighbor resolution** in CSV Feed columns. Users can now configure feed columns that resolve first-degree neighbors through relationships and extract their attributes at runtime, enabling enriched, flat CSV output without additional API calls.

### What changed

**New column mapping mode: Relationship (neighbor)**

Each entity type mapping within a feed column can now be toggled between two modes:
- **Direct** (existing behavior) — maps to a direct attribute of the entity (e.g., `indicator.value`, `indicator.confidence`)
- **Relationship** (new) — resolves a neighbor entity through a relationship and extracts an attribute from it (e.g., Indicator → `indicates` → Malware → `name`)

**Configurable multi-match behavior**

When multiple neighbors match (e.g., an Indicator `indicates` 2 Malware), users can configure per column:
- **Strategy**: "All (list)" to join all matches, or "First match" to take only the first
- **List separator**: configurable separator for joining multiple values (defaults to `,`), validated to differ from the feed's CSV separator to prevent parsing ambiguity

### Backend changes (5 files)

- **GraphQL schema** (`opencti.graphql`) — Extended `FeedMapping` with `relationship_type` and `target_entity_type`; extended `FeedAttribute` with `multi_match_strategy` and `multi_match_separator`
- **Elasticsearch index mapping** (`internalObject-registrationAttributes.ts`) — Registered new fields in the Feed entity attribute definitions for proper strict-mapping indexing
- **TypeScript types** (`store.d.ts`) — Updated `BasicStoreEntityFeed` with the new optional fields
- **Domain validation** (`feed.ts`) — Validates relationship_type/target_entity_type must both be set or both absent; validates multi_match_strategy and multi_match_separator (must differ from feed CSV separator)
- **Feed engine** (`httpRollingFeed.ts`) — Core runtime logic:
  - `resolveNeighborsForFeed()` — batch-queries relationships for all loaded entities across both directions (FROM and TO), then batch-resolves target entities via `elFindByIds`. Returns a pre-computed lookup map
  - Updated `buildCsvLines()` to handle neighbor-based mappings using the pre-resolved lookup map, applying the configurable multi-match strategy and separator

### Frontend changes (3 files + generated)

- **FeedCreation.tsx** and **FeedEdition.jsx** — Redesigned column configuration UI:
  - Entity type label with a clickable **Chip** toggle (Direct / Relationship)
  - In Relationship mode: 3-column grid with Relationship type → Target type → Attribute dropdowns, populated from `schemaRelationsTypesMapping` (both directions)
  - Multi-match strategy dropdown and list separator text field appear when any mapping uses Relationship mode
  - Dividers between entity type sections for visual clarity
- **FeedLine.jsx** — Updated Relay fragment with new fields

### Performance

- Batch relationship queries: one `fullRelationsList` call per unique (rel_type, target_type) pair — not per entity
- Batch entity resolution: one `elFindByIds` call for all target IDs
- Pre-computed lookup map before iterating CSV rows
- Bounded by existing `SIZE_LIMIT` (5000 entities)

### Backward compatibility

- All new fields are optional (nullable in GraphQL, optional in TypeScript)
- Existing feeds with only direct mappings continue to work unchanged
- `multi_match_strategy` defaults to `list`, `multi_match_separator` defaults to `,`

Closes #14680

## Test plan

- [ ] Create a feed with only direct attribute columns → verify CSV output is unchanged (backward compatibility)
- [ ] Create a feed with a relationship column (e.g., Indicator → indicates → Malware → name) → verify the neighbor's attribute appears in the CSV
- [ ] Test with multiple neighbors matching → verify comma-separated list output with "All (list)" strategy
- [ ] Switch to "First match" strategy → verify only one value appears
- [ ] Change the list separator (e.g., to `|`) → verify it is used in the CSV output
- [ ] Attempt to set list separator to same as CSV separator → verify validation error
- [ ] Toggle between Direct and Relationship mode → verify smooth switching in both directions
- [ ] Edit an existing feed with neighbor columns → verify data loads correctly in the edition form
- [ ] Test with a feed that has multiple entity types → verify per-type mapping works independently


opencti-platform/opencti-front/src/private/components/data/feeds/FeedCreation.tsx
```diff
@@ -168,6 +171,30 @@ const FeedCreation: FunctionComponent<FeedCreationFormProps> = (props) => {
...
+    const targets = new Set<string>();
+    schema.schemaRelationsTypesMapping.forEach((values, key) => {
+      if (values.includes(relType) || relType === 'related-to') {
+        const [from, to] = key.split('_');
+        if (from === entityType) targets.add(to);
+        if (to === entityType) targets.add(from);
+      }
+    });
+    return Array.from(targets).sort();
+  };
```
fellowseb: (code structure)

Consider extracting these to a utils file like `src/utils/Relation.ts` and passing `schemaRelationsRefTypesMapping` as a prop. This would prevent from redefining those at every render.

opencti-platform/opencti-front/src/private/components/data/feeds/FeedCreation.tsx
```diff
@@ -280,15 +314,16 @@ const FeedCreation: FunctionComponent<FeedCreationFormProps> = (props) => {
...
-          (m) => !m.attribute
-            || !m.type
-            || m.attribute.length === 0
-            || m.type.length === 0,
-        ).length > 0
       ) {
         return false;
       }
+      const invalidMappings = R.values(feedAttribute.mappings).filter((m) => {
+        if (!m.type || m.type.length === 0 || !m.attribute || m.attribute.length === 0) return true;
```
fellowseb: (nit)

`if (!m.type || !m.attribute) return true;` should be enough

opencti-platform/opencti-front/src/private/components/data/feeds/FeedEdition.jsx
```diff
@@ -154,6 +158,30 @@ const FeedEditionContainer = (props) => {
...
+    const targets = new Set();
+    schema.schemaRelationsTypesMapping.forEach((values, key) => {
+      if (values.includes(relType) || relType === 'related-to') {
+        const [from, to] = key.split('_');
+        if (from === entityType) targets.add(to);
+        if (to === entityType) targets.add(from);
+      }
+    });
+    return Array.from(targets).sort();
+  };
```
fellowseb: We could then here reuse the extracted utils

opencti-platform/opencti-graphql/src/domain/feed.ts
```diff
@@ -25,6 +27,15 @@ const checkFeedIntegrity = (input: FeedAddInput) => {
...
     }
     input.feed_attributes.forEach((f) => {
+      if (f.multi_match_strategy && !VALID_MULTI_MATCH_STRATEGIES.includes(f.multi_match_strategy)) {
+        throw ValidationError(`Invalid multi_match_strategy "${f.multi_match_strategy}", must be "first" or "list"`, 'multi_match_strategy');
+      }
+      if (f.multi_match_separator && f.multi_match_separator === input.separator) {
+        throw ValidationError(
+          `Multi-match separator for column "${f.attribute}" must differ from the feed CSV separator ("${input.separator}")`,
+          'multi_match_separator',
+        );
```
fellowseb, line 40: We could also add the check in the UI, preventing from adding/updating the feed if we see colliding separators. WDYT ?

opencti-platform/opencti-graphql/src/http/httpRollingFeed.ts
```diff
@@ -39,7 +47,174 @@ const escapeCsvField = (separator: string, data: string) => {
...
+  const isComplexKey = attributePath.includes('.');
+  const baseKey = isComplexKey ? attributePath.split('.')[0] : attributePath;
+  const data = (entity as any)[baseKey];
+  if (!isNotEmptyField(data)) return '';
+  if (isComplexKey) {
+    const [, innerKey] = attributePath.split('.');
+    const dictInnerData = data[innerKey.toUpperCase()];
+    return isNotEmptyField(dictInnerData) ? String(dictInnerData) : '';
+  }
+  if (Array.isArray(data)) return data.join(',');
```
fellowseb: Is it safe using `,` to join attribute values here ?

opencti-platform/opencti-graphql/src/http/httpRollingFeed.ts
```diff
@@ -39,7 +47,174 @@ const escapeCsvField = (separator: string, data: string) => {
   return data;
 };
 
-export const buildCsvLines = (elements: any[], feed: BasicStoreEntityFeed): string[] => {
+const neighborKey = (relType: string, targetType: string) => `${relType}:${targetType}`;
+
+const extractAttributeFromEntity = (entity: BasicStoreBase, attributePath: string): string => {
+  const isComplexKey = attributePath.includes('.');
+  const baseKey = isComplexKey ? attributePath.split('.')[0] : attributePath;
+  const data = (entity as any)[baseKey];
```
fellowseb, line 55: (thinking out loud, no action needed)

I believe this shows that maybe we could have a better type than `BasicStoreBase` to represent "any store object'.

fellowseb (commented):
Neat feature !
Tested on my side 👍 .

Consider augmenting `httpRollingFeed-test.ts` with a least one test with a populated neighbor map or better that calls `resolveNeighborsForFeed` to get the entire thing tested.

Also, from a UX perspective here are two suggestions:
- (bonus) Preventing from validating the form if a "List separator" matches the main "Separator"
-  (ultra bonus) Having a local state to keep track of the most recent "List separator" so that next  column definition gets populated with it instead of ` ,`

opencti-platform/opencti-front/src/private/components/data/feeds/FeedCreation.tsx
SamuelHassine: Done in 83c012d — extracted getTargetTypesForRelationship to src/utils/Relation.ts and reused the existing getRelationshipTypesForEntityType from the same file in both FeedCreation.tsx and FeedEdition.jsx. The inline functions are now removed.

opencti-platform/opencti-front/src/private/components/data/feeds/FeedCreation.tsx
SamuelHassine: Done in 83c012d - simplified to `if (!m.type || !m.attribute) return true;`

opencti-platform/opencti-front/src/private/components/data/feeds/FeedEdition.jsx
SamuelHassine: Done in 83c012d - both `FeedCreation.tsx` and `FeedEdition.jsx` now import and use `getRelationshipTypesForEntityType` and `getTargetTypesForRelationship` from `src/utils/Relation.ts`.

opencti-platform/opencti-graphql/src/domain/feed.ts
SamuelHassine: Done in 83c012d - added frontend validation in both FeedCreation and FeedEdition: the submit button is now disabled when a list separator collides with the CSV separator (`hasSeparatorCollision`), and the list separator text field shows an inline error/helperText when there's a collision.

opencti-platform/opencti-graphql/src/http/httpRollingFeed.ts
SamuelHassine: Good catch! Fixed in 83c012d - `extractAttributeFromEntity` now accepts an optional `listSeparator` parameter (defaults to `,`) instead of hardcoding it. The neighbor resolution path passes the configured `multi_match_separator` through.

opencti-platform/opencti-graphql/src/http/httpRollingFeed.ts
SamuelHassine: Agreed, this is a good observation. For now the `extractAttributeFromEntity` helper centralizes the extraction logic for both direct and neighbor paths. We can revisit the typing in a follow-up if needed.

SamuelHassine:
Added 3 new test cases in `httpRollingFeed-test.ts` covering neighbor-based CSV generation: list strategy with multiple neighbors, first-match strategy, and empty neighbors map. All addressed in 83c012d.

fellowseb (approved):

Issue #8207, opened 2024-08-29
Images embedded in report content are sized at 100% in PDF exports


damians-filigran:
## Description

When exporting a report to PDF, any inline images that were resized to fit on the page will lose their scaling and appear in the PDF at 100%, usually much larger than the space available

## Environment

1. SaaS
2. OpenCTI version: 6.2.15
3. OpenCTI client: Edge
4. Other environment details:

## Reproducible Steps

1. Create a new report
2. Go to content tab and create an HTML document
3. Write some content
4. Paste a screenshot of a tool into the page
5. Resize it to fit on the page
6. Export to PDF (simple format; no neighbours)

## Expected Output

Inline images appear on the PDF page as they did in the editor

## Actual Output

Inline images are scaled to be very large (100%?), so that they are too large for the page

## Additional information

On testing server under Report "RansomHub IOCs"

## Screenshots (optional)

![image](https://github.com/user-attachments/assets/0dc8c4b0-6a7d-4b4d-82ea-4c3296aaebdd)

damians-filigran:
Also, a larger screenshot (eg. knowledge graph) will be completely missing from the PDF export

Original HTML:
![image](https://github.com/user-attachments/assets/b9605a19-ab70-4d7a-a5a7-4dd5b38d6e37)


PDF contents:
![image](https://github.com/user-attachments/assets/ca5d320d-904a-401f-8824-4f9a305a9a5e)

romain-filigran:
Reproduced. The problem occurs with both ways of exporting to PDF: 
- embedded PDF exporter 
- export-report-pdf connector (https://github.com/OpenCTI-Platform/connectors/tree/master/internal-export-file/export-report-pdf)

romain-filigran:
related issue for the export-report-pdf connector: https://github.com/OpenCTI-Platform/connectors/issues/2587

PR #14715, opened 2026-03-02
[frontend] Improve management of images when moving from HTML to PDF (#8207)

lndrtrbn:
### Proposed changes

* Improve images in figures management

### Related issues

* #8207

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/utils/htmlToPdf/utils/pdfImageWidth.ts
+++ b/opencti-platform/opencti-front/src/utils/htmlToPdf/utils/pdfImageWidth.ts
@@ -26,12 +26,18 @@ const setImagesWidth = (content: string) => {
   updatedContent = updatedContent.replaceAll(/<figure.+?style=".*?width:([0-9\\.]+%).*?".*?>.*?<\/figure>/gi, (match, width) => {
     const widthValue = width.split('%')[0];
     const widthInPixels = (fullWidth * widthValue) / 100;
-    return match.replace('<img src="', `<img width="${widthInPixels}" src="`);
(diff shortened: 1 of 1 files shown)


opencti-platform/opencti-front/src/utils/htmlToPdf/utils/pdfImageWidth.ts
```diff
@@ -26,12 +26,18 @@ const setImagesWidth = (content: string) => {
   updatedContent = updatedContent.replaceAll(/<figure.+?style=".*?width:([0-9\\.]+%).*?".*?>.*?<\/figure>/gi, (match, width) => {
     const widthValue = width.split('%')[0];
     const widthInPixels = (fullWidth * widthValue) / 100;
-    return match.replace('<img src="', `<img width="${widthInPixels}" src="`);
+    let result = match;
+    result = result.replace(/ width="\d+"/, '');
+    result = result.replace(/ height="\d+"/, '');
+    return result.replace('<img ', `<img width="${widthInPixels}" `);
```
fellowseb, line 32: Aspect-ratio-changing edits, with specific `height` values, are not possible in CKEditor ?

fellowseb (approved):
Consider adding a unit test to `setImagesWidth` so that your use cases are documented. (Or to `htmlToPdf` with a spy on `htmlToPdfMake`)
A better testing strategy for pdf generation like this would be too have an integration test but would be too large for this fix. Maybe we can do it when moving away from CKEditor to make sure we have no regression.

lndrtrbn:
I will add the asked tests in an other PR

Issue #14707, opened 2026-03-02
Investigation: Expand knowledge popup: incorrect margins/rendering


romain-filigran:
## Description

In the **investigation graph**, when expanding the knowledge of an entity, a popup dialog appears to let the user select which knowledge to expand (target types and relationship types).

This popup is **not rendered correctly** — the margins are wrong, causing the content to be misaligned.

## Environment

OpenCTI version: 7.x

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Open an investigation graph.
2. Select an entity present in the graph.
3. Choose the **"Expand knowledge"** option.
4. Observe the popup dialog that appears.

## Expected Output

The popup should be properly rendered with correct margins and well-aligned content.

## Actual Output

<img width="1062" height="718" alt="Image" src="https://github.com/user-attachments/assets/1fd973f5-2b8f-49ef-9af6-2e763042128a" />

SamuelHassine:
@romain-filigran Can we switch, at some point, this "expand" component by true filters with built-in quick filters if necessary? This is really legacy and basically, under the hood, it's just filter. We need full fledged filtering system in the dialog with quick filters on relationship type or entity type? WDYT?

PR #14726, opened 2026-03-03
[frontend] Investigation: Expand knowledge popup: incorrect margins/rendering (#14707)

Gwendoline-FAVRE-FELIX:
### Proposed changes

* Fix margin.
* Adjust the width of the dialog box to match the guidelines.

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/14707

### DIFF
--- a/opencti-platform/opencti-front/src/components/CheckboxesField.tsx
+++ b/opencti-platform/opencti-front/src/components/CheckboxesField.tsx
@@ -43,7 +43,7 @@ const CheckboxesField = ({
     <FormControl component="fieldset" name={name}>
       <FormLabel component="legend">{label}</FormLabel>
 
-      <ButtonGroup size="small" sx={{ marginY: 1 }}>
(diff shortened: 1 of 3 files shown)


opencti-platform/opencti-front/src/components/CheckboxesField.tsx
```diff
@@ -43,7 +43,7 @@ const CheckboxesField = ({
     <FormControl component="fieldset" name={name}>
       <FormLabel component="legend">{label}</FormLabel>
 
-      <ButtonGroup size="small" sx={{ marginY: 1 }}>
+      <ButtonGroup size="small" sx={{ marginY: 1.5 }}>
```
Archidoit, line 46: we usually don't accept margin with a dot (only entire number)... but if REALLY needed let's keep that :/

Archidoit (approved):

Issue #14283, opened 2026-01-30
[Playbook] knowledge manipulation: specify target scope (all elements, main element only, or all except main element)


alice-debra:
## User story

As a playbook author I want to be able to control which elements a component applies to, so that I can modify the data in the bundle as I require. 

Detail: Add a new drop down menu to the listed component with the following options:

- All elements in the bundle,
- Only main element in the bundle,
- All except main element in the bundle.

Playbook components affected:
- Manipulate Knowledge
- Container Wrapper
- Security Coverage
- Share/Unshare with Organization
- Manage/Remove Access Restrictions
- Promote Observable to Indicator (n.b. All **observables** in the bundle)
- Extract Observables from Indicator (n.b. All **observables** in the bundle)

PR #14728, opened 2026-03-03
[backend] Split playbook components into separated files (#14283)

MTorbay-Filigran:
### Proposed changes

* Split the listed components of 14283 issue in separated files
*

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/14283

### Checklist


opencti-platform/opencti-graphql/src/modules/playbook/components/manipulate-knowledge-component.ts
```diff
@@ -0,0 +1,263 @@
...
+};
+
+interface UpdateValueConfiguration {
+  label: string;
+  value: string;
+  patch_value: string;
+}
+
+export const PLAYBOOK_MANIPULATE_KNOWLEDGE_COMPONENT: PlaybookComponent<ManipulateConfiguration> = {
+  id: 'PLAYBOOK_MANIPULATE_KNOWLEDGE_COMPONENT',
```
SouadHadjiat: This component id has been renamed from PLAYBOOK_UPDATE_KNOWLEDGE_COMPONENT to PLAYBOOK_MANIPULATE_KNOWLEDGE_COMPONENT. We can't rename component ids, otherwise all previously created playbooks with this component will fail, and we would need to migrate them.

SouadHadjiat (changes requested):
keep the component ids the same as before the refactoring

SouadHadjiat (approved):

Issue #13389, opened 2025-11-28
[Custom view] Introduce custom views


alice-debra:
Enable users to create and customize intelligence views that display the most meaningful data through reusable dashboard widgets.

- New tab: A "Custom Views" tab will be added to the platform (visible when configured)
- Configuration: Accessible via Settings → Customization → Entity Type (e.g., Intrusion Set)
- Customization: Users can build their views by adding and arranging existing dashboard widgets
- Flexibility: Each view can be tailored to specific entity types and user needs
- User-friendly: Direct configuration from settings with drag-and-drop widget management

Users will have personalized, widget-based views that present critical intelligence data in a format that best suits their workflow and requirements.

PR #14735, opened 2026-03-03
[backend] Create customView module (#13389)

fellowseb:
### Proposed changes

* Creates a `customView` module as a first chunk of work for https://github.com/OpenCTI-Platform/opencti/issues/13389
* Put in behind a `CUSTOM_VIEW` feature flag

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/13389

### Checklist

### Further comments

~The `editDefault` attribute seems unused, I'll check and remove if needed in another PR.~ Wrong, used in the front directly. Will update documentation.


opencti-platform/opencti-graphql/src/modules/customView/customView.graphql
```diff
@@ -0,0 +1,13 @@
+type CustomView implements InternalObject & BasicObject {
+    id: ID!
+    entity_type: String!
+    standard_id: String!
+    parent_types: [String!]!
+    metrics:[Metric]
```
Gwendoline-FAVRE-FELIX, line 6: Perhaps it would be useful to add:
created_at: DateTime!
updated_at: DateTime!

fellowseb: I actually added those at first but I wanted to make sure we do use them before exposing them

opencti-platform/opencti-graphql/src/modules/customView/customView-converter.ts
```diff
@@ -0,0 +1,22 @@
+import { STIX_EXT_OCTI } from '../../types/stix-2-1-extensions';
+import { buildStixDomain } from '../../database/stix-2-1-converter';
+import type { StixCustomView, StoreEntityCustomView } from './customView-types';
+import { cleanObject } from '../../database/stix-converter-utils';
+
+const convertCustomViewToStix = (instance: StoreEntityCustomView): StixCustomView => {
+  const stixDomainObject = buildStixDomain(instance);
```
Archidoit: a custom view is not a stix domain object, consider using buildStixObject instead

Archidoit: I've seen workspace are using buildStixDomain... strange... need to fix this in workspace-converter.ts as well

opencti-platform/opencti-graphql/src/modules/customView/customView-types.ts
```diff
@@ -0,0 +1,33 @@
...
+export interface StoreEntityCustomView extends StoreEntity {
+  name: string;
+  description: string;
+  manifest: string;
+  restricted_members: Array<AuthorizedMember>;
+}
+// endregion
+
+// region Stix type
+export interface StixCustomView extends StixDomainObject {
```
Archidoit: same, StixCustomView are not StixDomainObject
Need to fix this in workspace-types as well.... 
Custom views and workspaces are internal objects.

opencti-platform/opencti-graphql/src/modules/customView/customView.graphql
Archidoit: yes we add these two attributes for every internal objects usually

opencti-platform/opencti-graphql/src/modules/customView/customView.ts
```diff
@@ -0,0 +1,46 @@
...
+    id: 'customView',
+    name: ENTITY_TYPE_CUSTOM_VIEW,
+    category: ABSTRACT_INTERNAL_OBJECT,
+    aliased: false,
+  },
+  identifier: {
+    definition: {
+      [ENTITY_TYPE_CUSTOM_VIEW]: () => uuidv4(),
+    },
+    resolvers: {
```
Archidoit: It's in workspace.ts but I'm not sure we need this... what is it exactly ?

fellowseb: From what I understand it's a remnant of a way to generate a UUID for an entity based on one or more of its attributes. Doesn't seem useful in this basic case as we're using a generator function `() => uuidv4()`. Removing.

opencti-platform/opencti-graphql/src/modules/customView/customView-types.ts
fellowseb: Fixed

opencti-platform/opencti-graphql/src/modules/customView/customView-converter.ts
fellowseb: That make sense. Should've spotted that... Fixed.

I'll open the fix on workspaces in another PR as it seems impactful.

opencti-platform/opencti-graphql/src/modules/customView/customView.graphql
fellowseb: I see some entity types like `EmailTemplate` or `FintelTemplate` that don't, but OK I'll add them.

opencti-platform/opencti-graphql/src/modules/customView/customView-converter.ts
```diff
@@ -1,10 +1,10 @@
 import { STIX_EXT_OCTI } from '../../types/stix-2-1-extensions';
-import { buildStixDomain } from '../../database/stix-2-1-converter';
+import { buildStixObject } from '../../database/stix-2-1-converter';
 import type { StixCustomView, StoreEntityCustomView } from './customView-types';
 import { cleanObject } from '../../database/stix-converter-utils';
 
 const convertCustomViewToStix = (instance: StoreEntityCustomView): StixCustomView => {
-  const stixDomainObject = buildStixDomain(instance);
+  const stixDomainObject = buildStixObject(instance);
```
Gwendoline-FAVRE-FELIX: As a result, we can also change the name of the constant to const stixObject = buildStixObject(instance);

fellowseb: Changed

Gwendoline-FAVRE-FELIX (approved):

Issue #14747, opened 2026-03-03
Improve v7 retro compatibility with remote http connectors


richard-julien:
## Use case

v7 change token management.
With this it will allow upgrade remote connectors before opencti to v7 and so ensure a more easy migration

PR #14748, opened 2026-03-03
[worker/pycti] Improve v7 retro compatibility with remote http connectors (#14747)

richard-julien:
See #14747

### DIFF
--- a/client-python/pycti/connector/opencti_connector_helper.py
+++ b/client-python/pycti/connector/opencti_connector_helper.py
@@ -35,6 +35,7 @@ import time
 import uuid
 from collections import deque
 from enum import Enum
+from hmac import compare_digest
 from queue import Queue
 from typing import Callable, Dict, List, Optional, Union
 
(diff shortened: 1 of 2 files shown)


xfournet:
@richard-julien
The [graphql query here](https://github.com/OpenCTI-Platform/opencti/blob/efdef97194cf14d7fec5f5f7d308a1ee3dff9769/client-python/pycti/api/opencti_api_connector.py#L162) will fail on v6 since `jwks` doesn't exist in v6
A fallback to a graphql query without `jwks` need to be done

client-python/pycti/connector/opencti_connector_helper.py
```diff
@@ -744,31 +745,66 @@ def _data_handler(self, json_data: Dict) -> None:
...
+        :param token: The JWT token string to validate
+        :type token: str
+        :return: True if the JWT signature is valid and the token is not expired
+        :rtype: bool
         """
-        Returns True if the signature is valid and token is not expired.
-        Returns False otherwise.
+        unverified_header = jwt.get_unverified_header(token)
+        used_kid = unverified_header["kid"]
+        key = self.connector_jwks[used_kid]
```
xfournet, line 758: with v6 retro compat, connector_jwks can be None, we should be explicit if this case happen

PR #14790, opened 2026-03-05
[backend] fix migration for local strategy (#14737)

### LINKED ISSUE #14737: Disabled local authentication strategy is set to enable following conversion to V7
## Description

When we have a local authentication strategy setup in V6 and we update the platform to V7, the local strategy is created enabled in the UI instead of using the value from the environment variable (you would need to manually disable in the UI in order to make it disappear).

The expected result should be:
- if we have a working SSO authentication setup and we are Entreprise Edition/Trial, we set up the local disabled or enabled according to the value of the env variable
- if we have don't have a working SSO authentication or we have a community edition, the local should be enabled upon conversion regardless of its value in the env variable

**Workaround**: login as admin on OpenCTI platform (with access to security > Authentication) and edit local strategy configuration to disable it.

marieflorescontact:
### Proposed changes

* for an existing platform in 6.9.x, with local auth disabled => conversion set it to enabled, probably because it's config { disabled: true } and code looks for enabled here

### Related issues

* closes: https://github.com/OpenCTI-Platform/opencti/issues/14737

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/modules/authenticationProvider/authenticationProvider-migration.ts
+++ b/opencti-platform/opencti-graphql/src/modules/authenticationProvider/authenticationProvider-migration.ts
@@ -9,7 +9,7 @@ import type { AuthContext, AuthUser } from '../../types/user';
 import { AuthenticationProviderType } from '../../generated/graphql';
 import { logApp } from '../../config/conf';
 import { convertAllSSOEnvProviders } from './authenticationProvider-migration-converter';
-import { addAuthenticationProvider, getAllIdentifiers, resolveProviderIdentifier } from './authenticationProvider-domain';
+import { addAuthenticationProvider, findAllAuthenticationProvider, getAllIdentifiers, resolveProviderIdentifier } from './authenticationProvider-domain';
 import { isUserHasCapability, SETTINGS_SET_ACCESSES } from '../../utils/access';
 import { AuthRequired } from '../../config/errors';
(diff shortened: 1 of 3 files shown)


marieflorescontact:
@copilot add test coverage for this PR

opencti-platform/opencti-graphql/src/modules/authenticationProvider/authenticationProvider-migration.ts
```diff
@@ -17,6 +17,11 @@ import nconf from 'nconf';
 import { getSettings, updateCertAuth, updateHeaderAuth, updateLocalAuth } from '../../domain/settings';
 import type { BasicStoreSettings } from '../../types/settings';
 
+export const isLocalAuthEnabled = (envProviders: Record<string, any>): boolean => {
```
aHenryJard:
```suggestion
export const isLocalAuthEnabledInEnv = (envProviders: Record<string, any>): boolean => {
```

aHenryJard: I think we really need to be clear on env vs database

aHenryJard (approved):
Looks good to me, tested locally:
- on a new platform V7, local is enabled by default as expected
- on an existing platform with auth0 + local disabled, after moving to this branch the local is correctly disabled (only auth0 button is displayed)

PR #14867, opened 2026-03-09
[backend] abort engine ES operations when http request is aborted (#14872)

### LINKED ISSUE #14872: Add a way to abort elastic requests when HTTP request is cancelled
## Use case

Add a way to abort elastic requests when HTTP request is cancelled

JeremyCloarec:
### Proposed changes

* forward http request abort to engine ES search/bulk/update request
* open question: maybe we want to forward the cancel to other ES requests also?

### Related issues

* fix #14872
*

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/database/engine.ts
+++ b/opencti-platform/opencti-graphql/src/database/engine.ts
@@ -489,6 +489,31 @@ export const searchEngineInit = async (): Promise<boolean> => {
 };
 export const isRuntimeSortEnable = (): boolean => isRuntimeSortingEnable;
 
+/**
+ * Executes an engine operation with proper abort signal handling for both ElkClient and OpenSearch.
+ * - ElkClient: passes the signal as an option natively.
+ * - OpenSearch: manually hooks up the abort signal to call .abort() on the promise.
+ */
+const elExecuteWithAbortSignal = async (
+  abortSignal: AbortSignal | undefined,
+  elkOperation: (opts: { signal: AbortSignal | undefined }) => Promise<any>,
+  openSearchOperation: () => Promise<any>,
+): Promise<any> => {
+  if (engine instanceof ElkClient) {
+    const r = await elkOperation({ signal: abortSignal });
+    return oebp(r);
+  }
+  const promise = openSearchOperation();
+  if (abortSignal) {
+    abortSignal.addEventListener('abort', () => {
+      // OpenSearch client does not support abort signal, so we need to manually abort the request
+      (promise as any).abort();
+    });
+  }
+  const r_1 = await promise;
+  return oebp(r_1);
+};
+
 const BULK_MAX_RETRIES = 5;
 const BULK_INITIAL_DELAY_MS = 500;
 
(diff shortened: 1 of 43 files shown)


opencti-platform/opencti-graphql/src/types/user.d.ts
```diff
@@ -82,6 +82,6 @@ interface AuthContext {
   synchronizedUpsert?: boolean;
   previousStandard?: string;
   req?: Express.Request;
-  synchronizedUpsert?: boolean;
```
xfournet: it's not currently used in TS code, but it still exists ?

opencti-platform/opencti-graphql/src/http/httpServer.js
```diff
@@ -33,6 +33,23 @@ const CERT_KEY_CERT = conf.get('app:https_cert:crt');
...
+  };
+  req.once('aborted', () => {
+    if (!abortController.signal.aborted) {
+      abortController.abort();
+    }
+  });
+  res.once('close', abortIfDisconnected);
+  return abortController.signal;
+};
+
```
xfournet: if would rather extract the if(!aborted) abort() logic and add the extra condition directly in close method (so the comment is directly linked to the corresponding event method)

```suggestion
const createRequestAbortSignal = (req, res) => {
  const abortController = new AbortController();
  const abort = () => {
    if (!abortController.signal.aborted) {
      abortController.abort();
    }
  }; 
  req.once('aborted', abort);
  res.once('close', () => {
    // `close` is emitted for both success and disconnect; only abort on disconnect.
    if (!res.writableEnded) {
      abort();
    }
  });
  return abortController.signal;
};
```

opencti-platform/opencti-graphql/src/http/httpServer.js
```diff
@@ -138,7 +155,8 @@ const createHttpServer = async () => {
       app,
       path: `${basePath}/graphql`,
       context: async ({ req, res }) => {
-        const executeContext = await createAuthenticatedContext(req, res, 'api');
+        const requestAbortSignal = createRequestAbortSignal(req, res);
```
xfournet: should we create it for other http call ? eg taxi, csv, ... which can call ES ?

opencti-platform/opencti-graphql/src/database/engine.ts
```diff
@@ -487,21 +487,59 @@ export const searchEngineInit = async (): Promise<boolean> => {
...
+    const r = await elkOperation({ signal: abortSignal });
+    return oebp(r);
+  }
+  const openSearchOperationPromise = openSearchOperation();
+  const abortRequest = () => {
+    // OpenSearch client does not support abort signal natively, so abort the request when possible.
+    (openSearchOperationPromise as any).abort?.();
+  };
+  if (abortSignal?.aborted) {
+    abortRequest();
```
xfournet: did you check that in this case the openSearchOperation will directly terminate ?
else we should avoid to create the request

opencti-platform/opencti-graphql/src/database/engine.ts
```diff
@@ -487,21 +487,59 @@ export const searchEngineInit = async (): Promise<boolean> => {
...
+  } else if (abortSignal) {
+    abortSignal.addEventListener('abort', abortRequest, { once: true });
+  }
+  try {
+    const r_1 = await openSearchOperationPromise;
+    return oebp(r_1);
+  } finally {
+    if (abortSignal && !abortSignal.aborted) {
```
xfournet: why checking abortSignal.aborted ?

opencti-platform/opencti-graphql/src/types/user.d.ts
JeremyCloarec: it's duplicated: it's also declared line 82 :)

opencti-platform/opencti-graphql/src/database/engine.ts
JeremyCloarec: You're right, I'll check on the condition before starting the request

opencti-platform/opencti-graphql/src/database/engine.ts
JeremyCloarec: not necessary you're right 👍

xfournet (approved):

xfournet (approved):

PR #14894, opened 2026-03-11
[backend/frontend] Display name customization (#14873)

### LINKED ISSUE #14873: feat(customization): allow administrators to customize the display name of core entities
## Summary

Allow platform administrators to define a **custom display name** for any core entity type in OpenCTI, so that the platform's terminology can be adapted to match the organization's internal vocabulary — without altering the underlying STIX data model or internal entity type identifiers.

## Problem Statement

OpenCTI ships with a fixed set of entity names that reflect the STIX 2.1 standard and Filigran's own taxonomy (e.g., `Incident Response`, `Report`, `Malware`, `Threat Actor`). While this naming is standard and well-defined, many organizations use different terminology internally:

- A SOC team may refer to `Incident Response` cases as **"Generic Cases"** or **"Security Incidents"**
- A threat intelligence team may call `Reports` **"Intelligence Products"**
- A red team may refer to `Attack Patterns` as **"Techniques"**

Today, there is no way to adapt these labels without modifying the source code. This creates a friction point for adoption, especially in organizations that have strong internal vocabulary standards or that are migrating from other platforms with different terminology.

## Proposed Solution

Extend the existing **`Settings > Customization > Entity types`** section (already available via `SubTypes.tsx` / `EntitySettingSettings.tsx`) to allow administrators to set a **custom display name** (and optionally a **custom plural display name**) for each entity type.

### UX / UI Proposal

In `Settings > Customization > Entity types`, on the entity setting page for each entity (e.g., `Incident Response`), add a new **"Display name"** field:

| Field | Description |
|---|---|
| **Display name (singular)** | Custom label shown in the UI instead of the default name (e.g., `Generic Case`) |
| **Display name (plural)** | Custom plural label (e.g., `Generic Cases`) |

- If left empty, the default OpenCTI entity name is used (no breaking change).
- A **"Reset to default"** button should be available to revert to the original name.
- The custom name should appear everywhere the entity type label is rendered: navigation menus, list headers, breadcrumbs, entity detail pages, dashboards, filters, etc.

### Example

| Entity type (internal) | Default display name | Example custom display name |
|---|---|---|
| `Case-Incident` | Incident Response | Generic Case |
| `Report` | Report | Intelligence Product |
| `Threat-Actor-Group` | Threat Actor (Group) | Adversary Group |

## Technical Notes

The codebase already has a solid foundation to build on:

- **`entitySetting` module** (`opencti-platform/opencti-graphql/src/modules/entitySetting/`) manages per-entity-type configuration and is the right place to add a `custom_name` / `custom_name_plural` field.
- **`EntitySettingSettings.tsx`** is the UI component where the new field should be added.
- **`Entity.ts`** (`opencti-front/src/utils/Entity.ts`) and the i18n/label resolution layer should be updated to use the custom name when set, falling back to the default otherwise.
- The internal `entity_type` identifier (e.g., `Case-Incident`) must **not** change — only the display label is affected. This ensures STIX compatibility and API stability.
- The custom name should be **platform-wide** (not per-user), stored in the `EntitySetting` object, and accessible via GraphQL.

## Acceptance Criteria

- [ ] Administrators can set a custom singular and plural display name for any entity type in `Settings > Customization > Entity types`
- [ ] The custom name is displayed everywhere the entity type label appears in the UI (navigation, lists, filters, dashboards, breadcrumbs, entity headers)
- [ ] If no custom name is set, the default OpenCTI name is used (fully backward compatible)
- [ ] A "Reset to default" action is available per entity type
- [ ] The internal `entity_type` identifier is unchanged (STIX and API compatibility preserved)
- [ ] The custom name is accessible via the GraphQL API (on the `EntitySetting` type)

## Out of Scope

- Customizing entity **icons** (separate concern)
- Per-user or per-group display name overrides (platform-wide only for now)
- Renaming STIX properties or changing the data model

## Related Code

- `opencti-platform/opencti-front/src/private/components/settings/sub_types/EntitySettingSettings.tsx`
- `opencti-platform/opencti-graphql/src/modules/entitySetting/entitySetting-domain.ts`
- `opencti-platform/opencti-graphql/src/modules/entitySetting/entitySetting-types.ts`
- `opencti-platform/opencti-front/src/private/components/settings/sub_types/EntitySettingsOverviewLayoutCustomization.tsx`

Jipegien:
### Proposed changes

* Wire custom entity display names into the UI by replacing `t_i18n('entity_${type}')` calls with the `useEntityLabelResolver()` hook across the platform. When an administrator sets a custom display name in Settings > Customization > Entity types, it now takes effect everywhere: navigation menus, creation/deletion dialogs, filter dropdowns, list columns, graph toolbars, breadcrumbs, settings pages, and search.
* Add unit tests for `useEntityLabel` and `useEntityLabelResolver` hooks (12 test cases covering custom names, plural handling, fallback chains, and multi-setting lookup).
* Add user-facing documentation for the "Custom display name" feature in `docs/administration/entities.md`.
* Remove redundant "Reset to default" buttons from the entity settings form — clearing the field and saving already resets to the default name, consistent with other OpenCTI settings.
* Fix search on the entity types customization page (`/dashboard/settings/customization/entity_types`) to match against custom display names, not just the default i18n labels.

### Related issues

* closes #14873

### Checklist

### Further comments

This PR has a large diff (~195 files) because the original codebase had no abstraction for entity type display labels — every component independently called `t_i18n('entity_${type}')`. The migration was necessarily file-by-file.

An alternative considered was intercepting at the i18n layer itself (dynamically patching the react-intl message catalog when entity settings load). This would have required zero call-site changes, but was rejected because: (1) it couples the i18n system to Relay-dependent data, (2) plural key patterns are inconsistent across the codebase, and (3) it creates invisible behavior where `t_i18n` silently returns different values depending on API state.

The explicit hook-based approach is more maintainable and testable, though it relies on developer discipline to use `entityLabel()` instead of `t_i18n('entity_...')` in future code. A lint rule flagging the old pattern would be a good follow-up.

Known limitation: `observations/artifacts/Root.jsx` is a class component and cannot use the hook without refactoring to a functional component.

Note: custom display names are UI-only. STIX JSON exports, URLs, and GraphQL identifiers continue to use the standard STIX type identifiers, preserving interoperability.


opencti-platform/opencti-graphql/config/default.json
```diff
@@ -126,8 +127,8 @@
     "artifact_zip_password": "infected",
     "admin": {
       "email": "admin@opencti.io",
-      "password": "ChangeMe",
-      "token": "ChangeMe",
+      "password": "admin",
+      "token": "550e8400-e29b-41d4-a716-446655440000",
       "externally_managed": false
```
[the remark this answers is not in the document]
Jipegien: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/OpenCTI-Platform/opencti/pull/14894#discussion_r2920030759)

opencti-platform/opencti-graphql/config/default.json
```diff
@@ -1,6 +1,7 @@
 {
   "app": {
     "port": 4000,
+    "encryption_key": "LF7vM33p5UlgZseKtKGbGeodveCCOFnlnKTkm18NuCo=",
     "base_path": "",
```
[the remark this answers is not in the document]
Jipegien: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/OpenCTI-Platform/opencti/pull/14894#discussion_r2920030799)

Jipegien:
the e2e fail is not reproduced when playing the playwright test on my local environement

.claude/settings.local.json
```diff
@@ -0,0 +1,8 @@
+{
+  "permissions": {
```
aHenryJard: This looks like an unrelated change, can you tell me why it's required for the feature ? or else revert.

opencti-platform/docker
aHenryJard, line 1: I don't understand this change, why is it required ?

.claude/settings.local.json
Jipegien: should be ok now.

opencti-platform/opencti-front/src/private/components/analyses/external_references/ExternalReferenceCreation.tsx
```diff
@@ -196,7 +198,7 @@ const ExternalReferenceCreation: FunctionComponent<ExternalReferenceCreationProp
   const renderClassic = () => {
     return (
       <Drawer
-        title={t_i18n('Create an external reference')}
+        title={t_i18n('', { id: 'Create ...', values: { entity_type: entityLabel('External-Reference') } })}
```
Kedae, line 201: This change impacts the e2e testing

docs/docs/administration/entities.md
```diff
@@ -18,6 +18,27 @@ You can specify which entities to hide on a platform-wide basis from "Settings >
 An overview of hidden entity types is available in the "Hidden entity types" field in "Settings > Parameters."
 
 
+## Custom display name
+
+This configuration allows you to rename any entity type across the platform. When you set a custom display name, it replaces the default label everywhere in the interface: navigation menus, creation buttons and dialogs, list columns, filter dropdowns, knowledge graphs, and search results.
```
tobybutler405, line 23: I think it is worth clarifying that this will be applied for all users. As the previous section 'hidden in interface' can be applied to only certain groups of users. 

e.g. "This configuration allows you to rename any entity type across the platform for all users."

opencti-platform/opencti-front/src/private/components/entities/securityPlatforms/SecurityPlatformEditionContainer.tsx
```diff
@@ -20,11 +21,12 @@ const SecurityPlatformEditionContainer: FunctionComponent<securityPlatformContai
...
   const { t_i18n } = useFormatter();
+  const entityLabel = useEntityLabelResolver();
   const { editContext } = securityPlatform;
 
   return (
     <Drawer
-      title={t_i18n('Update a security platform')}
+      title={t_i18n('', { id: 'Update ...', values: { entity_type: entityLabel('Security-Platform') } })}
       onClose={handleClose}
       context={editContext}
```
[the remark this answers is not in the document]
Jipegien: @copilot apply changes based on [this feedback](https://github.com/OpenCTI-Platform/opencti/pull/14894#discussion_r2931529456)

opencti-platform/opencti-front/src/utils/hooks/useEntitySettings.tsx
```diff
@@ -15,7 +15,7 @@ export type EntitySetting = EntitySettingSettings_entitySetting$data;
 const useEntitySettings = (entityType?: string | string[]): EntitySetting[] => {
   const { entitySettings } = useAuth();
   const entityTypes = Array.isArray(entityType) ? entityType : [entityType];
-  return entitySettings.edges
+  return (entitySettings?.edges ?? [])
     .map(({ node }) => useFragment<EntitySettingSettings_entitySetting$key>(entitySettingFragment, node))
     .filter(({ target_type }: EntitySetting) => (entityType ? entityTypes.includes(target_type) : true));
```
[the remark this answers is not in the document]
Jipegien: @copilot apply changes based on [this feedback](https://github.com/OpenCTI-Platform/opencti/pull/14894#discussion_r2931529494)

opencti-platform/opencti-front/src/utils/hooks/useEntityLabel.ts
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * Hook that resolves the display label for an entity type.
+ * If a custom_name is set in EntitySettings, it takes precedence.
+ * Otherwise, falls back to the default i18n translation.
+ *
+ * @param entityType - The internal entity type identifier (e.g. 'Case-Incident')
+ * @param plural - Whether to return the plural form
+ * @returns The resolved display label
+ */
+const useEntityLabel = (entityType: string, plural = false): string => {
```
fellowseb, line 13: This hook is unused, dead code. To be removed along with its tests and the default export.

When this is done it will make more sense renaming the file to match the name of the hook.

opencti-platform/opencti-front/src/utils/hooks/useEntityLabel.ts
```diff
@@ -0,0 +1,62 @@
...
+ * resolve the display label for any entity type on the fly.
+ *
+ * For the singular form, call `resolve(entityType)`.
+ * For the plural form, call `resolve(entityType, defaultPluralLabel)` where
+ * `defaultPluralLabel` is the i18n key to use when no custom plural name is
+ * configured (e.g. `t_i18n('Reports')`).
+ *
+ * @returns A resolver function
+ */
+export const useEntityLabelResolver = (): ((entityType: string, defaultLabel?: string) => string) => {
```
fellowseb, line 42: In the docs and the settings panel you mention "display name" so I'd rather be coherent and name the hook `useEntityTypeDisplayName`, with no mention to the "resolver". Also, mentioning "Type" in the hook name and resolver function seems preferable, since an entity could have a label (and already has I guess).

We would use it like the following:

```
const entityTypeDisplayName = useEntityTypeDisplayName();
entityTypeDisplayName('Intrusion-Set')
```

If you prefer the "label" term then consider updating the docs and translation strings.

opencti-platform/opencti-front/src/utils/outcome_template/engine/useFileFromTemplate.test.ts
```diff
@@ -7,9 +7,16 @@ import * as env from '../../../relay/environment';
...
+import * as useDistributionGraphDataModule from '../../hooks/useDistributionGraphData';
 
 describe('Hook: useFileFromTemplate', () => {
   beforeAll(() => {
+    vi.spyOn(useDistributionGraphDataModule, 'default').mockImplementation(() => ({
+      buildWidgetProps: () => ({ series: [], redirectionUtils: [] }),
+      buildWidgetLabelsOption: () => [],
+      buildWidgetColorsOptions: () => [],
+      buildWidgetWordCloudOption: () => [],
+    }));
```
fellowseb, line 19: These changes seem unrelated to the PR.

opencti-platform/opencti-front/src/utils/hooks/useEntityLabel.ts
```diff
@@ -0,0 +1,62 @@
...
+      // If a defaultLabel is provided we treat this as a plural request
+      if (defaultLabel !== undefined && setting.custom_name_plural) {
+        return setting.custom_name_plural;
+      }
+      if (defaultLabel === undefined && setting.custom_name) {
+        return setting.custom_name;
+      }
+    }
+    // Use the provided default label, or fall back to the singular i18n key
+    return defaultLabel ?? t_i18n(`entity_${entityType}`);
```
fellowseb, line 58: Having to pass a `defaultLabel` for the plural form is kinda weird.
Also the name of the param (`defaultLabel`) doesn't reflect the behavior.

A better change would be to:

* introduce new `entity_Type_plural` translation keys for every entity type, that would provide the plural form of the default label
* change the API of the resolver to explicitly tell if we want the plural form or not:
```
const entityTypeDisplayName = useEntityTypeDisplayName();
entityTypeDisplayName('Intrusion-Set', { plural: true; }); // By default if no options are passed the singular form is returned
```

opencti-platform/opencti-front/src/private/components/analyses/groupings/GroupingCreation.tsx
```diff
@@ -122,7 +124,7 @@ export const GroupingCreationForm: FunctionComponent<GroupingFormProps> = ({
   const [commit] = useApiMutation<GroupingCreationMutation>(
     groupingMutation,
     undefined,
-    { successMessage: `${t_i18n('entity_Grouping')} ${t_i18n('successfully created')}` },
+    { successMessage: `${entityLabel('Grouping')} ${t_i18n('successfully created')}` },
```
fellowseb, line 127: nit: here we can use the existing `GROUPING_TYPE` var.
(I hope that in a near future we'll be able to have common constants to reference instead of these string literals everywhere...)

opencti-platform/opencti-front/src/private/components/pir/pir_overview/PirOverviewHistory.tsx
```diff
@@ -101,6 +103,10 @@ const PirOverviewHistory = ({ dataHistory, dataPir }: PirOverviewHistoryProps) =
         {history.map((historyItem) => {
           const { id, context_data, timestamp } = historyItem;
           const redirectURI = pirLogRedirectUri(context_data);
+          const rawType = context_data?.entity_type ?? '';
+          const entityTypeLabel = rawType && rawType[0] === rawType[0].toUpperCase()
```
fellowseb, line 107: From what I understand this check on the case of the first character is done to distinguish base entity types that are relationships from those that are not.

To make this code clearer we should:

* hide that logic in a `inferEntityTypeKind(entityType: string): 'object' | 'relationship'` function. We should note that it would be great to use the `isBasicObject` or `isBasicRelationship` backend functions here. This is another sign we need to share more code between FE and BE.
* use this new function in the ternary
* change `displayEntityTypeForTranslation` so it would handle only the relationships case; rename it to `displayRelationshipTypeForTranslation`

This same comment applies for all 4 locations using `displayEntityTypeForTranslation` and this weird check.
Also, are we sure the relationship case is relevant here and in the 4 locations ? If not (I don't have the context) we could skip it.

opencti-platform/opencti-front/src/private/components/settings/sub_types/entity_setting/EntitySettingSettings.tsx
```diff
@@ -98,8 +105,69 @@ const EntitySettingSettings = ({ entitySettingsData }: EntitySettingSettingsProp
       },
     });
   };
+
+  const handleSubmitCustomName = (name: string, value: string) => {
+    commit({
+      variables: {
+        ids: [entitySetting.id],
+        input: { key: name, value: value.trim().length > 0 ? [value.trim()] : [''] },
```
fellowseb, line 113: The check in the ternary operator seems useless. `input: { key: name, value: [value.trim()] }` does the same thing.

opencti-platform/opencti-front/src/utils/hooks/useEntityLabel.ts
```diff
@@ -0,0 +1,62 @@
...
+ * `defaultPluralLabel` is the i18n key to use when no custom plural name is
+ * configured (e.g. `t_i18n('Reports')`).
+ *
+ * @returns A resolver function
+ */
+export const useEntityLabelResolver = (): ((entityType: string, defaultLabel?: string) => string) => {
+  const { t_i18n } = useFormatter();
+  const allEntitySettings = useEntitySettings();
+
+  return (entityType: string, defaultLabel?: string): string => {
```
fellowseb, line 46: Strongly consider changing the type of `entityType` to a union of all literals for which we have a translation (`entity_xxx'). This would be a way to check you've made no typo/mistake and slowly introduces type safety for entity types.

fellowseb (commented):
### General comment

I think the approach of using a hook is good. The implementation still has a few quirks, cf my detailed comments.

>  Known limitation: observations/artifacts/Root.jsx is a class component and cannot use the hook without refactoring to a functional component.

The refactoring seems straightforward. Let's do it in another commit.

On another topic I believe we follow a convention when naming the branch when there's a linked issue: issue/1234-some-title. (We should actually update CONTRIBUTING.md to reflect that, Ill take it)

### Tests

Consider adding an E2E test where an admin would customize the label and navigate in several places checking the change is applied.

### Terminology discussion (just thinking out loud)

This PR makes me believe that using the term "entity" should be banned anywhere outside the database layer. Instead shouldn't we actually be referring to the STIX Domain Object types in this context ?

opencti-platform/opencti-graphql/src/modules/entitySetting/entitySetting.graphql
```diff
@@ -64,6 +64,9 @@ type EntitySetting implements InternalObject & BasicObject {
     search: String
   ): FintelTemplateConnection @auth(for: [SETTINGS_SETCUSTOMIZATION])
   requestAccessConfiguration: RequestAccessConfiguration @auth
+  # Custom display name
+  custom_name: String @auth
+  custom_name_plural: String @auth
```
aHenryJard, line 69: Can you please add or update a resolver test to cover this 2 new fields ?

aHenryJard: In this file https://github.com/OpenCTI-Platform/opencti/blob/2a894d93cc3ae4d951f33e5cc80f3b828f285778/opencti-platform/opencti-graphql/tests/03-integration/02-resolvers/entitySetting-test.js#L61

fellowseb, line 69: We could also go with a single field in the API: `custom_name` with two children fields `singular` & `plural`

opencti-platform/opencti-front/src/private/components/analyses/groupings/GroupingCreation.tsx
fellowseb: Addressed in other PR

opencti-platform/opencti-front/src/private/components/pir/pir_overview/PirOverviewHistory.tsx
fellowseb: Other strat: use fallback behavior like in `useEntityTranslation`

fellowseb:
Closed in favour of https://github.com/OpenCTI-Platform/opencti/pull/15199

Issue #14847, opened 2026-03-07
Live stream lines not vertically centered


SamuelHassine:
## Description

In integestion => live streams:

<img width="1561" height="261" alt="Image" src="https://github.com/user-attachments/assets/ecf9a829-618f-4e3c-b47a-8b1c031f3444" />

Archidoit:
Will be fixed in : https://github.com/OpenCTI-Platform/opencti/pull/14896

PR #14896, opened 2026-03-11
[backend/frontend] Fix drawer visible for all users (#14895)

### LINKED ISSUE #14895: consumer drawer visible for every user
## Description
consumer name or email should not be visible for all user
we need to allow access for `Manage data sharing` capability

## Reproducible Steps

1. Go do Data > Data sharing > live stream
2. click on a stream line
3. drawer opens 
4. consumer name or email is visble for everyone who can access datasharing menu

## Expected Output

We shouldn't see the consumer name or email

Image

marieflorescontact:
### Proposed changes

* needs TAXIIAPI_SETCOLLECTIONS capa for consumers api
* needs TAXIIAPI_SETCOLLECTIONS to open right panel
* migrate `Stream.tsx` to Datatable
*  refactor` StreamCollectionCreation` to remove old `sharedUpdater`

### Related issues
* fix #14895

### Checklist


xfournet:
there was an existing regression on the ... button that doesn't appear when the manage capa  is absent.

image


It should be present for access capa with only Open stream

image


we could add this fix in the same PR

opencti-platform/opencti-front/src/private/components/data/stream/PublicStreamLineDummy.jsx
xfournet, line 1: why creating a new file in JS rather than in TS ?

marieflorescontact: done

opencti-platform/opencti-front/lang/front/fr.json
```diff
@@ -346,6 +346,7 @@
   "Associated file": "Fichier associé",
   "Associated user": "Utilisateur associé",
   "At least 1 location or 1 sector is required": "Au moins 1 lieu ou 1 secteur est requis",
+  "At risk": "Au risque",
```
```suggestion
  "At risk": "A risque",
```

Archidoit:
> there was an existing regression on the ... button that doesn't appear when the manage capa is absent.
> 
> image
> It should be present for access capa with only Open stream
> 
> image
> we could add this fix in the same PR

This should have already been fixed there : https://github.com/OpenCTI-Platform/opencti/pull/14815

marieflorescontact:
> there was an existing regression on the ... button that doesn't appear when the manage capa is absent.


without "Manage data sharing" capa, popover is visible with only "Open stream"
image

is it ok @xfournet ?

xfournet:
@Archidoit indeed it notice that when backporting on 6.9.x where the #14815 fix is not present.
@marieflorescontact so yes it's ok

Archidoit:
Sorting arrows are displayed on several columns : 
image

Archidoit:
@marieflorescontact Thanks for the DataTable refacto. Thanks to this, your PR will also solve : https://github.com/OpenCTI-Platform/opencti/issues/14847

opencti-platform/opencti-front/src/components/dataGrid/components/DataTableComponent.tsx
```diff
@@ -149,6 +149,7 @@ const DataTableComponent = ({
         const percentWidth = column.percentWidth ?? extendedColumnsMap.get(key)?.percentWidth;
 
         return R.mergeDeepRight(extendedColumnsMap.get(key) as DataTableColumn, {
+          id: key,
```
ludovic: The root cause of the sorting issue is not in `DataTableComponent` but in the `dataColumns` definition in Stream.tsx.

`DataTableComponent` builds column objects by doing `R.mergeDeepRight(extendedColumnsMap.get(key), { ...column })`. For columns whose key doesn't exist in `defaultColumnsMap` (e.g. `stream_public`, `stream_live`, `consumers`, `filters`), `extendedColumnsMap.get(key)` returns `undefined`, so no `id` property is ever set on the resulting column object.

In `DataTableHeaders.tsx`, the sort indicator is shown based on `sortBy === column.id`. When `column.id` is `undefined` for multiple columns, they all match together and the sort arrow is displayed on all of them simultaneously.

The established convention — visible in `Playbooks.tsx` and other components using `DataTable` with custom keys — is to explicitly set `id` on any column whose key is not in `defaultColumnsMap`. The fix should live here rather than in the shared base component.

**Suggested change:**

```tsx
  const dataColumns = {
    name: {
      label: 'Name',
      percentWidth: 10,
      isSortable: true,
    },
    description: {
      label: 'Description',
      percentWidth: 20,
      isSortable: false,
    },
    id: {
      label: 'Stream ID',
      percentWidth: 25,
      isSortable: true,
      render: ({ id }: StreamLine_node$data) => (
        <ItemCopy content={id} variant="inLine" />
      ),
    },
    stream_public: {
      id: 'stream_public',
      label: 'Public',
      percentWidth: 10,
      isSortable: true,
      render: ({ stream_public }: StreamLine_node$data) => (
        <ItemBoolean
          label={stream_public ? t_i18n('Yes') : t_i18n('No')}
          status={stream_public}
        />
      ),
    },
    stream_live: {
      id: 'stream_live',
      label: 'Status',
      percentWidth: 10,
      isSortable: true,
      render: ({ stream_live }: StreamLine_node$data) => (
        <ItemBoolean
          label={stream_live ? t_i18n('Started') : t_i18n('Stopped')}
          status={stream_live}
        />
      ),
    },
    consumers: {
      id: 'consumers',
      label: 'Consumers',
      percentWidth: 10,
      isSortable: false,
      render: ({ consumers }: StreamLine_node$data) => { ... },
    },
    filters: {
      id: 'filters',
      label: 'Filters',
      percentWidth: 15,
      isSortable: false,
      render: ({ filters }: StreamLine_node$data) => { ... },
    },
  };
```

This aligns with how `Playbooks.tsx` handles the same pattern (`messages`, `playbook_running`) and avoids encoding a workaround for a consumer-side omission into the shared base component.

opencti-platform/opencti-front/src/private/components/data/Stream.tsx
```diff
@@ -24,127 +37,257 @@ const useStyles = makeStyles(() => ({
...
+    id: {
+      label: 'Stream ID',
+      percentWidth: 25,
+      isSortable: true,
+      render: ({ id }: StreamLine_node$data) => (
+        <ItemCopy content={id} variant="inLine" />
+      ),
+    },
+    stream_public: {
+      label: 'Public',
```
ludovic, line 175:
```suggestion
      id: 'stream_public',
      label: 'Public',
```

opencti-platform/opencti-front/src/private/components/data/Stream.tsx
```diff
@@ -24,127 +37,257 @@ const useStyles = makeStyles(() => ({
...
+      isSortable: true,
+      render: ({ stream_public }: StreamLine_node$data) => (
+        <ItemBoolean
+          label={stream_public ? t_i18n('Yes') : t_i18n('No')}
+          status={stream_public}
+        />
+      ),
+    },
+    stream_live: {
+      label: 'Status',
```
ludovic, line 187:
```suggestion
      id: 'stream_live',
      label: 'Status',
```

opencti-platform/opencti-front/src/private/components/data/Stream.tsx
```diff
@@ -24,127 +37,257 @@ const useStyles = makeStyles(() => ({
...
+      isSortable: true,
+      render: ({ stream_live }: StreamLine_node$data) => (
+        <ItemBoolean
+          label={stream_live ? t_i18n('Started') : t_i18n('Stopped')}
+          status={stream_live}
+        />
+      ),
+    },
+    consumers: {
+      label: 'Consumers',
```
ludovic, line 199:
```suggestion
      id: 'consumers',
      label: 'Consumers',
```

opencti-platform/opencti-front/src/private/components/data/Stream.tsx
```diff
@@ -24,127 +37,257 @@ const useStyles = makeStyles(() => ({
...
+              />
+            );
       },
-      filters: {
-        label: 'Filters',
-        width: '15%',
-        isSortable: false,
+    },
+    filters: {
+      label: 'Filters',
```
ludovic, line 216:
```suggestion
      id: 'filters',
      label: 'Filters',
```

ludovic (changes requested):

opencti-platform/opencti-front/src/components/dataGrid/components/DataTableComponent.tsx
marieflorescontact: thanks. done.

ludovic (approved):

PR #14903, opened 2026-03-12
[backend] add sha256 to file metadata & use sha256 to check for duplicate upload (#14877)

### LINKED ISSUE #14877: Live Stream Sync: File operations create infinite event loop in bidirectional sync
## Description
When two OpenCTI instances are configured with bidirectional live stream sync (A to B and B to A), uploading a file to an entity on Instance A triggers an infinite file replace loop between the two instances. The sync service account on each instance repeatedly replaces the file, generating an unbounded cycle of `storeUpdateEvent` events.

This is distinct from the Automation Manager file manipulation loop fixed in #14585 (released in 7.260305.0). That fix addressed the playbook/Automation Manager path only. This issue affects the **sync ingestion path**, where the root cause is different: the `upload()` function in `file-storage.ts` defaults `metadata.version` to `now()` when no version is provided by the caller. Since the sync ingestion code path does not preserve or pass the source file's version, every sync operation generates a new version timestamp. This defeats the `untouched` version check, causing the file to be re-uploaded and re-indexed every time, which emits a new `storeUpdateEvent` that streams back to the originating instance.

**Root cause code path:**
1. `stixCoreObjectImportPush` in `stixCoreObject.js` handles file upload during sync ingestion.
2. It calls `uploadToStorage()` / `upload()` in `file-storage.ts`.
3. In `upload()`, the version check (`if (!metadata.version) { metadata.version = now(); }`) assigns a fresh timestamp because sync ingestion does not pass the source file version.
4. The existing file's version is always older than `now()`, so the `untouched: true` early return is never reached.
5. The file is re-uploaded, re-indexed, and a `storeUpdateEvent` is emitted.
6. The other instance receives this event via the live stream, processes it, and repeats the cycle.

**Operational impacts:**
- Playbooks get overwhelmed with the volume of tasks triggered by the looping events
- Users subscribed to affected entities receive floods of notifications (customer had to disable email notifications entirely)
- S3 bucket operations spike, hitting storage usage limits

## Environment
6.9.24

## Reproducible Steps

1. Set up two OpenCTI instances (Instance A and Instance B), both on version 6.9.24 or later.
2. On Instance A, create a live stream filtered to Entity type = Incident Response (IR Cases).
3. On Instance B, create a live stream ingestion pointing to Instance A's stream.
4. On Instance B, create a live stream filtered to Entity type = Incident Response (IR Cases).
5. On Instance A, create a live stream ingestion pointing to Instance B's stream (bidirectional sync now established).
6. On Instance A, create an IR Case entity.
7. Wait for the IR Case to sync to Instance B.
8. On Instance A, navigate to the IR Case's Data tab and upload a file.
9. Observe the Activity log on both instances: the sync service account will repeatedly replace the file in an infinite loop.

## Expected Output
When a file is uploaded to an entity on Instance A:
- The file syncs to Instance B once.
- Instance B recognizes the file is unchanged (same version) and returns `untouched: true`.
- No further events are generated. The loop terminates after one sync cycle.

## Actual Output
- The file syncs to Instance B, which re-uploads it with a new `now()` version, generating a `storeUpdateEvent`.
- That event streams back to Instance A, which repeats the same process.
- This creates an infinite A → B → A → B loop of file replace operations.
- The Activity log on both instances shows the sync service account replacing the file repeatedly (6+ visible replacements and growing).
- Playbooks are overwhelmed, notifications flood, and S3 usage spikes.

## Additional information
**Related issues:**
- #14585 (Playbook: manipulating container files even if not needed) — fixed the same class of bug for the **playbook/Automation Manager path** in release 7.260305.0. The customer has this fix applied. The sync ingestion path was not covered by that fix.
- #2623 (Sync data looping, 2022) — similar symptom pattern (sync loop) but different root cause and era (OpenCTI 5.x). Included for historical context only.

**Suggested fix direction:**
The sync ingestion path needs to preserve and pass the source file's version when calling `uploadToStorage()`. This would allow the `untouched` version check in `upload()` to correctly detect that the file has not changed and return early without emitting a new event. This is analogous to how #14585 fixed the playbook path.

**Note:** This issue only affects **file operations** in bidirectional sync. Metadata-only sync operations (entity field updates, relationship changes) are not affected because they do not go through the `upload()` / `uploadToStorage()` code path.

JeremyCloarec:
### Proposed changes

* add sha256 in internal file metadata
* extract sha256 on upload
* check if existing document has the same sha256 to check for duplicates
* known limitation: sha256 of uploaded file in computed from in memory content, we might need to check for large files & move the content to disk if too large

### Related issues

* fix #14877 
*

### Checklist


opencti-platform/opencti-graphql/src/modules/internal/document/document.ts
```diff
@@ -34,6 +34,7 @@ const attributes: Array<AttributeDefinition> = [
     isFilterable: false,
     mappings: [
       { name: 'version', label: 'Version', type: 'date', mandatoryType: 'internal', editDefault: false, multiple: false, upsert: false, isFilterable: true },
+      { name: 'sha256', label: 'Sha256', type: 'string', format: 'short', mandatoryType: 'internal', editDefault: false, multiple: false, upsert: false, isFilterable: true },
```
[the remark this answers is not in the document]
xfournet: @JeremyCloarec same question :)

opencti-platform/opencti-graphql/src/database/file-storage.ts
```diff
@@ -594,10 +596,25 @@ export const upload = async (
     const draftPrefix = getDraftFilePrefix(draftContext);
     key = `${draftPrefix}${key}`;
   }
+
+  // Read the stream into a buffer to compute SHA256 hash and reuse for upload
+  const readStream = createReadStream();
+  const chunks: Buffer[] = [];
```
xfournet: we must handle big files, else it will be easy to kill the platform with OOME

opencti-platform/opencti-graphql/src/modules/internal/document/document.ts
```diff
@@ -34,6 +34,7 @@ const attributes: Array<AttributeDefinition> = [
     isFilterable: false,
     mappings: [
       { name: 'version', label: 'Version', type: 'date', mandatoryType: 'internal', editDefault: false, multiple: false, upsert: false, isFilterable: true },
+      { name: 'sha256', label: 'SHA-256', type: 'string', format: 'short', mandatoryType: 'internal', editDefault: false, multiple: false, upsert: false, isFilterable: true },
```
fellowseb, line 37: Shouldn't it be `isFilterable: false` ?

fellowseb (approved):

opencti-platform/opencti-graphql/src/database/file-storage.ts
```diff
@@ -608,18 +624,19 @@ export const upload = async (
 
   const creatorId = (currentFile?.metaData as FileMetadata)?.creator_id ? (currentFile.metaData as FileMetadata).creator_id : user.id;
 
-  // Upload the data
-  const readStream = createReadStream();
+  // Upload the data from the buffered content
+  const uploadReadStream = createReadStream();
```
xfournet, line 628: are you sure we can call createReadStream multiple time on a graphql upload file object ?

xfournet: yes it's possible since it's handled out of the box with graphql upload that write the content on disk

https://github.com/jaydenseric/graphql-upload?tab=readme-ov-file#architecture

Issue #14619, opened 2026-02-25
feat(ui): UI improvements follow-up


alice-debra:
## Use case

This parent issue tracks all UI improvements follow-up, including tasks and bugs.

PR #14914, opened 2026-03-13
[frontend] Display option labels instead of IDs in PlaybookFlowFieldArray (#14619)

Gwendoline-FAVRE-FELIX:
### Proposed changes

* Convert playbook field array options to { label, value } format so AutocompleteField can display the option names instead of their IDs after selection.

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/14619

### DIFF
--- a/opencti-platform/opencti-front/src/components/AutocompleteField.tsx
+++ b/opencti-platform/opencti-front/src/components/AutocompleteField.tsx
@@ -59,6 +59,7 @@ const AutocompleteField = <
     getOptionLabel,
     endAdornment,
     disabled,
+    renderTags,
   } = muiProps;
 
   const [, meta] = useField(name);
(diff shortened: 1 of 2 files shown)


SarahBocognano:
For @lndrtrbn : I suggested to Gwen to remove all the shenanigans for the Formik component as Field Option is natively taken into account by Autocomplete but I need your expertise on this one as ok it's "working" on paper but you are the sensei

lndrtrbn:
> For @lndrtrbn : I suggested to Gwen to remove all the shenanigans for the Formik component as Field Option is natively taken into account by Autocomplete but I need your expertise on this one as ok it's "working" on paper but you are the sensei

I need to make some tests to verify we did not make it on purpose on the first place but it seems a good idea indeed

lndrtrbn (changes requested):
The solution broke playbook components using selects, for example `Container Wrapper`.

Why it brokes? Because before the option (the thing saved in the configuration) was a string and now it is an object `{ value, label }`

lndrtrbn (approved):

Issue #14878, opened 2026-03-10
[ci] Handle release of LTS version


efaure:
## Use case

With LTS version we need to handle new format of tag : 

 - Trigger release build on tag including "lts" version
 - Do not block fail with "version mismatch" if pycti version is not similar to the tag, as it's expected for lts version (cf https://github.com/OpenCTI-Platform/release/pull/20)

PR #14950, opened 2026-03-16
[ci] Switch LTS tag format from 7.260309.0-ltsx to 7.260309.0-lts.x and improve pycti check (#14878)

efaure:
### Proposed changes

* Update tag regex to trigger Circle CI release build to replace format 7.260309.0-ltsx with  7.260309.0-lts.x 

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/14878

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -946,7 +946,7 @@ workflows:
       - build_pycti:
           filters:
             tags:
-              only: /[0-9]+(\.[0-9]+)+(\.[0-9]+)*(-lts[0-9]*)?/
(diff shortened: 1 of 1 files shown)


efaure:
@xfournet  I made 2 more update in this PR : 

 - Fix the pycti version check (to not use "post" version), and fail the step is the check is invalid
 - Update the regex to trigger release build to be more restrictive. The new one do not trigger on tag with: 
    - Unexpected prefix / suffix
    - Unexpected number of digit group (eg it will now refuse 7.260309 or 7.260309.0.1)

xfournet (approved):

Issue #14288, opened 2026-01-30
Filter on 'description' attribute in Live streams, Notifiers, and Playbooks


romain-filigran:
## Use case

In streaming-related features (Live streams, Notifiers, Playbooks), it is currently not possible to filter on the `description` attribute of entities.

This prevents use cases such as:

- Notify me when a Report description contains a given keyword (e.g. a ransomware name, customer name, project, sector).
- Trigger a Playbook when an Indicator description contains keyword X (“critical”, “actively exploited”, etc.).
- Expose a live stream of all entities (e.g. Indicators) whose description contains a specific keyword or pattern.

## Current Workaround

N/A

## Proposed Solution

Enable filtering on the description field for all entities that support it, in:
- Live streams
- Notifiers
- Playbooks

Supported operators for description:
- Empty
- Not empty
- Search
- Contains
- Not contains

PR #14952, opened 2026-03-16
[frontend/backend] Introduce filter on description in streams/playbooks/triggers (#14288)

Archidoit:
### Proposed changes
Introduce the 'description' filter for streams, playbooks and triggers :

- add 'description' in stix filtering testers, with all the operators for strings
- add backend tests
- update documentation

### Related issues
#14288

### Screenshot
image

### DIFF
--- a/opencti-platform/opencti-front/src/components/filters/FilterChipPopover.tsx
+++ b/opencti-platform/opencti-front/src/components/filters/FilterChipPopover.tsx
@@ -458,7 +458,8 @@ export const FilterChipPopover: FunctionComponent<FilterChipMenuProps> = ({
   };
 
   const displayOperatorAndFilter = (fKey: string, subKey?: string, disabled = false) => {
-    const availableOperators = getAvailableOperatorForFilter(filterDefinition, subKey);
+    const isStixFiltering = entityTypes?.includes('Stix-Filtering');
+    const availableOperators = getAvailableOperatorForFilter(filterDefinition, subKey, { isStixFiltering });
     const finalFilterDefinition = useFilterDefinition(fKey, entityTypes, subKey);
     return (
       <>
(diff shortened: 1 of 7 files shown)


opencti-platform/opencti-front/src/utils/filters/filtersUtils.tsx
```diff
@@ -757,7 +759,7 @@ export const getAvailableOperatorForFilterKey = (
     return ['eq', 'not_eq'];
   }
   if (isBasicTextFilter(filterDefinition)) {
-    if (filterDefinition.type === 'string') {
+    if (filterDefinition.type === 'string' || opts?.isStixFiltering) {
       return ['eq', 'not_eq', 'nil', 'not_nil', 'contains', 'not_contains',
         'starts_with', 'not_starts_with', 'ends_with', 'not_ends_with', 'search'];
```
fellowseb: This seems a little odd: couldn't we have a "STIX filter" affecting a non-string attribute ?

opencti-platform/opencti-front/src/components/filters/FilterChipPopover.tsx
```diff
@@ -458,7 +458,8 @@ export const FilterChipPopover: FunctionComponent<FilterChipMenuProps> = ({
   };
 
   const displayOperatorAndFilter = (fKey: string, subKey?: string, disabled = false) => {
-    const availableOperators = getAvailableOperatorForFilter(filterDefinition, subKey);
+    const isStixFiltering = entityTypes?.includes('Stix-Filtering');
```
fellowseb, line 461: For context could you elaborate a little about this 'Stix-Filtering' entity type ?

Archidoit: entityTypes contains the entity types concerned in the context, enabling to fetch the correct filter keys (available filter keys depend on the entity types, except for stix filtering). If we are in stream/triggers/playbooks, we fetch all the filter keys available in stix filtering, those the Stix-Filtering entity types in those cases

opencti-platform/opencti-front/src/utils/filters/filtersUtils.tsx
Archidoit: yes we can, but in this case, the available operators are the same as for dynamic filtering

fellowseb (approved):

Issue #14918, opened 2026-03-13
Improve data sharing in public mode


nino-filigran:
## Use case

We want to improve data sharing in public mode to be able to specify user associated with sharing, in order to give more control to administrators regarding which data is shared (similar than public dashboard). 

## Proposed Solution

- if user has the right to manage credentials, user will be able to set a data sharing as public
- if the data sharing is public, then a user must be picked 
- a max marking would need to be set too, like public dashboard. 

## If the feature request is approved, would you be willing to submit a PR?

Yes / No (Help can be provided if you need assistance submitting a PR)

PR #14975, opened 2026-03-16
[backend] refactor: migrate feed, taxiiCollection and streamCollection to new module pattern (#14918)

ludovic:
### Proposed changes

* Migrate `Feed`, `TaxiiCollection` and `StreamCollection` from the legacy domain/resolver pattern to the new TypeScript module pattern (co-located `*-types.ts`, `*-converter.ts`, `*-domain.ts`, `*-resolver.ts`, `*-graphql.ts`, `*.ts`, `*.graphql` under `src/modules/dataSharing/`)
* Remove the corresponding type definitions, queries and mutations from the global `opencti.graphql` schema (now declared via `extend type Query/Mutation` in each module's own `.graphql` file)
* Delete legacy files: `src/domain/feed.ts`, `src/domain/taxii.js`, `src/domain/stream.js`, `src/resolvers/feed.ts`, `src/resolvers/taxii.js`, `src/resolvers/stream.js`
* Update all consumers (`schema.js`, `sseMiddleware.js`, `cache.ts`, `httpRollingFeed.ts`, `httpTaxii.js`, `stixDomainObject.ts`) to import from the new module paths
* Keep re-exports of `ENTITY_TYPE_FEED`, `ENTITY_TYPE_TAXII_COLLECTION` and `ENTITY_TYPE_STREAM_COLLECTION` in `internalObject.ts` for backward compatibility with existing migrations

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/14918

### Checklist

### Further comments

This PR is a pure structural refactor — no functional behaviour is changed. The three `dataSharing` entities (`Feed`, `TaxiiCollection`, `StreamCollection`) are migrated to the same module pattern already used by `Workspace`, `PublicDashboard`, `DraftWorkspace`, etc.

The re-exports in `internalObject.ts` are intentional: existing migration files import these constants from that path and are never retroactively modified, so the re-exports preserve a stable public surface without touching historical migrations.


opencti-platform/opencti-graphql/src/modules/dataSharing/streamCollection-resolver.ts
```diff
@@ -0,0 +1,37 @@
...
+import { fetchStreamInfo } from '../../database/stream/stream-handler';
+
+const streamCollectionResolvers: Resolvers = {
+  Query: {
+    streamCollection: (_, { id }, context) => findById(context, context.user, id) as unknown as Promise<StreamCollection>,
+    streamCollections: (_, args, context) => findStreamCollectionPaginated(context, context.user, args) as unknown as Promise<StreamCollectionConnection>,
+    redisStreamInfo: () => fetchStreamInfo() as unknown as Promise<RedisStreamInfo>,
+  },
+  StreamCollection: {
+    authorized_members: (stream, _, context) => getAuthorizedMembers(context, context.user, stream as unknown as BasicStoreEntity),
```
marieflorescontact: it seems that Ts doesn't look the right type for stream:

Image

it should look into `streamCollection-types.ts` for the `BasicStoreEntityStreamCollection`

opencti-platform/opencti-graphql/src/modules/dataSharing/streamCollection-resolver.ts
```diff
@@ -0,0 +1,37 @@
...
+    consumers: (stream) => getStreamCollectionConsumers(stream.id),
```
marieflorescontact, line 23: we have probably the same issue for all others "as unknown as something"

marieflorescontact (commented):
functionnaly tested all good (CRUD)
we need to investigate the " as unkown as something" issue

opencti-platform/opencti-graphql/src/modules/dataSharing/streamCollection-resolver.ts
ludovic: Yes, I forgot to add the modules to graphql-codegen.yml.

marieflorescontact (approved):
looks good 🙂

PR #14980, opened 2026-03-17
[backend] Improve SSE resolutions (#15027)

### LINKED ISSUE #15027: Improve SSE resolutions
## Environment

4. Other environment details:

## Reproducible Steps

1. { e.g. Run ... }
2. { e.g. Click ... }
3. { e.g. Error ... }

richard-julien:
closes #15027

### DIFF
--- a/opencti-platform/opencti-graphql/src/graphql/sseMiddleware.js
+++ b/opencti-platform/opencti-graphql/src/graphql/sseMiddleware.js
@@ -498,6 +498,13 @@ const createSseMiddleware = () => {
     const { user } = req;
     const { id: eventId, data: eventData } = element;
     const { type, data: stix, message } = eventData;
+    // First verify the user has access to this relation's markings/organizations.
+    // This relation comes from the raw stream event and was not filtered by isStixMatchFilterGroup
+    // (isCurrentlyVisible was false), so we must explicitly check access rights here.
+    const isRelationAccessible = await isUserCanAccessStixElement(context, user, stix);
+    if (!isRelationAccessible) {
+      return;
+    }
     const isRel = stix.type === 'relationship';
     const fromId = isRel ? stix.source_ref : stix.sighting_of_ref;
     const toId = isRel ? stix.target_ref : stix.where_sighted_refs[0];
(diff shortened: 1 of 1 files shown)


opencti-platform/opencti-graphql/src/graphql/sseMiddleware.js
```diff
@@ -647,17 +654,23 @@ const createSseMiddleware = () => {
...
-                      const countRelatedContainers = await elCount(context, user, streamQueryIndices, queryOptions);
-                      // At least one container is matching the filter, so publishing the event
-                      if (countRelatedContainers > 0) {
-                        await resolveAndPublishMissingRefs(context, cache, channel, req, eventId, stix);
-                        await client.sendEvent(eventId, event, eventData);
-                        cache.set(stix.id, 'hit');
+                      // But first, verify the user has access to this element (markings, organizations, etc.)
+                      // isCurrentlyVisible=false could be due to filter mismatch OR marking restriction;
+                      // we must not send data the user is not allowed to see.
+                      const isAccessible = await isUserCanAccessStixElement(context, user, stix);
```
xfournet, line 660: we compute and check if stix is accessible only here.
Why for others stix usage that are done before, we don't check if it's accessible ?

aHenryJard (approved):
Tested locally, works as expected for:
- a report that contains observable with a marking: works as expected on stream sharing
- a relationship between malware with a marking: work as expected on stream sharing

opencti-platform/opencti-graphql/src/graphql/sseMiddleware.js
xfournet: @copilot  can you check if we should check stix accessibility for the other usage of the stix variable ?

SouadHadjiat (approved):
tested locally too, looks good for me
I tested the main usecase with a report containing an entity with marking that is modified, it's not published anymore in the stream, and it's published if marking is correct.

opencti-platform/opencti-graphql/src/graphql/sseMiddleware.js
xfournet: checked that every path is correctly covered by security access checks

xfournet (approved):

Issue #14966, opened 2026-03-16
Error when clicking enrichment button after adding an external reference


snitram59:
## Description

When adding an external reference and then clicking the enrichment button, OpenCTI returns the following error message:

`An unknown error occurred. Please provide a support package to your administrator or OpenCTI maintainers.`

This issue occurs on OpenCTI version `7.260309.0`.

## Environment

2. OpenCTI version: 7.260309.0

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Open an entity in OpenCTI.
2. Add an external reference to the entity.
3. Click on the enrichment button for this external reference.
4. Observe the error message.

## Expected Output

Clicking the enrichment button should trigger the enrichment process based on the external reference without returning an error.

## Actual Output

An error message is displayed:

`An unknown error occurred. Please provide a support package to your administrator or OpenCTI maintainers.`

The enrichment does not appear to start successfully.

## Additional information

<img width="1898" height="1023" alt="Image" src="https://github.com/user-attachments/assets/d3564435-b223-4e04-9063-a6b1b499957a" />

## Screenshots (optional)

romain-filigran:
Reproduced.

<img width="573" height="160" alt="Image" src="https://github.com/user-attachments/assets/e52b20fb-5da6-44cf-8501-afb541d813f4" />

PR #14987, opened 2026-03-17
[frontend] Error when clicking enrichment button after adding an external reference (#14966)

SarahBocognano:
### Proposed changes

* change the size "large' of the refresh button as it's not part of our custom theme (only small and default)

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/14966

### DIFF
--- a/opencti-platform/opencti-front/src/components/common/button/Button.utils.ts
+++ b/opencti-platform/opencti-front/src/components/common/button/Button.utils.ts
@@ -196,7 +196,7 @@ export const getButtonContentSx = () => ({
 });
 
 export const getSizeConfig = (theme: Theme, size: ButtonSize, iconOnly: boolean) => {
-  const config = theme.button.sizes[size];
+  const config = theme.button.sizes[size] ?? theme.button.sizes['default'];
 
   return {
     ...config,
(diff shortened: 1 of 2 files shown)


opencti-platform/opencti-front/src/components/common/button/Button.utils.ts
```diff
@@ -196,7 +196,7 @@ export const getButtonContentSx = () => ({
 });
 
 export const getSizeConfig = (theme: Theme, size: ButtonSize, iconOnly: boolean) => {
-  const config = theme.button.sizes[size];
+  const config = theme.button.sizes[size] ?? theme.button.sizes['default'];
```
JeremyCloarec: Is this change still needed if we change the size to default in ExternalReferenceEnrichmentLines? From the types defined, size should always exist in the sizes of the button

JeremyCloarec (approved):

opencti-platform/opencti-front/src/components/common/button/Button.utils.ts
fellowseb: > From the types defined, size should always exist in the sizes of the button

It's the proof that it's not always the case (:.

But yeah, although this doesn't hurt, I agree this is a bit too much defensive IMO.
I think we can remove this. If you choose to keep it (for extra safety), please add `| undefined` to the type of the parameter.

Also I haven't found other occurrences of wrong type passed in JSX files for the `size` prop.

fellowseb (approved):

labo-flg (approved):

PR #14988, opened 2026-03-17
[frontend] Error when clicking enrichment button after adding an external reference (#14966)

SarahBocognano:
* https://github.com/OpenCTI-Platform/opencti/issues/14966

### DIFF
--- a/opencti-platform/opencti-front/src/components/common/button/Button.utils.ts
+++ b/opencti-platform/opencti-front/src/components/common/button/Button.utils.ts
@@ -196,7 +196,7 @@ export const getButtonContentSx = () => ({
 });
 
 export const getSizeConfig = (theme: Theme, size: ButtonSize, iconOnly: boolean) => {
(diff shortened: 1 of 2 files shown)


opencti-platform/opencti-front/src/components/common/button/Button.utils.ts
```diff
@@ -196,7 +196,7 @@ export const getButtonContentSx = () => ({
 });
 
 export const getSizeConfig = (theme: Theme, size: ButtonSize, iconOnly: boolean) => {
-  const config = theme.button.sizes[size];
+  const config = theme.button.sizes[size] ?? theme.button.sizes['default'];
```
JeremyCloarec: Is this change still needed if we change the size to default in ExternalReferenceEnrichmentLines? From the types defined, size should always exist in the sizes of the button

JeremyCloarec (approved):

PR #15003, opened 2026-03-17
[frontend] When putting number widgets in the middle of a dashboard, it breaks horizontal alignement (#14856)

### LINKED ISSUE #14856: When putting number widgets  in the middle of a dashboard, it breaks horizontal alignement
## Description

When putting number widgets  in the middle of a dashboard, it breaks horizontal alignement

Image

Because the layout is not the same as other widgets.

SarahBocognano:
### Proposed changes

* Wrapped AuditNumber widget component in ContainerWrapper like other Widgets

### Related issues

* Closes: https://github.com/OpenCTI-Platform/opencti/issues/14856


lndrtrbn (commented):
Image

It seems the issue is not fixed for all types of widgets, widgets for entities and relationships still have the issue. Also for the audit widget, the title is duplicated and action buttons not totally aligned

opencti-platform/opencti-front/src/public/components/dashboard/stix_relationships/PublicStixRelationshipsNumber.tsx
```diff
@@ -67,6 +67,8 @@ const PublicStixCoreRelationshipsNumber = ({
   startDate,
   endDate,
   title,
+  variant,
+  height,
```
lndrtrbn: Why do you add those properties as they are not used when the component is called?

Same for the 2 components (entities, relationships)

opencti-platform/opencti-front/src/public/components/dashboard/stix_relationships/PublicStixRelationshipsNumber.tsx
```diff
@@ -81,20 +83,23 @@ const PublicStixCoreRelationshipsNumber = ({
   );
 
   return (
-    <>
+    <WidgetContainer
+      padding="medium"
+      height={height}
+      title={parameters?.title ?? title ?? t_i18n('Relationships number')}
```
lndrtrbn: I think here it's only "Relationships number" and title is used inPublicStixCoreRelationshipsNumberComponent

Same for the 2 components (entities, relationships)

lndrtrbn (approved):

Issue #14978, opened 2026-03-16
[Sighting] `x_opencti_negative` field is not updated on upsert


Lhorus6:
## Description

When a sighting is created between two entities with a specific `x_opencti_negative` value (e.g., `true`, setting the qualification to False Positive), re-ingesting the same sighting via the API with a different `x_opencti_negative` value (e.g., `false`) does not update the qualification.

All other upsertable fields behave as expected: `attribute_count` increments, `first_seen` and `last_seen` are updated. However, the `x_opencti_negative` field remains unchanged regardless of the new value provided.

## Environment

OpenCTI `7.260309.0`

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Create a sighting between two entities (e.g., a Domain-Name and an Organization) with `x_opencti_negative` set to `true` (qualification: False Positive).
2. Re-create the same sighting via the API with `x_opencti_negative` set to `false`.
3. Observe that the qualification remains False Positive.
4. Re-create the same sighting with a different `last_seen` value and `x_opencti_negative` set to `false`.
5. Observe that `last_seen` is updated but the qualification still remains False Positive.

You can use this script to reproduce, playing with the sightings inputs line 40 for `last_seen` the and 42 for the `x_opencti_negative`

[sightings_creation.py](https://github.com/user-attachments/files/26038510/sightings_creation.py)

## Expected Output

When a sighting is re-ingested via the API with a different `x_opencti_negative` value, the sighting qualification should be updated accordingly (e.g., from False Positive to True Positive).

## Actual Output

The `x_opencti_negative` field is never updated on upsert. The sighting qualification remains locked to the value set during initial creation. All other fields (`attribute_count`, `first_seen`, `last_seen`) are correctly updated.

PR #15006, opened 2026-03-17
[backend] Sighting x_opencti_negative upsert (#14978)

Archidoit:
### Proposed changes
When creating an already-existing sighting with x_opencti_negative different from the one already created, the field should be upserted.

### Related issues
#14978


fellowseb (approved):
👍 🚢

PR #15048, opened 2026-03-18
[backend] replace SYSTEM_USER with real user impersonation on public data sharing endpoints - Chunk B (#14918)

### LINKED ISSUE #14918: Improve data sharing in public mode

ludovic:
### Proposed changes

* Add a `resolvePublicUser` helper in `dataSharing-utils.ts` that resolves the configured user from cache and rejects internal system users
* Add domain validation: creating or editing a public sharing endpoint without configuring a `*_public_user_id` now throws a `FunctionalError`
* Add `*_public_user_id` and `*_public_authorized_markings` fields to `Stix*` interfaces and converters so they are serialized in STIX bundles

### Related issues
* Fixes #14918

### Checklist

### Further comments

This is part of a larger effort (issue #14918) to bring proper access control to public data sharing endpoints.


opencti-platform/opencti-graphql/src/modules/dataSharing/feed-domain.ts
```diff
@@ -83,6 +86,9 @@ export const findById: DomainFindById<BasicStoreEntityFeed> = async (context: Au
 };
 export const editFeed = async (context: AuthContext, user: AuthUser, id: string, input: FeedAddInput): Promise<BasicStoreEntityFeed> => {
   checkFeedIntegrity(input);
+  if (input.feed_public && !input.feed_public_user_id) {
+    throw FunctionalError('A user must be configured when the feed is public', { id });
```
marieflorescontact: what about old public feed? I mean without feed_public_user_id (no migration)
we need to handle this case

opencti-platform/opencti-graphql/src/modules/dataSharing/streamCollection-domain.ts
```diff
@@ -22,6 +23,9 @@ export const createStreamCollection = async (context: AuthContext, user: AuthUse
   if (input.filters) {
     validateFilterGroupForStixMatch(JSON.parse(input.filters));
   }
+  if (input.stream_public && !input.stream_public_user_id) {
+    throw FunctionalError('A user must be configured when the stream collection is public');
```
marieflorescontact, line 28: what about old public stream? I mean without stream_public_user_id (no migration)
we need to handle this case

opencti-platform/opencti-graphql/src/modules/dataSharing/taxiiCollection-domain.ts
```diff
@@ -28,6 +28,9 @@ const STIX_MEDIA_TYPE = 'application/stix+json;version=2.1';
 
 // Taxii graphQL handlers
 export const createTaxiiCollection = async (context: AuthContext, user: AuthUser, input: TaxiiCollectionAddInput) => {
+  if (input.taxii_public && !input.taxii_public_user_id) {
+    throw FunctionalError('A user must be configured when the Taxii collection is public');
+  }
```
marieflorescontact, line 34: well same comment

opencti-platform/opencti-graphql/src/graphql/sseMiddleware.js
```diff
@@ -167,22 +168,30 @@ const computeUserAndCollection = async (req, res, { context, user, id }) => {
 
 const authenticateForPublic = async (req, res, next) => {
   const context = await createAuthenticatedContext(req, res, 'stream_authenticate');
-  const user = context.user ?? SYSTEM_USER;
   req.context = context;
-  req.userId = user.id;
-  req.user = user;
```
marieflorescontact, line 173: I am not sure it is safe to remove user from req at this level because `createAuthenticatedContext` ( called just after) is used for non public stream/feed (despite his name...😅)
could we move the try catch bellow at this level?

ludovic: I don't see any consequence to this deletion in the code 🤔

opencti-platform/opencti-graphql/src/modules/dataSharing/feed-domain.ts
ludovic: I had kept in mind that we would do the migration; if we don't, it will cause problems. I'll remove those throws :+1:

opencti-platform/opencti-graphql/src/graphql/sseMiddleware.js
marieflorescontact: yes you're right

opencti-platform/opencti-graphql/src/modules/dataSharing/PUBLIC_USER_IMPERSONATION_PLAN.md
```diff
@@ -0,0 +1,131 @@
+# Plan — Public Data Sharing: User Impersonation (issue #14918)
```
marieflorescontact: do we really need this file in the PR? 🙂

opencti-platform/opencti-graphql/src/modules/dataSharing/streamCollection-domain.ts
```diff
@@ -22,6 +23,9 @@ export const createStreamCollection = async (context: AuthContext, user: AuthUse
...
+  }
```
marieflorescontact, line 29: should we add a check on user id (to be sure it exists)

marieflorescontact (approved):

PR #15073, opened 2026-03-19
[frontend] [Playbook] Component Title displays name and action (#14951)

### LINKED ISSUE #14951: [Playbook] Component Title displays name and action
As a user creating a component I need to see an accurate title of the component on the RHS so that I can modify the name and still know which component I am using and don’t think I am adding a new version. 

The right hand side 'Title' panel should also switch between Add and Update depending on if the user is adding a new one or updating an existing one.

Acceptance Criteria: The names used should be consistent when choosing from the Add components menu list and when the component is being created. 

Given: I am creating a new playbook
AND I have selected to add a component 
When the component is opened for editing
THEN I can see the title of the selected component in the title bar
AND the title bar displays that it is to be added

Given: I am editing a new playbook
When the component is opened for editing
THEN I can see the title of the selected component in the title bar
AND the title bar displays that it is to be Updated

SarahBocognano:
### Related issues

* closes https://github.com/OpenCTI-Platform/opencti/issues/14951


Archidoit (approved):

Archidoit:
Working well !! :)

opencti-platform/opencti-front/src/private/components/data/playbooks/playbookFlow/PlaybookFlowAddComponents.tsx
```diff
@@ -45,27 +45,40 @@ const PlaybookFlowAddComponents = ({
 }: PlaybookFlowAddComponentsProps) => {
   const { t_i18n } = useFormatter();
   const [componentId, setComponentId] = useState<string | null>(null);
```
lndrtrbn: Maybe we can remove this state if we store the full object now

opencti-platform/opencti-front/src/private/components/data/playbooks/playbookFlow/PlaybookFlowAddComponents.tsx
```diff
@@ -45,27 +45,40 @@ const PlaybookFlowAddComponents = ({
...
   };
 
   const isActionValid = action === 'config' || action === 'add' || action === 'replace';
   const hasSelection = selectedNode !== null || selectedEdge !== null;
   const open = isActionValid && hasSelection;
 
+  const isUpdate = action === 'config' && selectedNode?.type === 'workflow';
+  const componentName = selectedComponent?.name ?? selectedNode?.data?.component?.name;
+
+  const drawerTitle = `${isUpdate ? t_i18n('Update') : t_i18n('Add')} ${componentName} ${t_i18n('component')}`;
```
lndrtrbn: We should avoid concat translations like that because it works well in english but it is not garanted that every languages put the words in the same order. Instead use t_i18n with variables

lndrtrbn: For example in MalwareAnalysisDeletion.tsx
```
const deleteSuccessMessage = t_i18n('', {
    id: '... successfully deleted',
    values: { entity_type: t_i18n('entity_Malware-Analysis') },
  });
```

opencti-platform/opencti-front/lang/front/de.json
```diff
@@ -595,6 +597,7 @@
   "Completed tasks": "Abgeschlossene Aufgaben",
   "Completed works": "Abgeschlossene Arbeiten",
   "Complex": "Komplex",
+  "component": "bestandteil",
```
lndrtrbn: Seems to not be used anymore, you can remove it

lndrtrbn (approved):

PR #15080, opened 2026-03-19
[backend] validate filigran_chatbot_ai_cgu_status on settings edit (#15049)

### LINKED ISSUE #15049: Bad settingsEdit mutation can crash platform
## Description
A bad settingsEdit mutation change can crash the system.
It seems the value should be 'disabled' and not 'disable'.  As this is a toggle in the UI, should it not be a boolean, like the other ?

## Environment
1. 7.260309.0

## Reproducible Steps

1. `mutation DisableAI($id: ID!, $input: [EditInput]!) {
  settingsEdit(id: $id) {
    fieldPatch(input: $input) {
      id
      platform_ai_enabled  # boolean value
      filigran_chatbot_ai_cgu_status
    }
  }
}`

Vars

`{
		"id": "5ff70d99-a647-4d69-ad57-e3af454b58ee",
		"input": [
  		{
				"key": "filigran_chatbot_ai_cgu_status",
				"value": ["disable"]
 		  },
 		  {
   	    "key": "platform_ai_enabled",
				"value": ["false"] 
			}
    ]
}`

## Actual Output
Error on platform

JeremyCloarec:
### Proposed changes

* validate filigran_chatbot_ai_cgu_status value on settings edit
*

### Related issues

* fix #15049
*

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/domain/settings.js
+++ b/opencti-platform/opencti-graphql/src/domain/settings.js
@@ -226,7 +226,7 @@ export const settingsEditField = async (context, user, settingsId, input) => {
   const cguStatus = data.find((inputData) => inputData.key === 'filigran_chatbot_ai_cgu_status');
   if (cguStatus && cguStatus.value) {
     const validStatuses = Object.values(CguStatus);
-    if (!validStatuses.includes(cguStatus.value)) {
+    if (!validStatuses.includes(cguStatus.value[0])) {
       throw UnsupportedError(`Invalid CGU status, expected one of ${validStatuses.join(', ')}`);
     }
   }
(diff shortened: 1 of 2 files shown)


opencti-platform/opencti-graphql/src/domain/settings.js
```diff
@@ -250,6 +251,14 @@ export const settingsEditField = async (context, user, settingsId, input) => {
       }
     }
   }
+  const cguStatus = data.find((inputData) => inputData.key === 'filigran_chatbot_ai_cgu_status');
+  if (cguStatus && cguStatus.value) {
+    const validStatuses = Object.values(CguStatus);
+    if (!validStatuses.includes(cguStatus.value)) {
```
ludovic: cguStatus.value is actually an array
Because of this, the comparison `validStatuses.includes(cguStatus.value)` will always return false
You should use `cguStatus.value[0]` instead

ludovic (changes requested):
cguStatus.value is an array (defined as [Any]! in GraphQL), so validStatuses.includes will always fail and falsely reject valid inputs. We need to use cguStatus.value[0].

Also, could we add a positive test passing a valid enum value? Currently, the test passes purely by accident because every value triggers the error.

Archidoit:
Incorrect error message if we set several status 

image

aHenryJard (approved):
Tested locally, all good to me

ludovic (approved):

PR #15081, opened 2026-03-19
[backend] 'Original creation date' should be filterable for Cases (#15078)

### LINKED ISSUE #15078: Original creation date filter not available in Cases
## Description
'Original Creation date' filter is not available in IR, ROI or Requests for Takedowns.
This is available for reports but not in cases.

## Environment
1. SaaS 7.260318.0

## Expected Output
The problem is entities, events, reports, etc are often based on the Original Creation Date, which is not an option for filtering on. We need to be able to filter on the Original Creation Date to capture correct time frames based on when the event occurred.

## Actual Output
Not available to select

Archidoit:
### Proposed changes
'Original creation date' filter should be available for cases (case incident, case rfi, case rft).

[backend] The 'created' attribute is already defined for stix domain objects in the backend, no need to redefine it with isFilterable:false in the cases specific attributes definition.

### Related issues
#15078

### DIFF
--- a/opencti-platform/opencti-graphql/src/modules/attributes/stixCyberObservable-registrationAttributes.ts
+++ b/opencti-platform/opencti-graphql/src/modules/attributes/stixCyberObservable-registrationAttributes.ts
@@ -1,5 +1,5 @@
 import * as R from 'ramda';
-import type { AttributeDefinition } from '../../schema/attribute-definition';
+import { type AttributeDefinition, created } from '../../schema/attribute-definition';
 import { xOpenctiModifiedAt } from '../../schema/attribute-definition';
 import { schemaAttributesDefinition } from '../../schema/schema-attributes';
 import {
@@ -313,6 +313,7 @@ const stixCyberObservablesAttributes: { [k: string]: Array<AttributeDefinition> 
     { name: 'fingerprint_md5', label: 'Fingerprint MD5', type: 'string', format: 'short', mandatoryType: 'no', editDefault: false, multiple: false, upsert: true, isFilterable: true },
     { name: 'key_length', label: 'Key length', type: 'string', format: 'short', mandatoryType: 'no', editDefault: false, multiple: false, upsert: true, isFilterable: true },
     { name: 'comment', label: 'Comment', type: 'string', format: 'short', mandatoryType: 'no', editDefault: false, multiple: false, upsert: true, isFilterable: false },
+    created,
     { name: 'expiration_date', label: 'Expiration date', type: 'date', mandatoryType: 'no', editDefault: false, multiple: false, upsert: true, isFilterable: true },
   ],
   [ENTITY_AI_PROMPT]: [
--- a/opencti-platform/opencti-graphql/src/modules/attributes/stixCyberObservable-registrationAttributes.ts
+++ b/opencti-platform/opencti-graphql/src/modules/attributes/stixCyberObservable-registrationAttributes.ts
@@ -313,7 +313,6 @@ const stixCyberObservablesAttributes: { [k: string]: Array<AttributeDefinition> 
     { name: 'fingerprint_md5', label: 'Fingerprint MD5', type: 'string', format: 'short', mandatoryType: 'no', editDefault: false, multiple: false, upsert: true, isFilterable: true },
     { name: 'key_length', label: 'Key length', type: 'string', format: 'short', mandatoryType: 'no', editDefault: false, multiple: false, upsert: true, isFilterable: true },
     { name: 'comment', label: 'Comment', type: 'string', format: 'short', mandatoryType: 'no', editDefault: false, multiple: false, upsert: true, isFilterable: false },
-    { name: 'created', label: 'Created', type: 'date', mandatoryType: 'no', editDefault: false, multiple: false, upsert: true, isFilterable: false },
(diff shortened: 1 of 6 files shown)


opencti-platform/opencti-graphql/src/modules/attributes/stixCyberObservable-registrationAttributes.ts
```diff
@@ -313,7 +313,6 @@ const stixCyberObservablesAttributes: { [k: string]: Array<AttributeDefinition>
     { name: 'fingerprint_md5', label: 'Fingerprint MD5', type: 'string', format: 'short', mandatoryType: 'no', editDefault: false, multiple: false, upsert: true, isFilterable: true },
     { name: 'key_length', label: 'Key length', type: 'string', format: 'short', mandatoryType: 'no', editDefault: false, multiple: false, upsert: true, isFilterable: true },
     { name: 'comment', label: 'Comment', type: 'string', format: 'short', mandatoryType: 'no', editDefault: false, multiple: false, upsert: true, isFilterable: false },
-    { name: 'created', label: 'Created', type: 'date', mandatoryType: 'no', editDefault: false, multiple: false, upsert: true, isFilterable: false },
```
JeremyCloarec, line 316: Unless I missed something, we need to keep tihs for sitxCyberObservables no? Since SCO are not related to domain objects, they do not inherit 'created' from stixDomainObject attributes

opencti-platform/opencti-graphql/src/modules/case/case-incident/case-incident.ts
```diff
@@ -36,7 +36,6 @@ const CASE_INCIDENT_DEFINITION: ModuleDefinition<StoreEntityCaseIncident, StixCa
     { key: 'notes', width: 12, label: 'Notes about this entity' },
   ],
   attributes: [
-    { name: 'created', label: 'Created', type: 'date', mandatoryType: 'external', editDefault: true, multiple: false, upsert: true, isFilterable: false },
```
JeremyCloarec, line 39: It's true that created exists on stixDomainObhect attributes, but it's defined a bit different here: we have mandatoryType : 'external' instead of 'no' on SDO definition, and editDefault: true instead of false on SDO.
To keep the same behavior, I suggest keeping it as:
`{...created, mandatoryType: 'external', editDefault: true}` instead

Archidoit: I thought it was a mistake since isFilterable was not correctly defined... That's why I aligned on the main 'created' for sdo. But maybe you're right.

opencti-platform/opencti-graphql/src/modules/notification/notification.ts
```diff
@@ -35,7 +35,7 @@ const TRIGGER_DEFINITION: ModuleDefinition<StoreEntityTrigger, StixTrigger> = {
   attributes: [
     { name: 'name', label: 'Name', type: 'string', format: 'short', mandatoryType: 'external', editDefault: true, multiple: false, upsert: false, isFilterable: true },
     { name: 'description', label: 'Description', type: 'string', format: 'text', mandatoryType: 'no', editDefault: false, multiple: false, upsert: false, isFilterable: true },
-    { name: 'created', label: 'Created', type: 'date', mandatoryType: 'external', editDefault: false, multiple: false, upsert: false, isFilterable: false },
+    created,
```
JeremyCloarec: Same remark as above, created definition is a bit different here. Also, not sure if we want to make it filterable for notifications? 
I suggest using:
`{...created, mandatoryType: 'external', isFilterable: false, upsert: false}`

opencti-platform/opencti-graphql/src/modules/notifier/notifier.ts
```diff
@@ -19,7 +19,7 @@ const NOTIFIER_DEFINITION: ModuleDefinition<StoreEntityNotifier, StixNotifier> =
   },
   attributes: [
     { name: 'name', label: 'Name', type: 'string', format: 'short', mandatoryType: 'internal', editDefault: false, multiple: false, upsert: false, isFilterable: true },
-    { name: 'created', label: 'Created', type: 'date', mandatoryType: 'external', editDefault: false, multiple: false, upsert: true, isFilterable: false },
+    created,
```
JeremyCloarec: Same remark as above, created definition is a bit different here. Also, not sure if we want to make it filterable for notifiers? 
I suggest using:
`{...created, mandatoryType: 'external', isFilterable: false}`

opencti-platform/opencti-graphql/src/modules/notification/notification.ts
Archidoit: It's not really notification though the file name is confusing, it's triggers :) Since 'original creation date' is a column of the triggers lists, it's worth having a filter on it : 

image

opencti-platform/opencti-graphql/src/modules/attributes/stixCyberObservable-registrationAttributes.ts
Archidoit: you're right!

opencti-platform/opencti-graphql/src/modules/notification/notification.ts
JeremyCloarec: Okay that's true :) We only need to keep the mandatoryType and upsert then

JeremyCloarec (approved):

PR #15025, opened 2026-03-18
[backend] feat: add public_user_id and public_authorized_markings to data sharing entities (#14918)

ludovic:
### Proposed changes

* Add `taxii_public_user_id` field to `TaxiiCollection` (ES attributes, GraphQL schema, TS types)
* Add `feed_public_user_id` field to `Feed` (ES attributes, GraphQL schema, TS types)
* Add `stream_public_user_id` field to `StreamCollection` (ES attributes, GraphQL schema, TS types)

### Related issues

* Fixes #14918

### Checklist

- [x] I consider the submitted work as finished
- [ ] I tested the code for its functionality
- [ ] I wrote test cases for the relevant uses case (coverage and e2e)
- [ ] I added/update the relevant documentation (either on github or on notion)
- [ ] Where necessary I refactored code to improve the overall quality

### Further comments

This PR is the first part (data model only) of a multi-chunk fix for the public data sharing authorization bypass.

The fix introduces a user impersonation model similar to `publicDashboard`: instead of using `SYSTEM_USER` for public access, each public feed/collection stores a reference to a real user account (`*_public_user_id`) whose permissions will be used when serving public requests.

Subsequent PRs will cover:
- **Chunk B**: HTTP handlers (`httpTaxii`, `httpRollingFeed`, `sseMiddleware`) + domain validation
- **Chunk C**: Capability check (`SETTINGS_SETACCESSES`) + frontend UI
- **Chunk D**: Warning on deletion of the impersonated user
- **Chunk E**: Migration (sets `*_public_user_id = SYSTEM_USER.id` for existing public entities)

PR #15157, opened 2026-03-24
[backend/frontend] UI for public user in Public Data Sharing - Chunk C (#14918)

### LINKED ISSUE #14918: Improve data sharing in public mode

ludovic:
### Proposed changes

**Backend**

- Added *_public_user: Creator GraphQL fields to `taxiiCollection.graphql`, `streamCollection.graphql `and `feed.graphql,` resolved via loadCreator in the corresponding resolvers

 **Frontend**

- Added a `CreatorField` (conditional on *_public: true) to the creation forms for TAXII collections, CSV feeds and live streams

- Added a `CreatorField` to the edition forms for all three, pre-populated with the existing user via `convertUser`

- Made `*_public_user_id` required in Yup validation when `*_public: true`

- Updated Relay fragments to include the resolved `*_public_user { id, entity_type, name }` sub-object (replacing the raw ID fetch)

### Related issues

* Fixes #14918

### Checklist

### Further comments
Depends on
#15025 (Chunk A — data model)
#15048 (Chunk B — HTTP handler security)


opencti-platform/opencti-front/src/private/components/data/stream/StreamCollectionCreation.tsx
```diff
@@ -166,6 +172,14 @@ const StreamCollectionCreationForm = ({
                   name="authorized_members"
                 />
               )}
+              {values.stream_public && (
+                <CreatorField
+                  name="stream_public_user_id"
+                  label={t_i18n('User used to retrieve data')}
```
marieflorescontact: this needs to be added in translation (well, as soon as it will be validated by product off course 🙂)

marieflorescontact (commented):
locally tested ok, needs product validation

marieflorescontact (approved):

PR #15193, opened 2026-03-26
[backend] Changed vulnerabilities to vulnerability in security coverage vocabulary (#14676)

### LINKED ISSUE #14676: [backend] Backend inconsistency between "vulnerability" and "vulnerabilities" in openAEV and openCTI STIX bundles
## Description

GraphQL OCTI API **does not return vulnerability scores** due to expectations naming mismatch : vulnerability on openAEV side vs vulnerabilities in openCTI side.

The decision is to fix it on the OCTI side because the other types are singular (prevention, detection).

Image

## Expected Output

To be able to get the vulnerabilities scores on the openCTI side : 

Image

MTorbay-Filigran:
### Proposed changes

* Changed vulnerabilities to vulnerability in security coverage vocabulary
* Added a migration to do it in db

### Related issues

* [#14676 ](https://github.com/OpenCTI-Platform/opencti/issues/14676)

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/migrations/1774531353818-rename-coverage-vocabulary.js
+++ b/opencti-platform/opencti-graphql/src/migrations/1774531353818-rename-coverage-vocabulary.js
@@ -18,8 +18,8 @@ export const up = async (next) => {
         value: ['Vulnerability'],
       },
     ];
-    await editVocabulary(context, SYSTEM_USER, elementToChange.id, input);
     logMigration.info(`${message} > changing 'vulnerabilities' to 'vulnerability'.`);
+    await editVocabulary(context, SYSTEM_USER, elementToChange.id, input);
   }
   logMigration.info(`${message} > done.`);
   next();
--- a/opencti-platform/opencti-graphql/src/migrations/1774454302668-change-coverage-vocabulary.js
+++ b/opencti-platform/opencti-graphql/src/migrations/1774454302668-change-coverage-vocabulary.js
@@ -1,0 +1,30 @@
+import { executionContext, SYSTEM_USER } from '../utils/access';
+import { logMigration } from '../config/conf';
+import { editVocabulary, findById } from '../modules/vocabulary/vocabulary-domain';
+
+const message = '[MIGRATION] Rename coverage_ov vulnerabilities to vulnerability in vocabulary';
+export const up = async (next) => {
+  logMigration.info(`${message} > started`);
+  const context = executionContext('migration');
+  const elementToChange = await findById(context, SYSTEM_USER, 'vocabulary--836e90eb-4d7e-5747-88ef-1405dba8d471'); // 'vulnerabitilies' vocabulary id
+  if (elementToChange) {
+    const input = [
+      {
+        key: 'name',
+        value: ['vulnerability'],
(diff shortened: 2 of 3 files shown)


opencti-platform/opencti-graphql/src/migrations/1774531353818-rename-coverage-vocabulary.js
```diff
@@ -0,0 +1,30 @@
...
+        key: 'name',
+        value: ['vulnerability'],
+      },
+      {
+        key: 'description',
+        value: ['Vulnerability'],
+      },
+    ];
+    await editVocabulary(context, SYSTEM_USER, elementToChange.id, input);
+    logMigration.info(`${message} > changing 'vulnerabilities' to 'vulnerability'.`);
```
SouadHadjiat, line 21: small detail : I suggest to put this log above just before "editVocabulary"

opencti-platform/opencti-graphql/src/migrations/1774531353818-rename-coverage-vocabulary.js
```diff
@@ -0,0 +1,30 @@
+import { executionContext, SYSTEM_USER } from '../utils/access';
```
SouadHadjiat, line 1: suggestion of migration name "rename-coverage-vocabulary".
Also you can update the timestamp to today's date.

opencti-platform/opencti-graphql/src/migrations/1774531353818-rename-coverage-vocabulary.js
```diff
@@ -0,0 +1,30 @@
...
+import { logMigration } from '../config/conf';
+import { editVocabulary, findById } from '../modules/vocabulary/vocabulary-domain';
+
+const message = '[MIGRATION] Rename coverage_ov vulnerabilities to vulnerability in vocabulary';
+export const up = async (next) => {
+  logMigration.info(`${message} > started`);
+  const context = executionContext('migration');
+  const elementToChange = await findById(context, SYSTEM_USER, 'vocabulary--836e90eb-4d7e-5747-88ef-1405dba8d471'); // 'vulnerabitilies' vocabulary id
+  if (elementToChange) {
```
aHenryJard, line 10: If there is no element maybe we should add it ? Or else OpenAEV integration will not work I believe no ?

aHenryJard: unless there is already vulnerability but how to know that... I think it cannot happens that it's not existing yet right ?

SouadHadjiat (approved):

opencti-platform/opencti-graphql/src/migrations/1774531353818-rename-coverage-vocabulary.js
MTorbay-Filigran: Normally with the fix, when you restore the db you will have the correct vocab 'vulnerability', and for the previous instances you will only have 'vulnerabilities' unless the user changed it manually on the UI (and yes in this specific case, it would not take it, but we cannot find the term the user replaced vulnerabilities)

PR #15199, opened 2026-03-26
[frontend/backend] Customize entity type names (#14873)

### LINKED ISSUE #14873: feat(customization): allow administrators to customize the display name of core entities

fellowseb:
### Proposed changes

This is a follow-up to this PR: https://github.com/OpenCTI-Platform/opencti/pull/14894.

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/14873
* Fixes https://github.com/OpenCTI-Platform/opencti/issues/14873

### Checklist


opencti-platform/opencti-front/src/private/components/threats/intrusion_sets/IntrusionSetDeletion.tsx
```diff
@@ -55,7 +55,7 @@ const CampaignDeletion = ({ id, isOpen, handleClose }: { id: string; isOpen: boo
       submitDelete={submitDelete}
       isOpen={isOpen}
       onClose={handleClose}
-      message={t_i18n('Do you want to delete this intrusion set?')}
+      message={t_i18n('Do you want to delete this') + ' ' + translateEntityType('Intrusion-Set') + '?'}
```
fellowseb, line 58: Concatenations like this don't work in all languages. Instead we want to inject params using `t_i18n` 's options.

PR #15212, opened 2026-03-26
[backend] Validate workflow schema on save (#14578)

### LINKED ISSUE #14578: feat(workflow): add workflow schema validation on save
# Why

When a workflow definition is saved, its structure should respect a specific schema.

# What

Add a validation layer on a `WorkflowDefinitionSet` query checking the following rules:

- OK
    - multiple from in transitions
    - one to in transitions

- NOT OK
    - multiple to in transitions
    - status name / transition name / field / value length > 255
    - transition's from/to or initialState referencing a status that doesn't exist
        - in the workflow definition
        - in the status templates in DB
    - condition type that doesn't exist
    - side effect that doesn't exist
    - condition operator that doesn't exist
    - same event name referenced in multiple transitions
    - entityType that doesn't exist
    - workflow id that already exists
    - incompatible action/mode pair? (some may only be async or sync)

- Additional 
    - validate action params with a schema
    - Workflow schema validation in general
    - validate condition field against a list of supported fields
    - prevent status template deletion if used in a workflow
    - if defining a workflow on entityType DraftWorkspace, then there should be at least one transition with a validateDraft action

# How
- create a src/modules/workflow/workflow-validation.ts file
- add required `zod` schemas and validation helper functions
- run a check with them on a workflow definition save
- add tests

OctaveLaventure:
### Proposed changes

* add validation for workflow schema save

### Related issues

* closes #14578 

### Checklist


opencti-platform/opencti-graphql/src/modules/workflow/workflow-validation.ts
```diff
@@ -0,0 +1,182 @@
...
+          throw ValidationError('Condition value length > 255');
+        }
+      }
+    }
+
+    if (transition.actions) {
+      for (const action of transition.actions) {
+        validateAction(action, `transition ${transition.event}`);
+        if (action.type === 'validateDraft') {
+          hasValidateDraft = true;
```
frapuks, line 214: We must to check transitions.actions but also status.onEnter and status.onExit.

I have this simple workflow.
- Nothing in the transition
- `Validate draft` action on enter the last status
=> It will be rejected on save, even if i have an `validate draft` action

Image

Image

delemaf: For now we'll only allow validateDraft on transitions (we might remove from onEnter/onClose actions)

OctaveLaventure: as we discussed we need to remove draft validation from onExit/onEnter

delemaf (approved):

PR #15229, opened 2026-03-27
[frontend] fix x_opencti_detection initial value (#15084)

### LINKED ISSUE #15084: Playbook "Manipulate Knowledge" does not serialize boolean value on initial component creation when switch is left at default (false)
## Description

When creating a "Manipulate Knowledge" playbook component with Action type "Replace" and Field "Detection" (`x_opencti_detection`), leaving the Value switch at its default position (unchecked / false) results in the `value` key being omitted entirely from the saved component configuration. The backend executor receives `undefined` from `R.head(action.value)` and the replace operation silently does nothing.

The issue only occurs on initial component creation when the switch is never toggled. If the user toggles the switch to True, saves, then toggles back to False and saves again, the value serializes correctly.

## Environment
6.9.26

## Reproducible Steps

1. Navigate to **Data > Processing > Automation** (Playbooks)
2. Create a new playbook or open an existing one
3. Add a **"Manipulate Knowledge"** component
4. Set **Action type** to `Replace`
5. Set **Field** to `Detection` (`x_opencti_detection`)
6. **Leave the Value switch in its default position** (unchecked / false) -- do not toggle it
7. Click **"Create"**
8. Export the playbook configuration as JSON and inspect the component

## Expected Output

The saved component configuration includes the `value` key with the string `"false"`:

```json
{
  "op": "replace",
  "attribute": "x_opencti_detection",
  "value": "false"
}
```

## Actual Output

The `value` key is missing entirely from the saved configuration:

```json
{
  "op": "replace",
  "attribute": "x_opencti_detection"
}
```

The backend executor calls `R.head(action.value)` on `undefined`, and the replace operation silently does nothing. No error is raised.

## Workaround

Toggle the switch to **True**, save the component, then toggle back to **False** and save again. After this round-trip, the `value` key is correctly serialized in the configuration.

## Root Cause Analysis

Two contributing factors combine to produce this bug:

### Primary: SwitchField does not initialize form value on mount

**File:** `SwitchField.jsx`

`SwitchField` only calls `setFieldValue` inside its `onChange` handler. When no `initialValue` prop is provided and the user never toggles the switch, `onChange` never fires and `setFieldValue` is never called. The `value` key is never set in the Formik form state.

### Secondary: Validation bypass masks the missing value

**File:** `PlaybookFlowFieldActions.tsx`

The validation logic includes a special case for `x_opencti_detection` that bypasses the normal requirement for a non-empty `value` field. This allows the component configuration to save successfully despite the `value` key being absent from the form state.

### Execution path

**File:** `PlaybookActionValueField.tsx`

This file renders the `SwitchField` for boolean attributes like `x_opencti_detection`. It does not pass an explicit `initialValue` prop to `SwitchField`, relying on the switch's visual default state (unchecked) to represent `false`. However, the visual state and the form state are disconnected until `onChange` fires.

### Summary of affected files

| File | Role in bug |
|---|---|
| `SwitchField.jsx` | Does not initialize form value on mount when no `initialValue` is provided |
| `PlaybookFlowFieldActions.tsx` | Validation bypass for `x_opencti_detection` allows save without a value |
| `PlaybookActionValueField.tsx` | Renders the SwitchField without passing an explicit `initialValue` prop |

## Related Issues

| Issue | Status | Relationship |
|---|---|---|
| #11481 | Closed | Similar symptom (missing boolean value) but in the edit/load code path, not the creation path. Different root cause. |
| #14744 | Open | Playbook-related bug but different root cause (backend bundle contamination). Not a duplicate. |

**Prior support ticket:** OCTI1-2190 reported the same symptom from a different customer.

### Option A: Initialize form value on SwitchField mount (Recommended)

Add a `useEffect` in `SwitchField.jsx` that calls `setFieldValue(name, initialValue ?? false)` on mount. This ensures the form state always reflects the visual state of the switch, regardless of whether the user interacts with it.

**Scope:** `SwitchField.jsx`
**Risk:** Low. Aligns form state with visual state. May need regression testing on other forms that use SwitchField.

### Option B: Pass explicit initialValue from PlaybookActionValueField

Update `PlaybookActionValueField.tsx` to pass `initialValue={false}` (or `initialValue="false"`) to the `SwitchField` when rendering boolean attributes. This is a narrower fix scoped to the playbook creation form.

**Scope:** `PlaybookActionValueField.tsx`
**Risk:** Low, but only fixes the playbook creation path. Other consumers of SwitchField may have the same latent bug.

### Option C: Remove the validation bypass

Remove or tighten the special-case validation for `x_opencti_detection` in `PlaybookFlowFieldActions.tsx` so that a missing `value` key blocks the save. This would surface the error to the user at creation time rather than silently producing a broken configuration.

**Scope:** `PlaybookFlowFieldActions.tsx`
**Risk:** Medium. Would cause a validation error if Option A or B is not also applied, since the form state would still be empty. Best applied in combination with Option A or B.

**Recommended approach:** Option A (addresses the root cause globally) combined with Option C (defense in depth).

## Additional Information

This bug affects any boolean field rendered by SwitchField in the playbook "Manipulate Knowledge" creation form where the user accepts the default value without toggling. The `x_opencti_detection` field is the most commonly reported case, but the pattern could affect other boolean attributes if they are added to the playbook action form in the future.

JeremyCloarec:
### Proposed changes

* fix x_opencti_detection action initial value
*

### Related issues

* fix #15084
*

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/data/playbooks/playbookFlow/playbookFlowFields/playbookFlowFieldsActions/PlaybookActionValueField.tsx
+++ b/opencti-platform/opencti-front/src/private/components/data/playbooks/playbookFlow/playbookFlowFields/playbookFlowFieldsActions/PlaybookActionValueField.tsx
@@ -160,6 +160,7 @@ const PlaybookActionValueField = ({
           type="checkbox"
           name={formValueName}
           label={t_i18n('Value')}
+          initialValue={action.value?.[0]?.value === 'true'}
           onChange={(_: string, detection: string) => {
             setFieldValue(
               valueName,


CelineSebe (approved):

CelineSebe:
Tested locally 👌

PR #15281, opened 2026-03-31
[backend] normalize stix pattern(#13803)

### LINKED ISSUE #13803: STIX Pattern deduplication issue when using apostrophes in Object Paths
## Description

We are facing a deduplication issue related to the use of apostrophes in STIX Pattern object paths.

According to the STIX Patterning Language documentation, it appears that object paths can be written with or without apostrophes, and that both syntaxes are seemingly valid. However, in the current implementation, patterns that are semantically identical but differ only by the presence or absence of apostrophes are not deduplicated.

As a result, two logically equivalent STIX patterns can coexist in the platform, leading to inconsistencies and duplicate indicators.

Image

Image

## Environment

OCTI 6.9.2

## Reproducible Steps

1. Create an Indicator with "Main Observable Type = File" and Pattern = [file:hashes.'MD5' = 'e1d9c90de2568f34ad689c71dac09c62']
2. Create an Indicator with "Main Observable Type = File" and Pattern = [file:hashes.MD5 = 'e1d9c90de2568f34ad689c71dac09c62']

-> No deduplication

## Expected Output

STIX Patterns that are semantically equivalent should be deduplicated, regardless of whether apostrophes are used in object paths.

A normalization step could be applied to object paths before deduplication to ensure consistency.

## Actual Output

Deduplication logic treats patterns with and without apostrophes in object paths as different.

This results in duplicate indicators for the same logical condition.

marieflorescontact:
### Proposed changes

* Adds `normalizePropertyName `method
* refactor `cleanupIndicatorPattern`
* rename  `grabInterestingTokens` =>  `collectNormalizedTokens`
*  FIX : Normalize .'MD5' → .MD5 to avoid duplication
(Keeps  .'SHA-256' → .'SHA-256'  because according to STIX grammar, the hyphen cannot be used without quotation marks)
* Adds unit tests

### Related issues

* closes: https://github.com/OpenCTI-Platform/opencti/issues/13803

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/utils/syntax.js
+++ b/opencti-platform/opencti-graphql/src/utils/syntax.js
@@ -11,11 +11,10 @@ export const STIX_PATTERN_TYPE = 'stix';
 
 const unflatten = (data) => {
   const result = {};
-  // eslint-disable-next-line no-restricted-syntax,guard-for-in
+
   for (const i in data) {
     const keys = i.split('.');
     keys.reduce((r, e, j) => {
-      // eslint-disable-next-line no-nested-ternary,no-param-reassign,no-return-assign
       return r[e] || (r[e] = Number.isNaN(Number(keys[j + 1])) ? (keys.length - 1 === j ? data[i] : {}) : []);
     }, result);
   }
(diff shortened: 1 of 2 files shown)


ludovic (commented):
Good work on the normalization! Two questions:

Migration: existing indicators with .'MD5' in the database won't be renormalized automatically. Is a migration script planned to fix historical data, or is this considered out of scope?

Edit path: In `indicatorEditField`, the return value of `validateIndicatorPattern` is not used (the formattedPattern is discarded). This means editing an existing indicator's pattern would still store the non-normalized version. Should this be addressed here or in a follow-up?

marieflorescontact:
> This means editing an existing indicator's pattern would still store the non-normalized version

thanks for the review 🙂. good catch for the migration. asking product team if we want to merge already duplicated indicators with migration.

marieflorescontact:
No migration needed @ludovic

ludovic (approved):

PR #15293, opened 2026-03-31
[backend] Reorganize integration test helpers to get more coverage (#15292)

### LINKED ISSUE #15292: Improve backend integration test helpers to get coverage
Currently some backend integration tests use helpers that make HTTP calls to the main execution content to execute business logic there. That's an issue as that process is out of reach of the coverage. Instead we need to make sure the business logic (domain handlers) run in the test workers context.

fellowseb:
### Problem

We realized that some backend resolvers we not taken into account in the code coverage because the test cases are using helpers that make queries on the "Init Platform" set up during the `globalSetup` phase.
While there is a helper to query a lightweight server in the test worker it is not super well documented nor used.

### Proposed changes

* Move all query helpers that target the local lightweight server in `testQueryHelper.ts`
* Rename function to have a consistent convention: `queryAs....` for local helpers, `queryInitPlarformAs...` for helpers targeting the Init Platform
* Edit integration tests to use the local helpers when possible
* Add comments to explain the situation and the difference between the two sets of helpers
* Removed usage of axios/http clients in use cases. Treat those as implementation details. Will make refactorings easier.

Bonus side-effect:
* Increases the coverage slightly due to newly covered resolvers

Limitations:

* Some tests (draft or auth related) rely on express middleware logic to work correctly. Those keep using the `queryInitPlatform` helpers for now.

### Related issues

* Fixes https://github.com/OpenCTI-Platform/opencti/issues/15292

### Checklist

### Further comments

The next step is to investigate launching an HTTP server closer to the real one in local test workers so we can include middleware logic and thus use local helpers everywhere in test cases; being careful about the time to spawn that server.


aHenryJard (approved):
Thank you for this huge work.

PR #15314, opened 2026-04-01
[backend/frontend] User deletion and Public sharing - Chunk D (#14918)

### LINKED ISSUE #14918: Improve data sharing in public mode

ludovic:
### Proposed changes

*This chunk handles cleaning up public configs when a user is deleted and fixes UX issues in the edition forms.

**Backend**
 - Cleanup on user deletion: Added disablePublicSharingForDeletedUser inside userDelete. It automatically sets *_public = false (and stream_live = false) on any Feed, TAXII, or Stream using the deleted user, and clears *_public_user_id. 
 - Fix "System" UI display: *_public_user resolvers now properly return null instead of falling back to SYSTEM_USER when *_public_user_id is empty.
 - Edition validation: Backend now strictly rejects passing *_public = true during fieldPatch mutations if no *_public_user_id is defined. 

**Frontend**
 - Controlled Public Switch: Stream and TAXII edition forms now use controlled Formik state for the public switch.
 - Batched Save on Enable: Toggling public ON merely shows the user selector. Selecting a user then triggers a single mutation sending both *_public_user_id and *_public = true.
- Immediate Save on Disable: Toggling OFF disables public sharing immediately.
- Delete Warning: Added an alert in UserDeletionDialog.tsx warning that associated public feeds and streams will be disabled.

### Related issues

* Fixed #14918

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/modules/dataSharing/dataSharing-utils.ts
+++ b/opencti-platform/opencti-graphql/src/modules/dataSharing/dataSharing-utils.ts
@@ -1,9 +1,14 @@
 import { getEntitiesMapFromCache } from '../../database/cache';
 import { INTERNAL_USERS, SYSTEM_USER } from '../../utils/access';
 import { ENTITY_TYPE_USER } from '../../schema/internalObject';
-import { FunctionalError } from '../../config/errors';
+import { DatabaseError, FunctionalError } from '../../config/errors';
 import { logApp } from '../../config/conf';
 import type { AuthContext, AuthUser } from '../../types/user';
+import { elRawUpdateByQuery } from '../../database/engine';
+import { READ_INDEX_INTERNAL_OBJECTS } from '../../database/utils';
+import { ENTITY_TYPE_FEED } from './feed-types';
+import { ENTITY_TYPE_TAXII_COLLECTION } from './taxiiCollection-types';
+import { ENTITY_TYPE_STREAM_COLLECTION } from './streamCollection-types';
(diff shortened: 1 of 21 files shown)


opencti-platform/opencti-graphql/src/modules/dataSharing/dataSharing-utils.ts
```diff
@@ -40,3 +45,50 @@ export const resolvePublicUser = async (context: AuthContext, userId: string | n
...
+/**
+ * When a user is deleted, disables all public sharing entities (CSV feed, TAXII collection, live stream)
+ * that referenced that user as their public impersonation user.
+ * The public flag is set to false, stream_live is set to false for streams, and the user id field is cleared.
+ */
+export const disablePublicSharingForDeletedUser = async (context: AuthContext, userId: string): Promise<void> => {
+  await Promise.all(
+    PUBLIC_SHARING_CONFIGS.map(({ entityType, publicField, userIdField, liveField }) => {
+      const liveScript = liveField ? ` ctx._source.${liveField} = false;` : '';
+      return elRawUpdateByQuery({
```
marieflorescontact: We need to check with product team if we want to have an activity log for this action (this means that in that case we we need to use `publishUserAction()` method)

marieflorescontact (commented):
functionnaly tested ok.

marieflorescontact (approved):
looks good to me 🙂

PR #15331, opened 2026-04-01
[backend] Introduce @ff GraphQL directive to restrict access based on feature flag value (#15330)

### LINKED ISSUE #15330: Introduce new GraphQL directive to restrict access to endpoint based on Feature Flag
Introduce new GraphQL directive to restrict access to endpoint based on Feature Flag values.

Example:
```
type Query {
  doSomething: String! @ff(flags: ["SOME_FLAG", "SOME_OTHER_FLAG"])
}
```

This definition will make queries to `doSomething` throw neither one of the feature flags is set.

Optionally introduce a `softFail` option that will resolve to `null` instead of returning an error:
```
type Query {
  doSomething: String @ff(flags: ["SOME_FLAG", "SOME_OTHER_FLAG"], softFail: true)
}
```

This can be helpful for queries happening super early, even before (or while) retrieving the FFs in the frontend.

fellowseb:
### Proposed changes

* Introduces new GraphQL directive to restrict access to endpoint based on Feature Flag values.

Example:
```
type Query {
  doSomething: String! @ff(flags: ["SOME_FLAG", "SOME_OTHER_FLAG"])
}
```
This definition will make queries to `doSomething` throw if neither one of the feature flags is set.

* Adds a `softFail` option that will resolve to `null` instead of returning an error:

```
type Query {
  doSomething: String @ff(flags: ["SOME_FLAG", "SOME_OTHER_FLAG"], softFail: true)
}
```

This can be helpful for queries happening super early, even before (or while) retrieving the FFs in the frontend.
Is used for the custom views feature as the new `customViewsDisplayContext` query is used in the `RootPrivateQuery` that itself fetches the FFs.

* Does a little refactoring of the `authDirective`
* Due to `Dual Package Hazard` issues happening when running vitest in the backend I had to add an `alias` to force using the CJS version.

### Related issues

* Fixes https://github.com/OpenCTI-Platform/opencti/issues/15330

### Checklist

### Further comments

TODO if/when merged:

- Update docs in Notion
- Update MD files

### DIFF
--- a/opencti-platform/opencti-front/src/schema/relay.schema.graphql
+++ b/opencti-platform/opencti-front/src/schema/relay.schema.graphql
@@ -8,6 +8,8 @@ directive @allowUnlicensedLTS on OBJECT | FIELD_DEFINITION
 
 directive @constraint(minLength: Int, maxLength: Int, startsWith: String, endsWith: String, notContains: String, pattern: String, format: String, min: Int, max: Int, exclusiveMin: Int, exclusiveMax: Int, multipleOf: Int) on INPUT_FIELD_DEFINITION
 
+directive @ff(flags: [String!]!, softFail: Boolean = false) on FIELD_DEFINITION
+
 """Controls the rate of traffic."""
 directive @rateLimit(
   """Number of occurrences allowed over duration."""
--- a/opencti-platform/opencti-graphql/config/schema/opencti.graphql
+++ b/opencti-platform/opencti-graphql/config/schema/opencti.graphql
@@ -21,6 +21,7 @@ directive @constraint(
   exclusiveMax: Int
   multipleOf: Int
 ) on INPUT_FIELD_DEFINITION
+directive @ff(flags: [String!]!, softFail: Boolean = false) on FIELD_DEFINITION
 
 ### SCALAR
 
--- a/opencti-platform/opencti-graphql/script/script-generate-schema.js
+++ b/opencti-platform/opencti-graphql/script/script-generate-schema.js
@@ -3,7 +3,7 @@
 
 import fs from 'node:fs';
 import _ from '../src/modules/index';
-import { printSchema } from 'graphql/utilities';
+import { printSchema } from 'graphql';
 import createSchema from '../src/graphql/schema';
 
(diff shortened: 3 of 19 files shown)


lndrtrbn (commented):
I think it would be preferable, for easy review and maintenance, to avoid refactoring stuff not related to a new feature in the same PR but make a dedicated one instead _(referring to the authDirective)_.

Do you think it could be interesting to also add an option that allows to change the "all flags must be present" to "at least one"?

fellowseb:
> Do you think it could be interesting to also add an option that allows to change the "all flags must be present" to "at least one"?

I don't really see the case where you would want all flags to be present TBH 🤔. This can be added later if people have the need.

> I think it would be preferable, for easy review and maintenance, to avoid refactoring stuff not related to a new feature in the same PR but make a dedicated one instead _(referring to the authDirective)_.

Removed the refactoring.

lndrtrbn (approved):

fellowseb:
@SouadHadjiat you mentioned using `UnsupportedError` instead of `ForbiddenAccess` : I still think `ForbiddenAccess` makes more sense given it's a client error (403) whereas `UnsupportedError` is a server error. `UnsupportedError` is used for cases where the server doesn't _support_ cases either because of domain logic (`'Can apply avg on non numeric attribute'`) or other reasons independant from the user.

fellowseb:
Copilot's opinion:

> Great implementation! Regarding error handling for feature flag access restrictions, here are my recommendations:
> 
> **Error Name:** `ForbiddenAccess` (which you're already using ✓)
> 
> **HTTP Status Code:** `403 Forbidden`
> 
> **Rationale:**
> - **403 Forbidden** is the correct HTTP status code when a resource is inaccessible due to insufficient permissions or access restrictions. Since feature flags determine access eligibility, they function as permission gates.
> - The error name `ForbiddenAccess` accurately reflects that access is denied, distinguishing it from:
>   - `401 Unauthorized` (authentication failed - user isn't authenticated)
>   - `404 Not Found` (resource doesn't exist)

lndrtrbn (approved):

PR #15343, opened 2026-04-02
[frontend/backend] Display custom views tab(s) (#13389)

### LINKED ISSUE #13389: [Custom view] Introduce custom views

fellowseb:
### Proposed changes

Implements chunk `Users can view a and navigate to a "Custom views" tab on entities pages` of the ["Custom views" feature](https://github.com/OpenCTI-Platform/opencti/issues/13389)

A custom view is a dashboard an administrator creates for a given entity type. This dashboard is then viewable (not editable) by "normal" users.

#### User journeys

* Case 1: given there is one Custom views in DB for a specific Entity Type (ex.: Intrusion-Set); a participating user with the correct feature flag will see a  new tab appear on the SDO page with the name of the custom view. Clicking on the tab reroutes the user to the custom view `path` (`/[slug]-[id-without-dashes]`) and displays its content.
* Case 2: given there are multiple Custom views in DB for a specific Entity Type (ex.: Intrusion-Set); a participating user with the correct feature flag will see a `Custom views` tab appear on the SDO page. Clicking on the tab displays a drop-down menu with an item per custom view. Clicking on the menu items reroutes the user to the custom view `path` (`/[slug]-[id-without-dashes]`) and displays its content.

https://github.com/user-attachments/assets/a0688e0a-11d1-44b0-a9c0-2bc3ef3491d8 

#### Specs

Figma: https://www.figma.com/design/K2TFfOTI4MBgWzIyGT2AHV/Custom-views?node-id=119-150&t=y1ubGmkFiRj9H8fl-0
Notion: https://www.notion.so/filigran/New-TAB-with-a-custom-knowledge-view-2b18fce17f2a80f09a34db26a15ff8d1

##### Frontend

* Created a `useTabWithDropDown` hook that returns two components separately, for case 2: the returned Tab and the DropDown. Seems a little awkward but I had to do that given returning either a Fragment or an array of both elements under the `<Tabs>` component lead to styles not applied to the Tab.
* Created types like `DashboardManifest` and `DashboardConfig` to be reused in all places using dashboards (Dashboards, Public Dashboards & Custom views)
* Adds translation keys for the `Custom view` string

##### Backend

Data model changes:
* Adds a `path` property to the custom view entity type and expose it in the API. This will be a concatenation of a slug (from the `name`) with its ID, and is used in the URL bar of the user, it's a path à la Notion that allows changing the name/slug while still being able to open an old link.
* Adds a `target_entity_type` to CustomView entities to tell which entities are concerned by this custom view

### Related issues

* Fixes https://github.com/OpenCTI-Platform/opencti/issues/13389

### Manual testing

Feature branch deployment available here : https://feat-issue-13389-2-c.octi.staging.filigran.io/ .
Use "feat-issue-13389-2-c" credentials in 1Password.

Connect as `customers@filigran.io`.

To test case 1: go to a `Campaign` page. You should see a single custom view tab.
To test case 2 go to a `Intrusion Set` page. You should see two custom views under a dropdown tab.

### Checklist


opencti-platform/opencti-front/src/components/TabWithDropDownMenu.tsx
```diff
@@ -0,0 +1,103 @@
...
+  /** Should skip creating Tab and Menu components **/
+  skip?: boolean;
+};
+
+/**
+ * A special type of Tab that displays a drop-down menu on click.
+ * Unorthodoxly this functionality is exposed as a hook as we want to provide
+ * two distinct components given the <Menu> component cannot
+ */
+const useTabWithDropDownMenu = ({ skip, renderMenuItems, label, ...tabProps }: TabWithDropDownMenuProps) => {
```
esrevi: I don't have a suggestion yet, but we should avoid returning a component with a hook as it is considerer as antipattern, i try to suggest another solution later in the day

opencti-platform/opencti-front/src/private/components/common/stix_domain_objects/StixDomainObjectTabsBox.tsx
```diff
@@ -78,34 +81,61 @@ const TABS_INFO: readonly TabInfo[] = [{
   label: 'History',
 }];
 
-const StixDomainObjectTabsBox = ({ basePath, extraActions, tabs }: StixDomainObjectTabsBoxProps) => {
+const CONTAINER_STYLE = {
```
esrevi: in the rest of the codebase, we use lowercase names and type sx objects as SxProps, could we align with that convention?

opencti-platform/opencti-front/src/components/FeatureFlagged.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+interface FeatureFlaggedProps {
+  flags: string[];
+  Enabled: ReactNode;
+  Disabled: ReactNode;
+}
+
+/**
+ * Utility component to switch between two components based on the value
+ * of feature flags. If at least one othe given `flags` is enabled then
```
esrevi:
```suggestion
 * of feature flags. If at least one of the given `flags` is enabled then
```

opencti-platform/opencti-front/src/components/FeatureFlagged.tsx
```diff
@@ -0,0 +1,21 @@
+import { ReactNode } from 'react';
+import useHelper from '../utils/hooks/useHelper';
+
+interface FeatureFlaggedProps {
```
esrevi: I think a more standard react pattern would be to render the component as children and suggest a fallback as prop and also, imo FeatureFlag component is simpler for the name no need to describe a state with FeatureFlagged


```suggestion
interface FeatureFlagProps {
flags: string[]
children: ReactNode
fallback?: ReactNode // better than disabled which is more familiar with Suspense
}

const FeatureFlag = ({ flags, children, fallback = null }: FeatureFlagProps) => {
  const { isFeatureEnable } = useHelper();
  return flags.some(isFeatureEnable) ? children : fallback;
};
```

so you can use like

```
<FeatureFlag flags={['CUSTOM_VIEW']}>
  <CustomViewTabsWrapper />
</FeatureFlag>

<FeatureFlag 
flags={['CUSTOM_VIEW']}
fallback={<Component />} 
>
  <CustomViewTabsWrapper />
</FeatureFlag>
```

opencti-platform/opencti-front/src/components/NotionLikeRedirector.tsx
```diff
@@ -0,0 +1,74 @@
+import { ReactNode } from 'react';
+import { Navigate, useParams } from 'react-router-dom';
+
+export interface NotionLikePageInfo {
```
esrevi: imo, the name is too much linked to Notion, especially, if people don't know Notion ^^
I think we could find another name, maybe SlugRedirectRoute or something else

opencti-platform/opencti-graphql/src/modules/customView/customView-domain.ts
```diff
@@ -1 +1,103 @@
...
+function isCustomViewsAvailableForEntityType(entityType: string) {
+  const candidateTypes = [
+    ...schemaTypesDefinition
+      .get(ABSTRACT_STIX_DOMAIN_OBJECT),
+    ABSTRACT_STIX_CORE_RELATIONSHIP,
+    ABSTRACT_STIX_CYBER_OBSERVABLE,
+    ENTITY_HASHED_OBSERVABLE_ARTIFACT,
+  ];
+
+  // if (isFeatureEnabled('DRAFT_WORKFLOW')) {
```
esrevi: is it something to do later with drafts ?

opencti-platform/opencti-front/src/private/components/custom_views/CustomView.tsx
```diff
@@ -0,0 +1,51 @@
...
+};
+
+interface CustomViewComponentProps {
+  manifest: string;
+}
+
+/**
+ * Displays a custom view from its serialized content
+ */
+const CustomViewComponent = ({ manifest }: CustomViewComponentProps) => {
```
esrevi: the component is named CustomViewComponent, but the file is CustomView.tsx and it's used as CustomView in the app, "Component" is also redundant, i suggest rename to CustomView directly

opencti-platform/opencti-front/src/private/components/custom_views/RootCustomView.tsx
```diff
@@ -0,0 +1,49 @@
...
+  return (
+    <ErrorBoundary>
+      <Suspense fallback={<Loader />}>
+        {queryRef && <RootCustomViewComponent queryRef={queryRef} />}
+      </Suspense>
+    </ErrorBoundary>
+  );
+};
+
+export default RootCustomView;
```
esrevi, line 49: I'd suggest to also rename this component to RootCustomView as we're using as RootCustomView in CustomViewRedirector

opencti-platform/opencti-front/src/private/components/common/stix_domain_objects/StixDomainObjectTabsBox.tsx
```diff
@@ -105,7 +138,17 @@ const StixDomainObjectTabsBox = ({ basePath, extraActions, tabs }: StixDomainObj
...
             ))
         }
+        {shouldDisplayCustomViewTab ? (
+          <Tab
+            component={Link}
+            to={customViews[0].path}
+            value={CUSTOM_VIEW_TAB_VALUE}
+            label={customViews[0].name}
+          />
+        ) : null}
```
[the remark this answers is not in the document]
fellowseb: path is relative no need to preprend basePath

opencti-platform/opencti-front/src/public/components/PublicDashboard.tsx
```diff
@@ -38,7 +38,7 @@ const PublicDashboardComponent = ({
 
   const { publicDashboardByUriKey } = usePreloadedQuery(publicDashboardQuery, queryRef);
   const manifest = publicDashboardByUriKey?.public_manifest;
-  const parsedManifest: PublicManifest = JSON.parse(manifest ? fromB64(manifest) : '{}');
+  const parsedManifest: DashboardManifest = JSON.parse(manifest ? fromB64(manifest) : '{}');
   const { widgets, config } = parsedManifest;
```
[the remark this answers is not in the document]
fellowseb: I considered it actually but didn't want to risk the change given `deserializeDashboardManifestForFrontend` does extra work on filters; even though from what I see Public dashboards don't have filters. Don't want to break something unrelated to my work without better/clearer understanding of implications.

opencti-platform/opencti-front/src/components/TabWithDropDownMenu.tsx
```diff
@@ -0,0 +1,103 @@
...
+const LABEL_STYLE = {
+  textTransform: 'lowercase',
+  display: 'inline-block',
+  '&::first-letter': {
+    textTransform: 'uppercase',
+  },
+} as const;
+
+const ARROW_STYLE = {
+  fontStyle: '20px',
```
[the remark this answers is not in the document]
fellowseb: Fixed, thanks bro

opencti-platform/opencti-front/src/components/FeatureFlagged.tsx
fellowseb: OK even though I like the explicitness of naming the ON and OFF cases. Some could argue that, when following your suggestion, the "nominal" case would be when the ff is OFF; while some could argue the contrary.

FYI here's one suggestion from MistralAI that I like TBH:
```
<FeatureToggle
  flags=["new_dashboard_enabled"]
  on={NewDashboard}
  off={LegacyDashboard}
/>
```

opencti-platform/opencti-front/src/components/NotionLikeRedirector.tsx
fellowseb: Yeah true, changing

opencti-platform/opencti-graphql/src/modules/customView/customView-domain.ts
fellowseb: I was unsure if drafts should be part of the acceptable types for custom views. But no, so removing this.

opencti-platform/opencti-front/src/private/components/custom_views/CustomView.tsx
fellowseb: True that, fixing

opencti-platform/opencti-front/src/private/components/custom_views/RootCustomView.tsx
fellowseb: Agreed

opencti-platform/opencti-front/src/components/FeatureFlagged.tsx
fellowseb: Fixed

opencti-platform/opencti-front/src/private/components/common/stix_domain_objects/StixDomainObjectTabsBox.tsx
fellowseb: Fixed. Reverted extracting to constant when styles are not reused at least once.

opencti-platform/opencti-front/src/components/FeatureFlagged.tsx
fellowseb: Removed the component

PR #15390, opened 2026-04-07
[frontend][backend] List custom views in settings (#13389)

fellowseb:
### Proposed changes

Implements chunk `Settings: an administrator is able to see listed custom views for an entity type` of the ["Custom views" feature](https://github.com/OpenCTI-Platform/opencti/issues/13389)

A custom view is a dashboard an administrator creates for a given entity type. This dashboard is then viewable (not editable) by "normal" users.

#### User journeys

* Case 1: for Entity Types that are eligible to having Custom Views, given there are custom views in the database, when an admin goes to Settings > Customization > [Entity type] a "Custom views" section is displayed and lists the views already created.
* Case 2: for Entity Types that are eligible to having Custom Views, given there are no custom views in the database, when an admin goes to Settings > Customization > [Entity type] a "Custom views" section is displayed and displays "No items available".
* Case 3: for Entity Types that are NOT eligible to having Custom Views, when an admin goes to Settings > Customization > [Entity type] no "Custom views" section is displayed.
* Case 4: non admins don't have access to the Customization page

https://github.com/user-attachments/assets/afd921eb-0cd6-4647-a77e-8efb7e27b46b

#### Specs

Figma: https://www.figma.com/design/K2TFfOTI4MBgWzIyGT2AHV/Custom-views?node-id=119-150&t=y1ubGmkFiRj9H8fl-0
Notion: https://www.notion.so/filigran/Custom-view-Settings-an-administrator-is-able-to-see-listed-custom-views-for-an-entity-type-3188fce17f2a8052b31bcb9b0628e56d?source=copy_link

##### Frontend

* Introduces new `CustomViewSettings` and `CustomViewSettingsDataTable` components
* Introduces new `useDataTableLocalSort` hook to perform client-side sort
* Integrate to new setting panels

##### Backend

Data model changes:
* Adds a `customViews` query
* Adds integration test cases, 100% coverage

### Related issues

* Fixes https://github.com/OpenCTI-Platform/opencti/issues/13389

### Manual testing

Feature branch deployment available here : https://feat-issue-13389-3-l.octi.staging.filigran.io/.
Use "feat-issue-13389-3-l" credentials in 1Password.

Connect as `admin@filigran.io`.

Test `Intrusion-Set` and `Campaign` entity types for content.

### Checklist


opencti-platform/opencti-front/src/components/dataGrid/useDataTableLocalSort.ts
```diff
@@ -0,0 +1,48 @@
...
+    true,
+  );
+  const { sortBy, orderAsc } = viewStorage;
+
+  const sortedData = useMemo(() =>
+    sort(
+      data,
+      sortBy ?? initialValues.sortBy,
+      orderAsc ?? initialValues.orderAsc,
+    ), [data, sortBy, orderAsc],
```
esrevi: i think we should also include initialValues.sortBy and initialValues.orderAsc in the deps

opencti-platform/opencti-front/src/private/components/settings/sub_types/custom_views/types.ts
```diff
@@ -0,0 +1,3 @@
+import { CustomViewsSettings_customViews$data } from './__generated__/CustomViewsSettings_customViews.graphql';
```
esrevi: There are now two types.ts files in the repo with the custom views, i suggest to rename them to [component].types.ts so it is easier to scan through the project

opencti-platform/opencti-front/src/private/components/settings/sub_types/custom_views/types.ts
```diff
@@ -0,0 +1,3 @@
...
+
+export type CustomViewsSettingsEntry = CustomViewsSettings_customViews$data['customViews'][number];
```
esrevi: this type is only referenced once, does it worth extracting that type? or keep it locally in the file that uses it

opencti-platform/opencti-front/src/private/components/settings/sub_types/Root.tsx
```diff
@@ -1,50 +1,48 @@
...
-interface SubTypeTabsContext {
-  tabs: SubTypeTabs;
-}
+const ORDERED_TABS = [
+  'workflow',
+  'attributes',
+  'templates',
+  'overview-layout',
+  'custom-views',
+] as const satisfies Array<keyof SubTypeTabs>;
```
esrevi: suggestion to imply that the list should not be changed with using readonly
```readonly (keyof SubTypeTabs)[]```

File: opencti-platform/opencti-front/src/private/components/settings/sub_types/Root.tsx [hunk 5]
```diff
@@ -57,6 +55,7 @@ const RootSubType = () => {
           <Route path="templates" element={<FintelTemplatesManager />} />
```
esrevi: now the index redirection uses ORDERED_TABS, if we add, change a key, the route will point to a unknown path, I think we should add constants and use them for the ordered tabs and the paths to prevent mistakes

opencti-platform/opencti-graphql/src/modules/customView/customView.graphql
```diff
@@ -30,13 +30,35 @@ type CustomViewsDisplayContext {
...
+}
+
 # Queries
 type Query {
-  ## Platform user use cases
+    ## Platform user use cases
+    
+    # Provides context when browsing to know which custom views are
+    # available for which entity types
+    customViewsDisplayContext: [CustomViewsDisplayContext!] @auth @ff(flags: ["CUSTOM_VIEW"], softFail: true)
```
esrevi, line 52: just for my understanding, why we set softFail here to true and for customViewDisplay, i guess it is false by default?

opencti-platform/opencti-front/src/private/components/settings/sub_types/custom_views/CustomViewsSettings.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * Custom Views settings page.
+ */
+const CustomViewsSettings = () => {
+  const { t_i18n } = useFormatter();
+  const { customViewsSettings, subType } = useSubTypeOutletContext();
+  const { customViews } = useFragment(
+    customViewsFragment,
+    customViewsSettings as CustomViewsSettings_customViews$key,
```
fellowseb, line 29: Not sure why this assertion is required. Just like in your PR @esrevi .

opencti-platform/opencti-graphql/src/modules/customView/customView.graphql
fellowseb: Yes it's false by default. It's used on `customViewsDisplayContext` as it's used on the first query made in the app; that itself queries the feature flags. I realise it makes this endpoint definition a little tied to its usage but I think it's fine.

opencti-platform/opencti-front/src/private/components/settings/sub_types/Root.tsx
fellowseb: we had the same issue before tbh (:, but ok

opencti-platform/opencti-front/src/private/components/settings/sub_types/custom_views/types.ts
fellowseb: true ! moving to the other file

opencti-platform/opencti-front/src/components/dataGrid/useDataTableLocalSort.ts
fellowseb: oups, indeed

opencti-platform/opencti-front/src/private/components/settings/sub_types/Root.tsx
esrevi: yep, let s improve that then :p

opencti-platform/opencti-front/src/private/components/settings/sub_types/custom_views/CustomViewsSettings.tsx
esrevi: one other solution would be to type the useFragment I think like
```
  const { customViews } = useFragment<CustomViewsSettings_customViews$key>(
    customViewsFragment,
    customViewsSettings,
  );
```

esrevi (approved):

PR #15398, opened 2026-04-07
[backend/frontend] Improve data sharing in public mode (#14918)

### LINKED ISSUE #14918: Improve data sharing in public mode

ludovic:
### Proposed changes

* **Chunk A**
    * Add `taxii_public_user_id` field to `TaxiiCollection` (ES attributes, GraphQL schema, TS types)
    * Add `feed_public_user_id` field to `Feed` (ES attributes, GraphQL schema, TS types)
    * Add `stream_public_user_id` field to `StreamCollection` (ES attributes, GraphQL schema, TS types)

* **Chunk B**
    * Add a `resolvePublicUser` helper in `dataSharing-utils.ts` that resolves the configured user from cache and rejects internal system users
    * Add domain validation: creating or editing a public sharing endpoint without configuring a `*_public_user_id` now throws a `FunctionalError`
    * Add `*_public_user_id` and `*_public_authorized_markings` fields to `Stix*` interfaces and converters so they are serialized in STIX bundles

* **Chunk C**
    **Backend**
    
    - Added *_public_user: Creator GraphQL fields to `taxiiCollection.graphql`, `streamCollection.graphql `and `feed.graphql,` resolved via loadCreator in the corresponding resolvers
    
     **Frontend**
    
    - Added a `CreatorField` (conditional on *_public: true) to the creation forms for TAXII collections, CSV feeds and live streams
    
    - Added a `CreatorField` to the edition forms for all three, pre-populated with the existing user via `convertUser`
    
    - Made `*_public_user_id` required in Yup validation when `*_public: true`
    
    - Updated Relay fragments to include the resolved `*_public_user { id, entity_type, name }` sub-object (replacing the raw ID fetch)

* **Chunk D**
    This chunk handles cleaning up public configs when a user is deleted and fixes UX issues in the edition forms.
    
    **Backend**
    

    - Cleanup on user deletion: Added disablePublicSharingForDeletedUser inside userDelete. It automatically sets *_public = false (and stream_live = false) on any Feed, TAXII, or Stream using the deleted user, and clears *_public_user_id.
    - Fix "System" UI display: *_public_user resolvers now properly return null instead of falling back to SYSTEM_USER when *_public_user_id is empty.
    - Edition validation: Backend now strictly rejects passing *_public = true during fieldPatch mutations if no *_public_user_id is defined.

    **Frontend**
    
    - Controlled Public Switch: Stream and TAXII edition forms now use controlled Formik state for the public switch.
    - Batched Save on Enable: Toggling public ON merely shows the user selector. Selecting a user then triggers a single mutation sending both *_public_user_id and *_public = true.
    - Immediate Save on Disable: Toggling OFF disables public sharing immediately.
    - Delete Warning: Added an alert in UserDeletionDialog.tsx warning that associated public feeds and streams will be disabled.

### Related issues

* Fixes #14918 

### Checklist


opencti-platform/opencti-front/src/private/components/data/stream/StreamCollectionEdition.tsx
```diff
@@ -172,6 +178,15 @@ const StreamCollectionEditionContainer: FunctionComponent<{ streamCollection: St
                 name="restricted_members"
               />
             )}
+            {initialValues.stream_public && (
+              <CreatorField
+                name="stream_public_user_id"
+                label={t_i18n('Share data corresponding to the right associated with this user')}
```
aHenryJard:
```suggestion
                label={t_i18n('Share data corresponding to permissions associated with this user')}
```

opencti-platform/opencti-front/src/private/components/data/taxii/TaxiiCollectionCreation.tsx
```diff
@@ -126,8 +133,7 @@ const TaxiiCollectionCreation: FunctionComponent<TaxiiCollectionCreationProps> =
             authorized_members: [],
             taxii_public: false,
             include_inferences: true,
-            score_to_confidence: false,
-          }}
+            score_to_confidence: false, taxii_public_user_id: null }}
```
aHenryJard: maybe you need a new line here to be consistent with above lines

opencti-platform/opencti-front/src/private/components/data/taxii/TaxiiCollectionCreation.tsx
```diff
@@ -182,6 +188,14 @@ const TaxiiCollectionCreation: FunctionComponent<TaxiiCollectionCreationProps> =
                     name="authorized_members"
                   />
                 )}
+                {values.taxii_public && (
+                  <CreatorField
+                    name="taxii_public_user_id"
+                    label={t_i18n('Share data corresponding to the right associated with this user')}
```
aHenryJard: same here, unless it has been validated with product ? But I don't think we use the 'right' word overall.
```suggestion
                    label={t_i18n('Share data corresponding to permissions associated with this user')}
```

opencti-platform/opencti-graphql/src/http/httpRollingFeed.ts
```diff
@@ -276,6 +276,10 @@ export const buildCsvLines = (elements: any[], feed: BasicStoreEntityFeed, neigh
   return lines;
 };
 
+export const resolveUserForFeed = async (context: AuthContext, feed: BasicStoreEntityFeed): Promise<AuthUser> => {
+  return context.user ?? resolvePublicUser(context, feed.feed_public_user_id);
```
aHenryJard: I think you need to check if feed is public before try to resolve a public user, to be sure it's in a context of a public exposed feed.

opencti-platform/opencti-front/src/private/components/data/taxii/TaxiiCollectionCreation.tsx
ludovic: Yes, the wording comes from the Product, but I'll pass your comment on to Nino.

opencti-platform/opencti-graphql/src/http/httpRollingFeed.ts
ludovic: The upstream guards already prevent reaching this point without a valid user or a public feed, but adding the explicit check here makes the function self-contained and the intent clearer. Done!

JeremyCloarec (approved):

Issue #14285, opened 2026-01-30
[Playbook] Add filters to components that manipulate Knowledge when one of the all elements, main entity only, or all except main is selected.


alice-debra:
## Use case

**When** manipulating knowledge within a playbook, 

**I want** to be able to specify filters if I manipulate one of those: “all elements” present in the bundle, “only the main entity” present in the bundle, or “all entities without the main entity” present in the bundle

ancazzz:
Should be developer after https://github.com/OpenCTI-Platform/opencti/issues/14283 on the same feature branch

PR #15427, opened 2026-04-09
[backend] Playbooks: add filter for component Container Wrapper (#14285)

lndrtrbn:
### Proposed changes

* Add new attribute `applyWithFilters` on component config
* Add backend tests for the logic
* Update documentation

### Related issues

* #14285 

### Checklist

### DIFF
--- a/docs/docs/usage/automation.md
+++ b/docs/docs/usage/automation.md
@@ -229,6 +229,9 @@ By default, wrapping is applied to the entity triggering the playbook. You can c
 
 **Specificities of the component**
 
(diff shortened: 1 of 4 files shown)


MTorbay-Filigran (commented):
When a group of filters don't match anything (ex : entity type = campaign and entity type = malware), but match elements individually (in this exemple, malware and / or campaign in the bundle), an empty container is created. Suggestion : if no element match, it should not create a container.

MTorbay-Filigran (approved):

PR #15442, opened 2026-04-10
[deps] add tooling script to test repo for vulnerabilities (#15433)

labo-flg:
### Changes 

This pull request adds a new script, `scripts/snyk-vuln-check.sh`, to automate Snyk vulnerability scanning across the repository. The script handles dependency installation, branch management, authentication checks, and report generation for Snyk results.

New Snyk vulnerability scan automation:
* Added `scripts/snyk-vuln-check.sh`, a Bash script to automate Snyk scans, including:
  * Optional dependency installation for all key subprojects (`yarn` and `pip`).
  * Automatic detection and switching to the specified or current Git branch.
  * Authentication check for Snyk.
  * Running Snyk tests across all projects and saving the output to a timestamped report file in `.snyk-reports/`.
  * Safe handling of local changes (stashing and restoring as needed).

This tool enable us easily scan the repo for vulnerabilities on a specific branch.

### Related issues
* closes OpenCTI-Platform/release#30

### DIFF
--- a/scripts/snyk-vuln-check.sh
+++ b/scripts/snyk-vuln-check.sh
@@ -54,6 +54,13 @@ install_deps() {
   pip install -r requirements.txt
 }
 
+check_snyk_auth() {
+  log "Checking Snyk authentication..."
+  if ! yarn dlx snyk whoami &>/dev/null; then
+    err "Snyk is not authenticated. Please run 'snyk auth' first and try again."
+  fi
+}
+
 run_snyk() {
   log "Running Snyk scan..."
   cd "$REPO_ROOT"
@@ -81,6 +88,7 @@ REPORT_FILE="${REPO_ROOT}/.snyk-reports/snyk-${BRANCH//\//-}-$(git rev-parse --s
 
 [ "$SKIP_INSTALL" = true ] && log "Skipping dependency installation (--skip-install)" || install_deps
 
+check_snyk_auth
 run_snyk
 
 [ "$STASHED" = true ] && { log "Restoring stashed changes..."; git stash pop; }
--- a/scripts/snyk-vuln-check.sh
+++ b/scripts/snyk-vuln-check.sh
@@ -5,7 +5,9 @@ set -euo pipefail
 # Configuration
 # ─────────────────────────────────────────────
 SKIP_INSTALL=false
-SNYK_VERSION="1.1291.0"  # Pin Snyk version for reproducibility
+SNYK_VERSION="1.1291.0"
+PYTHON_MIN_MAJOR=3
+PYTHON_MIN_MINOR=9
 
 usage() {
   echo "Usage: $0 [--skip-install] [<branch>]"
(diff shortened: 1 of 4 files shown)


scripts/snyk-vuln-check.sh
```diff
@@ -0,0 +1,99 @@
...
+
+  cd "$REPO_ROOT/opencti-platform/opencti-graphql"
+  yarn install
+  yarn install:python
+
+  cd "$REPO_ROOT/opencti-platform/opencti-front"
+  yarn install
+
+  cd "$REPO_ROOT/opencti-worker/src"
+  pip install -r requirements.txt
```
OctaveLaventure: directly using `pip` can be kind of ambiguous depending on the env and the potentially diverse python/pip versions installed, running `python3 -m pip ...` and maybe a python version check would ensure consistency between commands

scripts/snyk-vuln-check.sh
```diff
@@ -0,0 +1,99 @@
...
+}
+
+# ─────────────────────────────────────────────
+# Main
+# ─────────────────────────────────────────────
+cd "$REPO_ROOT"
+
+if ! git diff --quiet || ! git diff --cached --quiet; then
+  log "Stashing local changes..."
+  git stash push -m "snyk-compare-stash"
```
OctaveLaventure: As there is a `set -euo pipefail` at the top and most importantly `set -e`, if something fails (like `install_deps` or `check_snyk_auth`) between this and line 94 where you restore stash, the script will fail without restoring the stashed changes or switching back to the previous branch, you may want to use a `trap` cleanup function, I can help if needed, I have a working example locally

scripts/snyk-vuln-check.sh
```diff
@@ -0,0 +1,99 @@
+#!/usr/bin/env bash
```
OctaveLaventure, line 1: We could add a `yarn check-vulnerabilities` command for this now that we have a `package.json` at the root

scripts/snyk-vuln-check.sh
```diff
@@ -0,0 +1,99 @@
...
+# ─────────────────────────────────────────────
+log() { echo -e "\n\033[1;34m>>> $*\033[0m"; }
+err() { echo -e "\033[1;31m[ERROR] $*\033[0m" >&2; exit 1; }
+
+install_deps() {
+  log "Installing dependencies..."
+
+  cd "$REPO_ROOT"
+  yarn install
+
```
OctaveLaventure: I think this is almost equivalent to `yarn deps` we added in NX, so maybe we can use that instead so we don't have to maintain this script's install steps.
We would need to add a little `project.json` to the `docs` to be iso with this script

scripts/snyk-vuln-check.sh
labo-flg: thanks ! I've added a trap cleanup :)

scripts/snyk-vuln-check.sh
labo-flg: done !

scripts/snyk-vuln-check.sh
labo-flg: done

PR #15443, opened 2026-04-10
[frontend] fix rectangle selection (#15432)

### LINKED ISSUE #15432: Rectangle Selection Tool Does Not Accurately Select Nodes in Knowledge Graph
## Description

The rectangle (box) selection tool in the OpenCTI Knowledge Graph does not select nodes accurately. When drawing a rectangle around a group of nodes, the tool does not reliably highlight the intended nodes. Instead, the selection behavior appears inconsistent and somewhat random - nodes outside the drawn boundary are selected while nodes inside the boundary are missed. In some cases, the selection appears to be fully inverted, capturing everything outside the rectangle rather than everything within it.

The free select (lasso) tool does not exhibit this behavior and correctly selects only the nodes within the drawn boundary.

This issue affects the Knowledge Graph view available on Reports, Groupings, and Cases (Incident Response, RFI, RFT).

## Reproducible Steps

1. Navigate to Analyses > Reports and open any Report containing multiple nodes
2. On the Knowledge tab, click the Graph view icon in the top-right corner of the content area to switch to the interactive graph
3. Ensure multiple nodes are visible in the graph canvas
4. Select the rectangle (box) selection tool from the bottom toolbar of the graph canvas
5. Click and drag to draw a rectangle around a group of nodes
6. Release the selection

## Expected Output

All nodes that fall visually within the drawn rectangle boundary are selected and highlighted after the selection is released.

## Actual Output

The selection does not match the drawn boundary. Observed behaviors include:
- Nodes inside the rectangle are not selected
- Nodes outside the rectangle are selected instead
- The overall selection appears random or inverted compared to the drawn area

## Workaround

Use the free select tool as an alternative to build a selection.

JeremyCloarec:
### Proposed changes

**Root cause**
The react-rectangle-selection library positions its visual selection box using pageX/pageY as CSS left/top on a position: absolute div. These are document-relative coordinates, but the CSS positioning resolves relative to the nearest positioned ancestor — not the document origin. Since the graph is nested inside the OpenCTI layout (with a side nav, top bar, etc.), the visual box was offset from the actual cursor position.

Fix in RectangleSelection.tsx:
1. Hidden the library's built-in selection box (display: 'none' via the style prop) since it can't be positioned correctly.
2. Render our own overlay div using the already-computed canvas-relative coords state. This overlay is positioned inside a new position: relative wrapper div, so position: absolute on the overlay resolves correctly relative to the graph area.
3. Added scroll offset correction (window.scrollX/window.scrollY) when computing canvas-relative coordinates from the library's pageX/pageY values and getBoundingClientRect() (viewport-relative), ensuring correctness even if the page is scrolled.

### Related issues

* fix #15432 
*

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/components/graph/components/RectangleSelection.tsx
+++ b/opencti-platform/opencti-front/src/components/graph/components/RectangleSelection.tsx
@@ -40,14 +40,18 @@ const RectangleSelection = ({
       const graphCanvas = document.querySelector(`#${graphId} canvas`);
       if (graphCanvas) {
         const { left, top } = graphCanvas.getBoundingClientRect();
+        // The library provides page coordinates (pageX/pageY) but
+        // getBoundingClientRect() is viewport-relative; account for scroll.
+        const offsetLeft = left + window.scrollX;
+        const offsetTop = top + window.scrollY;
         setCoords({
           origin: [
-            Math.min(selectionCoords.origin[0], selectionCoords.target[0]) - left,
-            Math.min(selectionCoords.origin[1], selectionCoords.target[1]) - top,
+            Math.min(selectionCoords.origin[0], selectionCoords.target[0]) - offsetLeft,
+            Math.min(selectionCoords.origin[1], selectionCoords.target[1]) - offsetTop,
           ],
           target: [
-            Math.max(selectionCoords.origin[0], selectionCoords.target[0]) - left,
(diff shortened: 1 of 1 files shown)


CelineSebe (approved):

CelineSebe:
locally tested 👌

PR #15465, opened 2026-04-12
[backend] Fix work status transitioning to complete too early (#8224)

### LINKED ISSUE #8224: The "completed" status in OpenCTI can be misleading
## Description

The "completed" status in OpenCTI can be misleading depending on the connector.

From what I understood, here is how it works:

- When a connector sends a bundle of data, it creates a work, i.e. one of the lines with the progress bar (visible in "Data > Ingestion > Connector > [any connector]").
- When the bundle has been ingested, the platform sets the work to "complete".

However, the work can still be manipulated thanks to its ID. So if a connector decides to send data later using the ID of an old work rather than creating a new one, you end up in the situation of a work in "Completed" that continues to ingest data.

This is the case with the GroupIB connector, and I believe we observed the same thing with Sekoia.

NB: I'm not sure how to categorize this, is it a bug? I'll let you decide if it isn't

## Environment

OCTI 6.2.18

## Reproducible Steps

Nothing very direct:
1. Activate a connector that produces this kind of behavior
2. Let it run
3. Look at the works and try to see this

image

fellowseb:
### Problem analysis

Here's a rough sequence diagram about what happens between the Connector, the Backend, the Worker, the Queue:

```mermaid
sequenceDiagram
    autonumber

    participant C as Connector
    participant B as Backend
    participant MQ as RabbitMQ
    participant W as Worker

    rect rgb(220, 235, 255)
        note over C, B: 1. Work Initiation
        C->>+B: initiate_work()<br/>→ mutation workAdd
        B-->>-C: work_id
    end

    rect rgb(220, 255, 225)
        note over C, MQ: 2. Bundle Submission
        C->>+B: send_stix2_bundle()<br/>→ mutation workEdit.addExpectations(work_id, count)
        B-->>-C: ack

        loop For each STIX2 bundle chunk
            C->>MQ: send_stix2_bundle()<br/>→ publish bundle to queue
        end
    end

    rect rgb(255, 245, 200)
        note over C, B: 3. Work Completion Signal
        C->>+B: to_processed()<br/>→ mutation workEdit.toProcessed(work_id)
        B-->>-C: ack
    end

    rect rgb(255, 225, 220)
        note over W, B: 4. Bundle Processing (per bundle)
        loop For each bundle in queue
            W->>+MQ: consume message
            MQ-->>-W: STIX2 bundle

            loop For each STIX2 object in bundle
                W->>+B: import object<br/>→ REST / GraphQL mutation
                B-->>-W: ack
            end

            W->>+B: mutation workEdit.reportExpectation(work_id)
            B-->>-W: ack
        end
    end
```

The steps are the following:
- The "Connector" calls the "workAdd" mutation on the "Backend", via the "initiate_work()" python helper
- The  "Connector" calls the "send_stix2_bundle" python helper that does two things:  calls the "workEdit.addExpectations" mutation on the "Backend" and puts multiple bundles on the "RabbitMQ" queue.
- The "Connector" calls the "workEdit.toProcessed" mutation on the "Backend", via the "to_processed" python helper
- The "Worker" gets an item on the "RabbitMQ" queue to treat it: for each item in the bundle it imports it into the "Backend" and then calls "workEdit.reportExpectation"

There is a race between the connector and the workers: the issue arises when the worker(s) finish all ongoing work (i.e. handle all items of the bundles on the queue, section `4` in the diagramm) before the connector has time to send all bundles (or more precisely, before the connector has time to increase the expectation count to its final value, section `2` in the diagram).
This race only happens when multiple bundles are sent during a work as the connector calls `workEdit.addExpectations` once per bundle in `send_stix2_bundle`.

### Proposed changes

We can fix the race by waiting for both the Connector and the Workers to have finished before proceeding to changing the status to `complete`:
- the Connector signals it has finished by calling the `workEdit.toProcessed` endpoint
- the Workers have finished when all expected items have been processed

Backend:
* Set a `is_processed` flag in redis when the connector calls `workEdit.toProcessed`. Access this flag in `workEdit.reportExpectation` and use as a gate to change the work status to `complete`.
* Add integrations tests to cover the race between connector and workers

### Potential BREAKING CHANGES

This fix could affect some connectors that don't correctly call `api.work.to_processed()` as said by the documentation (see here: https://docs.opencti.io/latest/development/connectors/?h=send_stix2_bundle#initiating-a-work-before-pushing-data) or the provided dev template (see here: https://github.com/OpenCTI-Platform/connectors/blob/master/templates/external-import/src/connector/connector.py#L181).
For those the Work could remain blocked in the `progress` state. We do however have a mechanism to clean up all old works so it's not a big deal IMO.

Also, I've opened another PR to fix all `3` external-import connectors that don't call `to_processed()`.

I suggest we advertise this change as potentially a breaking change to let on-prem admins fix their connectors if needed.
It's still a better solution than an alternative

### Related issues

* Fixes https://github.com/OpenCTI-Platform/opencti/issues/8224

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/domain/work.js
+++ b/opencti-platform/opencti-graphql/src/domain/work.js
@@ -331,11 +331,11 @@ export const updateExpectationsNumber = async (context, user, workId, expectatio
     logApp.warn('The work cannot be found in database, expectation cannot be updated.', { workId, expectations });
     return workId;
   }
+  await redisUpdateActionExpectation(user, workId, expectations);
   const params = { updated_at: now(), import_expected_number: expectations };
   let source = 'ctx._source.updated_at = params.updated_at;';
   source += 'ctx._source["import_expected_number"] = ctx._source["import_expected_number"] + params.import_expected_number;';
   await elUpdate(currentWork._index, workId, { script: { source, lang: 'painless', params } });
-  await redisUpdateActionExpectation(user, workId, expectations);
   // Ensure that work hasn't been deleted in the meantime in case of race condition
   const workAlive = await isWorkAlive(context, user, workId);
   if (!workAlive) {
@@ -389,11 +389,10 @@ export const updateProcessedTime = async (context, user, workId, message, inErro
     return workId;
   }
   const { expected, total, isMultiPartWork } = await redisGetWorkCompletionState(workId);
-  const workIsFinished = isWorkFinished(expected, total);
-  if (isMultiPartWork && !workIsFinished) {
+  const isComplete = isWorkFinished(expected, total);
+  if (isMultiPartWork && !isComplete) {
     await redisMarkWorkAsProcessed(workId);
   }
-  const isComplete = currentWork.import_expected_number === 0 || workIsFinished;
   const params = { processed_time: now(), message };
   let source = 'ctx._source["processed_time"] = params.processed_time;';
   if (isComplete) {
--- a/opencti-platform/opencti-graphql/src/domain/work.js
+++ b/opencti-platform/opencti-graphql/src/domain/work.js
@@ -384,9 +384,9 @@ export const updateProcessedTime = async (context, user, workId, message, inErro
   await redisMarkWorkAsProcessed(workId);
   const params = { processed_time: now(), message };
   const { expected, total } = await redisGetWorkCompletionState(workId);
-  const isComplete = isWorkFinished(expected, total);
+  const isComplete = currentWork.import_expected_number === 0 || isWorkFinished(expected, total);
   let source = 'ctx._source["processed_time"] = params.processed_time;';
-  if (currentWork.import_expected_number === 0 || isComplete) {
+  if (isComplete) {
     params.completed_number = total && !Number.isNaN(total) ? total : 1;
     source += `ctx._source['status'] = "complete";
                ctx._source['import_expected_number'] = params.completed_number;
--- a/opencti-platform/opencti-graphql/src/domain/work.js
+++ b/opencti-platform/opencti-graphql/src/domain/work.js
@@ -388,7 +388,9 @@ export const updateProcessedTime = async (context, user, workId, message, inErro
     logApp.warn('The work cannot be found in database, processed time cannot be updated.', { workId });
     return workId;
   }
-  await redisMarkWorkAsProcessed(workId);
+  if (currentWork.is_multipart && currentWork.status !== 'complete') {
+    await redisMarkWorkAsProcessed(workId);
+  }
   const params = { processed_time: now(), message };
   const { expected, total } = await redisGetWorkCompletionState(workId);
   const isComplete = currentWork.import_expected_number === 0 || isWorkFinished(expected, total);
--- a/opencti-platform/opencti-graphql/src/domain/work.js
+++ b/opencti-platform/opencti-graphql/src/domain/work.js
@@ -211,7 +211,13 @@ export const deleteWorkForSource = async (sourceId) => {
 
 export const createWork = async (context, user, connector, friendlyName, sourceId, args = {}) => {
   // Create the new work
-  const { receivedTime = null, background_task_id, fileMarkings = [], draftContext } = args;
+  const {
+    receivedTime = null,
+    background_task_id,
+    fileMarkings = [],
+    draftContext,
+  } = args;
+  const isMultiPartWork = args.isMultiPartWork === true;
   // Create the work and an initial job
   const { id: workId, timestamp } = generateWorkId(connector.internal_id);
   const work = {
@@ -234,6 +240,7 @@ export const createWork = async (context, user, connector, friendlyName, sourceI
     processed_time: null,
     completed_time: null,
     completed_number: 0,
+    is_multipart: isMultiPartWork,
     messages: [],
     errors: [],
     [buildRefRelationKey(RELATION_OBJECT_MARKING)]: [...fileMarkings],
@@ -245,7 +252,7 @@ export const createWork = async (context, user, connector, friendlyName, sourceI
   const createdWork = await loadWorkById(context, user, workId);
   // If work was created, initialize work on redis
   if (createdWork) {
-    await redisInitializeWork(createdWork.id);
+    await redisInitializeWork(createdWork.id, isMultiPartWork);
   }
   return createdWork;
 };
@@ -271,8 +278,8 @@ const isWorkFinished = (expected, total) => total >= expected;
 export const reportExpectation = async (context, user, workId, errorData) => {
   const timestamp = now();
   await redisUpdateWorkFigures(workId);
-  const { expected, total, isProcessed } = await redisGetWorkCompletionState(workId);
-  const isComplete = isProcessed && isWorkFinished(expected, total);
+  const { expected, total, isProcessed, isMultiPartWork } = await redisGetWorkCompletionState(workId);
+  const isComplete = (!isMultiPartWork || isProcessed) && isWorkFinished(expected, total);
   // Ensure that work hasn't been deleted in the meantime
   const workAlive = await isWorkAlive(context, user, workId);
   if (!workAlive) {
--- a/opencti-platform/opencti-graphql/src/domain/work.js
+++ b/opencti-platform/opencti-graphql/src/domain/work.js
@@ -3,7 +3,15 @@ import * as R from 'ramda';
 import { elDeleteInstances, elIndex, elLoadById, elPaginate, elRawDeleteByQuery, elUpdate, ES_MINIMUM_FIXED_PAGINATION } from '../database/engine';
 import { generateWorkId } from '../schema/identifier';
 import { INDEX_HISTORY, isNotEmptyField, READ_INDEX_HISTORY } from '../database/utils';
-import { isWorkCompleted, redisDeleteWorks, redisGetWork, redisInitializeWork, redisUpdateActionExpectation, redisUpdateWorkFigures } from '../database/redis';
+import {
+  redisDeleteWorks,
+  redisGetWork,
+  redisGetWorkCompletionState,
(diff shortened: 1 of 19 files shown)


JeremyCloarec (commented):
In my opinion, the current solution needs to be changed a bit. With the current implementation, all works are impacted, but this causes 2 issues:
1. internal works are impacted, but a lot of those internal works do not set the work to processed (drafts, form intakes, ingestion manager, task manager but there might be others). This is the reason for the current rule tests failing: the rule activation works are never set to complete since they are not marked as processed. We could change all of those internal processes to mark the work as processed, but this adds a lot more changes to the PR & most importantly I feel like it's not future-proof/safe in it's design (the defaults behavior relies on the process always updating the processing time)
2. I'm not confortable breaking custom customer connectors that might not call toProcessed on the work

My suggestion is to add a new flag on work creation, `isMultiPartWork`, defaulting to false. When the flag is set to true, it would check on the processed status to compute for the completion state, like you did in the PR. But this way, this behavior would be opt-in, and not break the current behavior for other connectors. 
Only drawback it that it would require a bit more work to update the connectors that will now need to mark their works as 'multiPart'

fellowseb:
> In my opinion, the current solution needs to be changed a bit. With the current implementation, all works are impacted, but this causes 2 issues:
> 
> 1. internal works are impacted, but a lot of those internal works do not set the work to processed (drafts, form intakes, ingestion manager, task manager but there might be others). This is the reason for the current rule tests failing: the rule activation works are never set to complete since they are not marked as processed. We could change all of those internal processes to mark the work as processed, but this adds a lot more changes to the PR & most importantly I feel like it's not future-proof/safe in it's design (the defaults behavior relies on the process always updating the processing time)
> 2. I'm not confortable breaking custom customer connectors that might not call toProcessed on the work
> 
> My suggestion is to add a new flag on work creation, `isMultiPartWork`, defaulting to false. When the flag is set to true, it would check on the processed status to compute for the completion state, like you did in the PR. But this way, this behavior would be opt-in, and not break the current behavior for other connectors. Only drawback it that it would require a bit more work to update the connectors that will now need to mark their works as 'multiPart'

I was actually looking into fixing 1.

For 2. I actually have already a commit to add that additional flag but it's really annoying given their are many existing connectors that do multi-part work without the flag: updating the known connectors will require going through all of them. But also their could be customer connectors. This means we'll need to have a flag with three states: `true` (apply current PR logic), `false` (error out when trying to send a second bundle or add more expectations) and `None` to support legacy race-risky state. I find this too complex.
If you default to `false` (and choose not to have a third state) then you're still breaking customers' existing connectors. I feel like leaving the current behaviour (race-risky) in the `false` case, blindly accepting multiple bundles, is error-prone.

Plus, as noted, even if a Work stays in `progress` status it will get deleted or forced to `complete` by the process we have in `ConnecterManager#closeOldWorks`.
The documentation has been stating the need for `to_processed` since at least v5.10.

>  I feel like it's not future-proof/safe in it's design

What doesn't feel safe ?
I prefer myself the design where the work is multi-part by default. It's more explicit:
```
- First you init the work
- Then you `addExpectations` and queue the work as needed
- Finally you notify you've queue all work required
```
But we can still have a flag to support one-shot works, maybe by passing a flag to `addExpectations`.

fellowseb:
> In my opinion, the current solution needs to be changed a bit. With the current implementation, all works are impacted, but this causes 2 issues:
> 
> 1. internal works are impacted, but a lot of those internal works do not set the work to processed (drafts, form intakes, ingestion manager, task manager but there might be others). This is the reason for the current rule tests failing: the rule activation works are never set to complete since they are not marked as processed. We could change all of those internal processes to mark the work as processed, but this adds a lot more changes to the PR & most importantly I feel like it's not future-proof/safe in it's design (the defaults behavior relies on the process always updating the processing time)
> 2. I'm not confortable breaking custom customer connectors that might not call toProcessed on the work
> 
> My suggestion is to add a new flag on work creation, `isMultiPartWork`, defaulting to false. When the flag is set to true, it would check on the processed status to compute for the completion state, like you did in the PR. But this way, this behavior would be opt-in, and not break the current behavior for other connectors. Only drawback it that it would require a bit more work to update the connectors that will now need to mark their works as 'multiPart'

I went with your suggestion 👍 . It was my first idea but I had the feeling the given the low consequences and presence of documentation for many versions that introducing a little breaking change would've been fine. Better to be more conservative

One thing I still don't like about this solution is the fact that we do let connectors send multiple bundles even if isMultiPartWork is `False`. Maybe I could a warning log in `addExpectations` when this happens, to track connectors to migrate.

opencti-platform/opencti-graphql/src/domain/work.js
```diff
@@ -369,10 +388,12 @@ export const updateProcessedTime = async (context, user, workId, message, inErro
     logApp.warn('The work cannot be found in database, processed time cannot be updated.', { workId });
     return workId;
   }
+  await redisMarkWorkAsProcessed(workId);
```
JeremyCloarec: We could save a call to redis in some cases, by only marking the work as processed on redis if the work isn't complete & if the work is multiPart

fellowseb: True, fixed

opencti-platform/opencti-graphql/src/domain/work.js
```diff
@@ -369,10 +388,14 @@ export const updateProcessedTime = async (context, user, workId, message, inErro
     logApp.warn('The work cannot be found in database, processed time cannot be updated.', { workId });
     return workId;
   }
+  if (currentWork.is_multipart && currentWork.status !== 'complete') {
+    await redisMarkWorkAsProcessed(workId);
+  }
```
JeremyCloarec: Here you are checking the status of the work on ES, but doing so you might miss the case where the work in not yet complete on ES & it is complete in the `isComplete` local check.
This condition should be moved after `isComplete` is computed, and should check `isComplete` instead of `currentWork.status`

fellowseb: If there's a race with `reportExpectation` and we're in the case where the work is not yet complete on ES the remainder of the function will do the update in ES itself.

I was a little concerned about those potential races actually that's why I preferred first not optimising (i.e.  always marking the work as processed in redis) and then being looser by looking at the value in ES instead of redis (i.e. marking the work as processed a little more often than needed). It didn't seem like a perf concern given it happens only once at the end of each work.
That being said I don't think there's a race concern with your suggestion, so changing.

opencti-platform/opencti-graphql/src/domain/work.js
```diff
@@ -369,10 +388,15 @@ export const updateProcessedTime = async (context, user, workId, message, inErro
     logApp.warn('The work cannot be found in database, processed time cannot be updated.', { workId });
     return workId;
   }
+  const { expected, total, isMultiPartWork } = await redisGetWorkCompletionState(workId);
+  const workIsFinished = isWorkFinished(expected, total);
+  if (isMultiPartWork && !workIsFinished) {
+    await redisMarkWorkAsProcessed(workId);
+  }
```
JeremyCloarec: I don't get why we're only checking for `workIsFinished` and not the full `isComplete` condition. The case where `currentWork.import_expected_number === 0` is also a condition where we shoulnd't write the work to processed on redis

opencti-platform/opencti-graphql/src/domain/work.js
```diff
@@ -369,10 +388,14 @@ export const updateProcessedTime = async (context, user, workId, message, inErro
...
+  const { expected, total, isMultiPartWork } = await redisGetWorkCompletionState(workId);
+  const isComplete = isWorkFinished(expected, total);
+  if (isMultiPartWork && !isComplete) {
+    await redisMarkWorkAsProcessed(workId);
+  }
   const params = { processed_time: now(), message };
   let source = 'ctx._source["processed_time"] = params.processed_time;';
-  const { isComplete, total } = await isWorkCompleted(workId);
-  if (currentWork.import_expected_number === 0 || isComplete) {
+  if (isComplete) {
```
JeremyCloarec, line 398: I'm sorry, but now I don't get why the `currentWork.import_expected_number === 0` condition was completely removed? For me we should keep it here, and add it on line 393, and have the condition be 
```
if (isMultiPartWork && currentWork.import_expected_number !== 0 && !isComplete) {
   await redisMarkWorkAsProcessed(workId);
}
```

I don't know if the `currentWork.import_expected_number === 0` condition is still relevant, it looks like the isWorkFinished condition already handles that case, but I think it's better to be conservative and keep the condition in case it was there for an edge case I'm not thinking of

fellowseb: I just got mixed up in the logic... Correct me if I'm wrong but this logic handles the case where we want to complete a work that didn't get any addExpectations calls.
If I'm correct we're relying on comparisons with NaN value(s) which is not great.

Added a commit with the following changes:
- for readibility make redisGetWorkComputationState return 0 instead of NaN for expected and total
- isWorkFinished deals correctly with those cases now (i.e. a work not started or where total is defined but expected is not is considered as finished)
- for cohérence use values in redis only, and move the redis call before the ES call in reportExpectation.

I added an integration test for that case too.

WDYT ?

JeremyCloarec: You're right, I think that covers it 👍 I'll aprove the PR

JeremyCloarec (approved):

opencti-platform/opencti-graphql/src/domain/work.js
fellowseb: Thank you (:

fellowseb: Rebased and tested locally using the `opencti` external-import connector 👍

JeremyCloarec (approved):

Issue #15410, opened 2026-04-08
feat(client-python): add objects query connection parameters to entity-based queries


kurtrwall:
### Use Case

When reading a Grouping, Report, or some other Entity that has a connection to "objects", all of the queries are formed with `objects(all: true)` which can overtax the server and cause errors for entities that have a large number of objects since it causes every object to be returned.

For example, the query for a Grouping looks something like the following and cannot be easily changed.
```
query Grouping($id: String!) {
    grouping(id: $id) {
        id
        standard_id
        name
        ...
        objects(all: true) {
              edges {
                  node {
                      ... on BasicObject {
                          id
                          entity_type
                          parent_types
                      }
...snip
```

By simply adding additional variables to the query string, we can parameterize that connection query to control what's queried.

### Current Workaround

Passing `custom_attributes` for each call or monkeypatching the instance `properties` or `properties_with_files`. Even that doesn't cover everything because it only allows you to control the variables passed to the connection query not be able to handle things like pagination.

### Proposed Solution

1. Add a `query_connection()` method to the client to handle pagination.
2. Add additional parameters to entity classes' query methods to control the behavior of the connection query.

### If the feature request is approved, would you be willing to submit a PR?

Yes

PR #15480, opened 2026-04-13
refactor: refactor entity classes w/ inheritance + mixins

kurtrwall:
The main intention is to prevent `objects` from being loaded into the response for every read of an entity. However, considering the repeated code, I took it upon myself to move things into a base class and mixins to make changes to the entities easier to maintain, especially in the future. I've also added some methods that seem to be technically available on many Entity's but not implemented.

Additionally, I've added the `query_connection` method to the api client since it's logic is repeated everywhere a "connection" api member is used.

> Attempts to solve for issue #15410 while maintaining backwards compatibility.

Considering this is a proposed solution for an issue I opened earlier but has yet to be "accepted", I did not add any tests for the changes. However, since this intended to maintain backwards compatibility, all the tests pass without any changes to the test code.

### Proposed changes

- Refactor entity classes for DRYness
- Adds `pycti/entities/base.py::Entity`: a base class for all the entity classes that provides shared methods and a standard instantiation to reduce code and better code maintainability.
  - `OVERRIDES` ClassVar is a new dictionary that holds values to override when building GraphQL queries.
  - `PROPERTIES`: ClassVar that the entity class will use as a default for the instance property `properties`.
  - `OBJECTS_PROPERTIES`: ClassVar that the entity class will use as a default for the instance property `objects_properties`.
  - `FILES_PROPERTIES`: ClassVar that the entity class will use as a default for the instance property `files_properties`.
- Adds `with_objects` (default `True`) to query methods to prevent retrieving _all_ the objects every time.
- Adds `list_objects` method to get an entity's `objects` via optional pagination via mixin.
- Adds `list_files` method to all entity classes that have files associated via mixin.
- Adds `query_connection` method to `OpenCTIApiClient` to query an entity connection.

**Changes**
- The default `properties` for entity classes has been moved to `PROPERTIES`.
- The entity classes' properties has removed the `objects` connection property into a separate variable that is injected when query methods pass the argument `with_objects=True` (the default to maintain backwards compatibility).

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/15410
*

### Checklist


kurtrwall:
The `black` formatting test that failed seems like would always fail. I installed the `pre-commit` hook (per the contribution docs) and it formats the code differently than when I run `black` on its own.

Please advise.

RBourgeoisGH:
Hi @kurtrwall

Thanks for your contribution but unfortunately we don't have the bandwidth to properly review and test refactoring touching so many files. If you want to address the associated issue we would encourage you to just perform the minimum amount of changes needed to fix it.

Thanks

aHenryJard:
This PR seems to be frozen, so I'm closing. Feel free to reopen if needed, but also keep in mind that smaller PR with minimum code changes have more change to be review quickly.

kurtrwall:
Apologies, I didn't see the comments.

Understandable, it does touch a lot of files, but that's the nature of the refactor. The core idea is that each subclass would only contain what's required to handle the related STIX type. As it stands now, there's a lot of repeated code. This has actually led me to building my own registry of predefined, customizable queries alongside a client library for the GraphQL API.

At any rate, thanks everyone for your time considering this. Cheers.

PR #15484, opened 2026-04-13
[backend] Upgrade to vitest@4 (#13771)

### LINKED ISSUE #13771: Use vitest4 for api tests

fellowseb:
### Proposed changes

Followed the migration guide:
https://vitest.dev/guide/migration.html#vitest-4

Integration tests comparison:

Before:
Capture d’écran 2026-05-19 à 12 26 06

After:
Capture d’écran 2026-05-19 à 12 23 04

There's an issue in the amount of test cases, but the correct count is with vitest@4 after the migration.

<details> <summary>Show the result of Copilot's findings:</summary>

```
● Here's a clear summary of the findings:

  Summary:

   - Before: 2,093 total tests across 190 files
   - After: 2,078 total tests across 190 files
   - Difference: 15 tests

  --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

  The only file with a different test count:

  ┌─────────────────────────────────────────────────────┬──────────────┬─────────────┬─────────┐
  │ File                                                │ Before       │ After       │ Diff    │
  ├─────────────────────────────────────────────────────┼──────────────┼─────────────┼─────────┤
  │ tests/03-integration/02-resolvers/subType-test.ts   │ 21 tests     │ 6 tests     │ −15     │
  └─────────────────────────────────────────────────────┴──────────────┴─────────────┴─────────┘

  (connector-test.ts also changed, but only 3 skipped → 3 todo, same total count.)

  --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

  Root cause — vitest 3 double-counting bug:

  The source file subType-test.ts has exactly 6 tests at both commits (master bc5b944 and PR refs/pull/15484/merge). The 15 "extra" tests in before.log were mis-attributed to subType-test.ts by vitest 3 when
  running with isolate: false.

  The verbose output for that file in before.log even reveals the leak — the only printed test name is:

   Add Request Access to an entity and create an RFI. > should enable platform organization

  …which actually belongs to requestAccess-test.ts.

  The PR (vitest 4 upgrade) fixes this counting bug. The 15 tests were never missing — they were being double-counted in vitest 3. In after.log, they're correctly counted only once, in their actual files.
```

and

```
  The config difference is the smoking gun:

  ┌──────┬──────────────────────────────────────┬────────────────────────────────────┐
  │      │ vitest 3 (before)                    │ vitest 4 (after)                   │
  ├──────┼──────────────────────────────────────┼────────────────────────────────────┤
  │ Mode │ poolOptions.forks.singleFork: true   │ maxWorkers: 1 + isolate: false     │
  └──────┴──────────────────────────────────────┴────────────────────────────────────┘

  They seem equivalent — both run everything in one process, sequentially — but they're not.

  What happened in vitest 3 with singleFork: true:

   1. requestAccess-test.ts runs and registers 15 it() calls — including one top-level it() outside any describe()
   2. setupFiles (testSetup.js) re-runs before each file to reset the app context
   3. BUT: vitest 3's singleFork mode doesn't fully clear the test registry between files — the previously registered it()/describe() callbacks from requestAccess-test.ts are still cached in the module scope
   4. When subType-test.ts loads, those 15 stale test registrations leak into its collection phase, so vitest 3 reports 21 tests (6 real + 15 leaked)
   5. The leaked tests even ran again — the stdout logs show subType-test.ts > Add Request Access to an entity and create an RFI. > ... — tests from requestAccess-test.ts executing a second time under
  subType-test.ts's context

  In vitest 4, the explicit isolate: false flag is intentional and paired with a fixed test state lifecycle — the registry is properly flushed between files, so there's no leak.

  In short: those 15 tests aren't missing from after.log — they were being double-counted (and double-run) in before.log due to a vitest 3 singleFork test registry leak. The after.log is actually the correct
  behavior.
```
</details>

And indeed when looking in the logs we see the wrongly attributed tests ! 🤯 

```
2026-05-19T09:47:43.6790576Z  ✓ tests/03-integration/02-resolvers/subType-test.ts (21 tests) 2473ms
2026-05-19T09:47:43.6796521Z    ✓ Add Request Access to an entity and create an RFI. > should enable platform organization  357ms
```

It had effects on the Stream integration test  (`raw-test.js`):

<details> <summary>Details:</summary>

- The number of `case-rfi` created dropped from 10 to 8. I looked at the stream of events on `master` and in contains 2 events with the same message as 2 other messages. They seem to come `case-rfi`s from the same `Add Request Access to an entity and create an RFI. > should create malware with restricted access` test case. This makes me think that the correct count is `8` and not `10`.
```
    {
      "lastEventId": "1779072010081-0",
      "id": "case-rfi--0eaa8f0c-1812-5685-8a54-18cca39a61a5",
      "message": "creates a Case-Rfi `Request Access for entity Malware Request Access by admin via organization TestOrganization`",
      "origin": "http://localhost:4010"
    },
    {
      "lastEventId": "1779072010347-0",
      "id": "case-rfi--32073434-b13d-5bc4-a5b0-0443b8dc5a72",
      "message": "creates a Case-Rfi `Request Access for entity Malware Request Access by anais@opencti.io via organization TestOrganization`",
      "origin": "http://localhost:4010"
    },
...
    {
      "lastEventId": "1779072030976-0",
      "id": "case-rfi--6ee23c75-1864-58fa-a6d1-8d7b65137233",
      "message": "creates a Case-Rfi `Request Access for entity Malware Request Access by admin via organization TestOrganization`",
      "origin": "http://localhost:4010"
    },
    {
      "lastEventId": "1779072031250-0",
      "id": "case-rfi--b44b9032-f1c2-514a-b03f-c38fa15ac40e",
      "message": "creates a Case-Rfi `Request Access for entity Malware Request Access by anais@opencti.io via organization TestOrganization`",
      "origin": "http://localhost:4010"
    }

```

- The same 2 `events` diff is visible in the deletion count
- The number of created `Malware` dropped by one (51 to 50). This time when I checked the stream of events on `master` I see a duplicated event (same ID, same attributes):
```
    {
      "lastEventId": "1779072030899-0",
      "id": "malware--c403e350-c1d7-5bbe-b964-857784703ce2",
      "message": "creates a Malware `Malware Request Access`",
      "origin": "http://localhost:4010"
    },
     {
      "lastEventId": "1779072030899-0",
      "id": "malware--c403e350-c1d7-5bbe-b964-857784703ce2",
      "message": "creates a Malware `Malware Request Access`",
      "origin": "http://localhost:4010"
    },

```
This makes me think again that the correct count should be the lower value.
- The number of updated Malware changed : didn't investigate why.

</details>

So yes all seems to indicate that `vitest@3` was not only counting some tests twice, but also running them twice !

### Related issues

* Fixes https://github.com/OpenCTI-Platform/opencti/issues/13771

### Checklist


opencti-platform/opencti-graphql/vitest.config.integration.ts
```diff
@@ -7,23 +6,21 @@ import { BaseSequencer, type TestSpecification } from 'vitest/node';
...
       reporter: ['text', 'json', 'html'],
       clean: false,
     },
-    poolOptions: {
-      forks: {
-        singleFork: true,
-      },
-    },
+    maxWorkers: 1,
+    isolate: false,
```
nvkhope, line 23: Shouldn't we set `isolate: true` to avoid the risk of shared module state across test files? I would like to have your opinion on it 😄

fellowseb: That was my first reflex TBH. It worked well but increased run time to ~40mins (!). Also it's important to note that currently (under vitest@3) we have disabled the isolation by setting the [poolOptions.forks.singleFork](https://v3.vitest.dev/config/#pooloptions-forks-singlefork) option.

Adding 23mins to each CI run seems a bit too much pragmatically speaking. Especially as it would be without concrete improvements.
Also note that we started an initiative with @MTorbay-Filigran in the CI/CD workgroup to work on those integration tests (isolation, parallelization, ...). We'll be happy to introduce you to it and onboard you on it if you want (:.

aHenryJard (approved):

PR #15499, opened 2026-04-14
[backend] Playbook: option to hide 'listen knowledge' playbooks from manual enrollment (#14579)

### LINKED ISSUE #14579: Playbook: option to hide 'listen knowledge' playbooks from manual enrollment
## Use case

If an entity matches the filter of a playbook setup with listen knowledge, the manual enrollment CTA will be present on the entity (fixed by this: https://github.com/OpenCTI-Platform/opencti/issues/13856). 

However you have some situations  where you don't want this to be present and to be trigger manually, since it may add other labels, change the data, while the playbook has already been triggered. 
Also, this is the reason why the playbook is created with a "listen knowledge" and not "manual enrollment"

SarahBocognano:
### Proposed changes

* Add a new "enrollInPlaybook" boolean attribute in schema configuration
* modifying the function who returns the playbooks available for enrollment to take this boolean into account
* add tests for listen knowledge component

### Related issues

* Closes: https://github.com/OpenCTI-Platform/opencti/issues/14579

### Checklist


opencti-platform/opencti-graphql/tests/01-unit/modules/playbook/listen-knowledge-component-test.ts
```diff
@@ -0,0 +1,144 @@
...
+    vi.spyOn(middleware, 'stixLoadById').mockResolvedValue(mockStixEntity);
+    vi.spyOn(cache, 'getEntitiesListFromCache').mockResolvedValue([buildPlaybook('PLAYBOOK_INTERNAL_DATA_STREAM', { enrollInPlaybook: false })]);
+    vi.spyOn(stixFiltering, 'isStixMatchFilterGroup').mockResolvedValue(true);
+
+    const result = await findPlaybooksForEntity(testContext, {} as any, 'entity-id');
+    expect(result).toHaveLength(0);
+    expect(stixFiltering.isStixMatchFilterGroup).not.toHaveBeenCalled();
+  });
+
+  it('should include playbook when enrollInPlaybook is undefined', async () => {
```
lndrtrbn: ~I think it should not be included when value is undefined, it seems preferable to have explicit `true` only~

opencti-platform/opencti-graphql/src/modules/playbook/playbook-components.ts
```diff
@@ -106,6 +106,7 @@ export interface StreamConfiguration {
   update: boolean;
   delete: boolean;
   filters: string;
+  enrollInPlaybook: boolean;
```
lndrtrbn: I think a name referring to the manual enrollment would be easier to understand, like `canEnrollManually`, what do you think?

opencti-platform/opencti-front/lang/back/de.json
```diff
@@ -318,6 +318,7 @@
   "Endd date": "Enddatum",
   "Enforce reference": "Referenz erzwingen",
   "Enrichment resolution": "Auflösung der Anreicherung",
+  "enrollInPlaybook": "Verfügbar für manuelle Einschreibung",
```
lndrtrbn: The translation here is responsible of the label in the form?

lndrtrbn:
Nice tests suite :+1: 
Could you update the documentation also? :pray:

opencti-platform/opencti-graphql/tests/01-unit/modules/playbook/listen-knowledge-component-test.ts
lndrtrbn: In fact, probably bad idea to keep retro-comaptibility

docs/docs/usage/automation.md
```diff
@@ -85,6 +85,7 @@ You can import a playbook in OpenCTI coming either from your own platform or ano
 
 With this event source, the playbook will be triggered on any knowledge event (create, update or delete) that matches the selected filters.
 Note that you are limited to a subset of filters available for stream events that contain STIX data objects.
+You can decide with the "Available for manual enrollment" if this playbook will be available in the manual enrollment of entities that match the filters you set here.
```
tobybutler405:
```suggestion
Use the "Available for manual enrollment" toggle if you want this playbook to also be available for manual enrollment when you have an entity selected that matches your event source filter criteria. This allows you to hide playbooks that are not appropriate for manual enrollment. 
```

tobybutler405 (dismissed):
One small change to the wording, thanks.

lndrtrbn (approved):

PR #15512, opened 2026-04-15
[backend/frontend] fix Service account name normalization in the catalog(#12845)

### LINKED ISSUE #12845: Service account name normalization is not correct in the catalog
## Description

Service account name normalization is not correct in the catalog

When having something like "CrowdStrike", it should normalize as "crowdstrike", not "crowd-strike", which is confusing and not respecting the brand. When it is "MITRE ATTACK", we can accept "mitre-attack".

That being said, the name of the "user" in OpenCTI is not restricted in terms of characters, why not just normalize email address of the user and let the name exactly as is ([C] name of account = name of the connector) for better value / experience?

CelineSebe:
### Proposed changes

* Removal of CamelCase word splitting ($1-$2) in sanitizeContainerName
* Using the connector’s original display name (display_name) when creating the service account, rather than the standardised ‘name’ in IngestionCreationUserHandling.tsx

### Related issues

* resolves #12845 

### Checklist


marieflorescontact (approved):
locally tested ok 🙂

JeremyCloarec (approved):

PR #15517, opened 2026-04-15
[frontend] fix: prevent using relay cache when checking connectivity (#15516)

### LINKED ISSUE #15516: The button of unregister/register in the OCTI is not updated accordingly
## Description

Refers to https://github.com/FiligranHQ/xtm-hub/issues/2112

Kroustille:
### Proposed changes

* add network only to relay to prevent using cached data

### Related issues

* Fixes #15516
* Refers to https://github.com/FiligranHQ/xtm-hub/issues/2112

### Checklist

### Further comments

https://github.com/user-attachments/assets/3c5ccb83-17d5-4503-9f48-e13d492a759b

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/settings/xtm-hub/XtmHubSettings.tsx
+++ b/opencti-platform/opencti-front/src/private/components/settings/xtm-hub/XtmHubSettings.tsx
@@ -40,6 +40,7 @@ const XtmHubSettingsComponent = () => {
   const { settings: xtmHubSettings } = useLazyLoadQuery<XtmHubSettingsQuery>(
     xtmHubSettingsQuery,
     {},
+    { fetchPolicy: 'network-only' },
   );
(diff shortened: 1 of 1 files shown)


opencti-platform/opencti-front/src/private/components/settings/xtm-hub/XtmHubSettings.tsx
```diff
@@ -40,6 +40,7 @@ const XtmHubSettingsComponent = () => {
   const { settings: xtmHubSettings } = useLazyLoadQuery<XtmHubSettingsQuery>(
     xtmHubSettingsQuery,
     {},
+    { fetchPolicy: 'network-only' },
```
jpkha, line 43: It’s quite strange that changing the fetchPolicy fixes the issue, since the request is actually being sent and returns the correct response.

That said, the solution works for me because we mainly need the correct data during loading.

jpkha (approved):

aHenryJard (approved):

PR #15557, opened 2026-04-17
[ci] add ci for testing-xtm-one (#15597)

### LINKED ISSUE #15597: Add CI to test XTM-One
Switch the test instances from using the main branch / rolling container tag to a new testing-xtm-one branch / tag for each product. The auto-deployment mechanism on commit already exists and works with main/rolling — we need to replicate it for the new branch.

linked to : https://github.com/FiligranHQ/platform/issues/2435

raugier:
Fixes https://github.com/FiligranHQ/platform/issues/2435

fixes #15597

### DIFF
--- a/.github/workflows/ci-testing-xtm-one.yml
+++ b/.github/workflows/ci-testing-xtm-one.yml
@@ -18,3 +18,22 @@ jobs:
         client_python_local: true
         publish_to_registry: true
       secrets: inherit
+
+  wf-redeploy-instance:
+      name: Restart testing-xtm-one instance
+      needs: wf-publish-image
+      runs-on: ubuntu-latest
+      steps:
+        - name: Install AWX cli
+          run: pip install awxkit
+
+        - name: Restart instance via AWX
+          run: |
+            awx --conf.host ${{ secrets.AWX_URL }} \
+              --conf.username ${{ secrets.AWX_USER }} \
+              --conf.password ${{ secrets.AWX_PASSWORD }} \
+              job_templates launch 'Deploy OCTI/OAEV' \
+              --inventory eu-west-production \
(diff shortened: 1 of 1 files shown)


efaure (changes requested):
I think you will also need to ensure the redeployment of the instance. 

Currently, we don't have something similar in github action (because it's done on Circle CI for testing and master-ff). 
I think the best would be to trigger awx job to restart an instance ( `https://awx.filigran.io/#/templates/job_template/169/details`). 
You can get inspiration from the feature branch workflow to see howto launch an AWX job  https://github.com/OpenCTI-Platform/opencti/blob/master/.github/workflows/test-feature-branch.yml 

Do not forget to check that in `filigran-test` the image pull policy is set to `always` in order to ensure the app will be refreshed with a restart

raugier:
> I think you will also need to ensure the redeployment of the instance.
> 
> Currently, we don't have something similar in github action (because it's done on Circle CI for testing and master-ff). I think the best would be to trigger awx job to restart an instance ( `https://awx.filigran.io/#/templates/job_template/169/details`). You can get inspiration from the feature branch workflow to see howto launch an AWX job https://github.com/OpenCTI-Platform/opencti/blob/master/.github/workflows/test-feature-branch.yml
> 
> Do not forget to check that in `filigran-test` the image pull policy is set to `always` in order to ensure the app will be refreshed with a restart

Added the instance redeployment (filigran-test-octi is on eu-west-production so I had to use the Upgrade OCTI/OAEV job).
Thanks for the reminder about the pull policy, just checked and it's already on "always" :+1:

efaure (approved):
It seems good. We'll need to merge it to try it !

Archidoit (approved):

Archidoit:
fixes #15597

PR #15565, opened 2026-04-20
[frontend][backend] Multiple fixes for custom views (#13389)

fellowseb:
### Proposed changes

This PR brings a few small fixes to the custom views feature branch:

* DataModel : remove `CustomViewInfoForSettings` and `CustomViewDisplay`  models to use `CustomView` instead. I originally created those models as different "views" of the same CustomView entity; but I didn't have in mind how GraphQL & Relay works (i.e. the frontend Relay store requiring unique {ID, type} pairs for caching).
* Allow null and undefined in local data table sorting hook
* Add icon in custom views table as per the specs
* Fix custom view tabs styling issues correctly by passing props to underlying `<Tab>` compoennt
* Fix console error due to invalid tab value in settings before redirection by using the `false` value in that case
* Remove unused authorized_members field as it deprioritized
* Custom views are sorted by name when displayed

### Related issues

### Checklist

### Further comments


opencti-platform/opencti-front/src/private/components/settings/sub_types/SubTypeMenu.tsx
```diff
@@ -16,16 +17,18 @@ const SubTypeMenu = ({ entityType, tabs }: SubTypeMenuProps) => {
 
   if (!hasAtLeastOneEnabledTab) return null;
 
+  const currentTab = getCurrentTab(location.pathname, `/dashboard/settings/customization/entity_types/${entityType}`);
+
   return (
     <Tabs
-      value={location.pathname}
+      value={currentTab ?? 'workflow'}
```
esrevi: The `currentTab ?? 'workflow'` fallback could break if `tabs.workflow` is disabled, MUI's `<Tabs>` would have no matching `<Tab>` for that value

opencti-platform/opencti-graphql/src/generated/graphql.ts
```diff
@@ -6443,43 +6442,23 @@ export type CustomView = BasicObject & InternalObject & {
...
-  description: Scalars['String']['output'];
-  id: Scalars['ID']['output'];
-  name: Scalars['String']['output'];
-  updated_at: Scalars['DateTime']['output'];
-};
-
 export type CustomViewsDisplayContext = {
   __typename?: 'CustomViewsDisplayContext';
-  custom_views_info: Array<CustomViewDisplayContext>;
+  custom_views_info: Array<CustomView>;
```
esrevi: is the name still relevant since it returns an array of CustomView? should this be customViews instead? 

just a notice, some types are in snake cases like here, but in CustomViewsSettings, we have customViews in camelCase. Maybe we could stick to that convention: https://www.apollographql.com/docs/graphos/schema-design/guides/naming-conventions (at least for the new fields ^^)

opencti-platform/opencti-graphql/src/modules/customView/customView-domain.ts
```diff
@@ -45,19 +45,24 @@ function isCustomViewsAvailableForEntityType(entityType: string) {
...
+
 // View Use Cases (all authed users)
 
 export const getCustomViewByIdForDisplay = async (
   context: AuthContext,
   user: AuthUser,
   customViewId: string,
 ) => {
-  return storeLoadById<BasicStoreEntityCustomView>(
+  const entity = await storeLoadById<BasicStoreEntityCustomView>(
```
esrevi: do we need that constant ? it is more straightforward to return the function otherwise

opencti-platform/opencti-graphql/src/modules/customView/customView-domain.ts
```diff
@@ -67,14 +72,11 @@ export const getCustomViewsDisplayContext = async (context: AuthContext, user: A
     [ENTITY_TYPE_CUSTOM_VIEW],
   );
   const customViewInfoMap = allCustomViewEntities.reduce((infoMap, customViewEntity) => {
-    const { id, name, slug, target_entity_type: targetEntityType } = customViewEntity;
-    const infos = infoMap.get(targetEntityType) ?? [];
-    // Build the relative path from the slug and the id
-    const path = `${slug}-${id.replaceAll('-', '')}`;
-    infos.push({ id, path, name });
-    infoMap.set(targetEntityType, infos);
+    const infos = infoMap.get(customViewEntity.target_entity_type) ?? [];
```
esrevi: personal preference, but I find reduce is harder to read than a for ... of :p 

suggestion, I think we could also make in one pass instead of two maps with filter and grouping like

```
for each entity
  if not available, skip
  else add to group```

File: opencti-platform/opencti-front/src/private/components/custom_views/useCustomViews.ts [hunk 5]
```diff
@@ -27,8 +27,11 @@ export const useCustomViews = (entityType: string) => {
   }
   const customViews = customViewsContextForType.custom_views_info ?? [];
   const getCurrentCustomViewTab = matchPath(customViews);
+  const sortedCustomViews = [...customViews].sort(
```
esrevi: do you think this could be sorted from the query?

opencti-platform/opencti-front/src/private/components/custom_views/CustomViewTab.tsx
```diff
@@ -1,22 +1,24 @@
...
+} & TabProps<'a'> & TabProps<'div'>;
 
-const CustomViewTab = ({ customViews, displayMode, dropDownMenuState, value }: CustomViewTabProps) => {
+const CustomViewTab = ({ customViews, displayMode, dropDownMenuState, value, ...tabProps }: CustomViewTabProps) => {
   const { t_i18n } = useFormatter();
   if (displayMode === 'single') {
     return (
       <Tab
+        {...tabProps}
+        key="custom-view"
```
esrevi: why do we have to set the key here ?

opencti-platform/opencti-front/src/private/components/custom_views/useCustomViews.ts
fellowseb: True, we could. It would indeed be more performant.

But it would mean running more logic in the `relay store updater` code in the frontend when adding/editing a custom view (appears in my next PR). Currently I only run the store updater when adding a new item. If we did the sorting server-side, we would need to make sure the updater keeps the sorting place. Also we would need to had an updater when a custom view gets edited (as the name could change). Let's discuss offline!

opencti-platform/opencti-front/src/private/components/custom_views/CustomViewTab.tsx
fellowseb: not needed, removing

opencti-platform/opencti-graphql/src/generated/graphql.ts
fellowseb: Yeah I see contradictions in every module / API I look at, made me sad but I guess it infected me too !
I got confused~~

opencti-platform/opencti-front/src/private/components/settings/sub_types/SubTypeMenu.tsx
fellowseb: True, plus it's just wrong. Sorry, what I put elsewhere is: `currentTab || false`

opencti-platform/opencti-graphql/src/modules/customView/customView-domain.ts
```diff
@@ -45,19 +45,24 @@ function isCustomViewsAvailableForEntityType(entityType: string) {
     && !ENTITY_TYPES_WITHOUT_CUSTOM_VIEWS.includes(entityType);
 }
 
+export function computeCustomViewPath({ slug, id }: BasicStoreEntityCustomView) {
+  return `${slug}-${id.replaceAll('-', '')}`;
```
Archidoit, line 62: we do we replace '-' by ''  in the path for ids ?

opencti-platform/opencti-front/src/private/components/custom_views/useCustomViews.ts
Archidoit: I agree with you @esrevi , sorting should be done in backend, like for all other listings

Archidoit (changes requested):

opencti-platform/opencti-front/src/private/components/settings/sub_types/SubTypeMenu.tsx
fellowseb: Fixed

opencti-platform/opencti-graphql/src/generated/graphql.ts
fellowseb: Changed. I still don't like that there's some hidden magic doing the translation from snake_case (ES attributes) to camelCase (GraphQL api).

opencti-platform/opencti-graphql/src/modules/customView/customView-domain.ts
fellowseb: Side-effect of the experiments with Graphql. Changed

opencti-platform/opencti-front/src/private/components/settings/sub_types/SubTypeMenu.tsx
```diff
@@ -16,16 +17,18 @@ const SubTypeMenu = ({ entityType, tabs }: SubTypeMenuProps) => {
...
       sx={{ paddingBottom: 2 }}
```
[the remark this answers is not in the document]
fellowseb: Already addressed

opencti-platform/opencti-front/src/private/components/custom_views/useCustomViews.ts
fellowseb: Done, done the sorting in the backend in the paginated `customViews` query

opencti-platform/opencti-front/src/private/components/custom_views/CustomViewTab.tsx
fellowseb: Done

opencti-platform/opencti-graphql/src/modules/customView/customView-domain.ts
```diff
@@ -67,14 +72,11 @@ export const getCustomViewsDisplayContext = async (context: AuthContext, user: A
     [ENTITY_TYPE_CUSTOM_VIEW],
   );
```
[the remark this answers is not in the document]
fellowseb: This function got removed to use a more generic `customViews` query instead. This query isn't specifically intented to display context anymore so it's open for consumers to query everything, including the manifest, at their own risk performance-wise.

opencti-platform/opencti-graphql/src/modules/customView/customView-domain.ts
fellowseb: Good question. The spec was to do "like Notion", but I guess it's a little incoherent with the fact we show UUIDs with the hyphens in URLs for everything else.
Asking Product then I'll change depending on feedback, but in another PR (this one already does too many things)

opencti-platform/opencti-graphql/src/modules/customView/customView-domain.ts
fellowseb: Got removed with the removal of `customViewsDisplayContext`

PR #15569, opened 2026-04-20
[frontend][backend] Enable/disable a custom view (#13389)

fellowseb:
### Proposed changes

Implements "Enable/disable custom view" chunk:

* Added a `enabled` field on the data model: boolean, not mandatory but the field is present (mandatory) on API and STIX output
* By default set to `false` on creation & duplication
* The settings datatable show the status in a dedicated column : "View is disabled" or "View is enabled"
* A quick action is available directly on the line
* On the CustomViewEdition page a icon button allows the same and the status is repeated
* The filtering is made server-side for the other query

### Related issues

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/settings/sub_types/custom_views/CustomViewEditionHeader.tsx
+++ b/opencti-platform/opencti-front/src/private/components/settings/sub_types/custom_views/CustomViewEditionHeader.tsx
@@ -1,11 +1,13 @@
 import { useState } from 'react';
 import { graphql, useFragment } from 'react-relay';
-import { Box, Tooltip } from '@mui/material';
+import Box from '@mui/material/Box';
+import Tooltip from '@mui/material/Tooltip';
 import Button from '@common/button/Button';
 import Tag from '@common/tag/Tag';
 import { useTheme } from '@mui/styles';
 import VisibilityIcon from '@mui/icons-material/Visibility';
 import VisibilityOffIcon from '@mui/icons-material/VisibilityOff';
+import IconButton from '@common/button/IconButton';
 import TitleMainEntity from '../../../../../components/common/typography/TitleMainEntity';
 import { CustomViewEditionHeader_customView$key } from './__generated__/CustomViewEditionHeader_customView.graphql';
 import Breadcrumbs from '../../../../../components/Breadcrumbs';
@@ -84,9 +86,9 @@ const CustomViewEditionHeader = ({ data, onCreateWidget, onImportWidget }: Custo
         />
         <Box sx={{ display: 'flex', alignItems: 'center', marginLeft: 'auto', gap: 1 }}>
           <Tooltip title={customView.enabled ? t_i18n('Disable') : t_i18n('Enable')}>
-            <Button variant="secondary" iconOnly disabled={mutating} onClick={handleToggleEnabled}>
+            <IconButton variant="secondary" size="default" disabled={mutating} onClick={handleToggleEnabled}>
               {customView.enabled ? <VisibilityOffIcon /> : <VisibilityIcon />}
-            </Button>
+            </IconButton>
           </Tooltip>
           <CustomViewMenu data={customView} />
           <DashboardWidgetConfig
--- a/opencti-platform/opencti-front/src/private/components/settings/sub_types/custom_views/CustomViewEditionHeader.tsx
+++ b/opencti-platform/opencti-front/src/private/components/settings/sub_types/custom_views/CustomViewEditionHeader.tsx
@@ -1,7 +1,12 @@
 import { useState } from 'react';
 import { graphql, useFragment } from 'react-relay';
-import { Box, Typography } from '@mui/material';
+import { Box, Tooltip } from '@mui/material';
 import Button from '@common/button/Button';
+import Tag from '@common/tag/Tag';
+import { useTheme } from '@mui/styles';
+import PlayArrowIcon from '@mui/icons-material/PlayArrow';
+import StopIcon from '@mui/icons-material/Stop';
+import TitleMainEntity from '../../../../../components/common/typography/TitleMainEntity';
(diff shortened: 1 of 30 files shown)


opencti-platform/opencti-graphql/src/modules/customView/customView-converter.ts
```diff
@@ -12,6 +13,8 @@ const convertCustomViewToStix = (instance: StoreEntityCustomView): StixCustomVie
     description: instance.description,
     manifest: instance.manifest,
     target_entity_type: instance.target_entity_type,
+    enabled: Boolean(instance.enabled),
```
esrevi, line 16: isn't instance.enabled already a boolean? or could it be undefined ? if so, do you know why we don't have to set instance?.enabled instead?

opencti-platform/opencti-front/src/private/components/settings/sub_types/custom_views/CustomViewEditionHeader.tsx
```diff
@@ -42,16 +52,42 @@ const CustomViewEditionHeader = ({ data, onCreateWidget, onImportWidget }: Custo
...
+          label={
+            customView.enabled
+              ? t_i18n('View is enabled')
+              : t_i18n('View is disabled')
+          }
+          labelTextTransform="none"
+        />
         <Box sx={{ display: 'flex', alignItems: 'center', marginLeft: 'auto', gap: 1 }}>
+          <Tooltip title={customView.enabled ? t_i18n('Disable') : t_i18n('Enable')}>
+            <Button variant="secondary" iconOnly disabled={mutating} onClick={handleToggleEnabled}>
```
esrevi: I think you can use IconButton here instead of Button

opencti-platform/opencti-front/src/private/components/settings/sub_types/custom_views/CustomViewForm.tsx
```diff
@@ -36,13 +37,17 @@ const CustomViewForm = ({
...
+  };
+
+  const validation = Yup.object().shape(validators);
 
-  const handleFieldSubmit = (setSubmitting: (v: boolean) => void) => (name: string, value: unknown) => {
-    onSubmitField(name, value);
+  const handleFieldSubmit = (
+    setSubmitting: (v: boolean) => void,
+  ) => (name: keyof typeof validators, value: unknown) => {
+    onSubmitField(name, validators[name].cast(value));
```
esrevi: i wonder if onSubmitField handler passed should be a promise or not? because setSubmitting is false without waiting the response of that call no?

opencti-platform/opencti-graphql/src/modules/customView/customView-converter.ts
fellowseb: I hesitated but went with `enabled` possibly `undefined`; but no the optional chaining operator would be necessary if `instance` would be `undefined`, which is not the case hehre :)

opencti-platform/opencti-front/src/private/components/settings/sub_types/custom_views/CustomViewForm.tsx
fellowseb: What I can do is mimic what's done for `onSubmit` and pass the function to the parent so they would call `setSubmitting(false)` in the completion callbacks.

opencti-platform/opencti-front/src/private/components/settings/sub_types/custom_views/CustomViewEditionHeader.tsx
fellowseb: Done

opencti-platform/opencti-front/src/private/components/settings/sub_types/custom_views/CustomViewForm.tsx
fellowseb: Done

opencti-platform/opencti-front/src/private/components/settings/sub_types/custom_views/CustomViewFormDrawer.tsx
```diff
@@ -33,7 +33,10 @@ const CustomViewFormDrawer = ({
     values,
     { setSubmitting, resetForm },
   ) => {
-    if (isEdition) return;
+    if (isEdition) {
```
esrevi, line 36: if (isEdition && isSubmitting)

opencti-platform/opencti-front/src/private/components/settings/sub_types/custom_views/CustomViewFormDrawer.tsx
```diff
@@ -58,12 +61,23 @@ const CustomViewFormDrawer = ({
...
   };
 
-  const handleEditField = (field: string, value: unknown) => {
-    if (!isEdition) return;
+  const handleEditField = (
+    field: string,
+    value: unknown,
+    { setSubmitting }: { setSubmitting: (isSubmitting: boolean) => void },
+  ) => {
+    if (!isEdition) {
```
esrevi, line 69: if (!isEdition && isSubmitting) {

opencti-platform/opencti-front/src/private/components/settings/sub_types/custom_views/CustomViewFormDrawer.tsx
fellowseb: Hum I'd rather not as even if `!isSubmitting` (which shouldn't happen) we only want to call the customViewAdd mutation endpoint only when not in edition mode

opencti-platform/opencti-front/src/private/components/settings/sub_types/custom_views/CustomViewFormDrawer.tsx
fellowseb: Hum I'd rather not as even if `!isSubmitting` (which shouldn't happen) we only want to call the customViewEdit mutation endpoint only when not in creation mode

PR #15570, opened 2026-04-20
[frontend][backend] Mark a Custom view as default to use it as the index on the entity page (#13389)

fellowseb:
### Proposed changes

Implements second half of the dedicated chunk that sets a Custom View as `default`:

* Added a `default` boolean field in the data model. Like `enabled` it's optional but mandatory in ougoing types (API & STIX), defaults to `false`.
* Adds a `SwitchField` in the update drawer
* Behavior for end-users: the default custom view is displayed as a first Tab in the SDO page. Other custom views remain at the right
* There's always one default custom view at any time. This constraint is guaranteed at the backend domain level (not possible at the DB level). Supported the eddition/creation/duplication cases.
* Added integration test case

### Related issues

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/custom_views/CustomViewTabs.test.tsx
+++ b/opencti-platform/opencti-front/src/private/components/custom_views/CustomViewTabs.test.tsx
@@ -259,7 +259,7 @@ describe('useCustomViewTabs', () => {
     );
     const defaultTabElem = screen.getByRole('tab', { name: /My default custom view/i });
     expect(defaultTabElem).toBeInTheDocument();
-    const othersTabElem = screen.getByRole('tab', { name: /Custom view/i });
+    const othersTabElem = screen.getByRole('tab', { name: /^Custom view$/i });
     expect(othersTabElem).toBeInTheDocument();
     await user.click(othersTabElem);
     const firstLinkElem = screen.getByRole('link', { name: /My first custom view/i });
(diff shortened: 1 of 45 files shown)


opencti-platform/opencti-front/src/private/components/custom_views/CustomViewTabs.test.tsx
```diff
@@ -44,19 +45,34 @@ const TestWrapper = ({ entityType, basePath }: TestWrapperProps) => {
     </MenuItem>
   ));
 
-  const renderCustomViewTab = () => {
-    if (displayMode === 'single') {
+  const renderDefaultCustomViewTab = () => {
+    if (defaultCustomView) {
```
esrevi: prefer early return if (!defaultCustomView) return null

fellowseb: Done

opencti-platform/opencti-front/src/private/components/custom_views/CustomViewTabs.test.tsx
```diff
@@ -46,18 +46,14 @@ const TestWrapper = ({ entityType, basePath }: TestWrapperProps) => {
...
-          component={Link}
-          to={defaultCustomView.path}
-          value={DEFAULT_CUSTOM_VIEW_TAB_VALUE}
-          label={defaultCustomView.name}
-        />
-      );
-    }
-
-    return null;
+    return defaultCustomView ? (
```
esrevi, line 49: i think we misunderstood, i suggested early return 
if (!defaultCustomView) return null

return <Tab ... />

but it doesn't matter a lot here ^^

esrevi (approved):

PR #15582, opened 2026-04-20
fix(backend): [Investigation] fix non-idempotent operation during concurrent stream sync (#15207)

### LINKED ISSUE #15207: fix: ensure report rel_object.internal_id values are deduplicated
## Description

The ingestion should deduplicate rel_object.internal_ids, but an account reported duplicated values [here](https://filigran.atlassian.net/browse/OCTI1-3145)

## Environment

- OpenCTI version: 6.9.20

## Reproducible Steps

No steps provided, analyze the Reports ingestion/creation code and try to find where a `rel_object.internal_id` field could eventually be created with multiple duplicated values, and to prevent it. If nothing comes up, close this issue until we get an actual repro case.

## Screenshots (optional)

OctaveLaventure:
### Proposed changes

* prevent relation duplication with parallel workers on ADD operations

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/15207

### Checklist

### Further comments

**Root cause**: `updateAttributeMetaResolved`'s `UPDATE_OPERATION_ADD` branch computed which refs to create by diffing against `updatedInstance[key]`, an in-memory snapshot that could be stale relative to concurrent writers, even though the write itself was correctly serialized by the function's existing lock. Two concurrent "add ref" calls targeting the same container (e.g. two workers processing the same STIX bundle) could each diff against pre-lock data and both decide the ref was missing, producing duplicate relationship documents.

**Fix**: after the lock is held, re-fetch the current relations for just the candidate targets (`fullRelationsList` scoped by `toId` + `baseData: true` for a cheap, indexed lookup) and diff against that instead of the stale snapshot. Also dedupe the emitted event payload so downstream stream-consumers (e.g. cross-instance stream-sync) never replay a literal duplicate `add` operation.

**Investigation**: verified via direct ES aggregation on `stix_meta_relationships` (not just UI) across a two-instance concurrent-write repro — 0 duplicates across 10 reports with this fix alone, vs. reproducible duplication without it.

Also bisected against a broader set of candidate fixes (upstream lock-before-load hardening in `internalCreateEntityRaw`/`stixObjectOrStixRelationship.ts`, dedup guards in `buildInnerRelation`/`elUpdateEntityConnections`, and more) — none of those were necessary once this fix was applied, since every ADD-path caller funnels through this one function's lock, which already serializes on the target entity's stable identity. Also, none were fixing the duplicating issue by themselves without this fix. It may be possible to optimize this ingestion process with a broader refactoring but it pulls many strings, far beyond this bug's scope. I suggest that we release this slightly less performant but understandable fix and keep the global refactor for another issue.


opencti-platform/opencti-graphql/src/database/middleware.ts
```diff
@@ -2605,9 +2605,14 @@ export const updateAttributeMetaResolved = async <T extends StoreObject>(
...
-          const currentIds = filteredList.map((o: any) => [o.id, o.standard_id]).flat();
-          const refsToCreate = refs.filter((r) => !currentIds.includes(r.internal_id));
+          // Re-evaluate existing refs from storage while we hold the lock.
+          // The loaded instance can be stale when concurrent workers update the same object.
+          const currentRels = await fullRelationsList(context, user, relType, {
+            indices: READ_RELATIONSHIPS_INDICES_WITHOUT_INFERRED,
+            fromId: initial.internal_id,
+          });
+          const currentRelToIds = new Set(currentRels.map((rel: BasicStoreRelation) => rel.toId));
+          const refsToCreate = refs.filter((r) => !currentRelToIds.has(r.internal_id));
```
JeremyCloarec: I agree that this would solve the issue, but I think we could go for a different solution to prevent having to load all relations here, and save some perf this way

The proper & ideal solution would be to only load the initial entity once, after the lock is acquired, and lock the entity earlier if the fully loaded entity is needed (in upsertElement for exemple). This requires a lot more refactoring though, but I think it's worth the cost

OctaveLaventure: I agree, I will explore this option, thanks!

delemaf:
On the testing part it's working, I've no duplicates ref, either in events UI or elastic ref_object ✅

gabrielebonadiman (approved):

PR #15592, opened 2026-04-20
[backend] encode to base64 the embedded storage links (#7281)

### LINKED ISSUE #7281: Be able to easily copy/paste images, upload etc. in markdown editor (+ more capabilities)
## Use case

Be able to easily copy/paste images, upload etc. in markdown editor (+ more capabilities)

Inspiration: GitHub issue form.

esrevi:
### Proposed changes

* Ensure STIX export converts markdown embedded image links from OpenCTI storage endpoints into inline base64 so exported content remains self-contained.
* Preserve backward compatibility and export by only rewriting targeted markdown fields during export and leaving original links unchanged when conversion cannot be completed
* These changes are needed for an upcoming branch where we want to export/import images from description fields that are saved in the storage

### Related issues

* Resolves #7281 

### Checklist


client-python/pycti/utils/opencti_stix2.py
```diff
@@ -166,6 +173,158 @@ def convert_markdown(self, text: str) -> str:
...
+                    storage_url,
+                    binary=True,
+                    serialize=True,
+                )
+                if not encoded_data:
+                    # Some HTTP stacks reject raw spaces/special chars in path; retry with encoded path.
+                    encoded_storage_path = quote(embedded_storage_path, safe="/")
+                    encoded_storage_url = urljoin(
+                        self.opencti.api_url.replace("graphql", ""),
+                        f"storage/get/{encoded_storage_path}",
```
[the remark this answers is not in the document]
esrevi: Not changing that, because some integration tests were failing when updating this to /graphql

client-python/pycti/utils/opencti_stix2.py
```diff
@@ -61,6 +63,11 @@
 STIX_EXT_MITRE: str = "extension-definition--322b8f77-262a-4cb8-a915-1e441e00329b"
 PROCESSING_COUNT: int = 4
 MAX_PROCESSING_COUNT: int = 100
+MARKDOWN_EXPORT_FIELDS: Tuple[str, ...] = (
+    "description",
+    "x_opencti_description",
+    "content",
```
JeremyCloarec, line 72: Is it intended for content to be here? I thought content was already handled in an other way?

client-python/pycti/utils/opencti_stix2.py
```diff
@@ -166,6 +173,158 @@ def convert_markdown(self, text: str) -> str:
...
+        descriptions = entity.get("descriptions")
+        if isinstance(descriptions, list):
+            entity["descriptions"] = [
+                (
+                    self._rewrite_markdown_embedded_storage_images(value)
+                    if isinstance(value, str)
+                    else value
+                )
+                for value in descriptions
+            ]
```
JeremyCloarec: What's this descriptions with an 's'? I think it can be removed? Or if necessary, add it to MARKDOWN_EXPORT_FIELDS

client-python/pycti/utils/opencti_stix2.py
```diff
@@ -166,6 +173,158 @@ def convert_markdown(self, text: str) -> str:
...
+        markdown_length = len(markdown)
+
+        while cursor < markdown_length:
+            image_start = markdown.find("![", cursor)
+            if image_start == -1:
+                rewritten_chunks.append(markdown[cursor:])
+                break
+
+            rewritten_chunks.append(markdown[cursor:image_start])
+            alt_end = markdown.find("]", image_start + 2)
```
JeremyCloarec: A bit of an edge case, but this doesn't work if the text contains an escaped ], like so ![A text like \\[this\\]]

client-python/pycti/utils/opencti_stix2.py
```diff
@@ -166,6 +173,158 @@ def convert_markdown(self, text: str) -> str:
...
+            if embedded_storage_path is None:
+                return full_match
+
+            guessed_mime_type = mimetypes.guess_type(embedded_storage_path)[0]
+            if guessed_mime_type is None:
+                return full_match
+            if guessed_mime_type is not None and not guessed_mime_type.startswith(
+                "image/"
+            ):
+                return full_match
```
JeremyCloarec: Was there a reason to check mime_type? I feel like it's overly defensive and might break integration with some images

client-python/pycti/utils/opencti_stix2.py
```diff
@@ -166,6 +173,158 @@ def convert_markdown(self, text: str) -> str:
...
+                mime_type = (
+                    guessed_mime_type
+                    or mimetypes.guess_type(embedded_storage_path)[0]
+                    or "application/octet-stream"
+                )
+                if not mime_type.startswith("image/"):
+                    return full_match
+                embedded_data_uri_by_path[embedded_storage_path] = (
+                    f"data:{mime_type};base64,{encoded_data}"
+                )
```
JeremyCloarec, line 249: I think we should refactor this to reuse existing logic that handles importFiles 
Image

client-python/pycti/utils/opencti_stix2.py
esrevi: actually, for a Note, when we export in stix format, the note's description is set in "content" key so that's the reason this field is concerned

client-python/pycti/utils/opencti_stix2.py
esrevi: actually, it seems that stixCyberObservable has a "descriptions" field in ENTITY_PROCESS (windows-service-ext) so this case could happen I think

client-python/pycti/utils/opencti_stix2.py
esrevi: nice catch ^^

client-python/pycti/utils/opencti_stix2.py
esrevi: I'm not sure about being able to use that function, because contrary of x_opencti_file where we have more file's metadata while the code i added only receives markdown URLs and parse, normalize them first, and the output is also different because importFiles exports attachments in x_opencti_files and in payload_bin but here, we want to rewrite markdown iimage URLs to data encoded, but i may miss something :)

client-python/pycti/utils/opencti_stix2.py
esrevi: ok, i remove that check, it was there because we only handle imported images in markdown

JeremyCloarec (approved):

PR #15596, opened 2026-04-21
feat(backend): allow deletion of `UPDATE_LINKED` in draft (#14358)

### LINKED ISSUE #14358: fix(draft-workspace): "update_linked" actions cannot be reverted when removed
## Description

In Drafts, entity creation and modification actions can be removed/reverted correctly. These actions can be selected in the Draft overview and deleted using the trash icon, which effectively reverts the changes applied to the platform.

However, any action categorized as "update_linked" cannot be reverted this way.

## Environment

OCTI 6.9.13

## Reproducible Steps

1. Create a Draft and open it
2. Create a Report
3. Modify a Report
4. In the Draft overview, select both Report (the update and the create) and click on the trash icon to remove them
-> This will revert the modification, i.e. delete the newly created report and revert the update on the existing one

Image

6. Now change the Author on a Report
7. In the Draft overview, select the Report (the "update_linked") and click on the trash icon to revert the modification
-> Nothing will happen, i.e. the Author will not be reverted to the previous one

Image

## Expected Output

Removing the action from the draft should revert the "linked_update" and restore the previous value (e.g., the original author).

## Actual Output

Nothing happens when deleting the "update_linked" action from the draft. The "linked_update" remains in the draft and the modified value (e.g., author) stays changed on the entity.

The only way to revert the change is to manually edit the entity and restore the previous value.

delemaf:
### Proposed changes

* Allow and handle deletion of `UPDATE_LINKED` in draft.

### Related issues

* #14358 

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/database/draft-engine.ts
+++ b/opencti-platform/opencti-graphql/src/database/draft-engine.ts
@@ -133,6 +133,83 @@ const removeDraftDeleteLinkedRelations = async (
   await completeDeleteElementsFromDraft(context, user, deleteLinkedRelationsInstances);
 };
 
+const elRemoveUpdateLinkedElementFromDraft = async (context: AuthContext, user: AuthUser, element: BasicStoreCommon): Promise<void> => {
+  if (element.draft_change?.draft_operation !== DRAFT_OPERATION_UPDATE_LINKED) {
+    return;
+  }
+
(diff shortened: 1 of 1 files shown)


JeremyCloarec:
Has it been discussed with product? When drafts were first implemented, we agreed that update_linked couldn't be removed because there is no way for the user to know which relations it's going to delete from the draft when the update_linked is removed. 
Is product team OK with the behavior implemented here, that removing an update_linked will silently remove relations from the draft?

PR #15610, opened 2026-04-21
build(frontend): replace custom esbuild-based build script by Vite and migrate front to ESM (#14044)

### LINKED ISSUE #14044: chore(frontend): migrate frontend to ESM
It's needed to modernize the stack and unlock some possibilities (eg be able to remove pdf worker js file from sources)
As a consequence it lead to removing production build with esbuild and use vite instead

### LINKED ISSUE #16608: fix(frontend): production build emits a single ~50 MB JS bundle — esbuild missing code splitting
### Summary

The production frontend build (`yarn build`) emits **a single, monolithic JavaScript bundle of ~50 MB** (uncompressed), despite the codebase using `React.lazy()` + `Suspense` extensively for route-level code splitting. As a result, every user — even one who only opens the login page or a single dashboard — is forced to download, parse, and execute the entire application (including heavy, rarely-used dependencies) before anything renders. This severely impacts Time to First Byte (TTFB), Time to Interactive (TTI), and bandwidth usage, especially on constrained networks.

### Environment

- **OpenCTI version:** `7.260423.0` (front package)
- **OS / Node:** Node ≥ 20
- **Build tool:** esbuild `0.28.0` (production), Vite `8.0.9` (dev only)

### Observed behavior

Running `yarn build` in `opencti-platform/opencti-front` produces a single chunk:

```
builder/prod/build/static/js/front-[hash].js   ~50 MB (minified)
```

`index.html` generated by the builder references exactly one `<script>` tag:

```html
<script defer="defer" src="/static/js/front-[hash].js"></script>
```

No additional chunks are emitted for the dozens of lazy-loaded route modules.

### Expected behavior

Dynamic `import()` calls (via `React.lazy()`) should produce separate, on-demand chunks so that:

- The initial bundle contains only the shell + login + dashboard
- Heavy feature modules (observations, cases, workspaces, graph explorers, etc.) load on navigation
- Heavy third-party libraries (see below) are isolated into vendor chunks and fetched only when needed

### Root cause

The production build is driven by `opencti-platform/opencti-front/builder/prod/prod.js`, which configures esbuild **without enabling code splitting**:

```js
// builder/prod/prod.js
await esbuild.build({
  entryPoints: ["src/front.tsx"],
  bundle: true,
  minify: true,
  sourcemap: true,
  target: ["chrome58"],
  // ❌ No `splitting: true`
  // ❌ No `format: "esm"` (required by esbuild for splitting)
  // ❌ No manual `metafile` analysis / chunking strategy
});
```

esbuild **only** splits dynamic `import()` expressions into separate chunks when `splitting: true` is set **and** the output `format` is `esm`. Neither is present, so all lazy imports are inlined into the single entry bundle. In other words, the existing `React.lazy()` infrastructure (e.g. `src/private/components/cases/Root.tsx` and ~67 other `Root.tsx` route files) is effectively dead from a performance perspective in production.

### Why this matters — the heavy hitters bundled into every load

The bundle contains several large dependencies that are only used by specific features/routes, yet are shipped to every visitor:

| Dependency | Used by | Approx. weight |
| --- | --- | --- |
| `monaco-editor` + `monaco-graphql` | GraphQL playground only | ~5–8 MB |
| `react-force-graph-3d` (+ `three`) | Knowledge graph 3D view | ~1–2 MB |
| `ckeditor5` | Rich text editor | ~2–3 MB |
| `pdfmake` + `html-to-pdfmake` | PDF export | ~1 MB |
| `react-pdf` | PDF viewer | ~1 MB |
| `@tiptap/*` (27 extensions) | TipTap editor | ~1 MB |
| `reactflow`, `recharts`, `apexcharts` | Charts / diagrams | ~1–2 MB combined |
| `react-syntax-highlighter` | Code blocks | ~500 KB+ |
| Moment + moment-timezone (legacy) | Date handling | ~300 KB |

Optimistically, splitting these out could reduce the initial bundle by **70–85%**, bringing first-load JS into the low single-digit MB range.

### Additional concern

The bundle also ships `moment` + `moment-timezone` alongside `date-fns` + `date-fns-jalali` — i.e. **three** date libraries. Migrating the remaining moment usage to `date-fns` would remove a sizable chunk from the single bundle.

### Proposed solutions (non-exhaustive)

1. **Enable code splitting in esbuild**
   ```js
   await esbuild.build({
     // ...
     splitting: true,
     format: "esm",
     metafile: true,            // enables bundle analysis
     outExtension: { ".js": ".mjs" },
   });
   ```
   Note: this requires the generated `index.html` script tags to use `type="module"`.

2. **Migrate the production build to Vite**
   `vite.config.mts` is already present and configured for dev. Vite natively produces optimized, split chunks (with manualChunks support), which would honor the existing `React.lazy()` calls out of the box and unify dev/prod tooling.

3. **Add manual chunking** for known heavy vendors (monaco, ckeditor5, force-graph-3d, pdfmake) so they can be cached independently and prefetched on demand.

4. **Audit dependencies** (moment vs date-fns, duplicate charting libs) to shrink the total surface before splitting.

5. **Add a bundle-size budget** (e.g. via `esbuild` metafile analysis or `size-limit`) to CI to prevent regression.

### Steps to reproduce

1. `cd opencti-platform/opencti-front`
2. `yarn install && yarn build`
3. Inspect `opencti-platform/opencti-graphql/public/static/js/`
4. Observe a single multi-MB JS file (e.g. `front-[hash].js` ≈ 50 MB unminified / several MB gzipped).
5. Open the deployed app in a browser → DevTools → Network → confirm the entire bundle loads before first paint.

### Related context

- `vite.config.mts` already lists `optimizeDeps.include` for the heavy libs above, indicating awareness of their cost — but these optimizations only apply to dev server, not the production esbuild pipeline.

### Willingness to contribute

I'd be happy to open a PR implementing option (1) and/or (2) with the corresponding `index.html` changes, plus a bundle-size report before/after. Please let me know which direction the maintainers prefer.

xfournet:
Fix #14044 
Fix #16608
Partially handle #16048

* Changes
  * Removed custom ESBuild scripts, now using Vite to create production builds
     * includes removal of no more needed dependencies
     * includes removal of monaco-editor worker support for ESBuild
     * adapt the script-stack-analysis.js script
     * adapt Dockerfiles
  * Switch opencti-front module to ESM
    * rename *.mjs/mts files -> *.js/ts
    * update scripts from require to import
    * remove CJS  __dirname usage in e2e scripts
    * need to add inconsistentCjsInterop config in Vite to support some CJS modules
    * take the opportunity to remove the committed pdf worker JS file, replaced by NPM package
    * take the opportunity to leverage new ESM/Vite feature to handle flags image directly in the front end
  * Migrate eslint-plugin-custom-rules to ESM
  * Bonus: since we have to retest GraphiQL, I aligned default theme of GraphiQL with platform default theme

* Test plan: must be executed twice, one time with Vite dev server (port 3000) and one time with Vite production build (yarn build in front, then use port 4000)
  * Smoke test on UI, check for graph, editors (HTML, MD)
  * Playground (Monaco editor worker), eg in http://localhost:3000/public/graphql
    * test basic behavior
    * ensure autocompletion is working (that involve the worker)
  * PDFViewer: eg in http://localhost:3000/dashboard/settings/customization/fintel_designs
    *  create a new template
    * check the PDF preview
  * country flags, eg in http://localhost:3000/dashboard/locations/countries
    * create a new country
    * add an alias that match the 2 letter country flag (eg `fr` from France)
    * ensure the expected flags in displayed in the country list
  * test the script-stack-analysis.js script


opencti-platform/opencti-graphql/script/script-stack-analysis.js
```diff
@@ -1,19 +1,18 @@
...
- * For the frontend, <yarn build:standalone> in opencti-front
+ * For the frontend, <yarn build> in opencti-front
  * Then you just have to put your json log of the error in a .env files at the root directory of opencti-graphql
  * .env => backend_log='{...}' or frontend_log='{...}'
  * and start the script yarn stack:analysis
  */
 
 const BACKEND_MAP = './build/back.js.map';
-const FRONT_MAP = '../opencti-front/builder/prod/build/static/js/front.js.map';
+const FRONT_MAP = '../opencti-front/dist/assets/front.js.map';
```
xfournet: path is wrong, need to be updated

opencti-platform/opencti-front/vite.config.ts
```diff
@@ -0,0 +1,98 @@
+import { defineConfig } from 'vite';
+import react from '@vitejs/plugin-react';
+import * as path from 'node:path';
+import relay from 'vite-plugin-relay';
+import { viteStaticCopy } from 'vite-plugin-static-copy';
+import viteMonacoEditor from 'vite-plugin-monaco-editor';
+
+// Handle ESM/CJS interop for vite-plugin-monaco-editor
+const monacoEditorPlugin = (viteMonacoEditor as unknown as {default: typeof viteMonacoEditor}).default;
```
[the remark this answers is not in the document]
xfournet: rejected: this will be detected when the library will be updated to ESM. Handling the proposed way, won't let us the chance to remove the interop code when it becomes unnecessary

opencti-platform/opencti-front/script/sort-translation-files.js
```diff
@@ -1,19 +1,19 @@
-const fs = require('fs');
-const path = require('path');
+import { readFile, readdir } from 'node:fs/promises';
+import path from 'node:path';
```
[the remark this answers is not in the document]
xfournet: updated in a different way to use only async promise based methods

opencti-platform/opencti-graphql/script/script-stack-analysis.js
```diff
@@ -1,19 +1,18 @@
-/* eslint-disable no-console */
 import { readFileSync } from 'fs';
```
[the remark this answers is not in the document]
xfournet: changing to async

opencti-platform/opencti-front/vite.config.ts
```diff
@@ -0,0 +1,98 @@
...
+  changeOrigin: true,
+  ws,
+});
+
+// https://vitejs.dev/config/
+export default defineConfig({
+  build: {
+    target: ['chrome58'],
+    sourcemap: true,
+  },
```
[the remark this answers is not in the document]
xfournet: this is handled by build scripts

opencti-platform/opencti-front/vite.config.ts
```diff
@@ -0,0 +1,100 @@
...
+const backProxy = (ws = false) => ({
+  target: process.env.BACK_END_URL ?? 'http://localhost:4000',
+  changeOrigin: true,
+  ws,
+});
+
+// https://vitejs.dev/config/
+export default defineConfig({
+  build: {
+    target: ['chrome58'],
```
ludovic, line 20: Unless there is a specific reason, I think it would be better to remove the `target`; which is particularly outdated here; and let Vite choose(default 'baseline-widely-available').

xfournet: yes i have hesitated to remove it right now, but the PR is already quite large, i think to do it in a separate PR.

opencti-platform/opencti-front/package.json
```diff
@@ -9,11 +10,10 @@
     "relay": "relay-compiler relay.config.json",
     "dev": "yarn relay && vite",
     "dev:vite:debug": "yarn relay && vite --debug plugin-transform",
-    "start": "yarn relay && node builder/dev/dev.js",
+    "start": "yarn dev",
     "esbuild": "node builder/prod/prod.js",
```
fellowseb: Can't we remove this script and the `esbuild` devdependency ? `builder/prod/prod.js` doesn't exist anymore

opencti-platform/opencti-front/src/utils/flags.ts
```diff
@@ -0,0 +1,17 @@
...
+const flagsByCode: Record<string, string> = {};
+for (const [path, url] of Object.entries(flagModules)) {
+  const code = path.split('/').pop()?.replace('.svg', '');
+  if (code) {
+    flagsByCode[code] = url;
+  }
+}
+
+export const getFlagUrl = (countryCode: string): string => {
+  return flagsByCode[countryCode.toLowerCase()] ?? '';
```
fellowseb: nit: if `countryCode` doesn't exist in the assets it will result in `img`s with empty `src` thus resolving to current document. Consider returning `undefined` and treat the case at call sites.

opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectContent.jsx
```diff
@@ -30,7 +30,7 @@ import HtmlDisplay from '../../../../components/HtmlDisplay';
 import useAttributes from '../../../../utils/hooks/useAttributes';
 import useHelper from '../../../../utils/hooks/useHelper';
 
-pdfjs.GlobalWorkerOptions.workerSrc = `${APP_BASE_PATH}/static/ext/pdf.worker.mjs`;
+pdfjs.GlobalWorkerOptions.workerSrc = new URL('pdfjs-dist/build/pdf.worker.min.mjs', import.meta.url).toString();
```
fellowseb: nit: I see the same assignment to this global in `PdfViewer.tsx`. Consider mutualizing the assignment in `src/utils/pdfWorker.ts` or similar to avoid one changing in isolation.

fellowseb (approved):
Nice. Tested following the plan 👍

opencti-platform/opencti-front/package.json
fellowseb: (Plus `dependenciesMeta`)

xfournet: unfortunately esbuild is still required by vite-plugin-monaco-editor (and it doesn't provide it, we must keep it in our own package.json)
i was not able to remove vite-plugin-monaco-editor too

fellowseb: I see :/.
But we can still remove the `esbuild` npm script though if I'm not mistaken

opencti-platform/opencti-front/src/utils/flags.ts
xfournet: yep it's was not good, indeed each call sites have already fallback case if no country code has been found. i changed to leverage that.

opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectContent.jsx
xfournet: ok, done

fellowseb (approved):

fellowseb (approved):

opencti-platform/opencti-front/vite.config.ts
xfournet: https://github.com/OpenCTI-Platform/opencti/issues/17050

PR #15612, opened 2026-04-21
[backend/frontend] Add support for selecting unique values in number, line, and area widgets (#11229)

### LINKED ISSUE #11229: Ability to perform unique count in Dashboard for count widget
## Use case

I want to perform a unique count of my active users & have this number within an OpenCTI dashboard to know if users of my platforms are actually using it.

## Proposed Solution

In the count Widget, in activity mode, in the last step "Parameter", called "Counting Method":
- It should contain 2 fields, both mandatory
   - Add a new option called  "Group result by": this parameter's goal is to run a query similar to the Distribution list. This way, the user could count, for instance, "All users having done a mutation, group result ". The default value is "No grouping"
   - Add a new option called "Counting method", with value "Count all/Unique count". The default value is "Count All"

As a result, it needs to be a specific query when performing unique count.
   
scarletmerlin123:
### Proposed changes

* Add ability to select unique vales based on a specified field to support getting distinct counts

### Related issues
fixes #11229 

### Checklist

### Further comments

This function uses the elastic cardinality aggregation to provide a count of unique values. Due to how the count is computed it becomes estimated after a certain threshold. A threshold of 40000 was chosen to handle most cases..


Archidoit:
fixes #11229

opencti-platform/opencti-front/lang/front/fr.json
```diff
@@ -1036,6 +1036,7 @@
   "Data sources": "Sources de données",
   "Data Sources | Techniques": "Sources de données | Techniques",
   "Data type": "Type de données",
+  "Datasets with more than 30000 unique values have estimated counts and may differ slightly from the amount of actual distinct values.": "Les ensembles de données avec plus de 30000 valeurs uniques ont des comptes estimés et peuvent différer légèrement de la quantité de valeurs distinctes réelles.",
```
```suggestion
  "Datasets with more than 30000 unique values have estimated counts and may differ slightly from the amount of actual distinct values.": "Les ensembles de données avec plus de 30000 valeurs uniques ont des comptes estimés qui peuvent différer légèrement du compte des valeurs distinctes réelles.",
```

Archidoit:
If perspective=activty&history, the attributes available should concern activity/history and be more relevant.

image


Indeed, when you select uniq count by marking definition, you will get 0 since it's the marking definition of the event that is taken into account.... See example below

## Suggestion
Remove attributes that concerns stix core objects (label, type, marking, author...) and keep attributes concerning events (related entity, user...)



## Example

In the below screenshot: the count widget is the number of event of scope 'creation' of the last 24 hours and the list widget contains the entities created in the last 24 hours.

- If I uniq count by attribute, it seems correct: 5 entities created today <=> 5 events of scope creation for 5 different entities

image

- If I uniq count by marking definition, the count is 0 whereas some created entities have a marking definition... It's because 0 events have a marking definition.

image


image

Archidoit:
Conclusion:

Working well for attribute = user / related entity / etc
Not working for attribute = markings / labels / etc (you should remove these possibilities, that concerns entities and not actvity&history)

scarletmerlin123:
@Archidoit Thanks for reviewing this. Are you suggesting to remove the options that aren't directly related to activity&history (author, labels, marking definitions, creator) from the options from all activity & history widgets or only those with the unique option?

Would you also want me to remove Type because it can only be History in for this perspective

Archidoit:
@scarletmerlin123 I suggest to remove them for all the attributes inputs of the history&activity widgets, I don't see any use case where it's relevant. Do you see some?

scarletmerlin123:
@Archidoit I agree, I don't see any real use cases for those 5 properties .

scarletmerlin123:
it also looks like Event_type is missing in the translation so it shows incorrectly in the select list. I'll add that in with this PR.

Archidoit:
Can you also make the following UI changes : 

- when selecting 'uniq values', the attribute input should have 'related entity' as default value instead of nothing
image


- The Toogle 'uniq values' should not be above the 'title' input but at the same location as the 'attribute' input

- Put the 'attribute' input on the same line as 'uniq values' 

- Add a 'InformationOutline icon just next 'uniq values' with a tooltip to explain



Below are some schemas to illustrate the different cases

## Schemas

### For line and area widget types

```
┌─────────────────────────────────────────────────────────────────────┐
│  ① Visualization ──── ② Perspective ──── ③ Filters ──── ④ Parameters│
├─────────────────────────────────────────────────────────────────────┤
│                                                                     │
│  Title                                                              │
│  ___________________________________________________________________│
│                                                                     │
│  Interval                                                           │
│  [ Day                                                          ▼ ] │
│                                                                     │
│  [●] Unique values  ⓘ    Attribute  [ Related entity           ▼ ] │
│                                                                     │
│  [○] Stacked    [○] Display legend                                  │
│                                                                     │
│  ⚠ Datasets with more than 30000 unique values have estimated      │
│    counts and may differ slightly from actual distinct values.      │
│                                                                     │
│                                              [ Cancel ]  [ Update ] │
└─────────────────────────────────────────────────────────────────────┘
```

```
┌─────────────────────────────────────────────────────────────────────┐
│  ① Visualization ──── ② Perspective ──── ③ Filters ──── ④ Parameters│
├─────────────────────────────────────────────────────────────────────┤
│                                                                     │
│  Title                                                              │
│  ___________________________________________________________________│
│                                                                     │
│  Interval                                                           │
│  [ Day                                                          ▼ ] │
│                                                                     │
│  [○] Unique values  ⓘ                                              │ 
│                                                                     │
│  [○] Stacked    [○] Display legend                                  │
│                                                                     │
│                                                                     │
│                                              [ Cancel ]  [ Update ] │
└─────────────────────────────────────────────────────────────────────┘
```


### For number widget type


```
┌─────────────────────────────────────────────────────────────────────┐
│  ① Visualization ──── ② Perspective ──── ③ Filters ──── ④ Parameters│
├─────────────────────────────────────────────────────────────────────┤
│                                                                     │
│  Title                                                              │
│  ___________________________________________________________________│
│                                                                     │
│  [●] Unique values  ⓘ      Attribute  [ Related entity    ▼ ]      │
│                                                                     │
│                                                                     │
│  ⚠ Datasets with more than 30000 unique values have estimated      │
│    counts and may differ slightly from actual distinct values.      │
│                                                                     │
│                                              [ Cancel ]  [ Update ] │
└─────────────────────────────────────────────────────────────────────┘
```


### Details of the 'uniq values' composant in one line

```
┌──────────────────────────────────────────────────────────────────┐
│                          │                                        │
│  [●] Unique values  ⓘ   │   Attribute  [ Related entity     ▼ ] │
│                          │                                        │
└──────────────────────────┴────────────────────────────────────────┘
         LEFT SIDE                      RIGHT SIDE
    (always visible)           (visible if toggle = ON)
                                 (hide if toggle = OFF)
```

Archidoit:
I don't understand what the 'search keyword' attribute does... 
image

Archidoit:
Can you also update the documentation ? (Widget configuration > parameters)

https://docs.opencti.io/latest/usage/widgets/?h=wid#widget-configuration

opencti-platform/opencti-front/src/private/components/widgets/WidgetCreationParameters.tsx
```diff
@@ -711,12 +729,7 @@ const WidgetCreationParameters = () => {
                           )
                           }
                         >
-                          {['entity_type',
-                            'context_data.id',
-                            'context_data.created_by_ref_id',
-                            'context_data.labels_ids',
-                            'context_data.object_marking_refs_ids',
-                            'context_data.creator_ids',
+                          {['context_data.id',
```
Archidoit: oh it's attributes that were already present... i'll check if we can really remove them, or how we can make them work

scarletmerlin123:
@Archidoit Thanks, your suggestion for moving the toggle next to the attribute selection is a good idea. For multiline/area it would allow the user to compare unique vs regular counts. I'll get that in, along with updating the documentation.

scarletmerlin123:
@Archidoit I did some additional testing for marking definitions and found that it seems to be working as expected. I tested this change on an environment that had older data and see that the unique count matches what is shown when making a distribution list widget.

image

I did a quick look at the data that had marking definitions in the history index and it was very limited. The only occurrences I found were from a 3 month period between July - September 2024, and mostly read event types. It seems like reporting this information in the history is no longer a feature in the system.

Archidoit:
I also have a request for change for the warning message.

- In the warning message we want to display the exact threshold. In your message it's always 30000, even if elasticsearch:cardinality_threshold has been set.
- Instead of displaying the message in the parameters section, we want it only to be displayed under the tooltip of a warning icon. This icon will be displayed near the widget title only if the widget amount is greather than the threshold.

image

opencti-platform/opencti-front/src/private/components/widgets/WidgetCreationParameters.tsx
Archidoit: you can add the attributes back, it's working well :) there is just a fix to do to make markings work but it's not linked to your issue so you can ignore it (i opened a bug for this : https://github.com/OpenCTI-Platform/opencti/issues/15770 )

Archidoit:
```suggestion
                          {['entity_type',
                            'context_data.id',
                            'context_data.created_by_ref_id',
                            'context_data.labels_ids',
                            'context_data.marking_definitions',
                            'context_data.creator_ids',
                            'context_data.search',
```

Archidoit (changes requested):

scarletmerlin123:
@Archidoit I made some updates to both the UI and functionality. 

- I made the threshold a fixed value rather than keeping it configurable. This prevented needing to add excessive complexity to display the warning icon/tooltip by the title of the widget. In testing there was no noticeable change in memory usage so its unlikely that the config would have made any difference with the queries that get run.
- Updated terminology from 'Unique' to 'Distinct' to improve clarity
- Updated switches to checkboxes
- I updated the UI layout for the widget parameters modal so that it doesn't need to render the same text for every checkbox if there are multiple dataselections

image

docs/docs/usage/widgets.md
```diff
@@ -78,7 +78,14 @@ This level of control ensures that your dashboard aligns precisely with your ana
 
 ![Widget parameters](assets/widget-parameters.png)
 
+For the "Activity & History" perspective there is the option to use the distinct values within the data when using a  "Number", "Line", or "Area" visualization. This enables the user to present additional; insight into the usage of the application.
```
Archidoit:
```suggestion
For the "Activity & History" perspective there is the option to use the distinct values within the data when using a  "Number", "Line", or "Area" visualization. This enables the user to present additional insight into the usage of the application.
```

docs/docs/usage/widgets.md
```diff
@@ -78,7 +78,14 @@ This level of control ensures that your dashboard aligns precisely with your ana
...
 
+- Checked: The widget will report the count of distinct values for the specified field
+
+!!! warning "Distinct values limitation"
+
+    Selecting to use distinct values for data sets with many distinct values may return results that are slightly different than actuals. To preserve performance, the query begins to estimate the count if there are greater than 40000 distinct values in the dataset.
```
Archidoit:
```suggestion
    Selecting to use distinct values for data sets with more than 40,000 distinct values may return results that are slightly different from actuals. Indeed, in this case, to preserve performance, the query begins to estimate the count.
```

opencti-platform/opencti-front/lang/front/fr.json
```diff
@@ -727,6 +727,7 @@
   "Correlated reports": "Rapports connexes",
   "Correlation view": "Vue des corrélations",
   "Count": "Nombre",
+  "Count the number of distinct values in a specified field": "Compter le nombre de valeurs distinctes dans un champ spécifié",
```
```suggestion
  "Count the number of distinct values in a specified field": "Compter le nombre de valeurs distinctes pour un champ donné",
```

opencti-platform/opencti-front/src/components/dashboard/WidgetContainer.tsx
```diff
@@ -4,7 +4,13 @@ import Card, { CardProps } from '../common/card/Card';
 import Label from '../common/label/Label';
 import ChartExportPopover from '../../private/components/common/charts/ChartExportPopover';
 import { ErrorBoundary } from '@components/Error';
+<<<<<<< issue/11229
```
Archidoit: to be removed, maybe due to the merge of master you did. It's better to rebase to avoid those kind of conflicts.

opencti-platform/opencti-front/src/components/dashboard/WidgetContainer.tsx
```diff
@@ -24,13 +31,38 @@ const WidgetContainer: FunctionComponent<WidgetContainerProps> = ({
   padding,
   chart,
   action,
+  warning,
 }) => {
+  const theme = useTheme();
+  const titleWithWarning = () => {
```
Archidoit: useless to use a function here, you can simply create a constant for titleWithWarning

opencti-platform/opencti-front/src/private/components/common/audits/AuditsMultiAreaChart.jsx
```diff
@@ -89,6 +93,12 @@ const AuditsMultiAreaChart = ({
     interval: parameters.interval ?? 'day',
     timeSeriesParameters,
   }), [startDate, endDate, fallbackDates, parameters.interval, timeSeriesParameters]);
+  const warning = () => {
```
Archidoit: same here, create a constant instead of a function

opencti-platform/opencti-front/src/private/components/common/audits/AuditsMultiLineChart.jsx
```diff
@@ -109,12 +118,23 @@ const AuditsMultiLineChart = ({
         </div>
       );
     }
+    const computeShowWarning = (data) => {
```
Archidoit: this function is written twice in different files. You can write a general version of it in widgetUtils.tsx.

opencti-platform/opencti-front/src/private/components/widgets/WidgetCreationParameters.tsx
```diff
@@ -178,8 +182,71 @@ const WidgetCreationParameters = () => {
...
 
+  const uniqueParameterEnabled = (perspective: WidgetPerspective | null | undefined, visualizationType: WidgetVisualizationTypes | ''): boolean => {
+    return perspective === 'audits' && (['number', 'line', 'area'].includes(visualizationType)) && !getCurrentAvailableParameters(type).includes('attribute');
+  };
   const maxResultCount = getMaxResultCount(type);
 
+  const auditAttributeSelectionSection = (uniqueParameterEnabled: boolean, dataSelectionIndex: number) => {
+    return (
+      <Grid container spacing={4} sx={{ width: '100%' }}>
+        {uniqueParameterEnabled ? (
```
Archidoit: rather than : uniqueParameterEnabled ? <...> : undefined

you can simply write : uniqueParameterEnabled && <...>

opencti-platform/opencti-front/src/private/components/widgets/WidgetCreationParameters.tsx
```diff
@@ -255,7 +322,30 @@ const WidgetCreationParameters = () => {
           </Select>
         </FormControl>
       )}
-
+      {uniqueParameterEnabled(dataSelection[0].perspective, type) ? (
```
Archidoit: same

opencti-platform/opencti-front/src/utils/widget/widget.d.ts
```diff
@@ -30,6 +30,7 @@ export interface WidgetDataSelection {
   sort_mode?: string | null;
   field?: string;
   relationship_type?: string;
+  unique: boolean;
```
Archidoit: make this attribute mandatory. Indeed, all the widget have not necessary defined 'unique'
```suggestion
  unique?: boolean;
```

opencti-platform/opencti-front/src/utils/widget/widgetUtils.tsx
```diff
@@ -322,3 +322,6 @@ export const isWidgetUsingRelationsAggregation = (
 
   return widgetTypesWithRelationsAggregation.includes(widgetType);
 };
+
+export const count_estimation_threshold = 40000;
```
Archidoit: constants should be in capital letters

opencti-platform/opencti-front/src/utils/widget/widgetUtils.tsx
```diff
@@ -322,3 +322,6 @@ export const isWidgetUsingRelationsAggregation = (
...
+export const count_estimation_warning = 'Datasets with more than 40000 distinct values have estimated counts and may differ slightly from the amount of actual distinct values';
```
Archidoit: same

opencti-platform/opencti-front/src/utils/widget/widgetUtils.tsx
Archidoit:
```suggestion
export const UNIQ_COUNT_ESTIMATION_THRESHOLD = 40000;
```

opencti-platform/opencti-front/src/private/components/common/audits/AuditsNumber.jsx
```diff
@@ -89,6 +96,7 @@ const AuditsNumber = ({
     selection.filters,
     { removeTypeAll: true, startDate, endDate, dateAttribute },
   );
+  const warning = showWarning && t_i18n(UNIQUE_COUNT_ESTIMATION_WARNING);
```
Archidoit:
```suggestion
  const warning = showWarning ? t_i18n(UNIQUE_COUNT_ESTIMATION_WARNING) : undefined;
```

opencti-platform/opencti-front/src/private/components/common/audits/AuditsNumber.jsx
```diff
@@ -63,6 +69,7 @@ const AuditsNumber = ({
   height,
   host,
 }) => {
+  const [showWarning, setShowWarning] = useState(0);
```
Archidoit: why is showWarning initialized to a number here? instead of 'false' ?

opencti-platform/opencti-front/src/private/components/common/audits/AuditsNumber.jsx
```diff
@@ -98,17 +106,19 @@ const AuditsNumber = ({
...
           ? <WidgetNoHostEntity host={host} />
           : (
               <QueryRenderer
                 query={auditsNumberNumberQuery}
-                variables={{ types, filters, startDate, endDate: dayAgo() }}
+                variables={{ types, filters, startDate, endDate: dayAgo(), field: selection.attribute, unique: selection.unique }}
                 render={({ props }) => {
                   if (props && props.auditsNumber) {
                     const { total, count } = props.auditsNumber;
+                    setShowWarning(dataSelection[0].unique && count > UNIQUE_COUNT_ESTIMATION_THRESHOLD);
```
Archidoit:
```suggestion
                    setShowWarning(selection.unique && count > UNIQUE_COUNT_ESTIMATION_THRESHOLD);
```

Archidoit:
For widgets with no 'uniq' field but with an 'attribute' field, the attribute field is too short. 
Example: Create a widget of type 'list (distribution)' and of perspective Activity & History. In the parameter tab : 

image

opencti-platform/opencti-front/src/private/components/common/audits/AuditsNumber.jsx
Archidoit:
```suggestion
                    setShowWarning(dataSelection[0].unique && total> UNIQUE_COUNT_ESTIMATION_THRESHOLD);
```
count is the number for the last 24h hours. Here it's 'total' that should be compared to the threshold

Archidoit (changes requested):

Archidoit:
For widget of type=number, perspective=audits&history, if 'uniq' is not checked, the 'attribute' field should not appear : 
image

Archidoit:
In case the number is higher than the threshold, a warning icon is displayed near the widget name : 
image

This icon should have the same design than the other warning icons of the platform, like for instance the one displayed in the filters tab of a widget of type 'list (distribution)' and of perspective knowledge : 
image

Archidoit:
The mock-ups I sent are slightly different from what you implemented for uniq and attribute fields. 

Here is an insight of what we except : 
image


Here is the actual UI in your branch : 

image


For multiple attributes, here is what we expect : 

image



(design validated with @alice-debra )

Archidoit:
I pushed a commit with little fixes and frontend tests.

opencti-platform/opencti-front/src/private/components/common/audits/AuditsNumber.jsx
Archidoit: fixed in my last commit

Archidoit: I fixed it in my last commit

opencti-platform/opencti-front/src/private/components/common/audits/AuditsNumber.jsx
Archidoit: I fixed it in my last commit

opencti-platform/opencti-front/src/private/components/common/audits/AuditsNumber.jsx
Archidoit: I fixed it in my last commit

scarletmerlin123:
> The mock-ups I sent are slightly different from what you implemented for uniq and attribute fields.
> 
> Here is an insight of what we except : image
> 
> Here is the actual UI in your branch :
> 
> image
> For multiple attributes, here is what we expect :
> 
> image
> (design validated with @alice-debra )

Thanks for reviewing! @Archidoit The reason behind laying it out the way I did was to reduce the amount of duplication on the screen. The label next to the checkbox and the tooltip are the same for every data selection, so in a situation where there are multiple data selections the content gets very repetitive.  I can revert it to follow the mockup if you prefer.

image

Archidoit:
> > The mock-ups I sent are slightly different from what you implemented for uniq and attribute fields.
> > Here is an insight of what we except : image
> > Here is the actual UI in your branch :
> > image
> > For multiple attributes, here is what we expect :
> > image
> > (design validated with @alice-debra )
> 
> Thanks for reviewing! @Archidoit The reason behind laying it out the way I did was to reduce the amount of duplication on the screen. The label next to the checkbox and the tooltip are the same for every data selection, so in a situation where there are multiple data selections the content gets very repetitive. I can revert it to follow the mockup if you prefer.
> 
> image

--> Indeed I understand. After checking with our design team, we can keep your UI. But this fix is required : 

For widget of type=number, perspective=audits&history, if 'uniq' is not checked, the 'attribute' field should not appear : 
image

scarletmerlin123:
@Archidoit I pushed updates for the remaining issues. I noticed the UI didn't look right when simply removing the attribute selection from auditsnumber when distinct is not selected. I updated the UI so when its a single dataSelection it displays inline, otherwise it will do the column based view. 
image

image
image

Archidoit:
I pushed little code improvements and UI modifications. 

Here are some screenshots of the new UI for audits widgets.

With a single data selection : 
image

image

With multiple data selections :
image

Archidoit (approved):
Everything is good for me! This will be a useful feature for audit widgets. Thanks for your contribution !
I will merge your PR during the day.

esrevi (approved):
LGTM

PR #15615, opened 2026-04-21
[backend] try catch during migration to handle corrupted notes (#15493)

### LINKED ISSUE #15493: Notes with identical content and created date collide due to missing attribute_abstract in standard_id
## Description

Notes in OpenCTI are deduplicated using only `content` + `created` fields to compute the `standard_id`. When different Notes (attached to different objects) share the same body text and creation date, the platform computes the same `standard_id` for both, causing a `FUNCTIONAL_ERROR` with message *"This update will produce a duplicate"* during ingestion.

This is particularly visible with the S3 connector (Accenture ACTI feed) where multiple vulnerabilities can produce Notes (Title, Analysis, History) with identical content and `created` timestamps. The connector generates unique STIX IDs (based on `x_acti_uuid`), but the platform recomputes its own `standard_id` from the actual note fields and ignores the connector's intended identity.

The root cause is that `attribute_abstract` is not part of the Note `standard_id` contribution in `identifier.js`. The abstract semantically represents the subject of a Note, and two Notes about different subjects should be treated as different entities even if they share body text.

## Environment

1. OS: Any
2. OpenCTI version: Current master
3. OpenCTI client: Connectors (S3 / any connector creating Notes)
4. Other environment details: N/A

## Reproducible Steps

1. Ingest a STIX bundle containing two Note objects with different `attribute_abstract` values but identical `content` and `created` fields
2. The first Note is created successfully
3. The second Note triggers `FUNCTIONAL_ERROR`: *"This update will produce a duplicate"*

## Expected Output

Both Notes should be created as separate entities since they have different abstracts (i.e., they describe different things).

## Actual Output

Only the first Note is created. Subsequent Notes with the same `content` + `created` fail with a duplicate error.

## Additional information

The fix involves adding `attribute_abstract` to the Note contribution in `identifier.js`:

```js
[D.ENTITY_TYPE_CONTAINER_NOTE]: [
  { src: 'content' },
  { src: 'created', dependencies: ['content'] },
  { src: 'attribute_abstract', dependencies: ['content'] }
],
```

Notes without an abstract gracefully fall back to the current `(content, created)` dedup. A data migration is needed to recompute `standard_id` for existing Notes that have an `attribute_abstract`.

The corresponding pycti `Note.generate_id()` and connectors (S3, SDK) also need alignment.

JeremyCloarec:
### Proposed changes

* try catch during migration to handle corrupted notes
*

### Related issues

* fix #15493
*

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/migrations/1776164753048-note_standard_id_abstract.js
+++ b/opencti-platform/opencti-graphql/src/migrations/1776164753048-note_standard_id_abstract.js
@@ -18,16 +18,21 @@ export const up = async (next) => {
   const callback = (notes) => {
     const op = notes
       .map((note) => {
-        const newId = generateStandardId(note.entity_type, note);
-        if (newId === note.standard_id) {
+        try {
+          const newId = generateStandardId(note.entity_type, note);
+          if (newId === note.standard_id) {
+            return [];
+          }
+          const previousStixIds = note.x_opencti_stix_ids ?? [];
+          const updatedStixIds = R.uniq([...previousStixIds, note.standard_id]);
+          return [
+            { update: { _index: note._index, _id: note._id } },
+            { doc: { standard_id: newId, x_opencti_stix_ids: updatedStixIds } },
+          ];
+        } catch (err) {
+          logApp.warn('[OPENCTI] Error during note rewrite, skipping', { err, note });
           return [];
         }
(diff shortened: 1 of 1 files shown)


aHenryJard (approved):
Looks good to me, tested locally it's running fine.

PR #15616, opened 2026-04-21
[backend] Add telemetry for active status of history/activity retention rules (#3920)

### LINKED ISSUE #3920: Retention policy on History knowledge
## Use case

Being able to create a retention policy on History knowledge.

### LINKED ISSUE #11399: Retention policy on Activity
## Use case

Ensure that we are able to have retention policies for Activity logs, in order to ensure that disk space can be removed.

CelineSebe:
### Proposed changes

* add telemetry on telemetry manager

### Related issues

* Resolves #11399 
* Resolves #3920 

### Checklist


opencti-platform/opencti-graphql/src/manager/retentionManager.ts
ludovic: Wrong date field for History entries `convertFiltersToQueryOptions` defaults to field: 'updated_at', but `History` entities use timestamp as their primary temporal field (declared isFilterable: true in HistoryDefinition), while updated_at is declared with isFilterable: false.

This means the `before` filter is applied on `updated_at` instead of `timestamp`, so the retention window will not correctly target entries by their actual age.

Fix (apply in both `getElementsToDelete` in `retentionManager.ts` and `checkRetentionRule` in `retentionRules-domain.ts`):


```suggestion
    const queryOptions = await convertFiltersToQueryOptions(jsonFilters, { before, field: 'timestamp' });
```

opencti-platform/opencti-graphql/src/modules/retentionRules/retentionRules-domain.ts
ludovic: same here

opencti-platform/opencti-front/src/private/components/settings/retention/RetentionEdition.jsx
ludovic, line 264: the style is different for this alert, is it wanted ? ({margin: '15px 15px 0 15px' } for other alerts

opencti-platform/opencti-graphql/src/manager/retentionManager.ts
CelineSebe: thanks :)

opencti-platform/opencti-graphql/src/manager/retentionManager.ts
```diff
@@ -69,7 +69,7 @@ export const getElementsToDelete = async (context: AuthContext, scope: string, b
     result = await paginatedForPathWithEnrichment(context, RETENTION_MANAGER_USER, 'import/pending', undefined, { first: RETENTION_BATCH_SIZE, notModifiedSince: before.toISOString(), exact_path: false });
   } else if (scope === 'history') {
     const jsonFilters = filters ? JSON.parse(filters) : null;
-    const queryOptions = await convertFiltersToQueryOptions(jsonFilters, { before });
+    const queryOptions = await convertFiltersToQueryOptions(jsonFilters, { before, field: 'updated_at' });
```
ludovic: You should use `field: 'timestamp'` instead of `field: 'updated_at'`.

In `historyManager.ts`, History entries store two distinct date fields:
- `updated_at` → the **indexing date** (when the document was written to ES)
- `timestamp` → the **actual event date** (when the action occurred in OpenCTI)

These can diverge in case of stream lag or reprocessing. Using `updated_at` as the retention cutoff means filtering on the indexing time, not the real age of the event.

`timestamp` is already used as the canonical sort field for History (`orderBy: ['timestamp']` in `historyManager.ts`), and is the only date field declared `isFilterable: true` in `HistoryDefinition`.

```ts
const queryOptions = await convertFiltersToQueryOptions(jsonFilters, { before, field: 'timestamp' });
```

(applies to both `getElementsToDelete` in `retentionManager.ts` and `checkRetentionRule` in `retentionRules-domain.ts`)

marieflorescontact (commented):
Functionnaly tested:
`opencti_retention_history_creation_count` is not incremented when history retention rule is created.

ludovic (changes requested):
The telemetry part seems to have become obsolete due to the changes in PR #15967:
* **The issue**: PR #15967 prevents users from creating custom rules for the `history` scope and automatically creates a single default rule instead. Therefore, the `retention_history_creation_count` telemetry gauge will always equal `1` on every platform (tracking the system setup/migrations), which doesn't reflect actual user behavior.
* **Suggestion**: I suggest removing the telemetry code from this PR, or refactoring it to track whether the default history rule is **active** (tracking the `active` status value rather than the rule's creation).

What do you think ?

CelineSebe:
> The telemetry part seems to have become obsolete due to the changes in PR #15967:
> 
> * **The issue**: PR [[backend/frontend] add active/inactive on retention policies (#15669) #15967](https://github.com/OpenCTI-Platform/opencti/pull/15967) prevents users from creating custom rules for the `history` scope and automatically creates a single default rule instead. Therefore, the `retention_history_creation_count` telemetry gauge will always equal `1` on every platform (tracking the system setup/migrations), which doesn't reflect actual user behavior.
> * **Suggestion**: I suggest removing the telemetry code from this PR, or refactoring it to track whether the default history rule is **active** (tracking the `active` status value rather than the rule's creation).
> 
> What do you think ?

Thanks! Indeed, it's no longer necessary to track the rule's creation. I'm modifying the telemetry code

marieflorescontact (commented):
I dont understand how Telemetry is supposed to work for `History` and` Activity retention manager`. I can't see any difference in the Telemetry file.  Could you exlpain how I can test and what is expected output ?

For `ACTIVE` History rretention manager:
````
          { // ACTIVE
              "name": "opencti_is_history_retention_rule_active",
              "description": "Whether a history retention rule is active on the platform",
              "unit": "boolean",
              "gauge": {
                "dataPoints": [
                  {
                    "attributes": [],
                    "startTimeUnixNano": "1780293770892000000",
                    "timeUnixNano": "1780293830894000000",
                    "asInt": 1
                  },
                  {
                    "attributes": [],
                    "startTimeUnixNano": "1780293830894000000",
                    "timeUnixNano": "1780293890909000000",
                    "asInt": 1
                  }
                ]
              }
            },
````

For `INACTIVE`
````
{ // INACTIVE
              "name": "opencti_is_history_retention_rule_active",
              "description": "Whether a history retention rule is active on the platform",
              "unit": "boolean",
              "gauge": {
                "dataPoints": [
                  {
                    "attributes": [],
                    "startTimeUnixNano": "1780294130910000000",
                    "timeUnixNano": "1780294190927000000",
                    "asInt": 1
                  },
                  {
                    "attributes": [],
                    "startTimeUnixNano": "1780294190927000000",
                    "timeUnixNano": "1780294250927000000",
                    "asInt": 1
                  }
                ]
              }
            },
````

CelineSebe:
@marieflorescontact Thanks for your report

How to test:

1- Run in DEV_MODE, or add telemetry_manager: { console_debug: true } in development.json to also log to console
2- Create a retention rule with scope history (or activity) and mark it active
3- Wait ~2 min → inspect the telemetry JSON file → expect "asInt": 1
4- Deactivate the rule (or delete it)
5- Wait ~2 min → expect "asInt": 0

marieflorescontact:
> @marieflorescontact Thanks for your report
> 
> How to test:
> 
> 1- Run in DEV_MODE, or add telemetry_manager: { console_debug: true } in development.json to also log to console 2- Create a retention rule with scope history (or activity) and mark it active 3- Wait ~2 min → inspect the telemetry JSON file → expect "asInt": 1 4- Deactivate the rule (or delete it) 5- Wait ~2 min → expect "asInt": 0

I just tested it again and the counter is not incremented it is always: "asInt": 0 for `opencti_retention_history_creation_count`
did you test it on you side ?

image

````
{
              "name": "opencti_retention_history_creation_count",
              "description": "Number of history retention rules created",
              "unit": "count",
              "gauge": {
                "dataPoints": [
                  {
                    "attributes": [],
                    "startTimeUnixNano": "1780327345944000000",
                    "timeUnixNano": "1780327405964000000",
                    "asInt": 0
                  },
                  {
                    "attributes": [],
                    "startTimeUnixNano": "1780327405964000000",
                    "timeUnixNano": "1780327465979000000",
                    "asInt": 0
                  }
                ]
              }
            },
````

CelineSebe:
@marieflorescontact Yes, I think it's works well on my side for `is_history_retention_rule_active` and for `is_activity_retention_rule_active`. O --> false, 1 --> true

image

marieflorescontact:
> @marieflorescontact Yes, I think it's works well on my side for `is_history_retention_rule_active` and for `is_activity_retention_rule_active`. O --> false, 1 --> true
> 
image

image


ok now that the unit has changed to be a boolean  instead of a count it works. It was broken but you fixed it. great.

ludovic (dismissed):

ludovic (approved):

PR #15626, opened 2026-04-22
[backend] Playbook: added filter to Create Indicator From Observable Component (#14283)

SarahBocognano:
### Proposed changes

* add applyWithFilter attribute
* add logic
* add tests
* add doc

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/14285

### Checklist


opencti-platform/opencti-graphql/tests/03-integration/01-database/playbookComponents/create-indicator-component-test.ts
```diff
@@ -268,4 +271,594 @@ describe('Create indicator component', () => {
...
+      expect(basedOnRels[0].target_ref).toBe(OBSERVABLE_ID_2);
+
+      // Main observable did not produce any indicator
+      expect(result.bundle.objects.filter((o) => o.type === 'intrusion-set')).toHaveLength(1);
+    });
+  });
+
+  // Propagation of marking refs, labels, created_by_ref, external_references
+
+  it('should propagate object_marking_refs from observable to indicator and relationships', async () => {
```
MTorbay-Filigran: It would be easier to read and understand what the tests are about if they were wrapped into describes like you did above with the 'when using filters on bundle containing multiple objects' cases

opencti-platform/opencti-graphql/src/modules/playbook/components/create-indicator-component.ts
```diff
@@ -79,164 +85,116 @@ export const PLAYBOOK_CREATE_INDICATOR_COMPONENT: PlaybookComponent<CreateIndica
     return R.mergeDeepRight<JSONSchemaType<CreateIndicatorConfiguration>, any>(PLAYBOOK_CREATE_INDICATOR_COMPONENT_SCHEMA, schemaElement);
   },
   executor: async ({ playbookNode, dataInstanceId, bundle }) => {
-    const { applyToElements, wrap_in_container, types } = playbookNode.configuration;
+    const { applyToElements, applyWithFilters, wrap_in_container, types } = playbookNode.configuration;
     const context = executionContext('playbook_components');
     const baseData = extractBundleBaseElement(dataInstanceId, bundle);
 
-    const observablesToApply = bundle.objects.filter((object) => isBundleElementInScope(object, applyToElements, dataInstanceId));
```
MTorbay-Filigran, line 86: You could have kept this line, and filter directly on the observablesToApply list with filterBundleElements before the loop, instead of using isBundleElementMatchFilters that filters elements one by one. So that the loop is only on relevant elements

MTorbay-Filigran (approved):

Issue #8707, opened 2024-10-18
Draft: ability to validate by a tier & view a diff panel 


nino-filigran:
## Use case

Ability to request a draft to be validated by a tier & ensure that a diff panel between what's in the platform VS what's in draft is present.

PR #15637, opened 2026-04-22
[graphql] Add draft workspace review queries and mutations (#8707)

ludovic:
### Proposed changes

Extend the `draftWorkspace` GraphQL API with queries needed for the review diff panel:

- Add `draftOperation` filter to `draftWorkspaceEntities` to list only created/updated/deleted objects.
- Add `draftWorkspaceContainerObjects(draftId, containerId)` — returns objects added or removed from a container inside a draft.
- Add `draftWorkspaceResolveIds(draftId, ids)` — resolves a list of ids to their representative names within a draft context.
- Add `draftWorkspaceEntityFields(draftId, entityId)` — returns the scalar fields of a draft entity.
- Add `draftWorkspaceEntityRelations(draftId, entityId)` — returns core/sighting relations created or deleted in a draft for a given entity.
- Add `draftWorkspaceEntityContainerRefs(draftId, entityId)` — returns containers that reference the entity via an `object` ref changed in a draft.
- Add `reviewsCount` to `DraftObjectsCount` (entities + observables + containers).
- Add `draft_updates_patch` to `DraftVersion` to expose the JSON diff of an updated entity.
- Add `includeDeletedInDraft` option to `RelationOptions` in `middleware-loader` to allow querying deleted-in-draft relations.
- Wire all domain helpers in `draftWorkspace-resolvers.ts`.

### Related issues

* Fixes https://github.com/OpenCTI-Platform/opencti/issues/8707 (part 2/3, depends on #15636).

### DIFF
--- a/opencti-platform/opencti-graphql/src/modules/draftWorkspace/draftWorkspace-domain.ts
+++ b/opencti-platform/opencti-graphql/src/modules/draftWorkspace/draftWorkspace-domain.ts
@@ -11,6 +11,7 @@ import { buildStixBundle } from '../../database/stix-2-1-converter';
 import { computeSumOfList, isDraftIndex, READ_INDEX_DRAFT_OBJECTS, READ_INDEX_HISTORY, READ_INDEX_INTERNAL_OBJECTS } from '../../database/utils';
 import { createWork, updateExpectationsNumber } from '../../domain/work';
 import {
+  DraftChangeType,
   type DraftWorkspaceAddInput,
   type EditContext,
   type EditInput,
@@ -513,7 +514,7 @@ export const listDraftContainerObjects = async (context: AuthContext, user: Auth
       entity_id: rel.toId,
       entity_type: entity.entity_type,
       representative_main: extractEntityRepresentativeName(entity as any),
-      draft_operation: op === DRAFT_OPERATION_CREATE ? 'add' : 'remove',
+      draft_operation: op === DRAFT_OPERATION_CREATE ? DraftChangeType.Add : DraftChangeType.Remove,
     });
   }
   return result;
@@ -557,7 +558,7 @@ type DraftEntityRelationResult = {
   relation_id: string; relationship_type: string;
   from_id: string; from_type: string; from_name: string;
   to_id: string; to_type: string; to_name: string;
-  draft_operation: string;
+  draft_operation: DraftChangeType;
 };
 
 export const getEntityRelations = async (
@@ -617,7 +618,7 @@ export const getEntityRelations = async (
     to_id: rel.toId,
     to_type: rel.toType,
     to_name: rel.toName,
-    draft_operation: rel.draft_change!.draft_operation === DRAFT_OPERATION_CREATE ? 'create' : 'delete',
+    draft_operation: rel.draft_change!.draft_operation === DRAFT_OPERATION_CREATE ? DraftChangeType.Create : DraftChangeType.Delete,
   }));
   const containerResults = containerRelations.map(({ rel, refOp }) => ({
     relation_id: rel.internal_id,
@@ -629,7 +630,7 @@ export const getEntityRelations = async (
     to_type: rel.toType,
     to_name: rel.toName,
     // The ref changed (added/removed from this container), not the relation itself
-    draft_operation: refOp === DRAFT_OPERATION_CREATE ? 'add' : 'remove',
+    draft_operation: refOp === DRAFT_OPERATION_CREATE ? DraftChangeType.Add : DraftChangeType.Remove,
   }));
   return [...directResults, ...containerResults];
 };
@@ -638,7 +639,7 @@ type DraftEntityContainerRefResult = {
   container_id: string;
   container_type: string;
   container_name: string;
-  draft_operation: string;
+  draft_operation: DraftChangeType;
 };
 
 export const getEntityContainerRefs = async (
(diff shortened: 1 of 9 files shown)


opencti-platform/opencti-graphql/src/modules/draftWorkspace/draftWorkspace-domain.ts
```diff
@@ -191,8 +195,18 @@ export const listDraftObjects = async (context: AuthContext, user: AuthUser, arg
   if (types.length === 0) {
     types.push(ABSTRACT_STIX_CORE_OBJECT);
   }
-  const newArgs: EntityOptions<BasicStoreEntity> = { ...listArgs, types, indices: [READ_INDEX_DRAFT_OBJECTS], includeDeletedInDraft: true };
   const draftContext = { ...context, draft_context: draftId };
+  const draftOperationFilter = draftOperation
+    ? addFilter(listArgs.filters, 'draft_change.draft_operation', draftOperation)
+    : listArgs.filters;
```
CelineSebe, line 202: The `draftOperation` parameter is passed directly as an Elasticsearch filter value without validation. A malicious user could inject an arbitrary value ?

suggestion:

```
const VALID_DRAFT_OPERATIONS = new Set([DRAFT_OPERATION_CREATE, DRAFT_OPERATION_UPDATE, DRAFT_OPERATION_DELETE]);
const isValidDraftOperation = (operation: string) => VALID_DRAFT_OPERATIONS.has(operation);

  if (draftOperation && !isValidDraftOperation(draftOperation)) {
    throw FunctionalError('Invalid draft operation', { draftOperation });
  }
```

opencti-platform/opencti-graphql/src/modules/draftWorkspace/draftWorkspace-domain.ts
```diff
@@ -473,3 +529,189 @@ export const draftWorkspaceEditContext = async (context: AuthContext, user: Auth
     return notify(BUS_TOPICS[ENTITY_TYPE_DRAFT_WORKSPACE].CONTEXT_TOPIC, draft, user);
   });
 };
+
+export const resolveIdRepresentatives = async (
+  context: AuthContext,
+  user: AuthUser,
+  args: { draftId: string; ids: string[] },
+): Promise<{ id: string; representative_main: string | null }[]> => {
+  const { draftId, ids } = args;
```
CelineSebe, line 539: The `draftWorkspaceResolveIds` request accepts an `ids` array of any size. A call with thousands of IDs would cause excessive load

suggestion:

```
const MAX_RESOLVE_IDS = 100;

if (ids.length > MAX_RESOLVE_IDS) {
  throw FunctionalError(`Too many IDs to resolve (max ${MAX_RESOLVE_IDS})`, { count: ids.length });
}
```

opencti-platform/opencti-graphql/src/modules/draftWorkspace/draftWorkspace.graphql
```diff
@@ -85,7 +93,12 @@ type Query {
         orderMode: OrderingMode
         filters: FilterGroup
         search: String
+        draftOperation: String
```
marieflorescontact: should we use existing Enum DraftOperation instead of String? (in opencti.graphql)

marieflorescontact:
```suggestion
        draftOperation: DraftOperation
```

opencti-platform/opencti-graphql/src/modules/draftWorkspace/draftWorkspace-domain.ts
```diff
@@ -191,8 +195,21 @@ export const listDraftObjects = async (context: AuthContext, user: AuthUser, arg
   if (types.length === 0) {
     types.push(ABSTRACT_STIX_CORE_OBJECT);
   }
-  const newArgs: EntityOptions<BasicStoreEntity> = { ...listArgs, types, indices: [READ_INDEX_DRAFT_OBJECTS], includeDeletedInDraft: true };
+  if (draftOperation && !getDraftOperations().includes(draftOperation)) {
+    throw FunctionalError('Invalid draft operation', { draftOperation });
```
marieflorescontact: this check could be done also by using Enum DraftOperation in .graphql instead of type string

marieflorescontact (commented):
Functionnaly tested ok for all API (meaning No error) (except for the `draft_updates_patch `new field that I did't know how to test )
but I am not sure about the expected result:
````
query {
  draftWorkspaceEntityFields(
    draftId: "d5336da8-2c21-45e0-956d-457989c96857"
    entityId: "edd6db69-1dbe-41af-983b-af8601942b37"
  ) {
    field
    values
  }
}
````
is it supposed to return fields that has not been changed ? (as for an empty description for instance ?)
````
 {
        "field": "description",
        "values": [
          ""
        ]
      },
````

ludovic:
Yes, this is intentional: `draftWorkspaceEntityFields` returns all current scalar fields of the entity as stored in the draft index (not only the changed ones). An empty `description: ""` is a valid value, it means the field is empty in the draft, which is relevant for an update scenario where it may have been explicitly cleared. The `draft_updates_patch` field (on `DraftVersion`) is the one that carries the actual diff of what changed.

opencti-platform/opencti-graphql/src/modules/draftWorkspace/draftWorkspace.graphql
ludovic: I've refactored the different types used; they were causing confusion

CelineSebe (approved):

PR #15662, opened 2026-04-23
[frontend/backend] graphql updated to avoid empty content for note (#15618)

### LINKED ISSUE #15618: A note can be created with an empty content even when it shouldn't be possible
## Description

We have a string constraint on content in GraphQL for note creation: the content must contain more than 2 characters.
However, in some cases, notes with an empty content can be created

## Environment

4. Other environment details:

## Reproducible Steps

1. Run this query:
```
mutation NoteCreationMutation(
  $input: NoteAddInput!
) {
  noteAdd(input: $input) {
    id
    standard_id
    entity_type
    parent_types
    attribute_abstract
    content
  }
}
```
With this input
```
{
	"input": {
		"confidence": 100,
		"content": "\n\n",
		"externalReferences": [],
		"likelihood": 50,
		"note_types": [],
		"objectLabel": [],
		"objectMarking": []
	}
}
```

2. { e.g. Click ... }
3. { e.g. Error ... }

## Expected Output

Content keeps its value of '\n\n'

## Actual Output

Content is empty when you query the note again

## Additional information

This is caused by the value.trim() here in prepareElementForIndexing

Image

The bug can occur on other string attributes also.

ValentinBouzinFiligran:
### Proposed changes

* format on graphql

### Related issues

* Fixes #15618

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/config/schema/opencti.graphql
+++ b/opencti-platform/opencti-graphql/config/schema/opencti.graphql
@@ -3864,7 +3864,8 @@ input NoteAddInput {
   stix_id: StixId
   x_opencti_stix_ids: [StixId]
   attribute_abstract: String
-  content: String! @constraint(minLength: 2)
+#  content: String! @constraint(minLength: 2)
+  content: String! @constraint(minLength: 2, pattern: "\\S[\\s\\S]*\\S")
   authors: [String]
   note_types: [String]
   likelihood: Int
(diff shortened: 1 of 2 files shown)


opencti-platform/opencti-graphql/config/schema/opencti.graphql
```diff
@@ -3864,7 +3864,7 @@ input NoteAddInput {
   stix_id: StixId
   x_opencti_stix_ids: [StixId]
   attribute_abstract: String
-  content: String! @constraint(minLength: 2)
+  content: String! @constraint(minLength: 2, pattern: "\\S[\\s\\S]*\\S")
```
JeremyCloarec: I think using format: "not-blank" instead is better to align with other apis :)

JeremyCloarec (approved):

PR #15671, opened 2026-04-23
[backend] add possiblity to mark playbooks to debug (#15668)

### LINKED ISSUE #15668: Playbook execution end 2 end tracking
## Use case

The objective is to determine where business rules execute can be lost.

The idea is to permit the tracking of a object based on it's id.
id -> log files -> all traces from event insertion to playbook and plyabook outcome upsert

To do that we need to:

    be able to track the insertion of the event in redis stream (global feature flag)
    track the paybook triggering
    track the playbook box execution
    track playbook upsert execution

We can have theses traces activated by playbook.
Feature flag + checkbox by playbook

JeremyCloarec:
### Proposed changes

* add possiblity to mark playbooks to debug
*

### Related issues

* fix #15668 
*

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/manager/playbookManager/playbookExecutor.ts
+++ b/opencti-platform/opencti-graphql/src/manager/playbookManager/playbookExecutor.ts
@@ -9,6 +9,7 @@ import { logApp } from '../../config/conf';
 import { UnsupportedError } from '../../config/errors';
 import { PLAYBOOK_COMPONENTS } from '../../modules/playbook/playbook-components';
 import type { StreamDataEvent } from '../../types/event';
+import { isDebugPlaybook } from './playbookManagerUtils';
 
 // Only way to force the step_literal checking
 // Don't try to understand, just trust
@@ -68,7 +69,6 @@ type ExecutorFn = {
   externalCallback?: {
     externalStartDate: Date;
   };
-  debugMode?: boolean;
 };
 
 export const playbookExecutor = async ({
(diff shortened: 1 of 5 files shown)


opencti-platform/opencti-graphql/src/manager/playbookManager/playbookExecutor.ts
```diff
@@ -120,6 +122,9 @@ export const playbookExecutor = async ({
         bundle: execution.bundle,
         forceBundleTracking: execution.forceBundleTracking ?? false,
       };
+      if (currentPlaybookInDebug) {
+        logApp.info(`[PLAYBOOK MANAGER] Registering step observation for playbook ${playbookId}`, { observation });
```
fellowseb, line 126: Side discussion:

Is there any guidelines about the logging ? I'm trying to find the right fit for `debug` vs `info`.
In my mind it would be nice to use the following rules (for managers, but can be extended:
- `info` level used to inform about actual work being done, when there is work only.
- `debug` level used to inform about init/recurring/cleanup steps done by managers.

In that way we wouldn't have logs polluted by recurring entries that happen every hour/minute in scheduled handlers. And we assume the manager is active and working as normal.

WDYT ?

fellowseb (approved):

opencti-platform/opencti-graphql/src/manager/playbookManager/playbookExecutor.ts
JeremyCloarec: There is no explicit guideline as of now, but I agree with your suggested guideline 👍 
The current issue we have in the app with debug level is that it's almost unusable in my opinion: currently, EVERY query to elastic is logged in debug mode, polluting way too much the logs... If this was changed, the logApp.info in this PR should be at debug level instead, but in the meantime that's why I put the logs at info level here :)

PR #15683, opened 2026-04-24
[backend] Standardize authentication management for feeds and synchronizers (#15682)

### LINKED ISSUE #15682: [Refactoring] Standardize authentication management for feeds and synchronizers

ludovic:
### Proposed changes

- Added generic `encryptDatabaseValue` and `decryptDatabaseValue` utilities in `platformCrypto.ts`.
- Updated the `connector.ts` domain and `syncManager.js` to process Synchronizer tokens through these utilities.
- Updated `ingestion-taxii`, `ingestion-csv`, and `ingestion-json` domains and their respective managers to process `authentication_value` through the same utilities.
- Added a DB migration (`1777898748017-encrypt-credentials-at-rest`) that encrypts all existing plain-text credentials at startup, ensuring no legacy unencrypted values remain in the database.
- Masking of `authentication_value` / `token` before returning to the GraphQL client is now handled at the resolver level (field resolvers on `IngestionCsv`, `IngestionJson`, `IngestionTaxii`), rather than in the domain layer.

### Related issues

* Fixes #15682 
*

### Checklist


opencti-platform/opencti-graphql/src/utils/platformCrypto.ts
```diff
@@ -265,3 +265,22 @@ const createPlatformCrypto = async () => {
 };
 
 export const getPlatformCrypto = memoize(createPlatformCrypto);
+
+const getGeneralKeyPair = memoize(async () => {
+  const factory = await getPlatformCrypto();
+  return factory.deriveAesKey(['general', 'credentials'], 1);
```
fellowseb: What's the rationale for this "general > credentials" derivation path ? Isn't there a risk that every use case will use this and thus in practice nullify the idea behind having multiple derivation paths (one for each use case) ?

I think I would rather have two derivation paths for this PR: `['ingestion', 'credentials']` and `['synchronizer', 'credentials']` , WDYT ?

opencti-platform/opencti-graphql/src/migrations/1776988800000-encrypt-credentials-at-rest.js
```diff
@@ -0,0 +1,76 @@
...
+  const encrypted = await encryptDatabaseValue(value);
+  await elUpdate(entity._index, entity.internal_id, { doc: { [fieldName]: encrypted } });
+};
+
+export const up = async (next) => {
+  logMigration.info(`${message} > started`);
+  const context = executionContext('migration');
+
+  const factory = await getPlatformCrypto();
+  const keyPair = await factory.deriveAesKey(['general', 'credentials'], 1);
```
fellowseb: Consider using `getGeneralKeyPair` to keep a single place of definition for the derivation path; or creating a `isDatabaseValueAlreadyEncrypted` utility next to it to address your need for this migration.

opencti-platform/opencti-graphql/src/utils/platformCrypto.ts
ludovic: Done, split into ['ingestion', 'credentials'] and ['synchronizer', 'credentials'].

opencti-platform/opencti-graphql/src/migrations/1776988800000-encrypt-credentials-at-rest.js
ludovic: Done, added `isIngestionCredentialEncrypted` and `isSynchronizerCredentialEncrypted` utilities in platformCrypto.ts, used directly in the migration.

fellowseb (approved):
Tested, works well 👍 .

Last thing: consider moving the crypto utils dedicated to specific domains (ingestion, synchronizer) to their respective modules.
You could still keep generic helpers in platformCrypto or a utils file next to platformCrypto:
```
type Encryptor = {
  encrypt: (data: Buffer, aad?: Buffer | undefined) => Promise<Buffer<ArrayBuffer>>;
};

type Decryptor = {
  decrypt: (data: Buffer, aad?: Buffer | undefined) => Promise<Buffer<ArrayBuffer>>;
};

export const encryptValue = async (encryptor: Encryptor, value: string | undefined | null) => {
  if (!value) return value;
  const clearDataBuffer = Buffer.from(value);
  const encryptedBuffer = await encryptor.encrypt(clearDataBuffer);
  return encryptedBuffer.toString('base64');
};

export const decryptValue = async (decryptor: Decryptor, value: string | undefined | null) => {
  if (!value) return value;
  return (await decryptor.decrypt(Buffer.from(value, 'base64'))).toString();
};
```

fellowseb (approved):

PR #15699, opened 2026-04-24
fix(backend/frontend): fix SMTP crash and infinite loading in XTM Hub connectivity check (#15681)

### LINKED ISSUE #15681: fix: checkXTMHubConnectivity crashes when SMTP email is not sent, causing infinite loading on the frontend
## Description

Front : 
When checkXTMHubConnectivity fails, the XTM Hub component does not appear, and the front displays infinite loading.
=> in this case there should be an error message.

Backend : 
A missing email makes the checkXTMHubConnectivity fail, which should not be the case.
=> Solution : add a try catch when no email (exemple when wrong SMTP email configuration) to avoid crashing and complete the request on the handleLostConnectivityEmail : 

Image

## Environment

4. Other environment details:

## Reproducible Steps

1. To reproduce infinite loading you can add an error in the checkXTMHubConnectivity request (xtm-hub.ts) :

Image

## Expected Output
=> No infinite loading on the front end when checkXTMHubConnectivity fails
=> request completes even if handleLostConnectivityEmail does not complete

JosephDoUrden:
### Proposed changes

* **Backend**: Wrap `handleLostConnectivityEmail` in a try/catch inside `checkXTMHubConnectivity`. When SMTP is misconfigured or the email fails to send, the error is now logged as a warning and the connectivity check completes normally instead of crashing.
* **Frontend**: Add `onError` callback to the `commitCheckConnectivity` mutation in `XtmHubSettings`. When the mutation returns an error, `setIsCheckDone(true)` is now called so the component exits the loading state and renders instead of spinning indefinitely.

### Related issues

* Fixes #15681

### Checklist

### Further comments

The backend fix is intentionally minimal: only the email-sending step is guarded, so SMTP failures do not propagate and abort the connectivity check. The rest of the function (status update, `xtm_hub_last_connectivity_check`, notification) continues as normal.

The frontend fix mirrors the existing `onCompleted` handler — both paths call `setIsCheckDone(true)`, ensuring the loader is dismissed whether the mutation succeeds or fails.

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/settings/xtm-hub/XtmHubSettings.tsx
+++ b/opencti-platform/opencti-front/src/private/components/settings/xtm-hub/XtmHubSettings.tsx
@@ -132,10 +132,15 @@ const XtmHubSettings: React.FC = () => {
   const [isCheckDone, setIsCheckDone] = useState(false);
 
   useEffect(() => {
-    commitCheckConnectivity({ variables: {},
+    commitCheckConnectivity({
+      variables: {},
       onCompleted: () => {
         setIsCheckDone(true);
-      } });
+      },
+      onError: () => {
+        setIsCheckDone(true);
+      },
+    });
   }, []);
 
   if (!isCheckDone) {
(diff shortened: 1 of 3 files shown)


Gwendoline-FAVRE-FELIX:
Thanks a lot for your contribution and for submitting this PR.
Before we can start the review, could you please sign your commit?

PR #15712, opened 2026-04-27
Issue/13389 5 preloaded query

fellowseb:
### Proposed changes

* Extract Initial preloaded query to another file
* Use the `@alias` Relay directive in the preloaded query for custom views so we can pass `queryData.customViews` to the `CustomViewsPreloadedDataContextProvider` and use Relay's garantee that it will get re-rendered only when custom-views-related data changes.
* This is the start of a pattern to refactor the entire `Root` component: move/create fragments in dedicated consumers (hook + context), create dedicated hook to access data instead of `useAuth`, use React context's to pass the queryData down but keep state management done by Relay

### Related issues

* 

### Checklist


opencti-platform/opencti-front/src/private/PrivateRootPreloadedQuery.tsx
```diff
@@ -0,0 +1,144 @@
...
+          theme_primary
+          theme_secondary
+          theme_text_color
+          theme_logo
+          theme_logo_collapsed
+          theme_logo_login
+        }
+      }
+    }
+    ...useCustomViews_data @alias(as: "customViews")
```
fellowseb, line 111: Using the `@alias` directive allows relying on the fact that our `CustomViewsPreloadedDataContextProvider` will get re-rendered only when custom views-related data gets changed.

PR #15725, opened 2026-04-27
[backend] update work completion in updateProcessedTime only if no expectation (#15737)

### LINKED ISSUE #15737: Work complete to early when using connector with QUEUE_PROTOCOL='api'
## Description

When using connector that split work in several pieces but into work work, when using QUEUE_PROTOCOL='api' option, the work get complete and deleted before all data has been ingested.

It lead to this error:
```
"Work is no longer alive, no request can be done within the context of this work"
```

## Environment

4. Other environment details:

## Reproducible Steps

1. { e.g. Run ... }
2. { e.g. Click ... }
3. { e.g. Error ... }

richard-julien:
Closes #15737

### DIFF
--- a/opencti-platform/opencti-graphql/src/domain/work.js
+++ b/opencti-platform/opencti-graphql/src/domain/work.js
@@ -395,15 +395,16 @@ export const updateProcessedTime = async (context, user, workId, message, inErro
     logApp.warn('The work cannot be found in database, processed time cannot be updated.', { workId });
     return workId;
   }
-  const { expected, total, isMultiPartWork } = await redisGetWorkCompletionState(workId);
-  const isComplete = isWorkFinished(expected, total);
+  const { expected, isMultiPartWork } = await redisGetWorkCompletionState(workId);
+  const isComplete = expected === 0; // Only complete if nothing expected
   if (isMultiPartWork && !isComplete) {
     await redisMarkWorkAsProcessed(workId);
   }
   const params = { processed_time: now(), message };
   let source = 'ctx._source["processed_time"] = params.processed_time;';
   if (isComplete) {
-    params.completed_number = total && !Number.isNaN(total) ? total : 1;
+    logApp.info('Work completed with no expectation', { workId, inError });
+    params.completed_number = 0;
     source += `ctx._source['status'] = "complete";
                ctx._source['import_expected_number'] = params.completed_number;
                ctx._source['completed_number'] = params.completed_number;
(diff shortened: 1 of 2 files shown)


JeremyCloarec:
Doesn't this go against what was decided on the implementation of this PR https://github.com/OpenCTI-Platform/opencti/pull/15465 ?
This fix/implementation will break the cases where the work is multipart, and the worker finishes processing the bundles before the connector calls toProcessed. In this case, the work will never be marked as complete.
Shouldn't the fix be donne on the connector side to properly mark the work as multipart instead?

opencti-platform/opencti-graphql/src/domain/work.js
```diff
@@ -395,20 +395,12 @@ export const updateProcessedTime = async (context, user, workId, message, inErro
...
   }
   const params = { processed_time: now(), message };
   let source = 'ctx._source["processed_time"] = params.processed_time;';
-  if (isComplete) {
-    params.completed_number = total && !Number.isNaN(total) ? total : 1;
-    source += `ctx._source['status'] = "complete";
-               ctx._source['import_expected_number'] = params.completed_number;
-               ctx._source['completed_number'] = params.completed_number;
-               ctx._source['completed_time'] = params.processed_time;`;
-  }
```
fellowseb, line 411: If I'm not mistaken I'm afraid by removing this we'll come into a case of a multi-part work that will never get completed if `toProcessed` is called after the last `reportExpectation`.

opencti-platform/opencti-graphql/src/domain/work.js
```diff
@@ -418,10 +410,5 @@ export const updateProcessedTime = async (context, user, workId, message, inErro
   }
   // Update elastic
   await elUpdate(currentWork._index, workId, { script: { source, lang: 'painless', params } });
-  // Remove redis work if needed
-  if (isComplete) {
-    await redisDeleteWorks([workId]);
-    logApp.info('Work completed', { workId, inError });
-  }
```
fellowseb, line 425: How about only removing this for now as a hotfix ?

opencti-platform/opencti-graphql/src/domain/work.js
richard-julien: It will be completed with the background process that close work when other work is started. I think its good enough

richard-julien: Without removing this, the job is marked as completed before in the normal case. I prefer to have the edge case handle sometimes by the manager instead of the normal case done wrong

opencti-platform/opencti-graphql/src/domain/work.js
richard-julien: Yes it could be if we still want to have completion of the work just after creation

aHenryJard (approved):
Tested locally for non regression on enrichment connector (hygiene), export stix connector, draft and background tasks.

opencti-platform/opencti-graphql/src/domain/work.js
fellowseb: Right !

JeremyCloarec (approved):

fellowseb (approved):

PR #15759, opened 2026-04-29
[backend] Fix playbook filtering cache management (#14285)

MTorbay-Filigran:
### Proposed changes

* Added the applyWithFilters config to the playbook cache filters + added some tests to cover the filtering with every entities such as labels, author, markings...

### Related issues

* #14285 

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/tests/03-integration/01-database/playbookComponents/container-wrapper-component-test.ts
+++ b/opencti-platform/opencti-graphql/tests/03-integration/01-database/playbookComponents/container-wrapper-component-test.ts
@@ -5,6 +5,7 @@ import { testBundleObject, testExecutor } from './playbook-components-test-utils
 import type { StixRelation } from '../../../../src/types/stix-2-1-sro';
 import { playbookBundleElementsToApply } from '../../../../src/modules/playbook/playbook-types';
 import { STIX_EXT_OCTI } from '../../../../src/types/stix-2-1-extensions';
+import type { StixDomainObject } from '../../../../src/types/stix-2-1-common';
 
 const componentConfig = (config: Partial<ContainerWrapperConfiguration>) => {
   return {
@@ -21,11 +22,13 @@ describe('PLAYBOOK_CONTAINER_WRAPPER_COMPONENT', () => {
   const MALWARE_ID = 'malware--09bd862a-f030-55f2-920a-900c4913d9ff';
   const CAMPAIGN_ID = 'campaign--6bcf59ca-70c8-55ae-ac7d-a6f9b107a35b';
   const REL_ID = 'relationship--08e64f51-e890-5bec-be34-3344746f1b0c';
+  const LABEL_ID = '98d475ca-0f72-4878-8d64-de2e094f007e';
 
   const BUNDLE_OBJECTS = () => [
-    testBundleObject({
+    testBundleObject<StixDomainObject>({
       id: MALWARE_ID,
       type: 'Malware',
+      labels: [LABEL_ID],
     }),
     testBundleObject({
       id: CAMPAIGN_ID,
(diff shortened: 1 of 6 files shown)


opencti-platform/opencti-graphql/tests/03-integration/01-database/playbookComponents/container-wrapper-component-test.ts
```diff
@@ -168,6 +172,21 @@ describe('PLAYBOOK_CONTAINER_WRAPPER_COMPONENT', () => {
       expect(reportResult.object_refs).toEqual([]);
     });
 
+    it('should add only the matching label element with label filter (all-elements)', async () => {
+      const result = await PLAYBOOK_CONTAINER_WRAPPER_COMPONENT.executor(testExecutor({
+        mainId: MALWARE_ID,
+        bundleObjects: BUNDLE_OBJECTS(),
+        configuration: componentConfig({
+          applyToElements: playbookBundleElementsToApply.allElements.value,
+          applyWithFilters: filterLabel,
```
Archidoit, line 181: add a test with applyWithFilters = undefined or null

MTorbay-Filigran: Normally we covered the case when there is a bundle with multiple elements, scoped with applyToElements and no filters such as here : 
```
it('should wrap all elements in Container except main element', async () => {
    const result = await PLAYBOOK_CONTAINER_WRAPPER_COMPONENT.executor(testExecutor({
      mainId: MALWARE_ID,
      bundleObjects: BUNDLE_OBJECTS(),
      configuration: componentConfig({
        applyToElements: playbookBundleElementsToApply.allExceptMain.value,
      }),
    }));

    const reportResult = result.bundle.objects
      .find((element) => element.type === 'report') as StixContainer;
    expect(reportResult.object_refs).toEqual([CAMPAIGN_ID, REL_ID]);
  });
```

opencti-platform/opencti-graphql/tests/03-integration/01-database/playbookComponents/create-indicator-component-test.ts
```diff
@@ -13,14 +13,16 @@ import type { StixInternalExternalReference } from '../../../../src/types/stix-2
 const OBSERVABLE_ID = 'domain-name--a1b2c3d4-0000-0000-0000-000000000001' as const;
 const INTRUSION_SET_ID = 'intrusion-set--1ad04810-ab05-5873-96f5-a89d19607e1c' as const;
 const REPORT_ID = 'report--b4754e7d-88b4-51d9-aac4-86edaad66c4d' as const;
+const LABEL_ID = '98d475ca-0f72-4878-8d64-de2e094f007e';
 
 const FAKE_PATTERN = "[domain-name:value = 'malicious.example.com']";
 
-const domainObservable = (id: StixId = OBSERVABLE_ID) =>
-  testBundleObject<StixCyberObject & { value: string }>({
+const domainObservable = (id: StixId = OBSERVABLE_ID, withLabel?: boolean) =>
```
lndrtrbn: It would be a better pattern to use an approach similar to the function `componentConfig` _(see L182 in manipulate test file)_ we use to construct the configuration with a partial object that you can directly spread into the base object instead of making a ternary for each attribute you want to add

opencti-platform/opencti-graphql/tests/03-integration/01-database/playbookComponents/create-indicator-component-test.ts
```diff
@@ -13,14 +13,16 @@ import type { StixInternalExternalReference } from '../../../../src/types/stix-2
...
+  testBundleObject<StixCyberObject & { value: string; labels: string[] }>({
     id,
     type: 'domain-name',
     value: 'malicious.example.com',
+    ...(withLabel ? { labels: [LABEL_ID] } : {}),
```
lndrtrbn: I am not sure of putting labels at this place. In STIX objects they are normally in the extensions. Do you find something that said otherwise?

lndrtrbn (approved):

PR #15766, opened 2026-04-30
[backend] Allow using a rate limit base on both IP and user-agent (#15765)

### LINKED ISSUE #15765: Apply rate limit threshold per IP and per user-agent
## Use case

In current OpenCTI implementation, the overall rate limit on application is only per IP.
It means that in organizations with users being behind a proxy it is diffcult to use.

It will be more easy to use if the limit is applied per unique ( IP + user-agent)

Yes

aHenryJard:
### Proposed changes

* Use user-agent in rate limit key in addition of IP

### Related issues

* closes #15765 

### Checklist


opencti-platform/opencti-graphql/src/http/httpUtils.ts
```diff
@@ -171,16 +174,97 @@ export const buildDefaultHelmetParameters = () => {
...
+  const now = Date.now();
+  const lastLogged = rateLimitLogThrottle.get(key);
+  if (lastLogged === undefined || now - lastLogged >= RATE_LIMIT_LOG_INTERVAL_MS) {
+    rateLimitLogThrottle.set(key, now);
+    logApp.info('[RATE-LIMIT] Rate limited request', { ip, userAgent });
+  }
+};
+
 export const buildRateLimiterOptions = (): Options => {
+  console.log('buildRateLimiterOptions');
```
CelineSebe: console.log should be removed

opencti-platform/opencti-graphql/src/http/httpUtils.ts
```diff
@@ -171,16 +174,97 @@ export const buildDefaultHelmetParameters = () => {
...
+  const ip = ipKeyGenerator(req.ip ?? 'unknown');
+  const userAgent = req.headers['user-agent'] ?? 'unknown';
+  return crypto.createHash('sha256').update(`${ip}|${userAgent}`).digest('hex');
+};
+
+/**
+ * Build a BlockList from CIDR ranges for efficient IP range matching.
+ */
+const buildIpRangeSkipList = (ranges: string[]): BlockList => {
+  console.log('buildIpRangeSkipList');
```
CelineSebe: same

opencti-platform/opencti-graphql/src/http/httpUtils.ts
```diff
@@ -171,16 +174,97 @@ export const buildDefaultHelmetParameters = () => {
...
+const RATE_LIMIT_LOG_INTERVAL_MS = 60_000; // 1 minute
+
+/**
+ * Log a rate-limit event for an IP + User-Agent pair at most once per minute.
+ */
+const logRateLimitThrottled = (ip: string, userAgent: string): void => {
+  const key = `${ip}|${userAgent}`;
+  const now = Date.now();
+  const lastLogged = rateLimitLogThrottle.get(key);
+  if (lastLogged === undefined || now - lastLogged >= RATE_LIMIT_LOG_INTERVAL_MS) {
```
CelineSebe, line 235: code suggestion: set up periodic cleanup (delete entries older than RATE_LIMIT_LOG_INTERVAL_MS)

```
 // Evict stale entries periodically
    if (rateLimitLogThrottle.size > MAX_THROTTLE_ENTRIES) {
      for (const [k, v] of rateLimitLogThrottle) {
        if (now - v >= RATE_LIMIT_LOG_INTERVAL_MS) rateLimitLogThrottle.delete(k);
      }
    }
```

opencti-platform/opencti-graphql/src/http/httpUtils.ts
```diff
@@ -171,16 +174,97 @@ export const buildDefaultHelmetParameters = () => {
...
+  return crypto.createHash('sha256').update(`${ip}|${userAgent}`).digest('hex');
+};
+
+/**
+ * Build a BlockList from CIDR ranges for efficient IP range matching.
+ */
+const buildIpRangeSkipList = (ranges: string[]): BlockList => {
+  console.log('buildIpRangeSkipList');
+  const blockList = new BlockList();
+  try {
```
CelineSebe: The try/catch block covers the entire loop. If even one CIDR entry is invalid, the subsequent entries will not be added

suggestion:

```
 for (const range of ranges) {
  try {
    if (range.includes('/')) {
      const [subnet, prefixStr] = range.split('/');
      const prefix = parseInt(prefixStr, 10);
      const type = subnet.includes(':') ? 'ipv6' : 'ipv4';
      blockList.addSubnet(subnet, prefix, type);
    } else {
      const type = range.includes(':') ? 'ipv6' : 'ipv4';
      blockList.addAddress(range, type);
    }
  } catch (e: any) {
    logApp.warn('[HTTP] Invalid IP range entry ignored in rate limit skip config', { range, error: e.message });
  }
}
```

opencti-platform/opencti-graphql/src/http/httpUtils.ts
aHenryJard: oops yes thanks

opencti-platform/opencti-graphql/src/http/httpUtils.ts
aHenryJard: I don't think that there will be that much entries, but why not ok.

CelineSebe (approved):

PR #15809, opened 2026-05-04
[frontend][backend] Export & import custom views (#13389)

fellowseb:
### Proposed changes

This PR implements both "Import a custom view" and "Export a custom view" chunks

Similar behavior to what is done in `Custom dashboards` except the exported JSON has `type: 'custom-view'`.

### Related issues

### Checklist


opencti-platform/opencti-graphql/src/modules/customView/customView-domain.ts
```diff
@@ -363,3 +366,59 @@ export const deleteCustomView = async (
...
+  file: Promise<FileHandle>,
+) => {
+  if (!isCustomViewsAvailableForEntityType(targetEntityType)) {
+    throw FunctionalError(
+      'Custom views cannot be created for given entity type', {
+        entityType: targetEntityType,
+      });
+  }
+  const parsedData: CustomViewExport = await extractContentFrom(file);
+  const { manifest } = parsedData.configuration;
```
fellowseb, line 400: Unclear if/why I should check the version of `openCTI_version` here. And if so against what version ?

opencti-platform/opencti-front/src/components/dashboard/DashboardContent.tsx
```diff
@@ -6,12 +6,16 @@ import DashboardViz from './DashboardViz';
...
-  isEditable: boolean;
+type DashboardContentProps = {
   helpers: ReturnType<typeof useDashboard>;
   host: WidgetHost;
-}
+} & ({
+  entity: DashboardLike;
+  isEditable: true;
+} | {
+  entity: Pick<DashboardLike, 'id' | 'manifest'>;
```
esrevi: i'd suggest creating a type such as type DashboardWithManifest = Pick<DashboardLike, 'id' | 'manifest'> for example, or another name ^^ because it s used in different part of the code

opencti-platform/opencti-front/src/components/dashboard/useDashboardImportExport.tsx
```diff
@@ -0,0 +1,56 @@
...
+  configType: string;
+  entity: Pick<DashboardLike, 'id' | 'name'>;
+}
+
+export const useDashboardExport = ({
+  onExport,
+  configType,
+  entity,
+}: useDashboardImportExportProps) => {
+  const handleExport = () => {
```
esrevi: should this be a async function ?

opencti-platform/opencti-front/src/components/dashboard/useDashboardImportExport.tsx
```diff
@@ -0,0 +1,56 @@
...
+    onExport(entity.id)
```
esrevi: can we use async syntax?

opencti-platform/opencti-front/src/components/dashboard/useDashboardImportExport.tsx
```diff
@@ -0,0 +1,56 @@
...
+import VisuallyHiddenInput from '../../private/components/common/VisuallyHiddenInput';
+import { DashboardLike } from './dashboard-types';
+
+interface useDashboardImportExportProps {
+  onExport: (entityId: string) => Promise<string | null>;
+  configType: string;
+  entity: Pick<DashboardLike, 'id' | 'name'>;
+}
+
+export const useDashboardExport = ({
```
esrevi: do we need this as a hook ? or plain js function is enough ? from what i see, we don't have a state in that hook ?

opencti-platform/opencti-front/src/components/dashboard/useDashboardImportExport.tsx
```diff
@@ -0,0 +1,56 @@
...
+        inputRef.current.value = '';
+      }
+    });
+  };
+
+  const handleImport = () => inputRef.current?.click();
+  return { _onChange, handleImport, _inputRef: inputRef };
+};
+
+export const DashboardHiddenImportInput = ({ helpers }: { helpers: ReturnType<typeof useDashboardImport> }) => {
```
esrevi: we should not have this component in the hook file

opencti-platform/opencti-front/src/components/dashboard/useDashboardImportExport.tsx
```diff
@@ -0,0 +1,56 @@
...
+  };
+  return { handleExport };
+};
+
+export const useDashboardImport = ({ onImport }: {
+  onImport: (file: File) => Promise<void>;
+}) => {
+  const inputRef = useRef<HTMLInputElement | null>(null);
+
+  const _onChange = (event: BaseSyntheticEvent) => {
```
esrevi: _onChange, _inputRef is a convention to tell it's private, but we export them, so there's no need imo, keep onChange, inputRef instead

opencti-platform/opencti-front/src/private/components/settings/sub_types/custom_views/CustomViewEditionHeader.tsx
```diff
@@ -19,6 +19,8 @@ import type { Widget, WidgetHost } from '../../../../../utils/widget/widget';
 import CustomViewMenu from './CustomViewMenu';
 import type { Theme } from '../../../../../components/Theme';
 import useCustomViewEdit from './useCustomViewEdit';
+import ExportButtons from 'src/components/ExportButtons';
```
esrevi: prefer relative paths

opencti-platform/opencti-front/src/components/ExportButtons.jsx
```diff
@@ -186,6 +186,8 @@ class ExportButtons extends Component {
       handleExportDashboard,
       investigationAddFromContainer,
       navigate,
+      exportToImage = true,
```
esrevi, line 168: won't there be consequences setting them to true by default ?

opencti-platform/opencti-front/src/private/components/settings/sub_types/custom_views/useCustomViewImportExport.tsx
```diff
@@ -0,0 +1,69 @@
...
+    .toPromise();
+  const result = data as useCustomViewImportExport_Query$data;
+  const exportString = result.customView?.toConfigurationExport;
+  if (!exportString) {
+    throw new Error('Failed to export custom view');
+  }
+  return exportString;
+};
+
+export const useCustomViewExport = (customView: { id: string; name: string }) => {
```
esrevi: very weird here a hook returning a hook, I think it should just be plain js

opencti-platform/opencti-front/src/private/components/settings/sub_types/custom_views/useCustomViewImportExport.tsx
```diff
@@ -0,0 +1,69 @@
...
+        }
+        resolve();
+      },
+    });
+  });
+  const helpers = useDashboardImport({ onImport });
+  return { importing, ...helpers };
+};
+
+export const CustomViewHiddenImportInput = ({ helpers }: { helpers: ReturnType<typeof useDashboardImport> }) => {
```
esrevi: this component has no reason to exist, it returns exactly DashboardHiddenImportInput

opencti-platform/opencti-front/src/private/components/settings/sub_types/custom_views/useCustomViewImportExport.tsx
```diff
@@ -0,0 +1,69 @@
...
+  return <DashboardHiddenImportInput helpers={helpers} />;
+};
+
+const onExport = async (id: string) => {
+  const data = await fetchQuery(customViewExportQuery, { id })
+    .toPromise();
+  const result = data as useCustomViewImportExport_Query$data;
+  const exportString = result.customView?.toConfigurationExport;
+  if (!exportString) {
+    throw new Error('Failed to export custom view');
```
esrevi: we throw an error here, should it return null instead?

opencti-platform/opencti-front/src/private/components/workspaces/WorkspaceCreation.tsx
```diff
@@ -57,32 +57,30 @@ interface WorkspaceCreationProps {
...
     ? `${settings.platform_xtmhub_url}/redirect/opencti_custom_dashboards?platform_id=${settings.id}`
     : '';
-
   const [commitImportMutation] = useApiMutation<WorkspaceCreationImportMutation>(importMutation);
-  const [commitCreationMutation] = useApiMutation(workspaceMutation);
   const navigate = useNavigate();
-
-  const handleImport = (event: BaseSyntheticEvent) => {
-    const importedFile = event.target.files[0];
+  const onImport = (file: File) => new Promise<void>((resolve, reject) => {
```
esrevi: prefer follow the convention, onSomething for props, and handleSomething if local function

opencti-platform/opencti-front/src/components/dashboard/useDashboardImportExport.tsx
fellowseb: It's on purpose: they're kinda private because  not intended to be used by consumer code

opencti-platform/opencti-front/src/components/dashboard/DashboardContent.tsx
fellowseb: Created a `ExportableDashboardLike` type with id + name, as the reason for this was really to be used by `useDashboardImportExport` hooks.

opencti-platform/opencti-front/src/components/dashboard/useDashboardImportExport.tsx
fellowseb: Hum we don't have the need right now. It's a bit too futureproofing but ok.
Done.

opencti-platform/opencti-front/src/components/dashboard/useDashboardImportExport.tsx
fellowseb: done

opencti-platform/opencti-front/src/private/components/settings/sub_types/custom_views/CustomViewEditionHeader.tsx
fellowseb: Well, there is a `src` alias in tsconfig (:. I'm changing but I think we should rather fix the tooling and configs than discuss stylistic issues in PRs

opencti-platform/opencti-front/src/private/components/settings/sub_types/custom_views/useCustomViewImportExport.tsx
fellowseb: Done, well caught

opencti-platform/opencti-front/src/components/ExportButtons.jsx
fellowseb: It doesn't change current behaviour

opencti-platform/opencti-front/src/components/dashboard/useDashboardImportExport.tsx
fellowseb: Hmm. I'm not a big fan of fragmenting related things in several files just "because". (That's also one of the reasons I don't like default exports.) In this case IMO it makes things easier to understand just having a single file. Anyways...extracted. Done.

opencti-platform/opencti-front/src/components/dashboard/useDashboardImportExport.tsx
fellowseb: Fixed, moved to non-hook in another file

opencti-platform/opencti-front/src/components/dashboard/useDashboardImportExport.tsx
fellowseb: Changed

opencti-platform/opencti-front/src/private/components/settings/sub_types/custom_views/useCustomViewImportExport.tsx
fellowseb: Changed

opencti-platform/opencti-front/src/private/components/settings/sub_types/custom_views/useCustomViewImportExport.tsx
fellowseb: Removed though aliases are something useful in general

opencti-platform/opencti-graphql/src/modules/customView/customView-domain.ts
fellowseb: So the `openCTI_version` is currently added on export but not checked on import.
We'll add the import check if/when we have a major/breaking change in the exported model.

PR #15814, opened 2026-05-04
feat(client): add example script to modify exclusion list (#15824)

### LINKED ISSUE #15824: docs: add an example Python script to update the exclusion list
### Documentation Request

_No response_

emmacgz:
### Proposed changes

* new script to modify an Exclusion List with a .txt file

### Related issues

* closes #15824 15824

### Checklist

### DIFF
--- a/client-python/examples/modify_exclusion_list.py
+++ b/client-python/examples/modify_exclusion_list.py
@@ -1,0 +1,46 @@
+import requests
+
+# Variables
+api_url = #TO_COMPLETE (ex : "http://opencti:4000")
+api_token = #TO_COMPLETE
+exclusion_list_id = #TO_COMPLETE (ex :"y50ea9c3-2aaa-4843-aa94-f0235162c6bf")
+txt_file_path = #TO_COMPLETE (ex : "/my_exclusion_list.txt")
+
+MUTATION = """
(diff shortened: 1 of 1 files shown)


client-python/examples/modify_exclusion_list.py
```diff
@@ -0,0 +1,46 @@
+import requests
+
+# Variables
+api_url = #TO_COMPLETE (ex : "http://opencti:4000")
+api_token = #TO_COMPLETE
+exclusion_list_id = #TO_COMPLETE (ex :"y50ea9c3-2aaa-4843-aa94-f0235162c6bf")
+txt_file_path = #TO_COMPLETE (ex : "/my_exclusion_list.txt")
```
aHenryJard, line 7: You need to use an actual example that works in CI.
For api_url and token you can use environment variable like it's done on other examples, for example here https://github.com/OpenCTI-Platform/opencti/blob/db704e2221ce65f9aea2335c59295e1460fc46a1/client-python/examples/add_external_reference_to_report.py#L9-L10

aHenryJard:
As seen in direct message, could you please move your example to documentation https://github.com/OpenCTI-Platform/opencti/blob/master/docs/docs/administration/exclusion-lists.md ? It's not a use of pycti so it's not appropriate to add in client python. We don't have better place for now.

Issue #14054, opened 2026-01-16
feat(workflow): ability to comment on a transition & view the comment in the UI


nino-filigran:
## Use case

Among one of the action needed, one would be to be able to drop a comment when applying a specific transition, to explain a rejection for instance.
At first, we do not plan to have a full lifecycle of this comment, but only a comment attached to a specific stauts and when the status is changed, the comment would be lost. 

## Proposed Solution

Ability to have in UI & Back End the abilty to add some text hen chosing a specific transition.
Ability to view the comment in UI
## If the feature request is approved, would you be willing to submit a PR?

Yes / No (Help can be provided if you need assistance submitting a PR)

PR #15829, opened 2026-05-05
[frontend/backend] ability to leave a comment on a transition (#14054)

frapuks:
### Proposed changes

* Add “comment” to the transition actions
* If “comment” is enabled, a comment can be entered when the status changes
* Save the comment to the history
* Display the comment if it was included in the most recent entry in the history

### Related issues

* #14054 

### Checklist


opencti-platform/opencti-graphql/src/modules/workflow/workflow-validation.ts
```diff
@@ -46,6 +46,7 @@ export const workflowSerializedTransitionSchema = z.object({
   from: z.union([z.string().max(255), z.array(z.string().max(255))]).nullable(),
   to: z.string().max(255).nullable(),
   event: z.string().max(255),
+  comment: z.string().max(255).optional(),
```
OctaveLaventure:
```suggestion
  comment: z.enum(['disable', 'allowed', 'required']).optional(),
```

OctaveLaventure (commented):
What happens when a transition fails? Is the comment still saved? if so, do we see it in the front?

opencti-platform/opencti-front/src/private/components/settings/sub_types/workflow/utils.ts
```diff
@@ -19,10 +19,13 @@ export type Status = {
   onExit?: Action[];
 };
 
+export type CommentMode = 'disable' | 'allowed' | 'required';
```
delemaf:
```suggestion
enum CommentMode {
	disable,
	allowed,
	required,
}
export type CommentModeType = keyof typeof CommentMode;
```

opencti-platform/opencti-front/src/private/components/settings/sub_types/workflow/TransitionForm.tsx
```diff
@@ -14,6 +14,18 @@ const TransitionForm = () => {
   const hasUpdateAuthorizedMembers = values.actions?.some((a) => a.type === WorkflowActionType.updateAuthorizedMembers);
   const hasValidateDraft = values.actions?.some((a) => a.type === WorkflowActionType.validateDraft);
 
+  const commentMode: CommentMode = values.comment ?? 'disable';
```
delemaf: You can use CommentMode enum

opencti-platform/opencti-front/src/private/components/settings/sub_types/workflow/hooks/useWorkflowInitialElements.ts
```diff
@@ -75,13 +75,14 @@ export const useWorkflowInitialElements = (
...
     const transitionNodes: Node[] = workflowDefinition.transitions
-      .map(({ from, to, event, conditions = {}, actions = [] }) => ({
+      .map(({ from, to, event, conditions = {}, actions = [], comment }) => ({
         id: `${WorkflowNodeType.transition}-${from}-${to}`,
         type: WorkflowNodeType.transition,
         data: {
           event,
           conditions,
           actions: parseActions(actions),
+          comment: (comment ?? 'disable') as CommentMode,
```
delemaf: You can use CommentMode enum

opencti-platform/opencti-graphql/src/modules/workflow/api/workflow-resolvers.ts
```diff
@@ -32,18 +35,27 @@ const workflowResolvers = {
     workflowDefinitionDelete: (_: any, { entityType }: { entityType: string }, context: AuthContext) => {
       return deleteWorkflowDefinition(context, context.user!, entityType);
     },
-    triggerWorkflowEvent: (_: any, { entityId, eventName }: { entityId: string; eventName: string }, context: AuthContext) => {
-      return triggerWorkflowEvent(context, context.user!, entityId, eventName);
+    triggerWorkflowEvent: (_: any, { entityId, eventName, comment }: { entityId: string; eventName: string; comment?: string | null }, context: AuthContext) => {
+      const normalizedComment = comment?.trim() ?? undefined;
```
delemaf, line 39: Maybe this is safer to prevent injection and weird content
https://github.com/OpenCTI-Platform/opencti/blob/6a92e4f40ff8d5d37cf493b9808dbaff3b272806/opencti-platform/opencti-front/src/utils/String.jsx#L298-L310

delemaf: Since `sanitize` is only a frontend function we'll let this part as it is

frapuks:
> What happens when a transition fails? Is the comment still saved? if so, do we see it in the front?

For now, the comment won't be saved until the transition is complete, so you won't see anything.

opencti-platform/opencti-front/src/private/components/settings/sub_types/workflow/TransitionForm.tsx
```diff
@@ -14,6 +14,18 @@ const TransitionForm = () => {
   const hasUpdateAuthorizedMembers = values.actions?.some((a) => a.type === WorkflowActionType.updateAuthorizedMembers);
   const hasValidateDraft = values.actions?.some((a) => a.type === WorkflowActionType.validateDraft);
 
+  const commentMode: CommentModeType = values.comment ?? CommentMode.disable;
+  const enableComments = commentMode !== 'disable';
+  const requireComments = commentMode === 'required';
```
OctaveLaventure: Comparing to `CommentMode.<value>` here and in a few other places rather than hardcoded strings like 'disable' or 'required' would be cleaner.

Why:
First, it makes it easier to quickly get to the enum and check/update other possible values.

Also, although it usually won't make a big difference as you should have a TS error if the string is not included in the enum, it's not impossible that after a few refactors later on the type inference could be downgraded to a more permissive type such as `string` in some places. In that case, it wouldn't trigger an error if the string was wrong anymore.

opencti-platform/opencti-front/lang/front/de.json
```diff
@@ -1488,6 +1494,8 @@
   "Enable AI powered platform": "KI-gestützte Plattform aktivieren",
   "Enable Ask Ariane": "Aktivieren Sie Ask Ariane",
   "Enable client certificate authentication": "Client-Zertifikatsauthentifizierung aktivieren",
+  "Enable comment": "Kommentar einschalten",
+  "Enable comments": "Kommentare einschalten",
```
OctaveLaventure: I think the `Enable comments` key is not used anywhere

opencti-platform/opencti-front/src/private/components/settings/sub_types/workflow/TransitionForm.tsx
frapuks: Use CommentMode in every places

opencti-platform/opencti-front/lang/front/de.json
frapuks: removed

OctaveLaventure (approved):
We could rename `disable` to `disabled` to match the wording of `allowed` and `required`, but if it's too much of a hassle to update don't bother with it

delemaf (approved):

PR #15836, opened 2026-05-05
fix(backend): mark work as complete on ES in updateProcessedTime (#15813)

### LINKED ISSUE #15209: fix(export): exporting countries in PDF triggers a second failing export
## Description

Exporting countries in pdf triggers a second failing export

## Environment

1. Testing
2. Demo

## Reproducible Steps

1. Go to Locations > Countries (or Regions)
2. Open Export panel
3. lauch a simple PDF export

## Expected Output

1 exported PDF with all the courntries selected

## Actual Output

1 exported PDF with all the courntries selected then a second export appears in the panel, spinning for 10 minutes before giving error
Image

Image

### LINKED ISSUE #15813: fix(export): exporting a report produces 2 files each time
## Description

https://github.com/user-attachments/assets/ddd7ff82-4feb-4d68-b5b8-43ffc5fbdfbb

I have tried exporting in stix or pdf, but I think other export type will result the same, it doubles the files, one file is successfully exported and the other is hanging on "processing" state

## Environment

1. OS (where OpenCTI server runs): Mac OS 10, Windows 10, Ubuntu 16.4,
2. OpenCTI version: { e.g. OpenCTI 1.0.2 } 7.260430.0
3. OpenCTI client: frontend
4. Other environment details:

## Reproducible Steps

1. Create a report
2. Export stix, pdf, csv or other
3. One file is exported correctly, but it doubles, with another file export process that hangs on processing

## Expected Output

Only one file is exported at the end of the process

### LINKED ISSUE #16049: fix(filters): `entity_type` filter unconditionally expands to `parent_types`, leaking History/Activity documents into concrete-type queries
## Description

The filter adapter `adaptFilterToEntityTypeFilterKey` in `opencti-platform/opencti-graphql/src/utils/filtering/filtering-completeSpecialFilterKeys.ts` unconditionally expands any `entity_type` filter to also match against `parent_types`. This is correct for **abstract** STIX types (e.g. `Stix-Cyber-Observable`, which must match its concrete subtypes via `parent_types`) but incorrect for **concrete** types such as `Work`.

When a concrete-type filter is expanded this way and the target Elasticsearch index also contains documents whose `parent_types` array references that concrete type (notably `History` and `Activity` documents in `INDEX_HISTORY`), those documents leak into the result set. The GraphQL resolver then tries to project them as the requested type, producing null fields and, where the schema demands non-null, hard resolver errors.

The most user-visible manifestation is on the **Exported Files** panel of any entity: every export produces what appears to be a duplicate "Pending" entry that never resolves. See **Actual Output** below.

## Environment

- OpenCTI version: 7.260515.0

## Reproducible Steps

1. Open any Case-Incident entity in the OpenCTI UI.
2. Click **Generate Export**, select format `application/vnd.oasis.stix+json`, choose any TLP marking, and submit.
3. Open the **Data** tab → **Exported Files** panel on the same entity.
4. Observe two entries with identical millisecond timestamps: one completes, one remains "Pending" indefinitely.
5. Optionally, run the following query in the GraphQL playground to confirm at the API layer:

```graphql
query RecentWorks {
  works(first: 30, orderBy: timestamp, orderMode: desc) {
    edges {
      node {
        id
        name
        timestamp
        received_time
        processed_time
        completed_time
        connector { id name connector_type }
      }
    }
  }
}
```

Observe that the majority of returned edges have `name: null`, `connector: null`, and IDs in Redis-stream format (e.g. `1779113707837-0`) rather than the expected `work_<connector-uuid>_<ISO-timestamp>` format.

## Expected Output

- The Exported Files panel shows exactly **one** entry per export request, which transitions through its lifecycle (`wait` → `progress` → `complete`).
- The `works` GraphQL query returns only documents whose `entity_type` is `Work`.

## Actual Output

- The Exported Files panel shows **two** entries per export with identical millisecond timestamps. One completes; the other remains "Pending" indefinitely. `work.to_processed` does not clear the pending entry; only `work.delete` removes it.
- The `works` query returns the real `Work` plus every `History`/`Activity` document in `INDEX_HISTORY` whose `parent_types` array contains `"Work"`.

In one captured run, 30 edges were returned: 2 real `Work` documents and 28 leaked `History` documents with null `name`, null `connector`, and Redis-stream-style IDs. A separate run that included the `status` field (non-nullable on the `Work` GraphQL type) produced 17 of 20 rows as `INTERNAL_SERVER_ERROR: Cannot return null for non-nullable field Work.status`, because the leaked History documents have no `status` field.

### Root cause

In `opencti-platform/opencti-graphql/src/utils/filtering/filtering-completeSpecialFilterKeys.ts`:

```ts
const adaptFilterToEntityTypeFilterKey = (filter) => {
  const arrayKeys = Array.isArray(filter.key) ? filter.key : [filter.key];
  const newFilter = { ...filter, key: arrayKeys.concat(['parent_types']) };
  return newFilter;
};
```

Downstream in `opencti-platform/opencti-graphql/src/database/engine.js` (`elPaginate` → `elQueryBodyBuilder` → `buildSubQueryForFilter`), the expanded filter renders as a multi-field match:

```json
{
  "multi_match": {
    "fields": ["entity_type.keyword", "parent_types.keyword"],
    "query": "work"
  }
}
```

This matches any document whose `parent_types` array contains `"Work"`, not just documents whose `entity_type` is `Work`.

#### Why this surfaces specifically for `Work` queries

- `Work`, `History`, and `Activity` documents share the same Elasticsearch index (`INDEX_HISTORY` / `opencti_history`).
- When a `Work` is created or updated, the corresponding `History` audit entry's `parent_types` array includes `"Work"`.
- The keyword normalizer lowercases values at index time, so case-sensitivity does not filter them out.

The same defect powers `worksForSource`, which backs the **Exported Files** UI panel — hence the user-visible symptom.

#### Why every previous symptom fits

- **Identical-ms "duplicate" entries in the UI**: the History audit entry for the Work creation shares the Work's timestamp (same write transaction).
- **`work.to_processed` does not clear the pending entry**: the second entry is not a Work; the mutation has nothing to update.
- **`work.delete` clears it**: deletes by ID, and the History document's ID is addressable through the resolver.
- **Affects all `INTERNAL_EXPORT_FILE` connectors**: every export creates one Work and at least one History audit entry; both leak together.
- **"Pending" never resolves**: the leaked History document has no `status`, so the UI defaults to "Pending" and no lifecycle event will ever advance it.

## Scope of impact

The defect affects **any** GraphQL query that filters on a concrete `entity_type` against an index containing documents whose `parent_types` reference that type. Known affected surfaces:

- `works` and `worksForSource` (Exported Files panel; Connectors → Works panel)
- Any query against `INDEX_HISTORY` that filters by concrete type
- Any future feature filtering on a concrete type that shares an index with related History/Activity documents

## Screenshots

Image

JeremyCloarec:
### Proposed changes

* mark work as complete on ES in updateProcessedTime. This reverts in part changes done in https://github.com/OpenCTI-Platform/opencti/pull/15725
*

### Related issues

* fix #15813 
*

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/domain/work.js
+++ b/opencti-platform/opencti-graphql/src/domain/work.js
@@ -395,12 +395,20 @@ export const updateProcessedTime = async (context, user, workId, message, inErro
     logApp.warn('The work cannot be found in database, processed time cannot be updated.', { workId });
     return workId;
   }
-  const { isMultiPartWork } = await redisGetWorkCompletionState(workId);
-  if (isMultiPartWork) {
+  const { expected, total, isMultiPartWork } = await redisGetWorkCompletionState(workId);
+  const isComplete = isWorkFinished(expected, total);
+  if (isMultiPartWork && !isComplete) {
     await redisMarkWorkAsProcessed(workId);
   }
   const params = { processed_time: now(), message };
   let source = 'ctx._source["processed_time"] = params.processed_time;';
+  if (isComplete) {
+    params.completed_number = total && !Number.isNaN(total) ? total : 1;
+    source += `ctx._source['status'] = "complete";
+               ctx._source['import_expected_number'] = params.completed_number;
+               ctx._source['completed_number'] = params.completed_number;
+               ctx._source['completed_time'] = params.processed_time;`;
+  }
   if (isNotEmptyField(message)) {
     if (inError) {
       source += 'ctx._source.errors.add(["timestamp": params.processed_time, "message": params.message]); ';
--- a/opencti-platform/opencti-graphql/tests/03-integration/02-resolvers/connector-test.ts
+++ b/opencti-platform/opencti-graphql/tests/03-integration/02-resolvers/connector-test.ts
@@ -242,7 +242,7 @@ describe('Connector resolver standard behaviour', () => {
     // Read work and verify status
     queryResult = await queryAsUserWithSuccess(USER_CONNECTOR, { query: READ_WORK_QUERY, variables: { id: workId } });
     expect(queryResult).not.toBeNull();
-    expect(queryResult.data.work.status).toEqual('progress');
+    expect(queryResult.data.work.status).toEqual('complete');
   });
   it('should delete work', async () => {
     // Delete the work
@@ -378,7 +378,7 @@ describe.todo('Connector sending multiple bundles during the same multi-part wor
 
       // Status should have changed to `complete`
       queryResult = await queryAsUserWithSuccess(USER_CONNECTOR, { query: READ_WORK_QUERY, variables: { id: workId } });
-      expect(queryResult.data.work.status).toEqual('progress');
+      expect(queryResult.data.work.status).toEqual('complete');
       expect(queryResult.data.work.completed_number).toEqual(5);
     });
   });


esrevi (approved):
tested, seems to solve the issue ! :)

fellowseb:
Let's re-discuss this in tomorrow's meeting. I'm a bit lost with the changes we made on works lately...

romain-filigran:
Hello. Can we have an update on this bug fix resolution ?

fellowseb (approved):

PR #15859, opened 2026-05-06
[backend] fix unsharing component (#15586)

### LINKED ISSUE #15586: "Unshare with organizations" playbook component removes previously granted_refs in the playbook run
# Description
The "Unshare with organizations" playbook component incorrectly overwrites any previously granted permission in the playbook. 

E.g. in a playbook that shares an entity with organisation A and then in a later component removes sharing permissions for organisation B the permissions granted to organisation A are overwritten. 

## Environment

7.260326.0

## Reproducible Steps

1. Configure a playbook that shares with an organisation [A]
2. Then subsequently add a component that "Unshare with organizations" playbook component.
3. Set it to unshare an entity with a specific organization [B].
4. Execute the playbook.
5. Observe the generated STIX bundle.
## Expected Output
The playbook component should share the playbook with organisation A and unshare the entity with organization B. 

## Actual Output
The playbook component removes the previously granted permission to organisation A. (From copilot: The generated STIX bundle contains a JSON Patch `remove` operation. This operation targets the entire `granted_refs` array [for the playbook execution])

[20260424_playbook_Issue 15586 STR Workflow.json](https://github.com/user-attachments/files/27047116/20260424_playbook_Issue.15586.STR.Workflow.json)

.

JeremyCloarec:
### Proposed changes

This pull request updates the logic for unsharing playbook components to ensure that only the specified organizations are removed from the `granted_refs` field, rather than attempting to remove the entire field or replace it with the full list. This change makes the unsharing process more precise and prevents unintended removal of access for other organizations.

Improvements to unsharing logic:

* In `opencti-graphql/src/modules/playbook/components/unsharing-component.ts`, the code now filters out only the specified organization IDs from the `granted_refs` array, and replaces the field with the updated list, improving the accuracy of the unsharing operation

### Related issues

* fix #15586 
*

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/modules/playbook/components/unsharing-component.ts
+++ b/opencti-platform/opencti-graphql/src/modules/playbook/components/unsharing-component.ts
@@ -55,15 +55,17 @@ export const PLAYBOOK_UNSHARING_COMPONENT: PlaybookComponent<UnsharingConfigurat
     for (let index = 0; index < bundle.objects.length; index += 1) {
       const element = bundle.objects[index];
       if (all || element.id === dataInstanceId) {
+        const currentElementGrantedRefs = element.extensions[STIX_EXT_OCTI].granted_refs ?? [];
+        const newGrantedRefs = currentElementGrantedRefs.filter((o) => !organizationIds.includes(o));
         const patchValue = {
-          op: EditOperation.Remove,
+          op: EditOperation.Replace,
           path: `/objects/${index}/extensions/${STIX_EXT_OCTI}/granted_refs`,
-          value: organizationIds,
+          value: newGrantedRefs,
         };
         const patchOperation = {
-          operation: patchValue.op,
+          operation: EditOperation.Remove,
           key: INPUT_GRANTED_REFS,
-          value: patchValue.value,
+          value: organizationIds,
         };
         applyOperationFieldPatch(element, [patchOperation]);
         patchOperations.push(patchValue);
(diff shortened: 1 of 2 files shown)


opencti-platform/opencti-graphql/src/modules/playbook/components/unsharing-component.ts
```diff
@@ -55,15 +55,17 @@ export const PLAYBOOK_UNSHARING_COMPONENT: PlaybookComponent<UnsharingConfigurat
...
         const patchValue = {
-          op: EditOperation.Remove,
+          op: EditOperation.Replace,
           path: `/objects/${index}/extensions/${STIX_EXT_OCTI}/granted_refs`,
-          value: organizationIds,
+          value: newGrantedRefs,
         };
         const patchOperation = {
-          operation: patchValue.op,
+          operation: EditOperation.Remove,
```
SarahBocognano: Just curious of why are you changing this with an enum rather than the patch value operation ?

JeremyCloarec: Because the patchValue now uses a Replace op and not a Remove op: we want the upsertOperation to execute a Remove on ingestion, but the jsonpatch.applyPatch below doesn't work properly with Remove operations (jsonpatch expects remove operations to give the index to remove in it's value, not the value to remove)

SarahBocognano (approved):

PR #15863, opened 2026-05-06
feat(backend): Allow lucene advanced search syntax (~ and *) (#12674)

### LINKED ISSUE #12674: feat(filters): improve wildcard prefix and fuzzy search option
## Use case

Current options wilcard prefix and fuzzy are in the conf and use globally in the platform.
This situation in large database lead to global performance retrogradation because options are apply to everything

## Current Workaround

Disable the options

## Proposed Solution

Remove options from the configuration and add it in the frontend UI to let users choose when to use it.

JeremyCloarec:
### Proposed changes

* Allow `~` and `*` syntax in search (not in form search, like for grouping name unicity check)
* add en empty state for search with link to documentation if textual search 
  * add a empty state without link to documentation if only filter search

### Related issues

* fix #12674

### Checklist

### Further comments

For validation, check that new syntax is properly handled in advanced search and in the search bar off dedicated pages (observables, etc.).
Ensure the advanced syntax is not handled in other search fields (filters, etc). Make sure to verify the "duplicate name" search (on the label when you create a new report for instance), it should not handle the new syntax to match only identical names


esrevi:
could we align vertically the icon with link text? :) 
Screenshot 2026-07-30 at 17 31 16

opencti-platform/opencti-front/src/components/dataGrid/components/DataTableSearchEmptyState.tsx
```diff
@@ -0,0 +1,60 @@
...
+type DataTableSearchEmptyStateProps = {
+  rawSearchTerm: string | undefined;
+};
+
+const DataTableSearchEmptyState = ({ rawSearchTerm }: DataTableSearchEmptyStateProps) => {
+  const { t_i18n } = useFormatter();
+  const theme = useTheme<Theme>();
+
+  return (
+    <>
```
esrevi: we can remove this <>

opencti-platform/opencti-front/src/components/dataGrid/components/DataTableSearchEmptyState.tsx
```diff
@@ -0,0 +1,60 @@
...
+          py: 6,
+          px: 2,
+          width: '100%',
+        }}
+      >
+        <DoNotDisturbOnOutlined sx={{ fontSize: 40, color: theme.palette.text.disabled, mb: 1 }} />
+        <Typography variant="h6" sx={{ color: theme.palette.text.secondary, mb: 2 }}>
+          {rawSearchTerm ? `${t_i18n('No results for')} "${rawSearchTerm}"` : t_i18n('No results')}
+        </Typography>
+        {!!rawSearchTerm
```
esrevi: I think we could remove !! since it's a string and we use above a ternary like that too 

```suggestion
        {rawSearchTerm
```

esrevi (dismissed):

opencti-platform/opencti-graphql/src/database/engine.ts
```diff
@@ -2187,12 +2182,9 @@ function processSearch(
 
   for (let searchIndex = 0; searchIndex < partialSearch.length; searchIndex += 1) {
     const partialElement = partialSearch[searchIndex];
-    const cleanElement = improvedFuzzyWildActivated ? specialElasticCharsEscapeWithFF(partialElement) : specialElasticCharsEscape(partialElement);
+    const cleanElement = specialElasticCharsEscapeWithFF(partialElement);
```
esrevi: rename without withFF ^^

opencti-platform/opencti-graphql/src/database/engine.ts
```diff
@@ -201,8 +201,6 @@ import { AbortError } from 'node-fetch';
 const ELK_ENGINE = 'elk';
 const OPENSEARCH_ENGINE = 'opensearch';
 export const ES_MAX_CONCURRENCY: number = conf.get('elasticsearch:max_concurrency');
-export const ES_DEFAULT_WILDCARD_PREFIX: boolean = booleanConf('elasticsearch:search_wildcard_prefix', false);
```
esrevi, line 204: do we want to remove permanently the two search types?

if yes, please remove in default.json
    "search_wildcard_prefix": false,
    "search_fuzzy": false,

opencti-platform/opencti-graphql/src/database/engine.ts
```diff
@@ -2158,7 +2159,7 @@ function processSearch(
   search: string,
   args: ProcessSearchArgs,
 ): { exactSearch: string[]; querySearch: string[] } {
-  const { useWildcardPrefix = ES_DEFAULT_WILDCARD_PREFIX } = args;
+  const { useWildcardPrefix } = args;
```
esrevi, line 2162: now, i m just confused, we removed wildcard prefix, but we still got this function name, is it intentional?

opencti-platform/opencti-graphql/src/database/engine.ts
```diff
@@ -1872,6 +1870,9 @@ const findElementsDuplicateIds = (elements: BasicStoreBase[]): string[] => {
 export const specialElasticCharsEscape = (query: string) => {
   return query.replace(/([/+|\-*()^~={}[\]:?!"\\])/g, '\\$1');
 };
+export const specialElasticCharsEscapeWithFF = (query: string) => {
```
Archidoit:
```suggestion
export const specialElasticCharsEscape = (query: string) => {
```

Archidoit:
The PR name and description are not aligned with the PR content anymore.

opencti-platform/opencti-front/src/components/dataGrid/components/DataTableBody.tsx
```diff
@@ -180,6 +184,8 @@ const DataTableBody = ({
     return null;
   }
 
+  const isSearching = !!tableSearchTerm || (filters?.filters ?? []).length > 0;
```
Archidoit: filters.filters.length can be > 0 whereas no filters is set. For instance if a default empty filters is displayed (exemple in Incidents list). In this case, we can have no Search term, no filters set, but still the 'empty state' display : 

Image

mcartaud: And it isn't the wanted behavior? I mean I hadn't though about this case, but seem fine to me. And we could elaborate whereas a default search is still a search. A probably better option though would be to have a default state everywhere if no data and to allow it to be overridden if emptyStateMessage is provided.
Would you prefer this?

opencti-platform/opencti-graphql/src/database/engine.ts
mcartaud: From what I understood, callers of the method can pass a `useWildcardPrefix` argument to this method. This is the case for `findAllMarkings` for instance. Removing the conf line 204 only remove a setting to apply a default value.
But I'll double check

opencti-platform/opencti-graphql/src/database/engine.ts
mcartaud: I believe so since they are accessible by typing `*` or `~` in the search bar now so no needs to activate it platform wide

opencti-platform/opencti-graphql/src/database/engine.ts
mcartaud: rename in specialElasticCharsEscapeWithLuceneSyntax since specialElasticCharsEscape is already defined above. Ok for you?

opencti-platform/opencti-front/src/components/dataGrid/components/DataTableBody.tsx
Archidoit: Don't know what product wants, but functionnally for me and for what i see in the code, we want 'no results' displayed when there are filters set or a search term entered. In this case, no filters are set (and no search term). There's just a filter key displayed as a shortcut.

opencti-platform/opencti-graphql/src/database/engine.ts
Archidoit: perfect !

Archidoit:
Apart from this little comment (not really blocking for me) : https://github.com/OpenCTI-Platform/opencti/pull/15863#discussion_r3689721257

Working well !

docs/docs/usage/search.md
```diff
@@ -27,9 +27,21 @@ Also, using the `Advanced search` button, it is possible to directly put filters
 
 !!! info "Advanced filters"
 
-    You have access to advanced filters all across the UI, if you want to know more about how to use these 
+    You have access to advanced filters all across the UI, if you want to know more about how to use these
     filters with the API or the Python library, [don't hesitate to read the dedicated page](../reference/filters.md)
 
+### Search syntax
+
+Keyword search (global search and contextual search) support small Lucene-like syntax to refine matching:
```
alice-debra: Keyword search (global and contextual) supports a subset of Lucene syntax to refine matching:

alice-debra (changes requested):

docs/docs/usage/search.md
alice-debra: Keyword search (global and contextual) supports a subset of Lucene syntax to refine matching:

mcartaud:
> Apart from this little comment (not really blocking for me) : [#15863 (comment)](https://github.com/OpenCTI-Platform/opencti/pull/15863#discussion_r3689721257)
> 
> Working well !

@Archidoit improved the filter to fix the case you mentioned and also add some documentation on syntax (reviewed by Alice)

opencti-platform/opencti-front/src/components/dataGrid/components/DataTableBody.tsx
```diff
@@ -180,6 +184,9 @@ const DataTableBody = ({
     return null;
   }
 
+  const isSearching = !!tableSearchTerm
+    || (filters?.filters ?? []).filter((filter) => filter.values.length > 0).length > 0;
```
Archidoit: not working for filters with the 'nil' and 'not_nil' operators... Those filters have filter.values = []

Archidoit:
```suggestion
    || (filters?.filters ?? []).filter((f) => ['nil', 'not_nil', 'has_changed', 'not_has_changed'].includes(f.operator ?? 'eq') || f.values.length > 0);
```

mcartaud: 🧠 Not easy to see, and to know what keys are available, a type for this operator param could be useful, but it is a bit beyond the scope :)

Archidoit: yes, would be better to create a constant FILTER_OPERATORS_WITH_NO_VALUES, but beyond the scope

Archidoit (approved):

Archidoit (approved):

alice-debra (approved):
All good

Issue #15644, opened 2026-04-22
fix(authentication): logout on session timeout does not record activity event history


scarletmerlin123:
## Description

The system fails to record logout events when the user is automatically logged out due to inactivity or clicking logout on the TimeoutLock modal. Logout events are only reported when the user actively logs out using the button in the top nav menu. 

## Environment

1. OS (where OpenCTI server runs): Mac OS 15
2. OpenCTI version:OpenCTI 7.260417.0
3. OpenCTI client: frontend
4. Other environment details: Chrome 147.0.7727.56

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Login to OpenCTI and allow the system to timeout
2. Check the Event history under dashboard/settings/activity/events to see if logout event was recorded

## Expected Output

Logout events triggered by the TimeoutLock modal should be recorded in activity.

## Actual Output

Logout events triggered by the TimeoutLock are not recorded.

nino-filigran:
@scarletmerlin123 you are using local auth, right?

scarletmerlin123:
Yes, this is with local auth.

nino-filigran:
Sorry I struggle to reproduce since yesterday: can you show me your configuration?

scarletmerlin123:
This is partially resolved by the PR #15748. The logout needs to happen during an active session to have it record the logout activity.

PR #15867, opened 2026-05-06
[frontend/backend] on session timeout, user is logout (#15252)

### LINKED ISSUE #15252: Logging out does not work when setting APP__SESSION_TIMEOUT
## Description

Logging out does not work when setting APP__SESSION_TIMEOUT

## Reproducible Steps

1. set  APP__SESSION_TIMEOUT = 120000 (2 minutes)
2. start platform and login

## Expected Output

logout after 2 min

## Actual Output

no logout

esrevi:
### Proposed changes

* screen recorded with session_timeout = 8000 and session_idle_timeout = 4000
https://github.com/user-attachments/assets/7829dc82-e163-4302-a4f3-ac5736706a3d

* without settings session_idle_timeout in the config, it is 0 by default, the user is logout directly
https://github.com/user-attachments/assets/679b0454-fa53-4851-9c47-0af6f6734e3c

* for session_timeout set under 2 mins, reduce the ttl toucheCache, otherwise, cap to 2 mins

### Related issues

* fixes #15252 , #15644 

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/database/sessionStore-redis.js
+++ b/opencti-platform/opencti-graphql/src/database/sessionStore-redis.js
@@ -16,7 +16,7 @@ class RedisStore extends Store {
     this.scanCount = Number(options.scanCount) || 100;
     this.serializer = options.serializer || JSON;
     // ttl < 2mins, divide by 10 to have a more frequent touch, otherwise touch every 2 mins
-    this.touchCache = new LRUCache({ ttl: Math.min(options.ttl / 10, 120000), max: 1000 });
+    this.touchCache = new LRUCache({ ttl: options.ttl < 120000 ? options.ttl / 10 : 120000, max: 1000 });
     this.locker = new AsyncLock();
   }
 
--- a/opencti-platform/opencti-graphql/src/database/sessionStore-redis.js
+++ b/opencti-platform/opencti-graphql/src/database/sessionStore-redis.js
@@ -1,4 +1,3 @@
-/* eslint-disable class-methods-use-this */
 import session from 'express-session';
 import { LRUCache } from 'lru-cache';
 import AsyncLock from 'async-lock';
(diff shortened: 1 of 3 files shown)


opencti-platform/opencti-graphql/src/database/sessionStore-redis.js
```diff
@@ -16,7 +15,8 @@ class RedisStore extends Store {
     this.prefix = options.prefix == null ? 'sess:' : options.prefix;
     this.scanCount = Number(options.scanCount) || 100;
     this.serializer = options.serializer || JSON;
-    this.touchCache = new LRUCache({ ttl: 120000, max: 1000 }); // Touch the session every 2 minutes
+    // ttl < 2mins, divide by 10 to have a more frequent touch, otherwise touch every 2 mins
+    this.touchCache = new LRUCache({ ttl: Math.min(options.ttl / 10, 120000), max: 1000 });
```
JeremyCloarec: if ttl is at 3 minutes, it will still keep options.ttl/10 (18 seconds) instead of keeping it at 2 minutes :)
I think you can add a specific condition that checks if ttl is < 2 minutes instead

esrevi: ah yes you are right, i ll make it simpler as you suggested :) thx !

JeremyCloarec (approved):

scarletmerlin123:
@esrevi I think there is still an issue where the time that the UI uses for the session expiration isn't the same as the session expiration recorded in redis. The epoch that the UI bases the time on is when the Timeoutlock modal is mounted, so it is after the actual session start time. The UI startDateEpoch could be offset so it ensures the session on the UI doesn't end after the actual session.

I tried looking at pulling the correct session expiration from the cookie but that also doesn't match what is in redis.  So I'll keep #15644 open while I make a PR to resolve the time inconsistency issues.

esrevi:
@scarletmerlin123  Thanks for the review! To clarify, the main goals of this PR were to fix the improper disconnection flow and ensure logout events triggered by the TimeoutLock modal are properly recorded in the activity log. 

For the startDateEpoch point, you are right that anchoring it to the moment TimeoutLock mounts creates a small drift compared to the actual Redis session expiry. This is a valid edge case but doesn't affect the core fixes, the user will still be logged out and the activity event will still be recorded, the countdown might just be slightly off imo. I suggest we track this case in a follow-up issue.

PR #15875, opened 2026-05-06
[frontend/backend] add origin filtering on live streams (#15874)

### LINKED ISSUE #15874: Add origin-based filtering on live streams
﻿## Use case

OpenCTI live streams currently filter events purely on the STIX payload (the `data` field of each Redis stream event). There is no way to scope a live stream to events produced by a specific user, group of users, or organization, even though every live event already carries this information in its `origin` envelope (`origin.user_id`, `origin.group_ids`, `origin.organization_ids`).

This is a missing capability for SOC integrations and platform-to-platform replication where consumers want to selectively forward only the events created by certain operators or teams.

## Current Workaround

None at the live stream level. The only workaround is to pre-filter on the consumer side after receiving every event, which wastes bandwidth and exposes data the consumer should not have to inspect.

## Proposed Solution

Add a new `origin_filters` capability on `StreamCollection` (separate from the existing STIX `filters`):

- A new optional `origin_filters` JSON field on `StreamCollection` (and on `StreamCollectionAddInput`).
- Strict server-side validation: only the three keys `members_user`, `members_group` and `members_organization` are allowed (reusing the existing activity-event member testers, which already read `event.origin.{user_id, group_ids, organization_ids}`).
- The SSE live handler runs the origin filter as a second gate, after the existing STIX filter check, around main-event publications. Recovery / dependency / container-fallback branches are intentionally not gated (they have synthetic origins).
- A new `Advanced options` accordion on the stream creation and edition forms (matching the style of the synchronizer `Advanced options` accordion), containing a dedicated filters widget restricted to the three origin keys, plus a clear warning that origin filtering does not apply during recovery.

### Known limitation (documented in the UI)

Recovery from Elasticsearch and synthetic dependency events have no real `origin`, so the origin filter is intentionally bypassed there. The user explicitly accepts this trade-off for the live use case.

## Additional Information

- Re-uses `testMembersUser/Group/Organization` from `filtering-activity-event/activity-event-testers`.
- Re-uses `MEMBERS_USER_FILTER` / `MEMBERS_GROUP_FILTER` / `MEMBERS_ORGANIZATION_FILTER` constants.
- Re-uses `isFilterGroupNotEmpty` and `testFilterGroup` helpers.
- Re-uses the `filterKeysSchema` registration for `History` for the value pickers.
- Re-uses the existing `Accordion` / `AccordionSummary` styled components (same as `SyncCreation`).
- A side bug is also fixed in `useSearchEntities`: `systemMembers` (system users) were leaking into Group / Organization filter pickers regardless of the requested `entityTypes`.

Yes

SamuelHassine:
### Proposed changes

* **Backend** - new optional `origin_filters` JSON field on `StreamCollection` (and on `StreamCollectionAddInput`), with strict server-side validation. The SSE live handler runs the origin filter as a second gate after `isStixMatchFilterGroup`, only around main-event publications. Recovery, container-fallback and dependency-publication branches are intentionally **not** gated (they carry synthetic origins).
* **Frontend** - new "Advanced options" accordion on the stream creation and edition forms (same custom `Accordion` / `AccordionSummary` styled components used by the synchronizer form), containing a dedicated `Filters` widget restricted to the three origin keys, plus a clear warning that origin filtering does not apply during recovery or for dependency events.
* **Side bug fix** - `useSearchEntities` was always merging `systemMembers` (system users) into `members_*` value pickers regardless of the requested `entityTypes`, so opening the Group or Organization filter showed system users. `systemMembers` are now only added when the filter targets `User`. This also fixes the same leak in the activity-events filter UI and any other consumer of the same code path.

#### Reuse maximised

* Re-uses `testMembersUser` / `testMembersGroup` / `testMembersOrganization` from `filtering-activity-event/activity-event-testers` (already read `event.origin.{user_id, group_ids, organization_ids}`).
* Re-uses `MEMBERS_USER_FILTER` / `MEMBERS_GROUP_FILTER` / `MEMBERS_ORGANIZATION_FILTER` constants and the `isFilterGroupNotEmpty` / `testFilterGroup` helpers.
* Re-uses the existing `filterKeysSchema` registration for `History` (`searchContext={{ entityTypes: ['History'] }}`) so the value pickers work without any new FE registry entry.
* Re-uses the existing custom `Accordion` / `AccordionSummary` (same as `SyncCreation`) for the styled "Advanced options" outline + chevron, and the existing `Filters` + `FilterIconButton` + `useFiltersState` trio (identical pattern to the existing main filter section).
* New BE module is a single ~46 line file (`filtering-stream-origin/stream-origin-filtering.ts`) containing the testers map + strict validator + matcher.

#### Known limitation (documented in the UI warning)

In recovery mode the synthetic events come from Elasticsearch with `origin: { referer: EVENT_TYPE_INIT }`, so origin filtering cannot apply there. Same for dependency events (`EVENT_TYPE_DEPENDENCIES`). The Advanced options panel shows an explicit warning about this trade-off.

### Related issues

* Fixes #15874

### Checklist

### Further comments

Design rationale for storing `origin_filters` as a separate JSON field rather than mixing the new keys into the existing `filters` field:

1. The existing `validateFilterGroupForStixMatch` registry is intentionally restricted to STIX-only keys; adding origin keys would have required loosening that contract everywhere `isStixMatchFilterGroup` is used (playbooks, triggers, etc.).
2. The two filters have different applicability windows: STIX filters apply in both live and recovery, origin filters only in live.
3. Two distinct UI sections clearly communicate the trade-off to the user.

The SSE gate is implemented by AND-ing the origin check inside each main-event `sendEvent` branch (rather than AND-ing into the outer `if/else if` chain conditions) so that an origin mismatch on the main event does **not** accidentally trigger the relation-fallback or container-fallback branches that would otherwise re-publish the event with its original `eventData` and defeat the filter.

Verification:
- 11/11 unit tests passing in `tests/01-unit/database/stream-origin-filtering-test.ts`
- `yarn check-ts` clean on both `opencti-graphql` and `opencti-front`
- `yarn lint` clean on both projects
- GraphQL types regenerated (`yarn build:types` + `yarn relay`)

### DIFF
--- a/client-python/pycti/connector/opencti_connector_helper.py
+++ b/client-python/pycti/connector/opencti_connector_helper.py
@@ -3254,7 +3254,7 @@ class OpenCTIConnectorHelper:  # pylint: disable=too-many-public-methods
         ):
             recover_iso_date = self.connect_live_stream_recover_iso_date
         # Generate the stream URL
-        url = url + "/stream"
+        url = url.rstrip("/") + "/stream"
         if live_stream_id is not None:
             url = url + "/" + live_stream_id
         self.listen_stream = ListenStream(
--- a/opencti-platform/opencti-front/lang/back/en.json
+++ b/opencti-platform/opencti-front/lang/back/en.json
@@ -676,6 +676,7 @@
   "Organization type": "Organization type",
   "Organizations": "Organizations",
   "Organizations management": "Organizations management",
+  "Origin filters": "Origin filters",
   "Origin of the case": "Origin of the case",
   "Original creation date": "Original creation date",
   "Other entities": "Other entities",
--- a/opencti-platform/opencti-front/lang/front/en.json
+++ b/opencti-platform/opencti-front/lang/front/en.json
@@ -3099,6 +3099,7 @@
   "Organizations restriction": "Organizations restriction",
   "Organizations sharing": "Organizations sharing",
   "Organizations splitter": "Organizations splitter",
+  "Origin filters only apply to live events. Recovery mode (when starting from a past date) and dependency events will ignore them.": "Origin filters only apply to live events. Recovery mode (when starting from a past date) and dependency events will ignore them.",
   "Origin of the case": "Origin of the case",
   "Original creation date": "Original creation date",
   "Originates from": "Originates from",
--- a/opencti-platform/opencti-front/src/private/components/data/stream/StreamCollectionCreation.tsx
+++ b/opencti-platform/opencti-front/src/private/components/data/stream/StreamCollectionCreation.tsx
@@ -7,10 +7,13 @@ import Alert from '@mui/material/Alert';
 import AlertTitle from '@mui/material/AlertTitle';
 import FormControlLabel from '@mui/material/FormControlLabel';
 import Switch from '@mui/material/Switch';
+import AccordionDetails from '@mui/material/AccordionDetails';
+import Typography from '@mui/material/Typography';
 import { FormikConfig } from 'formik/dist/types';
 import Button from '@common/button/Button';
 import Drawer, { DrawerControlledDialProps } from '@components/common/drawer/Drawer';
 import FormButtonContainer from '@common/form/FormButtonContainer';
+import { Accordion, AccordionSummary } from '../../../../components/Accordion';
 import { useFormatter } from '../../../../components/i18n';
 import TextField from '../../../../components/TextField';
 import ObjectMembersField from '../../common/form/ObjectMembersField';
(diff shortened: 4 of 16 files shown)


aHenryJard (approved):
Looks good to me. I have tested locally with:

analyst1: in Analyst group & Filigran Org
analyst2: in Analyst group and Filigran Org
anotherUser: in Analyst group and Org3 org

When creating a live stream with a filter on "Filigran" organization:

- When analyst1 do changes analyst2 can see them in stream but not anotherUser, admin with bypass can see changes too.

- When anotherUser do changes, analyst1 & analyst2 don't have it in stream.

The only comment that I have so far is UI, the organization is in lower case when everywhere else in the platform it start with an Upper case letter.

Image

ludovic (approved):

fellowseb:
One question I had was concerning the naming of the new `origin_filters` : I understand we didn't want to mix it with the STIX-dedicated `filters` field but I was wondering if it would have made sens to be more generic if we plan to add more non-STIX-related filters later on.

PR #15877, opened 2026-05-07
[backend] handle opinions when lower cased (#10828)

### LINKED ISSUE #10828: Update opinion in organization error
## Reproducible Steps

1. Organization
2. Give your opinion 
3. Try to change it 

Image

JeremyCloarec:
### Proposed changes

* frontend always sends opinions in lower case, even when vocab is modified to contain upper cases. Backend now handles lower cased opinions
*

### Related issues

* fix #10828
*

### Checklist


Archidoit:
working well

But in frontend, opinions defined in capital letters in taxonomies are always displayed in lower case in the object opinion edition form and in the object overview : 
image



image


image

JeremyCloarec:
> working well
> 
> But in frontend, opinions defined in capital letters in taxonomies are always displayed in lower case in the object opinion edition form and in the object overview : image
> image image

Yes, but I'm not sure if it is intended or not. To be defensive, I considered the display in lower case in the frontend to be expected and to not change the behavior, that's why I only pushed the fix on the backend side :)

Archidoit (approved):

Issue #6177, opened 2024-02-29
feat(case-management): custom fields and templates in Cases and Incidents


jmbodelon:
## Use case

In incident management it is essential to be able to create custom fields for each type of incident, allowing searches for any of them and also using them to create personalized dashboards.

## Current Workaround

The only workaround is to use the system's predefined fields or relationships with other entities or objects, and put those fields that do not fit in the description or as notes or labels.

## Proposed Solution

Use a similar strategy to TheHive with custom fields for alerts and cases, where you can add the fields you need to each template of a case type. Additionally, when promoting an alert to a case, custom fields are maintained.

Custom fields that can depend on others and be able to customize the design of the templates for each case would also be interesting.

## Additional Information

- **Custom fields in TheHive**: https://docs.strangebee.com/thehive/administration/custom-fields/

![image](https://github.com/OpenCTI-Platform/opencti/assets/1626951/f2f09072-712a-41bf-b329-e44e6b4f7581)

- **Case templates in TheHive**: https://docs.strangebee.com/thehive/user-guides/organisation/templates/case-templates/#new-case-template

![image](https://github.com/OpenCTI-Platform/opencti/assets/1626951/6ef72561-bd9e-4857-a42d-9c7093a847bb)

- **Conditional custom fields in Catalyst SOAR**: https://catalyst-soar.com/features/#conditional-custom-fields

![image](https://github.com/OpenCTI-Platform/opencti/assets/1626951/e1f8a688-2045-4d6a-bafc-c05a1660eada)

- **Custom fields in Cortex XSOAR**:
https://docs-cortex.paloaltonetworks.com/r/Cortex-XSOAR/8/Cortex-XSOAR-Administrator-Guide/Create-a-Custom-Incident-Field

![image](https://github.com/OpenCTI-Platform/opencti/assets/1626951/f1018a64-ec8a-432f-9dfe-d95ac22cbfc2)

- **Custom incident layouts in Cortex XSOAR**:
https://docs-cortex.paloaltonetworks.com/r/Cortex-XSOAR/8/Cortex-XSOAR-Administrator-Guide/Customize-Incident-Layouts

![image](https://github.com/OpenCTI-Platform/opencti/assets/1626951/4a53a293-2f25-46e2-b40a-14c105df8867)

## If the feature request is approved, would you be willing to submit a PR?

No, it is not a simple development.

nino-filigran:
[...]

PR #15884, opened 2026-05-07
feat(backend): custom field POC (nested) (#6177)

marieflorescontact:
### Proposed changes

* Create CustomField Module
* Add CustomField on Case-Incident (only int type for now, only for Case Incident)
*  Filters
* ONLY backend

### Related issues

* POC : https://github.com/OpenCTI-Platform/opencti/issues/6177

### Checklist


opencti-platform/opencti-graphql/src/modules/attributes/stixDomainObject-registrationAttributes.ts
```diff
@@ -352,3 +352,26 @@ const stixDomainObjectsAttributes: { [k: string]: Array<AttributeDefinition<any>
...
+    type: 'object',
+    format: 'nested',
+    mandatoryType: 'no',
+    editDefault: false,
+    multiple: true,
+    upsert: false,
+    isFilterable: false,
+    mappings: [
+      { name: 'field_id', label: 'Field ID', type: 'string', format: 'short', mandatoryType: 'internal', upsert: false, editDefault: false, multiple: false, isFilterable: false },
+      { name: 'field_name', label: 'Field name', type: 'string', format: 'short', mandatoryType: 'internal', upsert: false, editDefault: false, multiple: false, isFilterable: false },
```
aHenryJard, line 370: Either name or id is needed not both, we need to not duplicate that. Or move one to 'label' for UI display ?

PR #15887, opened 2026-05-07
feat(backend): add cleanup that logs users who are passively logged out due to session expiring (#15676)

### LINKED ISSUE #15676: feat(activity): record user logouts based on session expiration/termination
## Use case

The logout event should be recorded whenever all sessions are no longer valid for a user. Logout events are only recorded when a user manually triggers a logout and does not track when users are logged out due to inactivity or closing the browser.

## Current Workaround

None

## Proposed Solution

Periodically perform cleanup of the sessions in redis and record logouts for sessions that are expired.

Yes

scarletmerlin123:
### Proposed changes

* Add method for recording users who are logged out passively. Example: closing the tab or browser and not reopening OpenCTI before the session cookie expiration 

### Related issues

* Resolves #15676 

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/database/sessionStore-redis.js
+++ b/opencti-platform/opencti-graphql/src/database/sessionStore-redis.js
@@ -3,6 +3,7 @@ import { LRUCache } from 'lru-cache';
 import AsyncLock from 'async-lock';
 import { clearSessions, extendSession, getSession, getSessionKeys, getSessions, getSessionTtl, killSession, setSession } from './redis';
 import { logApp } from '../config/conf';
+import { publishUserAction } from '../listener/UserActionListener';
 
 const { Store } = session;
 
@@ -50,7 +51,8 @@ class RedisStore extends Store {
   touch(sid, sess, cb = noop) {
     const key = this.prefix + sid;
     const { touchCache } = this;
-    const sessionExtender = (done) => {
+    const sessionExtender = async (done) => {
+      await this._cleanUpSessions();
       const cachedTouch = touchCache.has(`touch-${key}`);
       if (cachedTouch) {
         return done(null, 'OK');
@@ -98,6 +100,22 @@ class RedisStore extends Store {
       return cb(null, keys);
     });
   }
+
+  _cleanUpSessions = async () => {
+    const sessions = await getSessions();
+    for (const session of sessions) {
+      if (session.redis_key_ttl <= 0) {
+        await this.destroy(session.redis_key_id);
+        await publishUserAction({
+          user: { ...session.user, origin: { socket: 'query', user_id: session.user.id } },
+          event_type: 'authentication',
+          event_access: 'administration',
+          event_scope: 'logout',
+          context_data: undefined,
+        });
+      }
+    }
+  };
 }
 
 export default RedisStore;


opencti-platform/opencti-graphql/src/database/sessionStore-redis.js
```diff
@@ -98,6 +100,22 @@ class RedisStore extends Store {
       return cb(null, keys);
     });
   }
+
+  _cleanUpSessions = async () => {
+    const sessions = await getSessions();
+    for (const session of sessions) {
+      if (session.redis_key_ttl <= 0) {
+        await this.destroy(session.redis_key_id);
```
OctaveLaventure, line 108: To double check but I think `this.destroy(session.redis_key_id)` won't work as `destroy()` adds a prefix to the key and the prefix is likely already included in getSessions()'s return value as we add it in the set() function in this file.

Also, is it really necessary as Redis already cleans up stale keys?

opencti-platform/opencti-graphql/src/database/sessionStore-redis.js
```diff
@@ -98,6 +100,22 @@ class RedisStore extends Store {
...
+        await publishUserAction({
+          user: { ...session.user, origin: { socket: 'query', user_id: session.user.id } },
```
OctaveLaventure, line 110: `user` is supposed to be an `AuthUser`, but in the session we only have the user's `id` and `otp_validated` keys, so you may want to use `getEntityFromCache` and get the full user

OctaveLaventure, line 110: Having a user is not guaranteed here, so you may want to add a line before such as `if (!session.user) continue;` to avoid errors

opencti-platform/opencti-graphql/src/database/sessionStore-redis.js
```diff
@@ -50,7 +51,8 @@ class RedisStore extends Store {
   touch(sid, sess, cb = noop) {
     const key = this.prefix + sid;
     const { touchCache } = this;
-    const sessionExtender = (done) => {
+    const sessionExtender = async (done) => {
+      await this._cleanUpSessions();
```
OctaveLaventure, line 55: Scanning all sessions on every user `touch` won't scale well with hundreds of users. A more scalable approach would be to do it periodically as a global clean up background task, as a subscription to redis updates or at least with a debounce mechanism

opencti-platform/opencti-graphql/src/database/sessionStore-redis.js
```diff
@@ -98,6 +100,22 @@ class RedisStore extends Store {
       return cb(null, keys);
     });
   }
+
+  _cleanUpSessions = async () => {
+    const sessions = await getSessions();
+    for (const session of sessions) {
+      if (session.redis_key_ttl <= 0) {
```
OctaveLaventure, line 107: If I understand the flow correctly, I think this will never be true as at that point the session data already has been cleared and getSessions won't return null data as it filters empty keys.

It will also include `-1` TTLs for keys without expiration dates, which is what is used for cookie-based sessions, so this condition should probably exclude `-1`.

As redis seems to clean up the keys automatically on expiration the output of `getSessions` will always have `redis_key_ttl >= 1` or `-1` (no expiration set). Basically if `getSessions` returns something it's likely not a session that needs to be cleaned up.

OctaveLaventure (commented):
Hello, thank you for your Pull Request!

With the current implementation, by the time you try to detect the expired session, the session data is gone, so this architecture won't work.

A potentially more idiomatic way would be to use shadow keys to keep the user's id after session expiration and periodically poll the sessions from the `platform_sessions` set, find the missing ones and read/delete the associated shadow keys data required for the audit log.

scarletmerlin123:
I'm going to look into alternatives for recording if a user passively logged out. Like @OctaveLaventure mentioned, there are issues with this approach and will still leave gaps.  Closing this PR.

PR #15890, opened 2026-05-07
feat(frontend): add offset to startDateEpoch so frontend session expires when backend session is still valid (#15889)

### LINKED ISSUE #15889: fix(authentication): session expiration time in frontend doesn't match expiration time in Redis
## Description

The frontend determines the current session expiration after mounting the TimeoutLock modal component. This causes a slight difference from the session expiration calculated from the session creation stored in Redis and causes the frontend to expire the session after the actual session is expired. 

## Environment

OS (where OpenCTI server runs): Mac OS 15
OpenCTI version:OpenCTI 7.260430.0
OpenCTI client: frontend
Other environment details: Chrome 147.0.7727.138

## Reproducible Steps

1. Compare local storage lockoutTracker startDateEpoch value and redis platforms_session value

## Expected Output
The frontend should not expire after the actual session expiration

## Actual Output

The frontend expires the session after the actual session is expired.

## Screenshots (optional)
Image

scarletmerlin123:
### Proposed changes

* Add 1 second offset to startDateEpoch so frontend session expires when backend session is still valid

### Related issues

* fixes #15889,  #15644
*

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/TimeoutLock.tsx
+++ b/opencti-platform/opencti-front/src/private/components/TimeoutLock.tsx
@@ -85,7 +85,7 @@ const TimeoutLock: React.FunctionComponent = () => {
     sessionLimit,
     idleCount: null,
     startDate: new Date(),
-    startDateEpoch: Date.now(),
+    startDateEpoch: Date.now() - 1000,
   });
   const [resetCounter, triggerReset] = useState(false);
   const interval = useRef<NodeJS.Timeout | null>(null);
--- a/opencti-platform/opencti-front/src/private/components/TimeoutLock.tsx
+++ b/opencti-platform/opencti-front/src/private/components/TimeoutLock.tsx
@@ -30,6 +30,13 @@ const timeoutLockRefreshQuery = graphql`
     }
   }
 `;
+const createTimeWithOffset = (): number => {
+  return Date.now() - ONE_SECOND;
+};
+
+const createOffsetTimeItem = (): { startDate: Date; startDateEpoch: number } => {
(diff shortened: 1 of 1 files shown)


opencti-platform/opencti-front/src/private/components/TimeoutLock.tsx
```diff
@@ -37,6 +37,7 @@ const timeoutLockRefreshQuery = graphql`
 const timeoutReducer = (state: TimeoutState, action: Action): TimeoutState => {
   const { idleLimit, sessionLimit } = state;
   let { idleCount, startDate, startDateEpoch } = state;
+  const offsetTime = Date.now() - ONE_SECOND;
```
ValentinBouzinFiligran: Hey @scarletmerlin123, great update — the fix looks solid now! 👍

One small idea to consider (totally optional, just a thought): since the { startDate, startDateEpoch } object with the offset is now created in two places, you could extract it into a small helper function to avoid the duplication:

```
const createTimeItem = () => ({
  startDate: new Date(),
  startDateEpoch: Date.now() - ONE_SECOND,
});

const newTimeItem = createTimeItem();
```

That way, if the offset ever needs to change, there's a single place to update it. It also makes the intent a bit clearer at the call sites.

That said, it's a minor suggestion — happy to go with the current approach if you prefer to keep it as is. What do you think?

scarletmerlin123: That's a good Idea. Made the revisions and pushed the update.

ValentinBouzinFiligran (approved):

PR #15901, opened 2026-05-08
[backend] check work aliveness only for background task works (#15886)

### LINKED ISSUE #15886: Bundle ingestion may be rejected due to closed work gate in backend
## Description

Closed work prevent futur related bundle to be ingested. This was initially done to stop ingesting works from canceled works, but in some case this reject messages that should be ingested.

* https://github.com/OpenCTI-Platform/opencti/pull/12451 changed the behavior, messages can be dropped when work is closed
* https://github.com/OpenCTI-Platform/opencti/pull/14456 move the gate from worker to backend, adding a log trace that make this behavior visible

xfournet:
### Proposed changes

Non background task works should always be considered alive

eg for connectors, it handles the race condition that happen on https://github.com/OpenCTI-Platform/opencti/blob/60b1fe80b57d7bafe3f4e524f3d8c05bd93ee4fe/opencti-platform/opencti-graphql/src/manager/connectorManager.js#L62 that delete the work that could be still referenced in queued elements (eg late or requeued elements)

### Related issues
* Fix #15886 

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/domain/work.js
+++ b/opencti-platform/opencti-graphql/src/domain/work.js
@@ -51,8 +51,12 @@ export const findById = (context, user, workId) => {
 };
 
 export const isWorkAlive = async (_context, _user, workId) => {
-  const redisWork = await redisGetWork(workId);
-  return redisWork?.is_initialized === 'true';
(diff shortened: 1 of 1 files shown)


xfournet:
Replaced by #15902

PR #15922, opened 2026-05-11
[backend/frontend] Workflow: autosave and publish (#15921)

### LINKED ISSUE #15921: feat(workflow): auto-save and publish
Auto-save workflow definition with publish mode

delemaf:
### Proposed changes

* Migrate from **workflow_content** to **published_version**, **draft_version**, **all_versions**
* Auto save
* Compile error

Before testing run:
```
POST /opencti_internal_objects-000001/_delete_by_query
{
  "query": {
    "bool": {
      "must": [
        { "term": { "target_type.keyword": "WorkflowDefinition" } }
      ]
    }
  }
}

POST /opencti_internal_objects-000001/_delete_by_query
{
  "query": {
    "bool": {
      "must": [
        { "term": { "entity_type.keyword": "WorkflowDefinition" } }
      ]
    }
  }
}
```
### Related issues

* #15921 

### Checklist


opencti-platform/opencti-graphql/src/modules/workflow/storage/workflow-definition-entity.ts
```diff
@@ -20,7 +58,42 @@ const WORKFLOW_DEFINITION_DEFINITION: ModuleDefinition<any, any> = {
...
+      label: 'Published Version',
+      type: 'object',
+      format: 'nested',
+      mandatoryType: 'no',
+      editDefault: false,
+      multiple: false,
+      upsert: true,
+      isFilterable: false,
+      mappings: versionMappings,
+    },
```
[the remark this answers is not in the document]
delemaf: Since we're on feature branch it's not relevent

opencti-platform/opencti-graphql/src/modules/workflow/workflow-validation.ts
```diff
@@ -108,38 +109,50 @@ export const validateWorkflowDefinitionData = async (
...
-) => {
+): Promise<WorkflowValidationError[]> => {
   let parsed;
+  const errors: WorkflowValidationError[] = [];
   try {
     parsed = JSON.parse(definitionStr);
   } catch (_error) {
-    throw ValidationError('Invalid workflow definition JSON');
+    errors.push({ type: 'INVALID_JSON', message: 'Invalid workflow definition JSON' });
   }
```
OctaveLaventure, line 119: Don't we want to return early if the JSON is invalid ? the `safeParse()` call that follows will return null anyway

opencti-platform/opencti-graphql/src/modules/workflow/domain/workflow-domain.ts
```diff
@@ -125,34 +231,91 @@ export const setWorkflowDefinition = async (
...
     if (existingWorkflow) {
+      // Add to version history (prepend new version to maintain chronological order)
+      const allVersions = existingWorkflow.all_versions || [];
+      const updatedVersions = [versionData, ...allVersions];
+
+      // draft_version is always in all_versions
       await updateAttribute(executionContext, executionUser, existingWorkflow.id, ENTITY_TYPE_WORKFLOW_DEFINITION, [
-        { key: 'workflow_content', value: [definition] },
+        { key: 'draft_version', value: [versionData] },
+        { key: 'all_versions', value: updatedVersions },
```
OctaveLaventure, line 265: `all_versions` grows indefinitely, I think we should only keep the last X versions just in case.

Even though workflows shouldn't change that often, we're not protected from a bug that would save in an infinite loop, which is quite possible as the save is triggered through hooks reacting to events on the react-flow part. For example if we allowed dragging react-flow elements. A debounce on the front could help too

opencti-platform/opencti-graphql/src/modules/workflow/domain/workflow-domain.ts
```diff
@@ -162,18 +325,98 @@ export const deleteWorkflowDefinition = async (
...
+      draftVersionId: draftVersion.id,
+    });
+  }
+
+  // CONSISTENCY GUARANTEE: published_version will be in all_versions (already there via draft)
+  // Copy draft_version to published_version and clear the draft (no more unpublished changes).
+  const updates: any[] = [
+    { key: 'published_version', value: [draftVersion] },
+    { key: 'draft_version', value: [] },
+  ];
```
OctaveLaventure: The type can probably be more strict than `any[]` here

opencti-platform/opencti-graphql/src/modules/workflow/domain/workflow-domain.ts
```diff
@@ -162,18 +325,98 @@ export const deleteWorkflowDefinition = async (
...
+  if (draftVersion.validation_errors && draftVersion.validation_errors.length > 0) {
+    throw FunctionalError('Cannot publish workflow with validation errors', {
+      entityType,
+      errorCount: draftVersion.validation_errors.length,
+    });
+  }
+
+  // Validate consistency BEFORE publishing
+  const allVersions = workflowDefinitionEntity.all_versions || [];
+  const draftInHistory = allVersions.some((v: any) => v.id === draftVersion.id);
```
OctaveLaventure: Is the `any` necessary ? I think it could be `WorkflowVersion`

OctaveLaventure: We should avoid single char variable names to improve readability (personal opinion 😁)

opencti-platform/opencti-graphql/src/modules/workflow/domain/workflow-domain.ts
```diff
@@ -260,9 +503,9 @@ export const getAllowedNextStatuses = async (
   context: AuthContext,
   user: AuthUser,
   entityId: string,
-): Promise<any[]> => {
+): Promise<unknown[]> => {
```
OctaveLaventure: Maybe `string[]` would work ?

opencti-platform/opencti-graphql/src/modules/workflow/domain/workflow-domain.ts
delemaf: I'll limit all_versions size to 100 entries and slice the oldest

OctaveLaventure (approved):
🚀

PR #15931, opened 2026-05-11
fix(decayrules): default rule recalculates incorrectly on indicator update (#13629)

### LINKED ISSUE #13629: fix(decay-rules): default rule recalculates incorrectly on indicator update
## Description

The built-in "Default" decay rule is producing inconsistent results after updating an Indicator. This one : 

Image

When an Indicator uses the default rule, any update of the score field triggers an incorrect recalculation of the rule, leading to an unexpected valid_until value.

This results in a mismatch between the expected behavior of the default decay rule and its actual outcome.

## Environment

OCTI 6.8.13

## Reproducible Steps

1. Create an Indicator with a score of 100 and using a main observable type that falls under the default decay rule. Example: main observable type Text.

Image 

Result:
- valid_from is set to today.
- valid_until is set to approximately one year from now → expected behavior.

Image

2. Now click on Update (top right corner).
3. Click inside the score field in the update form (just click inside like if you wanted to modify the score, but do nothing).
4. Click outside the form to close it.
5. Observe the result:
- The decay rule recalculates.
- valid_until is now few month later for no reason.

Image

## Expected Output

valid_until should follow the rule’s expected validity window (e.g., 1 year for a score of 100).

ValentinBouzinFiligran:
### Proposed changes

* compare in backend number with number
* mutation not triggered when score does not change

### Related issues

* Fixes #13629 

### Checklist


marieflorescontact (approved):
locally tested ok

PR #15970, opened 2026-05-12
[frontend] Fix input searching in playbook component (#15963)

### LINKED ISSUE #15963: label picker in playbook manipulate knowledge not working
## Description

Using the label selection dropdown a user can not type on initial load. The field is wiped and then reset. Once a user selects one label the user can then type correctly. 

## Environment

2. OpenCTI version: 7.260510.0
latest-octi-staging
demo instnace

## Reproducible Steps

1. Open a manipulate knowledge component in playbooks
2. click to add an action
3. add, label, then try to type a label 

## Expected Output

characters inputted into the label selection are shown and the dropdown is filtered corectly
## Actual Output

see video, characters entered are not shown and the list is not filtered. 
## Additional information

https://github.com/user-attachments/assets/6f26df23-9355-4610-9891-5d9080e3856d

This may also be related to this issue https://github.com/OpenCTI-Platform/opencti/issues/15927

Archidoit:
### Proposed changes
In the 'Manipulate knowledge' component of playbook : 
- fix input searching (when entering an input in a search field, the keyword should be written)
- align the 'action', 'field' and 'value' inputs
- the 'manipulate all elements included in the bundle' should switch if we click on its label only (not on the entire line after the label)

image

### Related issues
fixes #15963

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/data/playbooks/playbookFlow/playbookFlowFields/playbookFlowFieldsActions/PlaybookFlowFieldActions.tsx
+++ b/opencti-platform/opencti-front/src/private/components/data/playbooks/playbookFlow/playbookFlowFields/playbookFlowFieldsActions/PlaybookFlowFieldActions.tsx
@@ -53,6 +53,54 @@ const PlaybookFlowFieldActions = ({
     return a.op && a.attribute && a.value && a.value.length > 0;
   });
 
+  /**
+   * Returns the label for a given option.
+   * Handles both string values and FieldOption objects.
+   */
+  const getOptionLabel = (options: FieldOption[]) => (option: string | FieldOption): string => {
+    if (typeof option === 'string') {
+      return options.find((o) => o.value === option)?.label ?? option;
+    }
+    return option.label ?? '';
+  };
+
+  /**
+   * Reset the action when the operation changes.
+   * Clears attribute, value, and form display value.
+   */
+  const handleOperationChange = (index: number, val: FieldOption | null) => {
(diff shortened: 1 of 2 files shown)


opencti-platform/opencti-front/src/private/components/data/playbooks/playbookFlow/playbookFlowFields/playbookFlowFieldsActions/PlaybookFlowFieldActions.tsx
fellowseb, line 37: This default value looks weird. There's isn't a matching `t_i18n(capitalizeFirstLetter(op))` entry in the i18n dictionaries for this value. My guess is that it should be `['add', 'replace', 'remove']` or removed because unused.

fellowseb (approved):

opencti-platform/opencti-front/src/private/components/data/playbooks/playbookFlow/playbookFlowFields/playbookFlowFieldsActions/PlaybookFlowFieldActions.tsx
Archidoit: it enables to display 'Replace' instead of 'replace' in the form

Archidoit: and the values are in en.json

PR #15978, opened 2026-05-13
[backend] Add a liveness server with a dedicated port (#15977)

### LINKED ISSUE #15977: Add a liveness probe on OpenCTI
## Use case

The '/health' which is an readiness API may not be sufficient on some kubernetes configuration.

The feature request is to add an HTTP service on a new port that answer HTTP 200 OK as soon as the platform is alive.

- The port must be configuration in environment variable

https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/

aHenryJard:
### Proposed changes

This PR introduces a lightweight liveness HTTP server on a dedicated, configurable port so orchestrators (e.g., Kubernetes) can verify that the OpenCTI process is alive independently from the readiness /health endpoint. 

* Add a configurable liveness port, disabled by default

#### Test refactor
* Group all integration test files for the "http" module inside a module
* add an http utility for test and reuse it in taxii tests

### Related issues

* closes #15977 

### Doc
127 0 0 1_8000_deployment_liveness-probe__h=liveness


opencti-platform/opencti-graphql/tests/utils/httpUtils.ts
fellowseb, line 1: Rename to `httpUtils-test.ts` maybe ?

opencti-platform/opencti-graphql/src/http/httpLiveness.ts
fellowseb, line 1: Just my 2 cents but I think we could move this in a `health` (or platformHealth`) module instead of again adding stuff to the `/http` folder that should strictly be about HTTP pipeline stuff IMO.

docs/docs/deployment/liveness-probe.md
fellowseb, line 1: Love the docs ❤️

fellowseb (approved):

opencti-platform/opencti-graphql/tests/utils/httpUtils.ts
aHenryJard: it's not a test, it's a utility for tests. I try to start using fetch (instead of axios) for tests so I group them there.

PR #15990, opened 2026-05-13
[frontend] nested rel available with tech and functional date (#10110)

### LINKED ISSUE #10110: Can't display timeline in functional date with nested relationships
## Description
Can't display timeline with functional date and nested relationships

## Reproducible Steps
- Go to an entity , in the Knowledge tab
- At the bottom, go under the 'timeline' tab
- Click on the settings icon to select the date
- If you select 'functional date', you can't click on 'include nested relationships' (the date is automatically reset to 'technical date' if I click on the 'include nested relationships' toggle)

Image

ValentinBouzinFiligran:
### Proposed changes

* nested rel available with tech and functional date

### Related issues

* Fixes #10110 

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/common/stix_domain_objects/StixDomainObjectThreatKnowledge.tsx
+++ b/opencti-platform/opencti-front/src/private/components/common/stix_domain_objects/StixDomainObjectThreatKnowledge.tsx
@@ -201,14 +201,10 @@ const StixDomainObjectThreatKnowledge: FunctionComponent<
 
   const handleChangeTimeField = (event: SelectChangeEvent) => {
     setTimeField(event.target.value);
-    setNestedRelationships(
-      event.target.value === 'functional' ? false : nestedRelationships,
-    );
   };
 
   const handleChangeNestedRelationships = (event: React.ChangeEvent<HTMLInputElement>) => {
     setNestedRelationships(event.target.checked);
-    setTimeField(event.target.checked ? 'technical' : timeField);
   };
 
   const handleOpenTimeField = (event: React.MouseEvent) => {


opencti-platform/opencti-front/src/private/components/common/stix_domain_objects/StixDomainObjectThreatKnowledge.tsx
```diff
@@ -201,14 +201,10 @@ const StixDomainObjectThreatKnowledge: FunctionComponent<
 
   const handleChangeTimeField = (event: SelectChangeEvent) => {
     setTimeField(event.target.value);
-    setNestedRelationships(
-      event.target.value === 'functional' ? false : nestedRelationships,
-    );
```
SarahBocognano, line 206: It's working fine but do we know why this code was added in the first place ? Because it feels weird to just remove it

ValentinBouzinFiligran: This code was originally added as a defensive guard, likely under the assumption that nested/meta-relationships ("StixRefRelationship") didn't support a functional date ("start_time").

After digging into the code:
- On the backend, sorting by "start_time" already handles missing values gracefully via "missing: 0" (epoch) in Elasticsearch — no error, documents just fall to the bottom of the list.
- On the frontend, "StixDomainObjectTimeline" already has a fallback in "getDate()" that cascades from "start_time" → "first_seen" → "created_at" when the field is absent.

Both fallbacks were already in place before this PR, making the mutual-lock logic unnecessary  and actually harmful, since it was blocking a perfectly valid combination. Safe to remove.

SarahBocognano (approved):

PR #16018, opened 2026-05-15
[backend] Add automatic cleanup of old news feed items (#15643)

### LINKED ISSUE #15643: [XTMHub] Cleanup old newsfeed (older than 6 months)
This feature introduces the ability to remove items that are no longer relevant.

This ensures the newsfeed remains accurate, clean, and meaningful for end users.

Related to  https://github.com/FiligranHQ/xtm-hub/issues/2197

Antoine-Quintin:
### Proposed changes

* New `cleanOldNewsFeedItems(context, user, cutoffDate)` function in `news-feed-domain.ts` that deletes locally cached `NewsFeedItem` entities older than the cutoff. Uses `elPaginate` + `promiseMap` with concurrency 5, aligned with the pattern used in `loadAndSaveLatestNewsFeed` and #15960.
* Cleanup invocation added in `hubRegistrationManager`, called after `loadAndSaveLatestNewsFeed` at each tick (every hour), wrapped in its own try/catch so a cleanup failure doesn't break the manager.
* New configuration keys `hub_registration_manager:news_feed_cleanup_interval_value` (default `180`) and `hub_registration_manager:news_feed_cleanup_interval_unit` (default `"days"`). Pattern reuses the `value + unit` approach from the existing `retentionManager` for flexibility (days / hours / minutes / seconds — adjustable without code change).
* Continues the cleanup of free-string filters started in #15960 by using `FilterMode.And` and `FilterOperator.Lt` enums explicitly.
* 4 new unit tests in `newsFeed-test.ts` covering filter construction, batch deletion, empty case, and resilience to `ALREADY_DELETED_ERROR`.

### Related issues

* Closes #15643
* Related to FiligranHQ/xtm-hub#2197
* Companion PR on XTM Hub: FiligranHQ/xtm-hub#2325

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/manager/hubRegistrationManager.ts
+++ b/opencti-platform/opencti-graphql/src/manager/hubRegistrationManager.ts
@@ -1,12 +1,16 @@
 import { type ManagerDefinition, registerManager } from './managerModule';
-import conf, { booleanConf } from '../config/conf';
+import conf, { booleanConf, logApp } from '../config/conf';
 import { executionContext, HUB_REGISTRATION_MANAGER_USER } from '../utils/access';
 import { checkXTMHubConnectivity, loadAndSaveLatestNewsFeed } from '../domain/xtm-hub';
 import { XtmHubRegistrationStatus } from '../generated/graphql';
+import { cleanOldNewsFeedItems } from '../modules/xtm/hub/news-feed/news-feed-domain';
+import moment from 'moment';
 
 const HUB_REGISTRATION_MANAGER_ENABLED = booleanConf('hub_registration_manager:enabled', true);
 const HUB_REGISTRATION_MANAGER_KEY = conf.get('hub_registration_manager:lock_key') || 'hub_registration_manager_lock';
 const SCHEDULE_TIME = conf.get('hub_registration_manager:interval') || 60 * 60 * 1000; // 1 hour
+const NEWS_FEED_CLEANUP_INTERVAL_VALUE = conf.get('hub_registration_manager:news_feed_cleanup_interval_value') || 180;
+const NEWS_FEED_CLEANUP_INTERVAL_UNIT = conf.get('hub_registration_manager:news_feed_cleanup_interval_unit') || 'days';
 
 /**
  * If platform is registered, calls XTM Hub backend to check if the registration data is still valid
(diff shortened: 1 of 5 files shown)


opencti-platform/opencti-graphql/src/manager/hubRegistrationManager.ts
```diff
@@ -18,6 +22,27 @@ export const hubRegistrationManager = async () => {
   if (status === XtmHubRegistrationStatus.Registered) {
     await loadAndSaveLatestNewsFeed(context, HUB_REGISTRATION_MANAGER_USER);
   }
+
+  try {
+    const cutoffDate = moment()
+      .subtract(NEWS_FEED_CLEANUP_INTERVAL_VALUE, NEWS_FEED_CLEANUP_INTERVAL_UNIT)
```
ludovic: `moment().subtract(value, unit)` silently no-ops when the unit string is unrecognized, it returns `moment()` unchanged (i.e. _now_). This means a typo in the config (e.g. `"week"` instead of `"weeks"`, `"day"` instead of `"days"`) would set `cutoffDate = now`, causing the `creation_date < now` filter to match **every** `NewsFeedItem` ever stored and wipe the table completely.

Consider adding a startup guard:
```typescript
const VALID_UNITS = ['seconds', 'minutes', 'hours', 'days', 'weeks', 'months', 'years']; 
if (!VALID_UNITS.includes(NEWS_FEED_CLEANUP_INTERVAL_UNIT)) {
  throw new Error(`[XTMH] Invalid news_feed_cleanup_interval_unit: "${NEWS_FEED_CLEANUP_INTERVAL_UNIT}"`);
 }
 ```

File: opencti-platform/opencti-graphql/tests/01-unit/domain/newsFeed-test.ts [hunk 2]
```diff
@@ -196,4 +200,90 @@ describe('News feed', () => {
...
+
+    it('should swallow ALREADY_DELETED_ERROR without breaking the loop', async () => {
+      mockElPaginate.mockResolvedValueOnce({
+        edges: [
+          { node: { internal_id: 'item-1' } },
+          { node: { internal_id: 'item-2' } },
+        ],
+      });
+      const alreadyDeletedError = Object.assign(new Error('Already deleted'), {
+        extensions: { code: 'ALREADY_DELETED' },
```
ludovic:
```suggestion
        extensions: { code: 'ALREADY_DELETED_ERROR' },
```

opencti-platform/opencti-graphql/src/manager/hubRegistrationManager.ts
Antoine-Quintin: Good catch — that silent no-op would have been disastrous. Added a startup guard at module-load time that validates both news_feed_cleanup_interval_value (positive number) and news_feed_cleanup_interval_unit (one of the units moment().subtract actually understands). The manager now throws at boot if the config is invalid, before any cleanup ever runs.

opencti-platform/opencti-graphql/tests/01-unit/domain/newsFeed-test.ts
Antoine-Quintin: Done — using the exported ALREADY_DELETED_ERROR constant in the test. Side note: the actual constant value is 'ALREADY_DELETED_ERROR', not 'ALREADY_DELETED', so the previous test was technically passing for the wrong reason (the if branch never matched). The fix makes it test what it's supposed to test.

ludovic (approved):

PR #16021, opened 2026-05-15
[backend] Fix it test flake on sharing and on e2e (#15955)

### LINKED ISSUE #15955: Fix flake integration test "should Editor user access all objects"
```
 FAIL  tests/03-integration/02-resolvers/organization-sharing-test.ts > oganization-sharing-test > Organization sharing standard behavior for container > should Editor user access all objects
Error: Condition not met after 10 attempts - Please check that you have a test worker running
 ❯ awaitUntilCondition tests/utils/testQueryHelper.ts:307:11
    305| 
    306|   if (!isConditionOk === expectToBeTrue) {
    307|     throw new Error(`Condition not met after ${loopCount} attempts - $…

       |           ^
    308|   }
    309| };
 ❯ tests/03-integration/02-resolvers/organization-sharing-test.ts:128:7
```

https://github.com/OpenCTI-Platform/opencti/actions/runs/25716587365/job/75508205010?pr=15696

aHenryJard:
### Proposed changes

- Default is 10s and in the test it loop until 10s max so on some case it's not enough at 1s. Instead of waiting more I think it's possible to increase task manager frequency for tests.

- default session timeout is 20min, I think in some case (with some retries) it's not enough for the e2e tests.

https://github.com/OpenCTI-Platform/opencti/blob/71b54d97ac4d31f5494eec20380bfb939aff9793/opencti-platform/opencti-graphql/config/default.json#L260-L263

And

https://github.com/OpenCTI-Platform/opencti/blob/71b54d97ac4d31f5494eec20380bfb939aff9793/opencti-platform/opencti-graphql/tests/03-integration/02-resolvers/organization-sharing-test.ts#L128

### Related issues

* closes #15955 

### Checklist

### DIFF
--- a/.github/workflows/ci-test-frontend.yml
+++ b/.github/workflows/ci-test-frontend.yml
@@ -99,6 +99,7 @@ jobs:
           -e APP__CHILD_LOCKING_PROCESS__ENABLED=true \
           -e APP__ADMIN__TOKEN=bfa014e0-e02e-4aa6-a42b-603b19dcf159 \
           -e APP__APP_LOG__EXTENDED_ERROR_MESSAGE=true \
+          -e APP__SESSION_TIMEOUT=3600000 \
           -e APP__HEALTH_ACCESS_KEY=cihealthkey \
           -e REDIS__HOSTNAME=redis \
           -e REDIS__NAMESPACE=e2e-start \
--- a/.github/workflows/ci-test-frontend.yml
+++ b/.github/workflows/ci-test-frontend.yml
@@ -162,6 +162,7 @@ jobs:
           -e BACK_END_URL=http://opencti-e2e-start:4100 \
           -e E2E_TEST=true \
           -e TEAMS_WEBHOOK="teams-webhook-url" \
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-test-frontend.yml
```diff
@@ -99,6 +99,7 @@ jobs:
           -e APP__CHILD_LOCKING_PROCESS__ENABLED=true \
           -e APP__ADMIN__TOKEN=bfa014e0-e02e-4aa6-a42b-603b19dcf159 \
           -e APP__APP_LOG__EXTENDED_ERROR_MESSAGE=true \
+          -e APP__SESSION_TIMEOUT=3600000 \
```
fellowseb, line 102: From my recent experience this timer resets when there's activity; maybe it shouldn't I pinged Product about it

fellowseb (approved):

.github/workflows/ci-test-frontend.yml
aHenryJard: ok there is other fixes so I merge and then it will be better for e2e once this question is solved.

PR #16047, opened 2026-05-18
[backend/frontend] feat(xtmhub): Add a banner when platform is not registered (#16046)

### LINKED ISSUE #2139: 🎯 Display toast to user with capability when platform is not registered

### LINKED ISSUE #16046: [xtmhub] Add a banner when platform is not registered
## Description

Platform registration is a prerequisite for successful news feed fetch since we need the platform id to know which news feed items to deliver.
The goal of this chunk is to display a banner for users having manage xtm hub settings capability on OpenCTI to invite them to register the hub.

Linked to https://github.com/FiligranHQ/xtm-hub/issues/2139

carinelebas:
### Proposed changes
We want to add a banner when platform is not registered on xtmhub, and registration is possible for the user. The banner must be dismissible.
This implies some changes on TopBanner component:
- make it dismissible, using a logic similar to settingsMessageBanner
- move it from root component to index to be able to use hooks like useGranted and useNavigate, which requires to compute the banner height and pass it to the required components
- create a dedicated component to manage the 3 possible topBanners

### Related issues

* closes https://github.com/OpenCTI-Platform/opencti/issues/16046
* closes https://github.com/FiligranHQ/xtm-hub/issues/2139

### Checklist


opencti-platform/opencti-front/src/private/components/TopBannersManager.tsx
```diff
@@ -0,0 +1,37 @@
...
+  useBus(REGISTER_BANNER_DISMISSED_BUS, setIsDismissed, []);
+
+  const showLicenseBanner = shouldDisplayLicenseBanner(settings?.platform_enterprise_edition);
+  const showTrialBanner = shouldDisplayTrialBanner(settings);
+  const showRegisterBanner = isFeatureEnable('XTMHUB_NEWS_FEED_ENABLED') && shouldDisplayRegisterBanner(
+    settings,
+    isXTMHubAccessible,
+    isGrantedToXtmHub,
+    isDismissed,
+  );
```
CelineSebe: This code is duplicated in useTopBannerHeight. 

Suggestion: Create a single hook called `useTopBanner()` that returns both the visibility flags and the height, and use it in both places.

opencti-platform/opencti-front/src/private/components/xtm_hub/RegisterPlatformBanner.tsx
```diff
@@ -0,0 +1,40 @@
...
+  const navigate = useNavigate();
+  const { t_i18n } = useFormatter();
+  const { settings } = useContext(UserContext);
+
+  // Auto-reset dismissed state once the platform becomes registered
+  useEffect(() => {
+    if (settings?.xtm_hub_registration_status === 'registered') {
+      resetRegisterBannerDismiss();
+    }
+  }, [settings?.xtm_hub_registration_status]);
```
CelineSebe: This component will never be rendered if xtm_hub_registration_status === ‘registered’ (the parent TopBannersManager filters using shouldDisplayRegisterBanner). Therefore, this useEffect will never be triggered with the value ‘registered’.
Either move it to TopBannersManager or remove it.

opencti-platform/opencti-front/src/private/components/TimeoutLock.tsx
```diff
@@ -235,8 +237,8 @@ const TimeoutLock: React.FunctionComponent = () => {
       disableEscapeKeyDown={true}
       sx={{
         backdropFilter: 'blur(15px)',
-        marginTop: `${bannerHeightNumber}px`,
-        height: `calc(100% - ${bannerHeightNumber * 2}px)`,
+        marginTop: `${topBannerHeight + bannerHeightNumber}px`,
+        height: `calc(100% - ${topBannerHeight + bannerHeightNumber * 2}px)`,
```
CelineSebe, line 241: If the intention is to double both values (as was previously the case only for `bannerHeightNumber`),  parentheses should be added:

height: `calc(100% - ${(topBannerHeight + bannerHeightNumber) * 2}px)`,

opencti-platform/opencti-front/tests_e2e/auth.setup.ts
```diff
@@ -12,6 +12,12 @@ setup('authenticate as admin(at)opencti.io by default', async ({ page }) => {
   await page.goto('/');
   await loginPage.login();
   await expect(dashboardPage.getPage()).toBeVisible();
+
+  // Dismiss the RegisterPlatformBanner so its fixed-position button does not intercept clicks in any test
+  await page.evaluate(() => {
+    localStorage.setItem('register-banner-dismissed', 'true');
```
CelineSebe: You could import 'register-banner-dismissed' from bannerUtils.ts, if the key change the test will no longer work

opencti-platform/opencti-front/src/private/components/TopBannersManager.tsx
carinelebas: good idea ! done in [a139c03](https://github.com/OpenCTI-Platform/opencti/pull/16047/commits/a139c038c559771c9890f06e43710356794d97fe)

opencti-platform/opencti-front/src/private/components/xtm_hub/RegisterPlatformBanner.tsx
carinelebas: good point! useEffect is moved to TopBannersManager: [a139c03](https://github.com/OpenCTI-Platform/opencti/pull/16047/commits/a139c038c559771c9890f06e43710356794d97fe)

opencti-platform/opencti-front/tests_e2e/auth.setup.ts
carinelebas: good idea, done in [a139c03](https://github.com/OpenCTI-Platform/opencti/pull/16047/commits/a139c038c559771c9890f06e43710356794d97fe)

opencti-platform/opencti-front/src/private/components/TimeoutLock.tsx
carinelebas: Actually, we don't want to double topBannerHeight. BannerHeightNumber is doubled in this case because the banner is displayed both at the top and the bottom of the screen, which is not the case for the topBanner.

CelineSebe (approved):

PR #16052, opened 2026-05-18
fix(backend): fix getTracer undefined error when importing CSV with CSV Mapper (#16051)

### LINKED ISSUE #16051: fix(csv-mapper): "Cannot read properties of undefined (reading 'getTracer')" when importing CSV with CSV Mapper
## Description

When importing a CSV file using the CSV Import connector, the process crashes with the following error:

TypeError: Cannot read properties of undefined (reading 'getTracer')
at telemetry (tracing.ts:123)
at getEntitiesFromCache (cache.ts:167)
at getEntitiesListFromCache (cache.ts:177)
at getEntitySettingFromCache (entitySetting-utils.ts:137)
at validateCsvMapper (csvmapper-utils.ts:130)
at processCsvForWorkers (importcsv-connector.ts:74)

## Environment

4. Other environment details:

## Reproducible Steps

1.Create a CSV Mapper with at least one entity representation
2.Trigger a CSV import via the Import CSV connector
3.Observe the crash during the mapper validation phase

## Expected Output

CSV import processes normally (or fails with a meaningful validation error).

## Actual Output

Connector crashes with TypeError before any data is processed.

SarahBocognano:
### Proposed changes

* check context.tracing

### Related issues

* Fix: https://github.com/OpenCTI-Platform/opencti/issues/16051

### DIFF
--- a/opencti-platform/opencti-graphql/src/config/tracing.ts
+++ b/opencti-platform/opencti-graphql/src/config/tracing.ts
@@ -116,7 +116,7 @@ meterManager.registerMetrics();
 
 export const telemetry = async (context: AuthContext, user: AuthUser, spanName: string, attrs: object, fn: any) => {
   // if tracing disabled or context is not correctly configured.
-  if (!ENABLED_TRACING || !context) {
+  if (!ENABLED_TRACING || !context || !context?.tracing) {
     return fn();
   }
   // if tracing enabled


opencti-platform/opencti-graphql/src/config/tracing.ts
```diff
@@ -116,7 +116,7 @@ meterManager.registerMetrics();
 
 export const telemetry = async (context: AuthContext, user: AuthUser, spanName: string, attrs: object, fn: any) => {
   // if tracing disabled or context is not correctly configured.
-  if (!ENABLED_TRACING || !context) {
+  if (!ENABLED_TRACING || !context || !context?.tracing) {
```
fellowseb, line 119: What's the code path that brings to `tracing` not being defined ? I see `executionContext()` always instantiates one so I'm a bit surprised 🤔 . Maybe it's a specific resolver that passes another object as a context ? If so let's rather fix that resolver :).
I'd rather make sure we have types that don't lie x).

JeremyCloarec (approved):

PR #16056, opened 2026-05-19
[backend] add referer policy (#16020)

### LINKED ISSUE #16020: Update the referer request header to be more restricted

aHenryJard:
### Proposed changes

* Set referer header to the minimal secure version.

See https://developer.mozilla.org/en-US/docs/Web/Security/Practical_implementation_guides/Referrer_policy

### Related issues

* closes #16020 

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/tests/01-unit/http/httpUtils-test.ts
+++ b/opencti-platform/opencti-graphql/tests/01-unit/http/httpUtils-test.ts
@@ -95,7 +95,7 @@ describe('buildHelmetParameters coverage', () => {
       crossOriginOpenerPolicy: false,
       crossOriginResourcePolicy: false,
       referrerPolicy: {
-        policy: 'unsafe-url',
+        policy: 'strict-origin-when-cross-origi',
       },
       xFrameOptions: { action: 'deny' },
     });
@@ -123,7 +123,7 @@ describe('buildHelmetParameters coverage', () => {
       crossOriginOpenerPolicy: false,
       crossOriginResourcePolicy: false,
       referrerPolicy: {
-        policy: 'unsafe-url',
+        policy: 'strict-origin-when-cross-origin',
       },
       xFrameOptions: { action: 'deny' },
     });
@@ -157,7 +157,7 @@ describe('buildHelmetParameters coverage', () => {
       crossOriginOpenerPolicy: false,
       crossOriginResourcePolicy: false,
       referrerPolicy: {
-        policy: 'unsafe-url',
+        policy: 'strict-origin-when-cross-origin',
       },
       xFrameOptions: false,
     });
@@ -185,7 +185,7 @@ describe('buildHelmetParameters coverage', () => {
       crossOriginOpenerPolicy: false,
       crossOriginResourcePolicy: false,
       referrerPolicy: {
-        policy: 'unsafe-url',
+        policy: 'strict-origin-when-cross-origin',
       },
       xFrameOptions: { action: 'deny' },
     });
--- a/opencti-platform/opencti-graphql/tests/01-unit/http/httpUtils-test.ts
+++ b/opencti-platform/opencti-graphql/tests/01-unit/http/httpUtils-test.ts
@@ -95,7 +95,7 @@ describe('buildHelmetParameters coverage', () => {
       crossOriginOpenerPolicy: false,
       crossOriginResourcePolicy: false,
       referrerPolicy: {
-        policy: 'strict-origin-when-cross-origi',
+        policy: 'strict-origin-when-cross-origin',
       },
       xFrameOptions: { action: 'deny' },
     });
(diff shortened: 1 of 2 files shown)


opencti-platform/opencti-graphql/tests/01-unit/http/httpUtils-test.ts
```diff
@@ -95,7 +95,7 @@ describe('buildHelmetParameters coverage', () => {
       crossOriginOpenerPolicy: false,
       crossOriginResourcePolicy: false,
       referrerPolicy: {
-        policy: 'unsafe-url',
+        policy: 'strict-origin-when-cross-origi',
```
aHenryJard:
```suggestion
        policy: 'strict-origin-when-cross-origin',
```

fellowseb (approved):
What prevents us from using a stricter level ?

aHenryJard:
> What prevents us from using a stricter level ?

I already broke some custom onprem installation on previous changes for CSP, so I'm going slowly. Do you think we could use a higher level starting now ?

OctaveLaventure (commented):
Although `strict-origin-when-cross-origin` is generally a balanced default I would use the stricter `same-origin`, my rational is:
- Regarding your concerns after the issues with CSP updates, usually CSP are way more likely to cause urgent issues as they restrict the access to resources, while here it's only metadata that we would stop including, the blast radius is way smaller.
- We already force `no-referrer` in many links in the code.
- I don't think there is a quick way to know if users relies on that information, and if they do, it would be a bad idea IMO because this is a notoriously unreliable metadata and a there are more explicit ways to know where a request comes from such as query params. If it breaks something, it won't stop them from using the platform and anyway it's not a documented supported use-case, at least that I know of.
- I wouldn't use `strict-origin` or `strict-origin-when-cross-origin` as it would keep silently exposing users instances' URL, although they are discoverable with little effort it's still a potentially unwanted privacy leak. I guess an analyst opening a malicious link from the platform could potentially inform the target that they are investigated and by who for example.
- `no-referrer` would remove potentially useful debugging data, for example when debugging redirection loops or auth flows, so I wouldn't use it unless it's a compliance requirement.

fellowseb:
> > What prevents us from using a stricter level ?
> 
> I already broke some custom onprem installation on previous changes for CSP, so I'm going slowly. Do you think we could use a higher level starting now ?

As you wish. We can do this in two steps. Start low and set a higher barrier (`same-origin`) in a few weeks after announcing it in advance. (Can we do this in changelogs for instance ?)

aHenryJard:
> Although `strict-origin-when-cross-origin` is generally a balanced default I would use the stricter `same-origin`, my rational is:
> 
> * Regarding your concerns after the issues with CSP updates, usually CSP are way more likely to cause urgent issues as they restrict the access to resources, while here it's only metadata that we would stop including, the blast radius is way smaller.
> * We already force `no-referrer` in many links in the code.
> * I don't think there is a quick way to know if users relies on that information, and if they do, it would be a bad idea IMO because this is a notoriously unreliable metadata and a there are more explicit ways to know where a request comes from such as query params. If it breaks something, it won't stop them from using the platform and anyway it's not a documented supported use-case, at least that I know of.
> * I wouldn't use `strict-origin` or `strict-origin-when-cross-origin` as it would keep silently exposing users instances' URL, although they are discoverable with little effort it's still a potentially unwanted privacy leak. I guess an analyst opening a malicious link from the platform could potentially inform the target that they are investigated and by who for example.
> * `no-referrer` would remove potentially useful debugging data, for example when debugging redirection loops or auth flows, so I wouldn't use it unless it's a compliance requirement.

> Although `strict-origin-when-cross-origin` is generally a balanced default I would use the stricter `same-origin`, my rational is:
> 
> * Regarding your concerns after the issues with CSP updates, usually CSP are way more likely to cause urgent issues as they restrict the access to resources, while here it's only metadata that we would stop including, the blast radius is way smaller.
> * We already force `no-referrer` in many links in the code.
> * I don't think there is a quick way to know if users relies on that information, and if they do, it would be a bad idea IMO because this is a notoriously unreliable metadata and a there are more explicit ways to know where a request comes from such as query params. If it breaks something, it won't stop them from using the platform and anyway it's not a documented supported use-case, at least that I know of.
> * I wouldn't use `strict-origin` or `strict-origin-when-cross-origin` as it would keep silently exposing users instances' URL, although they are discoverable with little effort it's still a potentially unwanted privacy leak. I guess an analyst opening a malicious link from the platform could potentially inform the target that they are investigated and by who for example.
> * `no-referrer` would remove potentially useful debugging data, for example when debugging redirection loops or auth flows, so I wouldn't use it unless it's a compliance requirement.

Where do we force to no-referrer ? I can't find with a search, is it the `rel="noopener noreferrer"` in React ?

Also I know that referrer is use in some SSO flow, so I will go for the step by step as adviced by @fellowseb.

fellowseb:
>  Where do we force to no-referrer ? I can't find with a search, is it the rel="noopener noreferrer" in React ?

Yes that's where is happens (and should for all)

PR #16062, opened 2026-05-19
[backend/frontend] Fix public dashboard URI key not generated for non-Latin names (#16005)

### LINKED ISSUE #16005: Public dashboard URI key not generated for non-Latin names (Japanese, CJK, Cyrillic, Arabic, etc.)
## Description
When creating a public dashboard, the URI key is auto-generated from the dashboard name by lowercasing, replacing spaces with hyphens, and stripping non-[a-z0-9-] characters. If the name consists entirely of non-Latin characters (e.g. Japanese, Chinese, Korean, Arabic, Cyrillic), every character is stripped and the resulting URI key is an empty string.

A public dashboard with an empty URI key cannot be accessed via its public URL.

## Reproducible Steps
Navigate to Dashboards → Public dashboards
Click Create a public dashboard
Select a custom dashboard
In the Name field, enter a non-Latin name — e.g. 私のダッシュボード
Observe the Public dashboard URI KEY field remains empty
Compare with entering a Latin name like My Dashboard — URI key auto-populates as my-dashboard

## Expected Output
The URI key field should either:
Generate a valid slug (via transliteration, UUID fallback, or allowing Unicode in slugs), or
Show a validation error preventing creation when the generated slug is empty

## Actual Output
The URI key is silently set to an empty string
The dashboard is created but its public URL is broken / inaccessible

## Additional information
The slug function appears to be strictly Latin-only ([a-z0-9-])
This affects all non-Latin scripts: Japanese, Chinese, Korean, Arabic, Hebrew, Thai, Cyrillic, etc.
Mixed names (e.g. ダッシュボード-v1) lose the CJK portion and produce a partial/misleading slug
Workaround: Users can manually type a Latin URI key in the field

## Screenshots 

Image

With Japanese

Image

marieflorescontact:
### Proposed changes
* add a UUID fallback when the generated uri_key is empty.

After investigating the fix for the empty URI key issue with non-Latin dashboard names, I've chosen to keep the URI strictly restricted to [a-z0-9-] characters and add a UUID fallback when the generated slug is empty.
Allowing Unicode characters in URIs would introduce security risks.

### Related issues

* closes: #16005

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/workspaces/dashboards/public_dashboards/PublicDashboardCreationForm.tsx
+++ b/opencti-platform/opencti-front/src/private/components/workspaces/dashboards/public_dashboards/PublicDashboardCreationForm.tsx
@@ -22,6 +22,7 @@ import useAuth from '../../../../../utils/hooks/useAuth';
 import { ME_FILTER_VALUE } from '../../../../../utils/filters/filtersUtils';
 import { fromB64 } from '../../../../../utils/String';
 import FormButtonContainer from '../../../../../components/common/form/FormButtonContainer';
+import { generatePublicDashboardUriKey } from './public-dashboard-utils';
 
 const publicDashboardCreateMutation = graphql`
   mutation PublicDashboardCreationFormCreateMutation($input: PublicDashboardAddInput!) {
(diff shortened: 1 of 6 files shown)


SarahBocognano (approved):
Everything is working great :) Thanks for all the tests

PR #16064, opened 2026-05-19
[frontend] String utils to tsx (#14963)

### LINKED ISSUE #14963: Refactor String.jsx to TS

Archidoit:
### Proposed changes
Convert String.jsx to tsx

### Related issues
fixes #14963

### DIFF
--- a/opencti-platform/opencti-front/src/utils/String.tsx
+++ b/opencti-platform/opencti-front/src/utils/String.tsx
@@ -12,8 +12,16 @@ export const RELATIVE_DATE_REGEX = /^now([-+]\d+[smhHdwMy](\/[smhHdwMy])?)?$/;
 // the value to display if a field is empty or undefined
 export const EMPTY_VALUE = '-';
 
-export const truncate = (str, limit, truncateSpaces = true) => {
-  if (str === undefined || str === null || str.length <= limit) {
+export function truncate(str: string, limit?: number, truncateSpaces?: boolean): string;
+export function truncate(str: undefined, limit?: number, truncateSpaces?: boolean): undefined;
+export function truncate(str: null, limit?: number, truncateSpaces?: boolean): null;
+export function truncate(str: string | undefined | null, limit?: number, truncateSpaces?: boolean): string | undefined | null;
+export function truncate(
+  str: string | undefined | null,
+  limit?: number,
+  truncateSpaces = true,
+): string | undefined | null {
+  if (str === undefined || str === null || (limit && str.length <= limit)) {
     return str;
   }
   const trimmedStr = str.substring(0, limit);
@@ -24,9 +32,20 @@ export const truncate = (str, limit, truncateSpaces = true) => {
     0,
     Math.min(trimmedStr.length, trimmedStr.lastIndexOf(' ')),
   )}...`;
-};
+}
 
-export const adaptFieldValue = (value) => {
+/**
+ * Normalize a field value for use in GraphQL edit inputs.
+ * - If the value is an array, it is returned as-is.
+ * - If the value is nil (null/undefined), returns an empty string.
+ * - Otherwise, converts the value to its string representation.
+ *
+ * @param value The raw field value.
+ * @returns The normalized value as a string or array.
+ */
+export function adaptFieldValue(value: unknown[]): unknown[];
+export function adaptFieldValue(value: unknown): string | unknown[];
+export function adaptFieldValue(value: unknown): string | unknown[] {
   if (Array.isArray(value)) {
     return value;
   }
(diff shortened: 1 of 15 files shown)


opencti-platform/opencti-front/src/utils/String.tsx
```diff
@@ -24,21 +32,32 @@ export const truncate = (str, limit, truncateSpaces = true) => {
...
+ * Normalize a field value for use in GraphQL edit inputs.
+ * - If the value is an array, it is returned as-is.
+ * - If the value is nil (null/undefined), returns an empty string.
+ * - Otherwise, converts the value to its string representation.
+ *
+ * @param value The raw field value.
+ * @returns The normalized value as a string or array.
+ */
+export function adaptFieldValue(value: unknown[]): unknown[];
+export function adaptFieldValue(value: unknown): string | unknown[];
```
fellowseb: Shouldn't this overload case be
```
export function adaptFieldValue(value: unknown): string;
```
?

opencti-platform/opencti-front/src/utils/String.tsx
```diff
@@ -86,46 +105,53 @@ export const isValidStixBundle = (bundle) => {
...
 
-export const fromB64 = (str) => Base64.decode(str);
+export const fromB64 = (str: string): string => Base64.decode(str);
 
-export const fromBase64 = (str) => Base64.encode(str);
+export const fromBase64 = (str: string): string => Base64.encode(str);
 
-export const uniqWithByFields = R.curry((fields, data) => R.uniqWith(R.allPass(R.map(R.eqProps)(fields)))(data));
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export const uniqWithByFields = R.curry((fields: string[], data: any[]) => R.uniqWith(R.allPass(fields.map((f) => R.eqProps(f)) as any) as any, data));
```
fellowseb: Consider removing the usage of `ramda`, as migrating to TS but keeping linter escape hatches and `as any`s is a bit sad (:

```
export const uniqWithByFields = <T extends Record<string, unknown>>(fields: (keyof T)[]) => (data: T[]): T[] => {
  return data.filter((item, index) =>
    data.findIndex((other) =>
      fields.every((field) => item[field] === other[field])
    ) === index
  );
};
```

opencti-platform/opencti-front/src/utils/String.tsx
```diff
@@ -86,46 +105,53 @@ export const isValidStixBundle = (bundle) => {
...
 
-export const computeDuplicates = (fields, data) => R.groupWith(R.allPass(R.map(R.eqProps)(fields)), data);
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export const computeDuplicates = (fields: string[], data: any[]) => R.groupWith(R.allPass(fields.map((f) => R.eqProps(f)) as any) as any, data);
```
fellowseb: Same work here (:

opencti-platform/opencti-front/src/utils/String.tsx
fellowseb, line 1: Doesn't need to be a `tsx` file (:

opencti-platform/opencti-front/src/utils/String.tsx
```diff
@@ -86,46 +105,53 @@ export const isValidStixBundle = (bundle) => {
...
+    .replace(/[^a-zA-Z0-9 ]/g, '')
+    .replace(/(?:^\w|[A-Z]|\b\w)/g, (word, i) => {
+      return i === 0 ? word.toLowerCase() : word.toUpperCase();
+    })
+    .replace(/\s+/g, '');
 };
 
-export const renderObservableValue = (observable) => {
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export const renderObservableValue = (observable: any) => {
```
fellowseb: Could we get some typing here ?

Also from what I see a lot of these functions shouldn't be in a "string utils" lib 🤔

opencti-platform/opencti-front/src/utils/String.tsx
Archidoit: yes it does, because of the html code in it :/

opencti-platform/opencti-front/src/utils/String.tsx
Archidoit: done

opencti-platform/opencti-front/src/utils/String.tsx
Archidoit: done

opencti-platform/opencti-front/src/utils/String.tsx
Archidoit: will remove them :)

fellowseb (approved):

PR #16067, opened 2026-05-19
[backend] authorized_members in playbook should be an array (#15986)

### LINKED ISSUE #15986: Picker in Send to notifier component not accepting typing
## Description

Using the label selection dropdown a user can not type on initial load. The field is wiped and then reset. Once a user selects one label the user can then type correctly.

## Environment

2. OpenCTI version: 7.260510.0
latest-octi-staging
demo instance

## Reproducible Steps

1. Open a send notifer component in playbooks
2. click to add an action
3. add, label, then try to type a label 

## Expected Output

characters inputted into the label selection are shown and the dropdown is filtered corectly
## Actual Output

see video, characters entered are not shown and the list is not filtered. 
## Additional information

Almost identical to this issue https://github.com/OpenCTI-Platform/opencti/issues/15963 which suggests it is a common issue with the component - so other instances should be checked too

https://github.com/user-attachments/assets/a2fdcbd8-2910-4d19-a834-1bcd2a8f6edb

.

esrevi:
### Proposed changes

* Update authorized_members schema because frontend form expect an array instead of an object for the autocomplete
component

### Related issues

* Fixes #15986 

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/modules/playbook/playbook-components.ts
+++ b/opencti-platform/opencti-graphql/src/modules/playbook/playbook-components.ts
@@ -627,7 +627,7 @@ const PLAYBOOK_RULE_COMPONENT: PlaybookComponent<RuleConfiguration> = {
 
 export interface NotifierConfiguration {
   notifiers: string[];
-  authorized_members: object;
+  authorized_members: { value: string }[];
 }
 const PLAYBOOK_NOTIFIER_COMPONENT_SCHEMA: JSONSchemaType<NotifierConfiguration> = {
   type: 'object',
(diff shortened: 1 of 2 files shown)


opencti-platform/opencti-graphql/src/modules/playbook/playbook-components.ts
```diff
@@ -599,7 +639,17 @@ const PLAYBOOK_NOTIFIER_COMPONENT_SCHEMA: JSONSchemaType<NotifierConfiguration>
       $ref: 'Notifiers',
       items: { type: 'string', oneOf: [] },
     },
-    authorized_members: { type: 'object' },
+    authorized_members: {
```
fellowseb, line 642: Isn't this a potential breaking change in terms of API  or datamodel against already saved data ?

SarahBocognano (commented):
Everything is working well, I don't think we need a migration as we were already saving authorized_members as an array in database, it's just the wrong schema for react component. But maybe a small verification in database to be sure we weren't stocking an object before approval would be great :)

SarahBocognano (approved):

opencti-platform/opencti-graphql/src/modules/playbook/playbook-components.ts
esrevi: actually, it seems that authorized_members in ES are stored as array, this fixes the schema declaration which was wrong, so it should not break the API

PR #16104, opened 2026-05-21
feat(investigation): add select nodes by edge count to graph toolbar

### LINKED ISSUE #16927: feat(investigation): add ability to select graph nodes by edge count
## Use case

When working with a large investigation graph, an analyst often wants to isolate nodes based on how connected they are — e.g. finding hubs (high edge count) or leaf/orphan nodes (0–1 edges). Today there is no way to select nodes by their number of visible edges directly from the toolbar; the analyst has to eyeball the graph and click nodes one at a time, which does not scale past a handful of entities.

Common concrete tasks this would enable:
- Quickly find and select all pivots/hubs above a connectivity threshold (e.g. every node with exactly N relationships in the current view).
- Locate leaf nodes (edge count of 1) that may be candidates for pruning or further expansion.
- Combine with entity-type filtering so the operation is scoped (e.g. only Threat Actors with 3 edges).

## Current workaround

Manually inspecting each node and clicking to add it to the selection, or exporting the graph data and filtering externally. Both are slow and error-prone on non-trivial graphs.

## Proposed solution

Add a new toolbar action in the investigation graph's *Select Tools* section — **Select nodes by edge count** — placed to the right of the existing *Select Relationships of Selected Nodes* button. Clicking it opens a modal (consistent with the existing Expand modal) that lets the user:

- Enter a target **edge count** (number of visible edges in the current graph)
- Optionally filter by **entity type** using the same All/None checkbox pattern as the Expand modal

On submit, every node in the current graph matching both criteria is selected and the modal closes. Empty entity-type selection is a no-op, mirroring the existing Expand-modal guard.

## Additional information

Implementation submitted in PR #16104.

Yes — PR #16104 is already open with the implementation.

metac0rtex:
Closes #16927

## Description

Adds a **Select nodes by edge count** button to the investigation graph toolbar, positioned to the right of the existing *Select Relationships of Selected Nodes* button.

Clicking the button opens a modal (consistent with the existing Expand modal) where the user can:
- Enter a target **edge count** (number of visible edges in the current graph)
- Optionally filter by **entity type** using the same All/None checkbox pattern as the Expand modal

On submit, all nodes in the current graph that match both criteria are selected and the modal closes.

## Changes

- **`InvestigationSelectByEdgeForm.tsx`** (new) — Formik form with a number input for edge count and a `CheckboxesField` for entity types populated from the current graph nodes
- **`GraphToolbarSelectTools.tsx`** (modified) — adds the toolbar button, dialog, and selection handler; edge counting uses `source_id`/`target_id` from graph links; HTML input value is coerced to `Number` before comparison; empty entity type selection is a no-op (consistent with Expand modal behaviour)
- **`GraphToolbarExpandTools.tsx`** (modified) — removes the button/dialog/handler that was briefly placed here during development

## Behaviour details

- If no entity types are selected when **Select** is clicked, nothing happens (mirrors the Expand modal guard)
- Selecting **All** then submitting selects every node whose visible edge count equals the specified number
- The selection replaces any existing node/link selection

## Testing

1. Open an investigation with multiple nodes and relationships
2. Click the **Select nodes by edge count** button (hub icon) in the toolbar
3. Enter an edge count and optionally filter by entity type
4. Click **Select** — only nodes matching both criteria should be highlighted

### DIFF
--- a/opencti-platform/opencti-front/src/components/graph/components/GraphToolbarExpandTools.tsx
+++ b/opencti-platform/opencti-front/src/components/graph/components/GraphToolbarExpandTools.tsx
@@ -677,6 +677,7 @@ const GraphToolbarExpandTools = ({
           onReset={() => setIsExpandOpen(false)}
         />
       </Dialog>
+
     </>
   );
 };
--- a/opencti-platform/opencti-front/src/components/graph/components/GraphToolbarSelectTools.tsx
+++ b/opencti-platform/opencti-front/src/components/graph/components/GraphToolbarSelectTools.tsx
@@ -1,11 +1,13 @@
 import { SelectAll, SelectGroup, SelectionDrag } from 'mdi-material-ui';
-import { GestureOutlined, SwipeDown, SwipeUp, SwipeVertical, TouchApp } from '@mui/icons-material';
+import { DeviceHubOutlined, GestureOutlined, SwipeDown, SwipeUp, SwipeVertical, TouchApp } from '@mui/icons-material';
 import React, { useState } from 'react';
+import Dialog from '@mui/material/Dialog';
 import GraphToolbarOptionsList from './GraphToolbarOptionsList';
 import GraphToolbarItem from './GraphToolbarItem';
 import { useFormatter } from '../../i18n';
 import { useGraphContext } from '../GraphContext';
 import useGraphInteractions from '../utils/useGraphInteractions';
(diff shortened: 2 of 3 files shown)


Gwendoline-FAVRE-FELIX:
Thanks a lot for your contribution and for submitting this PR.
Before we can start the review, could you please :
- sign your commit;
- create an issue describing the request feature, and link this PR to it?

RBourgeoisGH:
Hi,
Thanks for your contribution, we noticed that there is still a few other checks failing. We invite you to fix them before we take another look.

Regards,

PR #16117, opened 2026-05-21
feat(frontend): added SecurityCoverageInformation component to relationship view (#13133)

### LINKED ISSUE #13133: feat(security-coverage): add security coverage metrics in the "has-covered" relationship view
## Use case

The basic relationship view is the same for every relationship, but in the case of the "has-covered" relationship (between a security coverage and a vulnerability for instance), there are metrics associated (detection, prevention) that are not displayed. 

The current data model as of 7.260401.0, means that even entities that are not covered (do not have executable tests) have a relationship of 'has covered' with no indication on the relationship view that the entity is not covered.

Manually created 'has covered' relationships can be populated initially with the coverage metric but can then not be updated by the user from the relationship editor view. 

## Current Workaround

Metrics are only visible through the Security Coverage Result tab itself.

Updating a manual security coverage metric is only possible from the specific linked entity on the security coverage overview page, this is unintuitive for users of the new Result tab. 

## Proposed Solution

If metrics are associated with this relationship, display them in the relationship details and available to be reported.

Enable these metrics to be editable, for relationships created manually. 

ADDITIONAL DETAILS: 

Linked to this it is possible in a manual security coverage to create bulk 'has covered' relationships, this does not allow the user to add a coverage metric for these relationships.

MTorbay-Filigran:
### Proposed changes

* Added a coverage information view in investigation (right bar) and relationship view for has-covered relationships
* Added the edition of the coverage in the same views
* Corrected the coverage logo in the investigation view
* Corrected the json exports for has-covered relationships and security coverage to add the coverage information 

### Related issues

* Fix #13133 

### Checklist


SouadHadjiat (approved):
Code review done, all good ✅
Waiting for Product validation

tobybutler405:
product validated

SouadHadjiat (approved):

PR #16144, opened 2026-05-22
fix(work): fix History/Activity leaking in works queries (#17349)

### LINKED ISSUE #16049: fix(filters): `entity_type` filter unconditionally expands to `parent_types`, leaking History/Activity documents into concrete-type queries

fellowseb:
### Proposed changes

* Fixes typo in query params passed to `elPaginate` in a few `work` resolvers.
* It didn't really have a visible effect in the app nor in client-python given there was always an additional filter that would filter on `work` entities; but resulted in an error when querying the `works` query in GraphQL playground.

### Related issues

* Fixes #16049 

### Testing instructions

Execute the `works` query in the GraphQL playground: if you have History/Activity entities it should crash without this fix.

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/database/engine.ts
+++ b/opencti-platform/opencti-graphql/src/database/engine.ts
@@ -4997,8 +4997,10 @@ export const getStats = (indices = READ_PLATFORM_INDICES) => {
 
 export const isEngineAlive = async () => {
   const context = executionContext('healthcheck');
-  const options = { types: [ENTITY_TYPE_MIGRATION_STATUS], connectionFormat: false };
-  const migrations = await elPaginate(context, SYSTEM_USER, READ_INDEX_INTERNAL_OBJECTS, options) as BasicStoreBase[];
+  const migrations = await elPaginate(context, SYSTEM_USER, READ_INDEX_INTERNAL_OBJECTS, {
+    types: [ENTITY_TYPE_MIGRATION_STATUS],
+    connectionFormat: false,
+  }) as BasicStoreBase[];
   if (migrations.length === 0) {
     throw DatabaseError('Invalid database content, missing migration schema');
   }
--- a/opencti-platform/opencti-graphql/src/database/engine.ts
+++ b/opencti-platform/opencti-graphql/src/database/engine.ts
@@ -4997,10 +4997,8 @@ export const getStats = (indices = READ_PLATFORM_INDICES) => {
 
 export const isEngineAlive = async () => {
   const context = executionContext('healthcheck');
-  const migrations = await elPaginate(context, SYSTEM_USER, READ_INDEX_INTERNAL_OBJECTS, {
-    types: [ENTITY_TYPE_MIGRATION_STATUS],
-    connectionFormat: false,
-  }) as BasicStoreBase[];
+  const options = { types: [ENTITY_TYPE_MIGRATION_STATUS], connectionFormat: false };
+  const migrations = await elPaginate(context, SYSTEM_USER, READ_INDEX_INTERNAL_OBJECTS, options) as BasicStoreBase[];
   if (migrations.length === 0) {
     throw DatabaseError('Invalid database content, missing migration schema');
   }
--- a/opencti-platform/opencti-graphql/tests/01-unit/domain/work-query-test.ts
+++ b/opencti-platform/opencti-graphql/tests/01-unit/domain/work-query-test.ts
@@ -1,0 +1,80 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import { ENTITY_TYPE_WORK } from '../../../src/schema/internalObject';
+
+const mockElPaginate = vi.fn();
+const mockAddFilter = vi.fn();
+
+vi.mock('../../../src/database/engine', async (importOriginal) => {
+  const actual = await importOriginal<typeof import('../../../src/database/engine')>();
+  return {
+    ...actual,
+    elPaginate: (...args: unknown[]) => mockElPaginate(...args),
+  };
+});
+
+vi.mock('../../../src/utils/filtering/filtering-utils', async (importOriginal) => {
+  const actual = await importOriginal<typeof import('../../../src/utils/filtering/filtering-utils')>();
+  return {
+    ...actual,
+    addFilter: (...args: unknown[]) => mockAddFilter(...args),
+  };
+});
+
+import { findWorkPaginated, worksForConnector, worksForSource } from '../../../src/domain/work';
+
+describe('Work domain query options', () => {
+  const context = {} as any;
+  const user = {} as any;
+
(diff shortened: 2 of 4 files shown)


opencti-platform/opencti-graphql/src/database/engine.ts
```diff
@@ -4987,8 +4987,10 @@ export const getStats = (indices = READ_PLATFORM_INDICES) => {
 
 export const isEngineAlive = async () => {
   const context = executionContext('healthcheck');
-  const options = { types: [ENTITY_TYPE_MIGRATION_STATUS], connectionFormat: false };
-  const migrations = await elPaginate(context, SYSTEM_USER, READ_INDEX_INTERNAL_OBJECTS, options) as BasicStoreBase[];
+  const migrations = await elPaginate(context, SYSTEM_USER, READ_INDEX_INTERNAL_OBJECTS, {
+    types: [ENTITY_TYPE_MIGRATION_STATUS],
+    connectionFormat: false,
+  }) as BasicStoreBase[];
```
fellowseb: Made those changes as using intermediate variables removes TS ability to check for extra props and so typos...

fellowseb: Reverting as I can't pass the 80% coverage because of it...

CelineSebe:
Tested locally, works well:
image

opencti-platform/opencti-graphql/tests/01-unit/domain/work-query-test.ts
```diff
@@ -0,0 +1,80 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
```
CelineSebe, line 1: It might be useful to add a test for the `worksForDraft` function

CelineSebe (approved):

PR #16145, opened 2026-05-22
fix(backend): fix force local auth env var not bypassing ldap config (#16131)

### LINKED ISSUE #16131: fix(authentication): APP__AUTHENTICATION__FORCE_LOCAL doesn't override faulty LDAP config
## Description

If a faulty LDAP configuration is set in the platform, the users are stuck on the login form (invalid LDAP URL for example).

The LDAP auth uses the basic login form to perform the auth queries, and even coupled with APP__AUTHENTICATION__FORCE_LOCAL: true, login form doesn't fall back to actual local auth but still tries to reach the LDAP server.

## Environment

1. OS (where OpenCTI server runs): Kube
2. OpenCTI version: Tested on 7.260428.0 but is probably true for all 7+ 
4. Other environment details:

## Reproducible Steps

Create new LDAP auth strategy 

Use "random" values for the fields : 
URL : 127.0.0.1
Bind DN:  cn=hello

## Expected Output

Without APP__AUTHENTICATION__FORCE_LOCAL: true 
Expected to fail (faulty LDAP config) 

With APP__AUTHENTICATION__FORCE_LOCAL: true
Expected to login through local auth

## Actual Output

With APP__AUTHENTICATION__FORCE_LOCAL: true
The platform still tries to log through LDAP server

troll-os:
### Proposed changes

* Added check to see if APP__AUTHENTICATION__FORCE_LOCAL is set when creating a new provider avoid enforcing it
* Added check to see if APP__AUTHENTICATION__FORCE_LOCAL is set when enabling providers at platform startup to avoid enforcing them 
* Added relevant logs for both cases

### Related issues

* https://github.com/OpenCTI-Platform/opencti/issues/16131

### Checklist

- [ ] I wrote test cases for the relevant uses case (coverage and e2e) (_I think it's already covered by existing tests_)

### DIFF
--- a/opencti-platform/opencti-graphql/src/modules/authenticationProvider/authenticationProvider-domain.ts
+++ b/opencti-platform/opencti-graphql/src/modules/authenticationProvider/authenticationProvider-domain.ts
@@ -30,7 +30,7 @@ import { notify, redisDeleteAuthLogHistory } from '../../database/redis';
 import conf, { BUS_TOPICS } from '../../config/conf';
 import { ABSTRACT_INTERNAL_OBJECT } from '../../schema/general';
 import { unregisterStrategy } from './providers';
-import {isAuthenticationForcedFromEnv, isLocalAuthForcedEnabledFromEnv} from './providers-configuration';
+import { isAuthenticationForcedFromEnv, isLocalAuthForcedEnabledFromEnv } from './providers-configuration';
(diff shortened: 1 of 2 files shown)


troll-os:
Open to discuss tests coverage

I dug in existing tests and failed to see how providers dynamic and startup loading are currently tested

Issue #15924, opened 2026-05-11
feat: dynamic token support for OAuth SMTP


Rlongares:
### Use case

Exchange Online will permanently remove support for Basic authentication with Client Submission (SMTP AUTH) gradually beginning with a small percentage of submission rejections for all tenants on March 1st 2026 and reaching 100% rejections on April 30th 2026, (previously September 2025).

After April 2026, we will remove support for Basic auth with the Client Submission (SMTP AUTH) endpoints:

smtp.office365.com
smtp-legacy.office365.com
Once Basic auth is permanently disabled, any clients or apps connecting using Basic auth with Client Submission (SMTP AUTH) will receive this response:

550 5.7.30 Basic authentication is not supported for Client Submission.

------------------------------
A recent update has added support for OAuth (https://github.com/OpenCTI-Platform/opencti/issues/12069). However, this implementation relies on a static token. In my case, the token needs to be refreshed periodically, as static tokens are not a viable option for us, but this use case is not currently supported.

### Current Workaround
No known workaround.

### Proposed Solution
OAuth support or additional documentation on the required configuration change.

### Additional Information
Reference: https://techcommunity.microsoft.com/blog/exchange/exchange-online-to-retire-basic-auth-for-client-submission-smtp-auth/4114750

Until general availability, this feature is currently available under Feature Flag

PR #16181, opened 2026-05-26
feat(backend): refresh token support for OAuth2 SMTP (#15924)

marieflorescontact:
### Proposed changes
 For OAuth2 the access token is refreshed at call time so that long-running instances do not fail once the initial token has expired

* We use` openid-client` library (already in use in the project) to smtp refresh access token
* No more static `oauthAccessToken `: replaced by `oauthIssuer `+ `oauthRefreshToken`
* we build a fresh transporter at each call to have a fresh token for Oauth2
* unit tests 

### Related issues

* closes: https://github.com/OpenCTI-Platform/opencti/issues/15924

### Checklist

### DIFF
--- a/docs/docs/deployment/configuration.md
+++ b/docs/docs/deployment/configuration.md
@@ -248,26 +248,36 @@ For a detailed list of exposed metrics, please refer to the [Telemetry](../deplo
 
 #### SMTP Service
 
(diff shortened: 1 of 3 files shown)


ludovic (commented):
Tested with a Google account, works perfectly!

ludovic (approved):

PR #16193, opened 2026-05-26
feat(client-python): export Malware STIX 2.0 using GraphQL (#15138)

### LINKED ISSUE #15138: feat(client-python): STIX 2.0 export via API for Malware
### Use Case

use new OCTI  STIX 2.0 exporter instead of client

### Current Workaround

_No response_

### Proposed Solution

_No response_

### Additional Information

_No response_

### If the feature request is approved, would you be willing to submit a PR?

None

marieflorescontact:
### Proposed changes

* Delegate Malware STIX 2.0 serialization to the GraphQL backend field malware(id) { toStix(version: stix_2_0) }.
* Add Malware.to_stix_2_0(id) in pycti/entities/opencti_malware.py
* Wire generate_export through the dispatch: short-circuit the legacy converter when the backend returns a payload.
* Skip the stix_nested_ref_relationship enrichment block in prepare_export for backend-managed types, to prevent duplicated operating_system_refs / sample_refs (backend already produces them deterministically).
* mode="full" orchestration is unchanged
* Add  tests 

### Related issues

* closes : https://github.com/OpenCTI-Platform/opencti/issues/15138

### Checklist

### Further comments
small script to help you test functionnaly : spec version should be 2.0 and *.refs should not be duplicated

````
# coding: utf-8

import json
import os

from pycti import OpenCTIApiClient

# Variables
api_url = os.getenv("OPENCTI_API_URL", "http://localhost:3000")
api_token = os.getenv("OPENCTI_API_TOKEN", "token-token-token-token-token")

# OpenCTI initialization
opencti_api_client = OpenCTIApiClient(api_url, api_token)

# Create Malware
malware = opencti_api_client.malware.create(name="My new malware")
# Create the STIX 2.0 bundle
bundle = opencti_api_client.stix2.get_stix_bundle_or_object_from_entity_id(
    entity_type="Malware",
    entity_id=malware["id"],
    mode="simple",
)
json_bundle = json.dumps(bundle, indent=4)

# Write the bundle
with open("malware_stix_2_0.json", "w") as f:
    f.write(json_bundle)

print(json_bundle)
````

### DIFF
--- a/client-python/pycti/entities/opencti_malware.py
+++ b/client-python/pycti/entities/opencti_malware.py
@@ -461,6 +461,33 @@ class Malware:
             )
             return None
 
+    def to_stix_2_0(self, **kwargs):
+        """Get Malware STIX 2.0 from the backend.
+        :param id: the id of the Malware
+        :type id: str
+        :return: Malware STIX 2.0, or ``None`` if not found
+        :rtype: dict or None
+        """
+        id = kwargs.get("id", None)
+        if id is None:
+            self.opencti.app_logger.error(
+                "[opencti_malware] to_stix_2_0: missing parameter: id"
+            )
+            return None
+        self.opencti.app_logger.info("Fetching Malware STIX 2.0", {"id": id})
+        query = """
+            query Malware($id: String!) {
+                malware(id: $id) {
+                    toStix(version: stix_2_0)
+                }
+            }
+        """
+        result = self.opencti.query(query, {"id": id})
+        malware = (result.get("data") or {}).get("malware")
+        if malware is None or malware.get("toStix") is None:
+            return None
+        return json.loads(malware["toStix"])
+
     def create(self, **kwargs):
         """Create a Malware object.
 
--- a/client-python/pycti/utils/opencti_stix2.py
+++ b/client-python/pycti/utils/opencti_stix2.py
@@ -59,6 +59,32 @@ ERROR_TYPE_TIMEOUT = "Request timed out"
 #: STIX Extension ID for OpenCTI custom objects and properties
 STIX_EXT_OCTI: str = "extension-definition--ea279b3e-5c71-4632-ac08-831c66a786ba"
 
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: client-python/pycti/entities/opencti_malware.py [hunk 1]
```diff
@@ -461,6 +461,33 @@ def read(self, **kwargs):
...
+                    toStix(version: stix_2_0)
+                }
+            }
+        """
+        result = self.opencti.query(query, {"id": id})
+        malware = (result.get("data") or {}).get("malware")
+        if malware is None or malware.get("toStix") is None:
+            return None
+        return json.loads(malware["toStix"])
+
```
JeremyCloarec, line 490: Since toStix on Graphql has a version parameter, this method could be refactored to a to_stix method, with the version being a parameter

client-python/pycti/utils/opencti_stix2.py
```diff
@@ -59,6 +59,32 @@
...
+_STIX_2_0_PRESERVED_KEYS = (
+    "createdBy",
+    "createdById",
+    "objectMarking",
+    "objectMarkingIds",
+    "objectOrganization",
+    "importFiles",
+    "importFilesIds",
+)
+
```
JeremyCloarec, line 87: I don't think this is the right way to go. We need to decide if we want to handle ref resolution on the backend side in the toStix method or not. I think adding a new parameter to toStix that enable ref resolution on the backend side is the way to go. This way, the whole prepare_export will be obsolete since it will also be done backend side

client-python/pycti/entities/opencti_malware.py
marieflorescontact: done

client-python/pycti/utils/opencti_stix2.py
marieflorescontact: I asked @richard-julien :the problem is the volume of linked objects, it may crash on the server side due to sizes that could be large.

PR #16222, opened 2026-05-28
[frontend/backend] Decay rule filters should not be mandatory (#16219)

### LINKED ISSUE #16219: [Decay Rules] Unable to visualize a Decay Rule definition
## Description

An exception occurred when trying to consult a decay rules with the following errors: 

```
{
    "errors": [
        {
            "message": "Cannot return null for non-nullable field DecayRule.decay_filters.",
            "locations": [
                {
                    "line": 22,
                    "column": 3
                }
            ],
            "path": [
                "decayRule",
                "decay_filters"
            ],
            "extensions": {
            },
        }
    ],
    "data": {
        "decayRule": null
    }
}
```

The issue only appears to impact old decay rules created before the introduction of the decay_filters field.
Newly created decay rules work correctly because they are initialized with a valid decay_filters value, while older rules still contain null or missing values for this field.

Image

## Environment

OpenCTI version: 7.260527.0

Archidoit:
### Proposed changes
- Decay rules filters should not be mandatory in graphql definition
- Fixes the error when displaying the decay rules created before the introduction of decay_filters
- e2e test to check the display of a decay rule overview
- add margin between 'update' button of a decay rule and the popover menu

### Related issues
fixes #16219

### DIFF
--- a/opencti-platform/opencti-graphql/tests/02-integration/modules/decayRules/decayRule-resolver-test.ts
+++ b/opencti-platform/opencti-graphql/tests/02-integration/modules/decayRules/decayRule-resolver-test.ts
@@ -131,6 +131,7 @@ describe('DecayRule resolver standard behavior', () => {
         expect(decayRule.decay_lifetime).toBeDefined();
         expect(decayRule.decay_revoke_score).toBeDefined();
         expect(decayRule.order).toBeDefined();
+        expect(decayRule).toHaveProperty('decay_filters');
         logApp.info('One built-in decay rule is', { decayRule });
         if (decayRule.name === TEST_FALLBACK_DECAY_RULE.name) {
           defaultDecayRuleId = decayRule.id || '';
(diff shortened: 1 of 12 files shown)


opencti-platform/opencti-graphql/tests/02-integration/modules/decayRules/decayRule-resolver-test.ts
```diff
@@ -129,7 +129,6 @@ describe('DecayRule resolver standard behavior', () => {
         expect(decayRule.id).toBeDefined();
         expect(decayRule.name).toBeDefined();
         expect(decayRule.decay_lifetime).toBeDefined();
-        expect(decayRule.decay_filters).toBeDefined();
```
nvkhope, line 132: Suggestion: Rather than removing entierly, maybe replace with a softer check like: 

`expect(decayRule).toHaveProperty('decay_filters');`

nvkhope (approved):

opencti-platform/opencti-graphql/tests/02-integration/modules/decayRules/decayRule-resolver-test.ts
Archidoit: done :)

nvkhope (approved):

PR #16229, opened 2026-05-28
[frontend] add Div extension to preserve <div> elements in RichTextEditor (#16228)

### LINKED ISSUE #16228: TipTap drops inherited parent div styles (font-size) when importing external HTML

ludovic:
### Proposed changes

*  add Div extension to preserve `<div>` elements in RichTextEditor

### Related issues

* Fixes #16228 
*

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/components/richTextEditor/extensions/Div.ts
+++ b/opencti-platform/opencti-front/src/components/richTextEditor/extensions/Div.ts
@@ -1,0 +1,45 @@
+import { Node, mergeAttributes } from '@tiptap/core';
+
+/**
+ * TipTap node extension that preserves <div> elements from externally-created HTML.
+ *
+ * Without this extension TipTap (ProseMirror) has no schema node for <div>, so it
+ * silently unwraps every <div> and discards its inline style / class attributes.
+ * This means font-size, color and layout styles defined on parent divs are lost.
+ *
+ */
+export const Div = Node.create({
+  name: 'div',
+  group: 'block',
+  content: 'block+',
+  defining: true,
+
+  addAttributes() {
+    return {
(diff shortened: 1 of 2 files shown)


CelineSebe:
Tested locally, works welle 👌

opencti-platform/opencti-front/src/components/richTextEditor/extensions/Div.ts
```diff
@@ -0,0 +1,45 @@
...
+  name: 'div',
+  group: 'block',
+  content: 'block+',
+  defining: true,
+
+  addAttributes() {
+    return {
+      style: {
+        default: null,
+        parseHTML: (element: HTMLElement) => element.getAttribute('style') || null,
```
CelineSebe, line 21: To improve security, you could sanitize the style attribute by allowing only a whitelist of safe CSS properties (font-size, color, margin, padding, text-align, etc.)

CelineSebe (approved):

opencti-platform/opencti-front/src/components/richTextEditor/extensions/Div.ts
ludovic: I understand the security concern, but after lookng into it, sanitizing or whitelisting CSS properties is highly likely to break the formatting of legitimate pasted HTML content (such as MJML newsletters, complex tables with specific widths, alignments, etc., which rely on width, max-width, vertical-align, display, and more).

Since the final HTML is already sanitized by DOMPurify in HtmlDisplay for read-only rendering, the actual risk is very limited. I think it's better to preserve the style attribute as-is to ensure maximum compatibility and avoid breaking the copy-paste experience for analysts.

PR #16255, opened 2026-05-28
[frontend] Hide trial banner 'Contact us' when not registered to XTM Hub and polish dialogs (#16254)

### LINKED ISSUE #16254: [Frontend] Trial banner: hide 'Contact us' when platform is not registered to XTM Hub and polish dialogs
## Description

When the platform runs on an Enterprise Edition trial license, the top banner reads `Your OpenCTI Enterprise Edition free trial is active: N Days remaining` and exposes a **Contact us** button. Clicking it opens a Formik dialog that ultimately calls the `contactUsXtmHub(message)` GraphQL mutation, which proxies the message to XTM Hub using the platform's `settings.xtm_hub_token` as authentication.

Two problems with the current implementation in `opencti-platform/opencti-front/src/private/components/LicenseBanner.tsx`:

### 1. The button is functionally broken when the platform is not registered to XTM Hub

The backend resolver short-circuits as soon as `settings.xtm_hub_token` is empty:

```ts
// opencti-platform/opencti-graphql/src/domain/xtm-hub.ts
export const contactUsXtmHub = async (context, user, message) => {
  const settings = await getEntityFromCache<BasicStoreSettings>(context, user, ENTITY_TYPE_SETTINGS);
  if (!settings.xtm_hub_token) {
    logApp.warn('[XTMH] Cannot contact XTM Hub: no token found');
    return { success: false };
  }
  ...
};
```

The frontend, however, does not inspect the response and always displays the `Thank you for reaching out, we'll get back to you shortly.` dialog, so the user believes the message was delivered while in reality nothing was sent.

A trial license without a registered platform (no Hub token) is a common scenario (license imported manually, Hub registration not completed, registration reset by the Hub on `not_found`), so the broken state is very reachable.

### 2. The two dialogs do not follow OpenCTI's design system

Looking at the dialogs side-by-side with `DeleteDialog`, `GraphToolbarRemoveConfirm`, `RequestAccessDialog` and the rest of the modernised forms:

- Both dialogs use raw `@mui/material/Dialog` + `DialogTitle` + `DialogContent` + `DialogActions` instead of the project-wide `@common/dialog/Dialog` wrapper, so spacing/padding are inconsistent (no top margin between the title and the message field, text glued to the dialog edges).
- The submit button is `<Button color="secondary">` from MUI, which renders as the OpenCTI EE/AI bright green — not the standard primary blue used by all other Formik dialogs in the app.
- The Cancel button is the default MUI variant rather than `variant="secondary"` from `@common/button/Button`.
- The contact form dialog incorrectly uses `Thank you!` as its title (the post-submit dialog should be the only one with that title).
- The post-submit dialog passes `title="Thank you!"` as a prop to MUI's `Dialog` — which is just an HTML attribute on the root element and not rendered as a heading, so the dialog has no visible title at all.

### Expected behaviour

- The **Contact us** button on the trial banner should only be visible when `settings.xtm_hub_token` is non-empty (i.e. the platform is actually registered to XTM Hub and the mutation can succeed). When there is no token, only the informational `N Days remaining` text should be shown.
- The two dialogs should reuse `@common/dialog/Dialog` and `@common/button/Button` like `DeleteDialog`, with proper title, spacing, and the primary blue Validate / Close buttons.

### Steps to reproduce

1. Run OpenCTI EE with a `trial` license but **without** completing the XTM Hub registration (or after `xtm_hub_token` was cleared by `resetRegistration`).
2. Observe the top banner: the **Contact us** button is displayed.
3. Click it, type any message, click **Validate**.
4. The success dialog appears as if the message was delivered.
5. Server logs show `[XTMH] Cannot contact XTM Hub: no token found`; nothing reaches the Hub.

### Environment

- OpenCTI server: `master` (reproduced on current `master`).
- OpenCTI client: frontend.

### Screenshots

See the screenshots attached to the related PR for the broken dialog layout/styling.

SamuelHassine:
### Proposed changes

* Hide the **Contact us** button on the EE trial banner when the platform is not registered to XTM Hub (`settings.xtm_hub_token` empty). Previously the button was always visible, but `contactUsXtmHub` short-circuits server-side without a token, so users saw a misleading "Thank you for reaching out" confirmation while no message was ever delivered to the Hub.
* Rebuild both the contact form and the post-submit thank-you dialog on top of the project-wide `@common/dialog/Dialog` + `@common/button/Button` design system primitives (same pattern as `DeleteDialog`, `GraphToolbarRemoveConfirm`, etc.), so we now have:
  - Proper dialog titles (`Contact us` for the form, `Thank you!` for the confirmation - previously the form was titled "Thank you!" and the confirmation had no title at all because the `title` prop was being passed to MUI's `Dialog`, where it just lands as an HTML attribute).
  - Consistent padding/spacing (no more text glued to the edges).
  - Primary blue **Validate** / **Close** buttons (instead of the off `color="secondary"` MUI green) and a `variant="secondary"` **Cancel** button.
  - Proper `setSubmitting` handling on success/error.

### Related issues

* Closes #16254

### Checklist

### Further comments

This PR keeps the behavioural change minimal (the button is now conditional on a single new boolean read from `RootSettings`). The visual refactor is intentionally scoped to the two dialogs already living in `LicenseBanner.tsx`; no other components are touched and no translation keys are added or removed (all strings already exist in the lang catalogs).

E2E/unit tests are not added here because the banner is gated behind a valid EE trial license + a fully provisioned XTM Hub token, which neither the CI license nor the test fixtures currently provide. Happy to wire something up if reviewers want it.

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/LicenseBanner.tsx
+++ b/opencti-platform/opencti-front/src/private/components/LicenseBanner.tsx
@@ -16,7 +16,6 @@ import { UserContext } from '../../utils/hooks/useAuth';
 import { daysBetweenDates, now } from '../../utils/Time';
 import { RootSettings$data } from '../__generated__/RootSettings.graphql';
 import TextField from '../../components/TextField';
-import { isNotEmptyField } from '../../utils/utils';
 
 export const LICENSE_OPTION_TRIAL = 'trial';
 
@@ -85,9 +84,11 @@ const computeBannerInfo = (
   if (eeSettings.license_type === LICENSE_OPTION_TRIAL) {
     const remainingDays = daysBetweenDates(now(), moment(eeSettings.license_expiration_date));
     const bannerColor = getBannerColor(remainingDays);
-    // The "Contact us" button only makes sense when the platform is registered
-    // to XTM Hub (xtm_hub_token present), since the contactUsXtmHub mutation
-    // short-circuits server-side without a token.
+    // The "Contact us" button only makes sense when the platform is actively
+    // registered to XTM Hub, since contactUsXtmHub short-circuits server-side
+    // without a valid Hub token. We rely on the registration status rather than
+    // the token alone, because a stale token can linger after a failed
+    // unregistration handshake.
     const showContactButton = canContactHub;
     return {
       buttonText: showContactButton ? t_i18n('Contact us') : undefined,
@@ -131,7 +132,7 @@ const LicenseBanner = () => {
     });
   };
 
-  const canContactHub = isNotEmptyField(settings?.xtm_hub_token);
+  const canContactHub = settings?.xtm_hub_registration_status === 'registered';
   const bannerInfo = computeBannerInfo(eeSettings, canContactHub, () => {
     setShowFormDialog(true);
   });
--- a/opencti-platform/opencti-front/src/private/components/LicenseBanner.tsx
+++ b/opencti-platform/opencti-front/src/private/components/LicenseBanner.tsx
@@ -1,15 +1,14 @@
 import moment from 'moment/moment';
 import React, { useContext, useState } from 'react';
 import { graphql } from 'react-relay';
-import Dialog from '@mui/material/Dialog';
-import DialogTitle from '@mui/material/DialogTitle';
-import DialogContent from '@mui/material/DialogContent';
 import DialogActions from '@mui/material/DialogActions';
-import Button from '@mui/material/Button';
+import Typography from '@mui/material/Typography';
+import { SxProps } from '@mui/material';
 import * as Yup from 'yup';
 import { Field, Form, Formik } from 'formik';
 import { FormikConfig } from 'formik/dist/types';
-import { SxProps } from '@mui/material';
+import Button from '@common/button/Button';
+import Dialog from '@common/dialog/Dialog';
 import { useFormatter } from '../../components/i18n';
 import TopBanner, { TopBannerColor } from '../../components/TopBanner';
 import useApiMutation from '../../utils/hooks/useApiMutation';
@@ -17,6 +16,7 @@ import { UserContext } from '../../utils/hooks/useAuth';
 import { daysBetweenDates, now } from '../../utils/Time';
 import { RootSettings$data } from '../__generated__/RootSettings.graphql';
 import TextField from '../../components/TextField';
+import { isNotEmptyField } from '../../utils/utils';
 
 export const LICENSE_OPTION_TRIAL = 'trial';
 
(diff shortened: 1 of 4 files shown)


SamuelHassine:
### Update — second commit (`d3fb6cec`)

While testing the fix on a real running platform I ran into a stronger reproduction of the original symptom: even after going through the **Unregister** flow on the Settings → XTM Hub tab, the trial banner kept showing the **Contact us** button. Live state on the platform was:

```json
{
  "xtm_hub_token": "88ec0c6a-...",  // ← still set
  "xtm_hub_registration_status": "lost_connectivity"        // ← set by a later checkXTMHubConnectivity poll
}
```

So the unregistration handshake (Hub tab → `postMessage({ action: 'unregister' })` → front `handleUnregistration`) had failed silently, leaving the platform with a stale token. The first commit's `isNotEmptyField(settings?.xtm_hub_token)` gate therefore evaluated to `true` and the button stayed visible.

I also manually ran the unregistration mutation against the server and confirmed a second, smaller bug:

```json
{
  "xtm_hub_token": "",                                    // ✓ cleared
  "xtm_hub_registration_status": "unregistered",          // ✓ cleared
  "xtm_hub_registration_user_id": "88ec0c6a-...",         // ✗ NOT cleared
  "xtm_hub_registration_user_name": "admin",              // ✗ NOT cleared
  "xtm_hub_registration_date": "2026-05-28T10:48:33Z",    // ✗ refreshed to NOW
  "xtm_hub_last_connectivity_check": "2026-05-28T10:48:33Z" // ✗ refreshed to NOW
}
```

`completeXTMHubDataForRegistration` in `opencti-graphql/src/utils/settings.helper.ts` was firing during unregistration because its guard (`tokenItem?.value && statusItem?.value`) only checks "are these two keys present", not "do they describe an actual registration". On unregister the front sends both keys with empty / `'unregistered'` values, the guard passes, and the helper re-stamps the registration user / dates as if we were registering.

This commit addresses both:

1. **`LicenseBanner.tsx`** — Switch the button gate from `xtm_hub_token` presence to `xtm_hub_registration_status === 'registered'`. This is the semantically correct check and is robust against stale tokens after a flaky unregistration handshake.
2. **`settings.helper.ts`** — Tighten `completeXTMHubDataForRegistration` to only enrich the input when the token value is non-empty AND status is `'registered'`. Added two unit tests in `tests/01-unit/utils/settings-test.ts`:
   - Empty token + `'unregistered'` status → no enrichment.
   - Empty token + `'registered'` status → no enrichment (defensive).
3. **`XtmHubTab.tsx#handleUnregistration`** — Now also clears `xtm_hub_registration_user_id`, `xtm_hub_registration_user_name` and `xtm_hub_last_connectivity_check`, matching the backend `resetRegistration` (in `domain/xtm-hub.ts`) so the unregister cleanup is fully symmetric, even if the backend helper is bypassed.

> Note: the underlying fragility of the postMessage handshake itself (the reason the user-triggered unregister never reached `handleUnregistration` in the first place) is broader and is being tracked separately — out of scope for this PR.

Local `vitest tests/01-unit/utils/settings-test.ts` → 4 passed.

PR #16269, opened 2026-05-28
feat(frontend): add main color picker to Tree widget

hugodocto:
### Proposed changes

* Add a "Main color" picker to the Tree widget in the dashboard widget create/update modal, under the Parameters step. The picker offers 8 colors taken from the existing palette in `utils/Charts.jsx`: blue, cyan, green, teal, amber, orange, red, purple.
* Store the choice on `WidgetParameters.mainColor` as a palette key. At render time it resolves to shade 600 in light mode and shade 400 in dark mode, matching the convention already used for the distributed palette in the same chart.
* When `mainColor` is unset, the widget keeps falling back to `theme.palette.primary.main`, so existing dashboards render exactly as before.
* The control is gated by the widget's `availableParameters` and is currently enabled only for `tree`. Adding the key to another visualization's `availableParameters` is enough to expose the same picker there later.

### Related issues

* N/A

### Checklist

### Further comments

Storing a palette key instead of a hex value lets the rendered color adapt to the active theme. The same key resolves to a darker shade in light mode and a lighter shade in dark mode through `resolveWidgetMainColor(theme, key)`, mirroring how `colors(theme.palette.mode === 'dark' ? 400 : 600)` is already used for the distributed palette.

When `distributed=true`, ApexCharts paints tiles from the multi-color palette and the main color is ignored. That is the existing behavior and is preserved. The new control sits next to the `distributed` toggle in the Parameters step.

The frontend wiring touches the three private Tree consumers (`StixCoreObjectsTreeMap`, `StixRelationshipsTreeMap`, `AuditsTreeMap`) and the two public-dashboard variants, so the chosen color is honored everywhere a tree is rendered, including on public dashboards. The backend change is limited to exposing the new field on the `WidgetParameters` GraphQL type. The workspace manifest itself is already a JSON blob, so no resolver, migration, or storage change is needed.

<details>
<summary>Screenshots</summary>

Dark theme, `mainColor: "red"` (resolved to `red[400]`):

Tree widget, red, dark theme

Light theme, `mainColor: "green"` (resolved to `green[600]`):

Tree widget, green, light theme

</details>

### DIFF
--- a/opencti-platform/opencti-front/src/components/dashboard/WidgetTree.tsx
+++ b/opencti-platform/opencti-front/src/components/dashboard/WidgetTree.tsx
@@ -3,7 +3,7 @@ import React, { useMemo } from 'react';
 import { useTheme } from '@mui/styles';
 import { ApexOptions } from 'apexcharts';
 import { useFormatter } from '../i18n';
-import { treeMapOptions } from '../../utils/Charts';
+import { resolveWidgetMainColor, treeMapOptions } from '../../utils/Charts';
 import { getMainRepresentative, isFieldForIdentifier } from '../../utils/defaultRepresentatives';
 import { simpleNumberFormat } from '../../utils/Number';
 
@@ -13,6 +13,7 @@ interface WidgetTreeProps {
   groupBy: string;
   onMounted?: OpenCTIChartProps['onMounted'];
   isDistributed?: boolean;
+  mainColor?: string | null;
 }
 
 const WidgetTree = ({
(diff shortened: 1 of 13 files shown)


aHenryJard:
Slack discussion on this PR can be found here: https://filigran-community.slack.com/archives/C06CC3T5Z37/p1781858275846469

RBourgeoisGH:
Hi,
Thanks for your contribution, we noticed that some tests have been failing for a while (and probably need to be rebased now), is it something you'll want to fix?

Thanks

PR #16284, opened 2026-05-29
[backend] Allow users to exit a draft they cannot edit (#16273)

### LINKED ISSUE #16273: [Draft Workspace] Users trapped in draft context, cannot exit
### Description

When a user navigates into a Draft Workspace created by another user (or a service account), they become trapped in the draft context and cannot exit. Clicking the **"Exit draft"** button returns the error:

> **"You are not allowed to do this."**

The "Exit draft" action is a navigation function — it should not be gated by any permission check. Additionally, the draft context persists in the user session, so manually navigating to the dashboard URL (e.g. `/dashboard`) redirects the user back into the draft view. Users cannot escape the draft without clearing all browser site data or using a different browser/incognito session.

### Environment

- **OpenCTI version**: 7.260522
- **Deployment**: On-premise (Kubernetes)
- **Edition**: Enterprise

### Steps to Reproduce

1. **User A** (or a service account) creates a Draft Workspace via API (e.g. data migration import using `draftWorkspaceAdd` or bundle upload to a draft context).
2. **User B** (a different user with `KNOWLEDGE_KNUPDATE` and Create/Update knowledge capabilities, including draft-specific overrides) navigates to the draft URL in their browser.
3. User B can see the draft contents (entities, observables, containers).
4. User B clicks the **"Exit draft"** button.
5. ❌ Error: **"You are not allowed to do this."**
6. User B attempts to navigate directly to `/dashboard` via the browser address bar.
7. ❌ The browser redirects back into the draft context — the user remains trapped.

### Expected Behaviour

- **"Exit draft"** should navigate the user back to the main knowledge base without any permission check — it is not a mutating action.
- Navigating to `/dashboard` should always exit the draft context and return the user to the live platform.
- Non-owner users should either:
  - (a) Be prevented from entering the draft view in the first place, OR
  - (b) Be able to freely exit the draft view once inside

### Actual Behaviour

- "Exit draft" is blocked by a permission/ownership check, trapping non-owner users.
- The draft context is session-sticky — it persists even when the user manually navigates away via the URL bar.
- The only workaround is to clear all site data (cookies, localStorage, sessionStorage) or use an incognito window.

### Impact

- **All users** who navigate into a draft they do not own become locked out of the entire platform.
- In environments where data migrations are being tested (draft created via API with large datasets), this can block an entire team from using the platform.
- Severity is high in pre-production/go-live scenarios where multiple team members need concurrent access.

### Additional Context

- The draft in question was created via API (CI/CD pipeline scripts), not through the UI.
- The draft contained ~28,500 entities, ~3,700 observables, and ~1,900 containers.
- The affected users have full `KNOWLEDGE_KNUPDATE` capabilities and draft-specific capability overrides.
- The issue affects multiple users simultaneously — anyone who navigates into the draft URL is trapped.

SamuelHassine:
### Proposed changes

* Fix the `Exit draft` action returning `You are not allowed to do this.` and trapping users in a draft context. Entering/exiting a draft only updates the user's own `draft_context` (a navigation action on the user entity, which is never part of the draft data). `userEditField` now performs the `draft_context` update **outside** of any draft context, so it is no longer gated by the draft authorized-members access check.
* Make `checkDraftInContext` self-heal: when a user is stuck in a draft they cannot access, their `draft_context` is cleared so they are no longer permanently trapped (e.g. when manually navigating to `/dashboard`).

### Related issues
* closes #16273

### How to test this PR

1. As an admin / via API / service account, create a draft workspace with restricted authorized members (e.g. give another user `User B` only `can view` access — or no access at all).
2. As `User B`, open the draft URL (e.g. `/dashboard/data/import/draft/<draftId>`). You are now in the draft context.
3. Click **Exit draft**.
4. Before this fix: the request fails with `You are not allowed to do this.` and navigating to `/dashboard` keeps redirecting back into the draft.
5. After this fix: `User B` exits the draft and returns to the live platform; the draft context is cleared from the session.

Automated coverage: `opencti-platform/opencti-graphql/tests/03-integration/01-database/draftWorkspace-exit-test.ts`
* a view-only user can exit a draft they cannot edit
* a user stuck in a draft they cannot access gets their `draft_context` self-healed.

### Checklist

### Further comments

The root cause was in `updateAttribute`: any update performed while in a draft context is checked against the draft's authorized members (`validateUserAccessOperation`). Since exiting a draft is itself an update to the user entity performed while the session is still in the draft context, a user without `edit` access to the draft (created by another user or a service account) could never clear their own `draft_context`. The fix scopes the bypass strictly to `draft_context` edits so all other self-edits keep their existing behaviour.

### DIFF
--- a/opencti-platform/opencti-graphql/src/domain/user.js
+++ b/opencti-platform/opencti-graphql/src/domain/user.js
@@ -997,7 +997,14 @@ export const userEditField = async (context, user, userId, rawInputs) => {
       inputs.push(input);
     }
   }
-  const { element } = await updateAttribute(context, user, userId, ENTITY_TYPE_USER, inputs);
+  // Editing the draft context (entering/exiting a draft) is a navigation action performed on the
+  // user's own entity, which is never part of the draft data. It must run outside of any draft
+  // context, otherwise a user who entered a draft they cannot edit would be blocked by the draft
+  // authorized-members access check and would be unable to exit it (see issue #16273).
+  const isDraftContextEdit = inputs.some((i) => i.key === 'draft_context');
+  const editContext = isDraftContextEdit ? { ...context, draft_context: undefined } : context;
+  const editUser = isDraftContextEdit ? { ...user, draft_context: undefined } : user;
+  const { element } = await updateAttribute(editContext, editUser, userId, ENTITY_TYPE_USER, inputs);
   const input = updatedInputsToData(element, inputs);
   const personalUpdate = user.id === userId;
   const actionEmail = ENABLED_DEMO_MODE ? REDACTED_USER.user_email : element.user_email;
--- a/opencti-platform/opencti-graphql/src/http/httpServer-draft.ts
+++ b/opencti-platform/opencti-graphql/src/http/httpServer-draft.ts
@@ -26,6 +26,13 @@ export const checkDraftInContext = async (executeContext: AuthContext) => {
 
       const isUserCanAccess = await isUserCanAccessStoreElement(executeContext, executeContext.user, draftWorkspace);
       if (!isUserCanAccess) {
+        if (executeContext.user.draft_context === executeContext.draft_context) {
+          // If user is stuck in a draft they cannot access, remove draft context from user so they are not trapped
+          await userEditField(executeContext, executeContext.user, executeContext.user.id, [{
+            key: 'draft_context',
+            value: '',
+          }]);
+        }
         throw FunctionalError(`Draft ${executeContext.draft_context} cannot be found`);
       }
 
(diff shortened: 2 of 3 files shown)


OctaveLaventure (approved):

SouadHadjiat (approved):
Tested locally, draft exit is now possible for user with "view-only" rights. 
I however don't think this is the right fix in the long term, if you try for example to update any other field of the user while in draft, we still throw "forbidden error" (which is not the case if the user has edit permissions on the draft, it's stays forbidden only for view-only user, even if the user is editing its own data via meEdit).

I think the part of the code that will need to be changed is here: 
Image

Approving since it's a critical bug.

labo-flg:
Merging to fix the critical bug rapidly, thanks !
@SamuelHassine Let's have a follow-up PR to fix this side effect if needed

PR #16288, opened 2026-05-29
fix(frontend/backend): refresh concerned users after organization update (#16195)

### LINKED ISSUE #16195: fix(dashboards): defining a default dashboard does not apply to already-created users until the user is refreshed
## Description

When a user is already created and at organisation level a dashboard is set as default, this user won't see on next login the default dashboard, unless the user refreshes the user by for instance updating the description of the user.

## Environment

Latest

## Reproducible Steps

1. Create a custom dasboard.
2. Add organisations in Authorize members
3. Go into the org that you have added as AM, and create a user that pertains to this org (user needs to have the access dashboard capability) 
4. Set the dashboard as default dashboard for this org 
5. Login with this user 
## Expected Output

User see the dashboard set by defaulr 

## Current output

User see the fallback default dashboard (the one by default when you launch octi for the first time)

## Note

When either the admin (or the user itself by going into user profile) updates the description, user is refreshed leading. As a result, the user sees the correct dashboard.

This means that any user pertaining to an org, if a default dashboard is set at org level, all the users should be refreshed in order to see the newly st default dashboard so that no manual intervention is needed to have this functionality working properly.

Flagging this as a regression since we never had any issue before on this, making me think that it used to work before (might be a mistake though)

Archidoit:
### Proposed changes
When an organization is updated, refresh the users belonging to this organization.
Example: when an organization default dashboard is modified, the users belonging to the organization should see the new default dashboard. Before, it wasn't the case, except if the admin modified the user (like the user description), to generate a user refresh.

Note that the user needs to reload the page to see the changes of dashboard.

Also add backend tests and put organization-test to ts

### Related issues
fixes #16195


opencti-platform/opencti-graphql/tests/utils/syncCountHelper.ts
```diff
@@ -68,7 +68,7 @@ export const testUpdatedCounter: Record<string, number> = {};
 testUpdatedCounter['marking-definition'] = 2;
 testUpdatedCounter.relationship = 8;
 testUpdatedCounter.campaign = 7;
-testUpdatedCounter.identity = 26;
+testUpdatedCounter.identity = 29;
```
Archidoit, line 71: (because of the tests added in organization-test)

opencti-platform/opencti-graphql/src/domain/group.js
```diff
@@ -23,9 +23,15 @@ import { cleanMarkings } from '../utils/markingDefinition-utils';
...
 export const GROUP_DEFAULT = 'Default';
 
+/**
+ * Refreshes the user cache for all members belonging to the given group.
+ * This ensures user sessions are reloaded with up-to-date group permissions.
+ */
 const groupUsersCacheRefresh = async (context, user, groupId) => {
   const members = await fullEntitiesThroughRelationsFromList(context, user, groupId, RELATION_MEMBER_OF, ENTITY_TYPE_USER);
-  await notify(BUS_TOPICS[ENTITY_TYPE_USER].EDIT_TOPIC, members, user);
+  if (members.length > 0) {
```
lndrtrbn: I am wondering why you do not want to refresh the cache when the array is empty?

opencti-platform/opencti-graphql/src/modules/organization/organization-domain.ts
```diff
@@ -53,6 +53,35 @@ export const editAuthorizedAuthorities = async (context: AuthContext, user: Auth
...
+
+export const organizationEditField = async (
+  context: AuthContext,
+  user: AuthUser,
+  organizationId: string,
+  input: EditInput[],
+  opts = {},
+) => {
+  const result = await stixDomainObjectEditField(context, user, organizationId, input, opts);
+  await organizationUsersCacheRefresh(context, user, organizationId);
```
esrevi, line 86: I think we should wrap this function in try/catch because if it fails, the field would have been changed, but organizationEditField will throw error because of the function refresh error

suggestion:
```
try {
await organizationUsersCacheRefresh(context, user, organizationId);
} catch {
logger.warn("refresh cache fail")
}
 ```

opencti-platform/opencti-graphql/src/domain/group.js
esrevi: I think you are right, if the last member is removed from the list, the cache will not be updated, maybe resync cache in delete function after removing one member?

Archidoit: indeed you're right ! I'm gonna remove this 'if'

opencti-platform/opencti-graphql/src/modules/organization/organization-domain.ts
Archidoit: done :)

esrevi (approved):

PR #15500, opened 2026-04-14
[backend] Upgrade opentelemetry deps to new major version (#16065)

fellowseb:
### Proposed changes

Followed the migration guide and CHANGELOGs:

* https://github.com/open-telemetry/opentelemetry-js/blob/main/doc/upgrade-to-2.x.md
* https://github.com/open-telemetry/opentelemetry-js/blob/36ce569862378f26d805ed4858d07a1e8ffd0e32/CHANGELOG.md?plain=1#L816
* https://github.com/open-telemetry/opentelemetry-js/blob/36ce569862378f26d805ed4858d07a1e8ffd0e32/experimental/CHANGELOG.md

### Related issues

* Fixes https://github.com/OpenCTI-Platform/opencti/issues/16065

* took stuff from https://github.com/OpenCTI-Platform/opencti/pull/12921

### Testing instructions

To test this PR do the following:

1. Tweak your `development.json` to have the following settings under `app` entry:
```
    "telemetry": {
      "pyroscope": {
        "enabled": true,
        "identifier": "OpenCTI",
        "exporter": "http://localhost:4040"
      },
      "tracing": {
        "enabled": true,
        "exporter_otlp": "http://localhost:4318/v1/traces"
      },
      "metrics": {
        "enabled": true,
        "exporter_otlp": "",
        "exporter_prometheus": 14269
      }
    },
```

2. Bring up the `telemetry`-related dependencies:
```
podman compose --profile telemetry up -d
```

3. Use OpenCTI in the app for a few seconds to generate telemetry data

4. Ensure `Pyroscope` is working by opening it locally on `http://localhost:4040/`. Select the `OpenCTI` service:

<details> <summary>Show screenshot</summary>
image
</details>

5. Ensure the `tracing` tool is still functional (Jaeger UI) on `http://localhost:16686/`. Select the `opencti-platform` service in the left panel. (The UI is ugly though)
<details> <summary>Show screenshot</summary>
image
</details>

6. Ensure Grafana (& Prometheus work), by displaying `technical metrics`:
- Open grafana on `http://localhost:3100/`, enter `admin/admin` as credentials (skip password update), and add a new Data Source in `Menu > Connections > Data sources`. Use `http://opencti-dev-telemetry-prometheus:9090` as Prometheus server URL and Save/test.
- Play with the open CTI app for a minute or so
- In Grafana show all metrics in `Drilldown > Metrics > All metrics`. You should see `nodej_`metrics and `opencti_api` metrics
[...]

PR #16330, opened 2026-06-01
chore(telemetry): improve queue to connector correlation (#16315)

### LINKED ISSUE #16315: refactor: use a clearer queue name to identify the corresponding creator
# Symptoms / detection

With composer and managed connector it's really hard to know the source that is causing ingestion stall in rabbit.

Would be great to rework queue names to have the start name of connectors. SAme for playbook and other queues that are just an uiid.

# Analysis / explanation

Image

# Resolution brainstorm

# GitHub links

fellowseb:
### Proposed changes

* Emit a new `opencti_connector_identity_info` metric from the connectors that contain following info:

```
opencti_connector_identity_info{
id="e1e82feb-4b0b-4358-89ca-31d2b627e6db",
name="OpenCTI Datasets",
scope="marking-definition,identity,location",
type="EXTERNAL_IMPORT"
} 1.0
```

This metric of type `Info` will be helpful to write Prometheus `recording rules`.

* Continues to enrich the local telemetry setup 
* Creates 3 recording records to enrich `rabbitmq` metrics with `connector name` and `connector type` info by doing joins resulting in 3 new "derived" metrics: `opencti:rabbitmq_detailed_queue_messages_persistent:enriched`, `opencti:rabbitmq_detailed_queue_consumers:enriched` and `opencti:rabbitmq_queue_messages_acked_total:enriched`.
These serve as a proof of concept before going to prod.

### Related issues

* Fixes #16315 

### How to test this PR

* Launch the local telemetry stack: `podman compose --profile telemetry up -d`
* Launch a OCTI instance with a connector and a worker
* Navigate to the local grafana instance on `http://localhost:3100/`; go to "Metrics > Drilldown" or "Explore" and checkout the new metrics.

### Checklist

### DIFF
--- a/client-python/pycti/connector/opencti_metric_handler.py
+++ b/client-python/pycti/connector/opencti_metric_handler.py
@@ -6,14 +6,14 @@ allowing monitoring of connector performance and health.
 
 from typing import Type, Union
 
-from prometheus_client import Counter, Enum, start_http_server
+from prometheus_client import Counter, Enum, Info, start_http_server
 
 
 class OpenCTIMetricHandler:
     """Handler for Prometheus metrics in OpenCTI connectors.
 
     This class manages Prometheus metrics for monitoring connector behavior,
-    including bundle sends, records processed, run counts, API pings, and errors.
+    including general info, bundle sends, records processed, run counts, API pings, and errors.
 
     When activated, it starts an HTTP server to expose metrics for scraping
     by Prometheus or compatible monitoring systems.
@@ -117,10 +117,16 @@ class OpenCTIMetricHandler:
                     namespace=namespace,
                     subsystem=subsystem,
                 ),
+                "identity": Info(
+                    "identity",
+                    "Identity and configuration of the connector",
+                    namespace=namespace,
+                    subsystem=subsystem,
+                ),
             }
 
     def _metric_exists(
-        self, name: str, expected_type: Union[Type[Counter], Type[Enum]]
+        self, name: str, expected_type: Union[Type[Counter], Type[Enum], Type[Info]]
     ) -> bool:
         """Check if a metric exists and has the expected type.
 
@@ -129,8 +135,8 @@ class OpenCTIMetricHandler:
 
         :param name: Name of the metric to validate
         :type name: str
-        :param expected_type: Expected Prometheus metric type (Counter or Enum)
-        :type expected_type: Union[Type[Counter], Type[Enum]]
+        :param expected_type: Expected Prometheus metric type (Counter, Enum or Info)
+        :type expected_type: Union[Type[Counter], Type[Enum], Type[Info]]
 
         :return: True if metric exists and has correct type, False otherwise
         :rtype: bool
@@ -198,3 +204,20 @@ class OpenCTIMetricHandler:
         if self.activated:
             if self._metric_exists(name, Enum):
                 self._metrics[name].state(state)
+
+    def set_info(
+        self,
+        connector_id: str,
+        connector_name: str,
+        connector_type: str,
+        connector_scope: str,
+    ) -> None:
+        if self.activated:
+            self._metrics["identity"].info(
+                {
+                    "id": connector_id,
+                    "name": connector_name,
+                    "type": connector_type,
+                    "scope": connector_scope,
+                }
+            )
--- a/opencti-platform/opencti-dev/grafana/datasources.yaml
+++ b/opencti-platform/opencti-dev/grafana/datasources.yaml
@@ -1,0 +1,11 @@
+apiVersion: 1
+datasources:
+  - name: OpenCTI Dev Prometheus
+    type: prometheus
+    access: proxy
+    url: http://opencti-dev-telemetry-prometheus:9090
+    isDefault: true
+    editable: true
+    jsonData:
+      httpMethod: POST
+      timeInterval: "15s"
--- a/opencti-platform/opencti-dev/prometheus/prometheus.yml
+++ b/opencti-platform/opencti-dev/prometheus/prometheus.yml
@@ -1,0 +1,25 @@
+global:
+  scrape_interval: "15s"
+
+scrape_configs:
+  - job_name: 'opencti-dev-telemetry-prometheus'
+    static_configs:
+      - targets: ['host.docker.internal:14269']
+        labels:
+          app: "opencti"
+  - job_name: "rabbitmq"
+    metrics_path: "/metrics/detailed"
+    params:
+      family: ["queue_coarse_metrics"]
+    static_configs:
+      - targets: ["host.docker.internal:15692"]
+        labels:
+          app: 'opencti'
+  - job_name: "connectors"
+    static_configs:
+      - targets: ["host.docker.internal:9095"]
+        labels:
+          app: "opencti"
+
+rule_files:
+  - "/etc/prometheus/rules/*.yml"
--- a/opencti-platform/opencti-dev/prometheus/rules/opencti_rabbitmq_enriched.yml
+++ b/opencti-platform/opencti-dev/prometheus/rules/opencti_rabbitmq_enriched.yml
@@ -1,0 +1,55 @@
+groups:
+  - name: opencti_rabbitmq_enriched
+    interval: 15s
+    rules:
+      - record: opencti:rabbitmq_detailed_queue_messages_persistent:enriched
+        expr: |
+          label_replace(
+            rabbitmq_detailed_queue_messages_persistent,
+            "connector_id",
+            "$3",
+            "queue",
+            "(.*)(listen|push)_([a-f0-9-]{36}).*"
+          )
+          * on(connector_id) group_left(name, type)
+          label_replace(
+            opencti_connector_identity_info,
+            "connector_id",
+            "$1",
+            "id",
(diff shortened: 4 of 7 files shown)


opencti-platform/opencti-dev/grafana/datasources.yaml
lndrtrbn, line 1: To help me understand, what this file used for?

opencti-platform/opencti-dev/prometheus/rules/opencti_rabbitmq_enriched.yml
lndrtrbn, line 1: To help me understand, what this file used for?

opencti-platform/opencti-dev/prometheus/prometheus.yml
lndrtrbn, line 1: To help me understand, what this file used for?

opencti-platform/opencti-dev/grafana/datasources.yaml
fellowseb: This is a Grafana config file used locally in dev when using the `podman compose --profile telemetry up -d` command in order to connect to the local Prometheus instance automatically at startup. `opencti-dev-telemetry-prometheus` is the name of the Prometheus service as defined in the Docker Compose file. Having this stack set up allows seeing how metrics, logs, traces and dashboards look like locally before setting them up in prod.

opencti-platform/opencti-dev/prometheus/rules/opencti_rabbitmq_enriched.yml
fellowseb: This is a  Prometheus config file used locally in dev when using the `podman compose --profile telemetry up -d`. More specifically it's defining three `recording rules`. A recording rule allows creating a derived metric from existing metrics.

In this case for instance it's defining a new `opencti:rabbitmq_detailed_queue_messages_persistent:enriched` by performing joins (just like JOIN statements in SQL) on two other metrics: `rabbitmq_detailed_queue_messages_persistent` a RabbitMQ metric used in current on-call dashboards, and `opencti_connector_identity_info` a new metric I've introduced in this PR to emit basic information about Connectors (name, type, scope).

Thanks to this we are able to display the name and type of the connector linked to a RabbitMQ queue in Dashboards.

opencti-platform/opencti-dev/prometheus/prometheus.yml
fellowseb: This is a Prometheus config file used locally in dev when using the `podman compose --profile telemetry up -d` command in order to tell Prometheus what endpoints it should scrape and how often. (i.e. the `targets`).

`host.docker.internal:14269` is the local address where the metric from the backend service (`opencti-graphql`) are emitted. If you wanna see this yourself follow the instructions in [this PR description](https://github.com/OpenCTI-Platform/opencti/pull/15500) and open a browser page on `http://localhost:14269`.

`host.docker.internal:15692` is the local address where the metric from the RabbitMQ queues are emitted.

`host.docker.internal:9095` is the local address where the metric from the Connector are emitted.


 `host.docker.internal` is used instead of `localhost` as we want to contact the host from within a container.

client-python/pycti/connector/opencti_metric_handler.py
```diff
@@ -198,3 +204,20 @@ def state(self, state: str, name: str = "state") -> None:
...
+
+    def set_info(
+        self,
+        connector_id: str,
+        connector_name: str,
+        connector_type: str,
+        connector_scope: str,
+    ) -> None:
+        if self.activated:
+            self._metrics["identity"].info(
```
T0t1n, line 216: Unlike inc / state, it accesses `self._metrics["identity"]` directly instead of going through `_metric_exists`. Minor, just for consistency: if `self._metric_exists("identity", Info)`

opencti-platform/opencti-dev/prometheus/prometheus.yml
```diff
@@ -0,0 +1,25 @@
...
+    metrics_path: "/metrics/detailed"
+    params:
+      family: ["queue_coarse_metrics"]
+    static_configs:
+      - targets: ["host.docker.internal:15692"]
+        labels:
+          app: 'opencti'
+  - job_name: "connectors"
+    static_configs:
+      - targets: ["host.docker.internal:9095"]
```
T0t1n, line 20: So, not sure I fully grok the whole stack yet, so take this with as a "rookie comment".
The connectors job points at a single static target host.docker.internal:9095. It won't stop anything from starting, but since connectors all default to port 9095, only one is actually scrapable at a time locally. Fine for a single connector, but the multi-connector case is exactly what #16315 is about, so I wanted to flag it in case we want the local setup to reflect that.

opencti-platform/opencti-dev/prometheus/prometheus.yml
```diff
@@ -0,0 +1,25 @@
...
+scrape_configs:
+  - job_name: 'opencti-dev-telemetry-prometheus'
+    static_configs:
+      - targets: ['host.docker.internal:14269']
+        labels:
+          app: "opencti"
+  - job_name: "rabbitmq"
+    metrics_path: "/metrics/detailed"
+    params:
+      family: ["queue_coarse_metrics"]
```
T0t1n, line 13: Heads up: the rabbitmq job only scrapes family: `["queue_coarse_metrics"]`, but `rabbitmq_detailed_queue_messages_persistent` and `rabbitmq_detailed_queue_consumers` (used by rules 1 & 2) live in the `queue_metrics` family. So those two series never make it into Prometheus and the :enriched rules stay empty.

Fix that worked for me locally:
   ```
 params:
      family: ["queue_coarse_metrics", "queue_metrics"]
```

After that, both `opencti:rabbitmq_detailed_queue_messages_persistent:enriched` and `opencti:rabbitmq_detailed_queue_consumers:enriched` show up, correctly joined with connector_id / name / type.

File: opencti-platform/opencti-dev/prometheus/rules/opencti_rabbitmq_enriched.yml [hunk 4]
```diff
@@ -0,0 +1,55 @@
...
+            opencti_connector_identity_info,
+            "connector_id",
+            "$1",
+            "id",
+            "(.*)"
+          )
+      - record: opencti:rabbitmq_queue_messages_acked_total:enriched
+        expr: |
+          label_replace(
+            rabbitmq_queue_messages_acked_total,
```
T0t1n, line 42: This rule doesn't produce anything, and it took me a while to see why. 

`rabbitmq_queue_messages_acked_total` (no `detailed_` prefix, unlike the other two) only exists on the aggregated `/metrics` endpoint, and only if `prometheus.return_per_object_metrics = true` is set in RabbitMQ's config (the `rabbitmq.config`) — there's no `detailed_` variant, so it's never scraped here. Fixing it means either finding a different source, or switching the whole setup to the per-object aggregated endpoint (which also means renaming the metrics in rules 1 & 2). 

That said, per-object aggregated metrics get expensive in prod with lots of queues — do we actually need acked?

T0t1n (commented):
So here we're getting pycti to push the new metric info, and we've got what we need to test it in dev — but what about the prod env config? The scrape setup + recording rules only land in opencti-dev, I don't see anything matching on the deployment side. How do we want to handle *prod* side of this configuration ?

client-python/pycti/connector/opencti_metric_handler.py
fellowseb: Yes as I didn't add a `name` parameter for simplicity I figured using `_metric_exists` wasn't needed.
I can generalize and add a `name` param though for consistency.

opencti-platform/opencti-dev/prometheus/rules/opencti_rabbitmq_enriched.yml
fellowseb: I focused my work on metrics that are used in our main Dashboard in prod: "OpenCTI Product Health". This dashboards has 3 visualizations (Under the Ingestion and RabbitMQ sections) that use rabbitmq metrics:
- `rabbitmq_queue_messages_acked_total`
- `rabbitmq_queue_messages_ready`
- `rabbitmq_detailed_queue_messages_persistent`
- `rabbitmq_detailed_queue_consumers`

That would mean that in prod we scrap both `/metrics` and `/metrics/detailed` , right ? I 'm missing the scrapping of the `/metrics` endpoint in this PR for dev. I'll try it out.

opencti-platform/opencti-dev/prometheus/prometheus.yml
fellowseb: Ah thanks, I must've missed it when switched from non-detailed metrics to detailed metrics 👍

fellowseb:
> So here we're getting pycti to push the new metric info, and we've got what we need to test it in dev — but what about the prod env config? The scrape setup + recording rules only land in opencti-dev, I don't see anything matching on the deployment side. How do we want to handle _prod_ side of this configuration ?

Indeed the idea of this PR is to set up a dev environment where we can experiment which metrics and recording rules.
For the prod side of things I opened a ticket and PR for adding a recording rule here: https://github.com/FiligranHQ/platform/issues/2714 & https://github.com/FiligranHQ/platform/pull/2716 .

PR #16351, opened 2026-06-02
docs(retention): add History and Activity scopes, active/inactive toggle (#15669)

### LINKED ISSUE #15669: Retention policies: active/inactive
## Use case

Some retention policies are more "technical" in the sense that they don't have filtrs (workfbench, files).
As a result, nabling multiple of them does not make sense.

Given we're introducing policies on History & Activty too that would work the same, the goal of the feature is to introduce a boolean active/inactive, so that these technical rules will always be present & the user would only be able to create only rules about knowledge with specific filters. 

## Current Workaround

None
## Proposed Solution

- Introduce a an active/inactive boolean
- When clicking on create a rule, only knowledge should be present.
- Activate/Deactivate should be available in the drawer & as a quick action
- The table should show the state (Active/inactive) of the rule in the rule list.
- Telemetry for technical rules (active)
- Migration
   - Every rule defined on all platforms should be set to active
   - All platform not having the rule Workbench or File should have it created, but disabled
   - Retention period for all rules created but disabled should be 365 days.

- New platforms
   - All new platforms should have the rules Workbench, Files, History & Activity created but disabled 
   - Retention period for all rules created but disabled should be 365 days.

CelineSebe:
### Proposed changes

***2 new scopes:**
- **History** — applies the retention rule on the history logs of knowledge entities
- **Activity** — applies the retention rule on the platform activity/audit logs

**Active / Inactive toggle:**
- Each retention rule now has an \`active\` boolean field
- A rule can be disabled without being deleted, allowing temporary suspension
- History and Activity rules are created as \`inactive\` by default to prevent accidental deletion of audit logs

### Changes to \`retentions.md\`

- Updated the Configuration section to list all 5 parameters including \`active\` and \`unit\`
- Expanded the Scopes section from 3 to 5 scopes (added History and Activity)
- Added a new **Enabling and disabling a rule** section explaining the \`active\` toggle use cases and the default-inactive behavior for History/Activity
- Added a new **History and Activity scopes**

### Related issues

* resolves #15669

### Checklist


docs/docs/administration/retentions.md
```diff
@@ -8,24 +8,55 @@ Note that the data deleted by an active retention policy will not appear in the
...
 ![Retention policy parameters](./assets/retention-policy-parameters.png)
 
 ## Scopes
 
-There are 3 possible scopes for a retention rule:
+There are 5 possible scopes for a retention rule:
+
+- **Knowledge**: The rule concerns all the entities of the platform. Users can define filters to target specific objects. Any object respecting the specified filters and that haven't been updated within the maximum retention duration will be permanently deleted.
+
+- **File**: The rule is applied on the global files (i.e. contained in Data > Import) that have been correctly uploaded (upload status is 'complete') and whose eventual imports have all been correctly completed. Such files are permanently deleted if they has been uploaded for a longer duration then the maximum retention duration.
```
nino-filigran: **File**: The rule is applied on the global files (i.e. contained in Data > Import) that have been correctly uploaded (upload status is 'complete') and whose eventual imports have all been correctly completed. Such files are permanently deleted if they has been uploaded for a longer duration then the maximum retention duration. The rule is created by default and set to Inactive if you had not created it before.

docs/docs/administration/retentions.md
```diff
@@ -8,24 +8,55 @@ Note that the data deleted by an active retention policy will not appear in the
...
+
+- **Workbench**: The rule is applied on the global workbenches (i.e. contained in Data > Import). The global workbenches that hasn't been updated for the maximum retention duration will be permanently deleted.
```
nino-filigran: - **Workbench**: The rule is applied on the global workbenches (i.e. contained in Data > Import). The global workbenches that hasn't been updated for the maximum retention duration will be permanently deleted. The rule is created by default and set to Inactive if you had not created it before.

docs/docs/administration/retentions.md
```diff
@@ -8,24 +8,55 @@ Note that the data deleted by an active retention policy will not appear in the
...
+
+- **History**: The rule is applied on the history logs of all knowledge entities. History log entries older than the maximum retention duration are permanently deleted.
```
nino-filigran: - **History**: The rule is applied on the history logs of all knowledge entities. History log entries older than the maximum retention duration are permanently deleted. The rule is created by default and set to Inactive if you had not created it before.

docs/docs/administration/retentions.md
```diff
@@ -8,24 +8,55 @@ Note that the data deleted by an active retention policy will not appear in the
...
+
+- **Activity**: The rule is applied on the platform activity logs (administration events such as login, logout, and security actions). Activity log entries older than the maximum retention duration are permanently deleted.
```
nino-filigran: - **Activity**: The rule is applied on the platform activity logs (administration events such as login, logout, and security actions). Activity log entries older than the maximum retention duration are permanently deleted. The rule is created by default and set to Inactive if you had not created it before.

nino-filigran (changes requested):

nino-filigran (approved):

PR #16366, opened 2026-06-03
[frontend/backend] Deleting a draft associated to the correct capability (#16309)

### LINKED ISSUE #16309: Delete draft linked to the wrong capability
## Description

It is possible to delete a draft with a read only access right. It should only be restricted to delete.

## Environment

4. Other environment details:

## Reproducible Steps

1. Have a user with only access knowledge
2. Go in the list of draft & select a draft 
3. Click on delete

## Expected Output

Delete should not be possible (option not enabled)
## Actual Output

Delete is possible.
IT should be tight to "Delete Knowledge" capability.

To clarify even more: 
- a user that has edit or manage on a draft via authorize members but no Delete should not be able to Delete a draft
- a user that has edit or manage on a draft via authorize members AND has Delete should be able to Delete a draft
- a user that has view only on a draft  AND has Delete should NOT be able to Delete a draft
- a user that only has delete as a capa in draft but not main should NOT be able to delete a draft.

Archidoit:
### Proposed changes
To delete a draft, the user should have the 'delete knowledge' capability
- Frontend fixes: The 'delete' menu item should not be displayed in the popover menu if the user has not the capability
- Graphql fixes: correct capa for the query
- Backend tests

### Related issues
fixes #16309

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/drafts/DraftPopover.tsx
+++ b/opencti-platform/opencti-front/src/private/components/drafts/DraftPopover.tsx
@@ -20,7 +20,7 @@ import { useGetCurrentUserAccessRight } from '../../../utils/authorizedMembers';
 import stopEvent from '../../../utils/domEvent';
 import useApiMutation from '../../../utils/hooks/useApiMutation';
 import useDeletion from '../../../utils/hooks/useDeletion';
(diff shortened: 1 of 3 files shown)


Archidoit:
Closing because duplicate with : https://github.com/OpenCTI-Platform/opencti/pull/16342

PR #16396, opened 2026-06-04
feat(pycti): add audit and log read functionality (#16395)

### LINKED ISSUE #16395: feat(client-python): add ability to read audit and log entities
### Use Case

To support reading the audit and log entries from a connector.

### Current Workaround

None

### Proposed Solution

Add audit and log entities to pycti with read only functionality

### Additional Information

pycti audit and log should only perform read functions. Only internal operations should be able to modify logs and audits. 

### If the feature request is approved, would you be willing to submit a PR?

Yes

scarletmerlin123:
### Proposed changes

* Add audit and log read functionality to pycti

### Related issues

* closes #16395 

### How to test this PR

Using the api, query for a list or individual logs or audits.

### Checklist

### Further comments

Chose to restrict audit and log functionality to read only. External applications should not be able to modify the history in the system directly.

### DIFF
--- a/client-python/pycti/entities/opencti_audit.py
+++ b/client-python/pycti/entities/opencti_audit.py
@@ -1,0 +1,182 @@
+from typing import Dict, List, Optional
+
+
+class Audit:
+    """Main Audit class for OpenCTI
+
+    Reads audits in the OpenCTI platform.
+
+    :param opencti: instance of :py:class:`~pycti.api.opencti_api_client.OpenCTIApiClient`
+    :type opencti: OpenCTIApiClient
+    """
+
+    def __init__(self, opencti):
+        """Initialize the Audit instance.
+
+        :param opencti: OpenCTI API client instance
+        :type opencti: OpenCTIApiClient
+        """
+        self.opencti = opencti
+        self.properties = """
+
+        """
+
+        self.properties = """
+            id
+            entity_type
+            event_scope
+            event_status
+            timestamp
+            user_id
+            raw_data
+            context_uri
+            user_metadata
+        """
+
+    def list(self, **kwargs) -> List[Dict]:
+        """List Audit objects.
+
+        :param filters: the filters to apply
+        :type filters: dict
+        :param search: the search keyword
+        :type search: str
+        :param first: return the first n rows from the after ID (or the beginning if not set)
+        :type first: int
+        :param after: ID of the first row for pagination
+        :type after: str
+        :param orderBy: field to order results by
+        :type orderBy: str
+        :param orderMode: ordering mode (asc/desc)
+        :type orderMode: str
+        :param customAttributes: custom attributes to return
+        :type customAttributes: str
+        :param getAll: whether to retrieve all results
+        :type getAll: bool
+        :param withPagination: whether to include pagination info
+        :type withPagination: bool
+        :return: List of Task objects
+        :rtype: list
+        """
+        first = kwargs.get("first", 500)
+        after = kwargs.get("after", None)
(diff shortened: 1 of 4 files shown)


scarletmerlin123:
I left off tests because I wasn't sure how to add while keeping it read-only.

client-python/pycti/entities/opencti_audit.py
```diff
@@ -0,0 +1,184 @@
...
+
+    def __init__(self, opencti):
+        """Initialize the Audit instance.
+
+        :param opencti: OpenCTI API client instance
+        :type opencti: OpenCTIApiClient
+        """
+        self.opencti = opencti
+        self.properties = """
+
```
marieflorescontact: this self.properties definition is will be  overwritten by the next one with the list of fields :)

client-python/pycti/entities/opencti_log.py
```diff
@@ -0,0 +1,184 @@
...
+    """
+
+    def __init__(self, opencti):
+        """Initialize the Log instance.
+
+        :param opencti: OpenCTI API client instance
+        :type opencti: OpenCTIApiClient
+        """
+        self.opencti = opencti
+        self.properties = """
```
marieflorescontact: same remark as for opencti_audit.py :)

client-python/pycti/entities/opencti_audit.py
```diff
@@ -0,0 +1,184 @@
...
+        """
+        id = kwargs.get("id", None)
+        custom_attributes = kwargs.get("customAttributes", None)
+        filters = kwargs.get("filters", None)
+        search = kwargs.get("search", None)
+        if id is not None:
+            self.opencti.admin_logger.info("Fetching audit with ID", {"id": id})
+            query = (
+                """
+                query AuditRead($id: String!) {
```
marieflorescontact: should it be AuditRead($id: ID!) ?

client-python/pycti/entities/opencti_log.py
```diff
@@ -0,0 +1,184 @@
...
+        """
+        id = kwargs.get("id", None)
+        custom_attributes = kwargs.get("customAttributes", None)
+        filters = kwargs.get("filters", None)
+        search = kwargs.get("search", None)
+        if id is not None:
+            self.opencti.admin_logger.info("Fetching log with ID", {"id": id})
+            query = (
+                """
+                query LogRead($id: String!) {
```
marieflorescontact: AuditRead($id: ID!) ?

marieflorescontact (commented):
thank you for your work :)
It works wells once AuditRead($id: String!) replaced by AuditRead($id: ID!)

client-python/pycti/entities/opencti_audit.py
scarletmerlin123: I reviewed the other classes and noticed that the ID type is only used with mutations with the exception of in opencti_api_work.py. Does it need to be an ID type?

marieflorescontact: Yes, it needs to be `ID! `here. The backend GraphQL schema explicitly defines these queries as `audit(id: ID!) `and l`og(id: ID!)` (unlike most other entities which accept String!, I don't know why ...). This is confirmed by the runtime error we encountered:
`Variable "$id" of type "String!" used in position expecting type "ID!"`

scarletmerlin123:
@marieflorescontact Thanks for reviewing! I made the changes that you suggested and the additional changes suggested by copilot.

marieflorescontact (approved):
thank you for your work :)

PR #16397, opened 2026-06-04
feat(frontend): custom view attribute widget - chunk 2 (#15848)

### LINKED ISSUE #15848: feat(custom-views): attribute widget for Custom Views
## Use case

Custom Views allow users to create configurable dashboard-like views displayed on entity pages, making it easy to visualize the most relevant relationships (e.g., latest vulnerabilities, malware used by an Intrusion Set, etc.).

However, Custom Views currently only support widgets that display **relationships, entity counts, or activity**. There is no way to display attributes of the entity itself such as its description, labels, markings, first_seen, last_seen, confidence, etc.

Users who build Custom Views to replace or complement the default entity tabs are forced to switch back to the Overview tab just to see basic entity attributes, breaking their workflow.

## Current Workaround

Users must navigate to the "Overview tab" and consult the hardcoded "Details" and "Basic Information" sections to view entity attributes. There is no way to embed this information within a Custom View.

This means Custom Views cannot serve as a "self-contained, single-pane-of-glass" for an entity, they always require tab-switching for basic attribute information.

## Proposed Solution

Add a new **"Attribute" widget type** to the widget library, available in the Custom Views context (entity-scoped dashboards).

**Configuration:**
- **Attribute picker**: when creating the widget, the user selects which attributes to display from a dynamic list based on the current entity type. This includes:
  - Type-specific attributes (e.g., `first_seen`, `last_seen`, `resource_level`, `primary_motivation` for an Intrusion Set)
  - Common attributes (e.g., `description`, `labels`, `markings`, `author`, `confidence`)
- **Layout options**: choice between 1-column or 2-column grid, with control over attribute ordering and positioning

SarahBocognano:
### Proposed changes

* Add logic related to custom attributes for a specific entity

### Related issues

* closes: https://github.com/OpenCTI-Platform/opencti/issues/15848

### Checklist


opencti-platform/opencti-front/src/components/dashboard/DashboardRawViz.tsx
```diff
@@ -1,16 +1,35 @@
...
+  host,
 }: DashboardRawVizProps) => {
+  const { isFeatureEnable } = useHelper();
+  const isCustomAttributesWidgetEnable = isFeatureEnable('CUSTOM_ATTRIBUTES_WIDGET');
+
+  const startDate = config?.relativeDate
+    ? computeRelativeDate(config.relativeDate)
+    : config?.startDate;
+
+  const endDate = config?.relativeDate
```
Archidoit: you can use computeStartEndDates

opencti-platform/opencti-front/src/private/components/widgets/WidgetCustomAttributesCard.tsx
```diff
@@ -0,0 +1,308 @@
...
+      const revoked = 'revoked' in data ? (data as { revoked?: boolean }).revoked ?? false : false;
+      return (
+        <ItemBoolean
+          status={revoked}
+          label={revoked ? t_i18n('Yes') : t_i18n('No')}
+          reverse
+        />
+      );
+    }
+    case 'created':
```
Archidoit: rather than listing every attribte, whouldn't it be possible to do a 'case' according to the 'attribute type'. This 'attribute type' information can be added for each attribute column definition (in WidgetListDefaultColumns.tsx).
Thus, if we add a new attribute, we'll only have to add it in WidgetListDefaultColumns, with is label and type.

opencti-platform/opencti-front/src/private/components/widgets/WidgetListsDefaultColumns.tsx
```diff
@@ -60,8 +60,23 @@ const availableWidgetColumns: Record<string, WidgetColumn[]> = {
...
     { attribute: 'container_content', label: 'Content' },
   ],
+  Campaign: [
+    { attribute: 'objective', label: 'Objective' },
+    { attribute: 'first_seen', label: 'First seen' },
+    { attribute: 'last_seen', label: 'Last seen' },
+  ],
   'Malware-Analysis': [
     { attribute: 'product', label: 'Product' },
+    { attribute: 'result_name', label: 'Result name' },
```
Archidoit: here, we can add the attribute type for each attribute
It's a suggestion to discuss

opencti-platform/opencti-front/src/utils/widget/widgetCustomAttributesRendererUtils.tsx
```diff
@@ -0,0 +1,620 @@
...
+          ))}
+        </Stack>
+      </FieldOrEmpty>
+    );
+  },
+};
+
+// ─── Grouping
+const groupingRenderers: EntityRenderers = {
+  context: (data) => {
```
Archidoit: with the attribute type being indicating in the attribute column definition, maybe we can assume context is of type vocab, and thus call a generic rendering function for open vocab

opencti-platform/opencti-front/src/utils/widget/widgetCustomAttributesRendererUtils.tsx
```diff
@@ -0,0 +1,620 @@
...
+        value={context}
+      />
+    );
+  },
+};
+
+// ─── Malware-Analysis
+const malwareAnalysisRenderers: EntityRenderers = {
+  product: (data) => {
+    const product = 'product' in data
```
Archidoit: this file has a lot of code repetition, consider using generic function.

For example for this line, we can use a generic function like : 

const getField = <T,>(data: unknown, key: string): T | undefined => (data as Record<string, unknown>)[key] as T | undefined;

opencti-platform/opencti-front/src/utils/widget/widgetCustomAttributesRendererUtils.tsx
```diff
@@ -0,0 +1,620 @@
...
+    const result = 'result' in data
+      ? (data as { result?: string }).result
+      : undefined;
+    return (
+      <FieldOrEmpty source={result}>
+        <Tag label={result ?? ''} />
+      </FieldOrEmpty>
+    );
+  },
+  result_name: (data) => {
```
Archidoit: other example here, if attribute is of type 'text', we call a generic function returning FieldOrEmpty+ExpandableMarkdown.

Or at least, if we don't add the attribute type, call a generic function returning FieldOrEmpty+ExpandableMarkdown.

Archidoit:
In a custom view with filter 'entity_type = campaign', the last_seen, first_seen and aliases columns are always empty

image


Even though these attributes are not for the listed campaigns : 
image

Archidoit:
For report entity types, 'content' is always empty even though the report content is not 

image


@SarahBocognano: Thank you ! The entities List attribute is not in my scope here ^^It seems like a bug. I can try to fix it if it's fast but I will not take much time on this widget. Only the "custom attributes" is to check

Archidoit:
For malware entity types, the 'is family' is always loading, and the other attributes of malwares are always empty : 
image

@SarahBocognano: Same that my message above :) Not in the scope

opencti-platform/opencti-front/src/private/components/widgets/WidgetColumnsCustomizationInput.tsx
```diff
@@ -1,26 +1,36 @@
...
 };
 
+export type WidgetColumnsLayout = '1' | '2';
+
 const WidgetColumnsCustomizationInput: FunctionComponent<WidgetConfigColumnsCustomizationProps> = ({
   availableColumns,
   defaultColumns,
   value = [],
   onChange,
+  isCustomView = false,
```
mcartaud: In general, I prefer to avoid having several props that are meant to work together : what will happen if you have only one of them? Since we have no documentation on component, I believe the signature should be self explanatory, so perhaps replace this with something like
```
iDontKnowTheNameOfTheParameter?: {
  layout,
  onLayoutChange
}
```

That way, only the first level props is optional, having it means you have in your custom view case and the other props are mandatory because you always want them in this case

opencti-platform/opencti-front/src/private/components/widgets/WidgetColumnsCustomizationInput.tsx
```diff
@@ -58,11 +68,24 @@ const WidgetColumnsCustomizationInput: FunctionComponent<WidgetConfigColumnsCust
...
       <AccordionDetails sx={{ background: 'none', paddingBlock: theme.spacing(2) }}>
+        {isCustomView && onLayoutChange && (
+          <Box sx={{ marginBottom: theme.spacing(2) }}>
+            <Typography variant="h4">{t_i18n('Layout')}</Typography>
+            <RadioGroup
+              row
+              value={layout}
+              onChange={(e) => onLayoutChange(e.target.value as WidgetColumnsLayout)}
+            >
+              <FormControlLabel value="1" control={<Radio size="small" />} label={t_i18n('1 column')} />
```
mcartaud: Nit: you could use a const for WidgetColumnsLayout, that way you could map on it here to create the options and still use it to type what you want (I believe)

opencti-platform/opencti-front/src/utils/widget/widget.d.ts
```diff
@@ -55,6 +55,7 @@ export interface WidgetDataSelection {
   sort_mode?: string | null;
   field?: string;
   relationship_type?: string;
+  layout?: '1' | '2' | null;
```
mcartaud: Can you use the type you defined earlier here? (or perhaps the type derived from the const if you followed my suggestion

opencti-platform/opencti-front/src/private/components/widgets/WidgetColumnsCustomizationInput.tsx
mcartaud: I see you changed the code here, I'll restart the review on the whole code changes rather than per commits, but I keep the message to discuss the "standard" if you want

opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectsCustomAttributes.tsx
```diff
@@ -0,0 +1,613 @@
...
+import WidgetNoHostEntity from '../../../../components/dashboard/WidgetNoHostEntity';
+import WidgetCustomAttributes from '@components/widgets/WidgetCustomAttribute';
+import WidgetNoData from '../../../../components/dashboard/WidgetNoData';
+import Loader, { LoaderVariant } from '../../../../components/Loader';
+import useQueryLoading from '../../../../utils/hooks/useQueryLoading';
+import { getCustomAttributesColumns } from '@components/widgets/WidgetListsDefaultColumns';
+import { StixCoreObjectsCustomAttributesQuery } from '@components/common/stix_core_objects/__generated__/StixCoreObjectsCustomAttributesQuery.graphql';
+
+export const stixCoreObjectsCustomAttributesQuery = graphql`
+  query StixCoreObjectsCustomAttributesQuery($id: String!) {
```
mcartaud, line 16: 😳, that's a long query

opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectsCustomAttributes.tsx
```diff
@@ -0,0 +1,613 @@
...
+  dataSelection,
+  parameters = {},
+  popover,
+  host,
+}: StixCoreObjectsCustomAttributesProps) => {
+  const { t_i18n } = useFormatter();
+  const rootRef = useRef(null);
+  const isPreviewMode = host?.kind === 'custom-view'
+    && Boolean(host.customViewTargetEntityId)
+    && host.previewMode;
```
mcartaud: Is it possible to have `previewMode === false` and `kind !== custom-view` and `!host.customViewTargetEntityId`? and in this case is it ok not be treat the view as in previewMode?
Not sure it is a problem, but I want to make sure it is either not possible or correctly handled :)

opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectsCustomAttributes.tsx
```diff
@@ -0,0 +1,613 @@
...
+
+  const selection = (dataSelection as WidgetDataSelection[])[0];
```
mcartaud: Why the `as`? It seem correctly typed
Also why using an array in the props if only the first entry is used? It feels better to have this component taking only one WidgetDataSelection and the doing this manipulation in the caller, if it is possible (i.e. we don't have to respect a specific interface for the component)
WDYT?

opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectsCustomAttributes.tsx
```diff
@@ -0,0 +1,613 @@
...
+  const layout: WidgetColumnsLayout = (selection.layout as WidgetColumnsLayout) ?? '1';
```
mcartaud: I don't think the as is necessary here since you already type the const

opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectsCustomAttributes.tsx
```diff
@@ -0,0 +1,613 @@
...
+
+  const resolvedEntityId = host?.kind === 'custom-view' ? host.customViewTargetEntityId : undefined;
+  const entityType = host?.kind === 'custom-view' ? host.customViewTargetEntityType : undefined;
+
+  const columns = (selection.columns && selection.columns.length > 0)
+    ? selection.columns as readonly WidgetColumn[]
```
mcartaud: Idem here for the `as`

opencti-platform/opencti-front/src/private/components/widgets/WidgetCustomAttributesCard.tsx
```diff
@@ -0,0 +1,270 @@
...
+          <FieldOrEmpty source={list}>
+            {list.map((v) => <ValueCopy key={v} value={v} />)}
+          </FieldOrEmpty>
+        );
+      }
+      if (!value) return empty();
+      return <ValueCopy value={value as string} />;
+    }
+    case 'cvss_score':
+      // Géré exclusivement dans entityTypeRenderers (nécessite 2 champs)
```
mcartaud: is it ok to have french comments?

opencti-platform/opencti-front/src/private/components/widgets/WidgetCustomAttributesCard.tsx
```diff
@@ -0,0 +1,270 @@
...
+      return (
+        <FieldOrEmpty source={list}>
+          <Stack direction="row" flexWrap="wrap" gap={1}>
+            {list.map((item) => <Tag key={item} label={item} />)}
+          </Stack>
+        </FieldOrEmpty>
+      );
+    }
+    case 'text_list': {
+      return <TextList list={value as string[] | undefined} />;
```
mcartaud: Are we confident the values have the correct type for each `attributeType`? Feels pretty easy to have a `NaN` or similar as a result

opencti-platform/opencti-front/src/private/components/widgets/WidgetCustomAttributesCard.tsx
```diff
@@ -0,0 +1,270 @@
...
+
+export type StixCoreObject = NonNullable<StixCoreObjectsCustomAttributesQuery$data['stixCoreObject']>;
+
+interface WidgetCustomAttributesCardProps {
+  column: WidgetColumn;
+  data: StixCoreObject | null | undefined;
+  isCustomViewReadOnly?: boolean;
+}
+
+const getField = <T,>(data: unknown, key: string): T | undefined =>
```
mcartaud: This method can probably be improved to avoid all the cast in the rest of the file. Perhaps by using a typeguard?

opencti-platform/opencti-front/src/private/components/widgets/WidgetCustomAttributesColumnsInput.tsx
```diff
@@ -0,0 +1,226 @@
...
+};
+
+const DraggableColumnItem: FunctionComponent<DraggableColumnItemProps> = ({
+  column,
+  index,
+  isLast,
+  label,
+  onRemove,
+}) => (
+  <Draggable key={column.attribute} draggableId={column.attribute ?? ''} index={index}>
```
mcartaud: I don't think the `key` field is needed here

opencti-platform/opencti-front/src/private/components/widgets/WidgetCustomAttributesColumnsInput.tsx
```diff
@@ -0,0 +1,226 @@
...
+                  <ListItemText primary={t_i18n(formatColumnName(column))} />
+                </ListItem>
+              ))}
+            </List>
+          </Box>
+
+          {/* Selected — 2/3, layout 1 or 2 columns */}
+          <DragDropContext onDragEnd={layout === '2' ? handleDragEndDoubleColumns : handleDragEndSingleColumn}>
+            {layout === '2' ? (
+              <Box sx={{
```
mcartaud: NIT: Perhaps you could extract those two layout parts in their own component in the same file, it would improve readability I believe

mcartaud (commented):
Ouf, large PR 💪
I mostly put code comments rather that on the feature itself because I don't have enough knowledge yet to make good comments

opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectsCustomAttributes.tsx
SarahBocognano: Yep, no choice here... GraphQL is great but it's a lot of repetitive stuff ^^

Archidoit:
You introduce a bug for list widgets in custom views. New columns were added, and there are always empty. Example: the empty columns for malwares in the screenshot below should be filled since those fields contain values for some malwares

image

Archidoit:
In the attribute widget for a campaign: the 'objective' field is displayed as light blue if it's empty, whereas it should be displayed white (see difference with markings in the screenshot below) : 
image

Archidoit:
> You introduce a bug for list widgets in custom views. New columns were added, and there are always empty. Example: the empty columns for malwares in the screenshot below should be filled since those fields contain values for some malwares
> 
> image

This bug is introduced by your PR since it was not possible to select those not-working attributes before.

Archidoit:
Same for other fields, example with malware analyses attributes here (only assignees is displayed correctly if empty) : 
image

opencti-platform/opencti-front/src/private/components/widgets/StixCoreObjectsCustomAttributes.tsx
```diff
@@ -41,6 +41,9 @@ export const stixCoreObjectsCustomAttributesQuery = graphql`
         name
         description
         aliases
+        confidence
```
mcartaud: How it this fixing tests? If it is only for that, I'd say we should fixes the tests themselves rather than fetching more data

opencti-platform/opencti-front/src/private/components/widgets/StixCoreObjectsCustomAttributes.tsx
```diff
@@ -117,38 +149,19 @@ const StixCoreObjectsCustomAttributes = ({
       showPreviewTag={isPreviewMode}
     >
       <div ref={rootRef} style={{ height: '100%', overflowY: 'auto' }}>
-        {isMissingHostEntity
+        {!resolvedEntityId
```
mcartaud: The check is different here because we can ends up in the missing entity case even if it is probably not an issue because we don't need one (was checked in previous version)

opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectsCustomAttributes.tsx
```diff
@@ -72,14 +72,12 @@ interface StixCoreObjectsCustomAttributesContentProps {
   queryRef: PreloadedQuery<StixCoreObjectsCustomAttributesQuery>;
   columns: readonly WidgetColumn[];
   layout: WidgetColumnsLayout;
-  isCustomViewReadOnly?: boolean;
```
mcartaud: Nit: this is quite hard to review : you added those changes 2 commits before and then removed them here, perhaps only dropping the commit would have make it easier for the reviewer?

opencti-platform/opencti-front/src/private/components/widgets/WidgetCreationParameters.tsx
```diff
@@ -33,6 +33,7 @@ import EntitySelectWithTypes from '../../../components/fields/EntitySelectWithTy
 import { FilterGroup } from '../../../utils/filters/filtersHelpers-types';
 import useAuth from '../../../utils/hooks/useAuth';
 import useHelper from '../../../utils/hooks/useHelper';
+import WidgetCustomAttributesColumnsInput, { WidgetColumnsLayout } from '@components/widgets/WidgetCustomAttributesColumnsInput';
```
mcartaud: I am not sure I understand the commit (and the commit description), perhaps merging it with the place you introduced usage of WidgetCustomAttributesColumnsInput and WidgetColumnsLayout would have make it easier to review

opencti-platform/opencti-front/src/private/components/widgets/WidgetCustomAttributesCard.tsx
```diff
@@ -84,6 +84,7 @@ const renderAttributeValue = (
       return <ItemCreators creators={data.creators ?? []} />;
 
     case 'objectAssignee': {
+      // objectAssignee est présent sur plusieurs types (Report, Task, Case...)
```
mcartaud: FR comment. I am confused, I put this comment on a previous commit, then removed it because the comment was removed and it comes back again, I am a bit lost 🙂
Or maybe it was elsewhere?

Edit: I think I selected the wrong commit to start with and did the whole PR again 😭

opencti-platform/opencti-front/src/private/components/widgets/StixCoreObjectsCustomAttributes.tsx
```diff
@@ -0,0 +1,154 @@
+import React, { useRef } from 'react';
+import { graphql } from 'react-relay';
```
mcartaud: I am confused, this file was deleted at some point, then comes back here, perhaps reverting the commit where you deleted it would make it easier to review, this makes the intent of the PR not very clear

opencti-platform/opencti-front/src/private/components/widgets/WidgetCustomAttributesCard.tsx
```diff
@@ -206,6 +207,7 @@ const renderAttributeValue = (
     }
 
     default: {
+      // Accès générique pour les champs non typés explicitement (aliases, custom fields...)
```
mcartaud: idem FR

mcartaud (commented):
Code looks good to me, but functionally @Archidoit approval will be more pertinent than mine

Archidoit:
@mcartaud Yes everything is good, except the design for some empty fields, Sarah is currently fixing this, than I'm gonna approve.

opencti-platform/opencti-front/src/private/components/widgets/WidgetCustomAttributesCard.tsx
```diff
@@ -165,8 +174,8 @@ const renderAttributeValue = (
 
   if (!data) {
     return (
-      <Typography variant="body2" sx={{ color: 'text.disabled', fontStyle: 'italic' }}>
-        —
+      <Typography>
+        -
```
Archidoit: you can use the EMPTY_VALUE constant

Archidoit (approved):

PR #16411, opened 2026-06-05
feat(authentication): force a password change (#9606)

### LINKED ISSUE #9606: feat(authentication): force a password change by an admin
## Use case

- Ability for an admin to force a password change on next login to ensure the passwords is updated and different from first login.
- Ability to define a pasword duration in order to force passwordchange after a certain time (force password rotation).

## Proposed Solution

Within policy, an adminsitrator will be able to decide how long a password is valid for (ex: 1y).
Once duration is set, at next login, the change password date will be calculated using (now + password duration validity). 
When the date is reached, user will have to change its password.

By "forcing" the password date to "today", it will force the password change to the current day, meaning it will act as a froce password beahvior for an admin. 

CelineSebe:
### Proposed changes

* Reset pwd by an admin 
* Add screens in UI that user needs to change the password 
* For user, redirection /change-password screen
* For admin:
      - in update drawer, in password, an option to reset password is present
      - in users list, a new column to know the pwd valid until (date)
      - in overview, a "pwd valid until"
      - in authentication, a new policy "Password validity duration in days (0 equals unlimited)"
* Bulk users to update force a password change
* Add FORCE_PASSWORD_CHANGE FF

### Related issues

* Resolves #9606 

### Checklist

### Further comments

https://feat-issue-9606-feat.octi.staging.filigran.io/


marieflorescontact (commented):
locally tested and it works well except for FF.

It seems that FF is not applied everywhere on frontend: when the feature is not enabled, I am still able to select Users and ask for a background task (wich does not process because backend is well protected by FF).  

Image

Then User overview is updated  with `Password Valid Until `date despite User is not redirected to change password screen (wich is normal beacuse feature is not supposed to be working)

Image

opencti-platform/opencti-front/src/relay/environment.jsx
```diff
@@ -133,6 +144,17 @@ export const defaultCommitMutation = {
 export const relayErrorHandling = (error, setSubmitting, onError) => {
   if (setSubmitting) setSubmitting(false);
   if (error && error.res && error.res.errors) {
+    const passwordChangeRequired = R.any(
+      (e) => R.pathOr(undefined, ['extensions', 'code'], e) === 'PASSWORD_CHANGE_REQUIRED',
+      error.res.errors,
+    );
```
marieflorescontact, line 149: Please consider remove ramda:

suggestion :
```
  const passwordChangeRequired = error.res.errors.some(
  (e) => e?.extensions?.code === 'PASSWORD_CHANGE_REQUIRED',
);
```

opencti-platform/opencti-front/src/public/components/login/AlertFlashError.tsx
```diff
@@ -1,18 +1,32 @@
 import { useCookies } from 'react-cookie';
+import { useFormatter } from '../../../components/i18n';
 import LoginAlert from './LoginAlert';
 
 const FLASH_COOKIE = 'opencti_flash';
 
+export const FLASH_CODE_MESSAGES: Record<string, string> = {
+  IP_NOT_ALLOWED: 'Your IP address is not allowed to access this platform',
+  PASSWORD_CHANGE_REQUIRED: 'You must change your password before continuing',
```
marieflorescontact: I don't know how to test this code.
I try to set a cookie `opencti_flash=PASSWORD_CHANGE_REQUIRED`, the message did not appear.
It works AUTH_ERROR (also by setting a browser cookie)

CelineSebe: That code is unnecessary after all. The “force password change” setting only applies to locally authenticated users who log in. I've removed that code.

marieflorescontact (approved):

PR #16452, opened 2026-06-08
fix(query): use terms query for keyword to avoid too_many_clauses (#16400)

### LINKED ISSUE #16400: fix: `buildCompleteUsers` issues an un-chunked `contact_information` query that exceeds Lucene `max_clause_count`, breaking user-cache build at scale
## Description

When the platform builds the user cache, `buildCompleteUsers()` (in `src/domain/user.js`) resolves the `Individual` entity linked to each Platform User by issuing a **single OpenSearch query** filtering `contact_information` against the email of **every** Platform User in the list:

```js
const contactInformationFilter = {
  mode: 'and',
  filters: [{ key: 'contact_information', values: clients.map((c) => c.user_email) }],
  filterGroups: []
};
const individualArgs = {
  indices: [READ_INDEX_STIX_DOMAIN_OBJECTS],
  filters: contactInformationFilter,
  noFiltersChecking: true
};
const individualsPromise = fullEntitiesList(context, SYSTEM_USER, [ENTITY_TYPE_IDENTITY_INDIVIDUAL], individualArgs);
```

Each user email becomes one boolean clause. Once a deployment has more Platform User accounts than Lucene's `indices.query.bool.max_clause_count` (default **1024**), the query throws `too_many_clauses` and the **user cache fails to build**.

Because authentication, the notification manager, and automation playbooks all depend on the user cache, this single failure cascades into a **full platform outage**.

## Impact

- **Deterministic outage** above ~1024 Platform Users — the cache build fails on **every** rebuild (startup and any cache invalidation), not intermittently.
- Affects large-tenant / MSSP-pattern deployments that provision a Platform User per managed customer or contact.
- **Deactivating users does not mitigate.** The user list feeding the cache build (`findAllUser` → `ENTITY_TYPE_USER`) has no `account_status` filter, so inactive accounts still contribute clauses. Only **deleting** accounts reduces the count.

## Root cause detail

The `contact_information` query exists solely to populate a single field on the resolved user — the back-link to the matching Individual entity:

```js
individual_id: individualMap.get(client.user_email)?.internal_id,
```

No other resolved user data depends on this query, which means the fix is well-isolated.

## Steps to Reproduce

1. Provision >1024 Platform User accounts.
2. Trigger a user-cache (re)build (restart, or any user create/update/delete that invalidates the cache).
3. Observe `too_many_clauses` (`search_phase_execution_exception`) in the logs, originating from the `Individual` / `contact_information` query.
4. Observe authentication, notifications, and automations failing platform-wide.

## Expected Behavior

The user-cache build should scale beyond the OpenSearch/Elasticsearch clause limit regardless of Platform User count.

## Suggested Fix

**Chunk** the `contact_information` resolution in `buildCompleteUsers` into batches below `max_clause_count` (e.g. 500-email batches) and merge the results into `individualMap`. The function already iterates and yields (`doYield()`); the `Individual` lookup needs the same batched treatment rather than a single query. Because the query feeds only the `individual_id` field, this change is isolated and low-risk.

## Related scaling concern (separate issue, same driver)

At high Platform User counts, the notification manager (`getNotifications`) auto-generates **2–3 native triggers per user** (`generateAssigneeTrigger`, `generatePlatformNotificationTrigger`, …) and evaluates them against every stream event — an `O(users × events)` load. This is **distinct** from the clause-count bug (it reads the cache via `getEntitiesListFromCache`; it does not build it), but it is driven by the same large account population and will remain a scaling wall after the chunking fix. Worth tracking separately.

## Environment

- OpenCTI **6.9.0** (verified present through 6.9.29, current head of the 6.9.x line — no fix exists yet)
- OpenSearch/Elasticsearch with default `max_clause_count` (1024)
- Deployment pattern: large Platform User population (per-customer / MSSP provisioning)

nvkhope:
### Proposed changes

* Batch (500) individual lookup queries to stay under the search engine clause limit

### Related issues

* closes #16400 

### How to test this PR

 1. Start the platform with OpenSearch and confirm it boots normally
 2. Create 1024+ users via the GraphQL API to exceed the default clause limit (e.g script)
 3. Restart the platform and check the logs:
  -- Before this fix: the platform logs a `max_clause_count` error from OpenSearch during cache initialization
  -- After this fix: the platform boots successfully, users are resolved in batches of 500
 
### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/database/engine.ts
+++ b/opencti-platform/opencti-graphql/src/database/engine.ts
@@ -189,7 +189,7 @@ import type {
 } from '../types/store';
 import type { BasicStoreSettings } from '../types/settings';
 import { completeSpecialFilterKeys } from '../utils/filtering/filtering-completeSpecialFilterKeys';
-import { IDS_ATTRIBUTES } from '../domain/attribute-utils';
+import { IDS_ATTRIBUTES, KEYWORD_TERMS_ATTRIBUTES } from '../domain/attribute-utils';
 import { schemaRelationsRefDefinition } from '../schema/schema-relationsRef';
 import type { FiltersWithNested } from './middleware-loader';
 import { pushAll, unshiftAll } from '../utils/arrayUtil';
@@ -2565,7 +2565,7 @@ export const buildLocalMustFilter = (validFilter: any) => {
     } else {
       // case where we would like to build a terms query
       const isTermsQuery = (operator === 'eq' || operator === 'not_eq') && values.length > 0 && !values.includes('EXISTS')
-        && arrayKeys.every((k) => (!k.includes('*') && (k.endsWith(ID_INTERNAL) || k.endsWith(ID_INFERRED))) || IDS_ATTRIBUTES.includes(k));
+        && arrayKeys.every((k) => (!k.includes('*') && (k.endsWith(ID_INTERNAL) || k.endsWith(ID_INFERRED))) || IDS_ATTRIBUTES.includes(k) || KEYWORD_TERMS_ATTRIBUTES.includes(k));
       if (isTermsQuery) {
         if (operator === 'eq') {
           for (let i = 0; i < arrayKeys.length; i += 1) {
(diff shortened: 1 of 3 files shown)


SouadHadjiat (commented):
Suggestion to change the query to use terms with all the values instead of multi_match boolean conditions for each value.

SouadHadjiat (approved):

PR #16505, opened 2026-06-10
feat(draft-workflow): send draft-workflow comment by notification to assignees and participants (#14054)

### LINKED ISSUE #14054: feat(workflow): ability to comment on a transition & view the comment in the UI

frapuks:
### Proposed changes

* Send comment by notification to assignees and participants
* Rework UI to display comment in the draft

### Related issues

* closes #14054 

### How to test this PR

- Have 2 users
- Create a draft workflow (settings / customizations / entity types / draft)
- Add minimum 2 status to have a transition
- Enable comment on the transition
- Use the User1 to create a draft and add the User2 as assignee
- Enter the draft
- Select the transition and write a comment
- Connect with the User2
- Go to the notifications
- See the notification with the comment
- Click on the notification to navigate to the draft

### Checklist


opencti-platform/opencti-front/src/private/components/drafts/DraftRoot.tsx
```diff
@@ -147,6 +160,18 @@ const RootDraftComponent = ({ draftId, queryRef, refetch }: RootDraftComponentPr
     }
   }, [draftContext, draftId, enterDraft, isDraftReadOnly, t_i18n]);
 
+  // Show a popup once per history entry if the last transition has a comment
```
nvkhope: _Question_: The useEffect writes the localstorage as soon as a new timestamp is detected, if there is any re-render or navigation before reading, it'll never show again

Could we handle the logic inside the dialog's onClose callback? 🤔  we could avoid using a useEffect and have a dedicated custom hook, which would be straightforward to test in isolation with renderHook

non-blocking 😄

nvkhope:
I have an error when clicking on the notification from the user 2 (even with the user 2 in the Manage access restriction) Did I do something wrong? 😄 

image

image

frapuks:
> I have an error when clicking on the notification from the user 2 (even with the user 2 in the Manage access restriction) Did I do something wrong? 😄
> 

I think you have a user without enough capabilities to access to the draft. The mimum needed is "create update knowledge" in draft (or in main)

opencti-platform/opencti-graphql/src/modules/workflow/domain/workflow-domain.ts
```diff
@@ -95,6 +100,65 @@ const validateVersionConsistency = (workflowEntity: WorkflowDefinitionEntity): v
...
+            }],
+          }],
+        };
+        return addNotification(context, SYSTEM_USER, notificationPayload);
+      }),
+    );
+    results
+      .filter((r): r is PromiseRejectedResult => r.status === 'rejected')
+      .forEach(({ reason }) => {
+        logApp.error('[OPENCTI-MODULE] Failed to send notification to recipient', { cause: reason });
```
aHenryJard:
```suggestion
        logApp.error('[OPENCTI-MODULE] Failed to send workflow notification to recipient', { cause: reason });
```

frapuks:
Ok for last(?) review.
Changes since last review :
- Update one error message (workflow-domain.ts)
- Prevent displaying the comment to the submitter
- Add dates to the notification

nvkhope (approved):

PR #16508, opened 2026-06-10
fix(search): fix search on activity/history (#16507)

### LINKED ISSUE #16507: fix(scope): Search is broken on activity
## Description

Whenever I search for a word in activity, the search result is empty.

## Environment
 LAtest

## Reproducible steps

1.Go in activity
2. Search for any word: "Login", "logout", "APT", "{youremail}"

## Expected output

You get some results.

## Actual output

The list is empty

CelineSebe:
### Proposed changes

* add a non-regression test to check historyFiltering generate top-level + nested clauses
* lGenerateFullTextSearchShould has been expanded
* add a unit test validating the presence of both top-level (Activity) and nested (history changes) search clauses,

### Related issues

* resolves #16507 
* link to #8237 

### How to test this PR
1.Go in activity
2. Search for any word: "Login", "logout", "APT", "{youremail}"

### Checklist


opencti-platform/opencti-graphql/tests/01-unit/domain/search-test.js
```diff
@@ -90,3 +90,19 @@ it('should buildLocalMustFilter build query from ids filter with terms', () => {
   };
   expect(query).toEqual(expectedQuery);
 });
+
+it('should generate search clauses for both activity and history fields in historyFiltering mode', () => {
```
aHenryJard, line 94: This test is really low level, could you please also add a test at domain level, for example a search on activity on event that occurs on new platform startup (so you are sure that they will be there)

opencti-platform/opencti-graphql/src/database/engine.ts
```diff
@@ -2166,26 +2168,57 @@ export const elGenerateFullTextSearchShould = (search: string, args: ProcessSear
...
-                  type: 'phrase',
-                  query: ex,
-                  lenient: true,
-                  fields: BASE_SEARCH_HISTORY,
-                },
+    pushAll(shouldSearch, cleanExactSearch.map((ex) => [
+      {
+        bool: {
+          must: [
+            { term: { 'entity_type.keyword': ENTITY_TYPE_ACTIVITY } },
```
SouadHadjiat: historyFiltering is used also to search on history entity type, not just activity, does history search still work?

CelineSebe: Yes, history search still works. The activity specific top-level clauses are scoped to Activity only. History search still relies on nested history changes matching.

I added an integration test that generates a real History event (field update) to check this.

SouadHadjiat: So for activity results we search on BASE_SEARCH_ATTRIBUTES, but not for history events. Is this intended? For example I tried to search on the message that we have in history events, and it doesn't work. Maybe a question for product?

CelineSebe: thanks for you review @SouadHadjiat 
@nino-filigran Do you think we want to align this change with history?

opencti-platform/opencti-graphql/src/database/engine.ts
```diff
@@ -2184,6 +2187,21 @@ export const elGenerateFullTextSearchShould = (search: string, args: ProcessSear
           ],
         },
       },
+      {
+        bool: {
+          must: [
+            { term: { 'entity_type.keyword': ENTITY_TYPE_HISTORY } },
```
SouadHadjiat: this could be simplified by putting ENTITY_TYPE_HISTORY in the previous condition along with ENTITY_TYPE_ACTIVITY in a terms query (instead of term)

CelineSebe: BASE_SEARCH_ATTRIBUTES is too broad for History and can widen behavior unexpectedly, or security issue.

I did that as a precaution

CelineSebe: I'm looking into this to simplify things a bit, for a refacto

SouadHadjiat: What differences do we have between activity and history for this query? If we want to restrict the fields, then why not for both? why just for history?

CelineSebe: because of that initial outcome:
https://github.com/OpenCTI-Platform/opencti/issues/8237

I push the changes

SouadHadjiat (approved):
tested locally

PR #16524, opened 2026-06-11
feat(indicator): upgrade yara to yara-x (#16523)

### LINKED ISSUE #16523: feat(indicator): yara lib is under maintenance, need to move to yara-x
## Use case

From https://github.com/virustotal/yara

Image

https://virustotal.github.io/yara-x/blog/yara-x-is-stable/

## Proposed solution

Upgrade to yara-x

Yes

aHenryJard:
### Proposed changes

* Upgrade yara python library to yara-x which is the new maintained version.

WARNING: this lib requires cargo to be installed... so if we move to it we need to update all the release chain !
even with `pip3 install --no-binary :all: yara-x` it's not working, it requires to patch the requirement.txt... not great.

=> This is a POC, we need to design to target solution

### Related issues

* closes #16523 

### Checklist

### DIFF
--- a/opencti-platform/Dockerfile_featurebranch
+++ b/opencti-platform/Dockerfile_featurebranch
@@ -67,14 +67,14 @@ COPY opencti-graphql/src/python/requirements.txt /opt/opencti/src/python/require
 
 RUN apk add --no-cache tini python3 python3-dev libmagic
 RUN set -eux; \
-    apk add --no-cache gcc git; \
+    apk add --no-cache gcc g++ git musl-dev cargo; \
     rm -f /usr/lib/python*/EXTERNALLY-MANAGED; \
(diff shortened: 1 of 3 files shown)


opencti-platform/Dockerfile_featurebranch
```diff
@@ -67,14 +67,14 @@ COPY opencti-graphql/src/python/requirements.txt /opt/opencti/src/python/require
 
 RUN apk add --no-cache tini python3 python3-dev libmagic
 RUN set -eux; \
-    apk add --no-cache gcc git; \
+    apk add --no-cache gcc g++ git musl-dev cargo; \
```
aHenryJard, line 70: Does not work very well on Alpine....

PR #16556, opened 2026-06-12
fix(backend): guard subscription filters against null payload to prevent redis sub leak (#16402)

### LINKED ISSUE #16402: fix(graphql): memory leak in subscription `withCancel` wrapper leaks `PubSubAsyncIterator` on cancellation
## Description

The `withCancel` higher-order function in `opencti-platform/opencti-graphql/src/graphql/subscriptionWrapper.ts` fails to tear down the wrapped `asyncIterator` when a subscription terminates. The underlying `PubSubAsyncIterator` (from `graphql-redis-subscriptions`) remains registered against Redis, and every subsequent event published to its topic accumulates in `pushQueue` with no consumer. Over time this produces unbounded RSS growth on frontend pods and ends in repeated OOM kills (exit 137).

Confirmed against current `master`: the file is byte-identical between 7.260318.0 (where this was reproduced) and `master`, so all releases in that range are affected.

## Environment

- Observed on **7.260318.0**.
- `subscriptionWrapper.ts` has only been modified twice in repo history (Apollo major version bump and an eslint reformat); neither touched the cancel path. **All versions back to the introduction of `withCancel` are affected.**
- Multi-pod frontend deployment (Kubernetes).
- Load-balanced WebSocket connections in front of the GraphQL subscription endpoint.
- Redis used as the pub/sub backend via `graphql-redis-subscriptions`.
- Observed: ~10 OOMKills on a single affected pod over the observation window; unbounded RSS growth visible in Grafana node memory metrics.

## Steps to reproduce

1. Deploy OpenCTI frontend with multiple replicas behind a load balancer that may reset idle WebSocket connections (or use a load balancer with a WebSocket idle timeout shorter than typical subscription lifetime).
2. Open one or more collaborative-edit views in the UI that establish GraphQL subscriptions wrapped by `withCancel` (e.g., a STIX domain object detail page).
3. Sever the WebSocket connection abruptly (load-balancer reset, idle timeout, or `kill -9` on the client) rather than letting the client issue a clean unsubscribe.
4. Continue publishing events on the affected topic (any edit to objects of the subscribed type).
5. Take a V8 heap snapshot of the frontend pod and search for retained `PubSubAsyncIterator` instances.

### Observations

- **8 orphaned `PubSubAsyncIterator` instances** retained on the STIX domain object edit topic.
- Each instance held `pushQueue.length = 536`, identical across all 8 — consistent with a single triggering event (LB reset or WebSocket idle timeout) that severed all consumers simultaneously while leaving the iterators registered against Redis.
- RSS on the affected pod grew monotonically until OOMKill.

## Expected behavior

When a subscription terminates (clean unsubscribe, client disconnect, or transport-level reset), the wrapped `asyncIterator.return()` should be invoked with correct `this` binding, which causes `PubSubAsyncIterator.return()` to call `emptyQueue()` internally — unsubscribing from Redis and clearing `pushQueue`. The iterator should then be eligible for GC.

## Actual behavior

The iterator is never torn down. Redis subscription remains registered. Every subsequent published event on the topic is pushed into the orphaned iterator's `pushQueue`, which grows without bound. The iterator and its queue are retained indefinitely, holding references that prevent GC of associated context.

## Potential Root cause

Two defects in the cancellation path of `withCancel` in `opencti-platform/opencti-graphql/src/graphql/subscriptionWrapper.ts`:

### Defect 1 — Lost `this` binding on `returnFn`

`returnFn` is extracted from `asyncIterator.return` as an unbound function reference, then invoked as `returnFn()`. This loses the `this` context that `graphql-redis-subscriptions`' `PubSubAsyncIterator.return()` requires — internally it calls `this.emptyQueue()` to unsubscribe from Redis and clear `pushQueue`. With `this` undefined, the unsubscribe never executes. The Redis subscription stays registered and the queue keeps filling.

### Defect 2 — Unguarded `onCancel()` blocks teardown on throw

`onCancel()` runs before `returnFn()` with no `try`/`catch` around it. If `onCancel()` throws — context cleanup failure, user session lookup error, anything — control never reaches `returnFn()` at all, and the iterator leaks via the same mechanism. The teardown is not guaranteed to run.

## Blast radius

`withCancel` is used by **15 confirmed resolver call sites**, covering every collaborative-edit subscription in the platform:

- STIX domain objects
- STIX cyber observables
- STIX core relationships
- Users
- Groups
- Settings
- Labels
- Kill chain phases
- External references
- Marking definitions
- Notifications
- (and additional subscription resolvers using the same wrapper)

The original reporter observed this behavior on STIX domain object subscriptions, but the leak mechanism is in the wrapper itself and applies identically to all 15 call sites. Any deployment with active collaborative-edit usage and non-clean WebSocket terminations is exposed.

Chehao:
### Proposed changes

  Null-guard every GraphQL subscription filter so dereferencing `payload.user` / `payload.instance` can never throw
  synchronously inside the filter:

  * `subscribeToUserEvents` — guard `payload` + `payload.instance`.
  * `subscribeToAiEvents` — guard `payload` + `payload.user` + `payload.instance`.
  * `subscribeToInstanceEvents` — guard `payload` + `payload.instance` up front; the `payload.user` check is scoped to
  the `!notifySelf` branch (the only branch that dereferences it), so a user-less / system-originated event still
  reaches `notifySelf` subscribers instead of being dropped at the top.
  * `subscribeToPlatformSettingsEvents` — add the previously-missing `payload` + `payload.instance` guard.

  Root cause: a synchronous throw inside a `withFilter` callback is surfaced by graphql-ws as a `4500` socket close
  that bypasses normal teardown, leaving the underlying redis subscription orphaned — connections accumulate over time
  (memory/connection leak).

  ### Related issues
  * fixes https://github.com/OpenCTI-Platform/opencti/issues/16402

  ### How to test this PR

  1. Open a GraphQL subscription (instance / user / AI / platform-settings events) over websocket.
  2. Publish an event whose payload has a missing/undefined `user` or `instance` (e.g. a system-originated edit).
  3. Before the fix: the websocket is force-closed with code `4500` and the redis subscription is orphaned
  (subscription/connection count keeps growing).
  4. After the fix: the filter returns `false`, the socket stays open, and the redis subscription count stays stable.
  For `subscribeToInstanceEvents` with `notifySelf=true`, a user-less system event is still delivered.

  ### Checklist

  ### Further comments

  This PR is the defensive guard on the consumer side. The producer-side contract (ensuring system-originated events
  carry a valid user / `SYSTEM_USER`) can be addressed separately. Rebased onto latest `master`.

### DIFF
--- a/opencti-platform/opencti-graphql/src/graphql/subscriptionWrapper.ts
+++ b/opencti-platform/opencti-graphql/src/graphql/subscriptionWrapper.ts
@@ -27,8 +27,8 @@ const withCancel = (asyncIterator: AsyncIterableIterator<any>, onCancel: () => v
 export const subscribeToUserEvents = async (context: any, topics: string | string[]): Promise<AsyncIterable<any>> => {
   const asyncIterator = pubSubAsyncIterator(topics);
   const filtering = await withFilter(() => asyncIterator, (payload) => {
-    if (!payload) {
-      // When disconnected, an empty payload is dispatched.
+    // A throw here closes the socket (4500) and orphans the redis sub; guard all fields.
+    if (!payload || !payload.instance) {
       return false;
     }
     return [payload.instance.user_id, payload.instance.id].includes(context.user.id);
@@ -41,8 +41,8 @@ export const subscribeToUserEvents = async (context: any, topics: string | strin
 export const subscribeToAiEvents = async (context: any, id: string, topics: string | string[]): Promise<AsyncIterable<any>> => {
   const asyncIterator = pubSubAsyncIterator(topics);
   const filtering = await withFilter(() => asyncIterator, (payload) => {
-    if (!payload) {
-      // When disconnected, an empty payload is dispatched.
+    // A throw here closes the socket (4500) and orphans the redis sub; guard all fields.
+    if (!payload || !payload.user || !payload.instance) {
       return false;
     }
     return payload.user.id === context.user.id && payload.instance.bus_id === id;
@@ -71,11 +71,15 @@ export const subscribeToInstanceEvents = async (
   const filtering = await withFilter(
     () => pubSubAsyncIterator(topics),
     (payload) => {
-      if (!payload) {
-        // When disconnected, an empty payload is dispatched.
+      // A throw here closes the socket (4500) and orphans the redis sub; guard before deref.
+      if (!payload || !payload.instance) {
         return false;
       }
       if (!notifySelf) {
+        // Only this branch needs the event user; a user-less (system) event must still reach notifySelf subs.
+        if (!payload.user) {
+          return false;
+        }
         return payload.user.id !== context.user.id && payload.instance.id === id;
       }
       return payload.instance.id === id;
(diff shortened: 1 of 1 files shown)


lndrtrbn:
Thanks for your contribution @Chehao! Looks like a good improvement indeed, could you sign your commit please? (and update the description based on your last push, it seems outdated)

lndrtrbn (approved):

PR #16576, opened 2026-06-12
feat(backend/frontend): add auto-reload support for public dashboards (#9109)

### LINKED ISSUE #9109: feat(dashboards): add auto-reload support for public dashboards
## Use case

When you want to set-up a Public Dashboard Monitor, you want to leave it there and have the dashboard always show the latest data.

## Current Workaround

Use Auto-Refresher web browser plug-ins, or use a web page that reloads a popup showing the dashboard. Neither is a good solution, or supported by the device showing the web page.

## Proposed Solution

Let the administrator enable or disable Public Dashboard Auto-Reloading, and interval of refresh. OR let the user that browsed to the Public Dashboard to set Auto-Reloading by himself with a dropdown or checkbox on the Public Dashboard page.

No

Gwendoline-FAVRE-FELIX:
### Proposed changes

* This PR adds auto-refresh to public dashboards by reusing the existing control (DashboardRefreshControl) directly on the public page.
Supported intervals include 1m, 5m, 15m, 30m, 1h, 12h, 1d, and Off (default value if not defined).
The refresh configured on the custom dashboard is correctly carried over when the public dashboard is created and can then be modified locally on the public page (without backend persistence).
Technically, public widgets now use usePublicDashboardViz + DashboardRefreshContext to trigger refetches with startTransition, which prevents the blink/loader flash during refresh.
A test has been added to usePublicDashboardViz to validate the refetch upon token change without triggering a loader.

### Related issues

* closes https://github.com/OpenCTI-Platform/opencti/issues/9109

### How to test this PR
* Create a custom dashboard and add many widgets, and select the refresh interval value
* Create a public dashboard from your custom dashboard
* Open the public dashboard in new window
* Update data on openCTI and check widgets updated in public dashboard according the refresh interval value

### Checklist


opencti-platform/opencti-front/src/components/dashboard/dashboard-types.ts
```diff
@@ -4,6 +4,8 @@ export interface DashboardConfig {
   startDate?: string | null;
   endDate?: string | null;
   relativeDate?: string | null;
+  refreshInterval?: number | null;
+  refresh_interval?: number | null;
```
Archidoit, line 7: why 2 refresh interval attributes with different semantic? maybe an error, keep only refreshInterval

opencti-platform/opencti-front/src/public/components/dashboard/PublicDashboardHeader.tsx
```diff
@@ -22,91 +24,98 @@ const PublicDashboardHeader = ({
...
+            variant="outlined"
+            disabled
+          >
+            <MenuItem value="none">{t_i18n('None')}</MenuItem>
+            <MenuItem value="days-1">{t_i18n('Last 24 hours')}</MenuItem>
+            <MenuItem value="days-7">{t_i18n('Last 7 days')}</MenuItem>
+            <MenuItem value="months-1">{t_i18n('Last month')}</MenuItem>
+            <MenuItem value="months-3">{t_i18n('Last 3 months')}</MenuItem>
+            <MenuItem value="months-6">{t_i18n('Last 6 months')}</MenuItem>
+            <MenuItem value="years-1">{t_i18n('Last year')}</MenuItem>
```
Archidoit: the list of the values available in 'menu item' is the same in DashboardTimeFilters.tsx. Maybe create a constante for the list of MenuItem ?

opencti-platform/opencti-graphql/src/modules/publicDashboard/publicDashboard-domain.ts
```diff
@@ -182,6 +182,12 @@ export const addPublicDashboard = async (
   }
 
   const parsedManifest = fromB64(dashboard.manifest ?? '{}');
+  const dashboardRefreshInterval = (dashboard as unknown as { refresh_interval?: number | null }).refresh_interval ?? 0;
+  parsedManifest.config = {
+    ...(parsedManifest.config ?? {}),
+    refreshInterval: dashboardRefreshInterval,
+    refresh_interval: dashboardRefreshInterval,
```
Archidoit, line 188: why having refreshInterval and refresh_interval ? useless duplication here

opencti-platform/opencti-graphql/src/modules/publicDashboard/publicDashboard-types.ts
```diff
@@ -89,6 +89,8 @@ export interface PublicDashboardCached extends BasicStoreIdentifier {
       startDate?: string;
       endDate?: string;
       relativeDate?: string;
+      refreshInterval?: number | null;
+      refresh_interval?: number | null;
```
Archidoit, line 92: same here

opencti-platform/opencti-front/src/components/dashboard/DashboardRefreshContext.tsx
```diff
@@ -0,0 +1,65 @@
+import { createContext, PropsWithChildren, useCallback, useContext, useMemo, useState } from 'react';
+
+interface DashboardRefreshContextValue {
+  // Integer incremented on every manual or auto refresh, so useDashboardViz can
+  // rely on this central refresh source. null when there is no dashboard-level
+  // refresh provider, signalling widgets to use their interval fallback instead.
```
mcartaud, line 6: Nit: using the `/** */` comment syntax make them available in completion or in the little modal opening when hovering a type, function, etc.
I would be in favor of always using this syntax for comments on types, methods, constants, etc and keeping the `//` syntax for comments inside a method explaining a tricky part

opencti-platform/opencti-front/src/components/dashboard/useDashboardRefresh.ts
```diff
@@ -69,25 +93,43 @@ const useDashboardRefresh = ({
...
-    }, msUntilNextTick);
+        interval = setInterval(() => {
+          triggerAutoRefresh();
+        }, refreshRate);
+      }, msUntilNextTick);
+    };
+
+    // Pause polling while the tab is hidden and reschedule (with catch-up) when
+    // it becomes visible again, so background tabs stop hitting the API.
+    const handleVisibilityChange = () => {
```
mcartaud, line 114: Nice 👍

opencti-platform/opencti-front/src/components/dashboard/DashboardRefreshContext.tsx
```diff
@@ -1,28 +1,42 @@
...
-export const useDashboardRefreshPendingState = () => useContext(DashboardRefreshContext).isRefreshing;
+export const useDashboardSetQueryPending = () => useContext(DashboardSetQueryPendingContext);
 
-export const useDashboardSetQueryPending = () => useContext(DashboardRefreshContext).setQueryPending;
+export const useDashboardRefreshPendingState = () => {
+  const pendingStateStore = useContext(DashboardPendingStateContext);
+  // useSyncExternalStore subscribes only to the pending-state store and updates
+  // this consumer when the store snapshot changes. This keeps widget trees that
+  // only need refreshToken from re-rendering on every query pending flip.
+  return useSyncExternalStore(
```
mcartaud, line 34: I am not sure I understand why this is needed, a context with only the boolean would not have been enough?
I'm up to some explanations, lots of complicate things happen here :)

esrevi: actually, this solves an issue when multiple widgets with different queries lifecycle finish that the refresh control button is listening for to show a loading state.

if there is one provider with one state, problem is on every query finishing, then every flip would change the provider's value, so react re renders every consumer of that context but we only want the refresh button to care about "isPending" value. So basically, widget A finishes -> whole tree re-renders, widget B finishes 10ms later -> whole tree re-renders again and so on, this makes flaky menu interaction in CI E2E test. 

So now, splitting in 3 contexts (each one is small and dedicated to a job),
- refreshContext makes the widgets to call refetch
- setQueryPendingContext permits the widgets to report their pending status but it should not cause re-renders because the value is a stable reference (Object.is is the same, no update fired, no re-render even if the function is called many times by widgets)
-  useSyncExternalStore lets DashboardRefreshControl to subscribe directly to the boolean status change, so before the split and this hook, every widget were re-rendering on each flag isPending flip, now, only the button is concerned

mcartaud (approved):

opencti-platform/opencti-graphql/src/modules/publicDashboard/publicDashboard-domain.ts
```diff
@@ -182,6 +182,11 @@ export const addPublicDashboard = async (
   }
 
   const parsedManifest = fromB64(dashboard.manifest ?? '{}');
+  const dashboardRefreshInterval = (dashboard as unknown as { refresh_interval?: number | null }).refresh_interval ?? 0;
```
fellowseb, line 185: Doesn't this assert mean we're missing `refresh_interval` on `BasicStoreEntityWorkspace` (and `StoreEntityWorkspace`) ?
Consider adding the field and removing the assert (:.

fellowseb (approved):

opencti-platform/opencti-graphql/src/modules/publicDashboard/publicDashboard-domain.ts
fellowseb: There are a few other `as unknown as` in this file easy/worth fixing too IMO !

esrevi: doing that in next PR!

PR #16577, opened 2026-06-12
feat(themes): add ability to control all theme elements (#16516)

### LINKED ISSUE #16516: feat(Themes): Allow for the complete control of theme modification
## Use case

The ability to completely control the theme would be beneficial for creating customized high-contrast themes or align with existing corporate branding. 

## Current workaround

Limited modifications are available via the theme creation and edit forms.

## Proposed solution

Add the ability for an advanced user to provide a json of the desired styles that overrides and extends the current theme. This includes the ability to customize mui component styles or specific parts of the existing theme. 

Additionally add the dark theme defaults as initial values on theme creation to make theme creation easier and have a way to restore a default theme if a theme needs to be reset and no backup was created. 

Yes

scarletmerlin123:
### Proposed changes

* Add field to theme creation/edition to allow passing json to provide specific overrides of styles
* Add default colors for theme creation to allow easy reset to a default theme 

### Related issues

* closes #16516 

### How to test this PR

1. create a new theme and verify that a new section for defining an override exists
2. Confirm the section warns the user of improperly formatted json, but continues to save
3. Confirm that when a proper override is provided that it uses that override value

example to override MUI paper background color :
```json
{
  "components": {
    "MuiPaper": {
      "styleOverrides": {
        "root": {
          "background-color": "#ED2100 !important"
        }
      }
    }
  }
}
```

### Checklist

### Further comments

This approach provides full control of theme properties without needing to modify the source code. The override extends and overrides the properties in the base theme.

If there is a syntax issue within the override that prevents it from producing the theme it will fall back to the defined theme until the syntax is corrected.


scarletmerlin123:
Closing, will revise approach to expand the built in theme editor to define parameters vs providing a json input.

PR #16580, opened 2026-06-12
feat(security): add IP allow-list support (#16602)

### LINKED ISSUE #16602: feat(security): Add security CIDR allow-list to allow login or access public information (like public dashboard)
## Use case

Add security CIDR whitelist to allow login or access public information (like public dashboard)

## Current workaround

No

## Proposed solution

In security > Authentication add IP access whitelist

Image

Allow user / group or organization to bypass

richard-julien:
### Proposed changes

* Add IP allow list support: administrators can configure a list of allowed IP addresses (individual IPs and CIDR ranges, IPv4/IPv6) from the Settings page
* Requests from IPs not in the allow list are blocked at the middleware level, with a configurable escape hatch (`app:ip_whitelist_enabled: false`) for recovery scenarios
* Exclusion mechanism: specific users, groups, or organizations can be exempted from the IP check, with automatic exclusion of the current admin and the default service accounts group on activation
* Rejection logs are throttled (1 per IP per minute, max 1000 entries) to prevent log flooding from browser polling
* Blocked access attempts are published as audit events

### Related issues

* closes #16602

### How to test this PR

1. Go to **Settings → Security → Authentications → IP access allow list**
2. Enable the allow list and add your current IP (e.g. `192.168.1.0/24`), save — confirm access is still granted
3. Remove your IP from the list, save — confirm you are redirected to the login page
4. Set `app:ip_whitelist_enabled: false` in config and verify all IPs bypass the check (escape hatch)
5. Re-enable, add a user/group to the exclusion list and verify they bypass the IP check
6. Attempt login from a blocked IP and verify the attempt is rejected and logged as an audit event

### Checklist

### Further comments

The allow list is enforced at the Express middleware level, before GraphQL resolvers are reached. Login requests (`Token` mutation) are validated separately via `checkIpAllowListForRequest` to block login attempts from unauthorized IPs. Internal routes (health checks, etc.) are excluded from the check.


opencti-platform/opencti-front/lang/back/de.json
```diff
@@ -486,6 +486,9 @@
   "Internal file": "Interne Datei",
   "Internal id": "Interne Kennung",
   "Investigated entities": "Untersuchte Entitäten",
+  "IP whitelist": "IP-Whitelist",
```
aHenryJard: Would be great to use allow list, at least on customer facing texts.

opencti-platform/opencti-graphql/src/http/ipWhitelistMiddleware.js
```diff
@@ -0,0 +1,283 @@
...
+    }
+  }
+  return true;
+};
+
+/**
+ * Checks if a source IP matches at least one entry in the whitelist.
+ * Uses ipaddr.js for robust IPv4/IPv6 and CIDR support.
+ */
+export const ipMatchesWhitelist = (sourceIp, whitelist) => {
```
aHenryJard, line 49: I wonder if we could not have some function in common with the https://github.com/OpenCTI-Platform/opencti/blob/c0cd63f50f0a462ba2f1c13466af1e7f7bf1015b/opencti-platform/opencti-graphql/src/modules/ingestion/ingestion-common.ts#L11

ludovic: I looked into it, but I don't think they share the same use case. extractHostFromUri is designed to parse URIs to extract domain names, whereas here we only deal with raw IP addresses and CIDR ranges via ipaddr.js. Since we don't have any URIs to parse in this middleware, sharing the function might not be applicable.

ludovic (approved):

labo-flg (approved):

PR #16630, opened 2026-06-16
feat(composer): warn and require user acknowledgement when deploying an unverified connector (#16558)

### LINKED ISSUE #16558: feat(connector catalog): Warn and require user acknowledgement when deploying an unverified connector
## Use case

The connector catalog (Composer) currently only allows deploying Verified connectors. We want to extend support to some 'Unverified' connectors as well.

When a user attempts to deploy an 'Unverified' connector, a warning dialog should be displayed, clearly stating that:
- This connector is not officially supported by Filigran.
- Filigran cannot be held responsible in case of issues.

The user must explicitly confirm they have acknowledged this before deployment proceeds.

## Acceptance Criteria

-  Unverified connectors are deployable from the catalog, with a clear visual distinction (e.g. badge).
-  A warning modal is shown before deploying an unverified connector.
-  Deployment proceeds only after explicit user acknowledgement.
-  No change to the current behaviour for Verified connectors.

Archidoit:
### Proposed changes
- Unverified connector visual distinction on the catalog card
image

- Unverified label in connector details

image

- Warning banner in the deploy drawer for unverified connectors

image

- If we deploy a not verified connector, after creating on 'create', a acknowledgement pop-up opens. The user needs to check the checkbox before deploying the connector:
image

image

### Related issues
fixes #16558


opencti-platform/opencti-front/src/private/components/data/IngestionCatalog/IngestionCatalogUnverifiedDeploymentPopover.tsx
```diff
@@ -0,0 +1,65 @@
...
+};
+
+const IngestionCatalogUnverifiedDeploymentPopover: React.FC<IngestionCatalogUnverifiedDeploymentPopoverProps> = ({
+  onClose,
+  isOpen,
+  onDeploy,
+}) => {
+  const { t_i18n } = useFormatter();
+  const [isAcknowledged, setIsAcknowledged] = useState(false);
+  return (
```
CelineSebe, line 28: isAcknowledged is stored in local state but is never set to false when the session is closed or opened.

```suggestion
  return (
  
   useEffect(() => {
    if (isOpen) {
      setIsAcknowledged(false);
    }
  }, [isOpen]);
  
```

Archidoit: indeed ! it's fixed

CelineSebe:
tested locally 👌

CelineSebe (approved):

PR #16700, opened 2026-06-19
fix(backend): add defaultNullValue to attribute definition & use it for indicators (#16682)

### LINKED ISSUE #16682: Upserting indicator without observable type remove the old value
## Description

When upserting an indicator, the attributes "observable type" and "detection" are overiden even if nothing is set in the input, leading to data loss.

## Environment

2. Version: 7.260615.0

## Reproducible steps

1. Create an indicator of type IPv4 
3. Try to create an other indicator with same name but no observable type
4. The upsert is done but observable type is not kept

## Expected output

Observable type and detection are kept

## Actual output

Observable type and detection are not kept

## Additional information

GraphQL mutations you can use to test:

```graphql
mutation IndicatorCreationMutation($input: IndicatorAddInput!) {
  indicatorAdd(input: $input) {
    id
    name
    x_opencti_detection
    x_opencti_main_observable_type
  }
}

# variables for initial creation
{
 "input": {
   "name": "142.11.200.1",
   "pattern": "[ipv4-addr:value = '5.6.43.2']",
   "pattern_type": "stix",
   "x_opencti_detection": true,
   "x_opencti_main_observable_type": "IPv4-Addr"
 }
}
# variables for upsert
{
 "input": {
   "name": "142.11.200.1",
   "pattern": "[ipv4-addr:value = '5.6.4.2']",
   "pattern_type": "stix",
 }
}
```

JeremyCloarec:
### Proposed changes

* it's now possible to set defaultNullValue on attributes: this value will be used on first creation (but not upsert) if input doesn't contain any value for the field and no default value was set in the entity settings
* indicator's x_opencti_main_observable_type now has defaultNullValue = 'Unknown' and x_opencti_detection has defaultNullValue = 'false' to keep the same behavior as before on first creation

### Related issues

* closes #16682

### Checklist


opencti-platform/opencti-graphql/src/schema/attribute-definition.ts
```diff
@@ -42,6 +42,7 @@ type BasicDefinition = {
   upsert_force_replace?: boolean; // For multiple, if upsert will for a replacement instead of cumulate information
   isFilterable: boolean; // If attribute can be used as a filter key in the UI
   editDefault: boolean; // TO CHECK ?????
+  defaultNullValue?: any; // If attribute has a default value when null (for example for the main observable type of indicators)
```
SouadHadjiat: why not just "defaultValue" ?

opencti-platform/opencti-graphql/src/modules/entitySetting/entitySetting-utils.ts
```diff
@@ -166,10 +166,29 @@ export const getDefaultValues = (attributeConfiguration: AttributeConfiguration,
...
+  return { ...input, ...Object.fromEntries(filledValues) };
+};
+
 export const fillDefaultValues = (user: any, input: any, entitySetting: any) => {
   const attributesConfiguration = getAttributesConfiguration(entitySetting);
   if (!attributesConfiguration) {
-    return input;
+    return fillDefaultNullValues(input, entitySetting);
   }
   const filledValues = new Map();
```
SouadHadjiat, line 190: to extract in a function getConfiguredDefaultValues

opencti-platform/opencti-graphql/src/modules/entitySetting/entitySetting-utils.ts
```diff
@@ -166,10 +166,29 @@ export const getDefaultValues = (attributeConfiguration: AttributeConfiguration,
...
+    }
+  });
+  return { ...input, ...Object.fromEntries(filledValues) };
+};
+
 export const fillDefaultValues = (user: any, input: any, entitySetting: any) => {
   const attributesConfiguration = getAttributesConfiguration(entitySetting);
   if (!attributesConfiguration) {
-    return input;
+    return fillDefaultNullValues(input, entitySetting);
```
SouadHadjiat: rename to getStaticDefaultValues

SouadHadjiat (commented):
Add an integration non regression test for the upsert case

SouadHadjiat (approved):

PR #16712, opened 2026-06-19
fix(draft-diff-panel): draft review panel not accessible without manage customization (#16706)

### LINKED ISSUE #16706: fix(scope): Draft review panel not accessible without manage customization
## Description

a User without manage customization does not have access to the panel review (panel review is accessible, but when clicking on an entity an error "page not found" is prompted to the user)
## Environment

Latest

## Reproducible steps

1. Create a draft
2. Go in that draft with a user that do not have manage customization but has "access knowledge"
3. edit that a draft with your user (add an entity)
4. Go in review tab
5. Click on the entity 

## Expected output

You can view the diff
## Actual output

Page not found. When going in grpahql explorer you can see that that you ahve an error "You are not allowed to do this".

ludovic:
### Proposed changes

* Add a new `attributeLabels` field on `EntitySetting` (GraphQL) protected by `@auth` only, returning just the `name` and `label` of each attribute
* Fix `DraftReviewDiffPanel` right panel being empty for users with only `KNOWLEDGE` capability by replacing the `attributesDefinitions` query (requires `SETTINGS_SETCUSTOMIZATION`) with the new `attributeLabels` query

### Related issues
* closes #16706 

### How to test this PR
1. Create a user with only the `KNOWLEDGE` capability (no `SETTINGS_SETCUSTOMIZATION`)
2. Open a draft workspace and navigate to the diff review panel
3. Verify that the right panel correctly displays entity attribute labels for changes
4. Also verify that a user with `SETTINGS_SETCUSTOMIZATION` still sees the panel correctly

### Checklist

### Further comments

`EntitySetting.attributesDefinitions` exposes the full attribute configuration (mandatory, default values, scale, etc.) and is intentionally restricted to `SETTINGS_SETCUSTOMIZATION`. The diff panel only needs attribute labels for display purposes, so a dedicated `attributeLabels` field accessible to any authenticated user is the minimal, non-breaking fix.

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/drafts/DraftReviewDiffPanel/DraftReviewDiffPanelContent.tsx
+++ b/opencti-platform/opencti-front/src/private/components/drafts/DraftReviewDiffPanel/DraftReviewDiffPanelContent.tsx
@@ -24,7 +24,7 @@ const draftReviewDiffPanelContentQuery = graphql`
   query DraftReviewDiffPanelContentQuery($entityType: String!) {
     subType(id: $entityType) {
       settings {
-        attributesDefinitions {
+        attributeLabels {
           name
           label
         }
@@ -52,7 +52,7 @@ const DraftReviewDiffPanelContentComponent: FunctionComponent<DraftReviewDiffPan
     queryRef,
   );
 
-  const labelMap = buildFieldLabelMap(subTypeData.subType?.settings?.attributesDefinitions);
+  const labelMap = buildFieldLabelMap(subTypeData.subType?.settings?.attributeLabels);
 
   const renderHeader = () => {
     const link = resolveLink(entity.entity_type);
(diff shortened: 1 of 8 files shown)


opencti-platform/opencti-front/src/private/components/drafts/DraftReviewDiffPanel/DraftReviewDiffPanelContent.tsx
```diff
@@ -24,7 +24,7 @@ const draftReviewDiffPanelContentQuery = graphql`
   query DraftReviewDiffPanelContentQuery($entityType: String!) {
     subType(id: $entityType) {
       settings {
-        attributesDefinitions {
+        attributeLabels {
```
SouadHadjiat: I wonder if we couldn't instead add these attributesLabels list to the rootPrivateQuery and then use "useEntitySettings" in the component, instead of fetching them with this query.

Image

nvkhope (commented):
I tested on my side, all good 👍

opencti-platform/opencti-front/src/private/components/drafts/DraftReviewDiffPanel/DraftReviewDiffPanelContent.tsx
ludovic: Good point 👍

PR #16743, opened 2026-06-22
feat(observables): auto-populate AS number from Autonomous-System name (#8383)

### LINKED ISSUE #8383: feat(ingestion): streamline the Autonomous System creation process
## Use case

AS are often received with the name "AS####". Given that number is mandatory, would be nice to auto-extract the number from the name.

mcartaud:
When creating an Autonomous-System observable, automatically parse and set the `number` field if the name matches the pattern `AS<number>` (e.g., typing "AS1234" fills in 1234 as the AS number).

### Proposed changes

* Auto populate AS number field if name field is provided and starts with `AS`
*

### Related issues

* closes #8383 

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/observations/stix_cyber_observables/StixCyberObservableCreation.jsx
+++ b/opencti-platform/opencti-front/src/private/components/observations/stix_cyber_observables/StixCyberObservableCreation.jsx
@@ -930,6 +930,29 @@ const StixCyberObservableCreation = ({
                               />
                             );
                           }
+                          if (
+                            attribute.value === 'name'
+                            && status.type === 'Autonomous-System'
+                          ) {
+                            const setDefaultAutonomousSystemId = (_, value) => {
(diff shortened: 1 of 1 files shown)


Gwendoline-FAVRE-FELIX:
It works well, but I have a question. What about the update? Do we want the number to change automatically if we change the name?

mcartaud:
> It works well, but I have a question. What about the update? Do we want the number to change automatically if we change the name?

Working on updates was planned but:
- we don't want to override value written by the user, so if something is in the number field, we must no update it
- the number field is a mandatory field, so it always have a value

So the feature is only useful at creation.
It was also planned to update CSV ingestion with that but same story, the number column is mandatory, not the name one, making only a small change to handle the same feature would have make the CSV configuration clumsy, so not doing it will be part of a more ambitious change someday

Gwendoline-FAVRE-FELIX (approved):

PR #16761, opened 2026-06-23
feat(draftworkflow): display metadata columns in restricted drafts (#16758)

### LINKED ISSUE #16758: fix(draft): Restricted drafts page don't show the new metadata columns
## Description

- Restricted drafts page don't show the new metadata columns

frapuks:
### Proposed changes

* Display metadata columns in restricted drafts

### Related issues

* closes #16758 

### How to test this PR

- Go to data / restriction / restricted draft
- See metadata colums
- Deactivate FF
- See old columns

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/data/restriction/RestrictedDrafts.tsx
+++ b/opencti-platform/opencti-front/src/private/components/data/restriction/RestrictedDrafts.tsx
@@ -10,14 +10,18 @@ import DataTable from '../../../../components/dataGrid/DataTable';
 import { DataTableProps } from '../../../../components/dataGrid/dataTableTypes';
 import { defaultRender } from '../../../../components/dataGrid/dataTableUtils';
 import { useFormatter } from '../../../../components/i18n';
+import ItemStatus from '../../../../components/ItemStatus';
 import type { Theme } from '../../../../components/Theme';
 import { hexToRGB } from '../../../../utils/Colors';
 import { computeValidationProgress } from '../../../../utils/draft/draftUtils';
 import { emptyFilterGroup, useBuildEntityTypeBasedFilterContext } from '../../../../utils/filters/filtersUtils';
+import useAuth from '../../../../utils/hooks/useAuth';
 import useConnectedDocumentModifier from '../../../../utils/hooks/useConnectedDocumentModifier';
+import useHelper from '../../../../utils/hooks/useHelper';
 import { usePaginationLocalStorage } from '../../../../utils/hooks/useLocalStorage';
 import { UsePreloadedPaginationFragment } from '../../../../utils/hooks/usePreloadedPaginationFragment';
 import useQueryLoading from '../../../../utils/hooks/useQueryLoading';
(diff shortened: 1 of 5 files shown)


opencti-platform/opencti-front/src/private/components/data/restriction/RestrictedDrafts.tsx
```diff
@@ -26,12 +30,40 @@ export const RestrictedDraftLineFragment = graphql`
         id
         entity_type
         name
+        description
```
nvkhope: I think description is not used

opencti-platform/opencti-front/src/private/components/data/restriction/RestrictedDrafts.tsx
```diff
@@ -198,7 +235,69 @@ const RestrictedDrafts = () => {
...
+    draft_status: {
+      id: 'draft_status',
+      label: 'Status',
+      percentWidth: 10,
+      isSortable: true,
+      render: (node: RestrictedDrafts_node$data) => (
+        node.workflowInstance?.currentStatus ? (
+          <ItemStatus status={node.workflowInstance.currentStatus} />
+        ) : (
+          <Chip
```
nvkhope: We could extract the chip into a var and reuse it in both columns

Archidoit:
Working well :)

frapuks:
Ready for 2nd review.
I fix Kevin comments and copilot comments.

 🧩 Column alignment
RestrictedDrafts now displays the same columns as Drafts:
• Without DRAFT_WORKFLOW FF: `name`, `creator`, `created_at`, `status`, `validation progress`
• With DRAFT_WORKFLOW FF: adds `createdBy`, `objectAssignee`, `objectParticipant` + `ItemStatus` for workflow
The GraphQL fragment was updated accordingly to fetch the new fields.

 ♻️ DraftStatusChip extraction
The `Chip` + `ItemStatus` render logic for draft status was duplicated in both pages. It's now extracted into a shared component: `src/private/components/common/draft/DraftStatusChip.tsx`
Both pages use it, keeping the visual behaviour identical.

 🐛 Race condition fix — OpenSearch UnsupportedError
`useRuntimeSortGuard` ran in a `useEffect`, but `useQueryLoading` also fires its first query in a `useEffect`. On first render, both effects flush in the same batch: the first (bad) query could be sent with a runtime-only `orderBy` (e.g. `createdBy`) before the guard had a chance to correct it — triggering an `UnsupportedError` on OpenSearch instances.
Fix: `useRuntimeSortGuard` now returns a `safeSortBy` value computed synchronously during render. Both Drafts and RestrictedDrafts use it directly in `queryPaginationOptions`. The `useEffect` is kept to repair the persisted localStorage state for subsequent reloads.

 🧹 Minor cleanup
• `isDraftWorkflowEnabled` extracted as a named boolean in both components (single source of truth for FF check)
• `Explicit RestrictedDrafts_node$data` type annotations removed from render callbacks (inferred, consistent with Drafts.tsx)

nvkhope (approved):

PR #16765, opened 2026-06-23
fix(form-submit): add no_split option to form submission (#16665)

### LINKED ISSUE #16665: fix(form-intake): entities created but not added to container's object_refs due to bundle split race condition
## Description

When a Form Intake is submitted and creates a container (e.g., Grouping) with additional entities (IOCs, Courses of Action, relationships), some of those entities end up being created but **not contained** in the container's `object_refs`.

This happens because the form bundle is split by the worker into individual sub-bundles that are re-queued to RabbitMQ. With multiple workers consuming the queue, the container sub-bundle can be processed before its dependencies have been ingested. After 3 failed retries (~9 seconds), the backend silently accepts the container **without** the unresolved references (`middleware.ts`, `retryNumber <= 2` condition).

The issue is non-deterministic and depends on worker count and queue load, making it difficult to reproduce in single-worker dev environments.

**Proposed fix:** Add `no_split: true` to the `pushToWorkerForConnector` call in `formSubmit` (`form-domain.ts` L513). Form bundles are small (a few objects), so there's no need to split them. This flag already exists and is used by the `taskManager`. It ensures the entire bundle is ingested atomically and sequentially by a single worker, eliminating the race condition.

## Environment

1. OS: Linux (GCP-hosted on-premise deployment)
2. Version: 7.260409.0
3. Other environment details: Multiple workers consuming the same RabbitMQ queue

## Reproducible steps

1. Configure a Form Intake with a container as main entity (e.g., Grouping) and additional entities (IOCs, Course of Action) with `includeInContainer: true`
2. Deploy with **multiple workers** (3+)
3. Submit the form with valid data
4. Check the created container

Note: difficult to reproduce with a single worker due to FIFO ordering. Requires multiple workers or significant queue load to trigger the race condition.

## Expected output

The container is created with all additional entities referenced in its `object_refs`. All entities are visible inside the container.

## Actual output

The container is created but some `object_refs` are silently dropped. The additional entities (IOCs, Courses of Action, relationships) exist in the platform but are **not** contained in the container.

Support logs show the pattern:
```
MISSING_REFERENCE_ERROR / GroupingAdd - unresolvedIds: ["course-of-action--...", "data-source--..."]
MISSING_REFERENCE_ERROR / GroupingAdd - unresolvedIds: ["data-source--..."]
MISSING_REFERENCE_ERROR / GroupingAdd - unresolvedIds: ["data-source--..."]
(silence — Grouping accepted without unresolved refs on 4th retry)
```

## Additional information

**Root cause in code:**

1. `form-domain.ts` L513 — `pushToWorkerForConnector` sends the bundle **without** `no_split: true`
2. `push_handler.py` L124 — worker splits bundles with >1 object and re-queues sub-bundles
3. `middleware.ts` L1148-1156 — after 3 retries (`retryNumber <= 2`), unresolved `object_refs` are silently dropped:

```typescript
// TODO this retry must be removed in favor of reworking the workers synchronization
if (isNotEmptyField(retryNumber) && expectedUnresolvedIdsNotDefault.length > 0 && retryNumber && retryNumber <= 2) {
    throw MissingReferenceError(...);
}
// retryNumber > 2 → accepted without unresolved refs
```

**Fix (1 line):**

```diff
// form-domain.ts L513
await pushToWorkerForConnector(connectorId, {
  type: 'bundle',
  applicant_id: user.id,
  content,
  work_id: work.id,
  draft_id: draftId,
  update: true,
+ no_split: true,
});
```

ludovic:
### Proposed changes

* Add `no_split: true` to the `pushToWorkerForConnector` call in `formSubmit` (`form-domain.ts`)
* Add unit test covering the `isDraft=false` path to assert `no_split: true` is passed

### Related issues
* closes #16665

### How to test this PR

**Unit test (local):**

 npx vitest run tests/01-unit/domain/form-submit-test.ts

**Functional test (requires multi-worker environment):**

The race condition only manifests with multiple workers consuming the same RabbitMQ queue simultaneously, so it cannot be reliably reproduced in a standard single-worker local setup.

To validate on a staging/multi-worker environment:
1. Configure a Form Intake with a container as main entity (e.g., Grouping) and additional entities (IOCs, Course of Action) with `includeInContainer: true`
2. Deploy with **3+ workers** on a loaded queue
3. Submit the form — verify the container contains **all** expected entities in its `object_refs`

### Checklist

### Further comments

Form intake bundles are small (a few objects) and must be ingested atomically: the container and all its referenced entities need to be processed by the same worker in a single pass.

Without `no_split: true`, the worker (`push_handler.py`) splits bundles with more than one object into individual sub-bundles that are re-queued to RabbitMQ. With multiple workers consuming the queue, the container sub-bundle can be processed before its dependencies have been ingested. After 3 failed retries (~9 seconds), `middleware.ts` silently accepts the container **without** the unresolved `object_refs`.

The `no_split` flag is already supported by the worker (see `push_handler.py` line 124: `if len(content["objects"]) == 1 or data.get("no_split", False)`) and is already used by the `taskManager` for the same reason. This fix aligns form intake with that existing pattern.

### DIFF
--- a/opencti-platform/opencti-graphql/src/modules/form/form-domain.ts
+++ b/opencti-platform/opencti-graphql/src/modules/form/form-domain.ts
@@ -515,6 +515,7 @@ export const formSubmit = async (
       work_id: work.id,
       draft_id: draftId,
       update: true,
+      no_split: true,
     });
 
     logApp.info('[FORM] Bundle sent to connector queue', { formId: form.id, workId: work.id, bundleId: bundle.id });
--- a/opencti-platform/opencti-graphql/tests/01-unit/domain/form-submit-test.ts
+++ b/opencti-platform/opencti-graphql/tests/01-unit/domain/form-submit-test.ts
@@ -2,6 +2,7 @@ import { vi, describe, it, expect, beforeEach } from 'vitest';
 import { formSubmit } from '../../../src/modules/form/form-domain';
 import * as draftWorkspaceDomain from '../../../src/modules/draftWorkspace/draftWorkspace-domain';
 import * as workDomain from '../../../src/domain/work';
+import * as rabbitmq from '../../../src/database/rabbitmq';
 import { BYPASS, SYSTEM_USER } from '../../../src/utils/access';
 
 const { mockStoreLoadById } = vi.hoisted(() => ({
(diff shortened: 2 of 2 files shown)


marieflorescontact (approved):
looks good

PR #16796, opened 2026-06-24
feat(security): add configuration for trust proxy in ip whitelisting (#16795)

### LINKED ISSUE #16795: feat(security): Add configuration for trust proxy in ip whitelisting
## Use case

Add configuration for trust proxy in ip whitelisting

richard-julien:
Fix #16795


ludovic (approved):
LGTM. I validated the IP whitelist behavior on staging. However, I wasn't able to specifically test trust_proxy_addresses as it requires an environment with a public-IP load balancer, which our staging doesn't have.

PR #16800, opened 2026-06-24
fix(playbook): component configuration remains when the component is replaced (#16016)

### LINKED ISSUE #16016: fix(playbook): component configuration remains when the component is replaced
## Description

In a playbook, you have the option of replacing a component with another one. When you do so, the configuration of the replaced component is not reset, and some parts of it can remain in the replacing component.

You can see it in the JSON file that you get when you export the playbook.

## Reproducible Steps

1. Create a playbook with a middle component that has a configuration set
2. Replace this component with another one that has different configuration options 
3. Import the playbook
4. In the JSON file, you will see that some of the configuration - playbook definition is not of your current component. 

In this exemple, the PLAYBOOK_CREATE_OBSERVABLE_COMPONENT has excludeMainElement set to true and newContainer to false, but those are not options that you can choose in this component : 

Image

## Expected Output

The configuration of replaced component should not appear anymore in the new ones.

## Actual Output

Configurations get mixed.

MTorbay-Filigran:
### Proposed changes

* Added a condition "componentIsReplaced" to initialize values and config when a component is replaced in playbook
* Moved the logic of initializing values in computeInitialComponentConfigValues in playbookComponents-utils.ts

### Related issues

* closes #16016 

### How to test this PR

To test :
- Create a playbook with a component (ex Manipulate knowledge with an action)
- Replace the component with a component with different configuration settings (ex extract observable)
- Export the playbook
- In the playbook definition, you should not see any elements of the first component settings (ex actions for manipulate knowledge component) in the component configuration :
Capture d’écran 2026-06-25 à 09 43 23

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/data/playbooks/playbookFlow/playbookComponents-utils.ts
+++ b/opencti-platform/opencti-front/src/private/components/data/playbooks/playbookFlow/playbookComponents-utils.ts
@@ -55,7 +55,7 @@ export interface NodeData {
   configuration?: PlaybookConfig;
   component?: PlaybookComponent;
 }
-interface computeInitialComponentConfigValuesParams {
+interface ComputeInitialComponentConfigValuesParams {
   action: string | null;
   currentConfig: PlaybookConfig | null;
   configurationSchema: PlaybookComponentConfigSchema | null;
@@ -63,7 +63,7 @@ interface computeInitialComponentConfigValuesParams {
   selectedComponent?: PlaybookComponent | null;
 }
 
-export type PlaybookFlowFormValues = PlaybookFlowFormData & Record<string, unknown>;
+export type PlaybookFlowFormValues = PlaybookFlowFormData;
 
 export const computeInitialComponentConfigValues = ({
   action,
@@ -71,7 +71,7 @@ export const computeInitialComponentConfigValues = ({
   nodeData,
   configurationSchema,
   selectedComponent,
-}: computeInitialComponentConfigValuesParams): PlaybookFlowFormValues => {
+}: ComputeInitialComponentConfigValuesParams): PlaybookFlowFormValues => {
   const initialValues: PlaybookFlowFormValues = {
     name: '',
     description: '',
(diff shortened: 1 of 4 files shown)


opencti-platform/opencti-front/src/private/components/data/playbooks/types/playbook-types.ts
```diff
@@ -10,6 +10,7 @@ export interface PlaybookConfig {
   applyWithFilters?: string;
   actions?: PlaybookUpdateAction[];
   triggerTime?: string;
+  applyToElements?: string;
```
aHenryJard: it seems from the code reading that the value is a limited set of string ('only-main' or 'all'), is it possible to list them or use an enum ?

opencti-platform/opencti-front/src/private/components/data/playbooks/playbookFlow/playbookComponents-utils.ts
```diff
@@ -46,3 +48,69 @@ export function groupAndSortPlaybookComponents(
...
     .filter((g) => g.items.length > 0);
 }
+
+export interface NodeData {
+  name?: string;
+  description?: string;
+  configuration?: PlaybookConfig;
+  component?: PlaybookComponent;
+}
+interface computeInitialComponentConfigValuesParams {
```
lndrtrbn:
```suggestion
interface ComputeInitialComponentConfigValuesParams {
```

opencti-platform/opencti-front/src/private/components/data/playbooks/playbookFlow/playbookComponents-utils.ts
```diff
@@ -46,3 +48,69 @@ export function groupAndSortPlaybookComponents(
...
+  action: string | null;
+  currentConfig: PlaybookConfig | null;
+  configurationSchema: PlaybookComponentConfigSchema | null;
+  nodeData?: NodeData;
+  selectedComponent?: PlaybookComponent | null;
+}
+
+export type PlaybookFlowFormValues = PlaybookFlowFormData & Record<string, unknown>;
```
lndrtrbn: Why do you need this `Record<string, unknown>`? That's open to any object property and if not necessary should be avoided

lndrtrbn: For what I see you should instead update the type `PlaybookFlowFormData` to contains the new prop `applyToElements`

lndrtrbn (approved):

PR #16802, opened 2026-06-24
fix(draft): bypass draft context for some draft element creation (#16801)

### LINKED ISSUE #16801: fix(draft): Draft creation from XTM browser extension can fail or be incomplete
## Description

The relation between XTM extension and the openCTI platform is not very reliable. Sometimes creating a new draft from the extension it completely fails (with an error message displayed on the webpage where the extension is used)
Sometimes the draft import is created but some entities are missing. It seem to depend on what is already present on the platform

## Environment

1. OS: all
2. Version: 7.260619.0
3. Other environment details: all browser

## Reproducible steps

For first case, 
- go [here](https://krebsonsecurity.com/page/2/)
- start the extension, scan the page
- select all
- click on import
- click on "create container"
- chose "report"
- activate "create as draft"
- click on "create container"

For second case, it is more complicated to have a reproduction since it seem to depends on whats already present on the platform and I didn't manage to reproduce it

## Expected output

The draft is created with all the elements I selected in my extension

## Actual output

For case 1, receive a "Cannot index unsupported element in draft context" error

mcartaud:
### Proposed changes

* bypass draft context for draft workspace creation
* fallback to bypassing draft context when searching elements to link to new draft element

### Related issues

* closes #16801 

### Checklist


mcartaud:
Fix is incorrect, the restrictions are to keep, issue is more returning info to the user or being able to have several draft in parallel

PR #16133, opened 2026-05-22
[backend/client] Make x_mitre_id case-insensitive for standard_id generation (#16132)

SamuelHassine:
### Proposed changes

* Make `x_mitre_id` case-insensitive (lowercase + trim) when computing the `standard_id` of **Attack Pattern** and **Course of Action**, by adding an `x_mitre_id` resolver in `stixBaseEntityContribution.resolvers` (`opencti-platform/opencti-graphql/src/schema/identifier.js`). Re-importing the same MITRE technique (e.g. `T1059` and then `t1059`, or with extra whitespace) now deduplicates the entity instead of creating a second one, consistently with how `name` is normalized for vulnerabilities and how `name` was made case-insensitive for Software in #15633.
* Mirror the same normalization in the Python client (`AttackPattern.generate_id` / `CourseOfAction.generate_id`) so STIX bundles produced by pycti directly compute the same canonical `standard_id` as the backend. When the normalized `x_mitre_id` is empty (whitespace-only input), the python client now also falls back to `name` to mirror the backend's `filteredIdContributions` empty-field filtering, so client and server stay aligned even on the edge case.
* Add migration `1779435787999-attack_pattern_case_insensitive_standard_id.js`, modeled on `1776859399668-software_case_insensitive_standard_id.js`: it recomputes the new `standard_id` for every Attack Pattern and Course of Action, merges colliding duplicates into a single canonical target via `mergeEntities` (the merge target is picked by relation count DESC — i.e. the most-connected entity — to minimize relation rewriting, ties broken by oldest `created_at` then lowest `internal_id`), and bulk-rewrites the remaining singletons' `standard_id` while archiving the previous one in `x_opencti_stix_ids` so older references keep resolving. Progress logs go through `logMigration.info` so the operator can follow the merge and bulk-rewrite loops from the migration log channel.
* Update the existing unit tests (`identifier-test.js` and `client-python` `test_bundle_ids_rewrite.py`) for the new canonical UUIDs of `MITREID` / `T1368` inputs, and add new assertions that explicitly cover case-insensitivity, surrounding-whitespace trimming, and the whitespace-only fallback to `name` for both Attack Pattern and Course of Action — locking client/server id parity in both test suites.

### Related issues

* Closes #16132

### Checklist

- [x] I consider the submitted work as finished
- [x] I tested the code for its functionality
- [x] I wrote test cases for the relevant uses case (coverage and e2e)
- [x] I added/update the relevant documentation (either on github or on notion)
[...]

PR #16820, opened 2026-06-25
feat(data-sanity): add a base manager for data sanity (#16313)

### LINKED ISSUE #16313: feat: data sanity manager as an alternative to some migrations
## Use case

We currently use migrations to run data sanity operations, ex: standard_id recomputing, deduplications. These operations can take a lot of time, and shouldn't block the platform initialization and start. 

## Proposed Solution

Create a data sanity manager, that runs tasks. When we want to run a task in the next release, we can add a migration that will only create this task, that the data sanity manager will run at the next start.

## Additional Information

This proposal comes from our migrations handling tech pain.

### LINKED ISSUE #16775: fix(migration): attack pattern migration is way too long
## Environment

2. Version: { e.g. 1.2.3 }
3. Other environment details:

## Reproducible steps

1. { e.g. Run ... }
2. { e.g. Click ... }
3. { e.g. Error ... }

aHenryJard:
### Proposed changes

- Add a “data sanity” manager that can register and execute data sanity operations (run-once / periodic / on-demand) without requiring a migration file. The fist implementation uses only 'run-once' script.
- implement as first script attackPatternCaseSensitiveDuplicatedId from #16775
- Allow to rerun a script that is 'run-once' on demand using graphql API
- Everything is under BYPASS capability
- Disable migration introduces in https://github.com/OpenCTI-Platform/opencti/pull/16133, is replaced by the operation in data sanity manager and this will avoid issues for upgrade that jump over to this version.
- Add a UI under feature flag 'DATA_SANITY_MANAGER'

#### Counter update

- attackPatternCaseSensitiveDuplicatedId test coverage do 1 merge of attack pattern

### Related issues

* closes #16313 
* closes #16775 

### How to test this PR

- Update the data sanity manager schedule to be more frequent for test purpose in development.json
```
"data_sanity_manager": {
    "interval": 5000
  }
```

Start the application, it should run without any issues.

#### Deeper test

- Start with OpenCTI version 7.260422.0
- Enable DATA_SANITY_MANAGER FF
- Create attack pattern using same x_mitre_id but one with lower case and one with upper case
- Update app and wait for manager to run

=> one of the attack pattern should have been merged.

### Checklist


opencti-platform/opencti-graphql/tests/03-integration/10-modules/dataSanity/dataSanityManager-test.ts
```diff
@@ -0,0 +1,181 @@
...
+          throw Error('This is raising error - mockRunOnceOperationThatFails');
+        },
+        execution_type: 'on_demand',
+      },
+    ]);
+  });
+
+  let runOnceFirstRunDate: Date;
+  let runOncePeriodicRunDate: Date;
+
```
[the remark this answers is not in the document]
aHenryJard: We don't shuffle tests

opencti-platform/opencti-graphql/src/modules/attributes/internalObject-registrationAttributes.ts
```diff
@@ -262,6 +262,7 @@ const internalObjectsAttributes: { [k: string]: Array<AttributeDefinition<any>>
     { name: 'local_auth', label: 'Local authentication', type: 'object', format: 'flat', mandatoryType: 'no', editDefault: false, multiple: false, upsert: false, isFilterable: false },
     { name: 'cert_auth', label: 'Client certificate authentication', type: 'object', format: 'flat', mandatoryType: 'no', editDefault: false, multiple: false, upsert: false, isFilterable: false },
     { name: 'headers_auth', label: 'Header authentication configuration', type: 'object', format: 'flat', mandatoryType: 'no', editDefault: false, multiple: false, upsert: false, isFilterable: false },
+    { name: 'data_sanity_configuration', label: 'Data sanity configuration', type: 'object', format: 'flat', mandatoryType: 'no', editDefault: false, multiple: false, upsert: true, isFilterable: false },
```
aHenryJard: this is not needed anymore, I need to remove

opencti-platform/opencti-graphql/src/utils/access.ts
```diff
@@ -81,6 +81,7 @@ export const REDACTED_USER_UUID = '31afac4e-6b99-44a0-b91b-e04738d31461';
 export const RESTRICTED_USER_UUID = '27d2b0af-4d1e-42ae-a50c-9691bf57f35d';
 const PIR_MANAGER_USER_UUID = '1e20b6e5-e0f7-46f2-bacb-c37e4f8707a2';
 const HUB_REGISTRATION_MANAGER_USER_UUID = 'e16d7175-17c7-4dae-bd3c-48c939f47dfb';
+const DATA_SANITY_MANAGER_USER_UUID = 'a1b2c3d4-5678-4abc-9def-0123456789ab';
```
ludovic: Maybe use a real random UUID?

opencti-platform/opencti-graphql/src/manager/dataSanityManager.ts
```diff
@@ -0,0 +1,106 @@
...
+      const errorMessage = e?.message || String(e);
+      logApp.error('[DATA_SANITY_MANAGER] Data sanity operation failed', { operation: operation.identifier, error: e });
+      await markOperationAsExecuted(context, DATA_SANITY_MANAGER_USER, operation.identifier, executionTimeMs, false, errorMessage).catch(() => {});
+    }
+  }
+};
+
+export const dataSanityHandler = async () => {
+  const context = executionContext(DATA_SANITY_MANAGER_CONTEXT);
+  logApp.info('[DATA_SANITY_MANAGER] Running data sanity manager handler', { manager: DATA_SANITY_MANAGER_ID });
```
ludovic, line 73: If the app crashes during an operation, the entity stays stuck in is_running: true forever and the UI shows "Running" indefinitely with no way to distinguish it from an actual run. Would it make sense to add a logApp.warn at manager startup to detect and alert on this case, so ops can trigger a manual force_run?

On the UI side, adding a last_run_start_date field (set in markOperationAsRunning) would allow the frontend to detect stalled operations and display a "Stuck" status instead of "Running" when the operation has been running for an unexpectedly long time. WDYT?

aHenryJard: I need to try, there is a try catch in the 2 handler, it was supposed to catch error, I will have a look to that.

ludovic (approved):

PR #16894, opened 2026-06-30
feat(telemetry): add workflow publish count tracking and telemetry integration (#16892)

### LINKED ISSUE #16892: feat(workflow): telemetry on the publish
## Use case

When publishing a worfklow, we need to have telemetry to know whether the feature is used or not. 
Therefore, we need to have a telemtry count on the publish CTA

delemaf:
### Proposed changes

* add workflow publish count tracking and telemetry integration 

### Related issues

* closes #16892

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/modules/workflow/domain/workflow-domain.ts
+++ b/opencti-platform/opencti-graphql/src/modules/workflow/domain/workflow-domain.ts
@@ -568,8 +568,6 @@ export const publishWorkflowDefinition = async (
 
   await updateAttribute(executionContext, executionUser, workflowDefinitionEntity.id, ENTITY_TYPE_WORKFLOW_DEFINITION, updates);
 
-  addWorkflowPublishCount();
-
   const updatedWorkflow = await storeLoadById(
     executionContext,
     executionUser,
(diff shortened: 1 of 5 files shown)


opencti-platform/opencti-graphql/src/modules/workflow/domain/workflow-domain.ts
```diff
@@ -567,6 +568,8 @@ export const publishWorkflowDefinition = async (
 
   await updateAttribute(executionContext, executionUser, workflowDefinitionEntity.id, ENTITY_TYPE_WORKFLOW_DEFINITION, updates);
 
+  addWorkflowPublishCount();
```
OctaveLaventure: We do this pretty early in this function, something could break below it. Do we want to count publish operations even in that case? Or only when successful? Maybe we can put the counter at the end of the function in that case. If not, we should rename the metric to reflect that it's the publish attempts, not only the successful ones.

We could also add a test for it to ensure telemetry can't be removed without us noticing. Just a little mock and assert it's called once on publish (and eventually check it's not called on fail if that's what we want).

OctaveLaventure (approved):

PR #16899, opened 2026-06-30
feat(intel): define a default template for export (#11970)

### LINKED ISSUE #11970: feat(fintel): define a default template for export
## Use case

Define a default template to avoid selection each time an HTML or PDF expot is made. 

mcartaud:
### Proposed changes

* Add the notion of default for fintel template
* allow to add default on creation, or on details page or on list table
* preselect the default template on export
* update custom view to have the same behavior

### Related issues

* closes #11970 

### Checklist


opencti-platform/opencti-front/src/private/components/settings/sub_types/fintel_templates/FintelTemplateForm.test.tsx
```diff
@@ -41,6 +40,7 @@ describe('Component: FintelTemplateForm', () => {
         name: 'MyFintelTemplate',
         description: null,
         published: false,
+        default: false,
```
mcartaud, line 43: Add a test for default case

opencti-platform/opencti-graphql/src/modules/fintelTemplate/fintelTemplate.graphql
```diff
@@ -34,6 +35,7 @@ input FintelTemplateAddInput {
   template_content: String
   start_date: DateTime
   fintel_template_widgets: [FintelTemplateWidgetAddInput!]
+  default: Boolean!
 }
```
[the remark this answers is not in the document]
fellowseb: That's actually a good remark.
WDYT about making it optional and default to `false` ?

opencti-platform/opencti-front/src/private/components/settings/sub_types/custom_views/CustomViewFormDrawer.tsx
```diff
@@ -47,21 +59,54 @@ const CustomViewFormDrawer = ({
...
+        .catch((err) => {
+          setSubmitting(false);
+          handleError(err);
+        });
+    } else {
+      doAdd(values, setSubmitting);
+    }
+  };
+
   const handleEditField = (
```
fellowseb, line 153: We could also apply the logic to show CustomViewReplaceDefaultDialog when updating the Custom View in the callback.

(It's out-of-scope of this fintel template ticket but as you already touch Custom Views.. (:)

opencti-platform/opencti-front/src/private/components/settings/sub_types/custom_views/CustomViewsSettingsDataTable.tsx
```diff
@@ -142,7 +155,7 @@ const CustomViewsSettingsDataTable = ({
 
   return (
     <DataTable
-      initialValues={DEFAULT_SORT_CONFIG}
+      initialValues={{ sortBy: 'default', orderAsc: false }}
```
fellowseb, line 158: nit: I see that for the Fintel Templates we mange to sort the table by `default` then by `name`. Could we do the same here ? (Not sure how you managed this)

EDIT: ok I see you've applied a client-side sort (`data={[...defaultTemplates, ...otherTemplates]}`). Not sure we can do this easily here 🤔 .

opencti-platform/opencti-graphql/src/modules/fintelTemplate/fintelTemplate-domain.ts
```diff
@@ -82,6 +84,80 @@ export const checkFintelTemplateWidgetsValidity = (fintelTemplateWidgets: Fintel
...
+    ENTITY_TYPE_FINTEL_TEMPLATE,
+    [{
+      key: 'default',
+      value: ['true'],
+    }],
+  );
+  const deDefaulted = await applyUniqueDefaultTemplateConstraint(context, user, settingsType, templateId);
+  return [element, ...deDefaulted];
+};
+
```
fellowseb: I think we should call `publishUserAction()` in this handler too in order to generate an activity log.

EDIT: actually you'd need a handler dedicated to `fintelTemplateSetDefault` where you'd do that (instead of the unnamed handler that's currently in fintel-template-resolvers)

fellowseb: I believe we also should call `notify` to publish the change to the event Stream (`/stream`)

EDIT: actually you'd need a handler dedicated to `fintelTemplateSetDefault` where you'd do that (instead of the unnamed handler that's currently in fintel-template-resolvers)

fellowseb: I know realize that what I've done for Custom Views, repeated here, is not transactional & concurrent-safe. If multiple clients call `setTemplateAsDefault` we could end up with multiple/no defaults.

From what I understand, we could use a redis lock to mitigate the issue:

```
export const setTemplateAsDefault = async (
  context: AuthContext,
  user: AuthUser,
  settingsType: string,
  templateId: string,
) => {
await canCustomizeTemplate(context);
const lock = await lockResources([`fintel-template-default:${settingsType}`]);                                                                                                                                try {
  // ... rest of the function making the mutations
} finally {
  await lock.unlock();
}
}
```

opencti-platform/opencti-graphql/src/modules/fintelTemplate/fintelTemplate-domain.ts
fellowseb, line 240: Given we've created a dedicated mutation endpoint to set the `default` field we need to throw a `FunctionalError` when trying to mutate `default` using `fintelTemplateEditField` IMO or we'll have constraint inconsistencies. WDYT ?

Also I see we took a different approach in https://github.com/OpenCTI-Platform/opencti/pull/16994, similar to what we did in Custom Views : to not have a specific mutation dedicated to the `default` field. I see benefits (and disadvantages) in both approaches TBH. What was your rationale here ?

fellowseb (changes requested):
Tested, works well !
Thanks for the updates to the Custom View feature too !

opencti-platform/opencti-graphql/src/modules/fintelTemplate/fintelTemplate-domain.ts
fellowseb: Actually I think the locking logic needs to be in the handlers, outside of `setTemplateAsDefault`:

- In `addFintelTemplate`:
```
const lock = input.default ? await lockResources([`fintel-template-default:${settingsType}`]) : null;                                                                                                                                try {
  const created = await createEntity(
    context,
    user,
    finalInput,
    ENTITY_TYPE_FINTEL_TEMPLATE,
  );
  if (input.default) {
    await setTemplateAsDefault(context, user, settings_type, created.id);
  }
} finally {
  if (lock) await lock.unlock();
}
```

- In the new handler for `fintelTemplateSetDefault`

opencti-platform/opencti-graphql/src/modules/fintelTemplate/fintelTemplate.graphql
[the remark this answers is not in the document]
mcartaud: Since it is not possible to have different client and server, I thought it was ok to make it mandatory because it is the target state and it makes the intent more clear (we don't have to guess what is the default, even though it is not a hard guess I agree).
But no very strong opinions, I can change it if you want, or if it is possible to have decoupled back and front somehow

opencti-platform/opencti-graphql/src/modules/fintelTemplate/fintelTemplate-domain.ts
mcartaud: Using a single route is more idiomatic with what is done elsewhere in the app, changing it

opencti-platform/opencti-graphql/src/modules/fintelTemplate/fintelTemplate-domain.ts
mcartaud: not relevant anymore with the suppression of the dedicated default mutation

mcartaud: not relevant anymore with the suppression of the dedicated default mutation

opencti-platform/opencti-front/src/private/components/settings/sub_types/custom_views/CustomViewFormDrawer.tsx
mcartaud: 😱 I didn't notice this update button opening the drawer 😭. It is not the same behavior that for templates : for them on the template list and on a template details you have a menu action "update" opening this drawer (with the default toggle hidden). I'm updating the custom view to have the same behavior

opencti-platform/opencti-front/src/private/components/settings/sub_types/custom_views/CustomViewsSettingsDataTable.tsx
mcartaud: Yes, it is different because fintelTemplates are fetched with all the data of the views, whereas custom views have their own paginated requests. I didn't want to go to far in the refactoring, this PR goes already far beyond its original spec.
I have two solutions (if we want to do something)
- I can sort server side
- I can sort alphabetically what I receive in the front

I think solution 2 is the worst, it makes the user think it is sort, but it is not
With solution 1 it forces to fetch all custom views server side and do the sort manually. I don't think we have that much custom view, but it kinda defeat the purpose of having paginated view.

Perhaps best solution is to keep it that way, WDYT?

fellowseb: let's keep it that way for now yes 👍

opencti-platform/opencti-graphql/src/modules/fintelTemplate/fintelTemplate-domain.ts
fellowseb: We've decided to use this applicative lock even though the better tradeoff would've certainly been to tweak the data model in order to put the "default fintelTemplate" state in single(ton) entity to avoid the issue altogether. There's no big consequence on the UX nor performance-wise. We just plan on adding a comment to notify readers and AI agent that the best way to model a "unique value for a field amongst multiple documents" is not this one.

fellowseb (approved):

fellowseb (approved):

fellowseb (approved):

PR #16907, opened 2026-06-30
feat(custom-field): Chunk 1 - POC → Production Ready (Backend) (#17178)

### LINKED ISSUE #6177: feat(case-management): custom fields and templates in Cases and Incidents

### LINKED ISSUE #17178: Chunk 1 — POC → Production Ready (Backend)
📋 Detailed spec reference: CF for a single entity — Phase 1 & 2

> Goal: Harden the nested POC backend (PR #15884) into production-quality code, without touching the admin UI.

Approach: Nested (custom_field_values: [{field_name, int_value, ...}])

Scope (Case-Incident only, under feature flag)

Delivered beyond initial scope (in the Chunk 1 PR)

> Scope of Chunk 5 (STIX) will be revised afterwards based on what remains once these are accounted for.

ludovic:
### Proposed changes

* Add the `CustomFieldDefinition` module — a new internal object that lets admins define typed custom fields (integer, string, boolean, date, select, multi_select, markdown) and assign them to any SDO entity type via CRUD APIs (`SETTINGS_SETCUSTOMIZATION` auth).
* Register `custom_field_values` as a nested attribute on `ABSTRACT_STIX_DOMAIN_OBJECT`, with an in-memory cache loaded at boot and refreshed on every definition CRUD operation.
* **Per-entity-type settings model**: `mandatory` and `default_value` are defined per entity type a field is attached to (not globally on the definition), via a nested `entity_type_settings: [{ entity_type, mandatory, default_value }]` attribute. `min_value` / `max_value` / `select_options` / `multiple` stay global (intrinsic to the field type). Exposed in GraphQL + STIX export so the front/validation can resolve settings per entity type.
* Add a centralized validator (`validateCustomFieldValues`) that checks field type, integer bounds, select options, mandatory presence (including when zero values are submitted), duplicates, and cardinality — wired on CaseIncident creation and on the generic `updateAttribute` middleware.
* Inject custom field filter keys dynamically into the schema for any SDO type, with nested query translation (`adaptFilterToCustomFieldFilterKey`) supporting `eq`/`not_eq`/`contains`/`not_contains`/`starts_with`/`not_starts_with`/`ends_with`/`not_ends_with`/range operators, and dashboard distribution support via `elAggregationNestedTermsWithFilter`.
* Fix a pre-existing (unrelated to this feature) bug in `isSpecificTypeAttribute` (`schema-attributes.ts`) that prevented typed sub-attribute resolution (e.g. boolean) on `format: 'nested'` object attributes — surfaced by boolean custom field filtering, but generic to any nested object attribute.
* Add `representative` field declaration on the `CustomFieldDefinition` GraphQL type (required by the `InternalObject` interface it implements).
* Cascade deletion: background task cleans stored values on all entities when a definition is deleted, or when a definition is detached from one of its entity types.
* Edition guards: block integer min/max bound edits that would invalidate already-stored values; block removal of a select option still in use by existing values.
* Add STIX export/import utilities: `flattenCustomFieldValuesForStix` (nested → flat for export) and `unflattenStixToCustomFieldValues` (flat → nested for ingestion, skips unknown fields, not yet wired into the ingestion pipeline — tracked for a later chunk). Export wiring done on the CaseIncident converter.
* Add `CUSTOM_FIELDS` feature flag (`ENABLED_FEATURE_FLAGS` / `isFeatureEnabled('CUSTOM_FIELDS')` in `conf.js`).

### Related issues

* closes #17178 

### How to test this PR

> This PR is backend-only — no admin UI (Chunk 2) or Case-Incident form/list UI wiring (Chunk 3) is included or required. Every step below is done directly against the GraphQL API (e.g. via the GraphQL Playground at `/graphql`, Insomnia, or Postman) for creating/attaching/setting custom fields, since no screen exists yet to do this by clicking around. Steps 4, 5 and 7 reuse pre-existing, generic OpenCTI screens (entity list filters, dashboard widget editor, export button) that are not part of this PR either — they already work off the schema/backend changes below.

1. Enable the feature flag: add `"CUSTOM_FIELDS"` to `app.enabled_dev_features` in your platform config (see `config/development.json` for an example), or via env var equivalent for that config path.
2. **CRUD definitions**: Use the GraphQL API to create a `CustomFieldDefinition` (e.g. `customFieldDefinitionAdd` with `name: "x_opencti_cf_score"`, `field_type: "integer"`, `mandatory`/`default_value` set via `customFieldDefinitionAddEntityType` for a given `entityType`). Verify list/edit/delete work. Verify `field_type` and `name` are immutable after creation.
3. **Validation**: Create a CaseIncident with `customFieldValues` via the `caseIncidentAdd` GraphQL mutation directly — verify type checking (wrong type rejected), integer bounds, select options enforcement, mandatory field presence (including when submitting zero values), and duplicate detection.
4. **Filtering**: In the (pre-existing, generic) CaseIncident list UI, verify `x_opencti_cf_score` is filterable (including `contains`/`not_contains` on string/select fields) and returns correct results, on entities created via step 3.
5. **Distribution**: Using the (pre-existing, generic) dashboard widget editor, create a widget with distribution on `x_opencti_cf_score` for CaseIncidents — verify the nested aggregation returns correct buckets.
6. **Cascade deletion**: Delete a `CustomFieldDefinition` (or detach it from one entity type) that has stored values — verify the values are removed from existing entities.
7. **STIX export**: Export a CaseIncident with custom field values (pre-existing, generic export action) and verify the STIX bundle contains `x_opencti_cf_score: <value>` in the extension properties (flat format). Note: the reverse (STIX **import** flattening values back into `custom_field_values`) is implemented as a utility but not yet wired into the ingestion pipeline.

### Checklist

### Further comments

This is the backend portion of **Chunk 1 — "CF for a single entity"** from the custom fields roadmap (Notion). It replaces the POC branch (`issue/6177`) which used hardcoded hacks (`CF_SCORE_KEY`, `CF_COMMENT_KEY`) with a fully dynamic architecture based on an in-memory cache of definitions.

**Key design decisions:**
- **Dynamic over hardcoded**: all POC `FIXME hack` patterns replaced with cache lookups from `getCustomFieldDefinitionByName()` / `getCustomFieldDefinitionsForEntityType()`.
- **Immutable fields**: `field_type` and `name` cannot be changed after creation to avoid data corruption in existing entities.
- **Per-entity-type settings**: `mandatory` / `default_value` live on the entity-type attachment (`entity_type_settings`), not on the definition itself — the same field can be mandatory on one entity type and optional on another.
- **Cache pattern**: definitions are loaded once at boot (`loadCustomFieldDefinitions` in `initialization.js`) and refreshed after every CRUD mutation — no DB round-trip on read paths.
- **STIX ingestion**: unknown custom fields are silently skipped (no auto-creation of definitions from STIX) — this behavior is already the case today since ingestion doesn't parse `x_opencti_cf_*` extension properties at all yet (utility exists, not wired).

**Not in scope (future chunks):**
- Frontend UI for custom field management and entity forms (Chunk 2/3)
- STIX ingestion wiring (flat → nested on incoming bundles) and Python client/worker support (Chunk 5)
- Live Stream validation (Chunk 5)
- Rollout to entity types beyond Case-Incident (Chunk 4)


CelineSebe:
All the tests works well:
1-CustomFieldDefinition CRUD: creation, listing, allowed patch/update, deletion.
2-Immutability: field_type and name are correctly locked after creation.
3-caseIncidentAdd validation: type checking, mandatory fields (including empty array submissions), select option enforcement, duplicate detection, and bounds checking when bounds are defined.

CelineSebe (approved):

opencti-platform/opencti-graphql/src/modules/customField/custom-field-domain.ts
```diff
@@ -0,0 +1,389 @@
...
+ * Load all custom field definitions from the database into memory.
+ * Called at platform startup and when definitions are modified.
+ */
+export const loadCustomFieldDefinitions = async (context: AuthContext): Promise<void> => {
+  const definitions = await fullEntitiesList<BasicStoreEntityCustomFieldDefinition>(
+    context,
+    SYSTEM_USER,
+    [ENTITY_TYPE_CUSTOM_FIELD_DEFINITION],
+  );
+  setCustomFieldDefinitionsCache(definitions);
```
marieflorescontact: why do we need to set a new cache here ? why not using the usual cache ?

opencti-platform/opencti-graphql/src/modules/customField/custom-field-domain.ts
```diff
@@ -0,0 +1,389 @@
...
+    user,
+    event_type: 'mutation',
+    event_scope: 'create',
+    event_access: 'administration',
+    message: `creates custom field definition \`${input.name}\``,
+    context_data: { id: created.id, entity_type: ENTITY_TYPE_CUSTOM_FIELD_DEFINITION, input },
+  });
+  // Reload the cache after creation
+  await loadCustomFieldDefinitions(context);
+  return notify(BUS_TOPICS[ABSTRACT_INTERNAL_OBJECT].ADDED_TOPIC, created, user);
```
marieflorescontact, line 117: I really don't understand why we use 2 caches . why don't we call
notify(BUS_TOPICS[ENTITY_TYPE_CUSTOM_FIELD_DEFINITION].ADDED_TOPIC, created, user);

after having initialize the cache in cacheManager.ts
by adding  writeCacheForEntity(ENTITY_TYPE_CUSTOM_FIELD_DEFINITION, platformDecayExclusionRules(context));

opencti-platform/opencti-graphql/src/manager/taskManager.js
```diff
@@ -543,6 +544,45 @@ const sharingOperationCallback = async (context, user, task, actionType, operati
...
 
+// Cascade cleanup of a deleted custom field definition: for each entity still holding a value
+// for the deleted field, rebuild its custom_field_values without the removed field and patch it.
+const customFieldValuesRemoveOperationCallback = async (context, user, task, operations) => {
+  const fieldName = operations[0]?.context?.values?.[0];
+  let totalProcessed = task.task_processed_number;
+  return async (elements) => {
+    const objects = [];
+    const ids = elements.map((e) => e.internal_id);
+    const loadedElements = await internalFindByIds(context, user, ids);
```
marieflorescontact: why do we need to reload all elements with internalFindByIds, it will return exactly the same as elements. Am I missing something ?

opencti-platform/opencti-graphql/src/modules/attributes/stixDomainObject-registrationAttributes.ts
```diff
@@ -352,3 +353,29 @@ const stixDomainObjectsAttributes: { [k: string]: Array<AttributeDefinition<any>
...
+    type: 'object',
+    format: 'nested',
+    mandatoryType: 'no',
+    editDefault: false,
+    multiple: true,
+    upsert: false,
+    isFilterable: false,
+    featureFlag: CUSTOM_FIELDS_FEATURE_FLAG,
+    mappings: [
+      { name: 'field_id', label: 'Field ID', type: 'string', format: 'short', mandatoryType: 'internal', upsert: false, editDefault: false, multiple: false, isFilterable: false },
```
marieflorescontact, line 371: shouldn't it be `upsert: true `?

opencti-platform/opencti-graphql/src/modules/case/case-incident/case-incident-converter.ts
```diff
@@ -21,6 +22,7 @@ export const convertCaseIncidentToStix_2_1 = (instance: StoreEntityCaseIncident)
       [STIX_EXT_OCTI]: cleanObject({
         ...caseIncident.extensions[STIX_EXT_OCTI],
         extension_type: 'new-sdo',
+        ...flattenCustomFieldValuesForStix(instance.custom_field_values),
```
marieflorescontact: I think flattenCustomFieldValuesForStix(instance.custom_field_values) should be called in buildStixDomain(instance)
should it be also added in stix 2.0 converter ?

opencti-platform/opencti-graphql/src/modules/attributes/stixDomainObject-registrationAttributes.ts
ludovic: Good catch, but this has no effect yet: connector-based STIX ingestion of custom fields isn't wired in any of the 3 chunks (unflattenStixToCustomFieldValues is currently unused). Leaving upsert: false for now to keep this chunk scoped to the GraphQL API; flagging as follow-up once ingestion is implemented.

opencti-platform/opencti-graphql/src/modules/case/case-incident/case-incident-converter.ts
ludovic: Good point, moved the call into `buildOCTIExtensions` (used by buildStixObject/buildStixDomain for all SDO/SRO/SCO types) so future entity types get it for free instead of each converter having to remember it. Checked` stix-2-0-converter.ts`: other recent extension fields like `pir_information`/`metrics` aren't backported there either, so following that same precedent — not adding custom fields to the legacy 2.0 export.

opencti-platform/opencti-graphql/src/manager/taskManager.js
ludovic: `elements` here come from taskQuery's `elList(..., { baseData: true })`, and `BASE_FIELDS` doesn't include `custom_field_values`. So elements don't actually have it — the reload fetches the full document. Added a comment to make this explicit.

opencti-platform/opencti-graphql/src/modules/customField/custom-field-domain.ts
ludovic: Good catch — I missed that because of a circular dependency issue in a unit test... I'm fixing it now.

opencti-platform/opencti-graphql/src/manager/taskManager.js
```diff
@@ -543,6 +544,47 @@ const sharingOperationCallback = async (context, user, task, actionType, operati
...
+      if (nextValues.length !== currentValues.length) {
+        objects.push({
+          id: element.standard_id,
+          type: convertTypeToStixType(element.entity_type),
+          extensions: {
+            [STIX_EXT_OCTI]: {
+              id: element.internal_id,
+              type: element.entity_type,
+              opencti_operation: 'patch',
+              opencti_field_patch: [{ key: 'custom_field_values', value: nextValues, operation: 'replace' }],
```
marieflorescontact: It would be better to use an `add` or `remove` instead of a `replace` to avoid data loss if a custom field is modified between the moment `nextValues` is calculated and the moment the worker processes the operation (this can take a while).

opencti-platform/opencti-front/lang/back/fr.json
```diff
@@ -647,6 +660,7 @@
   "Multi-match separator": "Séparateur multiple",
   "Multi-match strategy": "Stratégie multi-match",
   "Multipart": "Multipart",
+  "Multiple": "Plural",
```
marieflorescontact:
```suggestion
  "Multiple": "Multiple",
```

opencti-platform/opencti-graphql/src/manager/taskManager.js
marieflorescontact: you can use the `baseFields` in initial elList() to preventing reloading all elements here.

opencti-platform/opencti-graphql/src/modules/attributes/stixDomainObject-registrationAttributes.ts
marieflorescontact: if we create an entity that already exists , upsert will be necessary even without connector based stix ingestion

opencti-platform/opencti-graphql/src/modules/case/case-incident/case-incident-converter.ts
marieflorescontact: I think it is a mistake because if we don't add custom_field in stix_2_0_exporter (wich is now handle by Client pyhton) they won't be exported when we use connector json stix connector (export button in UI on all entites))

marieflorescontact (approved):

PR #16922, opened 2026-07-01
fix(jsonmapper): add maxRecords parameter to jsonMappingExecution (#16921)

### LINKED ISSUE #16921: fix(jsonmapper): Json mapper memory management improvement
## Description

Json mapper memory management improvement: limit the amount of memory used during json mapper test

## Environment

2. Version: { e.g. 1.2.3 }
3. Other environment details:

## Reproducible steps

1. { e.g. Run ... }
2. { e.g. Click ... }
3. { e.g. Error ... }

JeremyCloarec:
### Proposed changes

* add maxRecords parameter to jsonMappingExecution
*

### Related issues

* closes #16921

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/modules/ingestion/ingestion-json-domain.ts
+++ b/opencti-platform/opencti-graphql/src/modules/ingestion/ingestion-json-domain.ts
@@ -162,7 +162,8 @@ export const executeJsonQuery = async (context: AuthContext, ingestion: BasicSto
       });
       const paginationVariables = buildQueryObject(ingestion.query_attributes, { ...paginationData, ...responseHeaders }, false);
       nextExecutionState = { ...nextExecutionState, ...paginationVariables };
-      const paginationObjects = await jsonMappingExecution(context, ingestionUser || SYSTEM_USER, paginationData, jsonMapperParsed);
+      const maxObjects = maxResults === 0 ? 0 : maxResults - objects.length;
+      const paginationObjects = await jsonMappingExecution(context, ingestionUser || SYSTEM_USER, paginationData, jsonMapperParsed, maxObjects);
       if (paginationObjects.length > 0) {
         objects = objects.concat(paginationObjects);
       }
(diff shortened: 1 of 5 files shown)


opencti-platform/opencti-graphql/src/modules/ingestion/ingestion-json-domain.ts
```diff
@@ -162,7 +162,8 @@ export const executeJsonQuery = async (context: AuthContext, ingestion: BasicSto
       });
       const paginationVariables = buildQueryObject(ingestion.query_attributes, { ...paginationData, ...responseHeaders }, false);
       nextExecutionState = { ...nextExecutionState, ...paginationVariables };
-      const paginationObjects = await jsonMappingExecution(context, ingestionUser || SYSTEM_USER, paginationData, jsonMapperParsed);
+      const maxObjects = maxResults === 0 ? 0 : maxResults - objects.length;
```
lndrtrbn, line 181: If the objective is to not have negative value, shouldn't be: `const maxObjects = Math.max(0, maxResults - objects.length)` ?

JeremyCloarec: The Math.max is equivalent yes, but I find the ternary condition easier to read, I don't know about you? Basically here maxResult === 0 means we don't have a limit on max number of results => we keep 0, otherwise we compute the remaining amount of results to get. Doing it with the Math.max it's a bit harder to understand the case where maxResult === 0 I feel like

lndrtrbn: Well it's not equivalent because you can have negative value for maxObjects using the ternary, which is not possible with `Math.max`. I think the more robust solution is Math.max + a comment explaining that 0 means no limit

lndrtrbn: Or you cannot have objects.length > maxResults, and in that case add a comment above the ternary to explain

JeremyCloarec: Yes you can't have objects.length > maxResults here, since there is the condition objects.length < maxResults above in the while condition on line 155
I can add a comment on that, but I think that's maybe a bit too much since the while condition is just a few lines above no?

fellowseb (approved):

PR #16935, opened 2026-07-02
feat(json-mapper): improve extractWithRegexp (#16929)

### LINKED ISSUE #16929: feat(json mapper): improve json mapper extraction regexp

OctaveLaventure:
### Proposed changes

- 

### Related issues

* closes https://github.com/OpenCTI-Platform/opencti/issues/16929

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/parser/json-mapper.ts
+++ b/opencti-platform/opencti-graphql/src/parser/json-mapper.ts
@@ -14,6 +14,7 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 */
 
 import * as JSONPath from 'jsonpath-plus';
+import RE2 from 're2';
 
 import { v4 as uuidv4 } from 'uuid';
 import type { StixObject } from '../types/stix-2-1-common';
@@ -90,7 +91,7 @@ const extractComplexPathFromJson = async (
     }
   };
   data.extractWithRegexp = (regexp: string, groupIndex: number, value: string) => {
-    const myRegexp = new RegExp(regexp, 'g');
+    const myRegexp = new RE2(regexp, 'g');
     const matches = myRegexp.exec(value);
     if (matches != null) {
       return matches[groupIndex];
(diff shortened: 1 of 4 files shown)


opencti-platform/opencti-graphql/src/parser/json-mapper.ts
```diff
@@ -90,7 +91,7 @@ const extractComplexPathFromJson = async (
     }
   };
   data.extractWithRegexp = (regexp: string, groupIndex: number, value: string) => {
-    const myRegexp = new RegExp(regexp, 'g');
+    const myRegexp = new RE2(regexp, 'g');
     const matches = myRegexp.exec(value);
     if (matches != null) {
       return matches[groupIndex];
```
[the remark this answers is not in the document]
OctaveLaventure: fixed

opencti-platform/opencti-graphql/src/parser/json-mapper.ts
```diff
@@ -90,9 +91,9 @@ const extractComplexPathFromJson = async (
     }
   };
   data.extractWithRegexp = (regexp: string, groupIndex: number, value: string) => {
-    const myRegexp = new RegExp(regexp, 'g');
+    const myRegexp = new RE2(regexp, 'g');
     const matches = myRegexp.exec(value);
-    if (matches != null) {
+    if (matches != null && matches[groupIndex] != null) {
```
mcartaud, line 96: When looking at the type of matches, I find 3 different results and only 1 with a groupIndex property (from lib.es2018, other are from lib.es2022 and lib.es5).
Are you sure this is correct, because it seem we are targeting es2022+?

mcartaud: Seen together, final type is a combination of all three

mcartaud (approved):

opencti-platform/opencti-graphql/src/parser/json-mapper.ts
OctaveLaventure: as seen together the different types declarations are merged into one thanks to TS [declaration merging](https://www.typescriptlang.org/docs/handbook/declaration-merging.html), and RegExpExecArray is extending Array with a forced `string` type but it's not true at runtime as array out of bound index access will still return undefined

PR #16944, opened 2026-07-02
feat(workflow): add reset functionality to workflow management (#16812)

### LINKED ISSUE #16812: feat(Workflow): Ability to fully reset  a workflow
## Use case

It should be possible to fully reset a draft workflow in order to easily restart it from scratch.
## Proposed solution

Add a specific operation to reset the full worfkow. Publication might fail because it contains some delete statuses, but at least users will have the possibility to easily start from scratch. 

delemaf:
### Proposed changes
* Add reset functionality to workflow management

### Related issues
* closes #16812

### Checklist


OctaveLaventure (approved):

opencti-platform/opencti-front/src/private/components/settings/sub_types/workflow/Workflow.tsx
```diff
@@ -207,17 +207,14 @@ const Workflow = ({ queryRef, depsQueryRef }: { queryRef: PreloadedQuery<SubType
...
-      states: [],
-      transitions: [],
-    });
-    // Keep local state and autosave guard aligned with the empty graph state.
-    previousSchemaRef.current = JSON.stringify(transformToWorkflowDefinition([], [], workflowDefinition));
+    // Derive the schema from transformToWorkflowDefinition so that the payload and
+    // the autosave guard are always aligned: the effect recomputes the same function
+    // after setNodes([]) / setEdges([]) and will see an identical string, preventing
+    // a spurious follow-up mutation.
+    const emptySchemaString = JSON.stringify(transformToWorkflowDefinition([], [], workflowDefinition));
```
mcartaud, line 214: While I agree the fixes makes the values from the ref and on the `variables.definition` identical, which is good, it seem it will bring more a "undo to server state" functionality rather that a full reset : workflowDefinition comes from the server and will have a non null `initialState`. Perhaps passing the empty schema to the ref and `variables.definition` will be a better fix. 
Or if current worflowDefinition id and name are not important, an empty object should be enough

PR #16953, opened 2026-07-03
feat(custom-fields): Chunk 2 — Admin UI (Settings) (#17179)

### LINKED ISSUE #17179: Chunk 2 — Admin UI (Settings)
📋 Spec reference: 🚡 Add custom fields — MVP US.1 (create) + US.2 (attribute) + US.3 (edit/delete rules)

> Goal: Allow administrators to create, edit, delete and attribute custom field definitions from the UI (Settings), without using the raw GraphQL API.

Scope — A. Global admin screen (Settings > Taxonomies > Custom Fields)

Scope — B. Attribution per entity type (Settings > Customization > entity type > Attributes)

Backend support

Note on the model: mandatory and default_value are per entity type (US.2), not global on the definition. The global creation drawer (Scope A) only sets field-intrinsic constraints (type, min/max, options); mandatory + default_value are set at attribution time (Scope B).

Out of current scope (product decisions needed)

ludovic:
### Proposed changes

**A. Global admin screen (`Settings > Taxonomies > Custom Fields`)**
* New "Custom Fields" entry in the Taxonomies menu (gated by `SETTINGS_SETCUSTOMIZATION` + `CUSTOM_FIELDS` feature flag).
* List screen (`CustomFields.tsx`, DataTable): Label, Technical name, Entity types, Field type, with search + sort.
* Creation drawer (`CustomFieldCreation.tsx`): technical name auto-prefixed `x_opencti_cf_` (regex-validated), label, field type (Number / Text / Boolean / Date / Selection list / Multi-select / Markdown), description, field-intrinsic constraints (min/max for integer, options for select/multi_select). `mandatory` / `default_value` are **not** set here — they're per entity type (Scope B).
* Edition drawer (`CustomFieldEdition.tsx`, inline submit): editable label, description, min/max, select options — `field_type` and technical name are locked/immutable after creation.
* Delete (`CustomFieldPopover.tsx`) with a destructive confirmation dialog — removes the definition and triggers the cascade cleanup of stored values on all entities (background task from Chunk 1).

**B. Attribution per entity type (`Settings > Customization > entity type > Attributes`)**
* "Custom fields" card (`EntitySettingCustomFields.tsx`) listing the definitions attached to the current entity type: Label, Technical name, Field type, per-entity **Mandatory**, per-entity **Default value**.
* "Add custom field" dialog (`EntitySettingCustomFieldsAddDialog.tsx`): multi-select of existing definitions + a shared Mandatory toggle applied on attach (`customFieldDefinitionAddEntityType`).
* Per-row edit of the per-entity `mandatory` / `default_value` on an already-attached definition (`customFieldDefinitionUpdateEntityType`), with a type-aware default-value input (text/number/boolean/select/multi-select/markdown/date-time picker).
* Per-row remove with a destructive confirmation dialog — detaches the definition from this entity type only (`customFieldDefinitionRemoveEntityType`), distinct from the global delete in Scope A.
* **Date fields**: default value input supports a `@now` toggle (resolves to the current date/time at population time) in addition to a fixed date/time picker (new `DateTimePicker.tsx` common component, since custom field dates now carry time, not just a day).

**Backend support (new in this PR, on top of Chunk 1)**
* `customFieldDefinitionsForEntityType(entityType: String!)` query exposing `entity_type_settings` for entity forms to resolve `mandatory` + `default_value` per entity type.

### Related issues

* closes #17179

### How to test this PR

1. Enable the `CUSTOM_FIELDS` feature flag (see Chunk 1 PR for how).
2. **Global admin**: Go to `Settings > Taxonomies > Custom Fields`. Create a field of each type (integer, string, boolean, date, select, multi_select, markdown). Verify edit works and `field_type`/technical name are locked. Delete one and verify a confirmation dialog appears, warning that stored values will be removed.
3. **Attribution**: Go to `Settings > Customization > Case-Incident > Attributes`. Add one of the definitions created above, set it mandatory, set a default value. For a `date` field, verify you can choose between a fixed date/time or the `@now` toggle. Edit the per-entity mandatory/default value inline. Remove it from this entity type (verify the confirmation dialog mentions only this entity type, not a full delete).
4. Verify the custom field then appears correctly on the Case-Incident creation/edition/details screens (wired in Chunk 3).

### Checklist

### Further comments

This is the frontend admin UI portion of the custom fields feature (Notion roadmap), covering MVP US.1 (create), US.2 (attribute to entity type), and US.3 (edit/delete rules). It builds on the backend infra delivered in Chunk 1 and is consumed by the Case-Incident wiring in Chunk 3.

**Key design decisions:**
- **Two distinct screens for two distinct concerns**: the global admin screen (Scope A) only manages field-intrinsic properties (type, constraints) since those apply platform-wide; per-entity-type settings (`mandatory`, `default_value`) live exclusively in the entity's own Attributes tab (Scope B), since the same field can be mandatory on one entity type and optional on another.
- **Two separate delete actions**: a full delete (Scope A, removes the definition entirely) vs. a per-entity-type detach (Scope B, keeps the definition but stops applying it to this entity type) — both trigger the same cascade cleanup of stored values, scoped appropriately.
- **`@now` as a sentinel string**: rather than a schema/data-model change, `default_value` for date fields can store the literal string `@now`, resolved to the current date/time only at the point a form is populated — keeps the backend model untouched.


ValentinBouzinFiligran:
When creating a custom field of type Selection list / Multiple selection list and leaving Select options empty, the form correctly refuses to submit (as expected, since at least one option should be required). However, the field isn't highlighted in red / no visible validation error is shown, so there's no indication of why the creation silently fails. It would help to surface the "At least one option is required" validation message directly on the Select options field so users aren't left wondering what went wrong.

ValentinBouzinFiligran (approved):

PR #16981, opened 2026-07-06
fix(engine): avoid wrapping inner ES operations when calling retryElOperations (#16978)

### LINKED ISSUE #16978: fix: circuit_breaking_exception not triggering retry mechanism
## Description

The `elUpdate` primitive engine.ts function uses `retryElOperations`  but is wrapping the inner operation in a try/catch and rethrowing higher level `DatabaseError` exceptions with a different message and in consequence makes `retryElOperations` inoperent.

## Solution

Either one of the two:
- Change `retryElOperations` to look into the stringified `cause` 
- Stop try/catching and throwing different `DatabaseError` exceptions

fellowseb:
### Descrption

During on-call we've spotted some circuit breaking alerts triggered on `elUpdate` without the retry mechanism being effective even though the code seems to indicate it should be in place.

### Proposed changes

Modifies engine primitives that call `retryElOperations` after wrapping the ES operation in a try/catch block to rethrow a higher-level applicative Exception like `DatabaseError` as this makes `retryElOperations` ineffective. Instead, include the call to `retryElOperations` inside the try.

### Related issues

* closes https://github.com/OpenCTI-Platform/opencti/issues/16978

### How to test this PR

Run unit test suite.
Kinda super tricky trying to reproduce in a real application case.

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/database/engine.ts
+++ b/opencti-platform/opencti-graphql/src/database/engine.ts
@@ -529,44 +529,22 @@ const elExecuteWithAbortSignal = async (
 const BULK_MAX_RETRIES = 5;
 const BULK_INITIAL_DELAY_MS = 500;
 
-const collectErrorCandidates = (error: any): any[] => {
-  const candidates: any[] = [];
-  const queue: any[] = [error];
-  const visited = new Set<any>();
-  while (queue.length > 0) {
-    const current = queue.shift();
-    if (!current || visited.has(current)) {
-      continue;
-    }
-    visited.add(current);
-    candidates.push(current);
-    queue.push(current?.cause);
-    queue.push(current?.originalError);
-    queue.push(current?.extensions?.data?.cause);
-  }
-  return candidates;
-};
-
 const isTransitoryError = (error: any): boolean => {
-  const errorsToInspect = collectErrorCandidates(error);
-  for (let index = 0; index < errorsToInspect.length; index += 1) {
-    const currentError = errorsToInspect[index];
-    const statusCode = currentError?.statusCode ?? currentError?.meta?.statusCode ?? currentError?.status;
-    // 429: Too many requests, 503: Service unavailable, both can be transient and should be retried
-    if (statusCode === 429 || statusCode === 503) {
-      return true;
-    }
-    const errorCode = currentError?.code;
-    // All these error codes are commonly associated with transient issues that can occur in network communication
-    // or when the search engine is under heavy load, and thus are good candidates for retrying the operation.
-    if (['ECONNRESET', 'ECONNREFUSED', 'ETIMEDOUT', 'EPIPE', 'EAI_AGAIN'].includes(errorCode)) {
-      return true;
-    }
-    const errorMessage = currentError?.message ?? '';
-    // All these error messages are commonly associated with transient issues that can occur when the search engine is under heavy load
-    if (/circuit_breaking_exception|es_rejected_execution|too_many_requests|service_unavailable/i.test(errorMessage)) {
-      return true;
-    }
+  const statusCode = error?.statusCode ?? error?.meta?.statusCode ?? error?.status;
+  // 429: Too many requests, 503: Service unavailable, both can be transient and should be retried
+  if (statusCode === 429 || statusCode === 503) {
+    return true;
+  }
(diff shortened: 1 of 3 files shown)


opencti-platform/opencti-graphql/src/database/engine.ts
```diff
@@ -4050,27 +4050,25 @@ export const elUpdate = async (
...
+    return await elExecuteWithAbortSignal(
+      context?.requestAbortSignal,
+      (opts) => (engine as ElkClient).update(updateRequest, opts),
+      () => (engine as OpenClient).update(updateRequest),
+    );
   };
-  return retryElOperations(updateOperation);
+  return retryElOperations(updateOperation).catch((err: any) => {
+    throw DatabaseError('Update indexing fail', { cause: err, documentId, entityType, ...extendedErrors({ documentBody }) });
+  });
```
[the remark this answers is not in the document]
fellowseb: Done

opencti-platform/opencti-graphql/tests/01-unit/database/engine-retry-test.ts
```diff
@@ -0,0 +1,179 @@
...
+
+  beforeEach(async () => {
+    vi.clearAllMocks();
+    testMocks.info.mockResolvedValue({
+      version: { number: '8.19.1' },
+      tagline: 'You Know, for Search',
+    });
+    testMocks.putPipeline.mockResolvedValue({});
+    await searchEngineInit();
+  });
```
fellowseb, line 100: Unfortunately the current state of engine.ts makes it hard to test without such mocks.
A refactor to make it more testable would be a good idea.

opencti-platform/opencti-graphql/src/database/engine.ts
```diff
@@ -4130,23 +4094,21 @@ export const elReplace = async (
...
-      id: documentId,
-      index: indexName,
-      timeout: BULK_TIMEOUT,
-      refresh: true,
-    };
-    try {
-      if (engine instanceof ElkClient) {
-        return await engine.delete(deleteRequest);
-      }
+    if (engine instanceof ElkClient) {
```
gabrielebonadiman, line 4104: Small nit: both branches of the  instanceof  check do the same thing here, could we just drop the `if` and return 
```return await  engine.delete(deleteRequest)```
directly? Not blocking, just simplifies things a bit further.

Comment (fellowseb) [reply opencti-platform/opencti-graphql/src/database/engine.ts:4104]: Yes this appears in multiple places throughout `engine.ts` : it's due to TS not able to reconciliate union type of ElasticSearch and OpenSearch signatures:

```
  Each member of the union type '{ (this: That, params: DeleteRequest | DeleteRequest, options?: TransportRequestOptionsWithOutMeta | undefined): Promise<...>; (this: That, params: DeleteRequest | DeleteRequest, options?: TransportRequestOptionsWithMeta | undefined): Promise<...>; (this: That, params: DeleteRequest | DeleteRequest, options?: Transp...' has signatures, but none of those signatures are compatible with each other.
```

OS types for the `delete` operation:
```
  delete (params: API.Delete_Request, options?: TransportRequestOptions): TransportRequestPromise<API.Delete_Response>;
  delete (params: API.Delete_Request, callback: callbackFn<API.Delete_Response>): TransportRequestCallback;
  delete (params: API.Delete_Request, options: TransportRequestOptions, callback: callbackFn<API.Delete_Response>): TransportRequestCallback;
```

ES types for the same operation:
```
export default function DeleteApi(this: That, params: T.DeleteRequest | TB.DeleteRequest, options?: TransportRequestOptionsWithOutMeta): Promise<T.DeleteResponse>;
export default function DeleteApi(this: That, params: T.DeleteRequest | TB.DeleteRequest, options?: TransportRequestOptionsWithMeta): Promise<TransportResult<T.DeleteResponse, unknown>>;
export default function DeleteApi(this: That, params: T.DeleteRequest | TB.DeleteRequest, options?: TransportRequestOptions): Promise<T.DeleteResponse>;
```

fellowseb: There's no easy way to resolve this and it's out of scope fo this PR (:

gabrielebonadiman: 👍 thanks for this.

PR #16986, opened 2026-07-06
fix(engine): apply retryElOperations to elIndex (#16979)

### LINKED ISSUE #16979: fix(engine): elIndex does not benefit from retryElOperations utility
## Description

`elIndex` does not benefit from the `retryElOperations` strategy; thus can lead to circuit breaker alerts.

Is this an intended omition ? If not let's add it given we've spotted circuit breaking alerts during on-call for such use case.

fellowseb:
### Description

We've noticed during on-call that `elIndex` does not currently benefit from the retry mechanism and thus can lead to circuit breaking errors on the first attempt. This PR attempts to apply that retry mechanism.

### Proposed changes

* Wrapped primitive ES call in `retryElOperations`
* Updated unit tests suite

### Related issues

* closes #16979 

### How to test this PR

Run unit test suite.
Kinda super tricky trying to reproduce in a real application case.

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/database/engine.ts
+++ b/opencti-platform/opencti-graphql/src/database/engine.ts
@@ -4030,15 +4030,17 @@ export const elIndex = async (
   if (pipeline) {
     indexParams = { ...indexParams, pipeline };
   }
-  if (engine instanceof ElkClient) {
-    await engine.index(indexParams).catch((err: any) => {
-      throw DatabaseError('Simple indexing fail', { cause: err, documentId, entityType, ...extendedErrors({ documentBody }) });
-    });
-  } else {
-    await engine.index(indexParams).catch((err: any) => {
-      throw DatabaseError('Simple indexing fail', { cause: err, documentId, entityType, ...extendedErrors({ documentBody }) });
-    });
-  }
+  const indexOperation = async () => {
+    // Branching kept to please tsc
+    if (engine instanceof ElkClient) {
+      return await engine.index(indexParams);
+    } else {
+      return await engine.index(indexParams);
+    }
+  };
+  await retryElOperations(indexOperation).catch((err: any) => {
+    throw DatabaseError('Simple indexing fail', { cause: err, documentId, entityType, ...extendedErrors({ documentBody }) });
+  });
 
   return documentBody;
 };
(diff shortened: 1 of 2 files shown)


opencti-platform/opencti-graphql/src/database/engine.ts
```diff
@@ -4030,15 +4030,17 @@ export const elIndex = async (
...
+    // Branching kept to please tsc
+    if (engine instanceof ElkClient) {
+      return await engine.index(indexParams);
+    } else {
+      return await engine.index(indexParams);
+    }
+  };
+  await retryElOperations(indexOperation).catch((err: any) => {
+    throw DatabaseError('Simple indexing fail', { cause: err, documentId, entityType, ...extendedErrors({ documentBody }) });
+  });
```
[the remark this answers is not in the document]
fellowseb: DOne

gabrielebonadiman (approved):

JeremyCloarec (approved):

opencti-platform/opencti-graphql/src/database/engine.ts
```diff
@@ -4030,15 +4030,17 @@ export const elIndex = async (
...
-  }
+  const indexOperation = async () => {
+    // Branching kept to please tsc
+    if (engine instanceof ElkClient) {
+      return await engine.index(indexParams);
+    } else {
+      return await engine.index(indexParams);
+    }
+  };
+  await retryElOperations(indexOperation).catch((err: any) => {
```
T0t1n, line 4041: Not blocking, but one side effect worth considering: in elIndexFiles (file-search.js), the first elIndex call indexes the file content through the attachment pipeline, with file_data potentially being several MB. When it fails we fall back to indexing without the content.

With this change, if the failure is a circuit_breaking_exception caused by the payload itself (file too large for the ES available memry), we'ml now replay the exact same big request 5 times against an engine already under memory pressure, before going to the fallback. On that specific path the error is not really transitory.

Not sure it's worth handling in this PR, but maybe the file-content indexing path should skip the retry (or bail out early on circuit breaker) and go straight to the fallback. 

If circuit breaking errors persist after this fix, we should pay attention to the correlation of these log messages:
`logApp.error('Error on file indexing', ...)` in `elIndexFiles(...)`
and
`logApp.warn('[SEARCH] Engine transitory error, ...')` in `retryElOperations(...)`

fellowseb: Good point. I'll take a look and see if we need an opt-out option.

Issue #16708, opened 2026-06-19
fix(scope): indicator uppercase/lowercase pattern is modified by playbooks when extracting observables causing deduplication to fail


emmafiligran:
# Description

When an indicator is created with an uppercase value (for example a SHA-256 hash such as `E5F6D9D405819E6B05B5D8268A2E973294859AD65237EDE36AB612B536D0AC2D`), the **"Extract observables from indicator"** action creates the corresponding observable in lowercase.

As a result, when a subsequent playbook promotes this observable back into an indicator, the newly created indicator is also lowercase. Since deduplication appears to be case-sensitive, the new indicator is not deduplicated against the original uppercase indicator.

This leads to duplicate indicators representing the exact same value, differing only by letter casing.

# Steps to Reproduce

1. Create an indicator using an uppercase hash value.
2. Create a playbook triggered on indicator creation that:

   * Executes **"Extract observables from indicator"**
   * Adds a label to trigger the second playbook

3. Create a second playbook triggered by that label that:

   * Promotes the observable to an indicator.

4. Observe the value of the newly created indicator.

# Expected Behavior

The observable generated by **"Extract observables from indicator"** should preserve the original casing (uppercase in this example), or alternatively indicator values should be normalized consistently before deduplication.

This would ensure that when the observable is later promoted back to an indicator, the deduplication mechanism correctly identifies it as the same indicator.

# Actual Behavior

The observable created by playbook via **"Extract observables from indicator"** is converted to lowercase.

When this observable is promoted back to an indicator:

* The new indicator is lowercase.
* The original indicator remains uppercase.
* Deduplication does not occur.
* Two indicators are created for the same hash value.

# Impact

Duplicate indicators can be generated through automated playbook workflows, creating inconsistencies in the platform and reducing the effectiveness of the deduplication process and merging of information. 

# Additional Information

Please reach out directly if you need:

* Internal links to the playbooks
* Example entities demonstrating the issue
* Additional logs or screenshots

JeremyCloarec:
I'm not sure what the proper fix is here product wise: do we want to make indicator deduplication case insensitive or do we want to make observable/indicator generation keep the original lower/uppercases? @nino-filigran

nino-filigran:
Indeed, theproblem is not the deduplication. It's that the IOC/observable generation should keep the original lower/upper case.

nino-filigran:
After internal discussion, all hashes need to be case sensitive therefore we'll implement a fix that goes in this direction.

PR #17002, opened 2026-07-06
Revert "fix(backend): preserve original case for unknown hash algorithms (#16708)"

gabrielebonadiman:
Reverts OpenCTI-Platform/opencti#16948


gabrielebonadiman:
Closed in https://github.com/OpenCTI-Platform/opencti/pull/16948

PR #17028, opened 2026-07-07
fix(decay-rules): change decay excluded indicator creation (#13752)

### LINKED ISSUE #13752: fix(decay-rules): decay exclusions should set valid_until much later and not get revoked
## Description

With the new decay rule exclusion feature we are trying to utilizing it to say if the indicator has the Author = Our Org to exclude it from decaying. This part seems to work as we see the Lifecycle button on these indicators as org instead of blue now. 

However, we will see the "valid until" time matching our decay rule lifetime value of 45 days. Since this is ~45 days in the future I am not sure what will happen when this date hits. Since the excluded indicator valid until is ~45 days in the future we are unsure if it will become revoked at this time or not which would remove it from our platforms connected via stream connectors. We cannot let decay excluded indicators get revoked this way in order to utilize this feature.

Image

The other issue we already see with this is our external platforms like CrowdStrike add the expiration date from the valid until time which would impact it being valid in external platforms. Not sure if it is possible to say the year 9999 so it realistically won't hit that time or make it so this valid time will not impact these excluded indicators.

Image

## Environment

1. OS (where OpenCTI server runs): CentOS
2. OpenCTI version: 6.9.1
3. OpenCTI client: N/A
4. Other environment details:

## Reproducible Steps

1. Create a decay rule exclusion
2. Create an indicator that would match that exclusion
3. Observe the indicators Valid Until date

## Expected Output

Excluded indicators from the decay rule to have valid until an infinite amount of time so they do not get revoked.

## Actual Output

Image

JeremyCloarec:
### Proposed changes

This pull request makes a targeted adjustment to the indicator creation logic, specifically when a decay rule exclusion is applied. The main change ensures that the `valid_until` field is set to a special constant value and updates the order and source of certain indicator properties for consistency.

Key changes:

**Indicator creation logic:**
* When a decay rule exclusion is applied, the `valid_until` field is now explicitly set to `UNTIL_END_STR`, and the indicator's properties are merged from `resolvedIndicator` instead of the previous object. Additionally, `valid_from` is set using `toISOString()` for consistency.

**Imports update:**
* The `UNTIL_END_STR` constant is imported from `../../utils/format` to support the new logic for `valid_until`.

### Related issues

* closes #13752

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/modules/indicator/indicator-domain.ts
+++ b/opencti-platform/opencti-graphql/src/modules/indicator/indicator-domain.ts
@@ -52,7 +52,7 @@ import {
   checkDecayRules,
 } from '../decayRule/decayRule-domain';
 import { stixDomainObjectEditField } from '../../domain/stixDomainObject';
-import { checkScore, prepareDate, utcDate } from '../../utils/format';
+import { checkScore, prepareDate, UNTIL_END_STR, utcDate } from '../../utils/format';
 import { checkObservableValue, isCacheEmpty } from '../../database/exclusionListCache';
 import { stixHashesToInput } from '../../schema/fieldDataAdapter';
 import { REVOKED, VALID_FROM, VALID_UNTIL, X_DETECTION, X_SCORE } from '../../schema/identifier';
@@ -309,7 +309,9 @@ export const addIndicator = async (context: AuthContext, user: AuthUser, indicat
 
   if (isDecayActivated && exclusionRule) {
     finalIndicatorToCreate = {
-      ...indicatorToCreate,
+      valid_until: UNTIL_END_STR,
+      ...resolvedIndicator,
+      valid_from: validFrom.toISOString(),
       decay_exclusion_applied_rule: {
         decay_exclusion_id: exclusionRule.id,
         decay_exclusion_name: exclusionRule.name,


opencti-platform/opencti-graphql/src/modules/indicator/indicator-domain.ts
```diff
@@ -306,7 +306,9 @@ export const addIndicator = async (context: AuthContext, user: AuthUser, indicat
 
   if (isDecayActivated && exclusionRule) {
     finalIndicatorToCreate = {
-      ...indicatorToCreate,
+      valid_until: UNTIL_END_STR,
+      ...resolvedIndicator,
+      valid_from: validFrom.toISOString(),
       decay_exclusion_applied_rule: {
```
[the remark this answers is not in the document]
JeremyCloarec: it's intended

opencti-platform/opencti-graphql/src/modules/indicator/indicator-domain.ts
```diff
@@ -309,7 +309,9 @@ export const addIndicator = async (context: AuthContext, user: AuthUser, indicat
 
   if (isDecayActivated && exclusionRule) {
     finalIndicatorToCreate = {
-      ...indicatorToCreate,
+      valid_until: UNTIL_END_STR,
```
gabrielebonadiman, line 312: I hope I am not missing something here but I think there is a spread order bug.  `...resolvedIndicator`  comes after  `valid_until: UNTIL_END_STR` right?  so if  `resolvedIndicator.valid_until` is set it silently overwrites the sentinel. This reproduces the exact bug from #13752

```suggestion
finalIndicatorToCreate = {
  ...resolvedIndicator,
  valid_from: validFrom.toISOString(),
  valid_until: UNTIL_END_STR,
  decay_exclusion_applied_rule: { ... },
};
```

Could we also add a regression test? :)

JeremyCloarec: Will work on a test 👍 
The spread order is intended, but I'm waiting on the input of the product team to check if the behavior is what they are expecting.
The order of the spread is like that, because we still want to use the valid_until from the input if it's set. We only want to set the valid_until to UNTIL_END_STR if there is no valid_until in the input: before the fix, it was always using indicatorToCreate in the spread (so always setting a valid_until coming from the decay)

aHenryJard (commented):
I commented the issue, I'm not sure about the direction on this one.

aHenryJard (commented):
I think that this PR code is not up to date with latest discussion with Nino. The issue is more on the revoke part that should not be executed on excluded indicators we don't need an additional valid until.

JeremyCloarec:
> I think that this PR code is not up to date with latest discussion with Nino. The issue is more on the revoke part that should not be executed on excluded indicators we don't need an additional valid until.

Unless I missed something, my understanding is that the current PR implementation is what is intended from product team. Here is the relevant discussion:

Me: Hi 
I'm lost on this issue. Latest discussion I had with you Nino was:
Re nino! J'ai cette issue qui traine [fix(decay-rules): decay exclusions should set valid_until much later and not get revoked · Issue #1…](https://github.com/OpenCTI-Platform/opencti/issues/13752#issuecomment-4905408384), et j'aurais besoin de ton avis pour savoir exactement sur quel fix partir. Le user veut que si son indicator est dans une decay exclusion rule, il n'ait pas de valid_until et qu'il ne decay pas. Ca OK, mais quel devrait être le comportement si l'indicator est dans une decay exclusion rule mais possède un valid_until? Et quid de quand le decay manager est disabled? Actuellement avec ma PR, le comportement est comme ca:
Decay manager ON, decay exclusion rule ON, valid_until dans l'input ON: => on prend le valid_until de l'input
Decay manager ON, decay exclusion rule ON, valid_until dans l'input OFF: => on met le valid_until à l'année 5138
Decay manager ON, decay exclusion rule OFF => on prend le valid_until calculé par decay
Decay manager OFF => on prend quand même le valid_until calculé par le decay
C'est surtout sur le point 4 que c'est pas clair => est ce que dans ce cas là on veut aussi mettre l'année 5138?
 
Nino: Je pense que le point 4 c'est bon (et globalement cette approche se défend) car ça permet quand même de revoke quelque chose par défault. Tu risque autrement de te retrouver avec pleins d'IOC qui ne sont jamais revoke. 
 
Me: Ok, c'est le comportement de mon fix alors Peut etre je te laisse mettre un point sur l'issue pour valider ce comportement, Angélique voulait aussi ton avis sur ce fonctionnement justement
 
Me: So in my understanding, a decay rule excluded indicator can still revoke if a valid_until is specified in input. Is it now incorrect? Do we want all decay rule excluded indictors to never be revoked, even if a specific valid_until date is specified?
fix(decay-rules): decay exclusions should set valid_until much later and not get revoked · Issue #13752 · OpenCTI-Platform/opencti
Description With the new decay rule exclusion feature we are trying to utilizing it to say if the indicator has the Author = Our Org to exclude it from decaying. This part seems to work as we see t...
 
Nino: No it's not incorrect for me
IF a date is specified, we'll follow it. IT means that the score will not change while the IOC is still valid. But if the users would like to send "forever" valid IOCs, then they send them without valid until & they ensure they are taken by exclusion rules.

gabrielebonadiman (commented):
This looks good to me overall. One tiny add: it might be worth adding a regression test to back this up against future changes.

PR #17046, opened 2026-07-08
fix(ingestion): improve ingestion validation (#17020)

### LINKED ISSUE #17020: feat(ingestion): improve ingestion validation

mcartaud:
### Related issues

* closes #17020 

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/modules/ingestion/ingestion-json-domain.ts
+++ b/opencti-platform/opencti-graphql/src/modules/ingestion/ingestion-json-domain.ts
@@ -78,7 +78,18 @@ const buildQueryObject = (queryParamsAttributes: Array<DataParam> | undefined, r
   return params;
 };
 
-const replaceVariables = (body: string, variables: Record<string, object | string>) => {
+/**
+ * Normalises the result of getValueFromPath (typed as any) to a string or null.
+ * JSONPath can return an array — in that case the first element is used.
+ * Non-string, non-array values (objects, numbers, …) are rejected (return null).
+ */
+const toUrlString = (value: unknown): string | null => {
+  const candidate = Array.isArray(value) ? value[0] : value;
+  return typeof candidate === 'string' ? candidate : null;
+};
+
+const replaceVariables = (body: string | undefined | null, variables: Record<string, object | string>) => {
+  if (body == null) return '';
   const regex = /\$\w+/g;
   return body.replace(regex, (match) => {
     const variableName = match.substring(1);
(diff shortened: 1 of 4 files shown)


opencti-platform/opencti-graphql/src/modules/ingestion/ingestion-json-domain.ts
```diff
@@ -160,13 +178,14 @@ export const executeJsonQuery = async (context: AuthContext, ingestion: BasicSto
   let nextExecutionState = buildQueryObject(ingestion.query_attributes, { ...requestData, ...responseHeaders }, false);
   // region Try to paginate with next page style
   if (ingestion.pagination_with_sub_page && isNotEmptyField(ingestion.pagination_with_sub_page_attribute_path)) {
-    let url = getValueFromPath(ingestion.pagination_with_sub_page_attribute_path, requestData);
+    let url = toUrlString(getValueFromPath(ingestion.pagination_with_sub_page_attribute_path, requestData));
     while (isNotEmptyField(url) && (maxResults === 0 || objects.length < maxResults)) {
+      verifyIngestionUri(url as string);
```
fellowseb: ~TypeScript is unable apply the narrowing here ?~

Ah no I was mixing up with the `isNotEmptyField` util in the frontend.
How about updating the implementation of `isNotEmptyField` and `isEmptyField` to match the frontend impl ?

opencti-platform/opencti-graphql/src/modules/ingestion/ingestion-json-domain.ts
```diff
@@ -133,11 +144,18 @@ export const executeJsonQuery = async (context: AuthContext, ingestion: BasicSto
     timeout,
     responseType: 'json',
     certificates,
+    beforeRedirect: (options) => {
+      if (typeof options.href === 'string') {
```
fellowseb, line 158: Here we only look at the `href` option but the docs (https://axios.rest/pages/advanced/request-config#beforeredirect) show there could be other options set, no ? Like `protocol`, `hostname`, `port`, `path`. Maybe it's better to build a url (`new URL`) from those instead ?

Actually I don't see where `href` would come from 🤔 . My understanding is that `options` should match the Node request options but to check.

mcartaud: Axios has a dependency on `follow-redirects` that preserve some fields in redirection, href being one of them  (added a comment to make it clear)
https://github.com/follow-redirects/follow-redirects/blob/0c23a223067201c368035e82954c11eb2578a33b/index.js#L37
https://github.com/axios/axios/blob/c3f553c740ebf3dff5e22dae24e9caaafafddd2d/package.json#L142

I can add a fallback building our own `href` but `protocol`, `hostname` and other are also preserved by follow-redirect so it brings no additional value IMO
What do you think?

fellowseb: It wasn't clear for me if `href` could be undefined but `hostname`, `protocol` etc defined but if it's guaranteed to be defined then yeah we're good

opencti-platform/opencti-graphql/src/modules/ingestion/ingestion-json-domain.ts
```diff
@@ -163,27 +177,27 @@ export const executeJsonQuery = async (context: AuthContext, ingestion: BasicSto
...
-    let url = getValueFromPath(ingestion.pagination_with_sub_page_attribute_path, requestData);
+    let url = toUrlString(getValueFromPath(ingestion.pagination_with_sub_page_attribute_path, requestData));
     while (isNotEmptyField(url) && (maxResults === 0 || objects.length < maxResults)) {
+      verifyIngestionUri(url);
       logApp.info(`> Sub query: ${url}`);
       await wait(100); // Wait 100 ms between 2 calls
       const { data: paginationData } = await httpClient.call({
         method: ingestion.pagination_with_sub_page_query_verb ?? ingestion.verb,
-        url,
+        url: url as string,
```
fellowseb: nit: can now be removed I believed
```suggestion
        url,
```

fellowseb (approved):

opencti-platform/opencti-graphql/src/modules/ingestion/ingestion-json-domain.ts
mcartaud: 🤦‍♂️

PR #17070, opened 2026-07-09
fix(labels): respect user defined label case (#16680)

### LINKED ISSUE #16680: feat(UX):  Remove automatic text-transform: capitalize on labels, preserve user-defined casing
Summary
Remove the CSS property that automatically applies title case (or any text-transform capitalization) to labels, so that the casing defined by users at creation time is preserved and displayed as-is.

Problem
Users are consistently frustrated by the automatic capitalization applied to labels in the UI. When a label is created with a specific casing (e.g. malware, APT-Group, phishingKit), the display is silently overridden by a CSS rule — making it impossible to distinguish intentional casing choices.

This is particularly problematic for:

Technical labels that follow naming conventions (e.g. camelCase, snake_case, acronyms).
Taxonomy-aligned labels where casing carries semantic meaning.
Imported labels from external sources (MISP, STIX, etc.) that must remain consistent.
Expected Behavior
Labels should be rendered exactly as entered by the user. No CSS transformation should alter their appearance.

Proposed Fix
Remove or override the CSS property responsible for the automatic capitalization on label components.

mcartaud:
### Proposed changes

* Remove labels capitalization

#### Boyscoutting:
* filter out already selected labels when adding new ones
* display color of labels when adding new labels
Old behavior
Screenshot 2026-07-15 at 15 19 33
Screenshot 2026-07-15 at 15 19 40

New behavior
image

### Related issues

* closes #16680 

### Checklist


esrevi:
I think it misses one case, when I create a label, the color isn't applied directly in the input select

https://github.com/user-attachments/assets/bcd11e5a-dc32-4563-be0a-60ee5ecb3533

opencti-platform/opencti-front/src/private/components/common/stix_core_objects_or_stix_relationships/StixCoreObjectOrCoreRelationshipLabelsView.jsx
```diff
@@ -173,11 +176,14 @@ const StixCoreObjectOrCoreRelationshipLabelsView = (props) => {
                 key={label.id}
                 label={label.value}
                 color={label.color}
-                onDelete={canUpdateKnowledge ? () => (
-                  enableReferences
-                    ? handleOpenCommitDelete(label)
-                    : handleRemoveLabel(label.id)
-                ) : undefined
+                labelTextTransform="none"
+                onDelete={
```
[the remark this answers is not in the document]
esrevi: I think we could add in WidgetCustomAttributesCard, for objectLabel case this labelTextTransform="none" too so it matches the same behaviour

esrevi:
It seems the labels are not preserving the user case: 

https://github.com/user-attachments/assets/7895d07b-0630-4ece-afb4-58595140fc74

opencti-platform/opencti-front/src/components/common/tag/Tag.tsx
```diff
@@ -127,4 +127,8 @@ const Tag = ({
   );
 };
 
+type LabelTagProps = Omit<TagProps, 'labelTextTransform'>;
+
+export const LabelTag = (props: LabelTagProps) => <Tag {...props} labelTextTransform="none" />;
```
Archidoit: we usually create components that are used in other files than the one where they are defined in a separated file, LabelTag.tsx in this case (like it is done for ItemYears.tsx, even though the file is small and call almost directly <Tag>, as it is in your case)

Archidoit: Maybe we can rename this component to RawTag, as it can be used for other components than labels in the future.

Archidoit:
After creating a label with some capitals and a given color, the label case and color is not applied 

image


whereas:

image

Same, in Taxonomie > Labels, the created labels are all displayed with small letters:
image

mcartaud:
> It seems the labels are not preserving the user case:

Fixed it. Honestly don't understand how it could have worked before, but I saw some tags in uppercase in the platform for it is a mystery to me

mcartaud:
> After creating a label with some capitals and a given color, the label case and color is not applied
> 
> Same, in Taxonomie > Labels, the created labels are all displayed with small letters: 

Yes, I didn't test thoughtfully the label creation, I believed the issue was only on display because I had uppercase labels in my data, not sure how.
It should be fixed now

mcartaud:
@esrevi @Archidoit, it if fixed now, I missed the fact that main issue comes from the creation that lower case all new labels (at least the ones created from UI by user).
The caveat with new implementation is that "aLabel", "alabel" and "Alabel" are now 3 different labels. Is it ok? (I'll double check with product)

Archidoit:
Nice, perfect! @mcartaud Waiting for the product confirmation then  :)

Archidoit:
New created labels are in the correct color in the dropdown (red in the screenshot), but not in the input and in the report details (blue in the screenshot): 
image


image

Archidoit:
Label in minuscules are displayed with first letter in capital in entities list
- report details
image


- but report list
image

Archidoit:
In filters, the syntax is not the same between
- filter icon button (everything in small letters)
- input box (everything with capital first letter, than small letters)
- input values list (correctly displayed)

image

mcartaud:
> New created labels are in the correct color in the dropdown (red in the screenshot), but not in the input and in the report details (blue in the screenshot): 

~It is working well on my computer, can you show me?~

Discussed together, it is working when using hex color, but not plain string colors. After digging a bit, it is not working anywhere with this string color.
It feels like it is another issue, opened a bug [here](https://github.com/OpenCTI-Platform/opencti/issues/17238)

opencti-platform/opencti-front/src/components/common/tag/Tag.tsx
```diff
@@ -30,13 +30,27 @@ const Tag = ({
   const theme = useTheme();
   const defaultColor = theme.palette.severity?.default ?? '#004C66';
 
+  const isHexColor = (c: string) => /^#([0-9a-fA-F]{3}|[0-9a-fA-F]{6}|[0-9a-fA-F]{8})$/.test(c);
+
+  const toHex = (c: string): string => {
+    if (isHexColor(c)) return c;
+    try {
+      const ctx = document.createElement('canvas').getContext('2d');
```
esrevi: on every render loop, if the color is not a hex color, it will creates a canvas element, this is not good, because this component is called in datatable, where there can be for each row, many Tags to render. 

I would drop that from the component, and maybe, normalize it on backend on addLabel/edit label?

mcartaud:
> Nice, perfect! @mcartaud Waiting for the product confirmation then :)

Seen with product, we display labels as written by the user, but use lower cases version for unicity check (so not possible to have `Label` and `label` for instance

opencti-platform/opencti-front/src/components/common/tag/Tag.tsx
mcartaud: Yes, not fan of it either, but is only on the "add tag" dialog after a tag creation where we use a non hex color (ie "red"). After creation, the color is normalized as hex by backend so will be very seldom and I decided to use this rather than copy the normalizing code from backend.
But now that I am saying it, I fill I could probably get the normalized value from backend right after the creation rather that making this weird thing

esrevi:
just wondering on update drawer message error
image

and on create label
image

Should we set the same message error on update? to me it seems more understandable This label already exists rather than This update will produce a duplicate, but nothing happens at the end since it is not mutated in that case.

nice to have: do you think we could render in the error message the name of label duplicated so we know "coincoin" already exists because "coinCoin" exists?

mcartaud:
> Should we set the same message error on update? to me it seems more understandable This label already exists rather than This update will produce a duplicate, but nothing happens at the end since it is not mutated in that case.
> 
> nice to have: do you think we could render in the error message the name of label duplicated so we know "coincoin" already exists because "coinCoin" exists?

For the error message, it is a generic behavior for all entities, so I'd prefer not to change it without much thinking. For the other point, same response : generic behavior so changing it have large impacts. I could do frontend validation (which would allow me to display the duplicate name), but we are on a paginated view so, we would need a dedicated route to check unicity before update. Do you think it is worth it?

alice-debra:
Product validation done 🟢

opencti-platform/opencti-front/lang/front/fr.json
```diff
@@ -4667,6 +4667,7 @@
   "This is an open vocabulary that covers common capabilities that may be exhibited by a malware instance or family": "Il s'agit d'un vocabulaire ouvert qui couvre les capacités communes que peuvent présenter une instance ou une famille de logiciels malveillants",
   "This is your own account. If your IP is not in the allow list, you will be locked out of the platform after saving.": "Il s'agit de votre propre compte. Si votre adresse IP ne figure pas dans la liste d'autorisation, vous serez bloqué hors de la plateforme après l'enregistrement.",
   "This item is built-in": "Cet attribut est par défaut",
+  "This label already exists": "Cette étiquette existe déjà",
```
```suggestion
  "This label already exists": "Ce label existe déjà",
```

Archidoit: In French, label is generally translated to 'label' in opencti

Archidoit (approved):

PR #17076, opened 2026-07-09
fix(catalog-domain): Memory leak & CPU spikes in catalog/connector contract validation (AJV) leading to "JavaScript heap out of memory" crashes (#17075)

### LINKED ISSUE #17075: fix(catalog-domain): Memory leak & CPU spikes in catalog/connector contract validation (AJV) leading to "JavaScript heap out of memory" crashes
## Description

The connector catalog module (`catalog-domain.ts`) suffers from two compounding issues that cause uncontrolled memory growth on the OpenCTI front pod, eventually crashing the Node.js process with a V8 JavaScript heap out of memory error. This was identified via Pyroscope memory profiling (memory:inuse_space:bytes) on a production pod, where inuse space grew unbounded over ~15 minutes before the pod was killed and restarted.

The actual "check-then-set" pattern only guards against sequential calls. Under concurrent load (e.g. a GraphQL query resolving manager_contract_hash / manager_contract_image for N connectors in parallel), every concurrent call sees catalogMap as undefined before the first await `buildCatalogMap()` resolves, triggering N parallel, redundant rebuilds of the entire catalog — including re-reading files, re-parsing JSON, and recompiling every AJV schema in the catalog. `getSupportedContractsByImage()` has the same flaw.

## Environment

2. Version: { e.g. 1.2.3 }
3. Other environment details:

## Reproducible steps

1. { e.g. Run ... }
2. { e.g. Click ... }
3. { e.g. Error ... }

## Additional information

[ajv-validator/ajv#2554 — SchemaEnv cached twice when using addSchema](https://github.com/ajv-validator/ajv/issues/2554) — confirms AJV's internal cache stores SchemaEnv keyed by the full schema object, which is "particularly memory intensive" for large/frequently-regenerated schemas.
[ajv-validator/ajv#2555 — PR: "Use cache once" (noCache option)](https://github.com/ajv-validator/ajv/pull/2555) — proposed (unmerged) fix for #2554.
[ajv-validator/ajv#2135 — PR: "fix: use WeakMap to cache compiled schemas"](https://github.com/ajv-validator/ajv/pull/2135) — describes "memory leakage of compile function cache" and proposes switching the internal cache to a WeakMap so unreferenced schemas can be garbage-collected. Open since 2022, still unmerged — confirms this is a known, unresolved upstream limitation, not an isolated misuse on our side.

SarahBocognano:
### Proposed changes

* Memoize the in-flight Promise, not just its resolved value, in `getCatalogs()` and `getSupportedContractsByImage()`, so concurrent callers await the same build instead of triggering parallel rebuilds. Reset on failure to allow retry.
* Replace ad-hoc `ajv.compile()` calls with a keyed validator cache `(Map<cacheKey, ValidateFunction>)`, keyed by a stable identifier that accurately reflects the schema shape (e.g. contract slug + sorted required/optional field names). This bounds compilation to one per distinct schema shape actually encountered, instead of once per request.
* Clear the validator cache in `resetCatalogs()` to stay consistent when custom catalogs are reloaded.

### Related issues

* closes: https://github.com/OpenCTI-Platform/opencti/issues/17075

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/modules/catalog/catalog-domain.ts
+++ b/opencti-platform/opencti-graphql/src/modules/catalog/catalog-domain.ts
@@ -13,7 +13,6 @@ import { readFile } from 'node:fs/promises';
 import type { ValidateFunction } from 'ajv';
 
 const validatorCache = new Map<string, ValidateFunction>();
-const MAX_VALIDATOR_CACHE_SIZE = 500;
 
 /**
  * Compiles (or retrieves from cache) an AJV validator for a given schema.
@@ -23,10 +22,6 @@ const MAX_VALIDATOR_CACHE_SIZE = 500;
 const getOrCompileValidator = (cacheKey: string, jsonValidation: object): ValidateFunction => {
   let validate = validatorCache.get(cacheKey);
   if (!validate) {
-    if (validatorCache.size >= MAX_VALIDATOR_CACHE_SIZE) {
-      logApp.warn('Validator cache size limit reached, compiling without caching', { size: validatorCache.size });
-      return ajv.compile(jsonValidation);
-    }
     validate = ajv.compile(jsonValidation);
     validatorCache.set(cacheKey, validate);
   }
--- a/opencti-platform/opencti-graphql/src/modules/catalog/catalog-domain.ts
+++ b/opencti-platform/opencti-graphql/src/modules/catalog/catalog-domain.ts
@@ -79,10 +79,7 @@ const buildCatalogMap = async (): Promise<Record<string, CatalogType>> => {
               additionalProperties: contract.config_schema.additionalProperties,
             };
             try {
-              // Stable key: the contract is uniquely identified by its title/slug,
-              // and its shape doesn't change from one call to another
-              getOrCompileValidator(`catalog-contract:${contract.title}`, jsonValidation);
-            } catch (err) {
+              getOrCompileValidator(`catalog-contract:${catalog.id}:${contract.slug}`, jsonValidation);
               throw UnsupportedError('Contract must be a valid json schema definition', { cause: err });
             }
           }
--- a/opencti-platform/opencti-graphql/src/modules/catalog/catalog-domain.ts
+++ b/opencti-platform/opencti-graphql/src/modules/catalog/catalog-domain.ts
@@ -10,15 +10,37 @@ import filigranCatalog from '../../__generated__/opencti-manifest.json';
 import conf, { logApp } from '../../config/conf';
 import type { ConnectorContractConfiguration, ContractConfigInput } from '../../generated/graphql';
 import { readFile } from 'node:fs/promises';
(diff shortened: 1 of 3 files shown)


opencti-platform/opencti-graphql/src/modules/catalog/catalog-domain.ts
```diff
@@ -342,7 +372,15 @@ export const validateContractConfigurations = (
     additionalProperties: false,
   };
 
-  const validate = ajv.compile(jsonValidation);
+  // The key must exactly capture the shape of validationProperties, since it varies
+  // depending on which optional fields are actually present in the request.
+  const cacheKey = [
```
aHenryJard, line 372: We could maybe use a hash as key ?

opencti-platform/opencti-graphql/src/modules/catalog/catalog-domain.ts
```diff
@@ -10,15 +10,37 @@ import filigranCatalog from '../../__generated__/opencti-manifest.json';
 import conf, { logApp } from '../../config/conf';
 import type { ConnectorContractConfiguration, ContractConfigInput } from '../../generated/graphql';
 import { readFile } from 'node:fs/promises';
+import type { ValidateFunction } from 'ajv';
+
+const validatorCache = new Map<string, ValidateFunction>();
+const MAX_VALIDATOR_CACHE_SIZE = 500;
```
aHenryJard: Does the cache have one entry per connector ? I mean this number will be an issue when there is more than 500 connector in the catalog right ?

fellowseb: I would be in favour of removing this limit of 500 that's totally arbitrary. Especially as we are going to work on the new system next week...

esrevi: I agree that setting a max cache size is not necessary especially if we exceed that value, the cache becomes useless, we could remove it imo

fellowseb (approved):
Reproduced locally 👍

opencti-platform/opencti-graphql/src/modules/catalog/catalog-domain.ts
esrevi: We don't think hashing is worth it, slug and sorted field names come from connector contracts, so there should not be real collisions and hashing would add CPU overhead and make cache entries unreadable in logs.

fellowseb:
Memory profiling shows the fix works: we used `https://github.com/mcollina/autocannon` to run 100 concurrent requests requiring a call to `buildCatalogMap`. The catalog currently contains 277 contracts.
- Before the fix, after only 10 calls to `buildCatalogMap` we see +3000 catalog contracts retained in memory
- After the fix we only see each contract twice retained in memory. So maybe still room for improvement but as we are going to rework this feature soon it's a good first step.

PR #17082, opened 2026-07-09
feat(authentication): logout based on websocket connection (#17081)

### LINKED ISSUE #17081: feat(authentication): log users out after closing all browser tabs or windows
## Use case

Currently users are only logged out when they timeout or click the logout button, closing all tabs/windows keeps the user logged in until the session expires. This fails to record logouts and results in a gap in access restriction where users are not prompted to login if the site is reloaded before the original session expires. 

## Current workaround

Users must explicitly logout 

## Proposed solution

Utilize the websocket connection used by graphql to track connection and disconnection events for users. Upon disconnection the user should have their sessions killed, and logout activity logged.  

## Additional information

OpenCTI allows multiple sessions which complicates this process. Rather than removing individual sessions based on what is closed,  the system could clear all sessions once the user's last instance is closed. Though this will potentially still allow for users to not be prompted for a login upon reloading the page, it still relies on the user being active in one of their sessions. 

Yes

scarletmerlin123:
### Proposed changes

* Add on connection listener to record when a user is connected to the application. Log each connection and track how many connections each connected user has
* Add on disconnection listener to track when a user disconnects from the websocket connection. 
* When a disconnection occurs for a user, wait 4sec before continuing to ensure it was an intentional disconnect. 
* Remove the the connection from the users connection count. When all connections are removed for a user, trigger a log out function to record the logout activity, clear the redis user context, and kill all active sessions.

### Related issues

* closes #17081 

### How to test this PR

Test that user is logged out upon tab closure
- login as a user then close the browser tab. 
- The login screen should be presented when navigating back to the application

Test multiple tab/window handling
- login and open multiple tabs
- The user should be able to close tabs and not be logged out until the last tab is closed

Test multiple session handling
- Login using an incognito window and a regular browser window to trigger a single user having multiple sessions
- Close one of the windows
- The user should be able to remain active for the remaining window
- Close the final window and the login screen should be presented when navigating back to the application

Test activity logging
- Review the activity page for logout activity for the previous test cases and confirm that logs were recorded for the test cases

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/http/httpServer.js
+++ b/opencti-platform/opencti-graphql/src/http/httpServer.js
@@ -24,7 +24,7 @@ import { createAuthenticatedContext } from './httpAuthenticatedContext';
 import { getSettings } from '../domain/settings';
 import { isWorkAlive } from '../domain/work';
 import { computeLoaders } from './httpAuthenticatedContext';
-import { buildRateLimiterOptions } from './httpUtils';
+import { buildRateLimiterOptions, logOutUser } from './httpUtils';
 import { checkDraftInContext } from './httpServer-draft';
 import ipWhitelistMiddleware from './ipWhitelistMiddleware';
 
@@ -118,11 +118,44 @@ const createHttpServer = async () => {
     server: httpServer,
     path: `${basePath}/graphql`,
   });
+  const userConnections = new Map();
   wsServer.on('error', (e) => {
     throw e;
   });
+  const trackConnections = async (user, activity, timeoutId) => {
+    const userId = user.id;
+    if (activity === 'connection') {
+      if (!userConnections.has(userId)) {
+        userConnections.set(userId, 1);
+      } else {
+        userConnections.set(userId, userConnections.get(userId) + 1);
+      }
+    } else {
+      const userConnectionCount = userConnections.get(userId);
+      if (userConnectionCount === 1) {
+        await logOutUser(user);
+        userConnections.delete(userId);
+      } else if (userConnectionCount > 1) {
+        userConnections.set(userId, userConnections.get(userId) - 1);
+      }
+    }
+    if (timeoutId) {
+      clearTimeout(timeoutId);
+    }
+  };
   const serverCleanup = useServer({
     schema,
+    onConnect: async (ctx) => {
+      const session = await extractWsSessionContext(ctx);
+      await trackConnections(session.user, 'connection');
+    },
+    onDisconnect: async (ctx) => {
+      // timeout to prevent logouts due to refresh or network issues
+      const timeoutId = setTimeout(async () => {
+        const session = await extractWsSessionContext(ctx);
+        await trackConnections(session.user, 'disconnection', timeoutId);
+      }, 4000);
+    },
     context: extractWsSessionContext,
   }, wsServer);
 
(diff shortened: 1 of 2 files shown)


fellowseb:
Hi @scarletmerlin123 ,
Thanks for contributing !

This POC demonstrates that we can observe GraphQL WebSocket connections, but we should not use this implementation as the production solution IMO.
The main problem is that presence is tracked by user ID and logout calls `killUserSessions(user.id)`. This changes the existing logout semantics: closing one browser session could invalidate the user’snormal, incognito, other-browser, and other-computer sessions. Conversely, if another session still has a WebSocket connection, closing the last tab of the current session only decrements the user-wide counter and may not invalidate the session that was actually closed.

There are several other reliability issues: 
•  `userConnections` is local to one backend process. Connections on other nodes are invisible, so this does not work correctly in a clustered deployment or during rolling restarts.
• A disconnect is not proof that a tab was closed. Refreshes, laptop sleep, network changes, proxies, and backend restarts also disconnect WebSockets. 
• The four-second timer is hardcoded, is not cancelled by a reconnect—the timeout is cleared only after it has fired—and is probably too short for transient failures.
• There are no expiring presence leases, so a browser or backend crash may produce no usable disconnect signal and leave the session active until its normal timeout. 
• The final decision is not an atomic “does this exact session still have active tabs?” check, leaving room for connection/disconnection races. 
• The delayed callback calls  `extractWsSessionContext`  again. If the session has already expired or been explicitly destroyed, that can throw from the detached timer.
•  `delUserContext(user)`  is also user-scoped and could remove edit contexts belonging to another active browser session.
• The audit event does not distinguish an automatic presence-based logout from an explicit logout. 

Instead, we intend to implement session-scoped server-side presence:                                                                                                                                         

1. Associate each connection with the current server session ID, never only the user ID. The session ID remains server-side and is derived from the authenticated cookie. 
2. Register one presence per authenticated tab, with a per-tab ID generated in  `sessionStorage` . (To be assessed in really needed)
3. Store expiring presence leases in Redis so all backend nodes share the same view and crashes are eventually detected.
4. After a disconnect or lease expiry, wait for a configurable grace period and atomically verify that the exact session has no remaining tab presence. 
5. Destroy only that server session, using a dedicated session-scoped logout service. Do not call  `killUserSessions` .
6. Optionally send a best-effort  `pagehide`  departure signal from the frontend to accelerate clean tab closures, while keeping lease expiry as the authoritative fallback. 
7. Keep explicit logout and remote SSO logout behavior unchanged, and record a distinct automatic logout reason such as  presence_expired .

This preserves the current isolation between normal, incognito, other-browser, and other-computer sessions while handling refreshes, transient disconnections, crashes, and multi-node deployments more safely.

Also we want to make this behavior an opt-in accessible by admins in the Settings > Security panel. Turning it on would activate it for all users on the platform.

scarletmerlin123:
Hi @fellowseb,

Thank you for reviewing! I understand the concerns of implementing this in a production setting.

Are your proposed changes something that Filigran is actively working on? I'm also not sure if we would still end up with an incomplete audit trail because of reliability issues of using client side events for triggering the logs. 

Maybe as a low effort approach I could tie logging to where sessions get cleaned up. When a user interacts with the system it will purge all the expired sessions. I could update this function to report back the updated sessions and use them to write the logs for removed sessions. It won't give us the time resolution that we wanted, but would serve the purpose of providing a better audit trail to know roughly when users exited the system with minimal complexity. Any thoughts on doing this? 

https://github.com/OpenCTI-Platform/opencti/blob/084db588a64be1b49f838dfb6b89a80487558ce4/opencti-platform/opencti-graphql/src/database/redis.ts#L212-L224

PR #17091, opened 2026-07-10
feat(authentication): improvement when updating the validity policy (#9606)

### LINKED ISSUE #9606: feat(authentication): force a password change by an admin

CelineSebe:
### Proposed changes

* Replace with elRawUpdateByQuery if we have a large number of users

### Related issues

* resolves #9606 

### How to test this PR

1- create 500 users with a script
2- update the password validity duration
3- no latency after clicking on "update" button

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/domain/user.js
+++ b/opencti-platform/opencti-graphql/src/domain/user.js
@@ -747,9 +747,13 @@ export const adjustAllUsersPasswordValidUntil = async (context, oldDays, newDays
         script: {
           source: `
             if (ctx._source.password_valid_until != null) {
-              def current = ZonedDateTime.parse(ctx._source.password_valid_until);
-              def newDate = current.plusNanos(params.diffMs * 1000000L);
-              ctx._source.password_valid_until = newDate.format(DateTimeFormatter.ISO_OFFSET_DATE_TIME);
+              try {
+                def current = ZonedDateTime.parse(ctx._source.password_valid_until);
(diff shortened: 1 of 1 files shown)


ludovic:
Nice work on the performance side — switching to `elRawUpdateByQuery` for bulk updates makes a lot of sense.

One point about multi-node (cluster) deployments, which we officially support: since we now write directly to Elasticsearch instead of going through `updateAttribute`, we no longer emit the event that refreshes the user cache across all nodes. The `resetCacheForEntity(ENTITY_TYPE_USER)` added here only clears the cache on the local node.

In a clustered setup, this means users served by the other nodes keep a stale `password_valid_until` in their in-memory cache, so they won't be redirected to `/change-password` until a restart. Users on the node that handled the update work fine.

To be clear, this is not a regression from this PR — the old per-user loop already bypassed the cluster refresh. But since the feature is new, it feels like the right time to close the gap. This could be a good candidate for a follow-up PR rather than blocking this one.

ludovic (approved):

PR #17097, opened 2026-07-10
chore(docker): check that all image build on PR (#17096)

### LINKED ISSUE #17096: Add a build of all docker images on PR

aHenryJard:
### Proposed changes

* Add a build of all docker image on PR from local branch (issue) or renovate
* Do not store any artifact on workflow itself
* Reduce artifact store on action: the other workflow is doing a global artifact and has a 7 day retention rule (see https://github.com/OpenCTI-Platform/opencti/blob/77a6a575e30c5df30aa33ae7da9885b4effab3fd/.github/workflows/cd-build-rolling-images-on-merge.yml#L130l )

### Related issues

* closes #17096 

### How to test this PR

see https://github.com/OpenCTI-Platform/opencti/actions/runs/29811803972
from https://github.com/OpenCTI-Platform/opencti/pull/17232
### Checklist

### DIFF
--- a/.github/actions/docker-build-platform/action.yml
+++ b/.github/actions/docker-build-platform/action.yml
@@ -56,4 +56,4 @@ runs:
       with:
         name: docker-image-opencti-platform${{ inputs.docker_image_suffix }}
         path: /tmp/opencti-platform${{ inputs.docker_image_suffix }}.tar
-        retention-days: 7
+        retention-days: 1
--- a/.github/workflows/cd-check-images-on-pr.yml
+++ b/.github/workflows/cd-check-images-on-pr.yml
@@ -6,6 +6,8 @@ on:
     branches:
       - master
       - lts/*
+      # FIXME tmp
+      - issue/17096-check-images-on-pr
 
 jobs:
   # -----------------------
--- a/.github/workflows/cd-check-images-on-pr.yml
+++ b/.github/workflows/cd-check-images-on-pr.yml
@@ -27,6 +27,9 @@ jobs:
           - variant: fips
             suffix: "_fips"
     steps:
+      - uses: actions/checkout@v7
+        with:
+          ref: ${{ github.sha }}
       - uses: ./.github/actions/docker-build-platform
         with:
           checkout_ref: ${{ github.sha }}
@@ -50,6 +53,9 @@ jobs:
           - variant: fips
             suffix: "_fips"
     steps:
+      - uses: actions/checkout@v7
+        with:
+          ref: ${{ github.sha }}
       - uses: ./.github/actions/docker-build-worker
         with:
           checkout_ref: ${{ github.sha }}
--- a/.github/workflows/cd-check-images-on-pr.yml
+++ b/.github/workflows/cd-check-images-on-pr.yml
@@ -1,0 +1,58 @@
+name: OpenCTI CD Check
+# This workflow build docker images variant to validate before merge on master.
+# requires branches that follow convention (issue/*)
+
+on:
+  push:
+    branches:
+      - issue/*
+      - renovate/*
+
+jobs:
+  # -----------------------
+  # Build platform images
+  # -----------------------
+  build-platform:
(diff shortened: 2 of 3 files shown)


.github/actions/docker-build-platform/action.yml
```diff
@@ -56,4 +56,4 @@ runs:
       with:
         name: docker-image-opencti-platform${{ inputs.docker_image_suffix }}
         path: /tmp/opencti-platform${{ inputs.docker_image_suffix }}.tar
-        retention-days: 7
\ No newline at end of file
+        retention-days: 1
```
nvkhope, line 59: Question: all other workflows are going to have the retention days to 1?

.github/workflows/cd-check-images-on-pr.yml
```diff
@@ -0,0 +1,56 @@
...
+      matrix:
+        include:
+          - variant: standard
+            suffix: ""
+          - variant: ubi9
+            suffix: "_ubi9"
+          - variant: fips
+            suffix: "_fips"
+    steps:
+      - uses: ./.github/actions/docker-build-platform
```
nvkhope, line 31: I think we need to checkout before using the actions (otherwise you wont be able to find the action)

.github/workflows/cd-check-images-on-pr.yml
```diff
@@ -0,0 +1,56 @@
+name: OpenCTI CD Check
+# This workflow builds docker image variants to validate before merge on master or lts.
+
+on:
+  pull_request:
+    branches-ignore:
```
nvkhope:
```suggestion
    branches:
```
If we only want to run this workflow when targeting master... etc


https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#example-including-branches

aHenryJard: ah yes it was on event "push" before, I need to change that to the opposite, thanks !

aHenryJard: and so I need to remove gh-pages

.github/actions/docker-build-platform/action.yml
aHenryJard: yes except for the rolling consolidated artifact "opencti-rolling" that will be kept 7 days

nvkhope: Should we define the retention-days as param in each workflow then? 🤔

.github/workflows/cd-check-images-on-pr.yml
nvkhope: I agree 😄

.github/actions/docker-build-platform/action.yml
aHenryJard: Why ? This is a kind a temporary build, and it's up to the workflow using the action to keep it longer in an aggregated one if needed, it seems fine that way to me: the workflow has the ownership on retention for its usage.

For me action = can be used in several workflow
workflow = orchestrator of a whole process

We are in an action here.

.github/workflows/cd-check-images-on-pr.yml
aHenryJard: done.

.github/workflows/cd-check-images-on-pr.yml
aHenryJard: you are totally right

aHenryJard: done.

.github/actions/docker-build-platform/action.yml
nvkhope: Oh sorry, I got it now 😅 

We want a 1 day retention for the action artifacts, however, as you said, thanks to the upload-artifacts job in the cd-build-rolling-images-on-merge workflow for example, we consolidate them with the 7 day retention

nvkhope (approved):

PR #17172, opened 2026-07-16
fix(dashboard): Take interval into account in widgets (#17168)

### LINKED ISSUE #16857: fix(dashboard): Line graph visulaizations display improper X Axis labels
## Description
When using the Line visualization in a dashboard widget, regardless of the  interval setting, the X axis displays like days.   

Image

## Environment

1. 7.260624.0

## Reproducible steps

1. Create a dashboard widget. Configuration: Visualization - Line; Filters - in testing, Entity type = incidents AND Status = Analyzed; Parameters - Interval - Month was used in testing; Date attribute - created_at (Technical date) and created (Functional date) were used
2. Observe the X axis displays the interval as one would expect for a Day interval

## Expected output

X axis displays the appropriate interval. For a month interval, data is sum’ed and grouped on a monthly basis. For a week interval, data is sum'ed and grouped on a weekly basis. 

## Actual output
Data is displayed in daily format, no sum or grouping performed.

### LINKED ISSUE #17032: fix(dashboard): bar-chart date-histogram buckets duplicated when secondary and tertiary group-by (Label) added
# Description

In a custom dashboard, a vertical bar-chart widget on the 'Incident' entity type that groups by a date field ('first_seen') using a 'date-histogram 'Interval = Month (or Year)' AND adds a secondary/tertiary breakdown/group-by on 'Label'renders the same time bucket multiple times on the x-axis.

Observed example: December 2026 appears twice, January 2026 three times, May 2026 four times. The repeat count is 'variable' (2/3/4) rather than a flat 3x, despite only 3 configured label values. Which also asks if the sum up of the Incidents are correctly calculated.

Seems that aggregate totals are correct. Only the bucketing/distribution shape on the x-axis is wrong: date buckets are duplicated rather than each date bucket containing its label breakdown.

Looks like a regression. The same widget rendered correctly approximately one week prior to 2026-07-07, with no configuration changes on the reporting side.

## Environment

1. OS: OpenCTI Cloud/SaaS (hosted)
2. Version: 7.260701.0 

## Reproducible steps

1. Create a custom dashboard and add a vertical bar-chart widget.
2. Set the entity type to Incident.
3. Set the primary group-by to a date field (`first_seen`) with a date-histogram Interval of Month (or Year).
4. Add a secondary breakdown / group-by on Label (with Label being multi-valued across the dataset).
5. Observe the x-axis: individual month (or year) buckets are repeated multiple times, with a variable repeat count.

## Expected output

Each date-histogram bucket (e.g. each month) appears **once** on the x-axis, with the Label breakdown nested/stacked within that single bucket.

## Actual output

Individual date buckets are duplicated on the x-axis. The repeat count per bucket varies (2, 3, 4...) and appears to track the number of distinct Label values present in that time bucket. Aggregate totals remain correct.

## Additional information

Image

For customer screenshots, please ask internally. Thank you

### LINKED ISSUE #17168: fix(dashboards): Vertical-bar widget (relationships perspective) ignores "Month" interval and always buckets by day, regardless of dashboard time-range mode
## Description
A vertical-bar widget configured with Interval = Month does not aggregate data into monthly buckets. Instead, it renders one bar per day for every day that has data, across the entire selected range. This happens regardless of:

Whether the dashboard time-range mode is "Relative time" or a fixed range
Whether explicit start/end dates are set
The width of the selected range (reproduces even when the range is narrowed to a single calendar month 

The x-axis label formatter does appear to apply the monthly format (e.g., "June 2026"), but it's applied per daily bucket rather than per monthly bucket - resulting in the same month label repeated under multiple individual daily bars instead of one consolidated bar per month.

## Environment
1. 7.260710.0

## Reproducible steps

1. Create a dashboard, add a vertical-bar widget with:
    Perspective: relationships
    Relationship type filter: e.g. related-to
    Date attribute: start_time
    Interval: Month
    (This particular repro also uses a dynamicTo/regardingOf nested filter targeting specific entities via a part-of relationship - see attached widget config JSON; not yet confirmed whether the nested filter is required to trigger the bug or incidental.)

See : https://csm.octi.staging.filigran.io/dashboard/workspaces/dashboards/9b33a446-b001-4fdc-822f-bd3d90ae2964

## Expected output
With Interval = Month, data should be aggregated into one bar per calendar month.
This should render as exactly one bar for that month (or zero, if there's no data), not multiple daily bars all labelled with the same month.

## Actual output
One bar is rendered per day with data, for the entire range - never consolidated into monthly buckets.
Confirmed even in a 1-month window: ~10 distinct daily bars appear, all labelled "June 2026," instead of a single monthly bar.
Changing the dashboard time-range mode or dates has no effect on this behavior.

## Additional information

Image

Image

Image

Archidoit:
### Proposed changes
- Take into account the interval selected by the user in the 'parameters' tab instead of always displaying the widget with the 'day' interval for the widgets : vertical bar (entities and relationship perspectives), heat map (entities perspective), line (entities and relationship perspectives), area (entities perspective).

- Add frontend tests for Entities and Relationships vertical bar widgets

image

### Related issues
fixes #17168
fixes #17032
fixes #16857

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectsMultiAreaChart.tsx
+++ b/opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectsMultiAreaChart.tsx
@@ -13,6 +13,7 @@ import { Widget, WidgetDataSelection, WidgetHost, WidgetParameters } from '../..
 import { DashboardConfig } from '../../../../components/dashboard/dashboard-types';
 import { StixCoreObjectsMultiAreaChartTimeSeriesQuery } from '@components/common/stix_core_objects/__generated__/StixCoreObjectsMultiAreaChartTimeSeriesQuery.graphql';
 import { computeStartEndDates } from '../../../../components/dashboard/dashboard-viz-utils';
+import { getWidgetInterval } from 'src/utils/widget/widgetUtils';
 
 const stixCoreObjectsMultiAreaChartTimeSeriesQuery = graphql`
   query StixCoreObjectsMultiAreaChartTimeSeriesQuery(
@@ -122,7 +123,7 @@ const buildQueryVariables = (
   return {
     startDate,
     endDate,
-    interval: parameters?.interval ?? 'day',
+    interval: getWidgetInterval(parameters),
     timeSeriesParameters,
   };
 };
--- a/opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectsMultiAreaChart.tsx
+++ b/opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectsMultiAreaChart.tsx
@@ -9,7 +9,7 @@ import WidgetMultiAreas from '../../../../components/dashboard/WidgetMultiAreas'
 import Loader, { LoaderVariant } from '../../../../components/Loader';
 import useDashboardViz from '../../../../components/dashboard/useDashboardViz';
 import WidgetNoHostEntity from '../../../../components/dashboard/WidgetNoHostEntity';
-import { Widget, WidgetDataSelection, WidgetHost } from '../../../../utils/widget/widget';
+import { Widget, WidgetDataSelection, WidgetHost, WidgetParameters } from '../../../../utils/widget/widget';
 import { DashboardConfig } from '../../../../components/dashboard/dashboard-types';
 import { StixCoreObjectsMultiAreaChartTimeSeriesQuery } from '@components/common/stix_core_objects/__generated__/StixCoreObjectsMultiAreaChartTimeSeriesQuery.graphql';
 import { computeStartEndDates } from '../../../../components/dashboard/dashboard-viz-utils';
@@ -99,6 +99,7 @@ const DATA_SELECTION_TYPES = ['Stix-Core-Object'];
 const buildQueryVariables = (
   resolvedDataSelection: WidgetDataSelection[],
   config: DashboardConfig,
+  parameters?: WidgetParameters,
 ): StixCoreObjectsMultiAreaChartTimeSeriesQuery['variables'] => {
   const computed = computeStartEndDates(config);
   const startDate = computed.startDate ?? monthsAgo(12);
@@ -121,7 +122,7 @@ const buildQueryVariables = (
   return {
     startDate,
     endDate,
-    interval: 'day',
+    interval: parameters?.interval ?? 'day',
     timeSeriesParameters,
   };
 };
@@ -144,6 +145,7 @@ const StixCoreObjectsMultiAreaChart = ({
     refreshRate,
     query: stixCoreObjectsMultiAreaChartTimeSeriesQuery,
     config,
+    parameters,
     buildQueryVariables,
   });
 
--- a/opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectsMultiHeatMap.tsx
+++ b/opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectsMultiHeatMap.tsx
@@ -101,6 +101,7 @@ const DATA_SELECTION_TYPES = ['Stix-Core-Object'];
 const buildQueryVariables = (
   resolvedDataSelection: WidgetDataSelection[],
   config: DashboardConfig,
+  parameters?: WidgetParameters,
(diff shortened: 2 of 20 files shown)


opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectsMultiVerticalBars.tsx
```diff
@@ -120,7 +121,7 @@ const buildQueryVariables = (
   return {
     startDate,
     endDate,
-    interval: 'day',
+    interval: parameters?.interval ?? 'day',
     timeSeriesParameters,
   };
```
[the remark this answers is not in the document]
Archidoit: done

opencti-platform/opencti-front/src/private/components/common/stix_relationships/StixRelationshipsMultiLineChart.tsx
```diff
@@ -113,7 +114,7 @@ const buildQueryVariables = (
     operation: 'count',
     startDate,
     endDate,
-    interval: 'day',
+    interval: parameters?.interval ?? 'day',
```
esrevi: suggestion:

actually, there are some components using this interval that haven't covered this case. 

I would suggest to avoid having to add this test in all widgets that we extract this function to something like 
```
export const getWidgetInterval = (params?: WidgetParameters) => params?.interval ?? 'day' 
```
And use that function in all widgets having that pattern

Thus, I think one test on that function should cover all widgets

Archidoit: done !

opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectsMultiAreaChart.tsx
```diff
@@ -121,7 +122,7 @@ const buildQueryVariables = (
   return {
     startDate,
     endDate,
-    interval: 'day',
```
esrevi, line 124: should use getWidgetInterval :)

opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectsMultiHeatMap.tsx
```diff
@@ -119,7 +120,7 @@ const buildQueryVariables = (
   return {
     startDate: startDate ?? monthsAgo(12),
     endDate: endDate ?? now(),
-    interval: 'day',
```
esrevi, line 122: should use getWidgetInterval :)

opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectsMultiLineChart.tsx
```diff
@@ -118,7 +119,7 @@ const buildQueryVariables = (
   return {
     startDate,
     endDate,
-    interval: 'day',
```
esrevi, line 121: should use getWidgetInterval :)

opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectsMultiVerticalBars.tsx
```diff
@@ -120,7 +121,7 @@ const buildQueryVariables = (
   return {
     startDate,
     endDate,
-    interval: 'day',
```
esrevi, line 123: should use getWidgetInterval :)

opencti-platform/opencti-front/src/private/components/common/stix_relationships/StixRelationshipsMultiLineChart.tsx
esrevi: should use getWidgetInterval :)

opencti-platform/opencti-front/src/private/components/common/stix_relationships/StixRelationshipsMultiVerticalBars.tsx
```diff
@@ -112,7 +113,7 @@ const buildQueryVariables = (
     operation: 'count',
     startDate,
     endDate,
-    interval: 'day',
+    interval: parameters?.interval ?? 'day',
```
esrevi: should use getWidgetInterval :)

esrevi (approved):
tested, it works, gj ! :)

PR #17197, opened 2026-07-17
fix(dashboard): Fix number widget undefined title (#16456)

### LINKED ISSUE #16456: fix: number widget becomes inaccessible when title is cleared and saved
## Description

When editing a Number widget, removing the Title value and saving causes the dashboard to enter an error state.
The widget disappears from the dashboard and becomes inaccessible. Refreshing the page does not resolve the issue.

## Environment

1. OS: N/A (Cloud environment)
2. Version: 7.260529.0
3. Other environment details:
  a. Reproduced on a dashboard containing a number widget
  b. Issue occurs after editing an existing widget and saving an empty Title value

## Reproducible steps

1. Create a dashboard.
2. Create a Number widget.
3. Open the widget and select Update.
4. Navigate to Parameters.
5. Remove the Title value completely (empty string).
6. Save the widget.
7. Return to the dashboard.

## Expected output

One of the following should occur:
	• Empty titles should be validated and rejected before saving, or
	• The widget should be saved and rendered successfully without a title.

## Actual output

• The dashboard displays: “An unknown error occurred. Please provide a support package…”
• Refreshing the dashboard does not resolve the issue.
• The Number widget is no longer displayed on the dashboard.
• The widget becomes inaccessible for further editing or removal.

Archidoit:
### Proposed changes
When a user remove the title of a widget of type Number, we should not have an error, and a default title (depending on the perspective) should be displayed instead

Add tests for useEntityTranslation

image

### Related issues
closes #16456

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/common/audits/AuditsNumber.tsx
+++ b/opencti-platform/opencti-front/src/private/components/common/audits/AuditsNumber.tsx
@@ -20,10 +20,9 @@ import { useFormatter } from '../../../../components/i18n';
 import { dayAgo } from '../../../../utils/Time';
 import WidgetContainer from '../../../../components/dashboard/WidgetContainer';
 import WidgetNoData from '../../../../components/dashboard/WidgetNoData';
-import useEntityTranslation from '../../../../utils/hooks/useEntityTranslation';
 import WidgetNumber from '../../../../components/dashboard/WidgetNumber';
 import useDashboardViz from '../../../../components/dashboard/useDashboardViz';
-import { UNIQUE_COUNT_ESTIMATION_THRESHOLD, UNIQUE_COUNT_ESTIMATION_WARNING } from '../../../../utils/widget/widgetUtils';
+import { UNIQUE_COUNT_ESTIMATION_THRESHOLD, UNIQUE_COUNT_ESTIMATION_WARNING, useGetNumberWidgetTitle } from '../../../../utils/widget/widgetUtils';
 import type { WidgetDataSelection, WidgetHost, WidgetParameters } from '../../../../utils/widget/widget';
 import type { DashboardConfig } from '../../../../components/dashboard/dashboard-types';
 import { buildFiltersAndOptionsForWidgets, normalizeFilterGroupForBackend } from '../../../../utils/filters/filtersUtils';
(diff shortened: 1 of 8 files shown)


opencti-platform/opencti-front/src/utils/hooks/useEntityTranslation.ts
```diff
@@ -3,7 +3,10 @@ import { useFormatter } from '../../components/i18n';
 const useEntityTranslation = () => {
   const { t_i18n } = useFormatter();
 
-  const translateEntityType = (type: string) => {
+  const translateEntityType = (type: string | null) => {
+    if (type === '') {
+      return type;
+    }
```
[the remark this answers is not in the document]
Archidoit: done

opencti-platform/opencti-front/src/private/components/common/audits/AuditsNumber.tsx
```diff
@@ -164,9 +165,9 @@ const AuditsNumber: FunctionComponent<AuditsNumberProps> = ({
     parameters,
     buildQueryVariables,
   });
-
```
esrevi, line 161: I'd suggest to extract a function such as 
```
export const getWidgetNumberTitle = (
  title: string | null | undefined,
  defaultTitle: string,
): string => translateEntityType(isNotEmptyField(title) ? title : defaultTitle);
```
or getWidgetTitle if it seems more generic 

Should we also update DraftNumbers?

esrevi: then we can also add a test for that function too

opencti-platform/opencti-front/src/private/components/common/audits/AuditsNumber.tsx
```diff
@@ -32,6 +32,7 @@ import { UNIQUE_COUNT_ESTIMATION_THRESHOLD, UNIQUE_COUNT_ESTIMATION_WARNING } fr
 import type { WidgetDataSelection, WidgetHost, WidgetParameters } from '../../../../utils/widget/widget';
 import type { DashboardConfig } from '../../../../components/dashboard/dashboard-types';
 import { buildFiltersAndOptionsForWidgets, normalizeFilterGroupForBackend } from '../../../../utils/filters/filtersUtils';
+import { isNotEmptyField } from 'src/utils/utils';
```
esrevi: prefer relative import path *../../../../utils/utils*

opencti-platform/opencti-front/src/private/components/common/audits/AuditsNumber.tsx
Archidoit: I will create a hook since translateEntityType is a hook, and add tests :)

opencti-platform/opencti-front/src/private/components/common/audits/AuditsNumber.tsx
Archidoit: done :)

esrevi (approved):
lgtm !

PR #17211, opened 2026-07-19
perf(pycti): reduce request and queue amplification in high-volume workflows (#17212)

### LINKED ISSUE #17212: feat(client-python): improve high-volume ingestion and export performance
## Use case

High-volume `pycti` connectors and export workflows can spend a significant amount of time on avoidable per-object work: repeated GraphQL lookups and mutations, one-object AMQP fan-out for large same-level bundles, repeated export reads, and unnecessary payload buffering.

At sustained ingestion or export volume, these multipliers increase latency, queue traffic, backend request load, and memory usage even when the underlying business work is unchanged.

## Current workaround

The current workaround is to scale connectors, workers, and backend capacity or reduce bundle sizes manually. That helps absorb the load, but it does not remove the request and message amplification in the client paths themselves.

## Proposed solution

Improve the high-volume `pycti` paths while preserving existing behavior for current users:

* batch supported relation add/remove operations and prefetch repeated lookup data where the platform exposes a compatible bulk path
* reuse import/export scoped caches for repeated labels, vocabularies, external references, nested refs, and related export targets
* reuse AMQP publisher connections and send bounded same-dependency-level bundle chunks instead of one queue message per object
* keep listener concurrency opt-in and preserve old-platform fallbacks for new bulk relation capabilities
* stream large upload/download paths where transport permits and remove repeated hot-path allocations

The intended result is lower request volume, lower queue fan-out, and lower CPU/memory overhead without requiring existing connector or worker code changes.

## Additional information

PR #17211 implements this incrementally with focused benchmarks and regression coverage.

Representative measured improvements from that work include:

* selected STIX export accumulation for 10,000 objects: 39.9 s to 47 ms
* live 100-root full export: 108 GraphQL requests to 8 requests
* live 100 existing external-reference patch: 101 GraphQL requests to 2 requests
* modeled 50 ms async rule operation: 110,059 polls to 33 polls

Yes. PR #17211 is already open.

blauwers:
### Proposed changes

* This is not a breaking change for existing `pycti` users. Existing connector and worker code continues to work without modification, with compatibility defaults and old-platform fallbacks kept in place.
* Improve high-volume `pycti` ingestion and export paths by reducing repeated GraphQL work.
* Batch supported relation add/remove operations, prefetch repeated labels, vocabularies, external references, kill-chain phases, nested refs, and export targets, and reuse import/export scoped caches where the same data is requested repeatedly.
* Improve connector and worker throughput by reusing AMQP publisher connections, sending bounded same-dependency-level bundle chunks instead of one RabbitMQ message per object, and allowing opt-in bounded listener concurrency with delayed ACK/NACK settlement.
* Isolate connector request context so concurrent callbacks do not share work, draft, applicant, retry, or validation state.
* Reduce CPU and memory overhead in hot paths by linearizing list/deduplication work, streaming large upload/download paths where transport permits, and removing repeated extension lookups and small per-object allocations.
* Add focused benchmarks, regression coverage, and documentation for the new bundle chunking controls and `send_stix2_bundle()` return semantics.

### Related issues

* closes #17212

### How to test this PR

The full CI suite should run on the PR. For local review, the most relevant focused checks are:

1. Run the client Python suite:

   ```sh
   cd client-python
   python -m pytest
   ```

2. Run the focused backend validation for the bulk relation feature handshake and relation validation:

   ```sh
   yarn test:ci-unit tests/01-unit/utils/access-test.ts tests/01-unit/domain/stixObjectOrStixRelationship-test.ts
   yarn check-ts
   ```

3. Run the worker bundle chunking tests:

   ```sh
   cd opencti-worker
   python -m pytest tests/test_push_handler_bundle_chunking.py
   ```

4. Exercise a connector in AMQP queue mode and verify that:

   * large same-level bundles are chunked using `CONNECTOR_BUNDLE_SPLIT_MAX_OBJECTS` and `CONNECTOR_BUNDLE_SPLIT_MAX_BYTES`
   * setting `CONNECTOR_BUNDLE_SPLIT_MAX_OBJECTS=1` restores the previous one-object-per-message behavior
   * listener concurrency remains opt-in through `CONNECTOR_LISTEN_WORKER_COUNT`

### Checklist

### Further comments

This is a broad performance pass, but the changes were made incrementally and benchmarked one at a time. The PR focuses on paths that showed clear gains under representative high-volume scenarios rather than trying to optimize every possible code path.

Most changes are in `client-python`. The smaller platform and worker changes are limited to the capability-gated bulk relation API support and worker-side bundle chunking needed by the client optimizations.

Representative outcomes:

* Embedded `ObservedData` import with 4,000 observables: 346.5 ms to 9.4 ms.
* Selected STIX export accumulation with 10,000 objects: 39.9 s to 47 ms.
* AMQP listener benchmark with two 5 ms callbacks: 5,080 ms to 311 ms with worker count 2.
* Live 100-root full export: 108 GraphQL requests to 8 requests.
* Live 100 existing external-reference patch: 101 GraphQL requests to 2 requests.
* Live 100 ref removal patch: 100 requests to 1 request once the API feature cache is warm.
* Modeled 50 ms async rule operation: 110,059 polls to 33 polls.

Compatibility notes:

* Existing `pycti` callers do not need to change their code to use this branch.
* Listener concurrency still defaults to one worker.
* New bulk relation paths are capability-gated and keep old-platform fallbacks.
* AMQP bundle chunking defaults to 100 objects or 1 MB per chunk, and setting `*_BUNDLE_SPLIT_MAX_OBJECTS=1` restores the previous one-object chunking behavior.
* No connector or worker code changes are required for normal supported usage.

Validation notes:

* Each optimization was benchmarked before and after with focused regression coverage before commit.
* The branch passed the full clean CI-equivalent helper suite after the final performance fix.
* After rebasing onto current `master`, focused pycti tests, live vocabulary checks, backend unit tests, GraphQL typecheck, and Python formatting checks passed.


gabrielebonadiman:
Hello there @blauwers and thanks for the effort in this PR, the perf direction is solid and the numbers are compelling.

But at 36.7k additions / 3.2k deletions across 324 files, this isn't reviewable as one unit, and it currently has merge conflicts against master.

**Breakdown of what's in the diff:**

| Area | Lines |
|---|---|
| New benchmarks (`client-python/benchmarks`) | ~18.5k |
| Tests | ~11k |
| `pycti` library changes | ~6.4k |
| Worker + backend (bulk-relation capability, bundle chunking) | ~850 |

The last two are the real payload. The first two are useful evidence but shouldn't gate review of the core logic, and 150 new benchmark files is a lot to maintain going forward.

**Would you be open to splitting this into a stack of smaller PRs, each linked to its own issue?** It'll make review realistic and let us evaluate and merge each piece on its own merits, rather than as an all-or-nothing change. A rough breakdown that would help:

1. Backend bulk-relation capability + worker bundle chunking, small, isolated, easiest to land first.
2. `pycti` caching/prefetch, one lookup type at a time (labels, then vocabs, then external-refs, etc.) rather than all at once.
3. AMQP publisher reuse + opt-in listener concurrency.
4. Benchmarks/tests scoped per-PR above, not as a single upfront dump.

Rebase each against current `master` before resubmitting. Happy to review as soon as they're split, the underlying work looks genuinely valuable.

PR #17222, opened 2026-07-20
feat(dashboard): Saved filters in widgets: Handle export, duplication, and public dashboards (#15689)

### LINKED ISSUE #15689: feat(dashboards): reusable saved filters in dashboards
Dashboard widget configuration in OpenCTI requires users to manually define filters for each widget individually. When a team needs to monitor the same threat collection across multiple widgets or dashboards, they must recreate the same filter logic repeatedly. There is currently no way to reference a centrally-defined filter set as the data source of a widget. This leads to inconsistencies between widgets, high maintenance overhead when threat scopes evolve, and duplicated configuration effort - especially for teams managing multiple dashboards.

Archidoit:
### Proposed changes
---- Handle duplication of widgets with saved filters
- When duplicating a dashboard or widget: keep the saved filter reference (already working)
- Add tests for dashboard with saved filters duplication
.

---- Handle export of widgets with saved filters
- When exporting a widget with saved filters: resolve the saved filters ids and replace it by the saved filters content
- Add tests for toWidgetExport
- Display an error if trying to export a widget with a not found saved filter : 
image

.
----- Handle public dashboard creation from a dashboard with widgets with saved filters
- When creating a public dashboard from a dashboard containing saved filters: resolve the saved filters ids and replace it by the saved filters content
- The public dashboard should work the same way as the custom dashboard with saved filters
- Associated backend tests

.
---- Add translations for frontend error messages that were not translated

### Related issues
partially fixes #15689


opencti-platform/opencti-graphql/src/modules/dashboard/dashboard-utils.ts
```diff
@@ -32,11 +33,52 @@ export const checkDashboardConfigurationImport = (type: string, parsedData: Conf
...
+};
+
 export const exportDashboardWidget = async (context: AuthContext, user: AuthUser, manifest: string, widgetId: string) => {
   const parsedManifest = fromB64(manifest ?? '{}');
   if (parsedManifest && isNotEmptyField(parsedManifest.widgets) && parsedManifest.widgets[widgetId]) {
     const widgetDefinition = parsedManifest.widgets[widgetId];
     delete widgetDefinition.id; // Remove current widget id
+    await resolveSavedFiltersInDataSelection(context, user, widgetDefinition);
     await convertWidgetsIds(context, user, [widgetDefinition], 'internal');
     const exportConfigration = {
```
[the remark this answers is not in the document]
Archidoit: done

opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectAskAI.tsx
```diff
@@ -145,9 +145,9 @@ const StixCoreObjectAskAI: FunctionComponent<StixCoreObjectAskAiProps> = ({
     // check paragraphs value is correct
     if (action === 'container-report' || action === 'summarize-files') {
       if (Number.isNaN(paragraphs)) {
-        MESSAGING$.notifyError('Number of paragraphs should be a number');
+        MESSAGING$.notifyError(t_i18n('Number of paragraphs should be a number'));
       } else if (paragraphs <= 0) {
-        MESSAGING$.notifyError('Number of paragraphs should be greather than 0');
+        MESSAGING$.notifyError(t_i18n('Number of paragraphs should be greather than 0'));
```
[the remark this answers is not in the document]
Archidoit: done

opencti-platform/opencti-graphql/src/modules/publicDashboard/publicDashboard-domain.ts
```diff
@@ -138,6 +139,48 @@ export const getAllowedMarkings = async (
...
+    Object.keys(parsedManifest.widgets).forEach((widgetId) => {
+      parsedManifest.widgets[widgetId].dataSelection = parsedManifest
+        .widgets[widgetId]
+        .dataSelection.map((selection: any) => {
+          return {
+            ...(selection.label && { label: selection.label }),
+            ...(selection.attribute && { attribute: selection.attribute }),
+            ...(selection.date_attribute && { date_attribute: selection.date_attribute }),
+            ...(selection.number && { number: selection.number }),
+            ...(selection.centerLat && { centerLat: selection.centerLat }),
```
esrevi: if centerLat is equal to 0, the issue with this is we are going to strip that key from the object because it is considered as false when we do
```
...(0 && { "foo": "bar }}
```

I think it would be good to wrap with isNotEmptyField instead

```
...(isNotEmptyField(selection.centerLat) && { centerLat: selection.centerLat }),
```

opencti-platform/opencti-graphql/src/modules/publicDashboard/publicDashboard-domain.ts
```diff
@@ -138,6 +139,48 @@ export const getAllowedMarkings = async (
...
+ * to only keep display-related properties (no filters, no query data).
+ */
+const createPublicManifest = (parsedManifest: any) => {
+  if (parsedManifest && isNotEmptyField(parsedManifest.widgets)) {
+    Object.keys(parsedManifest.widgets).forEach((widgetId) => {
+      parsedManifest.widgets[widgetId].dataSelection = parsedManifest
+        .widgets[widgetId]
+        .dataSelection.map((selection: any) => {
+          return {
+            ...(selection.label && { label: selection.label }),
```
esrevi, line 168: prefer not mutate parsedManifest, a better pattern would be to build new object from it like that for example:

```
// break into a function to make it more readable with type to avoid any

  type PublicManifestSelection = {
    label?: string;
    attribute?: string;
    date_attribute?: string;
    number?: number;
    centerLat?: number;
    centerLng?: number;
    zoom?: number;
    columns?: unknown;
    [key: string]: unknown; // to be kept or not ?
  };

  type PublicManifestWidget = Record<string, unknown> & {
    dataSelection?: PublicManifestSelection[];
  };

 const sanitizeSelection = (selection: PublicManifestSelection): PublicManifestSelection => {
    return {
      ...(isNotEmptyField(selection.label) && { label: selection.label }),
      ...(isNotEmptyField(selection.attribute) && { attribute: selection.attribute }),
      ...etc
    };
  };

// generate the public widgets data
  const publicWidgets = Object.fromEntries(
    Object.entries(parsedManifest.widgets as Record<string, PublicManifestWidget>).map(([widgetId, widget]) => {
      const publicDataSelection = Array.isArray(widget.dataSelection)
        ? widget.dataSelection.map(sanitizeSelection)
        : widget.dataSelection;
      return [widgetId, { ...widget, dataSelection: publicDataSelection }];
    }),
  );

// new object
const publicManifest = {
 ...parsedManifest,
widgets: publicWidgets
}

return toB64(publicManifest)

```
so it prevents accidental regressions if same object is reused for another case and keeps function clear as it returns a new output

opencti-platform/opencti-graphql/src/modules/dashboard/dashboard-utils.ts
```diff
@@ -32,12 +33,53 @@ export const checkDashboardConfigurationImport = (type: string, parsedData: Conf
...
+) => {
+  const dataSelection = widgetDefinition.dataSelection;
+  if (dataSelection) {
+    await Promise.all(dataSelection.map(async (selection: any) => {
+      if (isNotEmptyField(selection.filters_id)) {
+        const savedFilter = await findSavedFilter(context, user, selection.filters_id as string);
+        if (!savedFilter) {
+          throw FunctionalError('Saved filter not found', { widget: widgetDefinition.id, savedFilter: selection.filters_id });
+        }
+        selection.filters = JSON.parse(savedFilter.filters);
```
esrevi: should we guard with try/catch if filters is malformed?

opencti-platform/opencti-graphql/src/modules/dashboard/dashboard-utils.ts
```diff
@@ -32,12 +33,53 @@ export const checkDashboardConfigurationImport = (type: string, parsedData: Conf
...
+      if (isNotEmptyField(selection.dynamicFrom_id)) {
+        const savedFilter = await findSavedFilter(context, user, selection.dynamicFrom_id);
+        if (!savedFilter) {
+          throw FunctionalError('Saved filter not found', { widget: widgetDefinition.id, dynamicFromSavedFilter: selection.dynamicFrom_id });
+        }
+        selection.dynamicFrom = JSON.parse(savedFilter.filters);
+        selection.dynamicFrom_id = undefined;
+      }
+      if (isNotEmptyField(selection.dynamicTo_id)) {
+        const savedFilter = await findSavedFilter(context, user, selection.dynamicTo_id);
```
esrevi: we have 3 await findSavedFilter, maybe we could use Promise.all to resolve them, so it would be done in parallel since they look independant

opencti-platform/opencti-graphql/src/modules/publicDashboard/publicDashboard-domain.ts
Archidoit: I didn't change this part of code... I just separated the part dedicated to the public manifest creation for more clarity. So mutating parsedManifest was already the behavior... Maybe I can change this if no consequence

opencti-platform/opencti-graphql/src/modules/publicDashboard/publicDashboard-domain.ts
Archidoit: Same : it was the old behavior, moved in a specific function. But good catch, I think we can improve this

opencti-platform/opencti-graphql/src/modules/dashboard/dashboard-utils.ts
Archidoit: done

opencti-platform/opencti-graphql/src/modules/publicDashboard/publicDashboard-domain.ts
Archidoit: done

esrevi:
Could be done in another issue: 
Some parts of the app don't have the saved filters feature, such as when we are in cards list view in threat-actors for example, or other parts such as locations, and entities (except security platforms who has this feature)

Otherwise everything seems to be good!

esrevi (approved):

PR #17224, opened 2026-07-20
chore(sbom): allow to merge node deps sbom with docker sbom (#17219)

### LINKED ISSUE #17219: feat(sbom): generate a SBOM on node application
- Current SBOM generated on optimized docker images is incomplet, it need to be merged with the one on node js app

Probably using cyclonedx cli
```
cyclonedx-cli merge --input-files sbom-1.json sbom-2.json --output-file sbom-final.json
```

aHenryJard:
### Proposed changes

* Add a SBOM generation on node application production dependencies (exclude dev dependencies)
* Merge it with the docker generated SBOM to have a full SBOM at the end

```
Target structure
────────────────────
Level 0  OpenCTI (root)
Level 1  └ OpenCTI Platform
            │─ (manufacturer set)
Level 2 = opencti components
		|-all platform container dependencies
		|-platform frontend dependencies
		|-platform backend dependencies
		|-all worker container dependencies
		|-all worker application dependencies
Level 3 = all dependencies (incide each corresponding level 2 components)
		            ├─ @isaacs/fs-minipass
		            ├─ @npmcli/arborist
		            └─ ... all 332, 
```

### Related issues

* closes #17219
 
### How to test this PR

See one of the last run: https://github.com/OpenCTI-Platform/opencti/actions/runs/30101153900

Get a generated SBOM on PR, download locally and look inside if it's compliant. Final sbom are the one with version like:
image

They are also present in the aggregated artifact **opencti-rolling**

### Checklist


.github/actions/generate-sbom/action.yml
```diff
@@ -76,20 +80,26 @@ runs:
...
+          SBOM_FILE="sbom-${SHORT_NAME}-container-${SAFE_TAG}.cdx.json"
           syft "${IMAGE_REF}" \
-            -c "${{ github.workspace }}/.syft.yaml" \
-            -o "cyclonedx-json=${SAFE_IMAGE}-${SAFE_TAG}.cdx.json"
+            -c "${{ github.workspace }}/.github/.syft.yaml" \
+            -o "cyclonedx-json@1.6=${SBOM_FILE}"
+
+          echo "Injecting manufacturer metadata into ${SBOM_FILE}"
+          jq --slurpfile mfr "${{ github.workspace }}/.github/sbom-manufacturer.json" \
+            '.metadata.manufacturer = $mfr[0]' "${SBOM_FILE}" > tmp.cdx.json && mv tmp.cdx.json "${SBOM_FILE}"
```
gabrielebonadiman: The manufacturer object is being injected into `metadata.manufacturer` here, but `cyclonedx-cli merge --hierarchical` doesn't carry that field into the merged output. I checked the actual artifacts and confirmed it's dropped for every component except platform/worker, which only show `supplier` (set natively by syft's `source.supplier` config, not by this line).

Can we set `.metadata.component.manufacturer` instead? That's the field ACN and IBM both expect at the component level, and it's what actually survives the merge.

.github/workflows/cd-build-rolling-images-on-merge.yml
```diff
@@ -5,9 +5,126 @@ name: OpenCTI CD Rolling
...
+
+      - name: Generate graphql SBOM
+        working-directory: opencti-platform/opencti-graphql
+        run: |
+          syft dir:. \
+            --config ../../.github/.syft.yaml \
+            --select-catalogers javascript \
+            --output cyclonedx-json@1.6=sbom-opencti-graphql.cdx.json
+          jq --slurpfile mfr ../../.github/sbom-manufacturer.json \
+            '.metadata.manufacturer = $mfr[0]
```
gabrielebonadiman: Same manufacturer issue as in generate-sbom/action.yml. Setting `.metadata.manufacturer` here too, which gets silently dropped once cyclonedx-cli merges everything. Downloaded the actual `sbom-opencti-platform-7.260722.0.cdx.json` from the run and neither the root `opencti` node nor `opencti-front`/`opencti-graphql` have any manufacturer info left.

Same fix as above, `.metadata.component.manufacturer` on both this step and the front one at line 84.

.github/workflows/cd-build-rolling-images-on-merge.yml
```diff
@@ -97,23 +215,95 @@ jobs:
...
+
+      - name: Download worker docker SBOM
+        uses: actions/download-artifact@v8
+        with:
+          name: sbom-opencti-worker${{ matrix.suffix }}
+          path: sbom-work/
+
+      - name: Install cyclonedx-cli
+        run: |
+          curl -sSfL https://github.com/CycloneDX/cyclonedx-cli/releases/download/v0.33.1/cyclonedx-linux-x64 -o "$RUNNER_TEMP/cyclonedx-cli"
```
gabrielebonadiman, line 270: We're curling a cyclonedx-cli binary straight off a GitHub release with no checksum check, in a job whose whole job is producing a trustworthy supply chain artifact. Worth pinning a sha256 and verifying it, feels a bit ironic otherwise.

gabrielebonadiman (commented):
Tested the merged SBOMs from the linked run, they're all valid CycloneDX 1.7 (cyclonedx-cli validate passes clean on all three variants).

One thing worth flagging before merge, not necessarily blocking: I checked the output against the ACN "Modulo SBOM" guidance (single level of depth, flat Level 0/Level 1). 

Our merged output is hierarchical (opencti > platform/front/graphql/worker > their own deps), which doesn't satisfy that flat model. Fine if IBM's ask is the only target for this PR, just flagging in case ACN compliance comes up later, it'd need a separate non-hierarchical output.

Also noticed the root opencti node ends up listing itself as its own dependency in the merged `dependencies` array. Traced it, it's not coming from sbom-parent.cdx.json (that file has no dependencies section at all), so it looks like something cyclonedx-cli's merge is generating on its own.

aHenryJard:
> Tested the merged SBOMs from the linked run, they're all valid CycloneDX 1.7 (cyclonedx-cli validate passes clean on all three variants).
> 
> One thing worth flagging before merge, not necessarily blocking: I checked the output against the ACN "Modulo SBOM" guidance (single level of depth, flat Level 0/Level 1).
> 
> Our merged output is hierarchical (opencti > platform/front/graphql/worker > their own deps), which doesn't satisfy that flat model. Fine if IBM's ask is the only target for this PR, just flagging in case ACN compliance comes up later, it'd need a separate non-hierarchical output.
> 
> Also noticed the root opencti node ends up listing itself as its own dependency in the merged `dependencies` array. Traced it, it's not coming from sbom-parent.cdx.json (that file has no dependencies section at all), so it looks like something cyclonedx-cli's merge is generating on its own.

Thank you, I'm going to merge this first step and we can iterate on the structure in a follow up PR if needed.

PR #17237, opened 2026-07-21
fix(ux): cannot load entities/observables anymore when there is a large number of them (#13437)

### LINKED ISSUE #13437: fix(ux): cannot load entities/observables anymore when there is a large number of them
## Description

Sometime when loading a large number of entities (e.g., Indicators, Stix-Core-Objects, or other entities), OpenCTI only displays a limited subset of the items. When scrolling down or navigating to the next page, a large blank space appears, even though the total number of entities/indicators is larger and should still be displayed.

I observed that the GraphQL API sends many repeated queries but only one actual data-fetching query for the entities. For example:

ContainerAddStixCoreObjectsLinesQuery is executed only once with count: 100, while other lighter queries such as StixCoreObjectHistoryLinesRefetchQuery continue to be called repeatedly.
As a result, the UI stops loading new items and shows empty gaps or blank content even though pageInfo.hasNextPage = true.

This behavior occurs consistently when the total dataset is large.

## Environment

1. OS (where OpenCTI server runs): Ubuntu 22.04
2. OpenCTI version: 6.8.15

## Reproducible Steps

1. Click the Request for takedown or incident response,...
2. add any of entity of this

Image

3. As you see total entity

Image

4. and opencti cant load anymore

Image

## Expected Output
OpenCTI should continuously load entities when scrolling or navigating pages.

No blank areas should appear as long as hasNextPage = true.

GraphQL should correctly request the next page with a new cursor.
No need to load more other lightweight GraphQL queries (e.g., history log queries) are triggered repeatedly instead

## Actual Output

OpenCTI loads only an initial subset of entities.

When scrolling or loading more data, the UI displays a blank space instead of additional items.

GraphQL requests show that the main data query is executed only once (with the same cursor), and no new paginated queries are triggered.

Other lightweight GraphQL queries (e.g., history log queries) are triggered repeatedly instead.

## Additional information
Issue occurs when database contains a very large dataset.

GraphQL response indicates globalCount is large and hasNextPage = true, but pagination does not continue.

Includes sample problematic queries for debugging:

ContainerAddStixCoreObjectsLinesQuery

StixCoreObjectHistoryLinesRefetchQuery

## Screenshots (optional)

Image

Image

delemaf:
### Proposed changes

* handle infinite scroll pagination on add entity/observable
* fix dummy placeholder line padding

### Related issues

* closes #13437

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/common/containers/ContainerAddStixCoreObjects.jsx
+++ b/opencti-platform/opencti-front/src/private/components/common/containers/ContainerAddStixCoreObjects.jsx
@@ -1,4 +1,4 @@
-import React, { useRef, useState } from 'react';
+import React, { useState } from 'react';
 import IconButton from '@common/button/IconButton';
 import Fab from '@mui/material/Fab';
 import { Add } from '@mui/icons-material';
@@ -141,7 +141,6 @@ const ContainerAddStixCoreObjects = (props) => {
   } = viewStorage;
   const contextFilters = useBuildEntityTypeBasedFilterContext(targetStixCoreObjectTypes, filters);
 
-  const containerRef = useRef(null);
   const [mappingSearch, setMappingSearch] = useState(null);
   const [currentSelectedText, setCurrentSelectedText] = useState(selectedText);
   if (currentSelectedText !== selectedText) {
@@ -350,7 +349,7 @@ const ContainerAddStixCoreObjects = (props) => {
       >
         <QueryRenderer
           query={containerAddStixCoreObjectsLinesQuery}
-          variables={{ count: 100, ...searchPaginationOptions }}
+          variables={{ count: 25, ...searchPaginationOptions }}
           render={({ props: renderProps }) => (
             <ContainerAddStixCoreObjectsLines
               data={renderProps}
@@ -364,7 +363,6 @@ const ContainerAddStixCoreObjects = (props) => {
               onDelete={onDelete}
               setNumberOfElements={helpers.handleSetNumberOfElements}
               mapping={mapping}
-              containerRef={containerRef}
               enableReferences={enableReferences}
               onLabelClick={helpers.handleAddFilter}
             />
@@ -430,7 +428,6 @@ const ContainerAddStixCoreObjects = (props) => {
           }
         }}
         title={t_i18n('Add entities')}
-        containerRef={containerRef}
       >
         <>
           {renderSearchResults(searchPaginationOptions)}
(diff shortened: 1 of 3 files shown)


opencti-platform/opencti-front/src/private/components/common/containers/ContainerAddStixCoreObjects.jsx
```diff
@@ -350,7 +349,7 @@ const ContainerAddStixCoreObjects = (props) => {
       >
         <QueryRenderer
           query={containerAddStixCoreObjectsLinesQuery}
-          variables={{ count: 100, ...searchPaginationOptions }}
+          variables={{ count: 25, ...searchPaginationOptions }}
```
mcartaud, line 352: This settings doesn't seem to work as expected (or at least has I understand it) : I have 63 entities, when opening the drawer, 2 requests are fired, first one returns a list of 25 items, second on a list of 38. And both requests are fired at drawer opening, not the first one at opening and second one letter during scroll as I would expect.

delemaf: It's mainly because your window is large (I've the same), you can display around 100 entities, so it's querying twice 25 + 38 => 63. It's not ideal, I agree but the point was to align this count and query to the other list:
https://github.com/OpenCTI-Platform/opencti/blob/ff7da5e6b114a8c0a44bd6214f01bb4fa057d37a/opencti-platform/opencti-front/src/private/components/data/Feed.tsx#L104
https://github.com/OpenCTI-Platform/opencti/blob/ff7da5e6b114a8c0a44bd6214f01bb4fa057d37a/opencti-platform/opencti-front/src/private/components/data/Taxii.tsx#L105-L106
https://github.com/OpenCTI-Platform/opencti/blob/ff7da5e6b114a8c0a44bd6214f01bb4fa057d37a/opencti-platform/opencti-front/src/private/components/observations/indicators/IndicatorEntities.jsx#L115-L116

We should change this everywhere (in another issue) but maybe the choice was made for small screen to not having 100 entities a once (it may load faster). 🙏

lndrtrbn (approved):

opencti-platform/opencti-front/src/private/components/common/containers/ContainerAddStixCoreObjects.jsx
mcartaud: Ok so real issue is not the number of entities here, but the ref.
I am not against aligning the setting with other part of the app. Good for me thanks :)

mcartaud (approved):

PR #17261, opened 2026-07-22
fix(draft): allow removing update_linked entities from draft (#17256)

### LINKED ISSUE #17256: fix(Draft): unable to remove elements from Draft
## Description

When you are the creator of the data in the draft AND that there is another author, remove from draft does not work, entities are not remove from draft & even change the operation to udpate_link. 

## Environment

Latest

## Reproducible steps

1. Import with a document some entities, using step by step (importDoc only / draft creation/ AM enabled with you only as can manage)
2. Ensure that among the list of data ingested in the draft, some already exists with  another creator. (Or simply create an empty draft, and add in the draft some entities that have at least 2 creators, you and another creator)
3. For these entities, click on the checkbox to mutli-select mutliple and select the option "remove from draft"

Alternative scenario: 
- For entities in already "update_link" state, removing the entities, regardless of whether you are the creator or not does not work either: take an entity that is update link, and remove from draft (will remain as update link)

## Expected output

When the task has run, entities are removed from draft. 
## Actual output

When it has run:
- entity is still there
- the draft operation has changed to update_link (and you can notice, this is not consistent, that you are not in the creator anymore).

If it is an observable: 
- it will create a duplicated line, of the same observable
- one linewill be the "initial" draft operation, with both the creators
- one line will be the "update_link", without me as creator

## Additional information

You can this draft to repro the behavior: https://latest.octi.staging.filigran.io/dashboard/data/import/draft/66d02d71-e528-486f-aad8-897bfaa610be/entities
The file in this draft on this environement has been already ingested, meaning the entities already exsit in this environement. You just need to re-import th same draft & you'll have the starting conditions of the draft.

amiraifori:
Removing an entity from a draft that is the creator of and remove from draft did not work when the entity had a related draft-created or draft-deleted relation still pointing to it: the entity's own changes were correctly reverted but it stayed stuck as update_linked forever, since elRemoveElementFromDraft threw UnsupportedError on any update_linked element instead of re-evaluating it.

Also cascade the same check after a draft-created relation is removed, so an update_linked entity is cleaned up automatically once nothing links to it anymore, regardless of the order elements are processed in during a multi-select "remove from draft".

Fixes #17256

### Proposed changes

* `elRemoveElementFromDraft` now handles the `update_linked` case by re-checking whether any draft-created/deleted relation still targets the element, instead of unconditionally throwing `UnsupportedError`. If nothing links to it anymore, it is fully removed from draft; otherwise it stays `update_linked` (no-op, no error).
* After a draft-created relation is removed (`elRemoveCreateElementFromDraft`), its former endpoints are re-checked and any `update_linked` element that is no longer linked to anything is cascaded-cleaned automatically, so the outcome no longer depends on the order in which a multi-selected batch (entity + relation) is processed.

### Related issues
* closes #17256

### How to test this PR

1. Create a draft, exit it, and create a report/entity live (outside the draft).
2. Re-enter the draft and update that entity (e.g. change its description) so it gets `draft_operation: update`.
3. Still in the draft, create a second entity and link it to the first one via a core relationship (e.g. `related-to`) — this relation is `draft_operation: create`.
4. Click "remove from draft" on the first entity: its own change is reverted but it should remain visible in the draft, now as `update_linked` (previously this looked the same as before but the state was a dead end).
5. Click "remove from draft" on the linking relation: the first entity should now disappear from the draft entirely (previously it stayed stuck as `update_linked` forever).
6. Also verify clicking "remove from draft" again directly on an `update_linked` entity no longer errors (previously threw `UnsupportedError`).

Covered by the new integration test `should remove an update_linked entity from draft once its linking relation is gone (issue #17256)` in `draftWorkspace-test.ts`.

### Checklist

### Further comments

I verified this compiles clean (`yarn check-ts`) and lints clean (`yarn eslint`), and wrote an integration test reproducing the exact scenario from the bug report. I was not able to run the integration test suite itself or exercise this against a live running instance in my environment (no Elasticsearch/Redis/RabbitMQ stack available), so I left "I tested the code for its functionality" unchecked — please run the test suite and/or manually verify against a live draft before merging.
EOF


amiraifori:
@nino-filigran Pleaseif you don't mind reviewing my PR which is linked to an issue you raised.

amiraifori:
@OctaveLaventure
Good day,
I do hope you are alright.

Please can I get a review on my PR?

amiraifori:
Hi @OctaveLaventure,

I’ve updated the branch with the latest changes from master, resolved the merge conflicts and adjusted the expected report counts to account for the added test.

The workflows are currently awaiting maintainer approval, so the updated changes still need verification. Could you please approve the workflows and review the PR when you have a chance?

opencti-platform/opencti-graphql/src/database/draft-engine.ts
```diff
@@ -68,9 +108,10 @@ const elRemoveCreateElementFromDraft = async (context: AuthContext, user: AuthUs
   const draftRelationsElementsImpact = await computeDeleteElementsImpacts(relationToRemove, [element.internal_id], relationsToRemoveMap);
 
   // Clean up all denormalized rel impact of relations deletion, then delete all relations
-  // TODO: clean up UPDATE_LINKED impacted elements that no longer need to be in draft => how to know that an update_linked element can be safely removed?
   await elRemoveRelationConnection(context, user, draftRelationsElementsImpact);
   await elDeleteInstances(context, [element, ...draftCreatedRelations]);
+  // The relations we just removed may have been the only reason another element was kept as UPDATE_LINKED
+  await cascadeCleanupNowUnlinkedElements(context, user, relationToRemove, [element.internal_id]);
```
OctaveLaventure, line 114: Nice fix for the create-relation-removal path. However, `elRemoveDeleteElementFromDraft` / `removeDraftDeleteLinkedRelations` (the DELETE_LINKED relation removal path) still carries its original `TODO: clean up UPDATE_LINKED impacted elements...` comments and doesn't call `cascadeCleanupNowUnlinkedElements` (or an equivalent). Is the same order-dependency bug reproducible there too (entity update_linked, then its linking relation is deleted rather than removed-from-draft-as-created)? If so, worth doing it in this PR, or confirming it's out of scope for this fix.

PR #17265, opened 2026-07-22
feat(rabbitmq): Proposal B - split stix bundles in node.js before pushing to worker queue (#17265)

### LINKED ISSUE #17259: feat(rabbitmq): ensure all bundles sent from backend to worker are split backend side
## Use case

ensure all bundles sent from backend to worker are split backend side

gabrielebonadiman:
### Proposed changes

* Ported pycti's `OpenCTIStix2Splitter` to TypeScript, validated diff-free against 18 bundles.
* Added `buildSplitMessages`: splits outgoing worker bundles into per-object messages.
* Renamed `pushToWorkerForConnector` → `pushBundleToWorker`, now the single gateway for split + expectations tracking; un-exported internal `send`.
* Fixed: malformed bundle content no longer blocks publishing, forwards unsplit instead.
* Updated all 9 call sites to the new function/name.
* Added direct test coverage on `pushBundleToWorker` (valid + invalid input).

### Related issues

* Fixes #17259

### How to test this PR

* Unit tests: `cd opencti-platform/opencti-graphql && npx vitest run --config vitest.config.ts tests/01-unit/database/rabbitmq-test.ts tests/01-unit/utils/stix2-splitter-test.ts`
* Send a multi-object STIX bundle through any of the updated call sites (e.g. `stixBundlePush` mutation) and confirm the worker receives one message per object instead of the raw bundle, and the work/job reaches "complete" only once all chunks are processed.

### Checklist

### Further comments


gabrielebonadiman:
Heads up on the diff size: `tests/data/stix2-splitter/mitre_att_capec.json` accounts for ~70k of the ~71.6k additions. It's a real-world MITRE ATT&CK/CAPEC fixture (same file used in client-python's own test suite) used to verify the splitter produces the exact same 7016 expectations as the Python implementation on a large, adversarial-scale bundle. The actual code change is small: rabbitmq.js (+33/-2) plus four ~1-line call-site updates to thread context/user through.

opencti-platform/opencti-graphql/src/database/rabbitmq.js
```diff
@@ -695,8 +697,51 @@ export const rabbitMQIsAlive = async () => {
...
+  }
+  const splitter = new Stix2Splitter();
+  const { bundles } = splitter.splitBundleWithExpectations(bundleContent);
+  if (bundles.length <= 1) {
+    return null;
+  }
+  return bundles.map((bundle) => ({ ...message, content: toBase64(bundle), no_split: true }));
+};
+
+export const pushToWorkerForConnector = async (connectorId, message, context, user) => {
```
JeremyCloarec: A bit of a nitpick, but we almost always use context & user as the first two parameters in the signature of methods, it could be great to keep it like this here for consistency sake

opencti-platform/opencti-graphql/src/database/rabbitmq.js
```diff
@@ -695,8 +697,51 @@ export const rabbitMQIsAlive = async () => {
...
   const routingKey = pushRouting(connectorId);
+  const splitMessages = buildSplitMessages(message);
+  if (splitMessages) {
+    if (message.work_id && context && user) {
+      await updateExpectationsNumber(context, user, message.work_id, splitMessages.length);
```
JeremyCloarec: I think this PR could to a great opportunity to centralize all updateExpectationsNumber in this method. Right now, some callers do it themselves depending on the state of the bundle (if it's nosplit or size = 1), but we could remove all calls to updateExpectationsNumber from callers and let pushWorkerForConnector handle all cases

gabrielebonadiman: I too thought that this can be an opportunity to centralise some of the logic 👍

opencti-platform/opencti-graphql/src/database/rabbitmq.js
```diff
@@ -451,8 +453,13 @@ const amqpExecute = async (execute) => {
  * Note: Ordering is maintained when callers use sequential awaits.
  * In rare edge cases around connection failures, duplicate delivery
  * is possible (at-least-once semantics). Consumers should be idempotent.
+ *
+ * Not exported on purpose: this is the low-level publish primitive for this module.
+ * Callers outside rabbitmq.js must go through pushToConnector or pushBundleToWorker,
+ * so every message to a worker/connector queue is guaranteed to go through their
+ * respective splitting/expectations/routing logic instead of bypassing it.
  */
```
gabrielebonadiman, line 461: just a comment for the sake of the pr. no need to bring this to master

opencti-platform/opencti-graphql/src/domain/stix.js
```diff
@@ -67,22 +67,22 @@ export const sendStixBundle = async (context, user, connectorId, bundle, work_id
...
-    await pushToWorkerForConnector(connectorId, {
+    await pushBundleToWorker(context, user, connectorId, {
       type: 'bundle',
       applicant_id: user.internal_id,
       content,
       work_id: target_work_id,
       update: true,
+      // Only this call is responsible for the work's expectations when it created it; a caller
+      // supplying an existing work_id is expected to manage expectations for that work itself.
+      trackExpectations: createdWorkHere,
```
gabrielebonadiman, line 85: just a comment for the sake of the pr. no need to bring this to master

opencti-platform/opencti-graphql/src/database/rabbitmq.js
```diff
@@ -695,9 +702,65 @@ export const rabbitMQIsAlive = async () => {
...
+/**
+ * Publishes a message to a connector's worker queue.
+ * When message.trackExpectations is true and message.work_id is set, the work's expected
+ * completion count is updated to reflect the actual number of STIX objects being sent - whether
+ * the bundle ends up split into several messages or sent as a single one. This is opt-in
+ * (rather than inferred from work_id's presence) because some callers, like CSV ingestion, track
+ * a work's expectations themselves under a different model (one expectation for the whole job) and
+ * must not have it incremented again here.
+ */
+export const pushBundleToWorker = async (context, user, connectorId, message) => {
```
JeremyCloarec, line 754: Was there a specific reason for the method name change?

opencti-platform/opencti-graphql/src/database/rabbitmq.js
```diff
@@ -695,9 +702,65 @@ export const rabbitMQIsAlive = async () => {
...
   const routingKey = pushRouting(connectorId);
-  return send(WORKER_EXCHANGE, routingKey, JSON.stringify(message));
+  const { messages, expectations } = buildSplitMessages(message);
+  const shouldTrackExpectations = message.trackExpectations && message.work_id && context && user && expectations > 0;
```
JeremyCloarec, line 757: I don't get the usage of message.trackExpectations. I think we want to always track expectations here no? Are there use cases where we have a work linked to the message but we don't want to update the expectations here? I see that trackExpectations is not set in playbook-components, but I'm not sure it was intended?

opencti-platform/opencti-graphql/src/database/rabbitmq.js
gabrielebonadiman: not really any specific reason but `pushToWorkerForConnector` didn't describe what it actually does anymore (it splits bundles and centralizes expectations tracking, not just "push  for a connector"). `pushBundleToWorker` says what it does. Purely a clarity rename :)

opencti-platform/opencti-graphql/src/database/rabbitmq.js
gabrielebonadiman: It's actually a non-issue here by design :) this call never sets `work_id` in the first place (same as before this PR), so `trackExpectations` never gets evaluated either way.

On the broader point though: opt-in is deliberate, not an oversight. CSV ingestion already tracks its own expectations upfront for the whole job, so if we defaulted to "always track when `work_id` is set," we'd double-count there. Making it opt-in keeps that one exception explicit instead of quietly working around it. Every new caller has to think about it once, but I'd rather that than a silent double-count bug.

JeremyCloarec: But csv ingestion doesn't track its own expectations though, and now that we split it in backend and not in worker, the worker doesn't add expectations either, resulting in an incorrect tracking on the work;
image

opencti-platform/opencti-graphql/src/utils/stix2-splitter.ts
```diff
@@ -0,0 +1,351 @@
...
+    let bundleData: StixObject;
+    if (useJson) {
+      try {
+        bundleData = JSON.parse(bundle as string);
+      } catch (e) {
+        throw new Error('File data is not a valid JSON', { cause: e });
+      }
+    } else {
+      bundleData = bundle as StixObject;
+    }
```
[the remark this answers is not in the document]
gabrielebonadiman: splitBundleWithExpectations()  now resets  cacheIndex / cacheRefs / elements / incompatibleItems  at the top of the call, so the instance is safe to reuse. Added a test that reuses one instance across two bundles and asserts results match a fresh instance.

opencti-platform/opencti-graphql/src/database/rabbitmq.js
```diff
@@ -695,8 +697,37 @@ export const rabbitMQIsAlive = async () => {
...
+  if (message.type !== 'bundle' || message.no_split) {
+    return null;
+  }
+  const splitter = new Stix2Splitter();
+  const { bundles } = splitter.splitBundleWithExpectations(fromBase64(message.content));
+  if (bundles.length <= 1) {
+    return null;
+  }
+  return bundles.map((bundle) => ({ ...message, content: toBase64(bundle), no_split: true }));
+};
```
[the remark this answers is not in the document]
gabrielebonadiman: Intentional: we tried forwarding unparseable bundles unsplit and it broke real CI (Backend/Integration + Rules/Taxii tests). Throwing is correct — an unparseable bundle was never valid to publish. Keeping it, tests cover it.

opencti-platform/opencti-graphql/src/database/rabbitmq.js
gabrielebonadiman: Fixed.  csv-bundler.ts  now sets  trackExpectations: true , so the backend adds real per-object expectations at split time. Also fixed a related dedup-collapse bug that could report the wrong count. Tests added.

opencti-platform/opencti-graphql/src/database/rabbitmq.js
gabrielebonadiman: Fixed. pushBundleToWorker(context, user, connectorId, message)  now leads with  context, user  for consistency.

opencti-platform/opencti-graphql/src/domain/stix.js
```diff
@@ -67,22 +67,21 @@ export const sendStixBundle = async (context, user, connectorId, bundle, work_id
...
     const content = Buffer.from(bundle, 'utf-8').toString('base64');
-    await pushToWorkerForConnector(connectorId, {
+    await pushBundleToWorker(context, user, connectorId, {
       type: 'bundle',
       applicant_id: user.internal_id,
       content,
       work_id: target_work_id,
       update: true,
+      // Only track expectations here if this call created the work.
+      trackExpectations: createdWorkHere,
```
JeremyCloarec: Even if work wasn't created here, we still want to track expectations I'd say

opencti-platform/opencti-graphql/src/manager/taskManager.js
```diff
@@ -257,17 +257,17 @@ const buildAndSendBundle = async (context, user, task, objects, opts) => {
...
     type: 'bundle',
     applicant_id: user.id,
     content,
     work_id: task.work_id,
     draft_id: task.draft_context ?? null,
-    no_split: opts.forceNoSplit ?? false,
+    no_split: forceNoSplit,
+    // Only this call is responsible for the work's expectations when the worker won't split
+    // anything further (single object, or splitting explicitly disabled for this task).
+    trackExpectations: objects.length === 1 || forceNoSplit,
```
JeremyCloarec: we always want pushBundleToWorker to handle expectations now that expectations tracking was removed above

opencti-platform/opencti-graphql/src/database/rabbitmq.js
JeremyCloarec: I think trackExpectations  can be completely removed now: it's not useful for any of the callers

gabrielebonadiman: yeah I remember I introduced this but it aged out 😅 , it was the safe move at the time, but now every single caller either already opts in or is unconditialonnaly true in practice so I will just go ahead and implement the removal from the

JeremyCloarec (approved):

PR #17269, opened 2026-07-22
fix(smtp): OAuth refresh token rotation (#15924)

ludovic:
### Proposed changes

* Persist a rotated SMTP OAuth refresh token when the identity provider returns a new one.
* Keep the existing refresh token when the provider returns none or returns the same value.
* Encrypt rotated tokens before storage and avoid overwriting a token that was already rotated concurrently.
* Add unit coverage for token rotation, encrypted persistence, and concurrent update protection.

### Related issues

* Part of #15924

### How to test this PR

* Run `yarn vitest run tests/01-unit/database/smtp-test.ts tests/01-unit/modules/smtpConfiguration/smtpConfiguration-domain-test.ts` from `opencti-platform/opencti-graphql`.
* Run `NODE_OPTIONS="--max-old-space-size=8192" ./node_modules/.bin/tsc --noEmit` from `opencti-platform/opencti-graphql`.

### Checklist

### Further comments

The access token remains an in-memory value. The database is updated only when the OAuth provider returns a non-empty refresh token that differs from the one used for the grant.

### DIFF
--- a/opencti-platform/opencti-graphql/src/database/smtp.js
+++ b/opencti-platform/opencti-graphql/src/database/smtp.js
@@ -6,7 +6,7 @@ import { getEntityFromCache } from './cache';
 import { ENTITY_TYPE_SETTINGS } from '../schema/internalObject';
 import { executionContext, SYSTEM_USER } from '../utils/access';
 import { isEmptyField } from './utils';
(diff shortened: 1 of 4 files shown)


marieflorescontact (approved):
locally tested ok :)

Issue #16214, opened 2026-05-27
chore(telemetry): improve observability


fellowseb:
### Description

This epic lists the tasks related to improving the Observability of OpenCTI. This can include working on :

- how/what/where data points are collected (metrics, traces, (structured) logs)
- what tools we use to visualize the data

PR #17282, opened 2026-07-23
fix(stream): fix number values in filter (#17275)

### LINKED ISSUE #17275: fix(stream): Can't access streams list
## Description
An error occurs when accessing live streams in demo

## Environment
demo

## Reproducible steps
Go under Data Sharing > Live streams. An error occurs.

link in demo : https://demo.octi.filigran.io/dashboard/data/sharing/streams

Image

JeremyCloarec:
### Proposed changes

This pull request addresses type safety issues when handling filter values in the filter components and utility functions. The main focus is to ensure that filter values are consistently treated as strings, which prevents potential runtime errors when performing string operations.

**Type Handling Improvements:**

* In `FilterValues.tsx`, ensures that the `value` used for filters is always a string by converting numeric values to strings before rendering.
* In `filtersUtils.tsx`, adds a defensive check in the `filterValue` function to return `null` if the value is not a string, preventing errors from string operations on non-string types.

### Related issues

* closes #17275

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/components/filters/FilterValues.tsx
+++ b/opencti-platform/opencti-front/src/components/filters/FilterValues.tsx
@@ -157,7 +157,10 @@ const FilterValues: FunctionComponent<FilterValuesProps> = ({
           fontFamily: 'Consolas, monaco, monospace',
         };
     const operatorOnClick = isLocalModeSwitchable ? () => handleSwitchLocalMode(currentFilter) : undefined;
-    const value = filtersRepresentativesMap.get(id) ? filtersRepresentativesMap.get(id)?.value : id;
+    let value = filtersRepresentativesMap.get(id) ? filtersRepresentativesMap.get(id)?.value : id;
+    if (typeof value === 'number') {
+      value = value.toString();
+    }
     const isRegardingOfFilter = parentFilter?.key === 'regardingOf' || parentFilter?.key === 'dynamicRegardingOf';
     return (
       <Fragment key={id}>
--- a/opencti-platform/opencti-front/src/components/filters/FilterValues.tsx
+++ b/opencti-platform/opencti-front/src/components/filters/FilterValues.tsx
@@ -157,10 +157,7 @@ const FilterValues: FunctionComponent<FilterValuesProps> = ({
           fontFamily: 'Consolas, monaco, monospace',
         };
     const operatorOnClick = isLocalModeSwitchable ? () => handleSwitchLocalMode(currentFilter) : undefined;
-    let value = filtersRepresentativesMap.get(id) ? filtersRepresentativesMap.get(id)?.value : id;
-    if (typeof value === 'number') {
-      value = value.toString();
-    }
+    const value = filtersRepresentativesMap.get(id) ? filtersRepresentativesMap.get(id)?.value : id;
     const isRegardingOfFilter = parentFilter?.key === 'regardingOf' || parentFilter?.key === 'dynamicRegardingOf';
     return (
       <Fragment key={id}>
--- a/opencti-platform/opencti-front/src/utils/filters/filtersUtils.tsx
+++ b/opencti-platform/opencti-front/src/utils/filters/filtersUtils.tsx
@@ -499,6 +499,11 @@ export const filterValue = (
   if (filterKey === 'relationship_type' || filterKey === 'type') {
     return t_i18n(`relationship_${value}`);
   }
+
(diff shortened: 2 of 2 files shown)


opencti-platform/opencti-front/src/components/filters/FilterValues.tsx
```diff
@@ -157,7 +157,10 @@ const FilterValues: FunctionComponent<FilterValuesProps> = ({
           fontFamily: 'Consolas, monaco, monospace',
         };
     const operatorOnClick = isLocalModeSwitchable ? () => handleSwitchLocalMode(currentFilter) : undefined;
-    const value = filtersRepresentativesMap.get(id) ? filtersRepresentativesMap.get(id)?.value : id;
+    let value = filtersRepresentativesMap.get(id) ? filtersRepresentativesMap.get(id)?.value : id;
+    if (typeof value === 'number') {
```
mcartaud: What about other types? boolean, etc
This is probably unlikely but if we are referring to the type, number should not be possible either

opencti-platform/opencti-front/src/utils/filters/filtersUtils.tsx
```diff
@@ -499,6 +499,11 @@ export const filterValue = (
   if (filterKey === 'relationship_type' || filterKey === 'type') {
     return t_i18n(`relationship_${value}`);
   }
+
+  // Defensive check to prevent errors on string manipulation after this call
+  if (typeof value !== 'string') {
+    return null;
+  }
   return value;
```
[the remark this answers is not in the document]
mcartaud: It seem to be a valid concern, what do you think?

opencti-platform/opencti-front/src/utils/filters/filtersUtils.tsx
```diff
@@ -499,6 +499,11 @@ export const filterValue = (
   if (filterKey === 'relationship_type' || filterKey === 'type') {
     return t_i18n(`relationship_${value}`);
   }
+
+  // Defensive check to prevent errors on string manipulation after this call
+  if (typeof value !== 'string') {
```
mcartaud: We should also add a case for number no? Like for boolean, returning `String(value)`?

opencti-platform/opencti-front/src/components/filters/FilterValues.tsx
JeremyCloarec: I could handle more types, but I tried to minimize the uses cases we're targeting for a fix: we shouldn't have any number value in the first place, but it looks like some filters have number values in their confidence filter Image. 
My guess was that it was related to a previous filter format that we no longer use: if you try to create a filter on confidence, you'll have a string value and not a number value anymore. Same thing for boolean filters, the value is set as a string and not a boolean

opencti-platform/opencti-front/src/utils/filters/filtersUtils.tsx
JeremyCloarec: My idea was that it's better to have a null return value/a deleted value displayed, since value should never be anything other than a string here. That way if it happens we'll be able to detect it and try to find out why we're sending something other than a string here

fellowseb: This defensive check is too late in the function, no ? Shouldn't we hoist it before other returns ?

fellowseb: The ideal IMO would be to normalize to a string here and log an error so we could monitor and fix those cases while not breaking current UX. Because I'm not sure "we'll be able to detect it" by returning null 🤔 .

The issue is we don't have a good way to log warnings/errors in the frontend without throwing in the component tree. How about normalizing to string for now, and treat the observability later (add a comment so we don't forget to add the log) ?
I created this ticket to improve the logging capabilities in the frontend: https://github.com/OpenCTI-Platform/opencti/issues/16214?issue=OpenCTI-Platform%7Copencti%7C17400

JeremyCloarec: Ok let's do observability on an other ticket 👍 
I've modified filterValue to normalize all values coming in

JeremyCloarec: Some other returns already expect value to not be a string (we're return JSON.stringify(value) above), so I think it's better to not normalize the value before the last return

fellowseb (approved):

PR #17331, opened 2026-07-26
fix(frontend): render non-hex label colors correctly (#17238)

### LINKED ISSUE #17238: fix(labels): plain text label colors are not properly handled
## Description

When creating a new label, user can use the color picker to select a color, or enter an arbitrary string. 
If this string is not an hex color code, the label display in Tags component is not able to display the color.
What is erroneous is that if user write "red" for instance in the color selection, the read color appear next to the color picker icon, making user think it is working, but after validating the color won't be shown when the label is displayed in a Tag component

Image

Image

Image

Image

## Environment

2. Version: { e.g. 1.2.3 }
3. Other environment details:

## Reproducible steps

1. Go to any observable
2. add a new tag, use the + button to create a new one
3. use `red` as color value
4. validate

## Expected output

The chosen color is shown everywhere

## Actual output

the label is gray everywhere it is displayed in a Tag (meaning in taxonomy page, we see the correct color)

## Additional information

The correct fix is probably to add a normalization step in the backend on label creation / update that will convert string color into hex color.
Another solution can be found in frontend by using something else than `sx` property to define color in Tag component

guillaume-flambard:
## Problem

When creating a label, the colour field accepts free text. If the value is not a hex code (say `red`), the colour picker preview shows it, but the Tag component renders the label grey: MUI's `alpha()` cannot parse a CSS colour name, throws, and callers fall back to the default colour. Labels already stored with a non-hex value are broken the same way.

Closes #17238

## Fix

Handle non-hex colours at the renderer instead of restricting the form:

- `normalizeLabelColor` in `src/utils/Colors.ts` maps any stored colour to a form MUI can parse: the 148 CSS named colours resolve to hex, 3/4-digit hex shorthand is expanded, 8-digit hex and `rgb()/hsl()` notations pass through.
- `Tag.tsx` normalizes before calling `alpha()`, with the existing try/catch as a last resort, so previously-created non-hex labels render correctly.
- The creation and edition forms go back to plain `required()`: free text stays allowed, garbage falls back to the default colour instead of being blocked.

## Tests

`src/utils/Colors.test.ts`:

- resolver coverage: hex passthrough and expansion, named colours, rgb()/hsl() passthrough, whitespace, malformed and empty values, full 148-name map
- the exact path from the issue: `alpha('red')` throws, `alpha(normalizeLabelColor('red'), 0.2)` returns `rgba(255, 0, 0, 0.2)`

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/settings/labels/LabelCreation.tsx
+++ b/opencti-platform/opencti-front/src/private/components/settings/labels/LabelCreation.tsx
@@ -71,8 +71,11 @@ const LabelCreation: FunctionComponent<LabelCreationProps> = ({
   const labelValidation = Yup.object().shape({
     value: Yup.string().required(t_i18n('This field is required')),
     color: Yup.string()
-      .matches(HEX_COLOR_REGEX, t_i18n('Invalid hexa color code'))
-      .required(t_i18n('This field is required')),
+      .required(t_i18n('This field is required'))
+      .matches(HEX_COLOR_REGEX, {
+        message: t_i18n('Invalid hexa color code'),
+        excludeEmptyString: true,
+      }),
   });
   const initialValues: LabelAddInput = {
     value: contextual ? inputValueContextual : '',
--- a/opencti-platform/opencti-front/src/private/components/settings/labels/LabelCreation.tsx
+++ b/opencti-platform/opencti-front/src/private/components/settings/labels/LabelCreation.tsx
@@ -16,6 +16,7 @@ import { commitMutation, defaultCommitMutation, handleErrorInForm } from '../../
 import { insertNode } from '../../../../utils/store';
 import Drawer, { DrawerControlledDialProps } from '../../common/drawer/Drawer';
 import { LabelAddInput, LabelCreationContextualMutation$data } from './__generated__/LabelCreationContextualMutation.graphql';
+import { HEX_COLOR_REGEX } from '../../../../utils/Colors';
 
 const labelMutation = graphql`
   mutation LabelCreationMutation($input: LabelAddInput!) {
@@ -69,7 +70,9 @@ const LabelCreation: FunctionComponent<LabelCreationProps> = ({
   const { t_i18n } = useFormatter();
   const labelValidation = Yup.object().shape({
     value: Yup.string().required(t_i18n('This field is required')),
-    color: Yup.string().required(t_i18n('This field is required')),
+    color: Yup.string()
+      .matches(HEX_COLOR_REGEX, t_i18n('Invalid hexa color code'))
+      .required(t_i18n('This field is required')),
   });
   const initialValues: LabelAddInput = {
     value: contextual ? inputValueContextual : '',
--- a/opencti-platform/opencti-front/src/utils/Colors.ts
+++ b/opencti-platform/opencti-front/src/utils/Colors.ts
@@ -269,6 +269,15 @@ export const itemColor = (
   return stringToColour(normalizedType, reversed);
 };
 
+/**
+ * Colours entered by users are rendered through MUI's alpha(), which only understands hex,
+ * rgb() and hsl(). A CSS colour name such as "red" makes alpha() throw, callers fall back to a
+ * default colour, and the user sees their choice silently ignored. Validate at the form instead.
+ */
+export const HEX_COLOR_REGEX = /^#([A-Fa-f0-9]{6}|[A-Fa-f0-9]{3})$/;
+
+export const isHexColor = (color?: string | null) => !!color && HEX_COLOR_REGEX.test(color);
+
 export const hexToRGB = (hex?: string, transp: number = 0.1) => {
   if (!hex) return `rgb(${50}, ${50}, ${50}, ${transp})`;
   const r = parseInt(hex.slice(1, 3), 16);
--- a/opencti-platform/opencti-front/src/utils/Colors.ts
+++ b/opencti-platform/opencti-front/src/utils/Colors.ts
@@ -331,6 +331,234 @@ export const isWashVisibleOn = (
   return dE >= MARKING_MIN_DELTA_E;
 };
 
+// The 148 CSS named colours, resolved to the hex the design system's Chip can read.
+export const CSS_NAMED_COLORS: Record<string, string> = {
+  aliceblue: '#f0f8ff',
+  antiquewhite: '#faebd7',
+  aqua: '#00ffff',
+  aquamarine: '#7fffd4',
+  azure: '#f0ffff',
+  beige: '#f5f5dc',
+  bisque: '#ffe4c4',
+  black: '#000000',
+  blanchedalmond: '#ffebcd',
+  blue: '#0000ff',
+  blueviolet: '#8a2be2',
+  brown: '#a52a2a',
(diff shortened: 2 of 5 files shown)


guillaume-flambard:
Pushed a small follow-up addressing Copilot’s Yup note: `.required()` now runs before `.matches()`, with `excludeEmptyString: true`, so an empty color shows the required message instead of “Invalid hexa color code”.

Archidoit:
This is not exactly how we want to solve this issue. We want to correctly handled non-hex color rather than forcing the user to use hex colors. What's more, your changes doesn't apply for labels already created with a non-hex color...

opencti-platform/opencti-front/src/private/components/settings/labels/LabelCreation.tsx
```diff
@@ -69,7 +70,12 @@ const LabelCreation: FunctionComponent<LabelCreationProps> = ({
   const { t_i18n } = useFormatter();
   const labelValidation = Yup.object().shape({
     value: Yup.string().required(t_i18n('This field is required')),
-    color: Yup.string().required(t_i18n('This field is required')),
+    color: Yup.string()
```
Archidoit, line 73: We want to autorize labels such as 'red', they should be handled rather than forbidden.

Archidoit (changes requested):

guillaume-flambard:
Thanks for the steer @Archidoit. You are right that the fix belongs at render time, and form validation would not help labels that already exist with a non-hex color. I am closing this rather than leave the wrong approach open. #17238 still needs the rendering fix, hopefully from someone with more bandwidth here than I have right now.

guillaume-flambard:
Thanks for the direction. Reworked in 5c76a26: instead of rejecting non-hex colors at the form, the renderer now handles whatever is stored.

normalizeLabelColor (src/utils/Colors.ts) resolves the 148 CSS named colours to hex, expands 3/4-digit shorthand, and passes rgb()/hsl() notations through, so MUI's alpha() and lighten() always receive something they can parse. Tag.tsx normalizes before tinting, so labels already created with a non-hex value render correctly too, and only a value nothing CSS can understand falls back to the default colour. The forms go back to just requiring a non-empty colour.

Tests in src/utils/Colors.test.ts cover the resolver and the exact failure path from the issue: alpha('red') throws, alpha(normalizeLabelColor('red')) does not.

guillaume-flambard:
Reopening: the rework above (5c76a26) landed a few minutes after this was closed, and it implements exactly the direction from the review. Render-time normalization, existing non-hex labels included, forms no longer force hex. The close crossed paths with the push, sorry for the crossed wires.

guillaume-flambard:
@Archidoit thanks for the steer, I reworked the PR around it. Labels like `red` are now handled rather than forbidden.

The form no longer rejects anything. `normalizeLabelColor` in `src/utils/Colors.ts` resolves a label colour to something the renderer can actually use:

- hex, with `#abc` and `#abcd` expanded to their 6 and 8 digit form
- the 148 CSS named colours
- `rgb()`, `rgba()`, `hsl()`, `hsla()`
- anything else returns `null`, and `Tag` falls back to the default colour instead of rendering a broken chip

`Tag.tsx` calls it in `getBackgroundColor`, so the preview and the rendered label finally agree, which was the original mismatch in #17238.

13 unit tests cover the cases. Rebased on current master, commits signed, checks green.

Ready for another look when you have a moment.

guillaume-flambard:
Reworked per @Archidoit feedback (render-time normalization via normalizeLabelColor, forms back to required only). Existing non-hex labels now render correctly, 13 tests cover resolver and Tag path. Ready for another look when you have a moment. d039229

guillaume-flambard:
@Archidoit before you look at this again: it is not the version you turned down, and the ground moved under it twice.

First, the rework you asked for. Validating at the form was the wrong place, and it did nothing for labels already stored with a non-hex colour. That is gone. The fix happens at render time and covers existing labels.

Second, #17664 landed in the meantime. The design system V1 rewrote `Tag.tsx`: the MUI block with `alpha()` and `lighten()` is gone, and the component now passes `color` straight to `@filigran/design-system`'s `Chip`. My earlier patch hooked into `getBackgroundColor`, which no longer exists, so the branch is rebased onto current `master` and the hook moved.

The bug survives that migration. `Chip` reads a colour through `parseDataColor`, whose regex is:

```js
const HEX_RE = /^#?(?:[0-9a-f]{3}|[0-9a-f]{6})$/i;
```

Anything else is ignored, the chip falls back to its `entity`/`severity` tone, and the library logs a warning saying so. A label stored as `red` still renders with the fallback colour, which is what #17238 reports.

What the branch does now, in one commit on top of `master`:

- `normalizeLabelColor` resolves a stored colour to `#rrggbb`, the only form the Chip reads. The 148 CSS colour names, hex of 3, 4, 6 or 8 digits, `rgb()`, `rgba()`, `hsl()` and `hsla()` are converted rather than forwarded. Anything else returns `null` and the Chip falls back exactly as it does today.
- The alpha channel is dropped rather than passed on, because `Chip` rejects `#rgba` and `#rrggbbaa` on purpose: the library comment says such a value would multiply into its own bounded wash.
- The leading `#` is optional on the way in. Their regex accepts a bare `70d907`, so a label already stored without the hash keeps rendering. I had this wrong in a first pass and caught it by testing against the library rather than against my own expectations.
- `Tag.tsx` normalises once, on the line that hands `color` to `Chip`.
- The forms are untouched. Nothing new is rejected at input time.

On verification: I did not install the front workspace, so beyond the vitest cases I ran every output through `parseDataColor` from `@filigran/design-system@1.1.0`, the version pinned in `package.json`. 165 values, all accepted, and the values the library refuses return `null` here. The unit tests assert the two properties directly as well: no output can violate the Chip regex, and all 148 names resolve.

Your review from 6 August still stands on the PR, so it shows as blocked. I can split this up if you would rather review it smaller.

guillaume-flambard:
@Archidoit one ping on this, then I will leave it alone.

It is still a single commit on top of `master`, tests green, and the version you can see is the rework from 3 September. What keeps the PR blocked is the review you left on 6 August: GitHub still counts it as standing, so the PR cannot move even when nobody objects to the current diff.

If a smaller diff would be easier to read, say the word and I will split the resolver from the tests. If it is simply not a priority right now, that is a fine answer too and I will stop here.

Archidoit:
@guillaume-flambard I was off and not really available lately. I'll have a look to your PR in the coming days :) Thanks again for your contribution !

opencti-platform/opencti-front/src/utils/Colors.ts
```diff
@@ -331,6 +331,234 @@ export const isWashVisibleOn = (
   return dE >= MARKING_MIN_DELTA_E;
 };
 
+// The 148 CSS named colours, resolved to the hex the design system's Chip can read.
+export const CSS_NAMED_COLORS: Record<string, string> = {
```
Archidoit: I'm not a big fan of having hardcoded values like this... You can have a look at other methods that convert directly for you.
Some methods create a canvas to fetch the hex color.

JS example:
const ctx = document.createElement('canvas').getContext('2d');
  ctx.fillStyle = name;
  return ctx.fillStyle; // returns hex (or rgb) string

PR #17344, opened 2026-07-27
fix(backend): use both range ends in computeRangeIntersection fallback (#17342)

### LINKED ISSUE #17342: fix(backend): computeRangeIntersection ignores a.end when ranges do not intersect
## Description

In `src/utils/format.js`, `computeRangeIntersection` builds the fallback span for two non-intersecting ranges as:

```js
const minStart = moment.min([a.start, b.start]);
const maxStop = moment.max([b.end, b.end]);   // <-- uses b.end twice, ignores a.end
```

The `maxStop` line uses `b.end` twice (copy-paste), so range `a`'s end is ignored. The `minStart` line just above correctly uses both (`[a.start, b.start]`).

## Impact

When the two ranges do not intersect and `a` ends **after** `b`, the computed end is wrongly truncated to `b.end` (which can even be before `minStart`). `computeRangeIntersection` is used by the inference rules `SightingObservableRule` / `SightingIndicatorRule` to set the `first_seen`/`last_seen` (and `start_time`/`stop_time`) of inferred relationships, so inferred edges get an incorrect validity window.

## Suggested fix

`const maxStop = moment.max([a.end, b.end]);`

## Environment

master branch

YAMRAJ13y:
## Proposed changes

`computeRangeIntersection` (`src/utils/format.js`) computed the fallback span for two non-intersecting ranges with `moment.max([b.end, b.end])`, which ignores `a.end` (copy-paste — the `minStart` line just above correctly uses `[a.start, b.start]`).

When the ranges do not intersect and `a` ends after `b`, the computed `end` was wrongly truncated to `b.end`. This function feeds the sighting inference rules (`SightingObservableRule`/`SightingIndicatorRule`) that set `first_seen`/`last_seen` (`start_time`/`stop_time`) on inferred relationships, so inferred edges got an incorrect validity window.

Fix: `moment.max([a.end, b.end])`.

## Related issues

Closes #17342

## How to test

```bash
yarn vitest run --config vitest.config.ci-unit.ts tests/01-unit/utils/format-utils-test.ts
```

Adds `computeRangeIntersection` tests for the intersecting case and the non-intersecting case where `a` ends after `b` (the latter fails without the fix). 10/10 tests pass.

### DIFF
--- a/opencti-platform/opencti-graphql/src/utils/format.js
+++ b/opencti-platform/opencti-graphql/src/utils/format.js
@@ -127,7 +127,7 @@ export const computeRangeIntersection = (a, b) => {
   }
   // No range intersection, get min/max to build the range
   const minStart = moment.min([a.start, b.start]);
-  const maxStop = moment.max([b.end, b.end]);
+  const maxStop = moment.max([a.end, b.end]);
   return { start: minStart.toISOString(), end: maxStop.toISOString() };
 };
 
--- a/opencti-platform/opencti-graphql/tests/01-unit/utils/format-utils-test.ts
+++ b/opencti-platform/opencti-graphql/tests/01-unit/utils/format-utils-test.ts
@@ -1,5 +1,24 @@
 import { describe, expect, it } from 'vitest';
-import { checkScore } from '../../../src/utils/format';
+import { buildPeriodFromDates, checkScore, computeRangeIntersection } from '../../../src/utils/format';
+
+describe('computeRangeIntersection tests', () => {
+  it('should return the overlapping window when the two ranges intersect', () => {
+    const a = buildPeriodFromDates('2020-01-01T00:00:00.000Z', '2020-06-01T00:00:00.000Z');
+    const b = buildPeriodFromDates('2020-03-01T00:00:00.000Z', '2020-09-01T00:00:00.000Z');
+    const result = computeRangeIntersection(a, b);
+    expect(result.start).toEqual('2020-03-01T00:00:00.000Z');
+    expect(result.end).toEqual('2020-06-01T00:00:00.000Z');
+  });
+
+  it('should span both ranges (min start, max end) when they do not intersect', () => {
+    // a ends AFTER b: the max end must be a.end, not b.end.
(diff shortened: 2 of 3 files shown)


xfournet:
Thanks for this — the diagnosis and the one-line fix are both correct.

The only thing blocking this PR was the `Backend / Rules and Taxii tests` job, failing on `tests/20-rules/relationToRelationBuilder-test.ts`. That test asserted the buggy behaviour: a created relation covering 2020 combined with an existing one covering 2019 — two periods that do not overlap — expected the inferred relation to stop on `2019-12-31`, i.e. the end of the existing relation alone, discarding the whole of 2020. In other words, the exact symptom you reported was frozen into an assertion.

I pushed a commit to your branch updating the three expectations in that file to the enclosing period `2019-01-01 → 2020-12-31`. Nothing else changed; your `format.js` fix and your unit tests are untouched.

For the record, we ran into this independently while removing `moment-range` in #16968, and reached the same conclusion. That PR is now scoped back to the dependency removal only, so your fix stays the one that lands. `git log -S 'moment.max([b.end, b.end])'` shows the expression was introduced in d1cd7b8e5e655af89063b3eeedc958a4c3e0e2c1 (2021-09-01), the very commit that added the fallback, and shipped from `5.0.0` onwards.

Reproduced on a dev platform with the rule engine enabled, using the `attribution_targets` rule, before and after the fix:

computeRangeIntersection: the non-overlapping fallback drops one of the two periods

Note that the impact is wider than the sighting rules mentioned in #17342: `computeRangeIntersection` also feeds `attribution_targets` and `attribution_use` (through `relationWithRelationBuilder.ts`), `localization_of_targets`, `indicate_sighted` and `infrastructure_observable_related`.

xfournet:
Rebased your branch onto current `master` — it was 260 commits behind. No conflicts: none of the three touched files had changed on `master` since your base. Your two commits are unchanged apart from their parent, and the unit tests still pass (10/10, with the non-intersecting case still failing when the fix is reverted).

Sorry for the force push on your branch; `maintainer_can_modify` was enabled, so I used it to unblock the CI rather than leave the PR waiting.

fellowseb (approved):

PR #17393, opened 2026-07-28
feat(accessibility): Resolve critical accessibility issues (#17312)

### LINKED ISSUE #17312: feat(accessibility): Resolve critical accessibility issues
## Use case

Upon scanning the application for section 508 compliance multiple critical accessibility issues were found. 

- Missing focus indicators
- Incorrect use of Modal component without a focus trap, dialog role, or close button
- QR code images with blank alt text
- OTP input cells with no accessible label
- Clickable div elements used as interactive controls
- Insufficient contrast for helper text
- Insufficient contrast for the severity "medium" color in the light theme 

## Current workaround

None

## Proposed solution

Resolve critical accessibility issues to meet WCAG guidelines

Yes

scarletmerlin123:
### Proposed changes

* Remove global focus outline removal and add style used by mui components when keyboard navigation is used
* Update <Modal> component to have correct focus trap, role, and a close button
* Add alt text to qr code
* Add aria-label to OTP input 
* Update interactive elements to use button instead of divs
* Update the color for contrast text to meet contrast requirements

### Related issues

* closes #17312 

### How to test this PR

- Requires manual review and use of a screen reader to verify label updates.

### Checklist


opencti-platform/opencti-front/src/components/dataGrid/components/DataTableHeader.tsx
```diff
@@ -107,13 +111,14 @@ const DataTableHeader: FunctionComponent<DataTableHeaderProps> = ({
       className={classes.headerContainer}
       style={{ width: cellWidth }}
     >
-      <div className={classes.label} onClick={throttleSortColumn}>
-        <Tooltip title={t_i18n(column.label)}>
-          <span>{t_i18n(column.label)}</span>
-        </Tooltip>
-        {sortBy && column.isSortable && (orderAsc ? <ArrowDropUp /> : <ArrowDropDown />)}
+      <div style={{ overflow: 'hidden' }} role="columnheader" aria-sort={sortBy ? (orderAsc ? 'ascending' : 'descending') : 'none'}>
```
mcartaud, line 114: The menu button that appear on hoover is not properly aligned anymore

Image

opencti-platform/opencti-front/src/components/filters/FilterValues.tsx
```diff
@@ -34,7 +37,23 @@ interface FilterValuesProps {
   filtersRestrictions?: FiltersRestrictions;
   host?: WidgetHost;
 }
-
+const useStyles = makeStyles((theme: Theme) => ({
```
mcartaud: theme should be prefixed with `_` if not used (`_theme`).

opencti-platform/opencti-front/src/components/filters/FilterValues.tsx
```diff
@@ -17,6 +17,9 @@ import FilterValuesForDynamicSubKey from './FilterValuesForDynamicSubKey';
 import { useTheme } from '@mui/material/styles';
 import { Stack } from '@mui/material';
 import type { WidgetHost } from '../../utils/widget/widget';
+import makeStyles from '@mui/styles/makeStyles';
+import clsx from 'clsx';
```
mcartaud: clsx is not a dependency of opencti-front project. Can you remove it because it is only used in this file and its behavior could be recreated with one or two util functions?

opencti-platform/opencti-front/src/components/markdownDisplay/MarkdownDisplay.tsx
```diff
@@ -267,9 +268,12 @@ const MarkdownDisplay: FunctionComponent<MarkdownWithRedirectionWarningProps> =
   } else {
     markdownDisplayContent = (
       <>
-        <div onClick={(event) => browseLinkWarning(event)}>
+        <Link
```
mcartaud, line 271: The same component is being used for instance when opening an observable history log entry or in a report notes. It use a link even if no action are provided, the wrapper should probably be around relevant elements in markdownRender rather than unconditionally here

Image

opencti-platform/opencti-front/src/components/markdownDisplay/MarkdownImagePreviewModal.tsx
```diff
@@ -95,53 +99,88 @@ const MarkdownImagePreviewModal: FunctionComponent<MarkdownImagePreviewModalProp
...
-            NavButton={({ onClick, next }) => {
-              const run = () => onClick?.();
-              if (next) {
-                nextActionRef.current = run;
-              } else {
-                prevActionRef.current = run;
-              }
+        <Box sx={{ position: 'relative', width: '100%' }}>
+          <IconButton
+            sx={{ color: 'rgba(255, 255, 255, 0.54)', position: 'absolute', top: 0, right: 0 }}
```
mcartaud, line 104: Is it possible to use a color from the them (like the lightGray used in PeriodicityField change? Or is the contrast not enough?

opencti-platform/opencti-front/src/components/ImageCarousel.tsx
```diff
@@ -150,13 +152,28 @@ const ImageCarousel: FunctionComponent<ImageCarouselProps> = ({ data }) => {
...
+      >
         <Box sx={modalStyle}>
-          <img
-            src={currentImage?.imageSrc}
-            alt={currentImage?.altText}
-            style={{ maxWidth: '80vw', maxHeight: '80vh' }}
-          />
+          <Box sx={{ position: 'relative' }}>
+            <IconButton
+              sx={{ color: 'rgba(255, 255, 255, 0.54)', position: 'absolute', top: 10, right: 10 }}
```
mcartaud, line 164: Idem color from theme

opencti-platform/opencti-front/src/private/components/data/playbooks/types/nodes/NodePlaceholder.tsx
```diff
@@ -26,11 +27,18 @@ const useStyles = makeStyles<Theme>((theme) => ({
     textAlign: 'center',
     cursor: 'pointer',
     '&:hover': {
+      backgroundColor:
+        theme.palette.mode === 'dark'
```
mcartaud, line 31: Is it possible to use a color from the theme?
It will not work well with custom themes

opencti-platform/opencti-front/src/components/filters/FilterValues.tsx
```diff
@@ -54,33 +73,32 @@ const FilterValues: FunctionComponent<FilterValuesProps> = ({
...
-      };
+  const onCLick = deactivatePopoverMenu ? () => { } : onClickLabel;
 
   // special case for nil/not_nil
   if (isOperatorNil) {
     return (
       <>
-        <strong
-          style={labelStyle}
+        <button
```
mcartaud: I am not sure of the origin of the change, but when selecting several elements in a filter, the design change:
- original 
Image

- with this PR
Image

opencti-platform/opencti-front/src/components/FilterIconButtonContainer.tsx
```diff
@@ -24,8 +24,8 @@ import { Stack } from '@mui/material';
 
 export type FilterIconButtonVariant
   = undefined // default variant (variant is undefined), for filters applied in datatables or widgets for instance
-    | 'small' // small variant, for filters in a datatable line for instance
-    | 'tag'; // for filters with a style similar as the Tag component, in an entity Overview for instance
+  | 'small' // small variant, for filters in a datatable line for instance
```
mcartaud: The file has been incorrectly formatted

opencti-platform/opencti-front/src/private/components/data/playbooks/types/nodes/NodePlaceholder.tsx
scarletmerlin123: It doesn't look like we have an existing that is an exact match for the transparency so would need to make a new property in the theme. It looks like palette.border.paper is close but will be a little more transparent.

This does bring up a point that maybe we should update the custom theme form to allow users to specify the mode. Currently the mode is set based on the name or falls back to dark mode which could cause issues if users don't follow the naming conventions.  
https://github.com/OpenCTI-Platform/opencti/blob/594dab16eca02020122f5c36c02e00bcfa7a007d/opencti-platform/opencti-front/src/components/AppThemeProvider.tsx#L122

opencti-platform/opencti-front/src/components/filters/FilterValues.tsx
scarletmerlin123: I updated to remove the use of clsx and makestyles in favor of updating it to use the sx prop on a Button component instead.

opencti-platform/opencti-front/src/components/markdownDisplay/MarkdownDisplay.tsx
scarletmerlin123: This is the same behavior as before, though the link component exposes the issue. Anywhere that we don't want links 'removeLinks' must be passed to MarkdownDisplay.  I added removeLinks to components that seem like they shouldn't have links. Are there any others that should be updated?

opencti-platform/opencti-front/src/components/markdownDisplay/MarkdownImagePreviewModal.tsx
scarletmerlin123: lightGray wouldn't have the transparency. This was taken from the existing button to open the image preview.

opencti-platform/opencti-front/src/components/markdownDisplay/MarkdownDisplay.tsx
mcartaud: On a second though, I don't think we are going into the correct direction here : you transform every usage of markdown into a link and let one opt-out this behavior, but in most cases it is used in a non interactive way or when interactive it is more as a button (to open a drawer, like in `Observable > History` entries  for instance) and clicking on them doesn't redirect the user. 

In addition, the current behavior seem fine : if a link in present in the markdown, it is properly rendered as a `<a />` element (at least in the places I checked), what problem are you trying to solve here?

opencti-platform/opencti-front/src/private/components/data/playbooks/types/nodes/NodePlaceholder.tsx
mcartaud: Indeed. I don't see how you can improve the code you did without making changes above the scope of this PR. GTM for now

opencti-platform/opencti-front/src/components/markdownDisplay/MarkdownDisplay.tsx
scarletmerlin123: The issue was it was being rendered as a div with a click handler instead of a interactive element. I may be incorrectly interpreting what MarkdownDisplay is doing. The intention here was to move from a div to a button link because this block seemed like it was specific to linking to external references because it looks like it is supposed to bring up the external link popover. However, like you pointed out all of the rendered markdown is being treated as a link. Seems like the best approach would be to make it an unstyled button element so it will be properly handled while using an interactive element that will be anouced to a screen reader.

mcartaud: Visually it is better, but this solution (whereas Link or Button usage) seem not very good for screen reader users, they will be overwhelmed by incorrect informations about falsy interactive elements and even on markdown containing real interactive element they will be informed they can click on the button but it does nothing, they'll have to go into the element, focus the link and then they'll be able to interact with it, like in current version. See attached videos (pardon my clumsiness with voice over navigation, I am not very used to it), the little clicks are when I try to interact with the button

This is the current state

https://github.com/user-attachments/assets/29d6fe18-936f-4861-9210-ec5d2ee34339

And with your changes

https://github.com/user-attachments/assets/cb4fe85d-f94a-4fa6-8e4c-16dd3cc4d7fa

scarletmerlin123: Thinking about it some more, it seems like the div itself wasn't really the issue because that isn't the interactive element.  The link within is the interactive element and the div is there to intercept the click. Probably fine to revert back to original implementation and add aria-hidden="true" so only the content within is picked up by the screen reader.

opencti-platform/opencti-front/src/components/markdownDisplay/MarkdownDisplay.tsx
```diff
@@ -270,14 +267,12 @@ const MarkdownDisplay: FunctionComponent<MarkdownWithRedirectionWarningProps> =
   } else {
     markdownDisplayContent = (
       <>
-        <Button
-          aria-label={t_i18n('Markdown')}
-          disableRipple
-          sx={{ all: 'unset' }}
+        <div
+          aria-hidden="true"
```
mcartaud: I don't think this is a good solution either because now the content is not selectable anymore by screen readers, making links that can be embedded in the markdown effectively unreachable for those users.
It also make tests to fail

mcartaud: Except this point the PR looks fine to me, if you want you can remove this change and do another PR for this issue

scarletmerlin123: Sounds good, I'll make a new issue for resolving accessibility concerns with Markdown display. Let me rebase and squash the commits.

opencti-platform/opencti-front/src/components/FilterIconButtonContainer.tsx
```diff
@@ -247,10 +247,10 @@ const FilterIconButtonContainer: FunctionComponent<
         );
         const isNotLastFilter = index < displayedFilters.length - 1;
 
-        const chipVariant = currentFilter.values.length === 0 && !NO_VALUES_FILTER_OPERATORS.includes(filterOperator ?? 'eq')
-          ? 'outlined'
-          : 'filled';
-        // darken the bg color when filled (quickfix for 'warning' and 'success' chipColor unreadable with regardingOf filter)
+          const chipVariant = currentFilter.values.length === 0 && !NO_VALUES_FILTER_OPERATORS.includes(filterOperator ?? 'eq')
```
mcartaud: small indentation issue here

scarletmerlin123: resolved

opencti-platform/opencti-front/src/components/dataGrid/components/DataTableHeader.tsx
```diff
@@ -107,13 +110,14 @@ const DataTableHeader: FunctionComponent<DataTableHeaderProps> = ({
       className={classes.headerContainer}
       style={{ width: cellWidth }}
     >
-      <div className={classes.label} onClick={throttleSortColumn}>
-        <Tooltip title={t_i18n(column.label)}>
-          <span>{t_i18n(column.label)}</span>
-        </Tooltip>
-        {sortBy && column.isSortable && (orderAsc ? <ArrowDropUp /> : <ArrowDropDown />)}
+      <div style={{ overflow: 'hidden', flex: 1 }} role="columnheader" aria-sort={sortBy ? (orderAsc ? 'ascending' : 'descending') : 'none'}>
```
mcartaud, line 114: One small issue I didn't noticed earlier : previously the whole header was clickable to add a sort or change its direction, now only the label is clickable

scarletmerlin123: updated so the button fills the available area

mcartaud: I am afraid it is not working, it is still only the label that is clickable

mcartaud: Change state : 

https://github.com/user-attachments/assets/0360d8a8-833d-4c61-bdf1-8ae6c94e041d

Current state

https://github.com/user-attachments/assets/08e72c5c-ae4f-4fde-a722-11d2c9f16d81

scarletmerlin123: Checking now and it seems to be working for me. Confirmed that the button element is the full width and has the width:100% style.  Wondering if something is cached in your browser.

Image

mcartaud: You're right, my mistake

mcartaud:
I am sorry, I noticed something else : an inconsistent behavior with the sub menu having a border but only on first page load, after clicking on one menu element it is not shown anymore (blank screen is me refreshing the page)

https://github.com/user-attachments/assets/9915d980-828b-431f-85e7-e2a17ff74ee4

scarletmerlin123:
This seems odd, the mouse actions are triggering :focus-visible when they shouldn't. I checked and the same issue exists in the #17581 PR when the focus changes are added, was hoping maybe the restructure of the nav wouldn't have this issue. It functions as intended using the keyboard, but incorrectly applies :focus-visible only on initial page load for mouse events. I'll keep looking to see if I can find the cause.

scarletmerlin123:
Resolved the issue. there was a typo where it was assigning the focus-visible style to the psudo class instead of as a class to be used by mui buttons that have the focusVisibleClassName prop.

SarahBocognano:
Capture d’écran 2026-08-20 à 14 57 25
Capture d’écran 2026-08-20 à 14 58 45
Capture d’écran 2026-08-20 à 14 59 07
Capture d’écran 2026-08-20 à 14 59 40
Capture d’écran 2026-08-20 à 15 01 09
Capture d’écran 2026-08-20 à 15 01 49
Capture d’écran 2026-08-20 à 15 01 19
Capture d’écran 2026-08-20 à 15 07 49
Capture d’écran 2026-08-20 à 15 06 16
Capture d’écran 2026-08-20 à 15 05 51
Capture d’écran 2026-08-20 à 15 05 25
Capture d’écran 2026-08-20 à 15 03 24
Capture d’écran 2026-08-20 à 15 02 31


Hello ! I'm taking a bit your review this week as my collegue is in holiday, I'm sorry I'm not really on tune with what's going on, but I've noticed several behavior in the platform on your branch, so I've put you all the capture here where I found this weird blue/white lines a little bit everywhere.
I don't know if you wanted that ? But as it's not uniformized everywhere, I though that maybe it was not.

FWI: This is all by navigating with keyboard (tab key)

scarletmerlin123:
@SarahBocognano Hi, That blue outline is the pseudo class that is getting added. it is added when a user uses keyboard navigation to navigate the app. The blue ring should now be showing on elements that previously didn't display a ring when using keyboard nav.

Edit: The blue ring is provided by the browser. I just checked and we don't currently have any global styles set for focus-visible so it is using the built-in style where not defined at the component level.

SarahBocognano (approved):

PR #17408, opened 2026-07-29
feat(widgets): Saved filters in widgets: refinements & introduction in custom views (#15689)

### LINKED ISSUE #15689: feat(dashboards): reusable saved filters in dashboards

Archidoit:
### Proposed changes
- When a saved filter content has changed, and we click on 'refresh' in a dashboard, the widget with the saved filters should be reloaded by taking into account the new filters content
- Frontend test: reloads with the freshly resolved data selection of a widget when refresh token changes
- Add scope in saved filters automplete tooltips

- Disable saved filters whose scope is not compatible with the widget perspective

image

- Keeps custom filters in local state when switching to saved mode and restores them when returning to custom mode

- Disable the 'saved filters' button in the Audits perspective
image

- Add the possibility to select saved filters in widgets in custom views

- Handle export of custom views with widgets with saved filters

Under the DASHBOARD_SAVED_FILTERS feature flag

### Related issues
fixes #15689


opencti-platform/opencti-front/src/components/dashboard/useDashboardViz.ts
```diff
@@ -40,11 +40,27 @@ const useDashboardViz = <TQuery extends OperationType>({
...
+  const prevRefreshTokenRef = useRef(refreshToken);
+
   // Stabilize the dataSelection dependency to avoid re-triggering the effect
   // on every render when the parent passes a new array reference with the same content.
   const dataSelectionSignature = useMemo(() => JSON.stringify(dataSelection), [dataSelection]);
 
-  useEffect(() => {
+  // Resolves the raw dataSelection into a query-ready form: it hydrates saved filters,
+  // injects the host entity context, and flags edge cases (missing host entity, preview
+  // mode, missing saved filters) via state setters. Resolution is asynchronous, so it
```
mcartaud: I am not sure those comments brings value : 
- if the behavior of `resolveDataSelection` should have some explanation, they probably should be as JSDoc of this method
- the rest on the return method to cancel is standard useEffect behavior

opencti-platform/opencti-front/src/components/dashboard/useDashboardViz.ts
```diff
@@ -57,12 +73,15 @@ const useDashboardViz = <TQuery extends OperationType>({
...
+        onResolved?.(result);
       }
     });
     return () => {
       cancelled = true;
     };
-  }, [filterKeysSchema, dataSelectionSignature, perspective, host]);
+  };
+
+  useEffect(() => handleResolveDataSelection(), [filterKeysSchema, dataSelectionSignature, perspective, host]);
```
mcartaud: This is pretty misleading, it is hard to understand why those elements are in the dependency when they are not passed anywhere. It is probably better to pass them as parameters of the `handleResolveDataSelection` to make it explicit, or to wrap the method in a useCallback with its own dependencies and just remove them from here

opencti-platform/opencti-front/src/components/dashboard/useDashboardViz.ts
```diff
@@ -40,11 +40,27 @@ const useDashboardViz = <TQuery extends OperationType>({
   const [isPreviewMode, setIsPreviewMode] = useState(false);
   const [isMissingSavedFilters, setIsMissingSavedFilters] = useState(false);
 
+  // refreshToken is an integer provided via context by DashboardContent and incremented
+  // by CustomDashboard on manual or auto refresh. When it changes, we force-reload
+  // regardless of whether query variables changed, so fresh data is always fetched.
```
mcartaud, line 45: I am not sure those comments helps understand what's going one, but if usage change someday, we'll have to maintain documentation

opencti-platform/opencti-front/src/components/dashboard/useDashboardViz.ts
```diff
@@ -126,34 +145,37 @@ const useDashboardViz = <TQuery extends OperationType>({
   }, [isPending, setQueryPending]);
 
   // Used by dashboard token refresh to rebuild variables from latest inputs
-  // before forcing the load.
-  const forceReloadWithFreshVariables = useCallback(() => {
-    if (!buildQueryVariables || !config || resolvedDataSelection.length === 0) {
+  // before forcing the load. Accepts an explicit resolved selection so a refresh can
```
mcartaud: idem comments. To be more explicit on comments, I think they are important as JSDoc when it explicit what's expected by a function or what is a type (or type field, or enum value, etc).
For 'simple' `//` comments, I think we should only use them when we do some strange thing for a reason not directly visible, for instance if we have to be defensive against some side effect from a totally different place, or things like that

opencti-platform/opencti-front/src/components/dashboard/useDashboardViz.ts
```diff
@@ -126,34 +145,37 @@ const useDashboardViz = <TQuery extends OperationType>({
...
     prevRefreshTokenRef.current = refreshToken;
 
     if (isMissingHostEntity || isMissingSavedFilters) {
-      return;
+      return undefined;
     }
 
-    forceReloadWithFreshVariables();
+    // Re-resolve on refresh, then force the reload with the freshly resolved data
+    // selection (and respecting the fresh edge-case flags) so the refreshed query uses
```
mcartaud: idem comments

opencti-platform/opencti-front/src/components/saved_filters/useBuildSavedFiltersOptions.ts
```diff
@@ -1,19 +1,48 @@
...
+ * saved-filters data returned by the API.
+ *
+ * The returned options are sorted with the user's own filters ("My filters")
+ * first, then those "Shared with me", each group ordered alphabetically by label.
+ *
+ * @param data  Raw saved-filter nodes from the `savedFilters` query.
+ * @param scope Optional widget perspective used to disable non-matching filters in widgets.
+ * @returns Sorted, display-ready autocomplete saved filters options.
+ */
+const useBuildSavedFiltersOptions = (
```
mcartaud, line 15: 👍

opencti-platform/opencti-front/src/private/components/settings/sub_types/custom_views/useCustomViewDashboardEdit.ts
```diff
@@ -101,15 +101,20 @@ const useCustomViewDashboardEdit = ({ customView }: {
...
-    const data = await fetchQuery(customViewExportWidgetQuery, { id, widgetId: widget.id })
-      .toPromise();
-    const result = data as useCustomViewDashboardEdit_WidgetExportQuery$data;
-    const exportString = result.customView?.toWidgetExport;
-    if (!exportString) {
-      MESSAGING$.notifyError(t_i18n('Failed to export widget'));
+    try {
+      const data = await fetchQuery(customViewExportWidgetQuery, { id, widgetId: widget.id })
+        .toPromise();
+      const result = data as useCustomViewDashboardEdit_WidgetExportQuery$data;
```
mcartaud: Is it possible to type `fetchQuery` rather than casting here?

opencti-platform/opencti-front/src/private/components/widgets/WidgetFilters.tsx
```diff
@@ -42,15 +43,22 @@ const WidgetFilters: FunctionComponent<WidgetFiltersProps> = ({ perspective, typ
   useEffect(() => {
     setDataSelection({
       ...dataSelection,
-      filters,
-      dynamicTo: filtersDynamicTo,
-      dynamicFrom: filtersDynamicFrom,
+      filters: useSavedFilter ? undefined : filters,
```
mcartaud: I find it a bit misleading to name the value of a state hook with `use...`, it is a convention for hook naming, but it is not hooks, they are boolean so they probably should have a name in `is...` or `has....`.
With this convention, we always have to think a little to understand why this hook is just used in a of or a ternary, is it to check if it undefined? etc.
In addition, react tooling make some checks for hooks, like they cannot be defined after dynamic parts of the code, what happens for those `use....` constants?

opencti-platform/opencti-front/src/private/components/workspaces/dashboards/CustomDashboard.tsx
```diff
@@ -94,24 +94,30 @@ const CustomDashboard = ({ data, noToolbar = false }: CustomDashboardProps) => {
...
     }
   };
 
   const onExport = async (id: string) => {
-    const data = await fetchQuery(dashboardExportQuery, { id })
-      .toPromise() as CustomDashboardExportQuery$data;
-    if (!data.workspace) {
+    try {
+      const data = await fetchQuery(dashboardExportQuery, { id })
+        .toPromise() as CustomDashboardExportQuery$data;
```
mcartaud: idem cast

opencti-platform/opencti-front/src/private/components/workspaces/useWorkspaceHandleExportJson.tsx
```diff
@@ -0,0 +1,46 @@
...
+`;
+
+const useWorkspaceHandleExportJson = () => {
+  const { t_i18n } = useFormatter();
+
+  const workspaceHandleExportJson = (workspace: workspaceToExport) => {
+    fetchQuery(WorkspaceExportHandlerQuery, { id: workspace.id })
+      .toPromise()
+      .then((data) => {
+        const result = data as workspaceExportHandlerQuery$data | null | undefined;
```
mcartaud: idem cast

opencti-platform/opencti-graphql/src/modules/dashboard/dashboard-utils.ts
```diff
@@ -150,8 +150,9 @@ export const importDashboardWidgetConfiguration = async (
 };
 
 // region workspace ids converter_2_1
-// Export => Dashboard filter ids must be converted to standard id
-// Import => Dashboards filter ids must be converted back to internal id
+// Export: dashboard widget filter ids are converted to STIX ids,
+// and saved filter references are resolved to inline filter content.
```
mcartaud: This would probably be more useful as JSDoc (`/** */`) comments because tooling will be able to show them on hoover or during completion

opencti-platform/opencti-graphql/src/modules/dashboard/dashboard-utils.ts
```diff
@@ -162,7 +163,10 @@ export const convertDashboardManifestIds = async (
   // Regeneration for dashboards
   if (parsedManifest && isNotEmptyField(parsedManifest.widgets)) {
     const { widgets } = parsedManifest;
-    const widgetDefinitions = Object.values(widgets);
+    const widgetDefinitions: Widget[] = Object.values(widgets);
+    if (from === 'internal') { // for exports: replace saved filters ids with the associated filters content
+      await Promise.all(widgetDefinitions.map((widgetDefinition) => resolveSavedFiltersInDataSelection(context, user, widgetDefinition)));
```
mcartaud, line 171: Just a warning on those methods (`resolveSavedFiltersInDataSelection`), I think they are usually bad because they mutate their parameters and it is not good practice, it makes the flow of data in the app hard to follow and all this code is not idempotent.
I would be in favor of trying to replace them with methods returning values rather than mutating their params (and more globally to try to be the most immutable as possible)
I don't think you should change it now since it is not directly linked to your change but it is more a "style" remark and to know if you agree with this

mcartaud (commented):
Only minor "style" remarks, works well functionally :)

mcartaud:
Only minor thing I noticed, but not sure it is to fix, just sharing : when you have a shared filter selected in some view (say observables) and this shared filter is updated from elsewhere, after I reload or navigate away and back, the shared filter is not updated, and I am asked to save it (with the yellow circle on the save icon) to its previous value.
It makes me wonder about the implementation, perhaps not storing the filters themselves in the local storage but the id of the shared filter would be better

alice-debra:
Product validation done 🟢

opencti-platform/opencti-front/src/components/dashboard/useDashboardViz.ts
Archidoit: it's not me who added this comment (i moved the code in a better place), so i will keep this comment :)

opencti-platform/opencti-front/src/components/dashboard/useDashboardViz.ts
Archidoit: indeed! i will turn it into a jsdoc comment :)

opencti-platform/opencti-front/src/components/dashboard/useDashboardViz.ts
Archidoit: I will turn this comment into a jsdoc comment :)

opencti-platform/opencti-front/src/components/dashboard/useDashboardViz.ts
Archidoit: done :)

opencti-platform/opencti-front/src/private/components/settings/sub_types/custom_views/useCustomViewDashboardEdit.ts
Archidoit: done :) better this way indeed

opencti-platform/opencti-front/src/private/components/workspaces/dashboards/CustomDashboard.tsx
Archidoit: done

opencti-platform/opencti-front/src/private/components/workspaces/useWorkspaceHandleExportJson.tsx
Archidoit: done

opencti-platform/opencti-graphql/src/modules/dashboard/dashboard-utils.ts
Archidoit: done :)

opencti-platform/opencti-graphql/src/modules/dashboard/dashboard-utils.ts
Archidoit: Indeed, I agree with you. There are several parts in the code mutating objects in the backend... Maybe i won't change this now, especially knowing this function can have huge impacts. Maybe better in a separated PR :)

opencti-platform/opencti-front/src/private/components/widgets/WidgetFilters.tsx
Archidoit: indeed, i was thinking about the verb 'use' without thinking about hooks ! I will rename useSavedFilter to isSavedFiltersMode (same for the other 'use') :)

opencti-platform/opencti-front/src/components/dashboard/useDashboardViz.ts
Archidoit: I'm gonna add a small comment before this line to explain

opencti-platform/opencti-front/src/components/dashboard/useDashboardViz.ts
```diff
@@ -83,6 +83,8 @@ const useDashboardViz = <TQuery extends OperationType>({
     };
   };
 
+  // Re-resolve selection inputs when schema, selection content, perspective, or host changes
+  // Because those changes make the result change
```
mcartaud, line 87: why not using `useCallback` to define the function? It does a function that will be recomputed if any of those dependencies change and you could remove them from the useEffect, making it more readable : useEffect is re-executed if handleResolveDataSelection is redefined, and handleResolveDataSelection is redefined if filterKeysSchema, dataSelectionSignature, perspective, host change.
Or am I missing something? It is possible those rules are tricky :)

Archidoit: done !

opencti-platform/opencti-front/src/components/dashboard/useDashboardViz.ts
```diff
@@ -81,11 +81,11 @@ const useDashboardViz = <TQuery extends OperationType>({
     return () => {
       cancelled = true;
     };
-  };
+  }, [filterKeysSchema, dataSelectionSignature, perspective, host]);
 
   // Re-resolve selection inputs when schema, selection content, perspective, or host changes
   // Because those changes make the result change
-  useEffect(() => handleResolveDataSelection(), [filterKeysSchema, dataSelectionSignature, perspective, host]);
+  useEffect(() => handleResolveDataSelection(), [handleResolveDataSelection]);
```
mcartaud: You should probably directly do `useEffect(handleResolveDataSelection, [handleResolveDataSelection]);` to avoid creating e new callback each time

mcartaud: `handleResolveDataSelection` dependency is probably missing line 186. On a side note, eslint has a plugin to lint react hook dependencies, we should probably use it because it is really easy to miss some of them and i can cause very hard to debug issue. I'll look into it

mcartaud (approved):

opencti-platform/opencti-front/src/components/dashboard/useDashboardViz.ts
Archidoit: done

Archidoit: done

PR #17415, opened 2026-07-30
fix(merge): avoid quadratic scan when filtering existing target relations during merge (#17413)

### LINKED ISSUE #17413: fix(scope): Merge hangs on entities with large relationship counts due to O(n×m) scan in filterTargetByExisting
## Description

`filterTargetByExisting` (used during entity merges) scans the full target relation list for every source relation via `R.find` + array `.includes`, instead of a hash lookup. This is O(n×m) instead of O(n+m). For entities with a large relationship count, the merge effectively hangs, holds its Redis lock indefinitely (auto-extended every ~50s), blocks playbooks/connectors on that entity, and can OOM the pod from the unbatched relation arrays held in memory.

## Environment

1. OS: on-prem k8s (customer-reported), reproducible on any OS
2. Version: 7.260715.0 (customer), present on `master` prior to fix
3. Other environment details: triggered by Data Sanity Manager's case-insensitive attack pattern dedup merge

## Reproducible steps

1. Create a target entity with a large number of relationships (e.g. 78k) and a duplicate source entity with few/no overlapping relationships
2. Trigger a merge of source into target (`attackPatternEdit` / Data Sanity Manager merge)
3. Merge takes tens of seconds to hours depending on relation count; entity stays locked, all writes to it fail with `LOCK_ERROR`

## Expected output

Merge completes in near-linear time relative to relation count (e.g. ~78k relations in well under a second).

## Actual output

Merge takes ~29s for 78k relations in isolated benchmark (worse in production with ES round-trips and concurrent retries piling up); observed as indefinite hang/deadlock in production, freezing all ingestion/playbooks touching the locked entity.

## Additional information

Root cause confirmed via benchmark comparing old vs new algorithm on identical synthetic data: 78k relations, old ~29,000ms, new (Map/Set indexed) ~13ms. Fix + benchmark test in PR #17406.

gabrielebonadiman:
### Proposed changes

* Replace O(n×m) linear scan (+O(n²) cache check) in `filterTargetByExisting` with an O(n+m) nested Map/Set index, fixing merge hangs on entities with large relationship counts
* Index is nested by `entity_type` → `internal_id` (not a concatenated string key) to avoid false-positive matches when either field contains a hyphen

### Related issues
* closes #17413

### How to test this PR

Run the unit suite:

npx vitest run --config vitest.config.ci-unit.ts tests/01-unit/database/merge-filterTargetByExisting-*

Or manually: merge a duplicate entity with tens of thousands of relations into a target and confirm the operation completes instead of hanging (previously observed on Attack-Pattern T1027, ~78k relations).

### Checklist

### Further comments

Root cause: `filterTargetByExisting` scanned all targets per source (`R.find`) plus a linear `.includes` cache check — quadratic, and the reason merges on high-relationship entities (77,964 in the reported case) never completed, holding the entity lock indefinitely via auto-extension.

Benchmarks (synthetic + real ES/Redis + real-shaped data matching the incident):
- 78k relations, isolated function: ~29s → ~13-25ms
- Real-code git-toggle (same commit, pre/post fix): confirms quadratic vs. near-linear growth
- Real ES/Redis, 5,000 seeded relations: full `mergeEntities()` in ~1.2s
- Incident-shaped data (target 77,964 distinct relations, duplicate source 2,000, 25% overlap): 48x speedup, identical output vs. frozen pre-fix reference
- Both-sides-large case (40k/40k, 30% overlap): 443x speedup

One known limitation: if relations degenerate into one entity holding tens of thousands of literal duplicate relations to a single other entity (not the reported shape), the per-bucket scan still dominates — still faster than before, but not full O(n+m). Not applicable to the reported incident, noting it for completeness.


JeremyCloarec (approved):
good job 👍

PR #17416, opened 2026-07-30
fix(client-python): accept scope as string or list in OpenCTIConnector (#17414)

### LINKED ISSUE #17414: fix(client-python): OpenCTIConnector constructor should accept scope as a list, not only as a comma-separated string
### Description

`OpenCTIConnector.__init__` (in `client-python/pycti/connector/opencti_connector.py`) only accepts `scope` as a comma-separated string, since it unconditionally calls `scope.split(",")`. This part assume that we send a string but convert into an array if the scope doesn't exist.

Some connector configurations already normalize the `CONNECTOR_SCOPE` value into a Python list before it reaches `OpenCTIConnector`. Passing that list crashes the constructor.

The fix should widen the constructor to accept both the legacy string form and an already-normalized list/tuple, while remaining fully backwards compatible.

### Environment

1. Python 3.12
2. OpenCTI v7.260729.0

### Reproducible steps

- Run OpenCTI
- Run VirusTotal connector

When the config form resolves scope to a list and passes it down into the constructor, it raises:

```sh
AttributeError: 'list' object has no attribute 'split'
```

## Expected output

- Make the constructor accept both forms and work without error

## Suggestion

```python
if not scope:
    self.scope = []
elif isinstance(scope, str):
    self.scope = scope.split(",")
else:
    # already a list/tuple (e.g. from the ListFromString Pydantic model)
    self.scope = list(scope)
```

helene-nguyen:
### Proposed changes

This PR widens the constructor to accept either form:
- a comma-separated `str` (legacy behavior, unchanged)
- an already-normalized `list`/`tuple`

* `client-python/pycti/connector/opencti_connector.py`: updated the `scope` handling in `OpenCTIConnector.__init__` to branch on type instead of assuming a string.
* Updated the  scope  docstring/type hints to document that  str ,  list , or  tuple  are now accepted.
* Updated tests

**Backwards compatibility**

• String input behaves exactly as before ( split(",") ).
• Empty/ None  input behaves exactly as before ( [] ).
• Only the list/tuple branch is new; no existing call sites are affected.

### Related issues

* closes #17414

### Checklist

### Further comments

Additional changes

- Add in the TOML file `--import-mode=importlib` gives every test file a unique module identity (keyed by file path, not by constructing a dotted package name), so the two same-named packages no longer collide.

### DIFF
--- a/client-python/pycti/connector/opencti_connector.py
+++ b/client-python/pycti/connector/opencti_connector.py
@@ -48,7 +48,9 @@ class OpenCTIConnector:
     :type connector_name: str
     :param connector_type: Type of connector (see :class:`ConnectorType`)
     :type connector_type: str
-    :param scope: Connector scope as a comma-separated string (e.g., "Report,Indicator")
+    :param scope: Connector scope, either as a comma-separated string
+    (e.g. ``"Artifact,Url"``) or as a list of strings
+    (e.g. ``["Artifact", "Url"]``. Both forms are normalized to a list.
     :type scope: str or list[str]
     :param auto: Whether the connector runs automatically on matching entities
     :type auto: bool
@@ -104,7 +106,7 @@ class OpenCTIConnector:
         :param scope: Connector scope, either as a comma-separated string
         (e.g. ``"Artifact,Url"``) or as a list of strings
         (e.g. ``["Artifact", "Url"]``. Both forms are normalized to a list.
-        :type scope: str
+        :type scope: str or list[str]
         :param auto: Whether the connector runs automatically
         :type auto: bool
         :param only_contextual: Whether to process only contextual data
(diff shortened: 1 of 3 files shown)


client-python/pycti/connector/opencti_connector.py
```diff
@@ -120,10 +125,13 @@ def __init__(
         self.id = connector_id
         self.name = connector_name
         self.type = ConnectorType(connector_type)
-        if scope:
+        if not scope:
+            self.scope = []
+        elif isinstance(scope, str):
```
mcartaud: Nit: we could add a trim on the striped value, because it is easy to make the mistake to add a space when writing a string

mcartaud (approved):

client-python/pycti/connector/opencti_connector.py
helene-nguyen: Thanks @mcartaud, I've added your suggestion for both cases :)

mcartaud (approved):

PR #17417, opened 2026-07-30
refactor(health): expose health and storage metrics through the prometheus subsystem (#17770)

### LINKED ISSUE #17770: feat(health): expose health and storage metrics through the prometheus subsystem
## Problem to solve
The `/health` endpoint probes ElasticSearch, S3/MinIO, RabbitMQ and Redis synchronously on every request, and operators have no visibility into storage usage or queue capacity from the monitoring stack.

## Proposed solution
Move health data ownership to the telemetry subsystem:
- Add a background monitor refreshing dependency connectivity on a periodic timer
- Collect storage and queue usage metrics on a separate, slower timer
- Export everything as Prometheus gauges on the existing metrics endpoint:
    - `opencti_dependency_up` (attribute `dependency`)
    - `opencti_elasticsearch_used_size_bytes`
    - `opencti_storage_used_size_bytes`
    - `opencti_queue_consumers` (attribute `connector_type`)
- Serve `/health` from that collected state, plus an optional `details=true` mode returning `es_used_size`, `s3_used_size` and `queue_consumers`

## Queue consumers
The platform reports the number of active listeners per queue type and does not aggregate them into an ingestion capacity value. Computing ingestion units is left to the requester, which owns that capacity model.

## Expected behavior
- `/health` performs no request to any dependency when queried
- `/health` returns `503` while no state has been collected yet, and `503` listing failing dependencies otherwise
- Detailed metrics are returned only when `details=true`, and never when the endpoint is `public`
- A metric that cannot be collected is reported as `null` and not exported, instead of being reported as `0`

## Implementation notes
- Refresh intervals configurable through `app:health_monitoring:dependency_check_interval` and `app:health_monitoring:usage_metrics_interval`
- Compute `s3_used_size` through paginated `ListObjectsV2`
- Group push queue consumers by their declared connector type, reporting queues without a type under `UNKNOWN`

## Related pull request
- https://github.com/OpenCTI-Platform/opencti/pull/17417

sbocahu:
## Summary
- Expose platform storage and queue metrics through the existing Prometheus metrics endpoint
- Move dependency connectivity checks into the same telemetry subsystem, refreshed by a periodic monitor
- Serve `/health` from the collected state instead of probing dependencies synchronously per request
- Add an optional `details=true` mode to `/health` returning `es_used_size`, `s3_used_size` and `queue_consumers`

## Architecture
A background monitor (`src/telemetry/platformHealthMetrics.ts`) owns all health state:
- Connectivity of ElasticSearch/OpenSearch, S3/MinIO, RabbitMQ and Redis is refreshed on a timer
- Storage and queue usage metrics are refreshed on a separate, slower timer
- Both are exported as Prometheus observable gauges and reused by `/health`

New Prometheus metrics:
- `opencti_dependency_up` (gauge, attribute `dependency`)
- `opencti_elasticsearch_used_size_bytes`
- `opencti_storage_used_size_bytes`
- `opencti_queue_consumers` (gauge, attribute `connector_type`)

## Queue consumers
The platform reports the raw number of active consumers per connector type and does **not** aggregate them into an ingestion capacity value. Deriving ingestion units is left to the requester, which owns that capacity model. Queues without a declared connector type are reported under `UNKNOWN`.

## Behavior
- `/health` no longer performs any request to ElasticSearch, S3, RabbitMQ or Redis
- `/health` returns `503` while the monitor has not collected an initial state, and `503` listing failing dependencies otherwise
- Detailed metrics are returned only when `details=true`, and only when the endpoint is not `public`
- A metric that cannot be collected is reported as `null` and is not exported to Prometheus, instead of being reported as `0`

## Configuration
| Key | Default | Description |
|:---|:---|:---|
| `app:health_monitoring:connectivity_interval` | `30000` | Interval between two connectivity checks |
| `app:health_monitoring:usage_metrics_interval` | `300000` | Interval between two usage metric collections, `0` disables them |

## Notes
- The previous per-process caches in `engine.ts` and `raw-file-storage.ts` were removed, since freshness is now owned by the monitor
- `s3_used_size` uses paginated `ListObjectsV2` (no native bucket-size API exists)

## Related issue

Closes #17770

### DIFF
--- a/opencti-platform/opencti-graphql/src/database/engine.ts
+++ b/opencti-platform/opencti-graphql/src/database/engine.ts
@@ -5110,6 +5110,44 @@ export const getStats = (indices = READ_PLATFORM_INDICES) => {
   return retryElOperations(statsOperation);
 };
 
+const ENGINE_USED_SIZE_CACHE_INTERVAL_MS = 300_000;
+let engineUsedSizeCacheInBytes = 0;
+let engineUsedSizeCacheDate = 0;
+let engineUsedSizeCachePromise: Promise<number> | null = null;
+
+const fetchEngineUsedSize = async (): Promise<number> => {
+  if (engine instanceof ElkClient) {
+    const engineIndicesStats = await engine.indices.stats({ index: '*', metric: ['store'], expand_wildcards: 'all' as any });
+    return Number(oebp(engineIndicesStats)?._all?.primaries?.store?.size_in_bytes ?? 0);
+  }
+  const engineIndicesStats = await engine.indices.stats({ index: '*', metric: 'store', expand_wildcards: 'all' as any });
+  return Number(oebp(engineIndicesStats)?._all?.primaries?.store?.size_in_bytes ?? 0);
+};
+
+export const getEngineUsedSize = async (): Promise<number> => {
+  const now = Date.now();
+  if (engineUsedSizeCacheDate > 0 && now - engineUsedSizeCacheDate < ENGINE_USED_SIZE_CACHE_INTERVAL_MS) {
+    return engineUsedSizeCacheInBytes;
+  }
+  if (engineUsedSizeCachePromise) {
+    return engineUsedSizeCachePromise;
+  }
+  engineUsedSizeCachePromise = retryElOperations(fetchEngineUsedSize)
+    .then((sizeInBytes) => {
+      engineUsedSizeCacheInBytes = sizeInBytes;
+      engineUsedSizeCacheDate = Date.now();
+      return sizeInBytes;
+    })
+    .catch((error) => {
+      logApp.warn('[SEARCH] Unable to fetch used size for health endpoint, returning last cached value', { cause: error });
+      return engineUsedSizeCacheInBytes;
+    })
+    .finally(() => {
+      engineUsedSizeCachePromise = null;
+    });
+  return engineUsedSizeCachePromise;
+};
+
 export const isEngineAlive = async () => {
   const context = executionContext('healthcheck');
   const options = { types: [ENTITY_TYPE_MIGRATION_STATUS], connectionFormat: false };
--- a/opencti-platform/opencti-graphql/src/database/raw-file-storage.ts
+++ b/opencti-platform/opencti-graphql/src/database/raw-file-storage.ts
@@ -233,3 +233,45 @@ export const rawListObjects = async (directory: string, recursive: boolean, cont
   }
   return s3Client.send(new s3.ListObjectsV2Command(requestParams));
 };
+
+const STORAGE_USED_SIZE_CACHE_INTERVAL_MS = 300_000;
+let storageUsedSizeCacheInBytes = 0;
+let storageUsedSizeCacheDate = 0;
+let storageUsedSizeCachePromise: Promise<number> | null = null;
+
+const fetchStorageUsedSize = async (): Promise<number> => {
+  let totalSize = 0;
+  let truncated = true;
+  let continuationToken: string | undefined;
+  while (truncated) {
+    const response = await rawListObjects('', true, continuationToken);
+    totalSize += (response.Contents ?? []).reduce((sum, object) => sum + (object.Size ?? 0), 0);
+    truncated = response.IsTruncated ?? false;
+    continuationToken = truncated ? response.NextContinuationToken : undefined;
+  }
+  return totalSize;
+};
+
+export const getStorageUsedSize = async (): Promise<number> => {
+  const now = Date.now();
+  if (storageUsedSizeCacheDate > 0 && now - storageUsedSizeCacheDate < STORAGE_USED_SIZE_CACHE_INTERVAL_MS) {
+    return storageUsedSizeCacheInBytes;
+  }
+  if (storageUsedSizeCachePromise) {
+    return storageUsedSizeCachePromise;
+  }
+  storageUsedSizeCachePromise = fetchStorageUsedSize()
+    .then((sizeInBytes) => {
+      storageUsedSizeCacheInBytes = sizeInBytes;
+      storageUsedSizeCacheDate = Date.now();
+      return sizeInBytes;
+    })
+    .catch((error) => {
+      logApp.warn('[FILE STORAGE] Unable to fetch used size for health endpoint, returning last cached value', { cause: error });
+      return storageUsedSizeCacheInBytes;
+    })
+    .finally(() => {
+      storageUsedSizeCachePromise = null;
+    });
+  return storageUsedSizeCachePromise;
+};
--- a/opencti-platform/opencti-graphql/src/http/httpPlatform.js
+++ b/opencti-platform/opencti-graphql/src/http/httpPlatform.js
@@ -15,7 +15,7 @@ import { basePath, DEV_MODE, ENABLED_UI, logApp, OPENCTI_SESSION, PLATFORM_VERSI
 import { sessionAuthenticateUser, userWithOrigin } from '../domain/user';
 import { checkIpWhitelistForRequest } from './ipWhitelistMiddleware';
 import { getXtmJwks } from '../domain/xtm-auth';
-import { downloadFile, getFileContent, isStorageAlive } from '../database/raw-file-storage';
+import { downloadFile, getFileContent, getStorageUsedSize, isStorageAlive } from '../database/raw-file-storage';
 import { loadFile } from '../database/file-storage';
 import { DEFAULT_INVALID_CONF_VALUE, executionContext, SYSTEM_USER } from '../utils/access';
 import { ENTITY_TYPE_SETTINGS } from '../schema/internalObject';
@@ -25,7 +25,7 @@ import { buildContextDataForFile, publishUserAction } from '../listener/UserActi
 import { internalLoadById } from '../database/middleware-loader';
 import { delUserContext, redisIsAlive } from '../database/redis';
 import { rabbitMQIsAlive } from '../database/rabbitmq';
-import { isEngineAlive } from '../database/engine';
+import { getEngineUsedSize, isEngineAlive } from '../database/engine';
 import createSseMiddleware from '../graphql/sseMiddleware';
 import initTaxiiApi from './httpTaxii';
 import initHttpRollingFeeds from './httpRollingFeed';
@@ -540,7 +540,15 @@ const createApp = async (app, schema) => {
           const rabbitMQAlive = healthCheckTimeout(rabbitMQIsAlive(), 'Timeout checking rabbitmq health');
           const redisAlive = healthCheckTimeout(redisIsAlive(), 'Timeout checking redis health');
           await Promise.all([engineAlive, storageAlive, rabbitMQAlive, redisAlive]);
-          res.status(200).send({ status: 'success' });
+          const withDetails = String(req.query?.details ?? '').toLowerCase() === 'true';
+          if (withDetails) {
+            const engineUsedSize = healthCheckTimeout(getEngineUsedSize(), 'Timeout checking elastic/opensearch used size').catch(() => null);
+            const storageUsedSize = healthCheckTimeout(getStorageUsedSize(), 'Timeout checking storage used size').catch(() => null);
+            const [esUsedSize, s3UsedSize] = await Promise.all([engineUsedSize, storageUsedSize]);
+            res.status(200).send({ status: 'success', es_used_size: esUsedSize, s3_used_size: s3UsedSize });
+          } else {
+            res.status(200).send({ status: 'success' });
+          }
         } else {
           res.status(401).send({ status: 'unauthorized' });
         }
--- a/opencti-platform/opencti-graphql/src/http/httpPlatform.js
+++ b/opencti-platform/opencti-graphql/src/http/httpPlatform.js
@@ -24,7 +24,7 @@ import { isEmptyField, isNotEmptyField } from '../database/utils';
 import { buildContextDataForFile, publishUserAction } from '../listener/UserActionListener';
 import { internalLoadById } from '../database/middleware-loader';
 import { delUserContext, redisIsAlive } from '../database/redis';
-import { rabbitMQIsAlive } from '../database/rabbitmq';
+import { getIngestionUnits, rabbitMQIsAlive } from '../database/rabbitmq';
 import { getEngineUsedSize, isEngineAlive } from '../database/engine';
 import createSseMiddleware from '../graphql/sseMiddleware';
 import initTaxiiApi from './httpTaxii';
@@ -544,8 +544,14 @@ const createApp = async (app, schema) => {
           if (withDetails) {
             const engineUsedSize = healthCheckTimeout(getEngineUsedSize(), 'Timeout checking elastic/opensearch used size').catch(() => null);
             const storageUsedSize = healthCheckTimeout(getStorageUsedSize(), 'Timeout checking storage used size').catch(() => null);
-            const [esUsedSize, s3UsedSize] = await Promise.all([engineUsedSize, storageUsedSize]);
-            res.status(200).send({ status: 'success', es_used_size: esUsedSize, s3_used_size: s3UsedSize });
+            const ingestionUnits = healthCheckTimeout(getIngestionUnits(executionContext('healthcheck'), SYSTEM_USER), 'Timeout checking ingestion units').catch(() => null);
+            const [esUsedSize, s3UsedSize, parallelIngestionUnits] = await Promise.all([engineUsedSize, storageUsedSize, ingestionUnits]);
+            res.status(200).send({
+              status: 'success',
+              es_used_size: esUsedSize,
+              s3_used_size: s3UsedSize,
+              ingestion_units: parallelIngestionUnits,
+            });
           } else {
             res.status(200).send({ status: 'success' });
           }
--- a/opencti-platform/opencti-graphql/src/database/rabbitmq.js
+++ b/opencti-platform/opencti-graphql/src/database/rabbitmq.js
@@ -11,6 +11,8 @@ import { getHttpClient } from '../utils/http-client';
 import { fullEntitiesList } from './middleware-loader';
 import { ENTITY_TYPE_BACKGROUND_TASK, ENTITY_TYPE_CONNECTOR, ENTITY_TYPE_SYNC } from '../schema/internalObject';
 import { ENTITY_TYPE_PLAYBOOK } from '../modules/playbook/playbook-types';
+import { CONNECTOR_INTERNAL_ENRICHMENT, CONNECTOR_INTERNAL_IMPORT_FILE } from '../schema/general';
+import { ConnectorType } from '../generated/graphql';
 import { s3ConnectionConfig } from './raw-file-storage';
 
 export const CONNECTOR_EXCHANGE = `${RABBIT_QUEUE_PREFIX}amqp.connector.exchange`;
@@ -509,6 +511,33 @@ export const getConnectorQueueSize = async (context, user, connectorId) => {
   }
   return targetQueues.length > 0 ? targetQueues.reduce((a, b) => (a.messages ?? 0) + (b.messages ?? 0)) : 0;
 };
+
+// Connector types whose active queue consumers represent parallel ingestion flows.
+// Uses the shared schema constants (and the ConnectorType enum for external import)
+// to avoid drifting from the canonical connector type definitions.
+const INGESTION_CONNECTOR_TYPES = new Set([
+  ConnectorType.ExternalImport, // external feeds ingestion
+  CONNECTOR_INTERNAL_IMPORT_FILE, // file import ingestion
+  CONNECTOR_INTERNAL_ENRICHMENT, // enrichment flows
+]);
+
+export const getIngestionUnits = async (context, user) => {
+  let stats = metricsCache.get('cached_metrics');
+  if (!stats) {
+    stats = await metrics(context, user);
+    metricsCache.set('cached_metrics', stats);
+  }
+  const ingestionPushQueues = stats.queues.filter((queue) => {
+    const queueName = queue?.name ?? '';
+    if (!queueName.startsWith(`${RABBIT_QUEUE_PREFIX}push_`)) {
+      return false;
+    }
+    const connectorType = queue?.arguments?.config?.type;
+    return INGESTION_CONNECTOR_TYPES.has(connectorType);
+  });
+  return ingestionPushQueues.reduce((sum, queue) => sum + (queue?.consumers ?? 0), 0);
+};
+
 export const getBestBackgroundConnectorId = async (context, user) => {
   let stats = metricsCache.get('cached_metrics');
   if (!stats) {


sbocahu:
Follow-up update with proper formatting (supersedes previous comment formatting glitches):

Addressed pending review comments in commit 8557652249.

Changes made:
- Updated `getEngineUsedSize` and `getStorageUsedSize` error handling to avoid masking cold-cache failures as `0`.
- Preserved stale cached values when available, while throttling retries after failures.
- Kept `/health?details=true` contract aligned with documented behavior (`null` on metric failures).
- Added unit tests for details gating (`public` vs private access key) and `null` detailed metrics on per-metric failures.

Note: I could not run the targeted test file locally in this environment because the `opencti-graphql` workspace is missing Yarn install state (`findPackageLocation` error).

sbocahu:
Follow-up fix pushed in commit 8dc73f7956:

- Fixed `tests/01-unit/http/httpPlatform-test.ts` route capture harness to only register string paths.
- This resolves the `TypeError: path.endsWith is not a function` failure in:
  - `httpPlatform: /health details behavior > should return null detailed metrics on per-metric failures when details=true`

The failure was in the test harness (fake `app.get`) rather than in `/health` implementation logic.

sbocahu:
Follow-up TypeScript fix pushed in commit 2052a042b6:

- Updated `nconf.get` test spies in `tests/01-unit/http/httpPlatform-test.ts` to use an optional key signature (`key?: string`).
- This resolves CI error `TS2345` about mock implementation type incompatibility.

opencti-platform/opencti-graphql/src/http/httpPlatform.js
```diff
@@ -112,6 +112,10 @@ export const decodeStoragePath = (fileParts = []) => fileParts
   })
   .join('/');
 
+export const shouldIncludeHealthDetails = (configAccessKey, detailsQueryParam) => {
```
aHenryJard, line 115: I don't think we should use the /health endpoint for technical metrics, is it possible to use the prometeus one /metrics instead ? Or else a new endpoint is also possible.

opencti-platform/opencti-graphql/src/database/engine.ts
```diff
@@ -5110,6 +5110,56 @@ export const getStats = (indices = READ_PLATFORM_INDICES) => {
   return retryElOperations(statsOperation);
 };
 
+const ENGINE_USED_SIZE_CACHE_INTERVAL_MS = 300_000;
+let engineUsedSizeCacheInBytes = 0;
+let engineUsedSizeCacheDate = 0;
+let engineUsedSizeFailureCacheDate = 0;
+let engineUsedSizeCachePromise: Promise<number> | null = null;
```
aHenryJard, line 5117: This cache will lead to unconsistency on the metrics, it's on a signle node so on cluster mode the number will be differents depending on the node. In addition it could be called on all nodes so the computation will be done N times instead of one time. It would be better to compute metrics on background with a manager and use Redis as cluster cache.

opencti-platform/opencti-graphql/src/database/raw-file-storage.ts
```diff
@@ -233,3 +233,57 @@ export const rawListObjects = async (directory: string, recursive: boolean, cont
   }
   return s3Client.send(new s3.ListObjectsV2Command(requestParams));
 };
+
+const STORAGE_USED_SIZE_CACHE_INTERVAL_MS = 300_000;
+let storageUsedSizeCacheInBytes = 0;
+let storageUsedSizeCacheDate = 0;
+let storageUsedSizeFailureCacheDate = 0;
+let storageUsedSizeCachePromise: Promise<number> | null = null;
```
aHenryJard, line 241: Same issue on local cache than the engine one.

opencti-platform/opencti-graphql/src/database/raw-file-storage.ts
```diff
@@ -233,3 +233,57 @@ export const rawListObjects = async (directory: string, recursive: boolean, cont
...
+  }
+  storageUsedSizeCachePromise = fetchStorageUsedSize()
+    .then((sizeInBytes) => {
+      storageUsedSizeCacheInBytes = sizeInBytes;
+      storageUsedSizeCacheDate = Date.now();
+      storageUsedSizeFailureCacheDate = 0;
+      return sizeInBytes;
+    })
+    .catch((error) => {
+      logApp.warn('[FILE STORAGE] Unable to fetch used size for health endpoint, returning stale cache or propagating failure', { cause: error });
```
aHenryJard, line 277: function in engine should be generic, so can be called by something else that the metric usage endpoint. Maybe just thow the error at this level and manage it at higher level like endpoint level.

sbocahu:
Reworked the feature following review feedback, pushed in commit 99cd26b86c.

Instead of caching metrics per process inside the `/health` endpoint, health data is now owned by the telemetry subsystem:

- Added `src/telemetry/platformHealthMetrics.ts`, a background monitor refreshing dependency connectivity and usage metrics on periodic timers.
- Exported the collected values as Prometheus observable gauges on the existing metrics endpoint: `opencti_dependency_up`, `opencti_elasticsearch_used_size_bytes`, `opencti_storage_used_size_bytes` and `opencti_ingestion_units`.
- Reworked `/health` to answer from that collected state, so it no longer issues synchronous requests to ElasticSearch, S3, RabbitMQ or Redis when queried.
- Removed the ad hoc size caches previously added to `engine.ts` and `raw-file-storage.ts`, since freshness is now owned by the monitor.
- Made both refresh intervals configurable through `app:health_monitoring`.

Validation:
- `tests/01-unit/http/httpPlatform-test.ts` and `tests/01-unit/telemetry/platformHealthMetrics-test.ts`: 25 tests passing.
- `yarn check-ts` and `yarn lint`: clean.

sbocahu:
Reworked ingestion units following feedback, pushed in commit d959974e27.

The platform no longer computes an aggregated ingestion units value. It now reports the raw number of active listeners per queue type, and the requester derives its own capacity model from them.

- Replaced `getIngestionUnits` with `getQueueConsumersByType`, which groups push queue consumers by their declared connector type instead of summing a hardcoded subset (`EXTERNAL_IMPORT` / `INTERNAL_IMPORT_FILE` / `INTERNAL_ENRICHMENT`).
- Replaced the `opencti_ingestion_units` gauge with `opencti_queue_consumers`, labelled by `connector_type`, so every type is observable individually.
- `/health?details=true` now returns `queue_consumers` (a map of connector type to consumer count) instead of `ingestion_units`.
- Queues without a declared connector type are reported under `UNKNOWN`.
- Dropped the now unused `ConnectorType` and connector type constant imports from `rabbitmq.js`.

Validation: 40 tests passing across `httpPlatform`, `platformHealthMetrics` and `rabbitmq` unit test files, including 6 new cases covering per-type aggregation, listen/non-platform queue exclusion, and missing type or consumer fields. `check-ts` and `lint` clean.

sbocahu:
Rebased onto the latest `master` (was 281 commits behind), branch head is now `8f536ccbd6`.

Three conflicts resolved, all in code that `master` had also touched:
- `httpPlatform.js` (x2): `master` added the `raw-file-storage` range helpers (`downloadFileRange`, `downloadLocalFileRange`) for the new PMTiles map route, while this branch removed the health probe imports. Kept both sides: the new range helpers stay, the health imports remain removed since `/health` now reads from the telemetry subsystem.
- `rabbitmq-test.ts`: `master` added the `buildSplitMessages` and `pushBundleToWorker` suites, this branch added the `getQueueConsumersByType` suite. Both suites kept.

Validation on the rebased tree:
- `yarn install` (picks up `sanitize-html`, newly required by `master`)
- 67 unit tests passing across `httpPlatform`, `platformHealthMetrics` and `rabbitmq` (40 from this branch + 27 added by `master`, all preserved)
- `yarn check-ts` and `yarn lint` clean

No functional change was made during the rebase: the telemetry subsystem is byte-identical to before, and the only delta in `httpPlatform.js` is `master`'s own new routes.

opencti-platform/opencti-graphql/src/database/engine.ts
xfournet: Addressed in the 09/03 rework (commit 99cd26b86c): the per-process cache was removed from `engine.ts`, `getEngineUsedSize` is back to a stateless function that queries ES/OpenSearch directly on every call — no more stale value returned on failure. Freshness and collection frequency are now owned by `platformHealthMetrics.ts` (background monitor, interval configurable via `app:health_monitoring`).

The redundant-computation-in-cluster point is still open though: each node runs its own timer with no coordination, so on N nodes we still get N independent `indices.stats` calls instead of one shared one (no Redis lock/cache). Leaving this thread open for your take on it.

opencti-platform/opencti-graphql/src/database/raw-file-storage.ts
xfournet: Same on the storage side: the local cache was removed (commit 99cd26b86c), `getStorageUsedSize` does a direct S3 scan on every call and propagates the error instead of returning a stale value. Freshness is now owned by the `platformHealthMetrics.ts` monitor.

The cluster-redundancy point is still open here too: each node does its own full bucket scan every 5 min with no Redis coordination — see my comment on `engine.ts`.

opencti-platform/opencti-graphql/src/database/raw-file-storage.ts
xfournet: Addressed in the 09/03 rework (commit 99cd26b86c): `getStorageUsedSize`/`getEngineUsedSize` are now generic functions that just propagate the error (see the "should propagate storage errors instead of reporting a size" test). Error handling and mapping to `null` on failure now happens at the monitor level (`collectUsageMetric` in `platformHealthMetrics.ts`), not in the function itself. Closing this one.

docs/docs/deployment/telemetry.md
```diff
@@ -22,6 +22,32 @@ The following metrics are exposed by the OpenCTI API.
 | `opencti_api_latency` | Histogram | Measures the latency of API query execution. | Milliseconds |
 | `opencti_api_direct_bulk` | Gauge | Measures the size of bulks for direct ingestion (fast path). | Count |
 | `opencti_api_side_bulk` | Gauge | Measures the size of bulks for absorption impacts (worker path). | Count |
+| `opencti_dependency_up` | Gauge | Reports the connectivity state of a platform dependency: `1` when the dependency answers, `0` when it fails. | Boolean |
```
xfournet, line 25: Keeping `dependency` here (metric `opencti_dependency_up`, label `dependency`) rather than switching to `connectivity`.

Rationale: this follows the standard Prometheus convention for target liveness — `<name>_up` plus a label naming the instance, the way node_exporter/blackbox_exporter expose `up`. `connectivity_interval` in the config key names the *action* (how often we probe), while `dependency` / `opencti_dependency_up` names the *result* being exposed (which dependency, up or down) — two different concepts, so I wouldn't rename either to match the other.

I also looked at whether `s3_used_size`/`es_used_size` (the `/health?details=true` fields) should be aligned to the `elasticsearch`/`storage` dependency identifiers used elsewhere — decided against it: `s3` is more specific and immediately clear to whoever reads the JSON payload (it says exactly what backend/protocol the size was measured through), whereas `storage` is intentionally the generic, protocol-agnostic name used for the connectivity/liveness dependency. Two different concerns, keeping both as they are.

opencti-platform/opencti-graphql/config/default.json
```diff
@@ -87,6 +87,10 @@
       }
     },
     "health_access_key": "ChangeMe",
+    "health_monitoring": {
+      "connectivity_interval": 30000,
```
xfournet: Naming: this key and `refreshConnectivityMetrics()` (in `platformHealthMetrics.ts`) are the only two places in this feature using "connectivity" — everything else that reads/writes this same state uses "dependency" (`dependencyStatuses`, `DependencyStatus`, `HealthDependency`, `HEALTH_DEPENDENCIES`, `dependencyProbes`, the `opencti_dependency_up` metric and its `dependency` label). Same single concept, two vocabularies depending on which file you're in — grepping "dependency" in config won't surface the interval that drives it.

Suggest renaming this key to `dependency_check_interval` (or `dependency_interval`) so "dependency" is the one term used end to end, config included. Companion comments on `configuration.md` and `platformHealthMetrics.ts`.

docs/docs/deployment/configuration.md
```diff
@@ -47,6 +47,8 @@ Here are the configuration keys, for both containers (environment variables) and
 | app:admin:password           | APP__ADMIN__PASSWORD           | ChangeMe              | Default password of the admin user                                                                                                                                               |
 | app:admin:token              | APP__ADMIN__TOKEN              | ChangeMe              | Default token (must be a valid UUIDv4)                                                                                                                                           |
 | app:health_access_key        | APP__HEALTH_ACCESS_KEY         | ChangeMe              | Access key for the `/health` endpoint. Must be changed - will not respond to default value. Access with `/health?health_access_key=ChangeMe`                                     |
+| app:health_monitoring:connectivity_interval | APP__HEALTH_MONITORING__CONNECTIVITY_INTERVAL | 30000 | Interval in milliseconds between two dependency connectivity checks (ElasticSearch/OpenSearch, S3/MinIO, RabbitMQ, Redis). The `/health` endpoint answers from the state collected by these checks. |
```
xfournet: Same naming remark as on `default.json`/`platformHealthMetrics.ts`: this is the only place (besides `refreshConnectivityMetrics()`) using "connectivity" for what's named "dependency" everywhere else in this feature (`dependencyStatuses`, `HealthDependency`, `opencti_dependency_up`...). Suggest renaming the key (and this doc row) to `app:health_monitoring:dependency_check_interval` for one consistent term across config, code and docs.

opencti-platform/opencti-graphql/src/telemetry/platformHealthMetrics.ts
```diff
@@ -0,0 +1,210 @@
...
+
+// Only dependencies that have actually been probed can be declared failing:
+// a not-yet-checked dependency is reported through the initialized flag instead.
+export const buildHealthFailures = (statuses: Record<HealthDependency, DependencyStatus>): string[] => {
+  return HEALTH_DEPENDENCIES
+    .filter((dependency) => statuses[dependency].checkedAt !== null && !statuses[dependency].isAlive)
+    .map((dependency) => `${dependency}: ${statuses[dependency].error ?? 'unavailable'}`);
+};
+
+export const refreshConnectivityMetrics = async (): Promise<void> => {
```
xfournet: Naming: this function (and `DEFAULT_CONNECTIVITY_INTERVAL_MS` above, the `connectivity_interval` config key it reads at line 175, and the two "Connectivity" log messages around line 180-181) are the only spots in this file using "connectivity". Everything this function actually writes to — `dependencyStatuses`, typed as `Record<HealthDependency, DependencyStatus>` — and everything downstream reading it (`opencti_dependency_up`, label `dependency`) is named "dependency".

Suggest renaming end to end: `refreshConnectivityMetrics` → `refreshDependencyStatus`, `DEFAULT_CONNECTIVITY_INTERVAL_MS` → `DEFAULT_DEPENDENCY_CHECK_INTERVAL_MS`, config key → `dependency_check_interval` (see companion comments on `default.json`/`configuration.md`), so "dependency" is the single term used from config to metric.

opencti-platform/opencti-graphql/src/http/httpPlatform.js
```diff
@@ -590,12 +587,18 @@ const createApp = async (app, schema) => {
...
-          const rabbitMQAlive = healthCheckTimeout(rabbitMQIsAlive(), 'Timeout checking rabbitmq health');
-          const redisAlive = healthCheckTimeout(redisIsAlive(), 'Timeout checking redis health');
-          await Promise.all([engineAlive, storageAlive, rabbitMQAlive, redisAlive]);
-          res.status(200).send({ status: 'success' });
+          // State is refreshed by the platform health monitor, never probed per request.
+          const { initialized, isHealthy, failures } = getPlatformHealthStatus();
+          if (!initialized) {
+            res.status(503).send({ status: 'error', error: 'Health monitoring not initialized yet' });
+          } else if (!isHealthy) {
+            res.status(503).send({ status: 'error', error: failures.join(', ') });
```
xfournet: `failures` is flattened into a single joined string here, so a consumer of `/health` only gets one opaque blob (e.g. "redis: Redis seems down, rabbitmq: RabbitMQ seems down") — no structured, per-dependency signal to tell which ones are actually down without parsing that string.

Worth noting the Prometheus side doesn't have this problem: `opencti_dependency_up` already exposes each dependency as its own time series via the `dependency` label, so a Prometheus consumer can already alert/aggregate per dependency. But anyone consuming `/health` directly (load balancers doing more than a binary check, or any monitoring that isn't scraping Prometheus) only sees the aggregate up/down + this string.

Suggest exposing the per-dependency booleans on the JSON response too — e.g. a `dependencies: { elasticsearch: true, storage: true, rabbitmq: false, redis: true }` field alongside (or instead of) `error` — built from the same `dependencyStatuses` that already backs `getPlatformHealthStatus()`/`buildHealthFailures()` in `platformHealthMetrics.ts`. Lets a caller build its own per-dependency stats/alerting instead of just knowing "something is down".

opencti-platform/opencti-graphql/src/database/engine.ts
sbocahu: Now addressed in b59492c51b — both halves of your point.

Collection no longer runs per node. `refreshUsageMetrics()` is private and only reachable through `syncUsageMetrics()`, which:

1. reads the shared value from Redis and adopts it if present (no computation at all);
2. on a miss, takes a `lockResource([...], { retryCount: 0 })` — nodes that lose the race return immediately and keep the previous cycle's values instead of piling on;
3. re-checks the cache *under* the lock, so a node that read just before another published doesn't recompute anyway;
4. publishes with `EX` set to `usage_metrics_interval`, which is what makes exactly one `indices.stats` call happen per interval cluster wide instead of N.

That also fixes the consistency half you raised: every node now reports the same figure, where before each had its own local number.

One behaviour worth knowing: a node that loses the lock race on a cold start reports `null` usage metrics until its next cycle. That was already the case before the first collection completed, so it isn't a regression, but it is visible on a fresh cluster.

opencti-platform/opencti-graphql/src/database/raw-file-storage.ts
sbocahu: Same fix on the storage side in b59492c51b — see the detail on `engine.ts`.

The full bucket scan is now behind the same Redis lock and shared value, so it runs once per `usage_metrics_interval` for the whole cluster rather than once per node. `getStorageUsedSize` itself stays generic and still just propagates its errors.

opencti-platform/opencti-graphql/config/default.json
sbocahu: Agreed, renamed in b59492c51b. The key is now `app:health_monitoring:dependency_check_interval` (`APP__HEALTH_MONITORING__DEPENDENCY_CHECK_INTERVAL`).

No back-compat fallback: `connectivity_interval` is introduced by this PR and has never shipped, so nothing is relying on the old name.

docs/docs/deployment/configuration.md
sbocahu: Renamed in b59492c51b — the row now documents `app:health_monitoring:dependency_check_interval`, matching the config key and the code.

opencti-platform/opencti-graphql/src/telemetry/platformHealthMetrics.ts
sbocahu: Renamed end to end in b59492c51b: `refreshConnectivityMetrics` → `refreshDependencyStatus`, `DEFAULT_CONNECTIVITY_INTERVAL_MS` → `DEFAULT_DEPENDENCY_CHECK_INTERVAL_MS`, the module-level `connectivityInterval` → `dependencyCheckInterval`, the two log messages, and the config key → `dependency_check_interval`.

`grep -ri connectivity` over the feature now returns nothing, so "dependency" is the only term from config to metric.

opencti-platform/opencti-graphql/src/http/httpPlatform.js
sbocahu: Good point, added in b59492c51b. `getPlatformHealthStatus()` now returns `dependencies: Record<HealthDependency, boolean>` built from the same `dependencyStatuses`, and `/health` exposes it as e.g. `{ "elasticsearch": true, "storage": true, "rabbitmq": false, "redis": true }`.

Two deliberate choices:

- **`error` is kept alongside it**, not replaced — the string is likely already consumed by existing monitoring.
- **It is gated behind `shouldIncludeHealthDetails()`**, the same gate as the usage metrics, rather than returned unconditionally. Naming which backend is down to an unauthenticated caller (`health_access_key: "public"`) is more disclosure than a binary up/down, which is the concern the bot raised earlier in this review.

It is returned on the 503 too, not only on the 200 — that is where knowing *which* dependency failed actually matters. Covered by a new test, "should name the failing dependencies on a 503 when details=true".

opencti-platform/opencti-graphql/src/telemetry/platformHealthMetrics.ts
```diff
@@ -0,0 +1,271 @@
...
+  const { queue_consumers } = usageMetrics;
+  return { ...usageMetrics, queue_consumers: queue_consumers === null ? null : { ...queue_consumers } };
+};
+
+export const startPlatformHealthMonitor = async (): Promise<void> => {
+  if (dependencyCheckInterval) {
+    return; // Already running
+  }
+  registerHealthGauges();
+  const dependencyCheckIntervalMs = conf.get('app:health_monitoring:dependency_check_interval') ?? DEFAULT_DEPENDENCY_CHECK_INTERVAL_MS;
```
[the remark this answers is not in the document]
sbocahu: Not applying this one — the premise is based on the issue description rather than on the branch.

`grep -rn "connectivity_interval\|dependency_check_interval"` over the repo returns four hits, all `dependency_check_interval`:

| Location | Key |
|---|---|
| `src/telemetry/platformHealthMetrics.ts:233` | `dependency_check_interval` |
| `config/default.json:91` | `dependency_check_interval` |
| `docs/docs/deployment/configuration.md:50` | `dependency_check_interval` |
| `docs/docs/deployment/telemetry.md:32` | `dependency_check_interval` |

So code, default config and docs agree, nothing is ignored, and the `?? DEFAULT_DEPENDENCY_CHECK_INTERVAL_MS` branch is only reached when the operator overrides nothing — in which case `default.json` already supplies the same 30000.

The stale reference was in the issue's implementation notes, which predated the review. Fixed in #17770.

On the substance: `connectivity_interval` was renamed at the reviewer's request (see the thread on `platformHealthMetrics.ts`), so that "dependency" is the single term used from config key to metric label, matching `dependencyStatuses`, `HealthDependency`, `opencti_dependency_up` and its `dependency` attribute. Going back to `connectivity_interval` would reintroduce exactly the inconsistency that was flagged.

No alias either: the key is introduced by this PR and has never been in a release, so no deployment can be relying on the old name.

opencti-platform/opencti-graphql/src/telemetry/platformHealthMetrics.ts
```diff
@@ -0,0 +1,271 @@
...
+    if (published !== null) {
+      usageMetrics = published;
+      return;
+    }
+    await refreshUsageMetrics();
+    await redisSetPlatformUsageMetrics(usageMetrics, usageMetricsTtlSeconds);
+  } finally {
+    await lock.unlock();
+  }
+};
```
aHenryJard: This looks like more and more like a cron manager implementation (including withTimeout, stop and start), except that you cannot choose on which node it's running (ingest/front) and it's not possible to disable it in case of production issue (you will have to wait for the bug hotfix + emergency release)

If the only missing piece to manager is the "start immediately" it can be added, it was missing for another topic anyway.

opencti-platform/opencti-graphql/src/database/engine.ts
```diff
@@ -5186,6 +5186,21 @@ export const getStats = (indices = READ_PLATFORM_INDICES) => {
   return retryElOperations(statsOperation);
 };
 
+// Branches are kept separate: ELK types the metric as an array, OpenSearch as a string,
+// and their client signatures are not mutually assignable.
+const fetchEngineUsedSize = async (): Promise<number> => {
+  if (engine instanceof ElkClient) {
+    const engineIndicesStats = await engine.indices.stats({ index: '*', metric: ['store'], expand_wildcards: 'all' as any });
```
aHenryJard: On shared elastic with several application installed I'm not sure it will be only OpenCTI size, I think you should only look at index with the ES_INDEX_PATTERN_SUFFIX and not '*'

opencti-platform/opencti-graphql/src/telemetry/platformHealthMetrics.ts
```diff
@@ -0,0 +1,271 @@
...
+    clearInterval(dependencyCheckInterval);
+    dependencyCheckInterval = null;
+  }
+  if (usageMetricsInterval) {
+    clearInterval(usageMetricsInterval);
+    usageMetricsInterval = null;
+  }
+  dependencyStatuses = buildInitialStatuses();
+  usageMetrics = buildInitialUsageMetrics();
+  logApp.info('[HEALTH] Platform health monitoring stopped');
```
aHenryJard, line 233: did you checked that redis lock is removed on stop ?

opencti-platform/opencti-graphql/src/database/engine.ts
sbocahu: Fixed in 6972e8d2f9: `fetchEngineUsedSize` now scopes to `${ES_INDEX_PREFIX}*` instead of `'*'`.

One correction: the constant that scopes to "OpenCTI's own indices" is `ES_INDEX_PREFIX` (defaults to `opencti`), not `ES_INDEX_PATTERN_SUFFIX` — the latter is the rollover suffix (`-000001`) appended at index-creation time, it doesn't identify ownership. `${ES_INDEX_PREFIX}*` is the exact pattern already used elsewhere in `engine.ts` for "all OpenCTI indices" (`elPlatformIndices`, `elPlatformTemplates`, `elCreateLifecyclePolicy`), so this keeps the same convention.

Updated the existing unit tests (`engine-used-size-test.ts`) to assert on the prefixed pattern.

opencti-platform/opencti-graphql/src/telemetry/platformHealthMetrics.ts
sbocahu: Checked: `USAGE_METRICS_LOCK` is acquired through `lockResource` (`redis.ts`), which wraps `Redlock` with a hard `maxTtl` (`app:concurrency:max_ttl`, defaults to 60000ms) — `redlock.acquire(locks, maxTtl)`. That TTL is enforced by Redis itself (`SET ... PX maxTtl`), independently of whether the app ever calls `.unlock()`.

So on the normal path, `syncUsageMetrics`'s `finally { await lock.unlock(); }` releases it right away. If `stopPlatformHealthMonitor()` happened to run while a collection was mid-flight, the lock isn't force-released by our code, but it self-expires within `max_ttl` regardless — same guarantee every other lock in the codebase relies on (e.g. the `draft_copy` lock in `engine.ts`). No permanent lock, worst case is a ~60s delay before another node picks up collection, and only in the narrow window of stopping exactly mid-collection.

opencti-platform/opencti-graphql/src/telemetry/platformHealthMetrics.ts
sbocahu: Fair comparison, but the manager framework's core assumption doesn't fit half of this feature: `cronHandler` in `managerModule.ts` acquires a cluster-wide `lockResources` lock so **exactly one node** runs the handler per tick.

That's fine for `syncUsageMetrics` (expensive, cluster-wide result — already built the same way with our own Redis lock). But `refreshDependencyStatus` must run independently on every node: it reports *that node's own* connectivity to ES/S3/RabbitMQ/Redis, not a cluster-wide fact. If node A can't reach Redis but node B can, both need to report their own truth — wrapping it in a standard manager would turn per-node liveness into "whichever node wins the lock decides for everyone," which is wrong. So this can't become a single manager as-is; at best it'd split into a manager for the usage-metrics half and something else, un-lock-gated, for the dependency-status half.

Agreed on "start immediately" being a real, separate gap in `managerModule` today (confirmed — it only schedules via `setIntervalAsync`, no immediate first run), and on the disable-without-a-release point: happy to add a plain `app:health_monitoring:enabled` config flag to close that specific gap without pulling in the whole manager framework.

sbocahu: @aHenryJard I split the concerns: storage metrics use the manager pattern, while no change to the dependencies connectivity metrics were made.

docs/docs/deployment/telemetry.md
sbocahu: resolved in a previous commit

aHenryJard (commented):
Tested locally:

- `es_used_size` number is good
- `s3_used_size` is good

queue_consumers numbers looks good.

The blocker for me is that when I stop redis locally, the /health endpoint does not answer it is just hanging for ever.

When I stop `elastic` or `minio` or `rabbitmq`, the /health is correctly marking it as error / false. As side note in that case `queue_consumers` is not exposed anymore.

The only annoying point I have is when redis is KO /health endpoint does not answer. It does not seem great for monitoring tools.

aHenryJard:
> Tested locally:
> 
> * `es_used_size` number is good
> * `s3_used_size` is good
> 
> queue_consumers numbers looks good.
> 
> The blocker for me is that when I stop redis locally, the /health endpoint does not answer it is just hanging for ever.
> 
> When I stop `elastic` or `minio` or `rabbitmq`, the /health is correctly marking it as error / false. As side note in that case `queue_consumers` is not exposed anymore.
> 
> The only annoying point I have is when redis is KO /health endpoint does not answer. It does not seem great for monitoring tools.

After checking on master branch, the behavior when redis is down is the same, not related to this PR, here is the issue https://github.com/OpenCTI-Platform/opencti/issues/18338

aHenryJard (approved):
Looks good to me

Issue #16785, opened 2026-06-23
feat(domain-specific-platform): Global configuration export


alice-debra:
Export the full OpenCTI platform configuration as a ZIP from the Configuration section, with a drawer to select which elements to include (V0 scope).

Generate one JSON file per category (platform, dashboards, feeds, playbooks, forms, custom views, branding, etc.), include a version field, and follow a clear naming convention.

Ensure secrets/credentials are excluded, and validate key design choices (format, ordering for import, logo handling, compatibility).

PR #17420, opened 2026-07-30
feat(domain-specific-platform): Global configuration export (#16785)

SarahBocognano:
… (#16785)

### Proposed changes

*
*

### Related issues

* closes

### Checklist


opencti-platform/opencti-graphql/src/migrations/singleton_internal_id.ts
```diff
@@ -0,0 +1,244 @@
...
+  expectedId: string;
+  label: string;
+  settingsType?: string;
+}
+
+const BY_NAME_ROWS: ByNameRow[] = [
+  // Theme
+  { entityType: ENTITY_TYPE_THEME, name: 'Dark', expectedId: THEME_DARK_ID, label: 'Theme' },
+  { entityType: ENTITY_TYPE_THEME, name: 'Light', expectedId: THEME_LIGHT_ID, label: 'Theme' },
+
```
fellowseb, line 155: After giving it some thought I'm not sure we should generate a unique stable internal_id for those default themes.
Seems counter intuitive UX-wise to overwrite any changes potentially made on the target platform.
I would say the same for all default or sample entities created at platform init. Only assign stable internal_id to built-in entities that are not modifiable by the user.

BUT, this behaviour needs to be confirmed at Product level. So let's write a list of all these "potential singleton" entities on the specs page and ask @alice-debra her thoughts 👍

PR #17424, opened 2026-07-30
fix(ipwhitelist): add superadmin user in ip whitelist exclusion (#17264)

### LINKED ISSUE #17264: fix(playbook):  playbook not appliying marking to observables
## Description
A playbook listens for connector creators and then applies a marking def before sending for ingestion. It is not consistently applying the marking def.

## Environment
1. 7.260715.0

## Reproducible steps

1. Have a playbook that listen on a specific creator, then applies a marking definition (custom marking definition, not TLP or PAP) & send for ingestion
2. Create some entity with this user (with or without marking)

## Expected output
Mdef from stage 2 to be applied

## Actual output
Marking definition is not consistlently applied.

Together with CSM, this is the playbook we used: 
1. See the playbook imported here: https://csm.octi.staging.filigran.io/dashboard/data/processing/automation/ae6d2965-6294-4011-aef7-e4506c248b01
2. (Testing the left branch).  It has been scoped to me as a creator, so adjust as needed. 
3. I tested on a software observable with no current marking def, then manually enrolled the playbook. The 'internal use' marking def was applied (from node 2 on the PB). Removing the applied marking and then adding one (manually) like TLP Amber before then re -enrolling the playbook, results in failures from then on.
4. Then trying a domain name observable, this time with a TLP Amber initially applied. Enrolling the PB results in nothing being applied at all.

It only worked once on this entity: https://csm.octi.staging.filigran.io/dashboard/observations/observables/abbe77d9-834d-492f-ae87-c2cfa064d5f2/overview

## Additional information
Original customer ticket: OCTI1-3870

JeremyCloarec:
### Proposed changes

This pull request updates the IP whitelist middleware to ensure that the platform super admin user is always excluded from IP restrictions, regardless of any exclusion list configuration. It also adds unit tests to verify this behavior.

### Security and Access Control

* Updated `isUserExcluded` in `ipWhitelistMiddleware.js` to always return `true` for the platform super admin user (`OPENCTI_ADMIN_UUID`), ensuring the super admin is never subject to IP restrictions. [[1]](diffhunk://#diff-eacd72b13f18882c0cb68795119269dd3850baf62f26f36ad54e1e363dde5c73R5) [[2]](diffhunk://#diff-eacd72b13f18882c0cb68795119269dd3850baf62f26f36ad54e1e363dde5c73L75-R84)

### Testing

* Added unit tests in `ipWhitelistMiddleware-test.ts` to confirm that the super admin is always excluded, even when exclusion IDs are empty or do not match. [[1]](diffhunk://#diff-1ff5ef895a17f0ff238d6074096a8ce7f70638fc3c0f4e9e0169da029475c979R7) [[2]](diffhunk://#diff-1ff5ef895a17f0ff238d6074096a8ce7f70638fc3c0f4e9e0169da029475c979R230-R244)

### Related issues

* closes #17264

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/domain/user.js
+++ b/opencti-platform/opencti-graphql/src/domain/user.js
@@ -1712,6 +1712,11 @@ const checkIpWhitelistOnLogin = async (context, loggedUser) => {
   // If IP matches whitelist, allow
   if (ipMatchesWhitelist(sourceIp, whitelist)) return;
 
+  // Always exclude platform superadmin
+  if (loggedUser.id === OPENCTI_ADMIN_UUID) {
+    return;
+  }
+
   // Check exclusion list
   const exclusionIds = settings.platform_ip_whitelist_exclusion_ids;
   if (isNotEmptyField(exclusionIds)) {
--- a/opencti-platform/opencti-graphql/src/http/ipWhitelistMiddleware.js
+++ b/opencti-platform/opencti-graphql/src/http/ipWhitelistMiddleware.js
@@ -219,6 +219,11 @@ const ipWhitelistMiddleware = async (req, res, next) => {
       });
     }
 
+    // Always exclude platform superadmin
+    if (authenticatedUser.id === OPENCTI_ADMIN_UUID || authenticatedUser.internal_id === OPENCTI_ADMIN_UUID) {
+      return next();
+    }
+
     // Authenticated (session or token): check exclusion list
     const exclusionIds = settings.platform_ip_whitelist_exclusion_ids;
     if (isNotEmptyField(exclusionIds)) {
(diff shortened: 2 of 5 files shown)


gabrielebonadiman:
qq - is the issue linked the correct one closed by this pr? 


qq 2 - Implementation wise I have no comments :) although, I might have a product question. should `superadmin` really bypass IP whitelisting unconditionally, with no opt-out besides the global  app:ip_whitelist_enabled  kill-switch? 

This removes `defense-in-depth` for the single highest-privileged account 😅  exactly the one where a compromised credential does the most damage. It might be I am not seeing the right issue but is this the intended trade-off, or should this be scoped tighter?

Issue #14716, opened 2026-03-02
feat(security-coverage): multi-OpenAEV security coverage results


romain-filigran:
## Use case

The current OpenCTI ↔ OpenAEV integration was designed for a 1:1 architecture (one OpenCTI instance connected to a single OpenAEV instance). It does not support a context where a single OpenCTI instance needs to orchestrate and consolidate Security Coverage results from multiple OpenAEV instances.

## Expected Outcome

- A Security Coverage in OpenCTI can target one or more OpenAEV instances
- Results are tagged / partitioned by source OpenAEV instance
- The Security Coverage UI provides a per-instance breakdown of results (detection status per Attack Pattern per OpenAEV instance)
- Users can compare coverage gaps across instances from a single view

PR #17455, opened 2026-07-31
fix(stix-export): Export Security Coverage Results when doing a Security Coverage full stix export (#14716)

MTorbay-Filigran:
### Proposed changes

* Updated the client Python so that the Security Coverage Results are resolved and exported as distinct objects of the bundle, and added the missing part in the backend to include the SCR in the SC bundle when exporting

### Related issues

* closes 

### How to test this PR

Import a stix bundle from a security coverage that has results, selecting full export :
Capture d’écran 2026-07-31 à 17 11 13

You should find in the bundle you receive the results as distinct objects in the bunlde objects list
Capture d’écran 2026-07-31 à 17 12 28

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/modules/securityCoverage/securityCoverage-domain.ts
+++ b/opencti-platform/opencti-graphql/src/modules/securityCoverage/securityCoverage-domain.ts
@@ -163,9 +163,17 @@ export const addSecurityCoverage = async (
 
 export const securityCoverageStixBundle = async (context: AuthContext, user: AuthUser, SecurityCoverageId: string) => {
   const objects = [];
-  const SecurityCoverage = await storeLoadByIdWithRefs(context, user, SecurityCoverageId) as StoreEntitySecurityCoverage;
+  const SecurityCoverage = await storeLoadByIdWithRefs(context, user, SecurityCoverageId) as StoreEntitySecurityCoverage & BasicStoreEntitySecurityCoverage;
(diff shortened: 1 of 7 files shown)


opencti-platform/opencti-graphql/src/modules/securityCoverage/securityCoverage-domain.ts
```diff
@@ -163,9 +163,17 @@ export const addSecurityCoverage = async (
...
   const stixSecurityCoverage = convertStoreToStix_2_1(SecurityCoverage);
   objects.push(stixSecurityCoverage);
+  const securityCoverageResultIds = SecurityCoverage[RELATION_RESULT_OF] ?? [];
+  if (securityCoverageResultIds.length > 0) {
+    const results = await storeLoadByIdsWithRefs(context, user, securityCoverageResultIds);
+    for (const resultElement of results) {
+      const stixSecurityCoverageResult = convertStoreToStix_2_1(resultElement);
+      objects.push(stixSecurityCoverageResult);
+    }
+  }
```
lndrtrbn: Not sure this part is required, at least it is not needed to make the full export to work. But maybe it's useful for something else, @SouadHadjiat what do you think?

SouadHadjiat: oh right, I don't think it's useful either for the export, and this will send to openAEV the results in the bundle

lndrtrbn (approved):

PR #17460, opened 2026-08-01
docs: explain bulk observable creation (#17354)

### LINKED ISSUE #17354: docs: clarify why bulk/multiple observable creation is unavailable for types with multiple key fields (e.g. Software)
**Description**

The UI supports creating multiple observables at once (bulk paste, one value per line) for observable types that are identified by a single key field, such as IPv4 Address, Domain Name, URL, and file hashes. This works well and is a great time-saver for analysts.

For some observable types, such as Software, bulk creation is not available. This is because Software (and similar types) are identified by more than one attribute (e.g. name, vendor, version, CPE, SWID), so there isn't a single key value that bulk paste can map to. The UI does surface a short inline note about this in the creation dialog, but it would help to have this explained more clearly in the product documentation.

**Request**

Could the documentation include a short section (e.g. in the Observables / Data Model section) explaining:
- Which observable types support bulk/multiple creation and why (single key field).
- Which types do not, and why (multiple identifying fields), using Software as an example.
- Where relevant, what the recommended workflow is for creating many observables of these "multi-key" types (e.g. via CSV mapper/import).

This would help users understand the behavior is by design rather than a limitation or regression, and reduce related support questions.

**Type**: Documentation enhancement

codewithdaniel1:
### Proposed changes

* Adds a new section to the Observables documentation explaining manual multiple observable creation.
* Clarifies that multiple creation is available when OpenCTI can map each input line to a clear identifying field, and that multi-field observable types such as `Software` should be created individually or imported with a CSV mapper.

### Related issues

* closes #17354

### How to test this PR

This is a documentation-only change.

Review the updated Observables documentation page and confirm that the new "Creating multiple observables" section explains:
* why some observable types support multiple creation
* why multi-field observable types such as `Software` do not support single-value bulk creation
* when to use CSV mappers for structured bulk imports

I also ran:

```bash
git diff --check
```

### Checklist

### Further comments

This PR is documentation-only, so automated test cases and code refactoring are not applicable.

### DIFF
--- a/docs/docs/usage/exploring-observations.md
+++ b/docs/docs/usage/exploring-observations.md
@@ -20,6 +20,25 @@ When clicking on the Observables tab at the top left, you see the list of all th
 
 ![Observables list](assets/observables-list-view.png)
 
+### Creating multiple observables
+
+When creating Observables manually, some observable types support multiple
+creation by pasting one value per line. This is available when OpenCTI can map
+each line to a single identifying field for the selected observable type, such
+as a domain name, an IP address, a URL, or a file hash.
+
+Some observable types cannot be created this way because they are identified
+by several fields rather than one pasted value. For example, `Software` can
+require information such as a name, vendor, version, Common Platform
+Enumeration (CPE), or Software Identification (SWID), so a single line of text
+cannot reliably describe one complete `Software` observable.
+
+For these observable types, use the standard creation form for individual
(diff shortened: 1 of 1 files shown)


nino-filigran:
Thanks for your input, looks good to me!

nino-filigran (approved):

Issue #17258, opened 2026-07-22
feat(userManagement): Merge users


nino-filigran:
## Use case

When setting up a platform, you might create some users that are creating data. and later on, when you properly onboard all users, using for instance SSO, you might need to reconciliate some existing users with some of these news ones. 

## Proposed solution

- Either have a proper functionality of merge user (merge should transfer data ownership, activity, history...)
- Or have a script allowing to do it.

## If the feature request is approved, would you be willing to submit a PR?

Yes / No (help can be provided if you need assistance submitting a PR)

PR #17507, opened 2026-08-04
feat(merge-users): feature flag, API surface and Bypass enforcement (#17493)

### LINKED ISSUE #17493: PR1 — Feature Flag + userMerge GraphQL mutation + Bypass gate
### Objective

Expose the merge capability as an API entry point, gated by a Feature Flag and the Bypass capability. This PR ships the surface only — no merge logic yet — so the endpoint and its security boundary can be reviewed in isolation.

### Out of scope

- Any actual data rewrite (delegated to later PRs).
- Dry-run report content (framework arrives in PR2).
### Dependencies

- None (first PR).
### Definition of Done

- Mutation is callable only with Bypass and only when the FF is on, and an unauthorized call fails loudly instead of returning a success-shaped result.
- UserMergeResult is proven to have the same shape in dry and real mode.
- Input validation is unit-tested.
- The follow-up query exists, is Bypass-gated and returns a typed placeholder; its behaviour against a real journal — readable from a separate session while a merge is still running — is covered once PR2 lands.
- Returns a typed placeholder result and does not mutate any data.

ludovic:
### Proposed changes

* Adds the `MERGE_USERS` feature flag, off by default (`app:enabled_dev_features` is empty in `config/default.json`).
* Adds the `userMerge` mutation and the `userMergeJournal` query, both `@auth(for: [BYPASS])` and behind the flag. Bypass and the flag are enforced a second time in the domain with an explicit throw, so an internal caller cannot bypass the GraphQL directives and a refused call cannot come back shaped like a success.
* Adds the guard-rails: no self-merge, no unknown user, no internal platform user (SYSTEM, RETENTION MANAGER, …) as source or target.
* Ships the API surface only. The engine is a stub that returns `FAILED` with an explicit message, so deploying this chunk alone cannot let an operator believe a merge happened. The real engine lands in #17494.
* `dryRun` defaults to `true` and `rightsStrategy` to `STRICT`, so calling the mutation with no options can neither rewrite data nor widen rights.
* `userMergeJournal(mergeId: ID)` takes an optional id: the merge runs synchronously and a dead HTTP connection does not stop it, so an operator who lost the connection also lost the id returned by the mutation.

Two deviations from the specification: the return type is `UserMergeResult!` where the spec writes `UserMergeResult`, and the internal-user guard-rail is beyond the specified scope.

### Related issues

* closes #17493 — part of #17258

### How to test this PR

Enable the flag, then start the platform:

```bash
yarn start
```

Call the mutation with a Bypass token. It must answer `FAILED` with the stub message, and never `SUCCESS`:

```bash
  -d '{"query":"mutation { userMerge(sourceId:\"<source>\", targetId:\"<target>\") { merge_id status message } }"}'
```

Then check the refusals:

* the same call with a non-Bypass token is rejected;
* the same call with the flag unset is rejected;
* `sourceId` equal to `targetId` is rejected;
* either id pointing at an internal platform user is rejected.

Automated coverage:

```bash
```

### Checklist

### Further comments

This is the base of a stack of 11 branches (#17493 → #17503). Reviewing it in isolation is the point: everything after it assumes the flag, the capability check and the result shape are settled here.

The stub deliberately reports `FAILED` rather than `SUCCESS`. A stub answering SUCCESS is exactly the `cleanAllEntityInconsistencies` behaviour this feature is specified against.

### DIFF
--- a/opencti-platform/opencti-graphql/src/modules/userMerge/userMerge-domain.ts
+++ b/opencti-platform/opencti-graphql/src/modules/userMerge/userMerge-domain.ts
@@ -85,6 +85,19 @@ export const userMerge = async (
   return executeUserMerge(sourceId, targetId, resolvedOptions);
 };
 
+// Bounds enforced here so that a future, real journal implementation cannot be abused
+// into a heavy read regardless of what a caller (GraphQL or otherwise) requests.
+const USER_MERGE_JOURNAL_DEFAULT_FIRST = 20;
+const USER_MERGE_JOURNAL_MIN_FIRST = 1;
+const USER_MERGE_JOURNAL_MAX_FIRST = 200;
+
+export const resolveUserMergeJournalFirst = (first?: number | null): number => {
+  if (first === undefined || first === null) {
+    return USER_MERGE_JOURNAL_DEFAULT_FIRST;
+  }
+  return Math.min(Math.max(first, USER_MERGE_JOURNAL_MIN_FIRST), USER_MERGE_JOURNAL_MAX_FIRST);
+};
+
 export const userMergeJournal = async (
   _context: AuthContext,
   user: AuthUser,
@@ -92,5 +105,5 @@ export const userMergeJournal = async (
   first?: number | null,
 ): Promise<UserMergeJournalEntry[]> => {
   assertUserMergeAllowed(user);
-  return readUserMergeJournal(mergeId ?? undefined, first ?? undefined);
+  return readUserMergeJournal(mergeId ?? undefined, resolveUserMergeJournalFirst(first));
 };
--- a/opencti-platform/opencti-graphql/src/modules/userMerge/userMerge-domain.ts
+++ b/opencti-platform/opencti-graphql/src/modules/userMerge/userMerge-domain.ts
@@ -1,0 +1,96 @@
+import { isFeatureEnabled, MERGE_USERS_FEATURE_FLAG } from '../../config/conf';
+import { ForbiddenAccess, FunctionalError, UnsupportedError } from '../../config/errors';
+import { storeLoadById } from '../../database/middleware-loader';
+import { ENTITY_TYPE_USER } from '../../schema/internalObject';
+import type { BasicStoreCommon } from '../../types/store';
+import type { AuthContext, AuthUser } from '../../types/user';
+import { INTERNAL_USERS, isBypassUser } from '../../utils/access';
+import { executeUserMerge, readUserMergeJournal } from './userMerge-engine';
+import { type UserMergeJournalEntry, type UserMergeOptions, type UserMergeResult, UserMergeRightsStrategy } from './userMerge-types';
+
+/**
+ * Options as they arrive from GraphQL: every field is optional there, because the schema
+ * carries the defaults. They are re-applied here so that a caller that is not the GraphQL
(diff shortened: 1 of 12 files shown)


ValentinBouzinFiligran:
Great PR, thanks for the work 🙌
The approach is clear and safe.

One small improvement (non-blocking from my side):
`userMergeJournal` currently accepts `first` without an explicit bound.
I’d recommend capping it in the domain layer (e.g., min 1, max 100/200) to avoid potential heavy reads once the journal is fully implemented.

Otherwise, looks good to me 👍

opencti-platform/opencti-graphql/src/modules/userMerge/userMerge.graphql
```diff
@@ -0,0 +1,55 @@
...
+type UserMergeResult {
+    id: ID!
+    source_id: ID!
+    target_id: ID!
+    dry_run: Boolean!
+    rights_strategy: UserMergeRightsStrategy!
+    status: UserMergeStatus!
+    started_at: DateTime!
+    completed_at: DateTime
+    message: String
```
marieflorescontact, line 30: what will be the usage of the message ?

marieflorescontact (commented):
I still have work review to do on this PR but I have question:
why did not you use background task instead of synchronous operation with a new tracking system in Redis ? I thought background task was more appropriate in this case. but maybe I am missing something.

opencti-platform/opencti-graphql/src/modules/userMerge/userMerge.graphql
ludovic: It carries the error message, if there is one.

ludovic:
> why did not you use background task instead of synchronous operation with a new tracking system in Redis ? I thought background task was more appropriate in this case. but maybe I am missing something.

Good question

Background tasks are element-scoped; an action applied to each element of a list or a query. A merge isn't that: it's ~40 register rows, update_by_query passes and ref-relation re-points, so it would need a new task type anyway. And the feature is meant to run on an idle platform with the flow dry-run => read the report => decide => run, which needs a synchronous answer. 
On Redis, it's not a tracking system for the merge itself; everything is written to Elasticsearch. Redis only holds the execution journal, which is diagnostic: what authorises deleting the source account is the coverage manifest, not the journal. Making it an indexed entity would permanently spend from the shared mapping budget for something that runs once, and it would carry a creator_id the merge rewrites too.

opencti-platform/opencti-graphql/src/modules/userMerge/userMerge-domain.ts
```diff
@@ -0,0 +1,109 @@
...
+  context: AuthContext,
+  user: AuthUser,
+  sourceId: string,
+  targetId: string,
+  options?: UserMergeOptionsInput | null,
+): Promise<UserMergeResult> => {
+  assertUserMergeAllowed(user);
+  assertValidUserMergeIds(sourceId, targetId);
+  await loadMergeableUser(context, user, sourceId, 'source');
+  await loadMergeableUser(context, user, targetId, 'target');
```
CelineSebe, line 83: loadMergeableUser calls for source/target are sequential, could be Promise.all for a minor latency win

CelineSebe (approved):

PR #17513, opened 2026-08-04
fix: use window.location.href instead of pathname to handle APP_BASE_PATH correctly (#17291)

### LINKED ISSUE #17291: fix(frontend): OpenCTI UI cannot download documents when APP_BASE_PATH is defined
## Description

In an OpenCTI instance running with a defined APP_BASE_PATH attempts to download a document result in a URL that contains two copies of the base path (e.g., ../opencti_dev/opencti_dev/storage/get/...)

## Environment

OS: Linux
Version: 7.260715.0, and probably at least several versions prior as well
Other environment details: None

## Reproducible steps

1. Select Data/Entities
2. Filter to entity_type=Report
3. Select a report/content
4. Click download

## Expected output

An actual file download.

## Actual output

Redirects back to the home page in a new browser tab.  Does not download the file.

## Additional information

None.

waterWang:
Fixes #17291

## Problem

When APP_BASE_PATH is configured, clicking download on a document creates a URL with a duplicated base path (e.g., /opencti_dev/opencti_dev/storage/get/...), causing the download to fail.

## Root Cause

The handleLink function in FileLine.tsx used window.location.pathname = url to navigate, which doesn't handle the base path correctly.

## Fix

Changed window.location.pathname = url to window.location.href = url, which correctly handles both absolute and relative URLs.

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/common/files/FileLine.tsx
+++ b/opencti-platform/opencti-front/src/private/components/common/files/FileLine.tsx
@@ -234,7 +234,7 @@ const FileLineComponent: FunctionComponent<FileLineComponentProps> = ({
     if (inNewWindow) {
       window.open(url, '_blank', 'noopener,noreferrer');
     } else {
-      window.location.pathname = url;
+      window.location.href = url;
     }
   };
   const generateIcon = () => {


romain-filigran:
Hey @waterWang. Thank you very much for your contribution, we really appreciate it!

Before we can merge your PR, we need you to sign our CLA (Contributor License Agreement), which allows us to incorporate your contribution into the project.

Is this something you're planning to do soon?

Additionally, your commits must be signed. Can you re-sign your commit?

mcartaud:
Thanks for your contribution. The fix does not fix the linked issue. I have implemented what seem to be the correct fix [here](https://github.com/OpenCTI-Platform/opencti/pull/17704).
I am closing this PR for now but if your fix is solving another aspect missing in my other PR please reopen it.
Thanks again

Issue #84, opened 2019-06-28
INTERNAL_SERVER_ERROR - "Int cannot represent non-integer value:"


fr33tux:
Hello,

## Description

Pages throws a "ERROR" and are not displayed correctly at the second request.

## Environment

1. OS (where OpenCTI server runs): Ubuntu 18.04.2, x64.
2. OpenCTI version: [opencti-20190628.tar.gz](https://github.com/OpenCTI-Platform/opencti/releases/tag/1.0.0)
3. OpenCTI client: frontend
4. Other environment details: Fresh installation.

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Install OpenCTI following the manual documentation
2. Click on any page, the first request usually works, but the second throws the "ERROR" message on the top left corner.

## Expected Output

Pages correctly displayed

## Actual Output

- "ERROR" is displayed on all pages:

![Screenshot from 2019-06-28 23-42-53](https://user-images.githubusercontent.com/5955898/60373284-ee958f80-99ff-11e9-8bb0-30747ac3e464.png)

The browser console throws this error:

```
Error: ""
    i construct.js:30
    a wrapNativeSuper.js:22
    n environment.js:39
    render environment.js:108
    render ReactRelayQueryRenderer.js:216
    React 11
    handleDataChange ReactRelayQueryRenderer.js:142
    error ReactRelayQueryFetcher.js:194
    error ReactRelayQueryFetcher.js:193
    error RelayObservable.js:553
    error RelayObservable.js:553
    r RelayObservable.js:178
    error RelayObservable.js:553
    error RelayObservable.js:553
    error RelayReplaySubject.js:105
    error RelayReplaySubject.js:104
    error fetchQueryInternal.js:126
    error RelayObservable.js:553
    error RelayObservable.js:553
    _error RelayModernQueryExecutor.js:190
    error RelayModernQueryExecutor.js:85
    error RelayObservable.js:553
react-dom.production.min.js:4408
    React 7
    unstable_runWithPriority scheduler.production.min.js:274
    React 7
    handleDataChange ReactRelayQueryRenderer.js:142
    error ReactRelayQueryFetcher.js:194
    error ReactRelayQueryFetcher.js:193
    error RelayObservable.js:553
    error RelayObservable.js:553
    r RelayObservable.js:178
    error RelayObservable.js:553
    error RelayObservable.js:553
    error RelayReplaySubject.js:105
    error RelayReplaySubject.js:104
[...]

Issue #81, opened 2019-06-28
Integrate with multiple MISP Instances


seifreed:
I would like to see the possibility to connect as many MISP instances as the admin desires. This could help people that have access to multiple of those instances.
 
## Problem to Solve

Nowadays only accepts connect to one MISP instance

## Proposed Solution

Add the possibility to connect more than one instance

SamuelHassine:
This issues has been moved to [the connectors repository](https://github.com/OpenCTI-Platform/connectors/issues/8).

PR #17514, opened 2026-08-04
feat(fds): replace the left navigation with the design-system Navbar (#17539)

### LINKED ISSUE #17539: feat(fds): replace the left navigation with the design-system Navbar
## Use case

The Filigran Design System (FDS) is being rolled out across the XTM product line so that shared components are maintained once instead of once per product. The left navigation is the first component migrated in OpenCTI.

Today OpenCTI maintains its own navigation: an inline JSX tree of roughly 900 lines, its own selection logic, its own collapsed-state persistence, and MUI menu items rather than real links. Every fix to navigation behaviour has to be made again in each product.

## Current workaround

Each product keeps its own implementation and they drift apart. A first migration was carried out on OpenAEV, which produced an implementation playbook; OpenCTI is the second application of it.

## Proposed solution

Replace the hand-written left navigation with the `Navbar` component from `@filigran/design-system`, with strict iso-functionality: same entries, same permissions, same submenus, same persisted collapsed state, same "show submenu icons" user setting.

What it brings:
- entries become real links, so Ctrl-click and middle-click open a new tab
- the inline JSX tree becomes a typed menu structure that permissions and feature flags traverse explicitly
- rail widths become a shared constant instead of an import from the navigation component itself
- full keyboard accessibility and correct ARIA semantics come from the library

Out of scope: the top bar, breadcrumbs and other menus are unchanged. Custom themes keep working — the navigation background and the selected-entry accent still follow a user-defined theme. No visual regression is expected for end users.

## Additional information

The library is a private git dependency, so this also wires the authentication needed to install it, including inside the Docker builds — which is specific to OpenCTI.

Yes — the PR is already open and green.

sandyghs:
## What changed, in plain terms

The left navigation is now the design system's `Navbar`, not a hand-built MUI
`Drawer`. Same entries, same permissions, same destinations — but the rows are
real links, so Ctrl-click and "open in a new tab" work everywhere, and the rail
is painted by the design system instead of by ~870 lines of product CSS.

Four things a user can notice:

- The collapsed rail is **48px** wide instead of 55px. That is the design
  system's real rail width; the seven floating toolbars that align themselves
  against it read the constant, so they shift 7px with it.
- Custom platform themes keep working exactly as before: a `theme_background`
  still paints the rail, a `theme_primary` still colours the selected row, and
  a draft still turns the accent orange.
- Standard installations now get the design system's own rail gradient rather
  than a reimplementation of it.
- The "Made by Filigran" signature keeps its two states: the wordmark when the
  rail is expanded, the emblem alone when it is collapsed — cropped from the
  same asset, so there is no second file to keep in sync with the brand.

### Proposed changes

* Replace the left navigation (`LeftBar.jsx` and its ~870 lines of styling)
  with `NavBar.tsx`, built on the design system's `Navbar`, `NavbarItem`,
  `NavbarSubmenu` and `ProductSwitcher`, keeping the menu structure, the three
  permission mechanisms, the feature flags and the `submenu_show_icons` user
  preference.
* Install the private design system in the containerised build: BuildKit build
  secrets for the three Dockerfiles, explicit per-run injection for the
  workflows that install outside an image.
* Re-anchor the end-to-end left-bar page object on what the component actually
  emits, and keep the rail's width contract for the eight files that read it.

### Related issues

* closes #17539

## Pinned library version

`486cec92c3abf006997ac269d34ff0fcc23f178f` — head of `main`. Verified
installable from a clean external project **and** from inside a container
before being pinned.

Bumped from `56f7e59823cae7d815a451206e3cb4cb1d31022d` to pick up library
PR #84, which gives button-rendered rail rows the pointer cursor. Before it,
**every** button-rendered row of this rail resolved to `cursor: default` — 14
expanded, 1 collapsed — while link rows got the hand cursor for free from the
user-agent sheet. That was a regression against the MUI navigation being
replaced, and no structural diff can see it. After the bump: 13/14 and 1/1.

The one row left out is the `ProductSwitcher` trigger, which is a sibling
component rather than a `NavbarItem`. Filed as observation 12 and deliberately
**not** compensated here — a host rule would hide the gap from every other
consumer.

Re-checked at the new pin, and recorded in `fds-migration/IMPLEMENTATION-LOG.md`:

- **No compensation could be retired.** `Navbar.tsx` and `NavbarSubmenu.tsx` are
  byte-identical between the two pins; `NavbarProps` still exposes no accent
  prop and `NavbarSubmenuProps` still exposes a single `open`/`onOpenChange`
  pair.
- **Nothing this rail depends on moved.** Set-diffing the compiled
  `dist/index.css` between the pins: 641 → 689 class selectors, 414 → 420
  custom properties, **zero removals** on either — so the accent guard that
  reads the installed stylesheet still holds.
- **One intentional visual change**, inside the product dropdown: library
  PR #81 moved `ProductSwitcher`'s panel onto `Menu`, so row padding-right goes
  8px → 16px, panel min-width 192px → 200px with a new cap and bounded height,
  divider margin 2px → 4px, group header band 24px → 32px, and disabled rows
  are painted with the disabled tokens instead of `opacity-60`. Row height
  (36px) and label size (14px) are unchanged — checked in the running platform.
- **Package size grew**, it did not shrink: 3 384 759 → 3 637 106 bytes
  (+252 347, +7.5%), 60% of it source maps and typings for the new `Header` and
  `Menu` primitives. There was no documentation to drop: the package has
  shipped `files: ["dist"]` since the initial monorepo scaffold, an ancestor of
  both pins.

## The point of this pull request

Beyond the navigation itself, this is the first time OpenCTI's continuous
integration installs the private design system. OpenCTI installs **inside
containers** — three Dockerfiles and four workflows — and a credential placed
on the host does not cross the container boundary. So:

- the three Dockerfiles receive the token as a **BuildKit build secret**, used
  only for the duration of the `yarn install` layer and never written to an
  image layer (verified: no match in image history, in any gitconfig, or
  anywhere on the filesystem of the built image);
- the workflows that install outside an image inject it explicitly per
  container run, with tracing disabled around the credential;
- `git` is installed where it was missing (`node:22-alpine` ships without it),
  including in the shared `builder` stage, because that image is exported as an
  artifact and reused by the rolling build.

## Compensations

Eight, each carrying a comment explaining why, an entry in
`fds-migration/LIBRARY-FEEDBACK.md`, and a testable removal condition.

| # | What | Why | Removal condition |
|---|---|---|---|
| 1 | Brand tokens overridden inline on `<nav>` (entries 1, 6) | `NavbarItem` paints its selected state from a fixed brand token, and the derived tint token is frozen against `:root`; OpenCTI's `theme_primary` is administrator-configurable and drafts switch the accent | `Navbar` gains an accent prop — pass the same value, delete the block; the token-name guard test passes unchanged |
| 2 | Submenu icons composed product-side (entry 2) | `asChild` — required for real links — makes the library's `icon`/`showIcon` inert, so `submenu_show_icons` had to be re-implemented | A slotted child keeps the library's icon handling — pass `icon`/`showIcon` again |
| 3 | Row body re-implemented, collapsed label kept `sr-only` (entry 7) | `asChild` replaces the library's own row content, including the hidden label the collapsed rail is navigated by | `asChild` composes the row body instead of replacing it |
| 4 | `open`/`onOpenChange` bound only while expanded (entry 10) | Collapsed, the same prop pair drives the hover flyout; controlling it made every flyout after the first unusable and wrote hover into the persisted menu state | The flyout gets its own state — bind unconditionally, hover three collapsed rows in turn, each opens and `selectedMenu` is unchanged |
| 5 | Sticky, definite-height geometry inline (entry 11) | The library's `<nav>` is laid out in flow and sized with a percentage height, which does not resolve against OpenCTI's shell: the rail was 776px tall in an 800px viewport and scrolled away with the page | The `<nav>` takes the height it is given — delete the `navStyle` geometry block, the rail still measures the full viewport and stays put while an inner container scrolls |
| 6 | Host rule `.app-navbar { flex: 0 0 auto }` (entry 8) | The rail does not defend its own width in a flex host | The `<nav>` carries `shrink-0` itself — delete the rule, the rail still measures 48/180px |
| 7 | Host rule raising portalled surfaces above the app bar (entry 5) | Flyouts and tooltips are portalled to `document.body` below OpenCTI's app bar | The library layers its portals above the host's own surfaces |
| 8 | `useFdsThemeScope` writes the theme classes on the document root | The library portals flyouts and tooltips to `document.body`; a subtree class leaves them unthemed. The hook's guard against targeting the root was amended, not bypassed | Never — this is the arbitrated mechanism |

**Conditional, not a compensation:** the legacy gradient is restored inline
**only** when an administrator configured a custom `theme_background`. Standard
installations get the real design-system rendering — the point of this pilot —
and custom-theme deployments lose nothing. The branch is commented as a
deliberate choice, not a workaround, and both cases were checked in the running
platform.

## Library observations filed

Twelve entries in `fds-migration/LIBRARY-FEEDBACK.md`, each with its removal
test. Three were added after hands-on passes on the running platform:

- **10 — accordion state and hover flyout state share one controlled prop.**
  Collapsed, `open`/`onOpenChange` drive the hover flyout; the 150ms delayed
  close of the row being left resolves against the same state snapshot as the
  next row's open, so the last one wins.
- **11 — the rail is laid out in flow and sized by percentage.** Both pilots had
  to write the same sticky, definite-height block in their host. Filed as an
  observation rather than a defect: it is arguably the host's responsibility,
  but two consumers out of two hit it.
- **12 — the pointer-cursor fix stopped at `NavbarItem`.** Library PR #84 fixed
  both of that component's render paths, which is 13 of this rail's 14
  button-rendered rows. The `ProductSwitcher` trigger declares no cursor of its
  own and still resolves to `default`. The ask is to generalise PR #84's own
  reasoning to every button the library renders.

Nothing in the library was modified by this pull request.

## Validation

Run locally on this exact commit: type-check clean, lint clean, **1337 unit
tests passing**. Two failures are **pre-existing** and reproduced identically
with this branch stashed:

- `src/utils/Time.test.ts` — timezone-dependent assertion, fails outside UTC.
- `src/buildConfig.test.ts` — requires `dist/index.html`, i.e. a prior build.

**41 unit tests** cover the new code and the amended theme hook, including the
behaviours most at risk of a silent regression: a submenu parent navigates
**only** while the rail is collapsed (as before); the selected-row accent is
still driven by the library token this code overrides — that guard reads the
installed stylesheet, because a rename upstream would otherwise lose the accent
with nothing turning red; and the rail's sticky, full-height geometry.

Continuous integration is green on this commit: **17/17 jobs** of `OpenCTI CI`
— frontend quality, backend, the three end-to-end groups, the client-python
matrix, the license check and the three container image builds, which is the
central proof here — plus `PR conventions checks` and `PR add labels`.
`PR check linked issue` is skipped by design on this base branch
(`pr-check-linked-issue.yml` only arms for `master` and `lts/*`). One red was
dated and re-run: `pycti test - python 3.12` failed once on a runner port
collision (`failed to bind host port … address already in use`), unrelated to a
front-end change, green on re-run.

Note for reviewers: `codeql-analysis.yml` and `cd-check-images-on-pr.yml`
restrict their trigger to `master`/`lts`, so they do not run here — a
pre-existing structural gap of this base branch, already recorded on the
OpenAEV side. Independent review is the only safety net for what they cover.

## How to test this PR

* Run the platform, then check both rail states: entries, permissions and
  destinations identical to `master`; Ctrl-click and middle-click open a new
  tab from any row; collapsed, hover moves from one submenu to the next and
  each flyout opens.
* Set a custom theme (`theme_background` + `theme_primary` in Settings →
  Customization): the rail keeps the configured background and the selected row
  the configured accent. Then switch back to a standard theme: the rail is
  painted by the design system.
* Toggle "Show submenu icons" in user preferences, in both rail states.
* `yarn test src/private/components/nav src/utils/hooks/useFdsThemeScope.test.tsx`
  — 41/41 passing.

## Not fixed here (mono-subject), dated in `fds-migration/IMPLEMENTATION-LOG.md`

- Two components read the collapsed-rail flag without subscribing to the toggle
  channel, so their offset goes stale. Pre-dates this change.
- `yarn i18n-checker` is declared in `package.json` but the script file does not
  exist and no job runs it.
- `bannerHeight` is built as the unitless string `'0'` when no banner is
  displayed, which makes any `calc()` containing it invalid — including the app
  shell's own `marginTop`. This rail derives its offsets from the numeric form
  instead; the shell's declaration is left alone.

### Checklist


sandyghs:
Update: I've bumped the design system pin on this PR, following feedback the OpenAEV team gave on the equivalent migration on their side. Since both products consume the same library, their review benefits this PR directly.

**What the feedback found.** Rows in the rail that render as buttons — the collapse toggle and the submenu triggers — didn't show the pointer cursor on hover, while rows rendering as links did. The browser gives links that cursor for free and gives buttons nothing, and the library ships no global reset, so nothing caught it. The old navigation, built on MUI, had it everywhere. Small, but a real iso-functionality regression: measured at 0 of 14 button rows before the bump, 13 of 14 after.

It was fixed in the library rather than patched here, so both products get it from the same place.

**Also in this bump**, since the pin moved forward: the ProductSwitcher now renders through the library's shared Menu. That changes its dropdown slightly — right padding 8→16px, minimum width 192→200px, separator 2→4px, group header 24→32px — with row height and label size unchanged. Nothing else moves: a set-diff of the compiled stylesheet between the two pins shows 641→689 class selectors and 414→420 custom properties, with zero removals on either side.

.github/actions/docker-build-platform/action.yml
```diff
@@ -50,6 +53,8 @@ runs:
         outputs: type=docker,dest=/tmp/opencti-platform${{ inputs.docker_image_suffix }}.tar
         tags: ${{ steps.meta.outputs.tags }}
         labels: ${{ steps.meta.outputs.labels }}
+        secrets: |
+          fds_git_token=${{ inputs.fds_git_token }}
```
natalie-natsu, line 57: What happens if you don't have the FDS_GIT_TOKEN? If a contributor wants to fork the project could he build it?

natalie-natsu (commented):
Lots of changes. It's nice to have tests. It will require another review on the components because for me most parts were ok, it is a lot of noise with these comments.

However, I can't approve anything before getting clear with the "FDS_GIT_TOKEN". How does it works with OpenAEV? I don't understand how we aren't soft locking projects with that. @aHenryJard have you seen this part?

.github/actions/docker-build-platform/action.yml
sandyghs: You're right to flag it — it's a real constraint today: the library is a private git dependency, so a fork can't build the front (in CI or locally). Two things frame it: no external contributor is affected yet, because fork PRs target master while this work lives on design-system/current; and the token was always an interim consumption mode. The structural fix is already planned — publishing the library to npm (workstream s5-release-machinery), which removes both the token and the private clone.

sandyghs:
> Lots of changes. It's nice to have tests. It will require another review on the components because for me most parts were ok, it is a lot of noise with these comments.

Fair point, and you were right to raise it. I've done a cleanup pass: each workaround is now a single line pointing at its numbered entry in fds-migration/LIBRARY-FEEDBACK.md, where the full context and the removal condition now live. These markers are temporary by design — each one gets deleted along with its workaround once the library ships the corresponding fix, and some of those fixes already landed this week. Nothing essential was lost, just the noise.

natalie-natsu (approved):
I accept only to unlock the PR, because this is the same action than a local merge commit. **In this current state, we can't approve any request other than POC.**

PR #17534, opened 2026-08-05
fix(public-dashboard): Improve ensurePublicContext in public-dashboards (#17533)

### LINKED ISSUE #17533: fix(public-dashboard): Improve ensurePublicContext in public-dashboards
## Description

Improvement of the function ensurePublicContext for public dashboards

## Environment

2. Version: { e.g. 1.2.3 }
3. Other environment details:

## Reproducible steps

1. { e.g. Run ... }
2. { e.g. Click ... }
3. { e.g. Error ... }

SarahBocognano:
### Proposed changes

* check if userInside of organization

### Related issues

* closes: https://github.com/OpenCTI-Platform/opencti/issues/17533

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/tests/01-unit/modules/publicDashboard/publicDashboard-distribution-unit-test.ts
+++ b/opencti-platform/opencti-graphql/tests/01-unit/modules/publicDashboard/publicDashboard-distribution-unit-test.ts
@@ -30,7 +30,10 @@ vi.mock('../../../../src/database/engine', () => ({
 
 vi.mock('../../../../src/database/redis', () => ({ notify: vi.fn() }));
 
-vi.mock('../../../../src/database/cache', () => ({ getEntitiesMapFromCache: vi.fn() }));
+vi.mock('../../../../src/database/cache', () => ({
+  getEntitiesMapFromCache: vi.fn(),
+  getEntityFromCache: vi.fn().mockResolvedValue({ platform_organization: null }),
+}));
 
 vi.mock('../../../../src/listener/UserActionListener', () => ({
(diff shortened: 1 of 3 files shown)


opencti-platform/opencti-graphql/tests/01-unit/modules/publicDashboard/publicDashboard-distribution-unit-test.ts
```diff
@@ -30,7 +30,10 @@ vi.mock('../../../../src/database/engine', () => ({
 
 vi.mock('../../../../src/database/redis', () => ({ notify: vi.fn() }));
 
-vi.mock('../../../../src/database/cache', () => ({ getEntitiesMapFromCache: vi.fn() }));
+vi.mock('../../../../src/database/cache', () => ({
```
gabrielebonadiman, line 33: One small question on this test update: since the mock returns `platform_organization: null` , `isUserInPlatformOrganization`  would return  true  regardless of the user's org membership.. just mentioning it because this test would pass even without the fix. Would it be possible to add a case with `platform_organization`  set and a user who isn't a member, to confirm `user_inside_platform_organization`  resolves to  false ?

SarahBocognano: I have added a dedicated unit test for isUserInPlatformOrganization covering a non-member user with platform_organization set, which resolves to false as expected. Kept it in a separate file since the distribution tests go through ensurePublicContext, which hardcodes the flag and can't exercise that logic anyway.

gabrielebonadiman (approved):

Issue #16059, opened 2026-05-19
feat(xtm-integration): connector lifecycle independence, release and versioning


alice-debra:
- Each connector can release on its own, when it's ready, with its own version
- Push the manifest in XTM Hub to have one source of truth

PR #17535, opened 2026-08-05
feat(merge-users):  dry-run/run framework, handler registry and coverage manifest (#17494)

### LINKED ISSUE #17494: PR2 — Merge engine: handler registry + dry-run/run framework + coverage manifest
### Objective

Build the merge engine framework that guarantees dry-run == real impact. Every category of work is a handler exposing dryRun() and run(), both routed through one shared computation — the dry-run is the run with the writes suppressed, not a second implementation of it. This PR ships the framework and the registry only — coverage grows as each later PR registers its handler.

### Out of scope

- Concrete handlers (each lands in its own PR).
- A "complete" dry-run — coverage is intentionally partial and grows PR by PR.
### Dependencies

- PR1 (mutation entry point).
### Definition of Done

- Engine tested via a mock handler proving discovery → dry pass → aggregation → coverage manifest, independently of real handlers (cf PoC#3 adaptivity test).
- Dry-run produces a report with covered vs uncovered sections and performs no writes.
- Replaying run() on an already-merged state is proven to be a no-op, including on a multiple:true field.
- Every handler is proven to route both modes through a single computation function — no handler re-implements its selection twice.
- A handler declaring a read on a field another handler writes is rejected at startup.
- A real pass launched against a state that changed since the dry pass fails instead of writing.
- The bulk wrapper fails on a forced version conflict instead of reporting success.
- An execution interrupted mid-run leaves a journal that names the last completed handler and the last completed user.
- Journal entries are readable from a separate session while the merge is still running.
- No handler rewrites the journal or the audit record.
- The register ships as a versioned constant covering every v2 row, and a run with no handler registered lists all of them as uncovered, by name.
- Coverage API returns uncovered rows by disposition and is unit-tested against a partial handler set — the actual MVP situation, where PR9/PR10/PR11 are missing.
- Reuses PoC#1 (footprint) and PoC#3 (skeleton).

ludovic:
### Proposed changes

* Adds the user reference register as a versioned constant: 101 rows, each with a stable id, a disposition, an entity and a field path. Row ids are never renumbered. The per-row analysis stays in the specification.
* Adds the `MergeHandler` contract. A handler exposes `compute()` (reads only) and `apply()` (writes only); the engine derives both execution modes from them, so dry-run is `compute` and the real run is `compute` then `apply`. A handler has no second code path that could drift from the first.
* Adds the handler registry. A handler declares the register rows it covers, the register version it was written against, and its read and write field sets. Registration rejects an unknown row id, a stale register version, a row already claimed by another handler, and a read set intersecting another handler's write set.
* Adds the two-pass engine. Every handler computes, the full report is produced, then every handler writes. Handlers are never interleaved, so a handler's computation always observes the same platform state in both modes. Before writing, each handler recomputes and its plan fingerprint is compared against the dry pass; a mismatch aborts without writing, since the platform is required to be at rest during a merge. Read/write disjointness is re-checked before each run, not only at registration. A failure returns a `FAILED` result carrying the merge id rather than throwing, because what was applied is only readable from the journal.
* Adds the coverage manifest, derived from the register rather than from the registered handlers, so an uncovered row is named rather than absent. `is_complete` is true only when every row is claimed. A disposition filter narrows the returned rows but never the counts. Exposed as `userMergeCoverage` and attached to every execution report.
* Adds the bulk write primitive: `conflicts: 'abort'`, `wait_for_completion: true`, `refresh: true`, throwing when `failures` is non-empty or `version_conflicts` is greater than zero, and returning the `updated` count.
* Adds the execution journal: one entry per handler and per pass, opened before the handler runs and closed with its outcome, so a handler that throws leaves a `FAILED` entry. Stored in Redis with a 30 day TTL, keyed per entry and indexed in a sorted set globally and per merge id, following the existing playbook-execution pattern.

No handler is registered in this chunk, so a merge still performs no write. Handlers land from #17495 onward.

### Related issues

* closes #17494 — part of #17258

### How to test this PR

Stacked on #17507, so review the last 7 commits only. Enable the flag and start the platform:

```bash
yarn start
```

With no handler registered, the coverage manifest must list all 100 register rows as uncovered, by name:

```bash
  -d '{"query":"{ userMergeCoverage { covered_count uncovered_count is_complete uncovered_rows { id disposition entity path } } }"}'
```

Filtering by disposition must narrow the returned rows while leaving the counts on the whole register:

```bash
  -d '{"query":"{ userMergeCoverage(disposition: TRANSFER) { covered_count uncovered_count uncovered_rows { id } } }"}'
```

Run a dry merge and read the journal back with the returned id:

```bash
  -d '{"query":"mutation { userMerge(sourceId:\"<source>\", targetId:\"<target>\", options:{dryRun:true}) { merge_id status report { coverage { is_complete } } } }"}'

  -d '{"query":"{ userMergeJournal(mergeId:\"<merge_id>\") { handler pass status updated } }"}'
```

The journal entries live in Redis and survive a platform restart for 30 days.

Automated coverage:

```bash
```

### Checklist

### Further comments

48 unit tests and 21 integration tests. Notable cases: with zero handler registered the manifest lists all 100 register rows as uncovered by name; a filtered manifest keeps its counts on the whole register; a real pass whose recomputation diverges from the dry pass writes nothing; per-disposition register counts are asserted as literals rather than recomputed from the array.

Commit by commit:

| Commit | Contents |
|---|---|
| `fc4278de` | user reference register, as a versioned constant (100 rows, v2) |
| `c8fa927b` | `MergeHandler` contract and handler registry |
| `5e539e76` | bulk write primitive |
| `68b90751` | execution journal |
| `422eeca9` | two-pass engine |
| `60558cc6` | coverage manifest and its API |
| `63ccabcb` | journal storage moved to Redis |

The precedent in the platform, `SanityOperation`, exposes `dryRun()` and `run()` as two methods each implementation writes, relying on both happening to call the same helper. That is a convention, and conventions drift the day someone fixes a bug in one branch and forgets the other. Here there is a single selection function, so "dry-run equals real impact" cannot drift.


opencti-platform/opencti-graphql/src/modules/userMerge/userMerge-engine.ts
```diff
@@ -1,55 +1,143 @@
...
+      handler: handler.identifier,
+    });
+  }
+  const updated = await handler.apply(handlerContext, plan);
+  return { ...plan, updated };
+};
+
 /**
- * Stubbed merge engine.
+ * Two full passes, never interleaved.
```
marieflorescontact, line 57: I don't understand the comment

opencti-platform/opencti-graphql/src/modules/userMerge/userMerge-engine.ts
```diff
@@ -1,55 +1,143 @@
...
- * honour, so the resolver and the domain guard-rails do not move when it lands.
+ * The dry pass is journalled too. An operator who lost the connection mid-dry-run still has
+ * a queryable trace of what was computed, and the two passes are distinguishable by dry_run.
  *
- * The placeholder deliberately reports FAILED rather than SUCCESS. Reporting success for
- * work that was never done is exactly the `cleanInconsistency` behaviour this feature is
- * specified against — an unimplemented engine must be impossible to mistake for a merge
- * that happened.
+ * No calling user is taken: whether a merge may be asked for at all is decided in the domain
+ * layer, and a merge rewrites references the caller has no reason to be allowed to read.
```
marieflorescontact, line 68: same, I don't understand  "the caller has no reason to be allowed to read"

marieflorescontact (commented):
I'm having trouble understanding the overall architecture of the User merge. There are a lot of new terms (handlers, registry, registers) I don't understand and the numerous comments are not really helping.  😅 I will need more time to review properly

opencti-platform/opencti-graphql/src/modules/userMerge/userMerge-engine.ts
ludovic: During a merge, we do two passes: first dry-run (read-only calculation presented to the admin), then real-run (writing). Right before writing, we re-run compute() to ensure the data hasn't changed in between (planFingerprint).

If the fingerprint changed, it means someone modified data on the platform between the dry-run review and the execution. If we proceeded anyway, we would write changes the operator never reviewed or approved. So instead of silently retrying or forcing the write, we abort immediately to ensure that 'what was reviewed in dry-run' strictly equals 'what actually gets written'.

opencti-platform/opencti-graphql/src/modules/userMerge/userMerge-engine.ts
ludovic: In OpenCTI, passing a user context applies that user's specific access rights and marking definitions.

However, the operator triggering the merge might not personally have access to every single private or restricted object created by the source user over the years. If the engine restricted its queries to what the calling operator is allowed to see, some references would be hidden, skipped, and left orphan after the merge.

Access authorization (BYPASS capability) is already strictly enforced in the domain layer (userMerge-domain.ts). Once authorized, the engine itself runs with platform-level access so it can migrate 100% of the user references across the entire database without blind spots.

ludovic:
Here is a quick breakdown to help you navigate the architecture:

**The Register** (userMerge-register.ts) = The Checklist A hardcoded, versioned constant of 100 items representing every single field across OpenCTI where a user ID exists (e.g. creator_id, sessions, history, etc.) along with what should happen to it (TRANSFER, INVALIDATE, RETAIN...). It lives in code so the engine knows what is not yet implemented.

**The Handlers** (userMerge-handler.ts) = The Workers Small domain-specific processors (e.g., STIX Objects handler, Redis session handler). A handler only has two functions: compute() (pure read/planning) and apply() (pure write). This prevents drift between dry-run and real-run.

**The Registry** (userMerge-registry.ts) = The Worker Directory In-memory registry loaded at startup. It validates that handlers cover valid register rows, don't overlap, and have no read/write conflicts.

**The Coverage Manifest** (userMerge-coverage.ts) = The Progress Meter Compares the 100 register rows against currently registered handlers. is_complete is true only when 100/100 rows are covered.

**The Journal** (userMerge-journal.ts) = The Flight Recorder Redis-based execution log (30-day TTL) tracking step-by-step progress per handler and merge ID so runs remain observable even if the HTTP connection dies.

Why PR1 & PR2 don't modify data yet: PR1 settled the GraphQL contract and Bypass permission. PR2 brings the two-pass engine and checklist framework. The actual migration logic for each entity type will arrive in small, bite-sized handler PRs (PR3 to PR11).

marieflorescontact:
I tested functionnaly the useMerge query. I got the response:

```
{
  "data": {
    "userMerge": {
      "id": "183055ab-86ae-4ac9-97fa-f38c1a28d848",
      "status": "SUCCESS",
      "dry_run": true,
      "report": {
        "merge_id": "183055ab-86ae-4ac9-97fa-f38c1a28d848",
        "registry_version": "v3",
        "total_updated": 0,
        "handlers": [],
        "coverage": {
          "is_complete": false,
          "total": 101
        }
      }
    }
  }
}
```


I just don't understand why coverage.total = 101 ? isn't it supposed to be equal to 100 ? well, maybe I am missing something 😅

ludovic:
> I tested functionnaly the useMerge query. I got the response:
> 
> { "data": { "userMerge": { "id": "183055ab-86ae-4ac9-97fa-f38c1a28d848", "status": "SUCCESS", "dry_run": true, "report": { "merge_id": "183055ab-86ae-4ac9-97fa-f38c1a28d848", "registry_version": "v3", "total_updated": 0, "handlers": [], "coverage": { "is_complete": false, "total": 101 } } } } }
> 
> I just don't understand why coverage.total = 101 ? isn't it supposed to be equal to 100 ? well, maybe I am missing something 😅

It is indeed 101; the register has changed, and I had forgotten to update the PR description.

ludovic:
### Mental Model & Architecture Overview

To help review this PR and understand the foundations of the User Merge feature, here is a breakdown of the 5 core concepts and the execution model:

#### 1. Core Concepts at a Glance

| Concept | File | Real-world Analogy | Purpose |
| :--- | :--- | :--- | :--- |
| **The Register** | `userMerge-register.ts` | **The Checklist** | An exhaustive, versioned constant of 99 rows listing every single place in OpenCTI referencing a user ID (e.g. `creator_id`, relations, sessions, filters). It describes the codebase, not what is implemented yet. |
| **The Handlers** | `userMerge-handler.ts` | **The Workers** | Domain-specific processors (e.g. Activity, Groups, Relationships). Each handler strictly exposes `compute()` (read-only) and `apply()` (write-only). |
| **The Registry** | `userMerge-registry.ts` | **The Directory** | Boot-time validator ensuring handlers match the current register version, don't claim duplicate rows, and don't read fields that another handler writes. |
| **The Coverage** | `userMerge-coverage.ts` | **The Progress Meter** | Compares registered handlers against the Register. `is_complete` becomes true only when 100% of the gating rows (`transfer` / `conditional`) are claimed. |
| **The Journal** | `userMerge-journal.ts` | **The Flight Recorder** | Redis-backed execution log (30-day TTL) tracking progress per handler and pass so runs remain observable even if the connection drops. |

---

#### 2. Execution Flow (Two-Pass Engine)

A merge is executed in two non-interleaved passes to guarantee that **what is reviewed in dry-run strictly equals what is written in real mode**.

```mermaid
sequenceDiagram
    autonumber
    actor Admin as Operator / Admin
    participant API as GraphQL / Domain Layer
    participant Engine as Merge Engine
    participant Handlers as Merge Handlers
    participant Redis as Redis Journal
    participant DB as OpenSearch / Redis Live

    Admin->>API: mutation userMerge(source, target, dryRun)
    API->>API: Verify BYPASS capability & Feature Flag
    API->>Engine: executeUserMerge()
    Engine->>Engine: Validate handlers are read/write disjoint

    rect rgb(240, 245, 255)
    Note over Engine,Handlers: PASS 1 — Dry-Run (Read-only planning)
    loop For each Handler
        Engine->>Redis: Open journal entry (RUNNING, dryRun=true)
        Engine->>Handlers: compute() [reads current platform state]
        Handlers-->>Engine: UserMergeHandlerPlan
        Engine->>Redis: Close journal entry (SUCCESS)
    end
    end

    alt options.dryRun == true
        Engine-->>Admin: Return complete Report & Coverage (0 writes)
    else options.dryRun == false
        rect rgb(255, 245, 240)
        Note over Engine,DB: PASS 2 — Real Run (Verification & Apply)
        loop For each Handler
            Engine->>Handlers: compute() & verify fingerprint == Pass 1
            alt Fingerprint diverged (platform moved)
                Engine-->>Admin: Abort immediately (nothing written)
            else Fingerprint matches
                Engine->>Redis: Open journal entry (RUNNING, dryRun=false)
                Engine->>Handlers: apply(plan)
                Handlers->>DB: Bulk write (conflicts: abort)
                Engine->>Redis: Close journal entry (SUCCESS, updated count)
            end
        end
        end
        Engine-->>Admin: Return Final Report + updated counts
    end
  ```
   
  #### 3. Key Safety Guarantees

* **Zero Drift between Dry-Run & Run (`compute` / `apply`)** : Unlike `SanityOperation` (where `dryRun()` and `run()` are two independent functions that can drift over time), handlers have a **single selection function** (`compute`). Dry-run executes `compute()`; real-run executes `compute()` then `apply(plan)`.
* **Divergence Guard (`planFingerprint`)** : Before any handler writes, its plan is recomputed. If data on the platform changed between the operator's dry-run review and the real execution, the engine aborts to prevent unapproved changes.
* **Read/Write Disjointness** : Validated at boot time. No handler can read a field that another handler writes, preventing execution order dependencies.
* **Strict Bulk Writes** : Uses `conflicts: 'abort'` and fails on any version conflict or partial failure, rather than silently skipping documents.

#### 4. Scope of this PR (Why are no handlers registered yet?)

This PR delivers the **scaffolding, contracts, and safety engine** only. With 0 handlers registered:
* The coverage manifest correctly exposes all 99 rows as uncovered (`is_complete: false`).
* A merge can be triggered safely and logs to the Redis journal, but writes 0 documents.
* Specific handlers (STIX objects, user preferences, groups, activities, etc.) land in subsequent isolated PRs (#17495 onward) where each domain's migration logic can be reviewed independently.

File: opencti-platform/opencti-graphql/src/modules/userMerge/userMerge-register.ts [hunk 3]
```diff
@@ -0,0 +1,170 @@
...
+  /** Outside the reach of a database merge (in-flight queue messages, opaque state). */
+  OutOfScope = 'out-of-scope',
+}
+
+export interface UserMergeRegisterRow {
+  /** Stable identifier a handler declares coverage against. Never reused, never renumbered. */
+  id: string;
+  disposition: UserMergeDisposition;
+  /** Entity or store the reference lives on. */
+  entity: string;
```
fellowseb: nit: This seems more like a description to me. Naming the field `entity` makes me want to store either an ID or type in that string.
Examples: `SSO session / refresh token`, `Edit context`...

opencti-platform/opencti-graphql/src/modules/userMerge/userMerge-register.ts
```diff
@@ -0,0 +1,170 @@
...
+ * no handler covers unless it holds the full list independently of the handlers. Without
+ * it, a report can only show what is done and never what is missing: the blind spots go
+ * invisible exactly when they are most numerous.
+ *
+ * Only what the engine needs at runtime is transcribed here — id, disposition, entity type,
+ * path, short label. The per-row analysis stays in the specification, which is the record of
+ * the reasoning that produced the list, not a document to keep in sync at runtime.
+ */
+
+export const USER_MERGE_REGISTRY_VERSION = 'v3';
```
fellowseb: Why are we already at v3 ?

fellowseb: Should it be `USER_MERGE_REGISTRY_VERSION` or `USER_MERGE_REGISTER_VERSION` ?

opencti-platform/opencti-graphql/src/modules/userMerge/userMerge-registry.ts
```diff
@@ -0,0 +1,89 @@
...
+import type { UserMergeHandler } from './userMerge-handler';
+import { findRegisterRow, USER_MERGE_REGISTRY_VERSION } from './userMerge-register';
+
+const HANDLERS: UserMergeHandler[] = [];
+
+const assertCoverageIsValid = (handler: UserMergeHandler): void => {
+  // Drift on the version: a row can be requalified — retain becoming transfer, say — without
+  // its id changing, so id existence alone is not enough. The v1 -> v2 revision requalified
+  // 12 rows without changing a single id.
+  if (handler.registryVersion !== USER_MERGE_REGISTRY_VERSION) {
```
fellowseb: How can these versions be different ? Are handlers persisted somewhere ?

opencti-platform/opencti-graphql/src/modules/userMerge/userMerge-registry.ts
```diff
@@ -0,0 +1,89 @@
...
+      if (intersection.length > 0) {
+        throw UnsupportedError('Merge handlers are not read/write disjoint', {
+          reader: reader.identifier,
+          writer: writer.identifier,
+          fields: intersection,
+        });
+      }
+    });
+  });
+};
```
fellowseb, line 41: All handlers are known statically right ?
So this check could (and should) be done once; not when executing a user merge.
Also this runs in O(N^2) loop iterations with N being the number of handlers, so executing it at every handler registration seems not super performant.

Instead we could do it in an init function called during platform init. (And put handlers registrations there too while at it). WDYT ?

Maybe I'm missing something, i'm still trying to grasp the entire PR stack.

opencti-platform/opencti-graphql/src/modules/userMerge/userMerge-registry.ts
```diff
@@ -0,0 +1,89 @@
...
+      if (existing) {
+        throw UnsupportedError('Two merge handlers claim the same register row', {
+          row_id: rowId,
+          handlers: [existing, handler.identifier],
+        });
+      }
+      claimedBy.set(rowId, handler.identifier);
+    });
+  });
+};
```
fellowseb, line 57: This check also could be done in an "init" instead of doing it at each registration

opencti-platform/opencti-graphql/src/database/redis.ts
```diff
@@ -1057,3 +1057,35 @@ export const redisDeleteXtmAgentResponse = async (cacheKey: string): Promise<voi
...
+  // mutation also lost the only way to name the run they need to follow.
+  await setKeyWithList(key, [USER_MERGE_JOURNAL_LIST, userMergeJournalList(mergeId)], entry, USER_MERGE_JOURNAL_TTL);
+  return entry;
+};
+
+export const redisUserMergeJournalRead = async (mergeId?: string) => {
+  const listId = mergeId ? userMergeJournalList(mergeId) : USER_MERGE_JOURNAL_LIST;
+  return keysFromList(listId, USER_MERGE_JOURNAL_TTL);
+};
+// endregion - user merge journal
```
fellowseb, line 1091: I know it's not yet a widely-approved convention but how about moving all this to a `userMerge-redis.ts` file in the module ? Way better for locality of reasoning & modularity.

opencti-platform/opencti-graphql/src/modules/userMerge/userMerge-coverage.ts
```diff
@@ -0,0 +1,146 @@
...
+
+/**
+ * Dispositions that a handler has to answer for.
+ *
+ * `retain` and `out-of-scope` are excluded by construction, not by tolerance: a retained row
+ * is one the merge deliberately does not rewrite, an out-of-scope one is unreachable. No
+ * handler will ever claim them, so counting them in would leave the deletion gate shut for
+ * good — including once every remaining transfer is implemented.
+ */
+const GATING_DISPOSITIONS = [UserMergeDisposition.Transfer, UserMergeDisposition.Conditional];
```
fellowseb, line 46: Shouldn't this array contain `UserMergeDisposition.Invalidate` too ?

opencti-platform/opencti-graphql/src/modules/userMerge/userMerge-journal.ts
```diff
@@ -0,0 +1,92 @@
...
+ * RUNNING that is indistinguishable from a node that died.
+ */
+export const withJournalEntry = async <T extends UserMergeHandlerOutcome>(
+  input: JournalEntryInput,
+  execute: () => Promise<T>,
+): Promise<T> => {
+  const entryId = await openJournalEntry(input);
+  try {
+    const outcome = await execute();
+    await closeJournalEntry(entryId, input.mergeId, { status: UserMergeStatus.Success, outcome });
```
fellowseb: If this `closeJournalEntry` call throws we'll try to mark the journal entry as `FAILED`; which doesn't really makes sense.

IMO this `closeJournalEntry` should be taken out of this try/catch block and attempted in a separate try/catch; and maybe we shouldn't rethrow when we failed to write to the journal ?

opencti-platform/opencti-graphql/src/modules/userMerge/userMerge.graphql
```diff
@@ -28,6 +89,8 @@ type UserMergeResult {
     started_at: DateTime!
     completed_at: DateTime
     message: String
+    # Absent when the execution failed before any handler ran.
+    report: UserMergeExecutionReport
 }
 
 type UserMergeJournalEntry {
```
fellowseb, line 94: Don't we want to expose the `output` (or `outcome` structure) to API consumers ? Including the updated_count

opencti-platform/opencti-graphql/src/modules/userMerge/userMerge-registry.ts
fellowseb: After reading further I don't see the point of `registryVersion` nor `USER_MERGE_REGISTRY_VERSION` at all given  all handlers are compiled in the same repository and statically known (& registered at init soon, see my other comment). My feeling is that we can remove this.

fellowseb:
I must say, I don't understand the choice of using Redis for the journal entries instead of using a new `ES` entity:
- The intended (long) durability period points towards using ES IMHO
- For me the "mapping budget" argument isn't really a problem: if we don't need to index/search on most/many of the journal entry fields we could use the `enabled: false` prop on object fields or `index: false` on fields of other types. (I've introduced a new `object` format (`raw`) to model the non-indexed object field in https://github.com/OpenCTI-Platform/opencti/pull/17769 BTW.
- The new entity could be excluded from all `userMerge` logic explicitly, especially in PR3.
- Also no need to make `source_user_id` and `target_user_id` fields ordinary user-reference fields. It's an audit log entry so we could keep them as strings.

Happy to discuss further pros & cons !

opencti-platform/opencti-graphql/src/modules/userMerge/userMerge-register.ts
ludovic: Fair point. The column is really a human-readable label for where the reference lives, and for several rows it isn't an entity type at all (SSO session / refresh token, accesses-to). Nothing reads it programmatically (handlers match on id) so renaming it to label costs nothing and stops it from promising a type it doesn't hold. I'll rename it. 👍

opencti-platform/opencti-graphql/src/modules/userMerge/userMerge-register.ts
ludovic: The register was revised twice while the specification was being reviewed, before any of this shipped: 
v1 → v2 requalified 12 rows (mostly retain that turned out to be transfer), 
v2 → v3 added the rows found in the Redis and session stores. 
Later chunks in the stack take it to v5. 
The number tracks the analysis, not releases.

ludovic: You're right, it's inconsistent: the constant versions USER_MERGE_REGISTER, while registry is the separate handler registry in userMerge-registry.ts

opencti-platform/opencti-graphql/src/modules/userMerge/userMerge-registry.ts
ludovic: They can't. Nothing is persisted; every handler is compiled here and every one of the 13 sets registryVersion: USER_MERGE_REGISTRY_VERSION, the same constant the check compares against, so the condition is never true. 
The intent was to catch a handler written against an older register whose rows were requalified without changing ids, but a recompile updates everyone at once, so it never fires. 
I'll drop it.

opencti-platform/opencti-graphql/src/database/redis.ts
ludovic: No objection from me, and I agree it reads better next to the code that uses it. I put it in redis.ts to follow what the file already does for every other Redis-backed feature. Since this cuts across the codebase rather than just this module, I'd rather not set the precedent alone; if the team is fine with the new convention, I'll move it.

opencti-platform/opencti-graphql/src/modules/userMerge/userMerge-journal.ts
ludovic: You're right, and it's worse than confusing: if the success write to Redis fails, the catch marks the handler FAILED even though it succeeded and already wrote to the database, then rethrows and aborts the merge. The operator ends up reading a journal that contradicts what is actually stored, at the exact moment they need it to be true.

I'll take the closing write out of the try, wrap it separately, and log without rethrowing when only the journal write failed; losing the trace of a successful handler is bad, but reporting it as failed is worse.

opencti-platform/opencti-graphql/src/modules/userMerge/userMerge-registry.ts
ludovic: You're right, running these per registration isn't useful; validating the handler set once is better. An init function (registerUserMergeHandlers()) actually lands in a later PR of the stack, so it isn't visible from here. I'll pull it into this PR and move the checks into it, along with dropping the redundant call in the engine.

opencti-platform/opencti-graphql/src/modules/userMerge/userMerge-coverage.ts
ludovic: No, and the doc comment right above should have said why; it only justifies retain and out-of-scope. An invalidate row is one whose reference dies with the source account: sessions, tokens, OTP secrets, cached rights, edit contexts. Deleting the user removes them, so no handler will ever claim them, and counting them in would keep the gate shut for good. I'll extend the comment to cover them.

opencti-platform/opencti-graphql/src/modules/userMerge/userMerge.graphql
ludovic: Yes, good catch. The journal is the only durable trace of a merge, so it should say what each handler actually did. I'll expose updated_count and the outcome.

ludovic:
> I must say, I don't understand the choice of using Redis for the journal entries instead of using a new `ES` entity:
> 
> * The intended (long) durability period points towards using ES IMHO
> * For me the "mapping budget" argument isn't really a problem: if we don't need to index/search on most/many of the journal entry fields we could use the `enabled: false` prop on object fields or `index: false` on fields of other types. (I've introduced a new `object` format (`raw`) to model the non-indexed object field in [feat(catalog): introduce new delivery mechanism for the catalog (#16059) #17769](https://github.com/OpenCTI-Platform/opencti/pull/17769) BTW.
> * The new entity could be excluded from all `userMerge` logic explicitly, especially in PR3.
> * Also no need to make `source_user_id` and `target_user_id` fields ordinary user-reference fields. It's an audit log entry so we could keep them as strings.
> 
> Happy to discuss further pros & cons !

The storage choice was settled with Angélique on two grounds: the mapping budget, and the fact that this is meant to run once per platform. Your index: false suggestion does weaken the first one; I'll grant you that. The second still holds for me: a new entity type is permanent platform surface for a one-shot operation.

Worth noting too that the merge itself doesn't only live in Redis: on completion it publishes an administration user action, so it's recorded in the platform activity log (in ES, with its own retention) carrying the merge id, source, target and strategy. That's the record answering "what happened to this account" months later. The Redis journal is the per-handler detail used to follow a run and post-mortem a failure, which is why it's TTL'd rather than kept.

There's also a practical cost. The journal isn't self-contained: a new entity carrying source_user_id needs its own row in the reference register, which means it reaches into the coverage manifest and the deletion gate further up the stack. That wouldn't be a local change in this PR.

The good news is that nothing here is a one-way door; the store sits behind userMergeJournal / UserMergeJournalEntry, so swapping it later is invisible to API consumers. I'd rather ship on Redis and treat the ES entity as a follow-up if the retention proves too short in practice. Does that work for you?

opencti-platform/opencti-graphql/src/modules/index.ts
```diff
@@ -174,3 +174,9 @@ import './userMerge/userMerge-graphql';
 import './workflow/api/workflow-graphql';
 import './customField/custom-field-graphql';
 // endregion
+
+// The handlers derive part of their targets from the schema, so they can only be registered
+// once every module above has declared its attributes.
+import { registerUserMergeHandlers } from './userMerge/userMerge-handlers';
+
+registerUserMergeHandlers();
```
fellowseb: Nit: I think we can follow the same "import with side-effect" pattern but with a new concept: the static init of a module:

Here we could import a `userMerge-init.ts` ES-module:
```
// region modules static init
import './userMerge/userMerge-init';
// endregion
```

And move the code here in the new init file.
That way we start a "clean" pattern.

nb: not a big fan of "import with side-effect" pattern, we might want to review this later. This way we could make sure the boot-order is correctly applied (Entities, then GraphQL schema, then static init)

opencti-platform/opencti-graphql/src/modules/userMerge/userMerge-register.ts
```diff
@@ -0,0 +1,170 @@
...
+ * no handler covers unless it holds the full list independently of the handlers. Without
+ * it, a report can only show what is done and never what is missing: the blind spots go
+ * invisible exactly when they are most numerous.
+ *
+ * Only what the engine needs at runtime is transcribed here — id, disposition, label, path.
+ * The per-row analysis stays in the specification, which is the record of the reasoning that
+ * produced the list, not a document to keep in sync at runtime.
+ */
+
+export const USER_MERGE_REGISTER_VERSION = 'v3';
```
fellowseb: Still needs to be removed I believe along with the usages (`register_version`). If we want to keep a version in the journal we could instead export the platform version.

opencti-platform/opencti-graphql/src/modules/index.ts
ludovic: Agreed, I'll move it to `userMerge-init.ts` under a `modules static init` region.

opencti-platform/opencti-graphql/src/modules/userMerge/userMerge-register.ts
ludovic: Agreed, removing it with both `register_version` fields. Nothing to replace on the journal
side: `UserMergeJournalRecord` carries no version today.

fellowseb (approved):
All tests put in the PR description pass (although some are ou of sync) 👍 .

Thank you for the explanation on the choice of Redis. Maybe, if needed, we could export the journal to the file storage in a future iteration; but ok 👍 .

One other general comment though:
I have the feeling that the coverage endpoint is a project/dev progress tracking API. Nothing that really is needed at runtime.
What will be the use case for this API ? Again I feel like the output data could be known at build-time. Instead it could be a build artefact or a documentation page udpated automatically.

PR #17549, opened 2026-08-06
feat(worker): Proposal D - add batched STIX object import for dependency-free entities (#85)

### LINKED ISSUE #85: Provide support for more Observable types
## Problem to Solve

The platform currently offers a limited set of observables that are commonly seen in incidents. The platform should offer more observable types which can provide more detailed information when describing incidents.

## Current Workaround

There is no current workaround. You can only use what the platform currently provides.

## Proposed Solution

I would like to see the following Observable types added to the platform, both as an option when adding a specific Observable and checkbox items when filtering a list of observables. An option could also include the ability for a user to specify new Observable types on their own for greater platform flexibility. Use the Mandiant APT1 report STIX json file as an example of some additional available Observable objects: https://oasis-open.github.io/cti-documentation/examples/example_json/apt1.json

File Path
File Name
Registry Key
PDB Path
Service
Scheduled Task
x509-certificate:issuer And x509-certificate:serial_number

## Additional Information

None

gabrielebonadiman:
### Proposed changes

* New `stixObjectsBatchImport(items)` GraphQL mutation: N dependency-free STIX objects created via one call instead of N. Scope limited to 4 leaf kinds (`stix_cyber_observable`, `malware`, `vulnerability`, `tool`), each still dispatched to the existing single-object domain functions (`addMalware`, `addStixCyberObservable`, `addVulnerability`, `addTool`) — no shared lock, no ES `_bulk` write, no ordering resolution. Deferred to v2.
* Worker: opt-in accumulate-window batching (`batch_import_enabled`/`batch_import_size`/`batch_import_timeout`), gated by a structural pre-filter (`is_batchable`/`_has_only_free_refs`) that only admits messages with no intra-batch dependency. Per-item ack/nack/requeue and work-expectation reporting preserved.
* client-python: thread-safe capture mode on `OpenCTIApiClient` (dry-runs the existing `import_item()` path to build the batch payload, no duplicated conversion logic).
* Telemetry: `opencti_worker_batch_size`, `opencti_worker_batch_items_total`, `opencti_worker_batch_fallback_total`.
* Feature-flagged off by default. Falls back to individual mutations for anything unsupported.

### Related issues
* closes #85

### How to test this PR

1. Set `worker.batch_import_enabled: true` in the worker config.
2. Import a report/bundle containing observables, malware, vulnerabilities, or tools.
3. Confirm a single `stixObjectsBatchImport` call handles the batchable items (check worker logs / `opencti_worker_batch_size` metric), non-batchable items still go through individual mutations.
4. `yarn vitest run --config vitest.config.ci-unit.ts tests/01-unit/domain/stixBatchImport-test.ts` (backend)
5. `pytest tests/test_push_handler_batching.py` (worker)

### Checklist

### Further comments

v1 is intentionally narrow: same domain functions, same locking, same ES writes as today — the only change is 1 network round-trip instead of N. Benchmarked live on a 78K-message report (batch on/off, same report re-imported): 76% wall-time reduction (4.2x), 100% DB-content parity. The win is entirely round-trip elimination, not lock/bulk-write savings — those aren't built yet (v2). Known gaps not yet hardened: no same-batch duplicate-key dedup (currently moot, batching is limited to non-referencing kinds), no retry/backoff for a failed batched item (reported + acked, not retried), and a bundle with no `x_opencti_seq` is currently trusted the same as `x_opencti_seq == 1`.


gabrielebonadiman:
Closing this, after further discussion with the team, we've decided not to pursue batching worker→GraphQL mutations in this form.

STIX bundles carry shared context (draft ID, applicant ID, playbook ID) that needs to stay tied to each object during processing. Grouping independent objects into a single batched mutation, as done here, breaks that association.

The performance data from this PR is genuinely useful and will inform future ingestion work.

PR #17573, opened 2026-08-06
fix(worker): prevent message queue consumer from stalling on unhandled exceptions (#17553)

### LINKED ISSUE #17553: fix(import-export): STIX import/export stalled on worker uncatch errors
## Description
STIX import broken on our internal staging environment. Root cause found via log/metric analysis.

A connector (likely MISP) sends malformed observables with only `vendor` set — no `name`/`cpe`/`swid`. Backend correctly rejects them (`syntax.js` → `FunctionalError: INCORRECT_OBSERVABLE_FORMAT`, pre-existing, not a regression). This one issue accounts for the vast majority of the worker's logged errors (`Unhandled exception` + `Unrecognized error during bundle import`), dwarfing everything else.

That connector's push queue is stuck and growing, while all other queues drain fine. Hypothesis: the poison message crashes the consumer thread, it requeues unacked, the worker's health-check restarts it, it re-crashes on the same message — a stall loop, not a slow queue.

PRs #17178 and #17451 (recent, temporally close to report) were checked and ruled out.

## Environment
- Internal staging environment
- One connector's dedicated RabbitMQ queue affected; others unaffected

## Steps to reproduce
1. Connector sends a `Software` observable with only `vendor` set
2. Backend rejects it (`INCORRECT_OBSERVABLE_FORMAT`)
3. Worker logs error; queue's consumer thread appears to crash instead of cleanly skipping the message

## Expected
Bad observable is skipped/rejected without affecting queue throughput.

## Actual
Queue backlog grows unbounded.

## Fix
1. Connector: always send `name` for Software observables
2. Worker: make consumer resilient to permanent validation errors — ack/skip, never crash the thread

## Notes
- Minor unrelated finding: `opencti_logger.py`'s `error()` always sets `exc_info=1`, can mislabel tracebacks in logs
- `stix-2-1-converter.ts` has no unit tests (unrelated gap, separate follow-up)

gabrielebonadiman:
### Proposed changes

* Fix `MessageQueueConsumer.consume_message` so a message is always ack'd or nack'd, even if `handle_message` raises an unexpected exception — previously such a message stayed unacknowledged and, under `prefetch_count=1`, silently stalled that queue's consumer forever.
* Harden `PushHandler.handle_message`'s `finally` block so a failure recording telemetry (counter/gauge) can no longer override the try/except's `ack`/`nack` return value and escape as an unhandled exception.

### Related issues
* closes #17553

### How to test this PR
* Feed the worker's push queue a message whose processing forces an unexpected exception past `handle_message`'s own error handling (e.g. mock `bundles_processing_time_gauge.record` to raise) and confirm the message is ack'd and the queue keeps consuming subsequent messages instead of stalling.
* Confirm normal `ack`/`nack`/`requeue` behavior for existing success/failure paths is unchanged.

### Checklist

### Further comments

Root-caused via Grafana log/metric analysis on the `testing` environment: a connector was sending malformed `Software` observables (missing `name`/`cpe`/`swid`), which the backend correctly rejects. That in itself was handled gracefully, but a separate latent bug meant any message whose processing raised an exception outside the handler's own try/except was left unacknowledged rather than ack'd or nack'd. Combined with `prefetch_count=1`, this permanently blocked that connector's queue from receiving further messages, causing an unbounded backlog while other queues drained normally. This PR closes that gap; a follow-up on the connector-side data mapping is tracked separately.


opencti-worker/src/message_queue_consumer.py
```diff
@@ -44,7 +44,14 @@ def ack_message(self, delivery_tag: int) -> None:
             )
 
     def consume_message(self, delivery_tag: int, body: str) -> None:
-        result = self.handle_message(body)
+        try:
+            result = self.handle_message(body)
```
gabrielebonadiman, line 48: `handle_message` is expected to catch its own errors and return "ack"/"nack"/"requeue". If something still escapes here, we must not leave the message unacknowledged: with prefetch_count=1 that would silently stall this queue's consumer forever. Ack it so the queue keeps flowing, and log loudly so the failure is not lost.

opencti-worker/src/push_handler.py
```diff
@@ -265,6 +265,12 @@ def handle_message(
             # Nack message and discard
             return "nack"
         finally:
-            self.bundles_global_counter.add(len(imported_items))
-            processing_delta = datetime.datetime.now() - start_processing
-            self.bundles_processing_time_gauge.record(processing_delta.seconds)
+            try:
```
gabrielebonadiman, line 268: Telemetry recording must never override the ack/nack decision above: if it raised here, the try/except's return value would be silently discarded and the exception would propagate instead, leaving the message unacknowledged.

opencti-worker/src/message_queue_consumer.py
```diff
@@ -44,7 +44,14 @@ def ack_message(self, delivery_tag: int) -> None:
...
+        except Exception as e:
+            self.logger.error(
+                "Unhandled exception while processing message, acking to avoid stalling the queue",
+                {"exception": e},
+            )
+            result = "ack"
```
aHenryJard, line 57: Why not requeue for some retry on error ? For example if an opencti node is not available it will throw an exception by maybe the message can be process later no ?

gabrielebonadiman: I thought about this, and I'd rather not introduce a more complex retry mechanism for something that only caused a stall and never actually broke the system otherwise (maybe I've been a bit too optimistic but 😄  ) that's how it was behaving before this bug, too.

Requeuing is also what worries me most: if the error is deterministic (not transient), we'd redeliver the same message forever, and the queue would look like it's moving while that slot never actually drains, which is effectively the same stall, just in a different shape.

I'd suggest we treat retry logic as a separate follow-up, once we have proper dashboards in place to observe when and why this error actually occurs. If it turns out to consistently be caused by things like `bad naming`/`malformed data`, retrying wouldn't help anyway so it's better to base that decision on real data rather than guessing now.

opencti-worker/tests/test_message_queue_consumer_resilience.py
```diff
@@ -0,0 +1,76 @@
+"""Unit tests for MessageQueueConsumer.consume_message's resilience to
```
aHenryJard, line 1: Thanks for testing coverage ! I don't think that worker tests are run on CI but it can be done in a follow up PR to get this one merged faster.

aHenryJard (approved):
Tested locally on the normal path works as expected, and also tested with tests.
All good to me, the 2 remaining copilot comment are either really edges cases (like logger not working) or existing code that moved.

The only think is that the CI does not run worker tests.

gabrielebonadiman:
@aHenryJard shall we open a new issue for adding worker test in the CI?

PR #17575, opened 2026-08-06
fix(xtm-hub): keep connectivity check alive when SMTP email fails (#15681)

### LINKED ISSUE #15681: fix: checkXTMHubConnectivity crashes when SMTP email is not sent, causing infinite loading on the frontend

shuvonsec:
## Summary

- Backend: wrap `sendAdministratorsLostConnectivityEmail` in try/catch inside `handleLostConnectivityEmail` so a SMTP failure no longer aborts `checkXTMHubConnectivity`. The email flag stays `true` so the next check can retry.
- Frontend: handle `onError` for the connectivity mutation in `XtmHubSettings` — show an error Alert instead of infinite loading.
- Test: add coverage that connectivity status still updates when email sending rejects.

Fixes #15681

## Type

- [x] Bug fix

## How to test

1. **Backend unit/integration**: run the XTM hub connectivity suite and confirm the new case passes:
   ```bash
   yarn test tests/03-integration/01-database/xtm-hub-test.ts
   ```
2. **Reproduce frontend infinite load (before) / error Alert (after)**:
   - Register a platform with XTM Hub (or mock `checkXTMHubConnectivity` to throw).
   - Force SMTP failure (bad SMTP config, or temporarily throw inside `sendAdministratorsLostConnectivityEmail`) while lost-connectivity email conditions are met (inactive hub + 24h since last check + email flag true).
   - Open **Settings → Filigran Experience → XTM Hub**.
   - Expect: page renders with an error Alert (no infinite Loader); GraphQL check still returns registration status when only the email step fails.
3. **Happy path**: with working SMTP / no email needed, XTM Hub card still loads normally after the connectivity check.

## Related Issue

Closes #15681

Made with [Cursor](https://cursor.com)

### DIFF
--- a/opencti-platform/opencti-front/lang/front/en.json
+++ b/opencti-platform/opencti-front/lang/front/en.json
@@ -4891,6 +4891,7 @@
   "Types of the targeting": "Types of the targeting",
   "Types of the targets": "Types of the targets",
   "Unable to validate remote OpenCTI URL": "Unable to validate remote OpenCTI URL",
+  "Unable to check XTM Hub connectivity. Please try again later.": "Unable to check XTM Hub connectivity. Please try again later.",
   "Unauthorized action, please refresh your browser": "Unauthorized action, please refresh your browser",
   "Unfix the nodes and re-apply forces": "Unfix the nodes and re-apply forces",
   "Unique identifier": "Unique identifier",
--- a/opencti-platform/opencti-front/src/private/components/settings/xtm-hub/XtmHubSettings.tsx
+++ b/opencti-platform/opencti-front/src/private/components/settings/xtm-hub/XtmHubSettings.tsx
@@ -1,6 +1,7 @@
 import { graphql, useLazyLoadQuery } from 'react-relay';
 import React, { useEffect, useState, useContext } from 'react';
 import { HubOutlined, MapOutlined, RocketLaunchOutlined, VideoLibraryOutlined, WidgetsOutlined } from '@mui/icons-material';
+import Alert from '@mui/material/Alert';
 import Typography from '@mui/material/Typography';
 import XtmHubTab from '@components/settings/xtm-hub/XtmHubTab';
 import Button from '@common/button/Button';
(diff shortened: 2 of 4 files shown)


shuvonsec:
Updated the connectivity-check error Alert copy to remove the SMTP mention (`fffbe7a02`) — SMTP failures are handled server-side and no longer fail the mutation, so the frontend message is now a generic retry prompt.

CelineSebe:
Thanks a lot for this fix, @shuvonsec, nice catch on isolating the SMTP failure so it no longer blocks the connectivity check, and good job adding backend test coverage for that regression case! 🙏

One small, non-blocking suggestion: the backend now has a test for the SMTP failure case, but there doesn't seem to be an equivalent frontend test verifying that the error Alert is displayed when commitCheckConnectivity fails (the onError branch in XtmHubSettings). Might be worth adding for symmetry, but not required to merge this fix.

Also FYI, the failing "Frontend e2e / e2e group0 tests" check looks unrelated to this change (fails on navigation.spec.ts / Infrastructures page visibility, not on XTM Hub), re-running CI to confirm it's flaky.

opencti-platform/opencti-front/lang/front/en.json
```diff
@@ -4891,6 +4891,7 @@
   "Types of the targeting": "Types of the targeting",
   "Types of the targets": "Types of the targets",
   "Unable to validate remote OpenCTI URL": "Unable to validate remote OpenCTI URL",
+  "Unable to check XTM Hub connectivity. Please try again later.": "Unable to check XTM Hub connectivity. Please try again later.",
```
CelineSebe, line 4894: Could you please add this translation to the files for all languages?

PR #17577, opened 2026-08-06
feat(merge-users): Scalar user references and public sharing user rewrite (#17495)

### LINKED ISSUE #17495: PR3 — Scalar root-ID attributes rewrite (creator_id, user_id, …)
### Objective

First real handler: rewrite scalar user-id attributes from source to target across all entities. This is the bulk of the "transfer" register rows and the simplest, highest-value coverage.

### Out of scope

- Filters, JSON manifests, STIX relations, RBAC (own PRs).
- Hardening resolveValidUser so that a disabled / tombstoned account can never serve a public endpoint. It currently checks only that the user is not an internal system user and that it exists in cache — it does not check account_status. Systematically transferring *_public_user_id means no endpoint points at the merged-away source after a merge, so this is defence-in-depth rather than an active hole. Track it as a separate platform ticket.
### Dependencies

- PR2 (engine).
### Definition of Done

- Dry-run reports exact per-attribute counts of what will be rewritten.
- The real run rewrites all listed scalar ids to the target.
- *_public_user_id transfer is covered by a test asserting the anonymous exposure diff is computed and reported, and that a target-broadens-markings case is blocked without acknowledgement.
- Reuses PoC#3.

ludovic:
### Proposed changes

* Adds the scalar user reference handler, the first one that writes. Its targets are derived from the schema: every attribute declared with format `id` and pointing at User. Discovery walks the entity types one by one rather than through `getIdAttributes()`, which deduplicates by attribute name and would collapse the several entities declaring `user_id` into a single entry.
* Adds a table alongside supplying what the schema cannot say. **Dispositions**: whether a discovered attribute is rewritten here or left to another chunk — `History`, `Activity` and `PirHistory` belong to the history chunk, `platform_ip_whitelist_exclusion_ids` is invalidated rather than transferred, and `User-Account.user_id` is the STIX observable property holding an account identifier on the observed system, declared as a user reference by mistake. **Lifecycle splits**: `work.status`, `BackgroundTask.completed` and `Notification.is_read` each split one attribute into two register rows with different dispositions. **Complements**: nine paths the declarations do not expose, each recording why it is missing.
* Reports a discovered attribute with no disposition as unassigned, failing a test rather than rewriting it silently.
* Shares one query builder between counting and updating, so what the dry run reported is what the real pass writes. `creator_id` is `multiple`, so the script removes the source before appending the target, which makes a replay a no-op.
* Adds the public sharing handler. `feed_public_user_id`, `taxii_public_user_id` and `stream_public_user_id` are transferred, and the handler reports what each endpoint exposes anonymously before and after the transfer.
* Adds a `blocking` flag on `UserMergeRightsAlert`, raised by the public sharing handler when the target user carries markings or organizations the source did not have. The engine checks it between the two passes and refuses the real pass until `acknowledgeExposureChange` is set.
* Moves handler registration from the userMerge module to the end of `modules/index.ts`, since the scalar handler reads the schema and userMerge is imported before the workflow and custom field modules.

The nine complements exist because `connector_user_id`, `NewsFeedItem.user_id` and `initiator_id` are declared with format `short`, `PublicDashboard.user_id` is not declared at all, `InternalFile.metaData.creator_id` sits under an object with no attribute definitions, and the workflow version `createdBy` fields are nested children declared without `entityTypes`. Fixing those declarations is out of scope: `format: 'id'` drives filter representative resolution, the filter key schema and history diff rendering, so changing it has effects well beyond this feature. A test fails once the schema starts yielding a complement, so the entry gets dropped rather than silently duplicating a discovered target.

Filters, JSON manifests, STIX relations and RBAC have their own chunks. Hardening `resolveValidUser` against a disabled account serving a public endpoint is tracked separately; transferring `*_public_user_id` means no endpoint points at the merged-away source, so that check is defence in depth.

### Related issues

* closes #17495 — part of #17258

### How to test this PR

Stacked on #17535, so review the last 7 commits only. Enable the flag and start the platform:

```bash
yarn start
```

Create some data owned by a source user — a few entities, an ingestion, a public feed — then run a dry merge. Each target reports a count, including the ones at zero, so the report names what was examined and not only what moved:

```bash
  -d '{"query":"mutation { userMerge(sourceId:\"<source>\", targetId:\"<target>\", options:{dryRun:true}) { merge_id status report { changes { register_row_id entity_type count exact detail } alerts { register_row_id kind message blocking } coverage { covered_count is_complete } } } }"}'
```

With a target user holding markings the source does not have, the public sharing handler must raise a `blocking` alert, and the real run must refuse:

```bash
  -d '{"query":"mutation { userMerge(sourceId:\"<source>\", targetId:\"<target>\", options:{dryRun:false}) { status message } }"}'
```

The same call with `acknowledgeExposureChange: true` proceeds. Replaying it must report zero updates, since the rewrite is idempotent — worth checking specifically on `creator_id`, which is a multi-valued field.

Automated coverage:

```bash
```

### Checklist

### Further comments

72 unit tests and 27 integration tests. Discovery resolves to 17 targets, 12 motivated exclusions and no unassigned attribute — the same coverage an enumerated list would give, derived rather than written down, so a properly declared attribute added later is covered without touching this code.

The blocking refusal sits in the engine rather than in the handler. Raising it in `apply()` would stop the merge after earlier handlers already wrote, and raising it in `compute()` would keep the difference out of the dry run report, which is the one place the operator can read it before deciding.

### DIFF
--- a/opencti-platform/opencti-graphql/src/modules/userMerge/userMerge-publicSharingHandler.ts
+++ b/opencti-platform/opencti-graphql/src/modules/userMerge/userMerge-publicSharingHandler.ts
@@ -1,0 +1,171 @@
+import { getEntitiesMapFromCache } from '../../database/cache';
+import { elRawCount } from '../../database/engine';
+import { ENTITY_TYPE_USER } from '../../schema/internalObject';
+import type { AuthContext, AuthUser } from '../../types/user';
+import { SYSTEM_USER } from '../../utils/access';
+import { ENTITY_TYPE_FEED } from '../dataSharing/feed-types';
+import { ENTITY_TYPE_STREAM_COLLECTION } from '../dataSharing/streamCollection-types';
+import { ENTITY_TYPE_TAXII_COLLECTION } from '../dataSharing/taxiiCollection-types';
+import { userMergeBulkUpdate } from './userMerge-bulk';
+import {
+  type UserMergeHandler,
+  type UserMergeHandlerContext,
+  type UserMergeHandlerPlan,
+  type UserMergePlannedChange,
+  type UserMergeRightsAlert,
+  USER_MERGE_TARGET_INDICES,
+} from './userMerge-handler';
+import { USER_MERGE_REGISTRY_VERSION } from './userMerge-register';
+
+export const USER_MERGE_PUBLIC_SHARING_HANDLER = 'public-sharing-user';
+
+interface PublicSharingTarget {
+  registerRow: string;
+  entityType: string;
+  userIdField: string;
+  label: string;
+}
+
+/**
+ * Mirrors the sharing configuration of the dataSharing module, which does not export it.
+ * The list is asserted against the module's own entity types by the handler tests.
+ */
+const PUBLIC_SHARING_TARGETS: PublicSharingTarget[] = [
+  { registerRow: 'feed.public-user-id', entityType: ENTITY_TYPE_FEED, userIdField: 'feed_public_user_id', label: 'csv feed' },
+  { registerRow: 'taxii-collection.public-user-id', entityType: ENTITY_TYPE_TAXII_COLLECTION, userIdField: 'taxii_public_user_id', label: 'Taxii collection' },
+  { registerRow: 'stream-collection.public-user-id', entityType: ENTITY_TYPE_STREAM_COLLECTION, userIdField: 'stream_public_user_id', label: 'live stream' },
+];
+
+const fieldPaths = PUBLIC_SHARING_TARGETS.map((target) => `${target.entityType}.${target.userIdField}`);
+
+const sharingQuery = (target: PublicSharingTarget, sourceId: string): Record<string, unknown> => ({
+  bool: {
+    must: [
+      { term: { [`${target.userIdField}.keyword`]: sourceId } },
+      { terms: { 'entity_type.keyword': [target.entityType] } },
+    ],
+  },
+});
+
+export interface UserMergeExposureDiff {
+  addedMarkings: string[];
+  removedMarkings: string[];
+  addedOrganizations: string[];
+  removedOrganizations: string[];
+}
+
+const names = (values: { id: string; name?: string; definition?: string }[]): string[] => {
+  return values.map((value) => value.definition ?? value.name ?? value.id);
+};
+
+const difference = (left: string[], right: string[]): string[] => left.filter((value) => !right.includes(value));
+
+/**
+ * What anonymous consumers of the endpoint would see after the merge, compared to before.
+ *
+ * A public sharing endpoint serves data through the configured user's own access, so moving
+ * the reference moves the exposure with it.
+ */
+export const userMergeExposureDiff = (source: AuthUser, target: AuthUser): UserMergeExposureDiff => {
+  const sourceMarkings = names(source.allowed_marking ?? []);
+  const targetMarkings = names(target.allowed_marking ?? []);
+  const sourceOrganizations = names(source.organizations ?? []);
+  const targetOrganizations = names(target.organizations ?? []);
+  return {
+    addedMarkings: difference(targetMarkings, sourceMarkings),
+    removedMarkings: difference(sourceMarkings, targetMarkings),
+    addedOrganizations: difference(targetOrganizations, sourceOrganizations),
+    removedOrganizations: difference(sourceOrganizations, targetOrganizations),
+  };
+};
+
+export const isExposureWidening = (diff: UserMergeExposureDiff): boolean => diff.addedMarkings.length > 0;
+
+const exposureMessage = (target: PublicSharingTarget, count: number, diff: UserMergeExposureDiff): string => {
+  const parts = [
+    diff.addedMarkings.length > 0 ? `markings gained: ${diff.addedMarkings.join(', ')}` : undefined,
+    diff.removedMarkings.length > 0 ? `markings lost: ${diff.removedMarkings.join(', ')}` : undefined,
+    diff.addedOrganizations.length > 0 ? `organizations gained: ${diff.addedOrganizations.join(', ')}` : undefined,
+    diff.removedOrganizations.length > 0 ? `organizations lost: ${diff.removedOrganizations.join(', ')}` : undefined,
+  ].filter((part) => part !== undefined);
+  return `${count} ${target.label}(s) will be served with the target user access (${parts.join('; ')})`;
+};
+
+const loadUser = async (context: AuthContext, userId: string): Promise<AuthUser | undefined> => {
+  const users = await getEntitiesMapFromCache<AuthUser>(context, SYSTEM_USER, ENTITY_TYPE_USER);
+  return users.get(userId);
+};
+
+/**
+ * Moves the user reference of the public sharing endpoints.
+ *
+ * Split from the scalar handler because the work is not the same: the value is rewritten the
+ * same way, but the decision attached to it is a change in what anonymous consumers can read.
+ * Clearing the field instead — the behaviour applied when a user is deleted — would leave the
+ * endpoints public while falling back to SYSTEM_USER, which is the worst of both outcomes.
+ */
+export const userMergePublicSharingHandler: UserMergeHandler = {
+  identifier: USER_MERGE_PUBLIC_SHARING_HANDLER,
+  covers: PUBLIC_SHARING_TARGETS.map((target) => target.registerRow),
+  registryVersion: USER_MERGE_REGISTRY_VERSION,
+  reads: fieldPaths,
+  writes: fieldPaths,
+  compute: async ({ context, sourceId, targetId }: UserMergeHandlerContext): Promise<UserMergeHandlerPlan> => {
+    const changes: UserMergePlannedChange[] = [];
+    const alerts: UserMergeRightsAlert[] = [];
+    const sourceUser = await loadUser(context, sourceId);
+    const targetUser = await loadUser(context, targetId);
+    const diff = sourceUser && targetUser ? userMergeExposureDiff(sourceUser, targetUser) : undefined;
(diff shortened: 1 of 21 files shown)


opencti-platform/opencti-graphql/src/modules/userMerge/userMerge-scalarDiscovery.ts
```diff
@@ -0,0 +1,44 @@
...
+      if (!(attribute.entityTypes ?? []).includes(ENTITY_TYPE_USER)) {
+        return;
+      }
+      const carrier = carriers.get(attribute.name) ?? { multiple: attribute.multiple === true, entityTypes: [], global: false };
+      if (isAbstract(entityType)) {
+        carrier.global = true;
+      } else if (!carrier.entityTypes.includes(entityType)) {
+        carrier.entityTypes.push(entityType);
+      }
+      carriers.set(attribute.name, carrier);
```
fellowseb: If I'm not mistaken we can have the same `attribute.name` defined on different entity types with different values for the `multiple` property.
So in this foreach loop the first occurence of `attribute.name` wins and sets the `multiple` prop for the carrier.

Similarly the `carrier.global` value would be set according to the first occurence's entity type; but nothing says all entity types would be aligned, right ?

opencti-platform/opencti-graphql/src/modules/userMerge/userMerge-scalarRules.ts
```diff
@@ -0,0 +1,169 @@
...
+
+export const USER_MERGE_SCALAR_COMPLEMENTS: UserMergeScalarComplement[] = [
+  {
+    id: 'connector-user-id',
+    registerRow: 'connector.user-id',
+    entityTypes: [ENTITY_TYPE_CONNECTOR],
+    path: 'connector_user_id',
+    shape: 'single',
+    missingBecause: "declared with format 'short'",
+  },
```
fellowseb, line 98: This userMerge feature seems like a nightmare in terms of modularity as we seem to be referencing other modules' internal details like this 😅 .
We could imagine dispatching the registration of register rows/handlers/targets/scalar complements to other modules. Currently entity/mapping definitions are already split in modules so consistency would indicate doing the same here.
That would invert the dependency direction between other modules and `userMerge` too.

WDYT ?

opencti-platform/opencti-graphql/src/modules/userMerge/userMerge-scalarDiscovery.ts
ludovic: Good catch on `multiple`! In practice, platform naming conventions prevent this today (*_id is always singular and lists use plurals/*_ids, with creator_id as the only historical exception), which is why no conflicts exist in the current schema.

That said, relying purely on implicit conventions is risky: locking the flag on the first occurrence is a latent bug that could silently pick the wrong Painless script down the line if someone ever breaks convention. I'll add an explicit boot-time check to fail fast if conflicting cardinalities are ever detected for the same attribute name.

Regarding global, it's actually an accumulated OR rather than "first wins" (carrier.global is set to true as soon as any abstract type is encountered and never reset). In practice, only creator_id hits this today, where being global is the intended behavior.

opencti-platform/opencti-graphql/src/modules/userMerge/userMerge-scalarRules.ts
ludovic: I completely understand the concern, seeing userMerge name paths from a dozen modules does look like coupling at first glance 😅

Two reasons behind the current design:

1. The complements are schema debt, not the target design. USER_MERGE_SCALAR_COMPLEMENTS only exists because some modules declare their user references incorrectly today: connector_user_id with format: 'short', publicDashboard.user_id not declared at all, workflow nested children declared without entityTypes. Fix those declarations and discovery yields them automatically, there is already a guard test failing as soon as a complement becomes redundant. The right follow-up is to fix the modules, not to maintain the list.

2. The register stays centralized on purpose. A merge is irreversible, so the real risk is a silent omission: a module forgetting to register one of its references leaves orphaned data behind with no alert. The register is the list of questions, handlers claim coverage as the answers, and an attribute discovered with no disposition is a hard boot error. Inverting the dependency means nobody owns exhaustiveness anymore. It is also what makes the global checks possible (e.g. read/write disjointness across all handlers).

To be precise on what actually goes away: once the schema is fixed, what remains in userMerge is one line per attribute in the disposition table pointing at a register row. That line is the deliberate part, the decision of what a merge does with that field. The mechanical part (path, shape, nesting) is what should disappear.

marieflorescontact (commented):
tested locally ok :)

opencti-platform/opencti-graphql/src/modules/userMerge/userMerge-publicSharingHandler.ts
```diff
@@ -0,0 +1,180 @@
...
+    diff.addedMarkings.length > 0 ? `markings gained: ${diff.addedMarkings.join(', ')}` : undefined,
+    diff.removedMarkings.length > 0 ? `markings lost: ${diff.removedMarkings.join(', ')}` : undefined,
+    diff.addedOrganizations.length > 0 ? `organizations gained: ${diff.addedOrganizations.join(', ')}` : undefined,
+    diff.removedOrganizations.length > 0 ? `organizations lost: ${diff.removedOrganizations.join(', ')}` : undefined,
+  ].filter((part) => part !== undefined);
+  return `${count} ${target.noun}(s) will be served with the target user access (${parts.join('; ')})`;
+};
+
+const loadUser = async (context: AuthContext, userId: string): Promise<AuthUser | undefined> => {
+  const users = await getEntitiesMapFromCache<AuthUser>(context, SYSTEM_USER, ENTITY_TYPE_USER);
```
fellowseb: I feel like we shouldn't rely on the cache during the userMerge to prevent cache-related inconsistencies but instead directly load from ES.

opencti-platform/opencti-graphql/src/modules/userMerge/userMerge-scalarRules.ts
```diff
@@ -0,0 +1,169 @@
...
+  {
+    id: 'public-dashboard-user-id',
+    registerRow: 'public-dashboard.user-id',
+    entityTypes: [ENTITY_TYPE_PUBLIC_DASHBOARD],
+    path: 'user_id',
+    shape: 'single',
+    missingBecause: 'not declared at all, though the entity stores it',
```
fellowseb, line 113: If not done already can we create an extra ticket for this issue ? This works because the mappings are global but it's not very clean/safe.
- We should throw if the mapping is not defined on the entity type itself
- and edit the public dashboard mappings

fellowseb: I guess we could/should create an issue for each complement.

ludovic: Done https://github.com/OpenCTI-Platform/opencti/issues/18201

opencti-platform/opencti-graphql/src/modules/userMerge/userMerge-publicSharingHandler.ts
ludovic: Done, switched to resolveUserById. The risk was moderate in practice (the merge runs on an idle platform with no users or connectors, so the cache is unlikely to be stale) but reading through the resolver removes the assumption entirely.

opencti-platform/opencti-graphql/src/modules/userMerge/userMerge-scalarRules.ts
fellowseb: 1. I'm not criticizing the design per se. I like the "register" pattern and I think it can be used to make discoverability and maintenance easier. I understand the reasons behind the need for _complements_. If we can fix them easily then good but I feel many won't ever be fixed as it would require data migration and so the risk isn't worth it.

2. I'm not sure I understand your point. There's the same risk of omission in the `userMerge` module than in a domain module. I would even say the contrary: in the perspective of maintaining this feature I feel that it would be easier to keep the domain-specific definitions/logic in their respective modules so that changes/evolutions scream at the face of the maintainer.
To answer your question about exhaustiveness: we could make the domain modules register the `register` rows that they "own" along with the handlers. This wouldn't prevent global checks at all to be done by the userMerge module. Simply each domain module would be responsible for its own part of the userMerge definitions.
Example: the entire content of `userMerge-publicSharing` along with the 3 related register rows could be moved to the `dataSharing` module.

I do however understand that it would require some changes in the way the app composition root is set up in order to do it cleanly. So maybe not a straightforward evolution but at least something to think about for a refactor after the first release.

Again, I'm mainly concerned about three aspects:
- maintainability (when/if the targeted data-model changes (migration) we'll need to update the register/targets/handlers...)
- discoverability (when adding a module/feature)
High code cohesion (i.e. modularity) is IMHO a way to increase them (but certainly not the only way. For instance I believe it's a good idea to add dedicated AI agent instructions/skills)

ludovic: You're right on the omission risk, I won't defend the exhaustiveness argument.

Context on why it's centralized: this started as a detached script for one customer, with an explicit constraint to keep the footprint on existing code minimal. Outside src/modules/userMerge the stack touches ~153 lines across 9 files, and no domain module. Your proposal is exactly the footprint that was ruled out, which gives us a clean trigger: it becomes legitimate once this is a permanent part of the product rather than a one-off. Matches your "after the first release".  Opened #18211 with the target and the constraints to preserve

One correction: the complements don't need data migration. All nine are declaration problems; a field declared format: 'short' while it stores an id; the data is already correct. Cheap to fix, the risk is that everything reading the declaration changes behaviour at once.

The AI instructions idea is the cheapest of the lot and I'd take it first.

fellowseb (approved):

PR #17581, opened 2026-08-06
feat(accessibility): Update Left nav to function with keyboard controls (#17548)

### LINKED ISSUE #17548: feat(accessibility): Update Left nav to function with keyboard controls
## Use case

Currently the left nav is not able to function using keyboard navigation. Additionally the requiring users to hover to open menus  when in the collapsed state is difficult for users with limited mobility.

## Current workaround

Users can currently use the expanded nav with limited keyboard control if they can click into the nav area first.

## Proposed solution

- Revise the left nav to be able to be tabbed into and have full keyboard control functionality when in both the expanded and collapsed states.
- Standardize the handling in the collapsed state so clicking the icon opens the menu instead of navigating to the first item within the menu.
- Remove the mouseenter event handler and rely on user click or keyboard action to open the menu
- Ensure menus have accessible labels and attributes to alert screen readers of popovers and expansions. 

Yes

scarletmerlin123:
### Proposed changes

* Update component to use List components to align with documentation that Menu isn't meant for navigation
* Update collapsible sections to have appropriate aria attributes to provide information for screen readers
* Update functionality of collapsed menu to align with how the uncollapsed menu works. The user needs to click the icon to view the available page options and does not automatically navigate to the first option.
* Nested the tooltip within the rendered label to resolve dom structure issues
* Removed disableRestoreFocus so keyboard users stay focused on the correct item in the menu
* Added necessary aria attributes for popover menu
* Added necessary attributes to allow tabbing through the left nav  
### Related issues

* closes #17548 

### How to test this PR

- Use keyboard controls to navigate through the left nav, expanding and collapsing sections, and following the links
- repeat the process with the nav collapsed and ensure can navigate through the nav and open menus using only keyboard controls
- confirm labeling using screen reader  

### Checklist

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/nav/LeftBarItem.tsx
+++ b/opencti-platform/opencti-front/src/private/components/nav/LeftBarItem.tsx
@@ -1,7 +1,7 @@
 import { ArrowDropDown, ArrowDropUp } from '@mui/icons-material';
-import { alpha, Collapse, ListItemIcon, ListItemText, MenuItem, MenuList, Popover, SxProps, Tooltip } from '@mui/material';
+import { alpha, Collapse, List, ListItemButton, ListItemIcon, ListItemText, MenuItem, MenuList, Popover, SxProps, Tooltip } from '@mui/material';
 import { useTheme } from '@mui/styles';
-import React, { useRef } from 'react';
+import React, { useRef, useState } from 'react';
 import { Link, useLocation } from 'react-router-dom';
 import { Theme } from '../../../components/Theme';
 import useDraftContext from '../../../utils/hooks/useDraftContext';
@@ -10,14 +10,14 @@ interface SubMenuItem {
   type?: string;
   link: string;
   label: string;
-  icon?: React.ReactNode;
+  icon?: React.ReactElement;
   exact?: boolean;
   granted?: boolean;
 }
 
 interface LeftBarItemProps {
   id: string;
-  icon: React.ReactNode;
+  icon: React.ReactElement;
   label: string;
   link: string;
   exact?: boolean;
@@ -32,6 +32,15 @@ interface LeftBarItemProps {
   isMobile: boolean;
   submenuShowIcons?: boolean;
   hiddenEntities?: string[];
+}
+
+interface RenderMenuItemParams {
+  itemLabel: string;
+  selected: boolean;
+  showIcon?: boolean;
+  fontSize?: 'default' | 'small';
+  forceShowText?: boolean;
+  itemIcon?: React.ReactElement;
 }
 
 const LeftBarItem: React.FC<LeftBarItemProps> = ({
@@ -55,7 +64,9 @@ const LeftBarItem: React.FC<LeftBarItemProps> = ({
   const location = useLocation();
   const theme = useTheme<Theme>();
   const draftContext = useDraftContext();
-  const anchorRef = useRef<HTMLLIElement | null>(null);
+  const anchorRef = useRef<HTMLDivElement | null>(null);
+  const [showSubItemMenu, setShowSubItemMenu] = useState<boolean>(false);
+  const [usingHover, setUsingHover] = useState<boolean>(false);
 
   const visibleSubItems = subItems.filter(
     (item) => item.granted !== false && (!item.type || !hiddenEntities.includes(item.type)),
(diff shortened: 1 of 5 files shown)


opencti-platform/opencti-front/src/private/components/nav/LeftBarItem.tsx
```diff
@@ -45,17 +45,14 @@ const LeftBarItem: React.FC<LeftBarItemProps> = ({
   selectedMenu,
   onClick,
   onMenuToggle,
-  onMenuOpen,
```
mcartaud, line 48: If those props are not used anymore, please remove them from the Props interface

opencti-platform/opencti-front/src/private/components/nav/LeftBarItem.tsx
```diff
@@ -210,15 +224,15 @@ const LeftBarItem: React.FC<LeftBarItemProps> = ({
...
+        <ListItemButton
           component={Link}
           to={link}
           dense
           onClick={onClick}
           sx={getMenuStyles(isParentSelected)}
         >
-          {renderMenuItem(icon, label, isParentSelected)}
-        </MenuItem>
+          {renderMenuItem(label, isParentSelected, undefined, undefined, undefined, icon)}
```
mcartaud: Is it possible to replace the props of renderMenuItem with a object like this? It is more pleasant to use than this list of undefined
```javascript
  interface RenderMenuItemParameters {
    itemLabel: string;
    selected: boolean;
    showIcon?: boolean;
    fontSize?: 'default' | 'small';
    forceShowText?: boolean; // For popover items
    itemIcon?: React.ReactElement;
  }

  const renderMenuItem = (
    {
      itemLabel,
      selected,
      showIcon = true,
      fontSize = 'default',
      forceShowText = false, // For popover items
      itemIcon,
    }: RenderMenuItemParameters,
  )
```

mcartaud (commented):
Works well overall, only a small behavior change on collapsed version : in original version, sub-menu opened on hover, now we need to click on the icon

opencti-platform/opencti-front/src/private/components/nav/LeftBarItem.tsx
scarletmerlin123: updated to use an interface and default values to clean this up

opencti-platform/opencti-front/src/private/components/nav/LeftBarItem.tsx
scarletmerlin123: removed unused functions

scarletmerlin123:
@mcartaud The move to require clicking was intentional. Hover activated menus are not accessible for keyboard users or anyone with limited mobility. Additionally, moving to require clicks mirrors the expanded menu functionality where clicking on a high level menu item reveals the subitems.

mcartaud:
> @mcartaud The move to require clicking was intentional. Hover activated menus are not accessible for keyboard users or anyone with limited mobility. Additionally, moving to require clicks mirrors the expanded menu functionality where clicking on a high level menu item reveals the subitems.

After discussing with product team, the hover behavior is widely used and we want to keep it (as well as the current click behavior which is to go on the first menu item). Do you think it is possible to keep the hover and click behavior whil improving keyboard navigation?

scarletmerlin123:
I updated it to add back the previous functionality. To do so did add some complexity since things need to behave slightly different if a user access the menu through a hover vs the keyboard.

mcartaud (commented):
Thanks for the changes. Two things I noticed:
- the icon of the menu has an aria-label but is also aria-hidden. This is fine when not collapsed, but will causes issues when collapsed. Is it possible to put the aria-label on a parent element (like the ListItemButton) to fix that?
- in collapsed mode, the tooltip will cover the first submenu entry. Is it possible to only display it on icon without submenu? It is the current behavior

scarletmerlin123:
Updated to only show the tooltip on items without sub menus. I also moved the aria-label to the listItemButton component. Because of how the Tooltip component works it will always pass down an aria-label even if the icon is aria-hidden="true". The parent icon button has the label now so this is fine and doesn't cause any usability issues. 

I also added something to restore focus to the button when leaving the menu and adjusted some aria attributes for the submenu popover.

mcartaud:
Looks good, but some tests are to fix

SarahBocognano (approved):

SarahBocognano:
Need to fix the e2e tests here and we'll be good to merge ! :)

scarletmerlin123:
@SarahBocognano I updated the failed test, though not sure if this will end up a widespread issue as more items get labeled. I'm having a difficult time running the e2e tests locally to see if this issue will pop up in other places. A quick glance at other getByLabel locators seems like "Entities" was the only one had multiple instances.

SarahBocognano:
@scarletmerlin123 I'll try to have a look to the e2e tests locally to see if I can make things works here ! (not sure of my skills on this one haha, but better try than do nothing :p)

PR #17682, opened 2026-08-12
feat(domain-specific-platform): Global configuration export (#16785)

### LINKED ISSUE #16785: feat(domain-specific-platform): Global configuration export

SarahBocognano:
### Proposed changes

* Add new functions in settings to be able to export configurations
* Add new function in theme in the backend to export by global export
* Add new function to export entity settings hidden types
* Update tests
* Add telemetry counts and stream event
* Update documentation on telemetry

### Related issues

* closes: https://github.com/OpenCTI-Platform/opencti/issues/16785

### How to test this PR
Go to Settings -> Filigran Experience -> Export Bootstrap Bundle Platform
Select the entities you want to export
Check if the search bar for specific export is working well
Export and see if everything is in your zip !

### Checklist


opencti-platform/opencti-graphql/src/modules/globalExport/globalExport-domain.ts
```diff
@@ -0,0 +1,279 @@
...
+    }
+  });
+
+  const archive = new ZipArchive();
+  const chunks: Buffer[] = [];
+  const zipReady = new Promise<void>((resolve, reject) => {
+    archive.on('data', (chunk: Buffer) => chunks.push(chunk));
+    archive.on('end', resolve);
+    archive.on('error', reject);
+  });
```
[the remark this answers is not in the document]
SarahBocognano: What do you think about this comment ? I think copilot is a bit of a drama queen here but your advice is welcome @fellowseb @esrevi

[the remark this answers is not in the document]
esrevi: from what I understand, it builds the zip in memory and we encode in base64 it and return that in the graphql response, imo, without limitation, this could increase a lot the memory usage, and on frontend we have to store that base64 in memory and decode it too. It feels not very efficient.

have you considered uploading the archive file to storage and return a url link instead? it could be better for that case?

opencti-platform/opencti-front/src/private/components/settings/experience/exportBundleInstances.tsx
```diff
@@ -0,0 +1,165 @@
+import { graphql } from 'react-relay';
```
esrevi, line 1: there is a mix of types and queries in this TSX file (without react component ^^) 
Could we split in three files? one for the types, one for the graphql queries, one for the configs  ? and it should be ts extension 
for example something like (to be checked if Relay is ok with that ^^)
exportBundleInstances.types.ts 
exportBundleInstances.queries.ts 
exportBundleInstances.config.ts

opencti-platform/opencti-front/src/private/components/settings/experience/GlobalExportBundleDrawer.tsx
```diff
@@ -0,0 +1,285 @@
...
+    });
+
+    setExporting(true);
+    try {
+      const result = await fetchQuery(
+        platformBundleDrawerExportQuery,
+        { entityTypes, selections },
+      ).toPromise() as PlatformBundleDrawerExportQuery$data;
+      if (result?.globalConfigurationExport) {
+        const blob = new Blob([base64ToBytes(result.globalConfigurationExport)], { type: 'application/zip' });
```
esrevi, line 116: if we change how we return the query with a file download, it may need some updates here

opencti-platform/opencti-front/src/private/components/settings/experience/GlobalExportBundleDrawer.tsx
```diff
@@ -0,0 +1,285 @@
...
+        const safeBundleName = bundleName.trim().replace(/[^a-z0-9-_]+/gi, '_').replace(/^_+|_+$/g, '').slice(0, 80);
+        const suffix = safeBundleName ? `_${safeBundleName}` : '';
+        const [day, month, year] = new Date().toLocaleDateString('fr-FR').split('/');
```
esrevi, line 119: i think we should not have fr-FR hardcoded

opencti-platform/opencti-front/src/private/components/settings/experience/GlobalExportBundleDrawer.tsx
```diff
@@ -0,0 +1,285 @@
...
+    >
+      {({ onClose: handleClose }) => (
+        <>
+          <Box sx={{ flexGrow: 1, overflowY: 'auto', p: 3, display: 'flex', flexDirection: 'column', gap: 2 }}>
+            <Typography variant="body2" color="textSecondary">
+              {t_i18n('Select all the elements to include in your configuration bundle')}
+            </Typography>
+
+            {EXPORT_INSTANCE_CONFIGS.map((config, index) => {
+              const previousGroup = index > 0 ? EXPORT_INSTANCE_CONFIGS[index - 1].group : undefined;
```
esrevi, line 155: I let you decide, but I think it could be better to extract components, can still live in that file, but this would help for quicker debug and also having tests :)

opencti-platform/opencti-front/src/private/components/settings/experience/GlobalExportBundleDrawer.tsx
```diff
@@ -0,0 +1,285 @@
...
+                    accordionSx={accordionSx}
+                  />
+                </React.Fragment>
+              );
+            })}
+
+            {EXPORT_CATEGORIES.map((category) => {
+              if (category.kind === 'placeholder') {
+                return (
+                  <Accordion key={category.key} disableGutters expanded={false} sx={{ ...accordionSx, opacity: 0.5 }}>
```
esrevi, line 179: for example, this could be extracted to 
CategoryPlaceholder then we'll got

```
if (category.kind === 'placeholder') {
return <CategoryPlaceholder category={category} />
}

if (category.kind === 'flat') {
return <CategoryFlat category={category} />
}
```

etc

Issue #127, opened 2019-07-08
OpenCTI and dependencies memory documentation 


richard-julien:
## Problem to Solve

Currently its not clear the amount of memory required to have openCTI running

## Current Workaround

No

## Proposed Solution

Write documentation about openCTI and dependencies.
- Grakn (4go) + Inner cassandra (4go)
- Elasticsearch (2go)
- Redis (512 Mo)
- RabbitMQ (256 Mo)
- OpenCTI API. (64 Mo)
- OpenCTI worker (64 Mo)
- OpenCTI connector (64 Mo)

## Hints
- Grakn: Open question https://discuss.grakn.ai/t/memory-configuration-in-docker/1245

> The Cassandra heap should be no less than 2GB, and no more than 50% of your system RAM
> Heaps smaller than 12GB should consider ParNew/ConcurrentMarkSweep garbage collection
> Heaps larger than 12GB should consider G1GC

- Elastic: https://www.elastic.co/guide/en/elasticsearch/reference/current/docker.html
- Redis: https://hub.docker.com/r/bitnami/redis/
- NodeJS: By default the memory limit in Node.js is 512MB. https://medium.com/@ashleydavis75/node-js-memory-limitations-30d3fe2664c0
- Python: https://docs.python.org/3.7/library/resource.html, or with container https://ss64.com/bash/ulimit.html / https://www.admin-linux.fr/controle-des-ressources-systemes-ulimit/

SamuelHassine:
@richard-julien This will be in a specific documentation page with links to documentation of dependencies on how enforce memory limits (ElasticSearch, Grakn, Redis and Minio).

richard-julien:
I think we need to enforce a default configuration that work with the minimal requirement. And of course explains this choice in the doc.

Issue #126, opened 2019-07-07
OpenCTI strategic roadmap


richard-julien:
## Problem to Solve

Everyone should be able to consult the strategic roadmap around openCTI

## Current Workaround

No

## Proposed Solution

Create the roadmap on https://www.opencti.io/roadmap

## Additional Information

The roadmap is a view of the openCTI strategic target.

SamuelHassine:
The strategic roadmap is now available on: https://www.opencti.io/en/roadmap/. In English and French.

Issue #39, opened 2019-05-22
Add aliases to the generic entity creation form


SamuelHassine:
## Problem to Solve

It could be very interesting to be able to add aliases to an entity directly during the creation in the report knowledge view.

## Current Workaround

Create the entity, go to the edition page and modify alias.

## Proposed Solution

Implement an alias field in the form and adjust the API.

## Additional Information

None.

Issue #38, opened 2019-05-22
Automatic graph organization on report


SamuelHassine:
## Problem to Solve

Be able to organize the report knowledge graph in one click and avoid the "stack" of all entities when adding new ones.

## Current Workaround

None.

## Proposed Solution

Implement graph organization algorithms on the report knowledge view.

## Additional Information

None.

PR #17727, opened 2026-08-16
feat(paper): adopt the library Paper on 25 call sites and put every card on the token surface (#17728)

sandyghs:
Adopts the design system `Paper` on the container surfaces that can carry it
without losing a behaviour, and puts the whole card family on the token surface.

Counted by `node fds-migration/scripts/count-surfaces.mjs`, which imports the
same analyser the conformity gate uses. That script is the authoritative method;
no figure here comes from a grep.

| | count |
|---|---|
| library `Paper` call sites | **25** in 19 files |
| files still importing MUI's `Paper` | 7 |
| `Card` wrapper sites moved onto the token surface | **219** in 166 files |

## What changes

- **Surfaces.** 25 call sites take the library `Paper`. The surfaces left out are
  listed with their measured reason — semi-transparent, gradient, accent-tinted
  border, floating, deliberately border-less.
- **Cards.** `Card` and `CardAccordion` stop painting the hardcoded `#0C1524`
  literal, which is no step of the elevation scale, and read the library's
  per-layer hook `--bg-elevation-default-layer-1` — the same lever a library
  `Paper` reads, so both answer a host redeclaring a layer. Their border reads
  the matching hook, `--border-elevation-subtle-soft-layer-1-transparency-15`,
  for the same reason: the semantic alias resolves from layer 0, the one variable
  the hook does not push, so a card edge would otherwise stay behind while every
  `Paper` moved.
- **Charts.** All nine factories, plus four colours that exist only to coincide
  with the carrying surface, follow the card instead of the old literal.
- **Login and customer themes.** Public surfaces move to layer 1;
  `useFdsThemeScope` redeclares the per-layer base for surface AND border, which
  is the only override the library honours.
- **Pin.** `f86e76e`, bringing library #126 and #127 (the Header's 24px leading
  padding, and the EE chip label reduced to 12px). Proved by the served bytes.

## Measured consequences, declared

**Token values moved by the bridge regeneration.** The pin bump regenerated
`src/components/fds-tokens.generated.ts`, which moved **14 token values** — 6 in
dark, 8 in light. Diffed against the base, per mode:

| token | mode | before | after | euclidean |
|---|---|---|---|---|
| `--color-feedback-alert-tertiary` | light | `#b8870a` | `#573e05` | 121.50 |
| `--color-filigran-tonic-secondary` | light | `#005744` | `#00241c` | 64.82 |
| `--color-feedback-warning-tertiary` | light | `#884106` | `#572a05` | 54.14 |
| `--color-feedback-warning-primary`, `--border-alert-warning`, `--icon-warning` | light | `#e6700f` | `#b8550a` | 53.57 |
| `--border-elevation-subtle-layer-1` | light | `#e4e5e7` | `#cacbce` | 44.46 |
| `--bg-elevation-disabled-layer-3` | dark | `#13213e` | `#182a4e` | 19.03 |
| `--bg-elevation-disabled`, `-layer-0`, `--bg-input-disabled` | dark | `#101b33` | `#0c1527` | 14.00 |
| `--bg-elevation-disabled-layer-1`, `-layer-2` | dark | `#13213e` | `#101b33` | 12.88 |
| `--gradient-warning` | light | `#e6700f → #f2be3a` | `#b8550a → #f2be3a` | n/a, a gradient |

`--gradient-warning` is listed for completeness and has **no consumer**: no
product line reads it, in any file outside the generated bridge itself.

For scale: the card literal this PR corrects moved 7.35 by the same measure, so
the disabled family in dark moves about twice as far.

**The token move is real and declared; its arrival on a pixel is not.** The chain
breaks at the last link, and the measurement says where.
`--bg-elevation-disabled` feeds `designSystem.background.disabled`
(`ThemeDark.ts:174`), and no product line reads that field. The disabled button
reads `palette.background.disabled`, which is a hardcoded `#363B46`
(`ThemeDark.ts:103`), so the moved token never reaches it. The disabled buttons
found rendered are secondary and tertiary variants, which `getDisabledSx` paints
`transparent` — measured `rgba(0, 0, 0, 0)`. No disabled PRIMARY button was found
rendered to close the measurement, so the last hop is stated as unproven rather
than as safe.

**Three tokens move when a customer theme is active**, and only these three:
`--bg-elevation-default-layer-1`, `--border-elevation-subtle-soft-layer-1` and
its `-transparency-15` derivation, all taking the customer's paper colour. The
last two mean a customised install has **no visible card edge**, because the
border resolves to the surface colour — shown, not fixed. That is true as of this
round and was not before: until the card border was moved onto the hook, cards
kept the built-in edge on a customer theme while panels had moved, and the
sentence claiming otherwise was contradicted by the measurement.

Expected, not a regression: the warning colour darkens in light mode only.

## Guards

The `paperPattern` motif in `check-fds-conformity.mjs` checks that a converted
file no longer imports MUI's `Paper` (deep default import included) and that no
padding is re-declared by hand on a library `Paper`. Both were seen red before
being trusted green — but only in one variant: `className="p-4"` in the leading
position passed green until this round, and the regex now covers it, proved red on
four shapes. MIXED files are declared through `mixed.allowMuiPaperFor`.

Gate: 54 checks, 0 failures. `tsc --noEmit` clean.

## Where to read the rest

- `fds-migration/PAPER-GAP-INVENTORY.md` — the reference: what was declined and
  why, the arbitrated rules, the sizing for the Card wave, the measurement traps.
- `fds-migration/LIBRARY-FEEDBACK.md` — entries **30 to 39**, each with a removal
  test. #37 blocks `title`/`action` adoption; #38 and #39 come from the navbar and
  the chip.


sandyghs:
**Declared re-run — `Backend / Integration tests`, `push` event, SHA `61e85da1d1`.**

Characterisation before re-running, so the re-run is a test of one hypothesis and
not a hope:

- **Zero backend files in the diff.** This PR touches `opencti-front` and
  `fds-migration` only.
- **The same job is green on the other event of the same SHA.** `pull_request`
  run: 17/17 success, `Backend / Integration tests` included.
- **What failed:** `tests/03-integration/04-manager/telemetryManager-test.ts:267`,
  `expect(await redisGetTelemetry(...))` — expected `1`, received `0`. One test
  file failed out of 214; 2123 tests passed.
- **Why it reads as a race:** the assertion reads a telemetry gauge out of Redis
  that a manager writes asynchronously. A gauge at 0 instead of 1 is a value not
  yet written, not a wrong value.
- Environment noise in the same log: `digest-mismatch` ×3,
  `ERROR (no HTTP response)` ×13, `ECONNREFUSED` ×28, and
  `close timed out after 5000ms` at the end of the run.

This is **not** the false red this branch met before: previously the platform
never started and no test ran. Here a test genuinely failed, which is why it is
being re-run rather than dismissed.

If the same test fails identically on the re-run, no second re-run will be
attempted — the question moves to whether this test is known to be flaky.

sandyghs:
**Re-run result — green.** `Backend / Integration tests` passed on the re-run, and
the `push` run for SHA `61e85da1d1` is now 17/17 success at both the job and the
run level.

The hypothesis stated before re-running is the one that held: the telemetry gauge
was read out of Redis before the manager had written it. Nothing in the diff was
involved — this PR touches `opencti-front` and `fds-migration` only, and the same
job had already passed on the `pull_request` event of the same SHA.

One re-run, no second attempt.

Head has since moved to `ff689e4ec8`, which adds a comment recording why the top
bar's Suspense placeholder is deliberately left under the rail. Cross-check on
that SHA: `push` **17/17**, `pull_request` **17/17**.

PR #17745, opened 2026-08-17
fix(user): improve org admin (#17742)

### LINKED ISSUE #17742: fix(user): improve org admin
## Description

improve org admin 

## Environment

2. Version: { e.g. 1.2.3 }
3. Other environment details:

## Reproducible steps

1. { e.g. Run ... }
2. { e.g. Click ... }
3. { e.g. Error ... }

JeremyCloarec:
### Proposed changes

* improve org admin
*

### Related issues

* closes #17742 

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/domain/user.js
+++ b/opencti-platform/opencti-graphql/src/domain/user.js
@@ -1069,6 +1069,7 @@ export const roleDeleteRelation = async (context, user, roleId, toId, relationsh
   return notify(BUS_TOPICS[ENTITY_TYPE_ROLE].EDIT_TOPIC, role, user);
 };
 
+const ORG_ADMIN_FORBIDDEN_EDIT_ATTRIBUTES = ['user_email', 'password'];
 // User related
 export const validateAndNormalizeEmailInput = async (context, userId, input) => {
   if (input.key === 'user_email') {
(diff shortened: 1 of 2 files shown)


opencti-platform/opencti-graphql/src/domain/user.js
```diff
@@ -1086,6 +1086,12 @@ export const userEditField = async (context, user, userId, rawInputs) => {
     if (userToUpdate.external && input.key === 'password_valid_until') {
       throw FunctionalError('Cannot force password change for external user', { userId });
     }
+    if (input.key === 'user_email') {
```
SouadHadjiat: maybe we could regroup with the next condition and have a list of "forbidden_keys_for_orga_admins" ? would be easier then for the next case to just add the key to the list.

But I wonder if orga admin shouldn't be able to update emails, since they can create users, isn't it weird that they couldn't then update their emails?

JeremyCloarec: Will add the forbidden keys constant 👍 

Good question, I assumed that the change was validated with product team but it might not be, I don't know

opencti-platform/opencti-graphql/src/domain/user.js
```diff
@@ -1103,13 +1104,13 @@ export const userEditField = async (context, user, userId, rawInputs) => {
     if (userToUpdate.external && input.key === 'password_valid_until') {
       throw FunctionalError('Cannot force password change for external user', { userId });
     }
+    // org admin can only edit forbidden attributes on its own user and not other users
+    if (!isUserHasCapability(user, SETTINGS_SET_ACCESSES) && ORG_ADMIN_FORBIDDEN_EDIT_ATTRIBUTES.includes(input.key) && user.id !== userId) {
```
Archidoit, line 1108: Just a remark. We ususally prefer having a list of authorized keys rather than a list of unauthorized keys. Because if a new key is added, it's more secure that it's not updatable by default if we forget to add it in one of the list (forgetting to add it in the list of authorized key is not a problem, forgetting to add it in the list of forbidden key do is...).

PR #17784, opened 2026-08-20
feat(manual_sc): Add the possibility to choose the entities for has covered relationship creation (#14716)

MTorbay-Filigran:
### Proposed changes

* Added new step to select tested entities linked to object covered in SC creation

### Test counters changes

ADD attack pattern : +9
ADD report : +3
ADD security coverage : +5
ADD security coverage result : + 5

DELETE attack pattern : +9
DELETE report : +3
DELETE security coverage : +5
DELETE security coverage result : + 5

### Related issues

* closes https://github.com/OpenCTI-Platform/opencti/issues/17824

### Checklist


opencti-platform/opencti-front/src/private/components/analyses/security_coverages/security_coverage_creation/SelectEntitiesToCoverStep.tsx
```diff
@@ -0,0 +1,244 @@
...
+      id
+      definition_type
+      definition
+      x_opencti_order
+      x_opencti_color
+    }
+  }
+`;
+
+const HAS_COVERED_TARGETS_TYPES = ['Attack-Pattern', 'Vulnerability', 'Artifact', 'Indicator'];
```
nvkhope, line 122:
```suggestion
const HAS_COVERED_TARGETS_TYPES = ['Attack-Pattern', 'Vulnerability', 'Artifact', 'Indicator', 'SecurityPlatform'];
```

opencti-platform/opencti-graphql/src/modules/securityCoverage/securityCoverageResult/securityCoverageResult-domain.ts
```diff
@@ -120,59 +113,52 @@ export const deleteSecurityCoverageResult = async (
...
+    return createListTask(context, user, {
+      description,
+      scope: 'KNOWLEDGE',
+      ids: selectedIds,
+      actions,
+      orderMode: 'asc',
     });
   }
 
+  const filters = selection?.filters ?? JSON.stringify({ mode: 'and', filters: [], filterGroups: [] });
```
lndrtrbn: There is a const `emptyFilterGroup` already existing in filtering-utils that you can reuse

opencti-platform/opencti-graphql/src/modules/securityCoverage/securityCoverageResult/securityCoverageResult-domain.ts
```diff
@@ -120,59 +113,52 @@ export const deleteSecurityCoverageResult = async (
...
  * @param user User making the request.
  * @param securityCoverageResultId ID of the security coverage result to populate.
+ * @param selection Entity selection (selected_ids, or filters/search/excluded_ids).
  * @returns The created background task.
  */
 export const createHasCoveredRelTask = async (
   context: AuthContext,
   user: AuthUser,
   securityCoverageResultId: string,
+  selection?: SecurityCoverageSelectedEntitiesInput | null,
```
lndrtrbn: You never check if selection exists or not, that means you can start a background task without any filters that will applied on all entities in the platform.

I think selection should not be nullable, if so this function should not be called at all. And also need to verify there is some filters, non empty, because we do not want to start ab background task on all entities

opencti-platform/opencti-graphql/src/modules/securityCoverage/securityCoverageResult/securityCoverageResult-domain.ts
lndrtrbn: In fact, if no filters we should not create a background task to avoid targeting all the platform.

opencti-platform/opencti-front/src/private/components/analyses/security_coverages/security_coverage_creation/SecurityCoverageCreation.tsx
```diff
@@ -0,0 +1,487 @@
...
+      id
+      name
+      active
+      connector_type
+      connector_scope
+    }
+  }
+`;
+
+const securityCoverageValidation = (t: (value: string) => string, isAutomated: boolean) => {
```
lndrtrbn: rename t to t_i18n to allow good parsing from auto translate script

Issue #17728, opened 2026-08-16
feat(fds): migrate the container surfaces to the design-system Paper


sandyghs:
## Use case

The Filigran Design System (FDS) is being rolled out across the XTM product line so that shared components are maintained once instead of once per product. After the left navigation and the admin top bar, container surfaces are the next component migrated in OpenCTI.

Today OpenCTI paints its panels with MUI `Paper`: 28 direct sites, each carrying its own padding, its own elevation and its own border decision. The values drift — nine of them use a padding no scale expresses, seven draw no border at all while twenty-one do. Every visual fix has to be made again, site by site, and again in each product.

## Current workaround

Each product keeps its own surface conventions and they drift apart. A first migration was carried out on OpenAEV (13 surfaces converted, measured iso on the rendered DOM), which produced the gap-inventory method and the arbitration list; OpenCTI is the second application of it.

## Proposed solution

Replace MUI `Paper` with the `Paper` component from `@filigran/design-system`, one measured wave at a time, starting with a gap inventory that ends on a STOP so every deviation is arbitrated before a single site is converted.

What it brings:

- padding becomes an explicit prop on a shared scale instead of per-site values
- elevation, surface colour and border come from the design tokens, so a custom customer theme repaints them consistently
- the header row (title + actions) becomes a real slot instead of per-product typography
- deviations are recorded and gated rather than rediscovered

Out of scope for this wave: the card wrapper and its 164 call sites (they need the `Card` component), accordions, semi-transparent containers, and floating surfaces that belong to dialogs and menus. No functionality, information or interaction is lost — a site that cannot be converted without a loss is listed, not forced.

## Additional information

Phase 0 is already open and green: the pin bump, three dead token references revived (all three silent — `tsc`, `eslint` and the build passed on the breakage, and one of them painted the selected navigation row in the default brand colour instead of the customer accent), and the 28 surfaces measured on the rendered DOM.

The first pull request also carries the groundwork the migration needs: the pin bump to the head of the library, and three dead token references revived. All three were silent — `tsc --noEmit`, `eslint` and the build passed on the breakage — and one of them declared an old token name, which painted the selected navigation row in the default brand colour instead of the customer accent. A rename breaks what a host *writes*, not only what it reads, and nothing signals it; the hunt is therefore reproducible, searching both silent shapes with a strict word boundary and cross-checking against the installed stylesheet.

## If the feature request is approved, would you be willing to submit a PR?

Yes — phase 0 is already open.

PR #17788, opened 2026-08-20
docs(fds-migration): record the Combobox reconnaissance and its blockers

sandyghs:
## What

Documentation only, two Markdown files under `fds-migration/`. **No component
code changed, nothing converted, no pin touched.** The Combobox adoption is
sequenced behind library work; this keeps the reconnaissance so the next session
starts from it.

## Why

The conversion stopped on its own conditions: the census disagreed with the
brief by more than the tolerance on the axis that defines the work list, and the
pivot turned out to be blocked by the same rule that sends a site back to MUI.
Both are now filed with their counts and their removal tests, and the library
gaps are on the table before the library session opens.

## Gate verdict

`node fds-migration/scripts/check-fds-conformity.mjs` — 54 checks, 0 issues
(52 OK, 2 SKIPPED) on this SHA. No build, test or type-check surface is touched.

## Known limits

Nothing is converted and no workaround marker is placed, because no file was
converted. The label and helper-text DOM delta is not measured — it needs a
converted screen. Two tooling defects are recorded rather than fixed here: this
product's JSX analyser is blind to type-parameterised call sites (fixed upstream
in `XTM-Foundation/filigran-design-system#131`, not in this product's forked
copy), and regenerating `fds-migration/` would drop Paper coverage unless paired
with a `libComponentUsage` declaration.

## Where to read

`fds-migration/IMPLEMENTATION-LOG.md`, entry dated 2026-08-20 — the numbers, the
two blockers, the two tooling defects. `fds-migration/LIBRARY-FEEDBACK.md`
entries 40 (data colour on the value chips, the blocking one), 41
(select-on-focus) and 42 (the remaining gaps and the delivery table).

## Arbitrations required

The PR title needs an issue number to satisfy this repo's title convention
(`type(scope): description (#issue)`), and no Combobox issue exists yet — the
four earlier design-system PRs reference #17539, the Paper one #17728.

### DIFF
--- a/fds-migration/IMPLEMENTATION-LOG.md
+++ b/fds-migration/IMPLEMENTATION-LOG.md
@@ -836,3 +836,118 @@ symbol.
 Both grafts are test-only. The rendered bar at this head is identical to the one
 measured above, so the checkpoint was not re-run: there is nothing new to look
 at, and re-photographing an unchanged bar would only look like evidence.
+
+## 2026-08-20 — Combobox reconnaissance: measured, nothing converted
+
+Session goal was to adopt the library `Combobox` on the `AutocompleteField`
+pivot and the direct MUI sites. **No product component code was written**: the
+session stopped twice on the conditions it was given, and the library work is
+now sequenced first. What follows is the measurement, so the next session does
+not repeat it.
+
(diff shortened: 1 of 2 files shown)


sandyghs:
PR for unnecessary documentation

PR #17791, opened 2026-08-20
fix(activity): refactor audits (#17789)

### LINKED ISSUE #17789: fix(activity): refactor activity
## Description

refactor activity

## Environment

2. Version: { e.g. 1.2.3 }
3. Other environment details:

## Reproducible steps

1. { e.g. Run ... }
2. { e.g. Click ... }
3. { e.g. Error ... }

JeremyCloarec:
### Proposed changes

* refactor audits
*

### Related issues

* closes #17789

### Checklist


opencti-platform/opencti-graphql/src/domain/log.ts
```diff
@@ -34,8 +34,10 @@ export const findById = async (context: AuthContext, user: AuthUser, logId: stri
   return storeLoadById<StoreHistory>(context, user, logId, ENTITY_TYPE_HISTORY, { historyFiltering: true });
 };
 
-export const findAudits = (context: AuthContext, user: AuthUser, args: QueryAuditsArgs) => {
-  let types = args.types ? args.types : isUserHasCapability(user, SETTINGS_SECURITYACTIVITY) ? [ENTITY_TYPE_ACTIVITY] : [ENTITY_TYPE_HISTORY];
+export const computeAuditTypes = (user: AuthUser, requestedTypes?: string[] | null): string[] => {
```
Archidoit, line 37: if requestedTypes = ['User'], the 'user' entry won't be filtered out. Consider checking that the result only contains audits types, if not throw an error.

Archidoit (changes requested):

opencti-platform/opencti-graphql/src/domain/log.ts
JeremyCloarec: It's true that we still do not verify that entity types here are audit types, like before, but audit typing restriction is handled in engine.ts thanks to the historyFiltering parameter set on all queries going out of log.ts :) I can add type filtering here also but it would be redudant with the historyFiltering

Archidoit: Ah yes okay, if there is already a filtering after, it's ok :) Maybe we can still add it as a safeguard for a future miseuse. But anyway, i'm approving the PR !

Archidoit (approved):

PR #17826, opened 2026-08-20
fix(notifications): filter out inactive users in notifications (#16148)

### LINKED ISSUE #16148: fix(notifications): notifications are still sent to expired or inactive accounts
## Use case

When a user account in OpenCTI is expired, disabled, or marked as inactive, the platform should automatically stop sending notifications to this account.

Currently, expired or inactive accounts continue receiving notifications (email and/or in-app notifications), even though these users should no longer interact with the platform.

This creates inconsistencies between account lifecycle management and the notification system, and may introduce unnecessary noise or compliance/security concerns.

## Current Workaround

There is currently no clean workaround available directly within OpenCTI.

Administrators must manually:

remove notification subscriptions,
disable notification channels,
or delete accounts entirely to ensure notifications stop being sent.

This approach is time-consuming and error-prone, especially in environments with many users or temporary accounts.

## Proposed Solution

Add a validation step in the notification pipeline to automatically verify whether a target account is:

active,
non-expired,
and enabled,

before sending any notification.

If the account is inactive or expired:

the user should automatically be excluded from notification recipients,
no email or in-app notification should be generated,
and scheduled notification digests should also ignore these accounts.

This validation should apply consistently across all notification mechanisms.

JeremyCloarec:
### Proposed changes
This pull request addresses a bug where expired, disabled, or inactive user accounts were still receiving notifications (both live and digest, via email and in-app). The fix ensures that only active user accounts receive notifications by adding a centralized validation in the notification pipeline. Unit tests have been added to verify the correct behavior.

**Notification recipient filtering:**

* Added the `isNotificationRecipientActive` function to check if a user account is active and not expired, using both account status and expiration date.
* Updated `getNotifications` to filter out users who are not active, ensuring that inactive, expired, or disabled accounts no longer receive notifications.

**Testing and validation:**

* Introduced comprehensive unit tests in `notification-recipient-active-test.ts` to verify that only active users receive notifications and that various edge cases (expired, disabled, or future expiration) are correctly handled.

**Configuration:**

* Imported `ACCOUNT_STATUS_ACTIVE` from the configuration to support the new recipient validation logic.

### Related issues

* closes #16148

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/manager/notificationManager.ts
+++ b/opencti-platform/opencti-graphql/src/manager/notificationManager.ts
@@ -6,7 +6,7 @@ import { type SizedNotifEvent, type StreamProcessor } from '../database/stream/s
 import { fetchRangeNotifications, storeNotificationEvent, createStreamProcessor } from '../database/stream/stream-handler';
 import { redisGetManagerEventState, redisSetManagerEventState } from '../database/redis';
 import { lockResources } from '../lock/master-lock';
-import conf, { booleanConf, logApp } from '../config/conf';
+import conf, { booleanConf, logApp, ACCOUNT_STATUS_ACTIVE } from '../config/conf';
 import { FunctionalError, TYPE_LOCK_ERROR } from '../config/errors';
 import { executionContext, INTERNAL_USERS, isUserCanAccessStixElement, isUserCanAccessStreamUpdateEvent, isUserInPlatformOrganization, SYSTEM_USER } from '../utils/access';
 import type { DataEvent, SseEvent, StreamNotifEvent, UpdateEvent } from '../types/event';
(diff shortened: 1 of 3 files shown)


tpasquet (commented):
Overall, the code looks good to me. Since I’m still new to the project, I think we should get an additional review from someone more familiar with the notification pipeline before merging.

This may not be necessary for this fix, but perhaps worth considering: should we also recheck the recipient’s current account status just before sending the notification?

A user could be active when the event is generated, then become locked or expired while it is queued or buffered. I may be missing an existing safeguard, but could that notification still be delivered?

Not necessarily a blocker for this PR, just something worth clarifying, or addressing in a follow-up if needed.

gabrielebonadiman (approved):

PR #17834, opened 2026-08-21
feat(chatbot): support human-in-the-loop tool approval in Ask Ariane (#17830)

### LINKED ISSUE #17830: feat(chatbot): support human-in-the-loop tool approval in Ask Ariane
## Use case

XTM One can now gate an agent's tool calls behind human approval: the turn pauses and waits for an operator decision before the tool runs. OpenCTI is the host of the Ask Ariane panel, so it has to carry that pause and the decision back and forth — otherwise a gated agent simply hangs from the platform's point of view.

## Current workaround

None. A turn that XTM One pauses for approval never resumes in OpenCTI.

## Proposed solution

Support the human-in-the-loop approval round trip in the chatbot integration:

- relay the approval-required pause and the operator's approve/deny decision through the chatbot proxy (`httpChatbotProxy.ts`, plus the route in `httpPlatform.js`),
- expose the agent's gating flag from the XTM One client (`xtm-one-client.ts`, `xtm-one.ts`),
- bump `@filigran/chatbot` to the release carrying the approval UI and wire it into `AskArianePanel`.

## Additional information

Upstream work this depends on:

- XTM One — issue https://github.com/XTM-One-Platform/xtm-one/issues/173, PR https://github.com/XTM-One-Platform/xtm-one/pull/2941 (per-tool human-in-the-loop approval for tool calls)
- Filigran UI — issue https://github.com/FiligranHQ/filigran-ui/issues/385, PR https://github.com/FiligranHQ/filigran-ui/pull/386 (collect human approval for gated tool calls mid-turn), shipped in `@filigran/chatbot` 3.9.0

Yes

fred-lr:
### Proposed changes

* Relay XTM One's approval pause and the operator's decision through the chatbot proxy: `POST /chatbot/messages/approve` (forwards the decisions body whole, passes upstream statuses through, requires a browser session so a token identity cannot approve its own proposal) and `GET /chatbot/conversations/:id/pending-approvals` (reload recovery — `approval_required` arrives once on the SSE stream).
* Never cache a turn that stopped for approval, or "I need approval before running X" gets pinned to an entity for the whole `ai:agents_refresh_timeout`.
* Declare `supports_approval_prompts` on XTM One registration so OpenCTI-contributed tools can gate normally.
* Bump `@filigran/chatbot` to 3.9.0 for the approval UI, and add the 129 widget i18n keys across all nine languages (second commit, separable).

### Related issues
* closes #17830

Upstream: XTM-One-Platform/xtm-one#2941, FiligranHQ/filigran-ui#386.

### How to test this PR

1. Point at an XTM One with per-tool approval enabled and ask Ask Ariane something that triggers a gated tool.
2. The turn pauses with an approval prompt — check Approve, Decline (with a reason), and "Yes, always".
3. Reload mid-prompt: the prompt comes back from `pending-approvals` and answering it still resumes the turn.
4. Against an XTM One with gating disabled, turns run straight through with no prompt.

### Checklist

### Further comments

Two sets of widget strings are left untranslated on purpose: the greeting and the relative-time labels are concatenated fragments that cannot be ordered correctly in another language. Both need an upstream fix in filigran-ui (one key with a placeholder); the time labels are hand-written as the closest idiomatic suffix in the meantime.

**Unattended callers.** AI Insights, NLQ search and playbook nodes never send `supports_tool_approval`, so a gated tool ends their turn with the plain approval message instead of a result. That is upstream's intended degradation, and playbooks already handle it: `parseStixBundle` rejects the prose and the node returns `output_port: 'unmodified'` with a warning, so the bundle passes through untouched.

The flag we declare gates OpenCTI-*contributed* tools; XTM One's own read-only tools, including STIX validation, are on the shipped whitelist and unaffected. What did need handling on our side is the cache — upstream calls this out explicitly — and that is what the no-cache guard and the read-time eviction in this PR are for.


opencti-platform/opencti-graphql/src/http/httpChatbotProxy.ts
```diff
@@ -312,6 +312,97 @@ export const postChatbotMessageSteer = async (req: Express.Request, res: Express
...
+// screen with the controls re-armed on anything else.
+export const postChatbotMessageApprove = async (req: Express.Request, res: Express.Response) => {
+  try {
+    const context = await authenticateAndVerify(req, res);
+    if (!context?.user) return;
+    // An approval is a person's consent, so it must come from a real browser
+    // session. Agent and service identities authenticate with an API token
+    // and have no business approving a tool call — least of all one they
+    // proposed themselves, which would defeat the whole control.
+    if (!context.user_with_session) {
```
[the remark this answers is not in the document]
fred-lr: Confirmed and fixed in f6f1db8. You're right about the precedence: `authenticateUserFromRequest` returns on the bearer branch (`src/domain/user.js:2244`) before it ever reaches the session branch at 2270, while `user_with_session` is set independently from `req.session?.user` in `httpAuthenticatedContext.js:86`. So a service token plus any user's cookie authenticated as the token identity and still passed the check.

Both routes now go through one guard that refuses any request carrying an `Authorization` header *and* requires the resolved identity to be the session's own:

```ts
const isBrowserSessionRequest = (req, context) => {
  if (req.headers.authorization) return false;
  if (!context.user_with_session) return false;
  const sessionUserId = req.session?.user?.id;
  return !!sessionUserId && sessionUserId === context.user?.id;
};
```

The widget only ever sends cookies (its one header override is `opencti-draft-id`), so nothing legitimate is affected. Two tests cover the combination you described and the mismatched-session case.

opencti-platform/opencti-graphql/src/modules/xtm/one/xtm-one.ts
```diff
@@ -62,6 +62,10 @@ export const registerWithXtmOne = async (context: AuthContext, user: AuthUser):
     enterprise_license_pem: pem,
     license_type: licenseType,
     business_vertical: 'cti',
+    // Ask Ariane renders `approval_required` prompts and posts verdicts back
+    // through `/chatbot/messages/approve`, so OpenCTI-contributed tools can
+    // gate normally instead of needing an administrator to exempt them.
+    supports_approval_prompts: true,
```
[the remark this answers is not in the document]
fred-lr: Added in f6f1db8, and the review prompted a second look at the value itself.

The flag is now derived rather than hardcoded: it is true only when the CGU is accepted and the licence validates — the same test `authenticateAndVerify` applies to every chatbot route. The flag promises somebody can be asked; with the panel unusable there is nobody to answer, and a gated tool would stall for the callers that never declare `supports_tool_approval` (AI Insights, playbook nodes).

Two tests now, one per side: the payload carries `true` on a licensed platform with the CGU accepted, and `false` when it is not.

fred-lr:
Pushed f6f1db8 addressing the review, including the suppressed comments — those were all correct, so recording the outcomes here since they have no thread to reply to.

**Suppressed comments, all fixed:**

* `de.json` — the waiting mini-game said `Wartungs-` (maintenance) against the off label's `Warte-`. Now `Das Warte-Minispiel aktivieren`.
* `ja.json` / `ko.json` — `Using tools…` rendered as "how to use tools" (`ツールの使用方法` / `도구 사용법`). Now `ツールを使用中…` and `도구 사용 중…`.
* Agent cache — right that the write guard cannot reach entries an earlier build wrote, since Redis outlives the deployment. Approval notices are now evicted on read via a new `redisDeleteXtmAgentResponse`, so the rollout clears them instead of replaying them for the rest of the TTL.

**Also found while re-reviewing:**

* `isApprovalRequiredMessage` only matched the notice at the start of the content. The agent can answer part of a question and *then* stop ("Here is what I found… I need approval before running `x`"), and a prefix test let that half-answer be cached with a stale approval request attached. Now matched anywhere, case-insensitively — caching one answer too few costs a re-run, caching this costs every user who opens the entity.
* `t("task")` / `t("tasks")` were missing from the dictionaries, so the background-task status labels degraded to `Collecte des résultats à partir de 3 tasks…` in French. Added across all nine files.

Verification: 72 unit tests pass across `httpChatbotProxy-test.ts` and `xtm-one-test.ts` (8 new), `tsc --noEmit` clean, ESLint clean, and all nine dictionaries at identical key sets (5543).

opencti-platform/opencti-front/src/private/components/chatbox/AskArianePanel.tsx
fellowseb, line 159: Not to change in this PR but it would've been better to put all the @filigran/chatbot related keys in separate JSON files to keep the separation clear. With this architecture we can't reliably say that all entries in the `lang/front` dictionaries must exist in the frontend code.

opencti-platform/opencti-graphql/src/database/redis.ts
```diff
@@ -1048,4 +1048,12 @@ export const redisSetXtmAgentResponse = async (cacheKey: string, content: string
...
   }
 };
+
+export const redisDeleteXtmAgentResponse = async (cacheKey: string): Promise<void> => {
+  try {
+    await getClientBase().del(`${XTM_AGENT_CACHE_KEY_PREFIX}${cacheKey}`);
+  } catch (err) {
+    logApp.warn('[XTM One] Agent response cache eviction failed', { cause: err });
+  }
+};
```
fellowseb, line 1058: Note: I'll open a RFC (and PR with instructions) to suggest moving all these to the dedicated module (in a [moduleName]-redis.ts file).
Not for this PR though.

opencti-platform/opencti-graphql/src/http/httpChatbotProxy.ts
```diff
@@ -312,6 +312,123 @@ export const postChatbotMessageSteer = async (req: Express.Request, res: Express
...
+// the bearer-token branch before it ever looks at the session (`domain/user.js`).
+// A request carrying a service token *and* any user's cookie therefore
+// authenticates as the token identity while still looking session-backed — which
+// would let an agent identity approve the very call it proposed. Require that the
+// identity actually resolved from the session, by matching it against the session
+// user and refusing any request that presents an Authorization header at all.
+const isBrowserSessionRequest = (req: Express.Request, context: AuthContext): boolean => {
+  if (req.headers.authorization) return false;
+  if (!context.user_with_session) return false;
+  const sessionUserId = (req as Express.Request & { session?: { user?: { id?: string } } }).session?.user?.id;
```
fellowseb: Consider instead enhancing the `Request` interface in `express.d.ts` 🙏

opencti-platform/opencti-graphql/src/http/httpChatbotProxy.ts
```diff
@@ -312,6 +312,123 @@ export const postChatbotMessageSteer = async (req: Express.Request, res: Express
...
+//
+// `context.user_with_session` alone does not answer that: it only records that
+// a session cookie was present, while `authenticateUserFromRequest` returns on
+// the bearer-token branch before it ever looks at the session (`domain/user.js`).
+// A request carrying a service token *and* any user's cookie therefore
+// authenticates as the token identity while still looking session-backed — which
+// would let an agent identity approve the very call it proposed. Require that the
+// identity actually resolved from the session, by matching it against the session
+// user and refusing any request that presents an Authorization header at all.
+const isBrowserSessionRequest = (req: Express.Request, context: AuthContext): boolean => {
```
fellowseb: Could be moved to `httpUtils.ts` right ? Without the context-specific (agents...) details in the comment.

opencti-platform/opencti-graphql/src/http/httpChatbotProxy.ts
```diff
@@ -312,6 +312,123 @@ export const postChatbotMessageSteer = async (req: Express.Request, res: Express
...
+    }
+    const jwt = await issueXtmJwt(context.user, XTM_ONE_URL);
+    const httpClient = getXtmClient('json', {
+      Authorization: `Bearer ${jwt}`,
+      'Content-Type': 'application/json',
+    });
+    const response = await httpClient.get(`/api/v1/platform/chat/conversations/${conversationId}/pending-approvals`, {
+      timeout: DEFAULT_XTM_TIMEOUT,
+    });
+    res.status(response.status).json(response.data);
```
fellowseb, line 404: What's the shape of the XTM One Api response ? Is there documentation about it somewhere ?

opencti-platform/opencti-graphql/src/http/httpChatbotProxy.ts
fred-lr: Agreed — `src/types/express.d.ts` already declares `Request.session` (`nonce`, `referer`), so the cast was re-declaring a subset of it locally. Adding `user?: { id?: string }` there lets line 329 read `req.session?.user?.id` directly, and any other handler that needs the session user gets it typed for free.

opencti-platform/opencti-graphql/src/http/httpChatbotProxy.ts
fred-lr: Right, nothing in it is chatbot-specific — it answers "did this identity resolve from a browser session?", which any route may need. Moving it to `httpUtils.ts` and keeping only the general rule in the comment: `authenticateUserFromRequest` returns on the bearer branch before it reaches the session branch, so `user_with_session` alone doesn't prove a browser. The chatbot-specific half — that this is what stops an agent identity approving its own proposal — belongs at the two call sites, and stays there.

opencti-platform/opencti-graphql/src/http/httpChatbotProxy.ts
fred-lr: Yes, documented on the XTM One side. The response is:

```json
{
  "conversation_id": "<uuid>",
  "proposals": [
    {
      "tool_call_id": "call_abc",
      "tool_name": "delete_entity",
      "tool_description": "…",
      "arguments": { "id": "…", "cascade": true },
      "input_schema": { },
      "source": "opencti"
    }
  ],
  "turn": "running" | "idle"
}
```

An empty `proposals` is the ordinary answer — most conversations aren't paused — so it isn't an error case. `tool_call_id` is the match key a decision has to name; the tool name can't be, since one turn can hold several calls to the same tool. `turn` exists because a decision on a *recovered* prompt resumes a turn whose stream the reload already closed: the widget polls while it reads `running` and fetches the conversation once it reads `idle`. Ownership failures are 404, not 403, so the endpoint never confirms someone else's conversation exists — which is why the handler passes upstream statuses through untouched.

Docs: deployment/[chat-approvals-in-clients.md](https://github.com/XTM-One-Platform/xtm-one/blob/feat/173-per-tool-approval/docs/docs/deployment/chat-approvals-in-clients.md) in the XTM One docs covers the client contract; the full wire format is in `specs/006-per-tool-approval/contracts/chat-approval-api.md`; and the live OpenAPI is at `{xtm_one_url}/api/openapi.json`.

opencti-platform/opencti-front/src/private/components/chatbox/AskArianePanel.tsx
fred-lr: Agreed, and the cost is exactly the one you name: with widget keys mixed into `lang/front`, "every key must be used in the front-end code" stops being a property anything can assert — a key retired by a widget bump just sits there. Every version bump adds to that. A separate `lang/chatbot/` set would keep both invariants checkable independently.

opencti-platform/opencti-graphql/src/database/redis.ts
fred-lr: Sounds good — `redisSetXtmAgentResponse` / `redisGetXtmAgentResponse` / `redisDeleteXtmAgentResponse` are a self-contained block and should move together when the RFC lands.

opencti-platform/opencti-front/src/private/components/chatbox/AskArianePanel.tsx
fellowseb: I created https://github.com/OpenCTI-Platform/opencti/issues/17910 as a follow-up on that matter.

opencti-platform/opencti-graphql/src/http/httpChatbotProxy.ts
fellowseb: I was just wondering if we should add minimal shape check of the response and forward only known fields; in order to reduce the API footprint and prevent potential malicious activity if XTM One servers are corrupted (I don't have a scenario in mind though).

fred-lr: This is a valid concern and we could investigate it in a follow up issue as it is not something new to this PR.
What makes it complicated here is that 3 players are concerned:
- OpenCTI chat proxy
- filigran-ui chatbot
- XTM One

fellowseb: True, resolving

fellowseb (approved):

PR #17857, opened 2026-08-24
fix(sync): handle oversized sync attachment payloads (#17778)

### LINKED ISSUE #17778: fix(sync): detect large files/images attached to entities and skip retrying when ERR_STRING_TOO_LONG raised
## Description

This bugfix request is the result of an issue spotted during on-call.
The customer instance is receiving the following errors:

```
{"category":"APP","cause":{"code":"UNKNOWN_ERROR","message":"Cannot create a string longer than 0x1fffffe8 characters","name":"Error","stack":"Error: Cannot create a string longer than 0x1fffffe8 characters\n    at Object.slice (node:buffer:689:37)\n    at Buffer.toString (node:buffer:868:14)\n    at Object.start (/opt/opencti/build/back.js:274661:157)\n    at process.processTicksAndRejections (node:internal/process/task_queues:103:5)"},"id":"83fc00e2-14ab-4971-bdd0-b3db559a1fdf","level":"error","manager":"SYNC_MANAGER","message":"[OPENCTI-MODULE] Sync manager event handling error, retrying...","source":"backend","timestamp":"2026-07-30T08:06:42.891Z","version":"7.260715.0"}
{"category":"APP","level":"info","message":"[OPENCTI] Sync 83fc00e2-14ab-4971-bdd0-b3db559a1fdf: stopping manager","source":"backend","timestamp":"2026-07-30T08:06:42.917Z","version":"7.260715.0"}
{
```

The root cause seems to be a very large attached file / markdown image resulting in too large string when calling this line: https://github.com/OpenCTI-Platform/opencti/blob/master/opencti-platform/opencti-graphql/src/manager/syncManager.js#L322.

## Proposed mitigation/solution

- Detect large files in `transformDataWithReverseIdAndFilesData`: add error logs in such case
- Catch ERR_STRING_TOO_LONG in that loop: stop retrying and add an error log in such case. Or fallback to sync the entity without the (large) attached files.

## Environment

2. Version: { e.g. 1.2.3 }
3. Other environment details:

## Reproducible steps

N/A

## Additional information

fellowseb:
### Proposed changes

* During Sync, on the receiving end, catches (`ERR_STRING_TOO_LONG`) exceptions (`Cannot create a string longer than`) when stringifying to base-64 attached files. When it happens, logs an errors and retries without the attached file. Also treated the case when the `JSON.stringify` could fail because the total payload exceeds 500~MB; throwing a `RangeError: Invalid string length`.

* Cases handled:
  * A single file is larger than the limit for `Buffer#toString('base64')` (`ERR_STRING_TOO_LONG`)
  * The base-64 encoding of the total payload exceeds the same limit (`ERR_STRING_TOO_LONG`)
  * The stringification of the total payload is larger than the JSON.stringify limit (`RangeError: Invalid string length`)
 
* This is a mitigation while working on a better solution.

### Related issues

* closes #17778

### How to test this PR

* Set up a sync between two platforms: on the source go to "Data Sharing" and create a public Stream. On the target, go to "Integrations" and create a "OpenCTI Streams" built-in connector (make sure you enter the admin token of source). Select "Perfect stream" option.
* Create a new Note entity with a large attached file that when base-64 encoded gets larger than 500MB by using the following command: `head -c 351000000 /dev/zero > ~/Downloads/sync-err-string-too-long-351MB.bin`.
* Witness the Note arriving on the receiving end and the sync process NOT being blocked (i.e. you can continue sending other entities).

Repeat the process with another case: attach multiple files with the sum larger than 500MB (ex: 5 x 150MB). You should fall in the `RangeError: Invalid string length` case.

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/manager/syncManager.js
+++ b/opencti-platform/opencti-graphql/src/manager/syncManager.js
@@ -36,6 +36,73 @@ const FILE_FETCH_TIMEOUT = conf.get('sync_manager:file_fetch_timeout') || 300_00
 
 const waitLoopTimer = new InterruptibleTimer();
 
+const isStringTooLongError = (error) => {
+  const errorMessage = error?.message ?? '';
+  return error?.code === 'ERR_STRING_TOO_LONG'
+    || errorMessage.includes('Cannot create a string longer than');
+};
+
+const dropAttachedFilesData = (syncData) => {
+  const files = syncData?.extensions?.[STIX_EXT_OCTI]?.files;
+  if (!Array.isArray(files) || files.length === 0) {
+    return [];
+  }
+
+  const droppedFiles = [];
+  for (let index = 0; index < files.length; index += 1) {
+    const file = files[index];
+    if (typeof file?.data === 'string' && file.data.length > 0) {
+      droppedFiles.push({
+        fileUri: file.uri,
+        dataLength: file.data.length,
+      });
+      delete file.data;
+    }
+  }
+  return droppedFiles;
+};
+
+const encodeEventPayloadToBase64 = (payload) => Buffer.from(payload, 'utf-8').toString('base64');
+
+const buildSyncEventContent = ({
+  syncId,
+  lastEventId,
+  eventType,
+  syncData,
+  eventContext,
+  encodeToBase64 = encodeEventPayloadToBase64,
+  logger = logApp,
+}) => {
+  const buildPayload = () => JSON.stringify({
+    id: lastEventId,
+    type: eventType,
+    data: syncData,
+    context: eventContext,
+  });
+  try {
+    return encodeToBase64(buildPayload());
+  } catch (encodingError) {
+    if (!isStringTooLongError(encodingError)) {
+      throw encodingError;
+    }
+    const droppedFiles = dropAttachedFilesData(syncData);
+    if (droppedFiles.length === 0) {
+      throw encodingError;
+    }
+    logger.error('[OPENCTI] Sync: Event payload too large, dropping attached files data and retrying.', {
+      id: syncId,
+      eventId: lastEventId,
+      entityId: syncData?.extensions?.[STIX_EXT_OCTI]?.id,
+      entityType: syncData?.extensions?.[STIX_EXT_OCTI]?.type,
+      droppedFilesCount: droppedFiles.length,
+      droppedFiles,
+      code: encodingError.code,
+      message: encodingError.message,
+    });
+    return encodeToBase64(buildPayload());
+  }
+};
+
 const hasEmbeddedStorageRef = (markdown) => {
   return markdown.includes('embedded/')
     || markdown.includes('/storage/get/embedded/')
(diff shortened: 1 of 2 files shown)


opencti-platform/opencti-graphql/src/manager/syncManager.js
```diff
@@ -34,6 +34,73 @@ const FILE_FETCH_TIMEOUT = conf.get('sync_manager:file_fetch_timeout') || 300_00
...
+  });
+  try {
+    return encodeToBase64(buildPayload());
+  } catch (encodingError) {
+    if (!isStringTooLongError(encodingError)) {
+      throw encodingError;
+    }
+    const droppedFiles = dropAttachedFilesData(syncData);
+    if (droppedFiles.length === 0) {
+      throw encodingError;
```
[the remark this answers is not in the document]
fellowseb: We're not treating the case where the content of the event payload itself (without attached files and markdown images) or markdown embedded images themselves would trigger ERR_STRING_TOO_LONG or RangeError errors in this PR.

opencti-platform/opencti-graphql/src/manager/syncManager.js
```diff
@@ -361,8 +449,13 @@ const syncManagerInstance = (syncId) => {
             while (!processed && running) {
               try {
                 const { data: syncData, previous_standard } = await transformDataWithReverseIdAndFilesData(sync, httpClient, stixData, eventContext);
-                const enrichedEvent = JSON.stringify({ id: lastEventId, type: eventType, data: syncData, context: eventContext });
-                const content = Buffer.from(enrichedEvent, 'utf-8').toString('base64');
+                const content = buildSyncEventContent({
```
gabrielebonadiman, line 452: if ` return encodeToBase64(buildPayload());` in the `catch (encodingError)` in the `buildSyncEventContent` fails (double fail for whatever the reason) then I think that this become an infinite loop as it's in a `while(true)` loop

```
const buildSyncEventContent = ({
  syncId, lastEventId, eventType, syncData, eventContext,
  encodeToBase64 = encodeEventPayloadToBase64, logger = logApp,
}) => {
  const buildPayload = () => JSON.stringify({
    id: lastEventId, type: eventType, data: syncData, context: eventContext,
  });
  try {
    return encodeToBase64(buildPayload());          // attempt 1
  } catch (encodingError) {
    if (!isStringTooLongError(encodingError)) {
      throw encodingError;
    }
    const droppedFiles = dropAttachedFilesData(syncData);
    if (droppedFiles.length === 0) {
      throw encodingError;
    }
    logger.error('[OPENCTI] Sync: Event payload too large, dropping attached files data and retrying.', { ... });
    return encodeToBase64(buildPayload());          // attempt 2 — NOT wrapped in try/catch
  }
};
```

```
let processed = false;
while (!processed && running) {
  try {
    const { data: syncData, previous_standard } = await transformDataWithReverseIdAndFilesData(...);
    const content = buildSyncEventContent({ syncId, lastEventId, eventType, syncData, eventContext });
    await pushBundleToWorker(...);
    await saveCurrentState(context, 'event', lastEventId);
    processed = true;
  } catch (processingError) {
    logApp.error('[OPENCTI-MODULE] Sync manager event handling error, retrying...', { cause: processingError, ... });
    await patchSync(context, SYSTEM_USER, syncId, { last_execution_date: ..., last_execution_status: 'error' });
    await syncLogger.error('Feed execution failed', ...);
    await wait(5000);   // no backoff, no max attempts
  }
}
```

gabrielebonadiman:
just for visibility - this is a mitigation (drop file data, keep syncing), not the fix. #17896 tracks the actual solution (stream file refs instead of embedding base64). Worth linking here so we remember to remove/simplify this path once that lands, rather than carrying both indefinitely.

opencti-platform/opencti-graphql/src/manager/syncManager.js
fellowseb: Hey @gabrielebonadiman thanks for the review,
Yeah it's the same comment Copilot made x). My main point was that I wanted to mitigate the large attached files case in this PR, not the whole situation.
If `return encodeToBase64(buildPayload());` still fails without the attached files it means:
- the payload of the event itself is ungodly large
- the markdown embedded images are ungodly large
- there's another system or programming error
For the first and third items: I'd prefer seeing the real case so we can fix it at the source (unlikely IMO).
For the second item: yes it should be mitigated but I simply don't want to do too many things in this PR.
Would removing all embedded images be a good mitigation ? 🤔 
I _could_ skip all items on error but I'm afraid that would just result in silently incomplete syncs.
WDYT ?

(Anyways if we were to skip we should do it only when `buildSyncEventContent` throws and let `pushBundleToWorker` failures retry)

gabrielebonadiman (approved):

Issue #163, opened 2019-07-27
Worker does not send ack when processing a long running task


SamuelHassine:
## Problem to Solve

If the import or export message implies a long-running task, the worker is not able to send the ack to RabbitMQ. That leads to process messages multiple times and infinite processing.

## Current Workaround

None.

## Proposed Solution

Implement a threaded process and use the pika `connection.sleep` to maintain the connection as [described here](https://github.com/pika/pika/issues/930).

## Additional Information

None.

richard-julien:
Agree. A threaded model will also permits to use the basic_consume for each queue. For me we need to have the control in the main thread (list queue, create thread, etc), and a specific thread for each queue we need to consume. With this approach we also should be able to display the number of active worker in the UI.

SamuelHassine:
Deal :)

Issue #162, opened 2019-07-26
Refactor the observables schema to match STIX2 references


SamuelHassine:
## Problem to Solve

Currently, the schema of observables is not fully compliant to [STIX2 (cyber observables)](https://docs.google.com/document/d/1IvkLxg_tCnICsatu2lyxKmWmh1gY2h8HUNssKIE-UIA/edit#heading=h.p49j1fwoxldc).

## Current Workaround

None.

## Proposed Solution

Refactor the schema and adapt the GraphQL API and the frontend display to the new attributes.

## Additional Information

None.

PR #161, opened 2019-07-26
[build] Add environment variables to the CircleCI workflow (#160)

SamuelHassine:
Add the environment variables for passing tests in the CircleCI workflow.

Issue #160, opened 2019-07-26
CircleCI tests not passing from PR


SamuelHassine:
## Description

The CircleCI tests are not passing when using PR because of the needs of environment variables that are provided only for building branches. 

## Environment

1. CircleCI
2. OpenCTI 1.1.0

PR #17884, opened 2026-08-24
feat(fds): convert the OpenCTI selection fields onto the design system (#17799)

sandyghs:
**DO NOT MERGE without Sandy's visual pass.**

## What is here

Every OpenCTI selection field that could be converted now composes
`filigran-design-system` instead of MUI: **275 of 287 mounts**, across 200+ files.
Library pin `bd076e8f`, taken from the base.

The 12 that remain each carry a reason — 2 legacy adapters that must outlive
their consumers, and 10 parked. Five multi-value selects are parked on top of
those and stop counting, because the census counts an adapter's mount once at the
adapter file instead of at each consumer; that is also why the denominator is
derived and reads 287. Full table and the arbitration list:
`fds-migration/CENSUS.md`.

```bash
node fds-migration/scripts/census-selection-fields.mjs opencti-platform/opencti-front/src
```

The census resolves each imported symbol's local name per file before counting.
Two earlier hand-carried counts were wrong — one missed
`import Select from '@mui/material/Select'` (37 mounts), the next missed the
aliased and default-plus-named forms (14 more). Every figure here comes from the
script.

## Accessibility is a gate, not a review note

The migration introduced a WCAG 4.1.2 defect and it is now fixed and guarded.
MUI expressed a Select's label association through `<Select label={...}>`, an
MUI-only prop dropped at conversion without carrying the text anywhere, and the
visible `<InputLabel>` was often left behind naming nothing.

```bash
node fds-migration/scripts/check-accessible-names.mjs
```

Red-before-fix: 20 findings — 12 Selects announced as "combobox" and nothing
else, 8 orphan InputLabels. Now clean, and it **runs in CI** as a step in the
frontend quality job, so it cannot regress unnoticed. It also covers Combobox
(no `ComboboxLabel` means `labelPosition="none"` plus an `aria-label`) and the
Select panel's own name.

Rule 1 was structurally dead until this round: its aria test scanned the whole
`<Select>` block, so the `aria-label` the panel rule mandates satisfied it.
Proven by mutation — removing a `SelectLabel` reported clean — and the repaired
rule immediately found two sites the dead one had hidden.

Sweeping for it found three Comboboxes with no accessible name at all
(`AIInsights`, `ResponseDialog`, `JsonFormArrayRenderer`) — each now named.

## Behaviour kept

- **`FilterAutocomplete`** — alt-click adds a filter NEGATED (`not_eq`/`not_nil`).
  That rides on the DOM event, which `ComboboxChangeMeta` carries.
- **`CustomField*`** — `freeSolo` with `options={[]}` was a tag entry, not a
  picker, so `allowCustomValue`.
- **`AutocompleteFreeSoloField`** — MUI offered a CLICKABLE suggestion row, so
  `onCreateOption`, not `allowCustomValue` (which renders a non-actionable hint).
- **Feed Relationship/Direct** — the toggle keeps two tones, and the chip is now
  wrapped in a button with `aria-pressed`, so it announces its state.

Two pre-existing defects surfaced and are recorded rather than silently fixed:
all four free-solo consumers submit `values[name].map((v) => v.value)`, so MUI's
bare-string Enter path would have submitted `undefined`; and
`VocabularyCreation`/`VocabularyEdition` passed a `classes.autoCompleteIndicator`
they never defined, so their clear button was showing.

## A red E2E is not proof the conversion is wrong

`Dashboard CRUD` was blamed on the widget Select cluster for three rounds, with a
clean A/B/A — converted red, reverted green, re-converted red. It was still the
wrong conclusion. The real error was a strict-mode violation:
`getByRole('combobox', { name: 'Attribute' })` matched TWO fields, because
`getByRole` matches the accessible name by SUBSTRING and the neighbour is named
"Date attribute" — a name that only existed once the accessible-name work gave
that Select a real label. The fix went in the selector, never in the labels.

## Chips

12 of 14 remaining MUI chips in these files now use the library Chip, tones read
from the code they replace: `theme.palette.error.main` and `success.main` →
`critical`/`low`; a `getChipColor()` returning `error|success|warning` →
`critical|low|high`; `color="warning"` → `high`; no colour → the neutral default.
Two stay on MUI and are listed for arbitration — one needs `onDelete`, the other
passes JSX where the library's `label` is typed `string`.

## CI

All 28 checks pass on `d8a72eecf2`, including `e2e group0` and `e2e group1`.

Two independent failures were masking each other in group1, which is why each
was misattributed in turn:

1. **The widget Select cluster.** `DashboardRelativeDateSelect` defaults
   `labelId='relative'` and renders `<InputLabel id="relative">Relative time`.
   MUI's `WidgetCreationParameters` declared three more `id="relative"` labels
   whose Selects each set `labelId="relative"`. Duplicate DOM ids resolve to the
   first match, so several unrelated Selects were all announced "Relative time" —
   a real pre-existing WCAG defect, and the name `DashboardWidgets.pageModel.ts`
   was written against. Converting gave each field its own name and broke a
   locator that depended on the bug. Fixed in the page model.
2. **The Author create row.** `CreatedByField` is parked on MUI, so
   `incidentResponse.spec.ts` could not call `createOption` on it — that
   affordance only exists on a converted mount. Restored to the MUI `+` flow.

**Still DO NOT MERGE** until the visual pass.


sandyghs:
Coordination on the design-system pin, from the parallel Button/Chip wave — no action needed on this PR's scope.

The FDS pin is being bumped `fb8c3cc` → **`83295d9`** in #17906 (issue #17905), as its own step. It carries #154 (Chip data colours), #156, #160 (ButtonGroup), #161, #162, #163 and #165.

Two things relevant to you:

- **Same pin for both waves.** Referencing the SHA here so the Combobox pilot and the Button/Chip wave land on one dependency state rather than two. #163 in particular touches Combobox (`labelPosition` + the `ComboboxLabel` required indicator), so it is worth knowing which side of the bump your visual proofs were taken on.
- **Your files are not touched, and will not be.** The Button/Chip wave excludes every file in this PR by path — re-derived today at 107 files, up from 36 when the wave was scoped. The one place the two waves could have collided is the shared `components/common/button/Button.tsx` wrapper: rewriting it would change rendering at ~170 call sites inside your 107 files without editing any of them, which would age your visual proofs under you. **That rewrite is deliberately held until this PR merges** (design lead decision), so nothing moves under your feet.

PR #17892, opened 2026-08-25
feat(merge-users): RBAC rights merge and union/strict option (#17496)

### LINKED ISSUE #17496: PR4 — RBAC rights merge + union/strict option
### Objective

Merge the source user's rights into the target with deduplication, and implement the union vs strict option from the decision log. This is the security-critical part: it must avoid both privilege escalation and silent access loss. It also covers the user's Individual, because that entity is part of the visibility filter and not a cosmetic profile card — a blind spot of the register, tracked here.

### The user's Individual — access control, not decoration

Each user owns an Individual (a STIX entity), created lazily by resolveUserIndividual on the first Note / Opinion / Case-Incident / Feedback authored without an explicit createdBy. It carries contact_information = user_email.

The link is not a stored identifier: individual_id is recomputed on every session build by joining contact_information against user_email. The email is the join key, which is why no register row can describe this link — it exists in no document. It is a genuine blind spot of the register, not an omission at a given row.

The Individual feeds the access filter: a user can see their own Individual and everything that Individual created (should clauses on internal_id and created-by in the org-restricted visibility filter, plus a direct grant in isOrganizationUnrestricted). Past Notes and Opinions carry createdBy = source Individual. Left untouched, the target resolves its own Individual from its own email and therefore cannot see that content — while PR3 has just made it the creator_id owner. The merge would produce a user who owns objects they cannot read. Collaborative ownership checks on Note and Opinion (createdBy === user.individual_id) fail on the whole history for the same reason.

### Out of scope

- STIX operational refs (PR5).
- Deleting the source / tokens (PR7, PR8).
### Dependencies

- PR2 (engine).
### Definition of Done

- Both strategies coded and tested (union and strict).
- Dedup verified (no duplicate memberships/roles).
- Dry-run surfaces every RBAC risk; hard-block cases prevent the real run.
- After the merge, the target can see the Notes and Opinions authored by the source, and passes the collaborative ownership check on them.
- Both branches are tested: target without an Individual (re-point) and target with one (merge).
- Exactly one Individual carries the target email, and none carries the source email.
- Reuses PoC#4.

ludovic:
### Proposed changes

* Merge the group and organization memberships of the source user into the target, with a `rightsStrategy` option: `STRICT` keeps the target rights untouched and drops the source ones, `UNION` adds the source rights on top. Memberships already held by the target are never duplicated, which keeps a replay a no-op.
* Remove the rights relations that should never start from a user (`has-role`, `has-capability`, `has-capability-in-draft`, `accesses-to`). They are reported as an alert and never reproduced on the target: copying a capability straight onto a user would grant a privilege the platform itself never grants.
* Transfer the authorized member entries of the source. A member can be listed several times under different group restrictions, so each source entry is carried over on its own and only collapses into a target entry sharing the same restriction, keeping the strongest access right. The merge is refused unless the operator acknowledges it when the target has no `KNOWLEDGE_KNUPDATE_KNMANAGEAUTHMEMBERS` capability.
* Rewrite `authorized_authorities` on organizations, scoped to organizations because the same field carries capability names on feeds, collections and background tasks.
* Report the rights differences in the dry-run: markings and organizations gained by the target, and those that were only reachable through the source.
* Merge the individual of the source user into the target's, so the target can read the objects it inherits from PR3 and passes the collaborative ownership check on them.
* Compute the public sharing exposure from the strategy instead of the current target session, and cover the endpoints already configured with the target, whose exposure also widens under `UNION`. The comparison covers the four sets `buildDataRestrictions` reads — markings, organizations, capabilities and groups — and is counted on the endpoints that are actually anonymous, the reference being rewritten on the private ones too. The authorized member entries and the individual carried over are stated rather than measured: they widen what the endpoint reads without changing any of those sets.

### Related issues

* closes #17496

### How to test this PR

1. Enable the feature flag: `APP__ENABLED_DEV_FEATURES='["MERGE_USERS"]'`.
2. Create two users in different groups, put the source user in an organization, and give the target a marking the source does not have.
3. Run the mutation in dry run with `rightsStrategy: STRICT`, then with `UNION`:
```graphql
mutation {
  userMerge(sourceId: "<source>", targetId: "<target>", options: { dryRun: true, rightsStrategy: UNION }) {
    status
    report { total_updated handlers { handler updated changes { register_row_id detail count } alerts { register_row_id kind message blocking } } }
  }
}
```
Under `STRICT` the `granted to the target` counts stay at zero; under `UNION` they match the source memberships, and the report lists the markings and organizations the target gains.
4. Run it for real, check that the target holds both sets of memberships, that the source holds none, and that replaying reports `total_updated: 0`.
5. Set `contact_information` of an individual to the source email and check that after the merge exactly one individual carries the target email and none carries the source one.

### Checklist

### Further comments

The rights projection is computed by the engine and passed in the handler context rather than read by the handlers. Every handler computes before any handler writes, so a handler asking the platform what the target can access only ever gets the pre-merge answer, and reading the rights handler output directly would break the disjunction rule between handlers.

Memberships go through the domain layer instead of a bulk index rewrite: they are denormalized onto both endpoints as `rel_<type>`, and a raw write would leave those arrays describing a membership that no longer exists.

The individual merge is covered by no register row, by construction: `individual_id` is recomputed at every session build by joining `contact_information` against `user_email`, so the link exists in no document. It is handled by a separate handler with an empty coverage.

This PR is stacked on #17577.


CelineSebe:
I tested the user merge flow with the UNION strategy, both in dry-run and for real.

The mutation returned SUCCESS. The dry-run correctly reported the organization and group memberships that would be transferred, with no blocking alerts or public-sharing exposure. After the real merge, the target user had successfully inherited the source user’s organization membership, while the source user no longer had it. The source user remaining present is expected for this PR.
I also tested the Individual handling: after associating an Individual with the source user’s email and running the merge, the Individual was correctly re-pointed to the target user’s email.

The tested behavior works as expected.
So, I approve this PR

CelineSebe (approved):

opencti-platform/opencti-graphql/src/modules/userMerge/userMerge-rightsHandler.ts
```diff
@@ -0,0 +1,347 @@
...
+    const restrictedMembers = await readRestrictedMembersPlan(context, sourceId, targetId);
+    changes.push({ register_row_id: RESTRICTED_MEMBERS_ROW, entity_type: '*', count: restrictedMembers.elements, exact: true, detail: 'transferred to the target' });
+    if (restrictedMembers.raised > 0) {
+      alerts.push(restrictedMembersAlert(restrictedMembers.raised));
+    }
+    alerts.push(...rightsGapAlerts(rights.source, rights.target, options.rightsStrategy, rights.labels));
+    const canManage = rights.target.capabilities.includes(BYPASS) || rights.target.capabilities.includes(KNOWLEDGE_KNUPDATE_KNMANAGEAUTHMEMBERS);
+    if (restrictedMembers.elements > 0 && !canManage) {
+      alerts.push(managementCapabilityAlert(restrictedMembers.elements));
+    }
```
fellowseb, line 300: Here we're using the target's pre-merge rights to emit an alert, right ? Shouldn't we be working with the rights after merge ?

opencti-platform/opencti-graphql/src/modules/userMerge/userMerge-rightsHandler.ts
```diff
@@ -0,0 +1,347 @@
...
+    return { handler: USER_MERGE_RIGHTS_HANDLER, changes, alerts };
+  },
+  apply: async ({ context, sourceId, targetId, options, sourceUser, targetUser }: UserMergeHandlerContext, plan: UserMergeHandlerPlan): Promise<number> => {
+    let updated = 0;
+    for (let i = 0; i < MEMBERSHIP_EDGES.length; i += 1) {
+      const edge = MEMBERSHIP_EDGES[i];
+      const edgePlan = await readEdgePlan(context, edge, sourceId, targetId, options.rightsStrategy);
```
fellowseb, line 307: Why are we calling `plan` again here and not using the `plan` ?

opencti-platform/opencti-graphql/src/modules/userMerge/userMerge-rightsHandler.ts
```diff
@@ -0,0 +1,347 @@
...
+        updated += 1;
+      }
+      for (let removed = 0; removed < edgePlan.removed.length; removed += 1) {
+        await deleteElementById(context, SYSTEM_USER, edgePlan.removed[removed].id, ABSTRACT_INTERNAL_RELATIONSHIP);
+        updated += 1;
+      }
+    }
+    for (let i = 0; i < DERIVED_EDGES.length; i += 1) {
+      const edge = DERIVED_EDGES[i];
+      const relations = await readDerivedEdge(context, edge, sourceId);
```
fellowseb, line 319: Same question for calling `readDerivedEdge` again

opencti-platform/opencti-graphql/src/modules/userMerge/userMerge-publicSharingHandler.ts
```diff
@@ -135,18 +241,45 @@ export const userMergePublicSharingHandler: UserMergeHandler = {
...
+          kind: 'exposure',
+          message: exposureMessage(target, keptPublicCount, kept, rights.labels, 'already configured with the target user have their exposure changed'),
+          blocking: isExposureWidening(kept),
+        });
+      }
+      if (publicCount > 0) {
+        alerts.push({
+          register_row_id: target.registerRow,
+          kind: 'exposure',
+          message: memberAccessMessage(target, publicCount, 'target', mergesTargetIndividual, 'transferred to the target user'),
```
fellowseb, line 274: Claude suggests to use `mergesSourceIndividual || mergesTargetIndividual` as later in the `individualHandler` we keep source's individual when/if there's no individual for the target.

opencti-platform/opencti-graphql/src/modules/userMerge/userMerge-rightsHandler.ts
ludovic: The plan object is the high-level report serialized for GraphQL and the journal (UserMergeHandlerPlan), so it only carries aggregated counters (changes: [{ count, register_row_id }]), not the actual relationship IDs. Re-fetching readEdgePlan here gives us the concrete internal_id and toId needed for createRelation and deleteElementById.

opencti-platform/opencti-graphql/src/modules/userMerge/userMerge-rightsHandler.ts
ludovic: Same

opencti-platform/opencti-graphql/src/modules/userMerge/userMerge-publicSharingHandler.ts
ludovic: The message specifically announces what the endpoint will ALSO serve (the marginal gain). A transferred endpoint was already served with the source user's access, so it was already serving what the source's individual created. Moving to the target only exposes new content if the target user actually had its own individual (mergesTargetIndividual). If the target has none and we repoint the source's, the endpoint serves the same individual content as before, so there is no extra exposure to report.

opencti-platform/opencti-graphql/src/modules/userMerge/userMerge-rightsHandler.ts
ludovic: Good catch! You're totally right, under UNION the target will inherit the capability if the source has it, so testing rights.projected.capabilities is much more accurate than rights.target.capabilities. I'll update it

marieflorescontact:
I tested locally. I hope my comments will be relevant :
when I run `userMerge` query with `dryrun = false`,  I loose source organization (I expected that target user being part of the source Orga  after the merge). 
For the trigger, wich is correctly transmited to target user, the `creator `is still `source user.` is it correct ?
target 2

ludovic:
> I tested locally. I hope my comments will be relevant :
> when I run `userMerge` query with `dryrun = false`, I loose source organization (I expected that target user being part of the source Orga after the merge).
> For the trigger, wich is correctly transmited to target user, the `creator `is still `source user.` is it correct ?

Regarding the setup, did you run the merge with the STRICT option (the default)? In that case, the result is indeed what’s expected. As for the trigger, it will be taken into account with the PR9.

marieflorescontact:
> > I tested locally. I hope my comments will be relevant :
> > when I run `userMerge` query with `dryrun = false`, I loose source organization (I expected that target user being part of the source Orga after the merge).
> > For the trigger, wich is correctly transmited to target user, the `creator `is still `source user.` is it correct ?
> 
> Regarding the setup, did you run the merge with the STRICT option (the default)? In that case, the result is indeed what’s expected. As for the trigger, it will be taken into account with the PR9.

yes, I used STRICT option. I was not sure what to expect for Organization behaviour. :)

fellowseb (approved):

PR #17916, opened 2026-08-26
feat(accessibility): Update the BulkRelationDialog to be a more accessible design (#17889)

### LINKED ISSUE #17889: feat(accessibility): Update the BulkRelationDialog to be a more accessible design
## Use case

The current dialog presents multiple issues related to accessibility:

- The screen reader doesn't state the entity that the relationship is from, making it harder for users to have appropriate context.
- The inputs are not labeled so users do not know what they are for
- The added items are not contained in a grid so keyboard users must tab through all inputs to navigate the list of added relationships
- No helper text or validation text is provided to alert users that they must create missing entities before submitting
- No helper text to clarify that users need to separate entities on new lines in the text area input 
- After creating missing entities focus resets and forces the user to navigate through all inputs again before pressing create
- The horizontal orientation of the form causes issues when using browser zoom or when using smaller screens

The layout should be revised to resolve these issues.
## Current workaround

None

## Proposed solution

Rearrange the layout to be vertically oriented and resolve issues above.

- Remove the use of the card display for the from entity and replace with accessible text
- Label all inputs
- add helper text to the "To" text area to clarify expected input
- Add visually persistent create missing entities button below the text area that becomes enabled if there are entities not in the platform.
- Resolve focus reset after opening create missing entities dialog
- Add form validation error if missing entities haven't been created
- add aria-live="polite" attribute to the created relationship section because it dynamically populates as users type
- Update layout of created relationship section to use a table/grid structure to allow labeling of inputs based on the column and allow easier keyboard navigation

## Additional information

Rough mockup:

Image

Yes

scarletmerlin123:
### Proposed changes

* Revised the layout to be vertical and be functional at high zoom levels
* Revised labeling and added helper text to clarify the form
* Added an alert for when users need to take the additional step of creating entities
* Resolve focus issues
* Resolved contrast issues with match status chips

### Related issues

* closes #17889 

### How to test this PR

* Confirm functionality of the modal
* Test using keyboard navigation and a screen reader

### Checklist

### Further comments

image


opencti-platform/opencti-front/src/private/components/common/bulk/dialog/BulkRelationDialog.tsx
```diff
@@ -520,37 +536,85 @@ const BulkRelationDialog: FunctionComponent<BulkRelationDialogProps> = ({
...
+              flex: 1,
+              p: 1,
+              minHeight: '100px',
+              overflowY: 'scroll',
+              background: theme.palette.background.accent,
+              border: `1px solid ${theme.palette.border.main}`,
+            }}
+            >
+              {bulkEntityList.length > 0 ? (
+                <List disablePadding sx={{ height: '100%' }} ref={listRef}>
```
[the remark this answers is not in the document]
scarletmerlin123: In testing adding aria-live="polite" was still intrusive. Given the forms workflow it isn't necessary to provide feedback because the generated list of items is a 1:1 relationship with the input and will be revealed as they move to the next step in the form.

opencti-platform/opencti-front/src/private/components/common/bulk/BulkSelectRawLineData.tsx
```diff
@@ -109,13 +105,22 @@ const BulkSelectRawLineData: FunctionComponent<BulkSelectRawLineDataProps> = ({
...
-      <Box sx={{ minWidth: `${entityNameHeaderWidth}px` }}>
+      <Box
+        tabIndex={0}
+        sx={{
+          '&:focus-visible': {
+            boxShadow: `0 0 0 2px ${theme.palette.text.primary}`,
+          },
+        }}
+      >
+        <span style={{ fontSize: '0.9rem', fontWeight: 400 }} id={`representation-label-${entityIndex}`}>{t_i18n('Representation')}</span>
```
mcartaud: Is it possible to use values from the theme here (font size and weight) (same question for change line 145)

opencti-platform/opencti-front/src/private/components/common/bulk/dialog/BulkRelationDialog.tsx
```diff
@@ -478,40 +459,60 @@ const BulkRelationDialog: FunctionComponent<BulkRelationDialogProps> = ({
...
-              <Box id="entityCard" sx={{ display: 'flex', justifyContent: 'center', padding: '0 10px', flexDirection: 'column' }}>
-                <EntityRelationshipCard
-                  entityName={stixDomainObjectName}
-                  entityType={stixDomainObjectType}
+
+        <DialogContent id="container" sx={{ height: '80vh' }}>
+          <Box component="form" sx={{ display: 'flex', height: '100%', width: '100%', minHeight: 350, flexDirection: 'column', gap: 2 }}>
+            <Box>
+              <FormControl sx={{ minWidth: 120 }}>
+                <TextField
```
mcartaud: Is it possible to use something else than a textField here? Perhaps a title or something like that.
Also is it possible to use the whole width of the dialog for it?

opencti-platform/opencti-front/src/private/components/common/bulk/dialog/BulkRelationDialog.tsx
```diff
@@ -478,40 +459,60 @@ const BulkRelationDialog: FunctionComponent<BulkRelationDialogProps> = ({
...
-              <Box id="relationArrow" sx={{ display: 'flex', justifyContent: 'center', padding: '0 20px', flexDirection: 'column', minWidth: '200px' }}>
-                <Select disabled={isSubmitting} onChange={handleChangeSelectedRelationType} value={selectedRelationType}>
+                <FormHelperText id="bulk-relation-create-from-helper">{t_i18n('The entity that will be used to create a relationship.')}</FormHelperText>
+              </FormControl>
+            </Box>
+            {/* Relationship Type */}
+            <Box>
+              <FormControl sx={{ minWidth: 120 }}>
+                <InputLabel id="bulk-relationship-create-relation-type-label">{t_i18n('Relationship type')}</InputLabel>
+                <Select
```
mcartaud, line 480: idem for the width of the select

opencti-platform/opencti-front/src/private/components/common/bulk/dialog/BulkRelationDialog.tsx
```diff
@@ -520,37 +521,85 @@ const BulkRelationDialog: FunctionComponent<BulkRelationDialogProps> = ({
...
+                          border: `1px solid ${theme.palette.primary.dark}`,
+                          backgroundColor: theme.palette.background.paper,
+                          borderRadius: 0.5,
+                          mt: 0.5,
+                          mb: 0.5,
+                          ':focus': {
+                            border: `1px solid ${theme.palette.primary.main}`,
+                          },
+                        }}
+                        key={uuid()}
```
mcartaud: This feels incorrect. In my understanding, it will create a new uuid on each render so the key will always be different and React won't be able to do its performance optimizations.
We should use a stable property, perhaps the index or an id if one is available?

opencti-platform/opencti-front/src/private/components/common/bulk/dialog/BulkRelationDialog.tsx
```diff
@@ -183,6 +170,9 @@ const BulkRelationDialog: FunctionComponent<BulkRelationDialogProps> = ({
   const [missingEntity, setMissingEntity] = useState<missingEntityType>();
   const [isFirstLoadDone, setIsFirstLoadDone] = useState<boolean>(false);
 
+  const listRef = useRef<HTMLUListElement | null>(null);
```
mcartaud: what is the purpose of this listRef? It didn't seem to be used

opencti-platform/opencti-front/lang/front/fr.json
```diff
@@ -5025,6 +5030,7 @@
   "Type affinity": "Type d'affinité",
   "Type and press Enter to add items": "Tapez et appuyez sur Entrée pour ajouter des articles",
   "Type or copy paste data in this area.": "Tapez ou copiez-collez des données dans cette zone.",
+  "Type or copy paste data. Each entity should be on a new line.": "Tapez ou copiez les données collées. Chaque entité doit être sur une nouvelle ligne.",
```
mcartaud: Écrivez ou copiez les données. Chaque entité doit être sur une nouvelle ligne.

opencti-platform/opencti-front/lang/front/fr.json
```diff
@@ -4686,6 +4689,7 @@
   "The design {name} is currently set as default. Do you want to replace it?": "Le thème {name} est actuellement défini comme thème par défaut. Souhaitez-vous le remplacer ?",
   "The disconnection process is still in progress. Closing this dialog will terminate the disconnection. Are you sure you want to close?": "La déconnexion est toujours en cours. La fermeture de cette boîte de dialogue mettra fin à la déconnexion. Voulez-vous vraiment fermer cette boîte de dialogue ?",
   "the draft processes tab": "l'onglet des processus du brouillon",
+  "The entity that will be used to create a relationship.": "L'entité qui sera utilisée pour créer une relation.",
```
mcartaud: L'entité qui sera utilisée pour créer les relations.

opencti-platform/opencti-front/lang/front/fr.json
```diff
@@ -1021,6 +1022,7 @@
   "Create Authentication": "Créer une authentification",
   "Create custom view": "Créer une vue personnalisée",
   "Create dashboard": "Créer un tableau de bord",
+  "Create entities before submitting.": "Créez des entités avant de les envoyer.",
```
mcartaud: Créez les entités avant de valider.

opencti-platform/opencti-front/lang/front/fr.json
```diff
@@ -3854,6 +3856,7 @@
   "Relations to threats": "Relations vers des menaces",
   "Relationship": "Relation",
   "Relationship log details": "Détails du journal des relations",
+  "Relationship match status": "Statut de la correspondance de la relation",
```
mcartaud: Status de la relation

opencti-platform/opencti-front/src/private/components/common/bulk/dialog/BulkRelationDialog.tsx
```diff
@@ -478,40 +459,60 @@ const BulkRelationDialog: FunctionComponent<BulkRelationDialogProps> = ({
...
+                      readOnly: true,
+                    },
+                  }}
                 />
-              </Box>
-              <Box id="relationArrow" sx={{ display: 'flex', justifyContent: 'center', padding: '0 20px', flexDirection: 'column', minWidth: '200px' }}>
-                <Select disabled={isSubmitting} onChange={handleChangeSelectedRelationType} value={selectedRelationType}>
+                <FormHelperText id="bulk-relation-create-from-helper">{t_i18n('The entity that will be used to create a relationship.')}</FormHelperText>
+              </FormControl>
+            </Box>
```
mcartaud, line 475: Is it possible to add a bit of vertical padding between the entity label, relationship type and the text field? It'd help understanding the different sections

opencti-platform/opencti-front/src/private/components/common/bulk/dialog/BulkRelationDialog.tsx
scarletmerlin123: You're right, the id should be stable. In this case there isn't a real id yet and the index isn't stable. I updated the generated entries to have an id to use as the key. Since it's on the item it should only update when the item changes .

opencti-platform/opencti-front/lang/front/fr.json
scarletmerlin123: This text was replaced with a different line of text since it no longer has the helper text

opencti-platform/opencti-front/src/private/components/common/bulk/dialog/BulkRelationDialog.tsx
scarletmerlin123: Increased the gap from 2 to 4 (16px -> 32px)

opencti-platform/opencti-front/src/private/components/common/bulk/dialog/BulkRelationDialog.tsx
```diff
@@ -291,6 +281,7 @@ const BulkRelationDialog: FunctionComponent<BulkRelationDialogProps> = ({
             setIsFirstLoadDone(true);
           }
           return [...acc, {
+            id: uuid(),
```
mcartaud, line 284: If the element is already existing, this will override its id. It is probably not an issue, but we better use the correct id to avoid creating a new element by mistake

opencti-platform/opencti-front/src/private/components/common/bulk/dialog/BulkRelationDialog.tsx
```diff
@@ -301,6 +292,7 @@ const BulkRelationDialog: FunctionComponent<BulkRelationDialogProps> = ({
           }];
         }
         return [...acc, {
+          id: uuid(),
```
mcartaud, line 295: idem

opencti-platform/opencti-front/src/private/components/common/bulk/dialog/BulkRelationDialog.tsx
```diff
@@ -478,40 +461,46 @@ const BulkRelationDialog: FunctionComponent<BulkRelationDialogProps> = ({
...
-                  entityType={stixDomainObjectType}
-                />
-              </Box>
-              <Box id="relationArrow" sx={{ display: 'flex', justifyContent: 'center', padding: '0 20px', flexDirection: 'column', minWidth: '200px' }}>
-                <Select disabled={isSubmitting} onChange={handleChangeSelectedRelationType} value={selectedRelationType}>
+
+        <DialogContent id="container" sx={{ height: '80vh' }}>
+          <Box component="form" sx={{ display: 'flex', height: '100%', width: '100%', minHeight: 350, flexDirection: 'column', gap: 4 }}>
+            <Box>
+              <Typography variant="body1">{t_i18n('Relationships will be created from the')} {t_i18n(`entity_${stixDomainObjectType}`)}: </Typography>
```
mcartaud, line 473: Small design remark : the name of the entity is smaller than the description above, can you do the opposite to make the important part stand out?
To match the current design system effort, we should use
Relationships will be created from the Security Coverages: -> 12px, secondary text color (#AFB0B6)
"entity label"-> 14px, primary text color (#F2F2F3)

If the properties exists in the design system but don't have the correct value yet, please use them, we have an ongoing design system effort that will be merged soon, otherwise, i'll handle the change when needed. Thanks

opencti-platform/opencti-front/src/private/components/common/bulk/dialog/BulkRelationDialog.tsx
scarletmerlin123: Updated property to be lineId so it has no overlap with the real data property.

opencti-platform/opencti-front/src/private/components/common/bulk/dialog/BulkRelationDialog.tsx
scarletmerlin123: Updated, though #AFB0B6 appears to be text.light. Also updated the text area, labels, and helper text to use the new design system.

mcartaud (approved):
Thanks @scarletmerlin123

PR #17927, opened 2026-08-27
feat(workflow): Make workflow publish create/resolve Status (#17898)

### LINKED ISSUE #17898: feat(workflow): Make workflow publish create/resolve Status
## Proposed solution 
- topological + cycle-tolerant per-state ordering (order field)
- publish-time full-status-mapping
- deletion guard

delemaf:
## Proposed changes

Implements topological + cycle-tolerant per-state ordering for workflow Status entities, and makes workflow publish create/resolve the associated `Status` records with a full status mapping.

- Adds `workflow-ordering.ts`: computes an exact `order` for every state reachable from `initialState`, even inside cycles — a single DFS marks the "back edges" that close each cycle, and a topological longest-path pass (Kahn's algorithm + DP) over the remaining DAG then assigns every state a deterministic order. There is no more manual `order` requirement or ambiguous/rejected case for cyclic states.
- Extends `workflow-domain.ts` to compute and persist the full status mapping at publish time, and to guard deletion of statuses/workflows referenced by the invariant.
- The `order` field on a state is now legacy-only, kept for backward-compatible round-tripping — it's never required and never used to resolve ambiguity, since the order is always derived automatically from the transition graph.
- Adds a `workflowStatusCleanupManager` to hard-delete orphaned `Status` records once their grace period (set by `reconcileOrphanedStatuses` on republish) has elapsed, re-verifying orphan status right before deletion.
- Adds unit test coverage for ordering, domain, validation, resolvers, status, and the cleanup manager.

### How to test
1. **Publish creates/resolves full Status mapping**
   - Draft a workflow with states `A -> B` (a transition to `B` that isn't listed in `states`), publish it.
   - Confirm a `Status` is created for both `A` and `B` (not just states explicitly listed), and that `B` shows up when assigning an entity's workflow status.

2. **Ordering: cycles are resolved automatically, no manual order needed**
   - Draft states `A -> B -> A` (a simple cycle) plus an unrelated `C -> D` chain, with no manual `order` set on any state.
   - Publish: confirm it succeeds and every state — including the cyclic `A`/`B` — gets a computed order automatically (no "contains a cycle" or missing-order validation error).
   - Confirm `C -> D` still gets its own correct topological order, independent of the unrelated cycle.
   - Separately, confirm a state that's genuinely **unreachable** from `initialState` (not part of any cycle, just never connected) is still rejected with a clear unreachable-state validation error, distinct from cycle handling.

3. **Removing a status still in use is blocked**
   - Publish `A -> B`, assign an entity to `B` (an ending state).
   - Save a draft removing `B`, attempt to publish it → should fail with `STATE_IN_USE`, even though `B` was only reachable via a transition (not listed in `states`).

4. **Status deletion guard is scoped correctly**
   - Create two entity types (e.g. `Incident` and `Report`) sharing the same `StatusTemplate`.
   - Confirm deleting the `Report` status succeeds when only `Incident`'s workflow actually references that template.

### Related

Closes #17898


opencti-platform/opencti-graphql/src/manager/workflowStatusCleanupManager.ts
```diff
@@ -0,0 +1,65 @@
...
+const WORKFLOW_STATUS_CLEANUP_MANAGER_KEY = conf.get('workflow_status_cleanup_manager:lock_key') || 'workflow_status_cleanup_manager_lock';
+const SCHEDULE_TIME = conf.get('workflow_status_cleanup_manager:interval') || 86400000; // 1 day
+
+/**
+ * Hard-deletes `Status` records whose grace period (set by `reconcileOrphanedStatuses` on
+ * republish, see workflow-domain.ts) has elapsed. Each candidate is re-verified as still orphaned
+ * right before deletion, since state can change during the grace window. Idempotent by
+ * construction: re-running the check-then-delete on an already-deleted record is a no-op.
+ */
+export const workflowStatusCleanupHandler = async () => {
```
OctaveLaventure, line 23: Currently this doesn't check for statuses used inside of drafts, we should discuss what to do about that

opencti-platform/opencti-graphql/src/modules/workflow/domain/workflow-ordering.ts
```diff
@@ -0,0 +1,156 @@
...
+  visit(initialState);
+  return onCycle;
+};
+
+/**
+ * Computes a display/validation order for every state reachable from `initialState`: the length
+ * of the longest simple path from `initialState` to it.
+ *
+ * A state's value is `null` only if it lies on a cycle reachable from `initialState`, or if the
+ * DFS never reached it before the step cap was hit. Callers must fall back to a manually supplied
```
OctaveLaventure: We should remove the manual order and references to it, and always rely on topological ordering

opencti-platform/opencti-graphql/src/modules/workflow/domain/workflow-domain.ts
```diff
@@ -592,11 +758,21 @@ export const publishWorkflowDefinition = async (
     throw FunctionalError('No draft version to publish', { entityType });
   }
 
-  // Check for validation errors
-  if (draftVersion.validation_errors && draftVersion.validation_errors.length > 0) {
+  // Re-validate at publish time rather than trusting the draft's stored `validation_errors`,
+  // which were computed at save time and can be stale (e.g. new validation rules added since, or
+  // DB state — like status templates — changed after the draft was last saved). Publishing an
+  // invalid definition (e.g. missing statusId / missing manual order) must be blocked here.
+  const freshValidationErrors = await validateWorkflowDefinitionData(
```
OctaveLaventure, line 768: We should discuss how to handle already published workflows that won't pass this validation anymore

opencti-platform/opencti-graphql/src/manager/workflowStatusCleanupManager.ts
```diff
@@ -0,0 +1,65 @@
...
+    },
+  });
+
+  let errorCount = 0;
+  for (let i = 0; i < candidates.length; i += 1) {
+    const status = candidates[i];
+    try {
+      const stillOrphaned = await isStatusOrphaned(context, WORKFLOW_MANAGER_USER, status);
+      if (stillOrphaned) {
+        await internalDeleteElementById(context, WORKFLOW_MANAGER_USER, status.id, ENTITY_TYPE_STATUS);
```
OctaveLaventure: **Invalidate the Status cache after cleanup deletion**

`internalDeleteElementById` does not publish the internal-object deletion notification, and Status deletions do not produce the STIX event that could otherwise propagate this change. If the Status cache was populated before this run, `findById` and `findByType` can continue returning the deleted record until another invalidation or restart.

Please capture the deleted element and publish `BUS_TOPICS[ABSTRACT_INTERNAL_OBJECT].DELETE_TOPIC`, matching `statusDelete` and `statusTemplateDelete`.

opencti-platform/opencti-graphql/src/domain/status.ts
```diff
@@ -203,6 +205,13 @@ export const statusTemplateEditField = async (context: AuthContext, user: AuthUs
 };
 export const statusDelete = async (context: AuthContext, user: AuthUser, subTypeId: string, statusId: string) => {
   validateSetting(subTypeId, 'workflow_configuration');
+  const status = await storeLoadById<BasicWorkflowStatus>(context, user, statusId, ENTITY_TYPE_STATUS);
+  if (status) {
+    const isUsedInWorkflow = await isStatusTemplateUsedInWorkflows(context, user, status.template_id);
```
OctaveLaventure: **Scope the deletion guard to this Status's entity type and scope**

`isStatusTemplateUsedInWorkflows` searches all workflows using only the template ID, but a Status is specific to a template, entity type, and scope. For example, an Incident workflow using template T now prevents deleting an unused Report status backed by T, even when Report has no workflow. An unrelated request-access status can be blocked in the same way.

Please check the workflow mapping applicable to this specific Status. The global template-reference check is appropriate for deleting the shared StatusTemplate, not every Status backed by it.

opencti-platform/opencti-graphql/src/modules/workflow/domain/workflow-domain.ts
```diff
@@ -557,6 +561,186 @@ export const deleteWorkflowDefinition = async (
...
+ * Only the Global scope is reconciled here — existing `Status` records in other scopes are
+ * left untouched.
+ */
+export const ensureFullStatusMapping = async (
+  context: AuthContext,
+  user: AuthUser,
+  entityType: string,
+  definitionData: WorkflowDefinitionData,
+): Promise<void> => {
+  const states = definitionData.states ?? [];
```
OctaveLaventure: **Include implicit states in the full status mapping**

Validation accepts transition endpoints backed by existing StatusTemplates even when they are absent from `definitionData.states`, and the engine registers those states. With templates A/B, `initialState: A`, `states: [{ statusId: A }]`, and transition A -> B, publishing succeeds but this loop creates only A's Status. With an empty `states` array, the early return skips the mapping altogether.

Please reconcile every canonical state ID, including initial/transition states, or explicitly reject incomplete declarations before publishing. Otherwise publish-time full-status-mapping is not guaranteed.

opencti-platform/opencti-graphql/src/modules/workflow/domain/workflow-ordering.ts
```diff
@@ -0,0 +1,156 @@
...
+    color.set(state, 'gray');
+    stack.push(state);
+    (adjacency.get(state) ?? new Set<string>()).forEach((neighbor) => {
+      const neighborColor = color.get(neighbor);
+      if (neighborColor === 'gray') {
+        const ancestorIndex = stack.indexOf(neighbor);
+        for (let i = ancestorIndex; i < stack.length; i += 1) {
+          onCycle.add(stack[i]);
+        }
+      } else if (neighborColor === undefined) {
```
OctaveLaventure: **Detect all members of overlapping cycles**

For transitions A -> B, B -> A, A -> C, C -> B in that order, this traversal returns `{ A: null, B: null, C: 1 }`. C also belongs to a cycle (A -> C -> B -> A), but its edge to already-black B is ignored here. Consequently, publishing overrides an explicit manual order such as `C.order = 7` with the computed value `1`.

Please identify cycle membership using strongly connected components rather than only gray-node back-edge stack segments, and cover this overlapping-cycle case.

opencti-platform/opencti-graphql/src/modules/workflow/domain/workflow-ordering.ts
```diff
@@ -0,0 +1,156 @@
...
+  };
+  dfs(initialState, 0, new Set([initialState]));
+
+  const cyclicStates = statesOnCycles(initialState, adjacency);
+  const result = new Map<string, number | null>();
+  // A state is null only if it's cycle-entangled, or if the DFS never computed an order for it
+  // (unreachable via any acyclic-terminating branch, or not yet visited when the step cap hit) —
+  // states that did get a real value before the cap was hit elsewhere keep that value.
+  reachable.forEach((state) => {
+    result.set(state, (cyclicStates.has(state) || !longestOrder.has(state)) ? null : (longestOrder.get(state) as number));
```
OctaveLaventure: **Do not treat partially explored orders as final after the cap**

Visiting a state before the step cap does not mean its longest path has been computed. A complete 15-state DAG (s0 through s14, with edges from each lower index to every higher index and destinations enumerated descending) hits the default 5,000-step cap and returns `s2 = 1, s3 = 1` despite edge s2 -> s3. Without the cap, their orders are 2 and 3. These incomplete numeric values override manual fallbacks and are persisted during publish.

Please compute DAG orders topologically instead of enumerating paths, or preserve capped values only when they are proven final.

opencti-platform/opencti-graphql/src/modules/workflow/domain/workflow-domain.ts
```diff
@@ -1130,6 +1328,40 @@ export const cleanupEntityWorkflow = async (
...
+    const content = version.content;
+    let parsed;
+    try {
+      parsed = typeof content === 'string' ? JSON.parse(content) : content;
+    } catch (_error) {
+      // Malformed content is not this function's concern; skip rather than false-positive.
+      continue;
+    }
+    const states: Array<{ statusId?: string }> = parsed?.states ?? [];
+    if (states.some((state) => state.statusId === statusTemplateId)) {
```
OctaveLaventure: **A state used in a transition can still be deleted**

A workflow can use B without listing it in `states`, for example:

```js
states: [{ statusId: A }],
initialState: A,
transitions: [{ from: A, to: B, event: 'next' }]
```

This is valid when both templates exist. The updated publish code correctly creates a Status for B, but this deletion check only looks at `states`. It therefore treats B as unused and allows deleting its Status or template, leaving the published workflow with a broken reference.

The same mismatch appears in two other places:
- **Cleanup:** removing B from `states` while keeping the transition can schedule its Status for deletion, even though the workflow still uses it.
- **Validation:** that same edit can incorrectly produce `STATUS_IN_USE` when an entity is assigned to B, even though B has not actually been removed from the workflow.

**Suggested fix:** use `extractCanonicalStateIds` wherever we decide which states a workflow uses. It includes `initialState` and transition endpoints as well as `states`, and `ensureFullStatusMapping` already uses it. Apply the same definition of "used" to validation, deletion guards, and cleanup.

opencti-platform/opencti-graphql/src/modules/workflow/domain/workflow-domain.ts
```diff
@@ -592,11 +761,21 @@ export const publishWorkflowDefinition = async (
...
   }
 
-  // Check for validation errors
-  if (draftVersion.validation_errors && draftVersion.validation_errors.length > 0) {
+  // Re-validate at publish time rather than trusting the draft's stored `validation_errors`,
+  // which were computed at save time and can be stale (e.g. new validation rules added since, or
+  // DB state — like status templates — changed after the draft was last saved). Publishing an
+  // invalid definition (e.g. missing statusId, unreachable state) must be blocked here.
+  const freshValidationErrors = await validateWorkflowDefinitionData(
+    executionContext,
```
OctaveLaventure, line 769: **Publishing can bypass the "status is still assigned" error**

Example:
1. Publish A -> B, where B is an end state, and assign an Incident to B.
2. Save a draft that removes B. Validation correctly returns `STATUS_IN_USE`.
3. Publish that draft. Publication succeeds even though the Incident is still assigned to B.

**Why:** during publication, `validateWorkflowDefinitionData` uses the saved draft as the "old" definition. It compares the draft with itself, sees no removed states, and no longer returns the error from step 2. The separate WorkflowInstance check does not stop this example because it deliberately skips end states.

**Suggested fix:** when publishing, compare the draft against the currently **published** definition and check whether entities are still assigned to any removed statuses. Keep the fresh validation; give its removal check the correct comparison baseline.

opencti-platform/opencti-graphql/src/manager/workflowStatusCleanupManager.ts
```diff
@@ -0,0 +1,68 @@
...
+    },
+  });
+
+  let errorCount = 0;
+  for (let i = 0; i < candidates.length; i += 1) {
+    const status = candidates[i];
+    try {
+      const stillOrphaned = await isStatusOrphaned(context, WORKFLOW_MANAGER_USER, status);
+      if (stillOrphaned) {
+        const { element: deleted } = await internalDeleteElementById(context, WORKFLOW_MANAGER_USER, status.id, ENTITY_TYPE_STATUS);
```
OctaveLaventure: **Cleanup can delete a Status while publishing brings it back**

For a Status B whose deletion deadline has passed, these operations can overlap:
1. Cleanup checks B and decides it is no longer used.
2. An admin republishes a workflow that uses B again. Publishing clears B's deletion mark and confirms that its Status exists.
3. Cleanup deletes B based on its earlier check.
4. Publishing finishes successfully, but the published workflow now references a missing Status.

Clearing `to_be_deleted_at` does not cancel a deletion already in progress. The cleanup scheduler lock prevents two cleanup runs from overlapping, but it does not block publishing. The deletion's own Status lock also does not repeat the earlier usage check.

**Suggested fix:** make publishing and cleanup use a shared lock for the same workflow/entity type. Hold it through publishing's reconciliation, Status mapping, and publication update; cleanup should reload the Status and recheck its deadline and usage under that same lock before deleting it. This makes the intended "restore wins" behavior reliable.

opencti-platform/opencti-graphql/src/modules/workflow/domain/workflow-domain.ts
```diff
@@ -557,6 +563,177 @@ export const deleteWorkflowDefinition = async (
...
+  // states (referenced only via initialState/transitions) have no manual override and default
+  // to 0 when the topological order is ambiguous.
+  const manualOrderByStatusId = new Map(
+    (definitionData.states ?? [])
+      .filter((state): state is WorkflowSerializedState & { statusId: string } => !!state.statusId)
+      .map((state) => [state.statusId, state.order]),
+  );
+
+  for (const statusId of canonicalStateIds) {
+    const order = computedOrder.get(statusId) ?? manualOrderByStatusId.get(statusId) ?? 0;
```
OctaveLaventure: **Generate orders for cyclic states too, without manual input**

The intended behavior is that **all orders are generated by the backend**, including states in cycles. A missing manual `order` should not prevent publication.

Currently, `computeStateOrder` returns `null` for cycle members, so this line uses a supplied manual order or defaults to `0`. In the browser scenario with A <-> B and a reachable C -> D branch, publishing preserved supplied values A=10 and B=20. Removing B's manual value then changed its persisted order to 0, although the graph had not changed. Those values are therefore coming from manual input/defaults, not automatic ordering of the cycle.

**Suggested fix:** include cyclic states in the automatic ordering policy and remove the dependence on `manualOrderByStatusId`. Publishing the same graph should not depend on whether manual orders were supplied. Update the PR description and scenario 2 to expect successful publication with generated orders, rather than rejection when a manual order is absent.

This corrects my earlier interpretation of the test instructions: the issue is incomplete automatic ordering, not missing validation for manual orders.

OctaveLaventure (approved):

PR #17949, opened 2026-08-28
fix(segregation): ids of entities that do not meet the filter of the stream are no longer shared in the object_ref list (#17576)

### LINKED ISSUE #17576: fix(segregation): Live stream share objects' stix IDs contained in container even though they do not meet the filter of the stream
## Description

Follow up of: https://github.com/OpenCTI-Platform/opencti/issues/16814

A live stream bound to a user (U1) whose maximum marking is TLP:AMBER+STRICT emits TLP:RED objects, object values, and reference translations that U1 is not authorized to access.  Full TLP:RED objects are emitted via dependency-resolution create events.
## Environment

Latest

## Reproducible steps

1. On platform A, Create a user U1 in a single organization, with a maximum marking capability of TLP:AMBER+STRICT (no TLP:RED access). Confirm U1 cannot see TLP:RED objects in the UI.
2. Create a live stream configured as a public stream, with the "Share data corresponding to permissions associated with this user" field set to U1.
3. Set the stream filter to entity type = report.
4. On platform B, connect to the stream and launch ingestion.
5. As a separate admin user (with TLP:RED access), add TLP:RED entity to a Report that U1 can see, within platform A. **Ensure that the same entity already exists within platform B, with TLP:RED**
6. With a user of PlatformB, go into the created report: you will see the entity TLP:RED in it, even though the stream is supposed to only share TLP:AMBER.+strict

curl.exe -N -H "Authorization: Bearer flgrn_octi_tkn_xxxx" "https://[opencti uri]/stream/[stream id]?no-dependencies=true"

## Expected output

You should not see the entity in platform B. 

## Actual output

You see the entity TLP:RED that you have added in platform A, in platform B. 

## Additional information

This seems to be due to the way we fetch the entities: we perform an in regards of, to load refs of the containers, without filtering on what the user in charge of sharing the stream is supposed to see. Therefore, the issue might be wider than only the marking, can potentially apply to any restriction of any refs (after a discussion with @JeremyCloarec )

MTorbay-Filigran:
### Proposed changes

Before sending events in stream with createSseMiddleware, sendEventWithFilteredObjectRefs filters object_refs to keep the references already published to that consumer, which are exactly the ones resolveAndPublishDependencies was able to resolve with the rights of the stream's user, and add entities missing from the cache.

* Cached entities :
Before sending an event to the stream, object_refs are filtered to keep the references present in the connection cache. That cache is populated just above by resolveAndPublishDependencies, which resolves references with the rights of the stream user, so anything the consumer cannot access was never resolved, never published as a dependency, and is therefore dropped here. This prevents leaking the ids of restricted objects and avoids sending references to objects the consumer never received.

* Entites evicted from cache :
As the cache is limited (5000 entities by default), sendEventWithFilteredObjectRefs fetches the entities accessible for the user that are not in the cache and add them in the object_refs filtered.

This prevents a container from exposing the STIX ids of objects the consumer is not allowed to see, such as an object restricted by a marking above the stream public user's allowed markings.

### Related issues

* closes #17576 

### How to test this PR

- Create a user with restriction TLP:AMBER
- Create a public stream restricted with this user's rights :
- Capture d’écran 2026-08-28 à 10 08 34
- Capture d’écran 2026-08-28 à 10 08 53
- Start the stream and open it
- Create a report, add several entities : one with no marking, one with marking tlp:amber, one with marking tlp:red
- In the stream, you should never see the id of the entity marked with tlp:red in object_refs.

### Checklist


opencti-platform/opencti-graphql/src/graphql/sseMiddleware.js
```diff
@@ -282,6 +282,33 @@ export const resolveMissingReferences = async (context, user, missingRefs, cache
...
+          user,
+          objectsRefsNotFoundInCache,
+          { baseFields: ['internal_id', 'standard_id'], indices: READ_STIX_INDICES, toMap: true, mapWithAllIds: true })
+        )
+      : {};
+    objectsRefs = objectsRefs.filter((ref) => cache.has(ref) || objectRefsAccessibleNotFoundInCache[ref] !== undefined);
+    eventToSend.eventData.data.object_refs = objectsRefs;
+  }
+  return client.sendEvent(eventToSend.eventId, eventToSend.eventType, eventToSend.eventData);
+};
```
fellowseb, line 310: I see a problem here: we're using the cache as a source of "access rights" truth. It shouldn't though because the user can see their markings/orgs change while the stream is running. We should not rely on the cache here IMO.

```
export const sendEventWithFilteredObjectRefs = async (
  context,
  user,
  client,
  eventToSend,
) => {
  let objectsRefs = eventToSend.eventData.data.object_refs;
  if (objectsRefs?.length > 0) {
    const uniqueObjectRefs = [...new Set(objectsRefs)];
    const objectRefsAccessible = uniqueObjectRefs.length > 0
      ? (await elFindByIds(
          context,
          user,
          uniqueObjectRefs,
          { baseFields: ['internal_id', 'standard_id'], indices: READ_STIX_INDICES, toMap: true, mapWithAllIds: true })
        )
      : {};
    objectsRefs = uniqueObjectRefs.filter((ref) => objectRefsAccessible[ref] !== undefined);
    eventToSend.eventData.data.object_refs = objectsRefs;
  }
  return client.sendEvent(eventToSend.eventId, eventToSend.eventType, eventToSend.eventData);

};
```

SouadHadjiat: well in that case we might already publish information that the user doesn't have access to anymore. The cache is just a "cache" of the dependencies that the user has access to at the time of the request, some of them will be published along with the stream event. This cache is per request, it doesn't last more.

fellowseb: Yes but a request in this context can run virtually forever, right ?
That's the intent of
```
      Connection: 'keep-alive',
```
and the use of heartbeats, if I'm not mistaken.
My message was related to that case: someone opens a long-running never-ending connection, but in the midst of the event sending the markings/organizations definitions change.

fellowseb: Maybe we could imagine using the cache but then subscribing to changes that could affect access rights to invalidate the cache ? That seems complex.
Also the current `user` passed in the handler is never updated too. I don't think we should use it in the stream processor callback code itself but always read user from the applicative cache instead.

But I do see the potential performance issue here. We can discuss the perf/security tradeoffs offline.

SouadHadjiat: I understand your concerns. This would be a bigger change. We thought with Marie of using the cache like it's been already used in this function. The main fix is to filter the object refs with objects that are accessible to the user. The user rights changing in the meantime is another issue (to be reproduced and assessed)

MTorbay-Filigran: Conversation continued live. For conclusion we decided to validate the PR as it is and open another issue for the user rights subject.

fellowseb (approved):

SouadHadjiat (approved):

PR #17962, opened 2026-08-28
feat(filters): Enable dynamic regarding of nested filter update in widgets (#17709)

### LINKED ISSUE #17709: fix(Dashboard): Entities list widget filters become non-editable after reload when using nested dynamic "In regards of" filters
## Description

Complex dynamic filters in an **Entities list dashboard widget** become non-editable after the dashboard page is reloaded.

The issue was initially reported for version `7.260722.0` and has now been reproduced internally on the demo instance running `7.260811.0`.

The filter chips are rendered correctly, but clicking them does nothing. The user can only delete the whole filter expression and rebuild it from scratch.

## Environment

- Reproduced on customer production: `7.260722.0`
- Reproduced internally on demo instance: `7.260811.0`
- Customer reports it does **not** happen on their test environment: `7.260609.0

## Reproducible steps

-> if you wish to have the JSON file widget, I'll share it with you privately.

1. Import or create an **Entities list** dashboard widget.
2. Configure a complex filter using:
   - `entity_type`
   - one or more dynamic **In regards of** filters
   - nested `regardingOf` conditions
3. Save the widget / dashboard.
4. Before reloading, click the filter chips.
   - They are editable.
5. Reload the dashboard page.
6. Click the same filter chips again.

## Expected output

After reload, all filter chips should remain editable:
- Clicking an existing `entity_type` filter should reopen the filter editor.
- Clicking an existing dynamic `In regards of` filter should reopen the dynamic filter editor.
- Persisted filters should be fully rehydrated into an editable UI state.

## Actual output

After reload, the filters are displayed but cannot be edited:
- Clicking `entity_type` or `In regards of` filter chips does nothing.
- No filter editor / popover opens.
- The only workaround is to delete the filter expression and recreate it.

## Additional information

The customer observed that when importing a widget from another environment where editing works:
- Filters are editable immediately after import.
- After reloading the page, the same filters become non-editable.

The customer also confirmed they have the same permissions on both environments.

Archidoit:
### Proposed changes
We should be able to update the nested filter of a 'in regards of (dynamic)' filter inside a widget.

### How to test
- Create a widget with a 'in regards of (dynamic)' filter.
- Update the widget
- Try to update the subfilter of the 'in regards of (dynamic)' filter. The update should be possible.
image

### What was the issue
The issue was that, in a widget with a 'in regards of (dynamic)' filter, the subfilter was not updatable. Example: when you clicked on the 'entity type' key of the subfilter in the screenshot below, nothing opens. Now a pop-up opens.
image

### Related issues
fixes #17709

### DIFF
--- a/opencti-platform/opencti-front/src/utils/filters/filtersUtils.tsx
+++ b/opencti-platform/opencti-front/src/utils/filters/filtersUtils.tsx
@@ -583,12 +583,32 @@ export const normalizeFilterGroupForFrontend = (
 ): FilterGroup => {
   return {
     ...filterGroup,
-    filters: filterGroup?.filters?.map((f) => ({
-      ...f,
-      id: uuid(),
-      key: Array.isArray(f.key) ? f.key[0] : f.key,
-      values: f.values.map((v) => v || 'todo: delete this'),
-    })),
+    filters: filterGroup?.filters?.map((f) => {
+      const key = Array.isArray(f.key) ? f.key[0] : f.key;
+      // build values
+      let values: FilterValue[];
+      if (key === 'dynamicRegardingOf') { // add id in dynamic regarding of subfilter for React rendering purposes
+        values = f.values.map((dynamicRegardingOfValue) => {
+          if (dynamicRegardingOfValue.key === 'dynamic') { // values with 'dynamic' key contains filters
+            return {
+              ...dynamicRegardingOfValue,
+              values: dynamicRegardingOfValue.values.map((filterValue) => normalizeFilterGroupForFrontend(filterValue)),
+            };
+          } else {
+            return dynamicRegardingOfValue;
+          }
+        });
+      } else {
+        values = f.values.map((v) => v || 'todo: delete this');
+      }
+      // return the filter with normalized key and values, and add an id
+      return {
+        ...f,
+        id: uuid(),
+        key,
+        values,
+      };
+    }),
     filterGroups: filterGroup?.filterGroups?.map((fg) => normalizeFilterGroupForFrontend(fg)),
   } as FilterGroup;
 };
@@ -609,7 +629,7 @@ export const serializeFilterGroupForBackend = (
 
 /**
  * Parse a filterGroup as given by the backend (backend format, i.e. with array keys),
- * And turns it into the frontend format (single key).²
+ * And turns it into the frontend format (single key).
  * @param filterGroup
  */
 export const deserializeFilterGroupForFrontend = (
--- a/opencti-platform/opencti-front/src/utils/filters/filtersUtils.tsx
+++ b/opencti-platform/opencti-front/src/utils/filters/filtersUtils.tsx
@@ -592,7 +592,7 @@ export const normalizeFilterGroupForFrontend = (
           if (dynamicRegardingOfValue.key === 'dynamic') { // values with 'dynamic' key contains filters
             return {
               ...dynamicRegardingOfValue,
-              values: dynamicRegardingOfValue.values.map((filterValue) => normalizeFilterGroupForFrontend(filterValue)),
+              values: dynamicRegardingOfValue.values.map((filterValue: GqlFilterGroup) => normalizeFilterGroupForFrontend(filterValue)),
             };
           } else {
             return dynamicRegardingOfValue;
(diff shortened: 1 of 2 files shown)


opencti-platform/opencti-front/src/utils/filters/filtersUtils.tsx
```diff
@@ -583,12 +583,32 @@ export const normalizeFilterGroupForFrontend = (
...
   return {
     ...filterGroup,
-    filters: filterGroup?.filters?.map((f) => ({
-      ...f,
-      id: uuid(),
-      key: Array.isArray(f.key) ? f.key[0] : f.key,
-      values: f.values.map((v) => v || 'todo: delete this'),
-    })),
+    filters: filterGroup?.filters?.map((f) => {
+      const key = Array.isArray(f.key) ? f.key[0] : f.key;
```
mcartaud, line 587: Why this isArray check? The type says it can only be an array

opencti-platform/opencti-front/src/utils/filters/filtersUtils.tsx
```diff
@@ -583,12 +583,32 @@ export const normalizeFilterGroupForFrontend = (
...
+            return dynamicRegardingOfValue;
+          }
+        });
+      } else {
+        values = f.values.map((v) => v || 'todo: delete this');
+      }
+      // return the filter with normalized key and values, and add an id
+      return {
+        ...f,
+        id: uuid(),
```
mcartaud, line 607: I know it was here before and we probably don't want to update it now, but we should be careful with this usage of uuid() because it is often something used as key for React and if it is not stable it could lead to unnecessary re-renders. It the custom dashboard page it is ok, the method is called only once at logging, but it is error prone

opencti-platform/opencti-front/src/utils/filters/filtersUtils.tsx
```diff
@@ -583,12 +583,32 @@ export const normalizeFilterGroupForFrontend = (
...
+    filters: filterGroup?.filters?.map((f) => {
+      const key = Array.isArray(f.key) ? f.key[0] : f.key;
+      // build values
+      let values: FilterValue[];
+      if (key === 'dynamicRegardingOf') { // add id in dynamic regarding of subfilter for React rendering purposes
+        values = f.values.map((dynamicRegardingOfValue) => {
+          if (dynamicRegardingOfValue.key === 'dynamic') { // values with 'dynamic' key contains filters
+            return {
+              ...dynamicRegardingOfValue,
+              values: dynamicRegardingOfValue.values.map((filterValue) => normalizeFilterGroupForFrontend(filterValue)),
```
mcartaud: could you add a type (TS warning)?

opencti-platform/opencti-front/src/utils/filters/filtersUtils.tsx
Archidoit: Indeed, but with the filter key typing difference between front and back, and because the modifications are sometimes complex, it avoids errors in case of bad conversions. It's a guard.

opencti-platform/opencti-front/src/utils/filters/filtersUtils.tsx
Archidoit: I don't see the ts warning :/ can you screenshot it maybe?

opencti-platform/opencti-front/src/utils/filters/filtersUtils.tsx
Archidoit: yes the method is only called once for each widget filters. Noted !

opencti-platform/opencti-front/src/utils/filters/filtersUtils.tsx
mcartaud: Not a big one, but 🤷
It comes from filter.value which is an array of any, but the method `normalizeFilterGroupForFrontend` takes a `GqlFilterGroup`. Probably in the `dynamicRegardingOf` case we always have `GqlFilterGroup` in `dynamicRegardingOfValue.values`, but writing it makes it more obvious (would be nice to have a fully typed definition of Filters but that's not something for now :) 

image

Archidoit: thanks ! it's fixed

mcartaud (approved):

PR #17975, opened 2026-08-28
feat(backend): use the app encryption key to derive the session signature key (#15353)

### LINKED ISSUE #15353: feat(security): use app encryption to generate the session signature key
## Description

With the introduction of app encryption key, we could generate the session signing key instead of defaulting on admin password

xfournet:
#### Session signature key

* The session cookie signature key is now derived from the platform encryption key (`app:encryption_key`), through the `session:signature` derivation path, instead of falling back on the admin password when `app:session_secret` was not set. The derived key has a full 256 bits of entropy, is automatically identical on all the platform nodes, and never has to be configured nor rotated by hand.
* The undocumented `app:session_secret` configuration and the `app:admin:password` fallback are removed. `app:session_secret` was referenced nowhere else in the repository, not even in `config/default.json`.
* `deriveSecret(path, version, length = 32)` is added to the platform crypto key factory, for the third party libraries that require raw secret material — here `express-session`, whose `secret` option is unavoidably a raw string. It is the only derivation that hands out key material, hence the comment restricting its use. Rotation stays possible through the derivation path version, like every other derived key.
* The session middleware is created lazily (`getSessionMiddleware()`), since the key derivation is asynchronous. The session store is therefore built on first use instead of at import time; neither store constructor has side effects.

#### Session module clean-up

Done along the way, in the same module:

* **The store is no longer reachable from outside.** It was exported through `applicationSession.store`, while a single call site actually needed it — the `sessionKill` resolver, only to strip the store key prefix off the session id. That prefix handling moves next to the store, so the resolver is now just `await killSession(id)`. The public API is metadata plus termination; the store, which exposes the full session content and `set()`, stays private.
* **The store is wrapped once, as promisified operations.** `express-session` is callback-only end to end (the `Store` interface, the session object methods, the middleware itself) with no promise API and no `util.promisify.custom` hook, so the conversion happens in one place next to the store creation, and the wrapper states the three store operations the platform actually relies on out of the eight in the interface.
* **The session policies become operations of the module.** The concurrent sessions limit and the "kill every session but the current one" scenario were composed in the user domain from the list and bulk kill primitives, which had to be public for that. They are now `killUserSessionsOverLimit(userId, max)` and `killOtherUserSessions(userId, currentSessionId)`, so the bulk kill stays private. The domain keeps what is its own — reading the setting — and the `enforceSessionLimit` wrapper, reduced to adapting its arguments for a single caller, is dropped.
* **The kill operations return only what their callers use**: a count for the concurrent limit, nothing for the "other sessions" scenario. `killSession` and `killUserSessions` keep returning the killed sessions, which the `sessionKill` and `userSessionsKill` mutations need for their audit action and their GraphQL result.
* **Hand written promise wrappers are replaced by `util.promisify` / `events.once`, and ramda by the standard library** (`Object.groupBy`, `Array.find`). Two incidental fixes come with it: the websocket session parsing never resolved when the store returned an error or no session, and `findSessions` silently returned nothing for object shaped stores — the `Store` interface allows the sessions to be returned either as an array or as an object, so the wrapper normalizes it.

#### Tests

* `tests/01-unit/database/session-test.ts` drives real HTTP nodes against a shared session store, the encryption key being the only difference between them: same key resumes the session, a different key rejects the cookie and yields an empty session. It also covers the two session policies, which had no test at all before — no test in the repository referenced `enforceSessionLimit`, and neither `user-test.ts` nor the password expiry suite mentioned sessions.
* `tests/01-unit/utils/platformCrypto-test.ts` covers `deriveSecret`: determinism, requested length, separation across seeds, paths and versions, no collision with the other derivations of the same path, and input validation.

### Related issues

* closes #15353

### How to test this PR

The key is used by `express-session` to sign (HMAC-SHA256) the session id carried by the `opencti_session` cookie. It authenticates the session id only: the session content lives in the session store. So a node accepts a session when **both** the signature verifies (same encryption key) **and** the session id is found in its store (`app:session_manager: shared` with the same Redis and `redis:namespace`).

1. Nominal: with an `app:encryption_key` configured and no `app:session_secret`, log in and browse. Sessions, subscriptions (websocket) and `Settings > Security > Sessions` session kill behave as before.
2. Restart the platform with the same encryption key: the session cookie stays valid, no new login (derivation is deterministic).
3. Cluster: start two nodes with the same encryption key, `app:session_manager: shared` and the same Redis (same `redis:namespace`), reached on the same host and `app:base_path`. A session created on the first node is accepted by the second one. Without a shared session store, the second node cannot find the session and requires a login — that requirement is pre-existing and unrelated to the signature key.
4. Key change: restart with a different `app:encryption_key`, then reload the UI. The cookie signature no longer verifies, a new empty session is created and the login page is shown. Beware that `APP__ENCRYPTION_KEY` in the environment takes precedence over the JSON configuration files (`nconf.env()` is registered before `nconf.file()`), and that a running process keeps the key it booted with.
5. Concurrent sessions limit: set `platform_session_max_concurrent` in the settings and log in beyond it — the oldest sessions are terminated, and the login event reports how many.
6. Unit tests: `yarn test:ci-unit`.

To check what a given key derives to, without starting the platform:

```
node -e "const c=require('crypto');const k=Buffer.from(process.env.APP__ENCRYPTION_KEY,'base64');c.hkdf('sha256',k,Buffer.alloc(0),Buffer.from('session:signature:secret::v1'),32,(e,b)=>console.log(Buffer.from(b).toString('base64')))"
```

### Checklist

### Further comments

**Upgrade impact.** The session signature key changes, so every existing session cookie becomes invalid and users have to log in again once. Deployments that had explicitly set `app:session_secret` will see it silently ignored, with the same one-time consequence. No new configuration requirement is introduced: `app:encryption_key` is already needed for the platform to start, for API token hashing and credentials encryption.

**Pre-existing behaviours kept as is, worth a separate look:**

* `killOtherUserSessions` kills nothing when there is no current session — the original filter carried a `currentSessionId &&` guard that short circuited for every session. Arguably it should kill them all, since that is exactly the case where a forced password change matters most. Behaviour is preserved here, and the new test pins it, so whoever changes it will see a red test rather than a silent surprise.
* The in memory session store (`app:session_manager` other than `shared`) does not really support session listing and killing: session ids come from `redis_key_id`, and its `destroy` never calls back because `LRUCache.get` is called with a callback.
* `req.session.save()` is called without a callback, so the session write is neither awaited nor checked for errors.
* `domain/user.js` still holds 50 ramda usages over 2397 lines. Converting them would have drowned this change and concentrated regression risk on the most sensitive file of the platform, so it is left out.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/opencti-platform/opencti-graphql/src/database/session.js
+++ b/opencti-platform/opencti-graphql/src/database/session.js
@@ -1,12 +1,20 @@
 import session from 'express-session';
 import nconf from 'nconf';
-import * as R from 'ramda';
+import { promisify } from 'node:util';
 import conf, { booleanConf, OPENCTI_SESSION } from '../config/conf';
 import SessionStoreMemory from './sessionStore-memory';
 import RedisStore from './sessionStore-redis';
+import { getPlatformCrypto } from '../utils/platformCrypto';
+import { memoize } from '../utils/memoize';
 
 const sessionManager = nconf.get('app:session_manager');
-const sessionSecret = nconf.get('app:session_secret') || nconf.get('app:admin:password');
+
+// The session cookie signing key is derived from the platform encryption key.
+// It is then automatically shared between all the platform nodes and does not rely on any user provided secret.
+const getSessionSecret = async () => {
+  const factory = await getPlatformCrypto();
+  return factory.deriveSecret(['session', 'signature'], 1);
+};
 
 const createMemorySessionStore = () => {
   return new SessionStoreMemory({
@@ -18,17 +26,24 @@ const createRedisSessionStore = () => {
     ttl: conf.get('app:session_timeout'),
   });
 };
-const createSessionMiddleware = () => {
+const createApplicationSession = async () => {
   const isRedisSession = sessionManager === 'shared';
   const store = isRedisSession ? createRedisSessionStore() : createMemorySessionStore();
   const isSessionCookie = conf.get('app:session_cookie') ?? false;
   const sessionTimeout = isSessionCookie ? undefined : conf.get('app:session_timeout');
   return {
-    store,
+    // The store API is callback based and gives a full read/write access to every user session:
+    // only the promisified operations needed by this module are exposed.
+    sessionStore: {
+      prefix: store.prefix,
+      // the store interface allows an array or an object of sessions, depending on the implementation
+      all: async () => Object.values(await promisify(store.all.bind(store))()),
+      destroy: promisify(store.destroy.bind(store)),
+    },
     session: session({
       name: OPENCTI_SESSION,
       store,
-      secret: sessionSecret,
+      secret: await getSessionSecret(),
       proxy: true,
       rolling: true,
       saveUninitialized: false,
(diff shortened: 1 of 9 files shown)


fellowseb:
During my tests I couldn't pass the following:

> - Start a second node with the same encryption key: a session created on one node is accepted by the other.

Still had to log in on the second node.

> - Change app:encryption_key: existing session cookies are rejected and users have to log in again.

Changing `app:encryption_key` does not log the user out.

opencti-platform/opencti-graphql/src/database/session.js
fellowseb, line 1: (One day we'll need to move the content of this file out of `database/` and into its own module, or a larger but more relevant one)

fellowseb (approved):

fellowseb:
> During my tests I couldn't pass the following:
> 
> > * Start a second node with the same encryption key: a session created on one node is accepted by the other.
> 
> Still had to log in on the second node.
> 
> > * Change app:encryption_key: existing session cookies are rejected and users have to log in again.
> 
> Changing `app:encryption_key` does not log the user out.

Re-tested, all good

Issue #17926, opened 2026-08-27
feat(fds): implement input


sandyghs:
## Use case

The Filigran Design System now ships form field components (Input, Textarea)
that are the reference implementation of the design language across all Filigran
products. OpenCTI currently renders its form fields with MUI TextField, which
diverges from the design system on placeholder tone, field height, label
placement and focus behaviour.

Adopting the library components brings OpenCTI in line with the design system,
removes the need for per-site style overrides, and makes future design changes
propagate automatically instead of requiring a product-side sweep.

Measured surface on the current tip:
- 60 standalone MUI TextField mounts (excluding Autocomplete render inputs)
- 54 multiline sites (11 raw + 43 through the shared pivot)

## Current workaround

Every form field is rendered with MUI TextField, each carrying its own style
overrides. Divergences from the design system are corrected site by site when
noticed, which does not scale and drifts again on the next design change.

## Proposed solution

Convert the MUI TextField mounts to the design system Input and Textarea
components, in waves:
- standalone single-line fields → Input
- multiline fields → Textarea
- sites owned by other in-flight branches are left untouched until those merge

Sites that cannot be converted mechanically (dynamic type, prop spreads,
slotProps escape hatches) are listed with an explicit reason rather than
rewritten.

## Additional information

This work is part of the design system migration and follows the same pattern as
the Breadcrumbs adoption (#17820) and the ongoing selection field conversion
(#17884).

PR #17979, opened 2026-08-29
feat(fds): thaw wave — the field sites #17884 used to own (#17926)

sandyghs:
Stacked on #17946 (`a4e25d4863`) per Sandy's option A — **merges after it**.

## Delivered

| Family | Before | After |
| --- | --- | --- |
| Checkbox | 26 MUI / 45 FDS | **6 MUI / 65 FDS** |
| Input (TextField sites) | 540 / 619 | **541 / 619** |

19 Checkbox sites converted. `edge`/`disableRipple` dropped rather than translated — both compensate for MUI's 9px padding box, which the 16px control does not have. Handlers moved to `onCheckedChange` only where the event was read for `target.checked` and nothing else.

Only each element's own span is rewritten, never the surrounding JSX — the restriction that the #17976 Switch sweep lacked when it swallowed an enclosing `Box`, a `Typography` and a sibling while still parsing.

Plus the `askIa` typo: one occurrence, consumed nowhere, made the prop unplaceable and pushed the site to MUI.

## Abandoned, with reasons (7 Checkbox)

| Reason | Count |
| --- | --- |
| inside `FormControlLabel` — needs unwrapping, file-by-file | 3 |
| carries `style` / `sx` — geometry rule | 2 |
| handler is a reference with an MUI signature | 2 |

## Not started

Textarea (24 multiline) and the direct Switch sites. See the report — the Switch lists must be re-derived against the merged #17976, which this base does not contain.

Gates: `check-ts` 0, lint clean, `FormControlLabel` injection sweep clean.


sandyghs:
Superseded by #17981 — closing so the two are not open at once.

Ownership of this branch moved to the overnight session. It was **merged**
into #17981 rather than rebased forward, because #17946 had already landed
almost all of it: of the nine commits here, one applied clean, one was dropped
by git as *already upstream*, and the first — which carries the same title as
#17946's own merge commit — conflicted in four files.

The merge measured the real remainder: **5 files, +21/−27**, all of it now in
#17981.

- `WorkbenchFileContent.jsx` — 12 Checkbox mounts onto the library
- `StixCoreObjectQuickSubscription.tsx`, `TriggerLiveCreation.tsx` — the
  `renderOption` checkbox (display-only, `checked` alone)
- `FileIndexingConfiguration.tsx` — one Checkbox, `onChange` → `onCheckedChange`
- `WorkbenchCreation.tsx` — the `askIa` → `askAi` typo that was keeping that
  site off the Input path

All 14 conflicts were resolved to the base's side, for one reason: **the base
is ahead of this branch, not behind it.** `EntitySelect`, `HiddenTypesField`,
`DataTableLine`, `FilterChipPopover`, `DataTableToolBar`,
`WorkspaceTurnToContainerDialog` and the two widget inputs were all converted
by #17946, so this branch's side of each hunk was the older MUI code.

The `scripts/check-utility-classes.js` path proposed here is dropped: the gate
is already in the base at `script/check-utility-classes.js`, which is the
directory this workspace actually uses.

The branch is left in place; nothing was pushed to it.

Issue #57, opened 2019-06-11
Observables : scoring/rating


Fred-certeu:
Please replace every line in curly brackets { like this } with appropriate answers, and remove this line.

## Problem to Solve

We use it for selecting observable at SOC level.
Some observables correspond to legitimate IP addresses, domain names, etc. Their use in consuming devices will generate false positive.
Some observables might be 'malicious' for sometimes.
The rating/scoring will indicate 'how malicious' a give observable is.

## Current Workaround

N/A

## Proposed Solution

Scoring/rating: numerical value associated to each observable.
Range: [0-100]
(0: non malicious)
(100: malicious)
The logic to calculate or set the rating can be different depending on the organisation.
It should be subject to a separate 'issue'

## Additional Information

N/A

PR #18001, opened 2026-08-30
feat(fds): redesign the tab-scoped right bars, keep their position (#17539)

sandyghs:
> **Now stacked on #18019**, which is where `utils/fdsLayer.ts` lives.

**This PR was replaced.** Its previous content moved the tab-scoped right bars
below the tabs; Sandy's verdict on the deployed branch was that the page content
passes underneath and it reads broken. F1 asks for the opposite — the bars back
in their ORIGINAL place, with the redesign kept. The structural move was never
merged, so the tip still had the original placement; this branch touches only
the *inside* of the bars.

### Knowledge bar — Figma node 7472:48226 (`navbarSecondary`)

- Rows are the library `NavbarItem` — the component the node itself places here
  (its instances are named `navbar menu item`). 36px row, 14px label, 16px left
  inset; hover and selected tones are the library's, not MUI's `selected`.
- Group headers are `NavbarTitle`: Content/caption at 10px in
  `--text-default-disabled`. Rendered `<p>`, not the default `<h2>` — this bar
  is chrome beside the entity's heading outline, and a heading here would land
  out of order (WCAG 1.3.1).
- 1px left edge on `--border-elevation-subtle-soft`, 4px top inset.

### The layer, via the shared helper

`RIGHT_BAR_LAYER` is **1**, not the drawers' 2: a tab-scoped bar is chrome
beside the content, not a panel over it. Pinned by two independent hexes from
the node — `--bg-elevation-default` #0d172b and `--bg-elevation-highlight`
#182a4e are the layer-1 values; a bare alias resolves layer 0 (#070d18), which
is what the bars painted before.

Both bars use `fdsLayerClass(RIGHT_BAR_LAYER)` **with `layerInputVars`**, on
`slotProps.paper` rather than the deprecated `PaperProps`. The vars matter: the
class and the three input aliases must sit on the same node or fields inside a
bar resolve at the root (LIBRARY-FEEDBACK #57).

### The rest of the F5 sweep

Three more raw `Drawer` mounts hold real fields and take the same layer-1
treatment: `GraphToolbar` (a SearchInput), `EntitiesDetailsRightBar` (a Select),
`ContainertKnowledgeTimeLineBar` (a SearchInput). The pure-chrome ones —
`StixCoreObjectContentBar`, `NavToolbarMenu`, `WorkbenchFileToolbar`,
`settings/sub_types/ToolBar.tsx`, `data/ToolBar.jsx` — get nothing: no field of
their own, surfaces not reported, and repainting them uninvited is the mistake
F23 records.

### Content view switcher

Was `ButtonGroup size="sm"` (24px), visibly smaller than everything else on that
row. It is `md` (36px) now, and its negative-margin pull onto the tabs strip was
re-derived by measurement rather than guessed: at -70px the group's centre sat
3px above the strip's, so it is -67px — centre on centre, delta 0. Glyphs stay
16px, because the library fixes the item's icon slot at `size-4` for both sizes.

### Positioning

Untouched: fixed, right-attached, full height, page content beside them.
`getPaddingRight` is exactly as the tip has it.

### Measured on the authenticated stack (1600x900)

| | measured | target |
|---|---|---|
| paper background | `rgb(13,23,43)` #0d172b | `--bg-elevation-default-layer-1` |
| highlight inside the bar | `rgb(24,42,78)` #182a4e | `--bg-elevation-highlight-layer-1` |
| left edge | #2b4f8d @ 15% | `--border-elevation-subtle-soft` |
| geometry | x=1400, y=0, h=900, w=200 | right edge, full height |
| row | 36px, label 14px, pad-left 16px | 36 / 14 / 16 |
| group header | 28px, 10px, #95969d, `<p>` | caption, disabled tone |
| switcher | 36px, centre delta 0 vs the tabs strip | md, aligned |

### E2E

The bar's rows are real links with `aria-current` now, not MUI `MenuItem`s with
`role=menuitem` — a menu role implies an application menu with roving focus,
which permanent navigation never was. Page objects selecting the old role timed
out. Swept by checking every `getByRole('menuitem')` in the suite against the
bar's own section labels: three selectors moved to `role: 'link'`; all the
others name real popup-menu commands and are untouched.

FIXLOG: F1 awaiting Sandy's validation of both halves; part of F5.

### DIFF
--- a/opencti-platform/opencti-front/src/components/ThemeDark.ts
+++ b/opencti-platform/opencti-front/src/components/ThemeDark.ts
@@ -554,8 +554,15 @@ const ThemeDark = (
           '&.Mui-error .MuiOutlinedInput-notchedOutline': {
             borderColor: 'var(--border-input-error)',
           },
+          // Disabled follows the library `Input`, which goes TRANSPARENT with a
+          // disabled border — not `--bg-input-disabled`. Painting that token
+          // put a light grey slab in the middle of a dark form (seen on the
+          // public-dashboard URI KEY field, which is always disabled).
           '&.Mui-disabled': {
-            backgroundColor: 'var(--bg-input-disabled)',
+            backgroundColor: 'transparent',
+            '& .MuiOutlinedInput-notchedOutline': {
+              borderColor: 'var(--border-elevation-disabled)',
+            },
           },
         },
         input: {
--- a/opencti-platform/opencti-front/src/components/ThemeLight.ts
+++ b/opencti-platform/opencti-front/src/components/ThemeLight.ts
@@ -552,8 +552,15 @@ const ThemeLight = (
           '&.Mui-error .MuiOutlinedInput-notchedOutline': {
             borderColor: 'var(--border-input-error)',
           },
+          // Disabled follows the library `Input`, which goes TRANSPARENT with a
+          // disabled border — not `--bg-input-disabled`. Painting that token
+          // put a light grey slab in the middle of a dark form (seen on the
+          // public-dashboard URI KEY field, which is always disabled).
           '&.Mui-disabled': {
-            backgroundColor: 'var(--bg-input-disabled)',
+            backgroundColor: 'transparent',
+            '& .MuiOutlinedInput-notchedOutline': {
+              borderColor: 'var(--border-elevation-disabled)',
+            },
           },
         },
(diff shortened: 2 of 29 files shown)


sandyghs:

one merge.

needed.

PR #18014, opened 2026-08-30
fix(fds): the two twice-reported findings, F15 and F8 (#17808)

sandyghs:
Two findings from Sandy's visual pass that she reported **twice**, plus the
FIXLOG that now tracks every finding of that pass.

### F15 — custom dashboard Refresh control (icon/label misalignment + missing gap)

One root cause for both halves: the control was wrapped in a MUI `ButtonGroup`,
which joins its children edge-to-edge (hence the zero gap) and styles them as
MUI buttons — which the library `Button` and `Select` are not (hence the
baseline misalignment). Replaced by a plain flex row with `gap: 1`, and the
refresh glyph is sized to the spinner it alternates with.

### F8 — `type=number` fields still MUI (validity period, max concurrent sessions)

The previous round converted only the fields blocked by `slotProps`, so these
two were never in scope. This round enumerates all 90 `type="number"` inputs and
classifies each as pivoting / MUI-fallback / raw:

- `PeriodicityField` bypassed the pivot entirely → now the library `Input` with
  `isTypeNumber` (designed steppers).
- `AuthenticationGlobalSettings` was blocked by `inputProps` → `min={0}`.

### FIXLOG

`fds-migration/FIXLOG.md` registers every finding of the pass, F1–F26, with a
status each. Nothing leaves the log by omission: an item ends FIXED (with how it
was verified on screen), FIXED-UNVERIFIED, or BLOCKED with a reason.

Both fixes here are **FIXED-UNVERIFIED** — code is in, not yet exercised on the
authenticated stack.

### DIFF
--- a/fds-migration/FIXLOG.md
+++ b/fds-migration/FIXLOG.md
@@ -15,7 +15,7 @@ Classes: **(a)** regression → fix · **(b)** fixed-at-tip → name the PR ·
 | # | Screen | What's wrong | Class | Status |
 |---|---|---|---|---|
 | F1 | Entity tabs, Content tab | Right bars: return to ORIGINAL position (right edge, full height), content back in place — "everything goes underneath". Keep the redesign. | a | OPEN |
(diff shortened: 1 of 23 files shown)


sandyghs:

one merge.

needed.

Issue #17729, opened 2026-08-16
fix(fds): revive the token references the rename killed and measure the container surfaces


sandyghs:
## Use case

The design-system pin is several library releases behind. Bringing it up to date carries a token-rename campaign (17 alpha tokens renamed), the navigation aligned on its Figma component sets, and a lower Paper border — but a rename also breaks host code silently, and OpenCTI proved to be the worse case.

Three token references were dead in this product and none of them was signalled: `tsc --noEmit`, `eslint` and the build all pass on the breakage.

- one reads the old name → a dangling `var()`
- one **declares** the old name → the selected navigation row was painted in the default brand colour instead of the customer accent. This is a shape OpenAEV did not hit: a rename breaks what a host *writes*, not only what it reads
- one asserts the old name in a test → the test stayed green on a dead reference

Before any surface is converted, the migration also needs a measured picture of what the product actually does today, so that every deviation from the library is arbitrated rather than discovered site by site.

## Current workaround

None. A dead token reference degrades a colour silently and no tool reports it; the only signal is a user noticing a wrong colour. The gap between the product's surfaces and the library's Paper is currently unknown.

## Proposed solution

Three things, no conversion:

1. **Bump the pin** to the head of the library, and prove it by the served bytes rather than by the lockfile. The token bridge is regenerated by its own script and its freshness verified by hashing.
2. **Revive the three dead references**, and make the hunt reproducible: both silent shapes (a `var()` inside a string, a utility class as a literal) are searched with a strict word boundary and cross-checked against the installed stylesheet — a naive search gives a false green, because the old name is a prefix of the new one.
3. **Measure the 28 container surfaces** on the rendered DOM, in the real theme, with the application's complete stylesheet stack, and record every gap that needs a decision before conversion.

Expected and shown, not a regression: the warning colour darkens in light mode only, which improves its contrast on white.

Out of scope: no surface is converted. The inventory ends on a STOP; conversion waves follow in their own pull requests.

## Additional information

The navigation guard went red on a token the library now uses on purpose. It was narrowed onto its stated intent rather than weakened, so anything a future pin introduces is still red by default.

## If the feature request is approved, would you be willing to submit a PR?

Yes — the PR is already open.

sandyghs:
Closing as absorbed into #17728. The work described here — the pin bump, the three dead token references and the surface measurement — is delivered in the same pull request as the first conversion wave, so it is tracked under the umbrella issue rather than separately.

PR #18015, opened 2026-08-30
fix(fds): repair the date-filter interval popover, F21 (#17729)

sandyghs:
F21 of the visual pass: the date filter's *within / interval* popover. Two real
defects, both regressions from an earlier conversion pass.

### 1. Adornment glyphs spilling out of the field

The `endAdornment` held **library IconButtons inside a MUI outlined field**, and
the size mismatch had been compensated by hand with `marginRight: -8` and
`marginLeft: 4, marginRight: -16`. Those negative margins pushed both glyphs
OUT of the field's padding box — the spill visible in Sandy's capture. Replaced
by a flex row with a real 4px gap.

### 2. Two fields fighting for focus

`autoFocus` was hardcoded `true` in `RelativeDateInput`, which `DateRangeFilter`
renders **twice**. Both fields claimed focus on open. It is a prop now, and only
*From* takes it.

**Verified on screen** (authenticated stack, Observables → filter *Date de
création*): the adornment's right edge now sits **10px inside** the field
instead of outside it, and exactly one element holds focus after the popover
opens.

This branch also carries the authoritative copy of `fds-migration/FIXLOG.md` —
identical to the copy in #18014, so the two merge in either order.


sandyghs:

one merge.

needed.

Issue #17664, opened 2026-08-11
feat(fds): integrate Filigran Design System v1


labo-flg:
## Description

Filigran new Design System is implemented as a new library of UI components, shared between all products in the XTM Suite.

PR #18021, opened 2026-08-30
fix(fds): navbar active state, EE chip size, AI glyph, integration surface (#17664)

sandyghs:
> **Stacked on #18020** (and through it on #18019 / #18003).

Five details from the visual pass, each root-caused rather than nudged, plus one
regression caught while exercising the previous PR.

### F14 — navbar sub-item inside a custom dashboard

`Custom dashboards` carried `exact: true`, so the moment you opened a dashboard
(`/…/dashboards/<id>`) the sub-item went inactive. Prefix matching is safe
between the two dashboard links — `/…/dashboards_public` does not start with
`/…/dashboards/` — so the flag simply should not have been there. Removed on
Public dashboards too, which had the same defect.

**Verified:** inside a dashboard, `Tableaux de bord personnalisés` is
`aria-current="page"` and `Tableaux de bord publics` is not.

### F24 — Settings right-nav EE chips

Take the library's `sm` variant. Measured 18px tall.

### F18 — "Aperçu par IA" glyph

`FiligranIcon`'s `size="small"` is MUI's 20px, and this glyph's viewBox is not
square, so it rendered **14×20 inside a 24px button** — a pixel taller than the
button itself, and off its centre line. It asks for 16 now, which is the
library's own small-button glyph size.

### F23 — integration blocks

The homogenisation pass pointed `paperBg` at `--bg-elevation-default`, which in
dark mode is `#070d18`: the **layer-0** surface, darker than the page the boxes
sit on. That is the background nobody asked for. Reverted to
`background.paper`, exactly what these boxes had before. The token border stays —
it was not part of the report, and it replaced a border derived from the *text*
colour, which is what kept these boxes from matching a real Paper.

### F17 (spacing half) — public dashboard dialog

16px between fields. Scoped to the reviewed surface; the app-wide
`fieldSpacingContainerStyle` is still 20px, so generalising the rhythm stays
Sandy's call. **The "+" button half of F17 is NOT in this PR** — see the FIXLOG:
no `+` control exists in that drawer or on the public-dashboards list, so the one
in the capture needs pointing at before it can be changed.

### Regression caught while exercising #18020

The outlined restyle gave disabled fields `--bg-input-disabled`, which in dark
mode is a light grey slab — plainly visible on the always-disabled URI KEY
field. The library's own `Input` goes **transparent with a disabled border**
instead. Fixed here rather than left for the next round.

FIXLOG: F14, F18, F23, F24 → FIXED. F17 → PARTIAL (spacing fixed, "+" blocked
on identifying the control).

### DIFF
--- a/opencti-platform/opencti-front/src/components/ThemeDark.ts
+++ b/opencti-platform/opencti-front/src/components/ThemeDark.ts
@@ -554,8 +554,15 @@ const ThemeDark = (
           '&.Mui-error .MuiOutlinedInput-notchedOutline': {
             borderColor: 'var(--border-input-error)',
           },
+          // Disabled follows the library `Input`, which goes TRANSPARENT with a
+          // disabled border — not `--bg-input-disabled`. Painting that token
+          // put a light grey slab in the middle of a dark form (seen on the
+          // public-dashboard URI KEY field, which is always disabled).
           '&.Mui-disabled': {
-            backgroundColor: 'var(--bg-input-disabled)',
+            backgroundColor: 'transparent',
+            '& .MuiOutlinedInput-notchedOutline': {
+              borderColor: 'var(--border-elevation-disabled)',
+            },
           },
         },
         input: {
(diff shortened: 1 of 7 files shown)


sandyghs:

one merge.

needed.

PR #18023, opened 2026-08-30
fix(fds): put the whole filter bar on one line, everywhere (#17664)

sandyghs:
> **Stacked on #18021** (and through it on #18020 / #18019 / #18003).

F13 and F20, both **twice reported**. Two independent causes, both found by
measuring the row rather than by adding another `alignItems`.

### The Add-filter picker was claiming the whole line

The library `Combobox` **root** carries `flex w-full flex-col`, so inside a flex
row it takes the full width and pushes the search field, the funnel and the
chips onto lines of their own. The inner `ComboboxField` had already been pinned
to 200px in an earlier pass — the root is the part that was never told. It is
`w-50 shrink-0` now, and tailwind-merge drops the `w-full` it replaces.

### The chips were a sibling of the toolbar, not part of it

Both list engines rendered them *below* the row: `ListLines` after the
`parameters` row's closing tag, and `DataTable` as `DataTableDisplayFilters`
after the whole toolbar. Sandy's rule is "everything must sit on ONE line", so
in both engines the chips moved INSIDE the filter row, before the filler that
right-aligns the view controls. Both rows wrap, so a long chip run still
degrades gracefully instead of overflowing.

### Why the previous round could not reach these pages

It added `alignItems: center` to two specific containers. The defect was never
alignment: it was one child taking the full width, and another being outside the
row entirely. Recorded in the FIXLOG so this is not re-derived a third time.

### Verified on screen — three pages, three engines

| page | engine | before | after |
|---|---|---|---|
| Triggers | `ListLines` | four lines | one: search / Add filter / funnel / two chips / count / two buttons |
| Reports | `DataTable` | two lines | one, including the saved-filter select, pagination and export |
| Threat actors (group), card view | card list | two lines | one |

FIXLOG: F13, F20 → FIXED.

### DIFF
--- a/opencti-platform/opencti-front/src/components/ThemeDark.ts
+++ b/opencti-platform/opencti-front/src/components/ThemeDark.ts
@@ -554,8 +554,15 @@ const ThemeDark = (
           '&.Mui-error .MuiOutlinedInput-notchedOutline': {
             borderColor: 'var(--border-input-error)',
           },
+          // Disabled follows the library `Input`, which goes TRANSPARENT with a
+          // disabled border — not `--bg-input-disabled`. Painting that token
+          // put a light grey slab in the middle of a dark form (seen on the
+          // public-dashboard URI KEY field, which is always disabled).
           '&.Mui-disabled': {
-            backgroundColor: 'var(--bg-input-disabled)',
+            backgroundColor: 'transparent',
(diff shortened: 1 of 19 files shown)


sandyghs:

one merge.

needed.

PR #18039, opened 2026-08-31
feat(merge-users): disable the source, emit the merge trace and gate the source deletion (#17499)

### LINKED ISSUE #17499: PR7 — Finalize: disable source + merge trace (source deletion coverage-gated, post-MVP)
### Objective

Finalize a successful merge: disable the source and leave a readable trace of the operation. The source user is kept (disabled, not deleted). Deleting it is a separate, coverage-gated operation that is out of the MVP.

Note: an earlier version of this issue specified a dedicated UserMerge audit record as the evidential artefact and made it the second condition of the deletion gate. That design was dropped on 2026-08-31 — it came from the technical refinement rather than from any product decision, and the gate needs a computation rather than a memory. See the Notion page for the full reasoning.

### Why deletion is not in the MVP

PR9 (filters), PR10 (JSON/Base64 manifests) and PR11 (edge cases) are post-MVP, so register rows a handler must claim are still uncovered when this PR runs. Deleting the source at that point breaks things silently: a filter pointing at a deleted user matches nothing, it does not raise an error. Triggers stop firing, Feed/Stream/TAXII sharing returns empty, dashboard widgets show zero — with nothing in the logs.

Drafts make it worse: draft_updates_patch is persistent WIP that stopping the workers does not flush, so validating a pre-merge draft re-injects the deleted UUID into live data long after the merge is closed and its report archived.

Functionally, keeping the source costs nothing: what the client needs at go-live is that the new SSO account finds and owns its past data (PR3–PR6). A disabled source row does not get in the way — and it keeps the operation reversible.

The product decision of 2026-07-24 reads "on merge initial user deleted if script ran successfully". Disable-then-delete is a deliberate departure from it, justified by the partial coverage above, and confirmed on 2026-08-31: deletion remains the endpoint, it is simply gated rather than automatic.

### Scope — what is expected

- Mark the source as disabled, without deleting it, and without adding any attribute to the User model. Disabling uses an existing field: `Expired` and not `Locked`, since `Locked` exists as a constant but is absent from the default status table and writing it would put the account in a state the platform does not offer.
- Disabling is a precondition, not a finalization step. An active source can log back in and create data after its objects have been transferred, so it runs first, before any reference is moved. It is itself a handler, so it appears in the dry-run report and in the journal like every other step.
- Emit the activity trace "user A has been merged into user B", carrying source, target, actor, strategy and outcome. Its two known limits are accepted because nothing irreversible depends on it: activity events are Enterprise Edition only, and they are purgeable by a retention rule.
- The execution journal written by the engine (PR2) remains the diagnostic artefact of the run.
- The target's profile attributes always win: name, firstname, lastname, user_email and the rest of the identity fields are never overwritten by the source's values. This is a merge policy, not a technical limitation — the merge moves owned objects, it does not reconcile identities. Enforced at handler registration, so it applies to every handler written later.
- No re-merge guard, deliberately. Re-running the same pair, or chaining, is allowed and converges: it is the catch-up path once coverage grows.
- The identity guarantee is an occupancy, not an invalidation: a disabled account occupies its email and blocks session creation. Once the source is deleted the email is free again — documented, not engineered around.
- Abort cleanly on any prior handler failure. This is a traceability guarantee, not an atomicity one: the platform has no rollback primitive, so what is required is that the journal names exactly what was applied and what was not.

### Deletion — the gate ships here, the deletion itself in PR11

- The gate is exposed as a read, `userMergeSourceDeletionReadiness(sourceId, targetId)`, returning whether the deletion is allowed and naming the blockers. A mutation that validates and then always refuses would be a trap for its caller.
- Hard gate, two conditions:
  - Coverage is complete: no register row with disposition transfer or conditional is left uncovered by a registered handler, read from the code-side register shipped by PR2.
  - Nothing is left to move: a dry-run on the pair is executed at deletion time and every handler reports zero. The source is still present at that point — it is only disabled — so the dry-run has both of its operands.
- The destructive half is written in PR11, behind this same gate. Until PR9/PR10/PR11 land the gate refuses every time, so a deletion written now could not run and could not be exercised against a real platform.

### Out of scope

- The source deletion itself (PR11).
- Session / token / runtime invalidation (PR8 — now the primary security control, since the source survives in DB).
- The handlers themselves (PR3–PR6).

### Dependencies

- PR3 (scalars), PR4 (RBAC), PR5 (STIX).
- Deletion additionally requires PR9, PR10 and PR11 — enforced by the coverage gate, not by reviewer vigilance.

### Definition of Done

- Merge finalizes with the source disabled and kept; the source is never deleted in the MVP, and no attribute is added to the User model.
- The source is disabled before any reference is moved, and the disable is itself a handler.
- The activity trace "A merged into B" is emitted on every successful merge.
- Re-running a merge on the same pair is allowed and converges — it is the catch-up path once coverage grows.
- The gate refuses unless coverage is complete and a dry-run on the pair reports zero everywhere, and names which of the two failed.
- No handler can write a user identity field, enforced at registration.
- The disabled source occupies its email and blocks session creation; the post-deletion limit is documented rather than engineered around.
- Failure path leaves the execution journal readable and the merge explicitly marked as not successful — no claim of rollback.

ludovic:
### Proposed changes

* Disable the source account **before** any reference is moved, as a handler, so it appears in the dry-run report and in the journal like every other step. `Expired` and not `Locked`: `Locked` exists as a constant but is absent from the default status table.
* Emit the activity trace *"user A has been merged into user B"* on every successful run, from the domain layer — the engine deliberately runs with no calling user, and the activity listener only retains actions whose origin is a real request.
* Refuse at registration any handler declaring a write on a user identity field. A merge moves owned objects, it does not reconcile identities.
* Gate the source deletion on two conditions: coverage is complete **and** a dry-run on the pair plans nothing. **The deletion itself is not exposed here** — it lands in PR11, once the gate can actually open.
* Drop the two register rows describing the audit-record entity that was abandoned during scoping (101 → 99 rows, `v3` → `v4`).

### Related issues
* closes #17499

### How to test this PR

```graphql
mutation { userMerge(sourceId: "<A>", targetId: "<B>", options: { dryRun: false }) { status } }
query { userMergeSourceDeletionReadiness(sourceId: "<A>", targetId: "<B>") { allowed blockers } }
```

After the merge, A is `Expired`, its sessions are gone, and an administration entry *"merges user a@… into user b@…"* is filed on A in the activity log (EE only). The readiness query refuses today and names why: register rows a handler must claim are still uncovered until PR9–PR11 land.

### Checklist

### Further comments

**The source is disabled, not deleted, and that is a deliberate departure from the 2026-07-24 product decision** ("initial user deleted if script ran successfully"). At this point PR9–PR11 have not landed, so rows a handler must claim are still uncovered: deleting now would leave references pointing at a user that no longer exists — silently, since a filter naming a deleted user matches nothing rather than raising. Deletion remains the endpoint; it is gated rather than automatic.

**The gate re-computes rather than remembers.** An earlier version of the scoping specified a dedicated audit-record entity attesting that a full-coverage run took place. It was dropped: a record attests to the past and reads *complete* even when references to the source appeared after the merge (a manual operation, an import, a draft validated late). A dry-run at deletion time cannot be wrong about the current state, and it has both of its operands because the source is still there. Consequence: no new entity, no data-model change.

**Re-merging the same pair stays allowed, deliberately** — it is the catch-up path once coverage grows. Handlers are idempotent, so a second run is a no-op on what was already applied.

Stacked on #17951 (PR6). Review only the last five commits.


opencti-platform/opencti-graphql/src/domain/user.js
```diff
@@ -1415,7 +1416,21 @@ export const deleteAllNotificationByUser = async (userId) => {
...
   // check rights
-  await loadUserToUpdateWithAccessCheck(context, user, userId);
+  const userToDelete = await loadUserToUpdateWithAccessCheck(context, user, userId);
+
+  // The four cascades below all select by a reference to the account being deleted. On an account
+  // a merge has emptied that is the wrong trade: they are no-ops when the merge moved everything,
+  // and destructive when it missed something — a Trigger or a Workspace the merge failed to
+  // re-point now belongs to the target, and the cascade deletes it rather than skipping it.
+  // Refusing here keeps a coverage gap re-runnable instead of turning it into a deletion.
+  // Deliberately not absolute: clearing `merged_into` through the ordinary field patch re-opens
```
marieflorescontact, line 1426: if I understand well, if merge goes not well, `merged_into` will be set for User and it will be impossible to delete it, except by field patching  manuallly the `merged_into` field ? is it ok

opencti-platform/opencti-graphql/src/modules/userMerge/userMerge-sourceDisableHandler.ts
```diff
@@ -0,0 +1,90 @@
...
+   * of the status change, which a raw index write would not do.
+   */
+  apply: async ({ context, sourceId, targetId }: UserMergeHandlerContext, plan: UserMergeHandlerPlan): Promise<number> => {
+    const planned = plan.changes.some((change) => change.register_row_id === SOURCE_DISABLE_ROW && change.count > 0);
+    if (!planned) {
+      return 0;
+    }
+    await userEditField(context, SYSTEM_USER, sourceId, [
+      { key: ACCOUNT_STATUS_FIELD, value: [ACCOUNT_STATUS_EXPIRED] },
+      { key: USER_MERGED_INTO_FIELD, value: [targetId] },
```
marieflorescontact: If userA is merged into userB and then userB is merged into userC, should we change the `merge_into` value on userA from userB to userC ?

opencti-platform/opencti-graphql/src/domain/user.js
ludovic: Yes, that's correct
If a merge fails mid-flight during the write pass, some references may already have been transferred. Allowing standard userDelete at that point would trigger destructive cascades on objects that now belong to the target.
The expected recovery path is to re-run the merge
However, if an operator really wants to force-delete the source anyway, clearing merged_into via standard field edit re-opens userDelete. It was deliberately left editable as a safety hatch so an admin is never stuck.

PR #18084, opened 2026-09-03
fix(user): normalize email to avoid duplicate (#18075)

### LINKED ISSUE #18075: fix(user): normalize email to avoid duplicate
## Environment

2. Version: { e.g. 1.2.3 }
3. Other environment details:

## Reproducible steps

1. { e.g. Run ... }
2. { e.g. Click ... }
3. { e.g. Error ... }

marieflorescontact:
### Proposed changes

* Create a utils method that normalize email to avoid duplicates 

### Related issues

* closes #18075 

### Checklist

### DIFF
--- a/opencti-platform/opencti-graphql/src/domain/user.js
+++ b/opencti-platform/opencti-graphql/src/domain/user.js
@@ -876,24 +876,19 @@ export const sendEmailToUser = async (context, user, input) => {
 };
 
 export const addUser = async (context, user, newUser) => {
-  let userEmail;
   const userServiceAccount = newUser.user_service_account;
-  if (newUser.user_email && !userServiceAccount) {
-    userEmail = normalizeEmail(newUser.user_email);
+  if (!newUser.user_email && !userServiceAccount) {
+    throw FunctionalError('User cannot be created without email');
+  }
+  const userEmail = newUser.user_email ? normalizeEmail(newUser.user_email) : `automatic+${uuid()}@opencti.invalid`;
+  if (isEmptyField(userEmail)) {
+    throw FunctionalError('The email you have provided is not valid');
+  }
+  if (newUser.user_email) {
     const existingUser = await elLoadBy(context, SYSTEM_USER, 'user_email', userEmail, ENTITY_TYPE_USER);
     if (existingUser) {
       throw FunctionalError('User already exists', { user_id: existingUser.internal_id });
     }
-  } else if (userServiceAccount) {
-    userEmail = newUser.user_email ? normalizeEmail(newUser.user_email) : `automatic+${uuid()}@opencti.invalid`;
-    if (newUser.user_email) {
-      const existingUser = await elLoadBy(context, SYSTEM_USER, 'user_email', userEmail, ENTITY_TYPE_USER);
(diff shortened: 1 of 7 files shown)


opencti-platform/opencti-graphql/src/utils/email.ts
```diff
@@ -0,0 +1 @@
+export const normalizeEmail = (email: string): string => email.trim().toLowerCase();
```
ludovic: The function is called by different JS files; it is preferable to add a guard.

```suggestion
export const normalizeEmail = (email?: string | null): string => {
  if (!email || typeof email !== 'string') return '';
  return email.trim().toLowerCase();
};
```

opencti-platform/opencti-graphql/src/domain/user.js
```diff
@@ -878,13 +879,19 @@ export const addUser = async (context, user, newUser) => {
...
     }
   } else if (userServiceAccount) {
-    userEmail = newUser.user_email ? newUser.user_email : `automatic+${uuid()}@opencti.invalid`;
+    userEmail = newUser.user_email ? normalizeEmail(newUser.user_email) : `automatic+${uuid()}@opencti.invalid`;
+    if (newUser.user_email) {
+      const existingUser = await elLoadBy(context, SYSTEM_USER, 'user_email', userEmail, ENTITY_TYPE_USER);
+      if (existingUser) {
+        throw FunctionalError('User already exists', { user_id: existingUser.internal_id });
+      }
+    }
```
ludovic: This code could be refactored and improved using guard clauses; what do you think?

```ts
  const userServiceAccount = newUser.user_service_account;
  if (!newUser.user_email && !userServiceAccount) {
    throw FunctionalError('User cannot be created without email');
  }
  const userEmail = newUser.user_email ? normalizeEmail(newUser.user_email) : `automatic+${uuid()}@opencti.invalid`;
  if (isEmptyField(userEmail)) {
    throw FunctionalError('The email you have provided is not valid');
  }
  if (newUser.user_email) {
    const existingUser = await elLoadBy(context, SYSTEM_USER, 'user_email', userEmail, ENTITY_TYPE_USER);
    if (existingUser) {
      throw FunctionalError('User already exists', { user_id: existingUser.internal_id });
    }
  }
```

marieflorescontact: yes I know but because of backport needed I prefered avoid refactoring code . do you think I should refactor any way ?

opencti-platform/opencti-graphql/src/utils/email.ts
marieflorescontact: fixed

opencti-platform/opencti-graphql/src/domain/user.js
ludovic: I understand, but I think the refactoring is localized and shouldn't have any more impact than the code you added. Plus, adding guards could be a benefit.

marieflorescontact: done

SarahBocognano (approved):

PR #18091, opened 2026-09-03
fix(filters): Bulk action filters for Security coverage results (#18090)

### LINKED ISSUE #18090: fix(scope): select all in security coverage result tab selects everything in the platform
## Description

select all in security coverage result tab selects everything in the platform

## Environment

1. Version: 7.260901.0

3. Other environment details:

## Reproducible steps

1. Open security coverage adn go to the result tab
2. select all and perform a bulk action e.g. share wtih organisation
3. observe that there is no query set for the background task so that all data is to be processed. 

## Expected output

only selected entities are processed

## Actual output

Task created processing all the data in the platform

## Additional information

Image

Archidoit:
### Proposed changes
In a Security Coverage overview, in the 'Result' tab, when selecting all the list (via the main checkbox), only the security coverage results should have the bulk action done on them, not the whole platform.

image

### Related issues
fixes #18090

### DIFF
--- a/opencti-platform/opencti-front/src/private/components/analyses/security_coverages/SecurityCoverageResult.tsx
+++ b/opencti-platform/opencti-front/src/private/components/analyses/security_coverages/SecurityCoverageResult.tsx
@@ -8,7 +8,7 @@ import {
   SecurityCoverageResultLinesPaginationQuery$variables,
 } from '@components/analyses/security_coverages/__generated__/SecurityCoverageResultLinesPaginationQuery.graphql';
 import { usePaginationLocalStorage } from '../../../../utils/hooks/useLocalStorage';
-import { emptyFilterGroup } from '../../../../utils/filters/filtersUtils';
+import { emptyFilterGroup, useBuildEntityTypeBasedFilterContext } from '../../../../utils/filters/filtersUtils';
 import { SecurityCoverageResultLines_data$data } from '@components/analyses/security_coverages/__generated__/SecurityCoverageResultLines_data.graphql';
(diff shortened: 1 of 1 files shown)


opencti-platform/opencti-front/src/private/components/analyses/security_coverages/SecurityCoverageResult.tsx
```diff
@@ -268,15 +269,7 @@ const SecurityCoverageResultComponent = ({ id }: SecurityCoverageResultProps) =>
...
-  const queryPaginationOptions = {
-    ...paginationOptions,
-  } as unknown as SecurityCoverageResultLinesPaginationQuery$variables;
-
-  const queryRef = useQueryLoading<SecurityCoverageResultLinesPaginationQuery>(
-    securityCoverageResultLinesQuery,
-    queryPaginationOptions,
-  );
-
+  const userFilters = useBuildEntityTypeBasedFilterContext('stix-core-relationship', filters);
```
JeremyCloarec, line 272: I was going to say that it should maybe be the specific 'has-covered' relationship instead of all abstract 'stix-core-relationship' type, but it looks like the view already displays all relationships and not only 'has-covered' ones. So the context filter correctly aligns with the view :) Not sure if it was intended or if it's an oversight on the view though

JeremyCloarec (approved):

opencti-platform/opencti-front/src/private/components/analyses/security_coverages/SecurityCoverageResult.tsx
Archidoit: Indeed ! I keep the behavior as it is, ie all the relationships involving the current Security coverage

PR #18103, opened 2026-09-04
feat(workflow): add workflowMigrationPreview endpoint (#17550)

### LINKED ISSUE #17550: feat(workflow): Create migration endpoint to migrate workflow definition
## Use case

Creation of an endpoint to migrate current statuses in entity customization to be able to use Worfklow definition.

frapuks:
### Proposed changes

* Add an endpoint to preview the conversion from old workflow to new workflow for an entity type.

### Related issues

* closes #17550 

### How to test this PR

- Open the graphQL playground
- use a query to trigger the new endpoint

### Checklist


opencti-platform/opencti-graphql/src/modules/workflow/domain/workflow-domain.ts
```diff
@@ -413,6 +414,30 @@ export const getWorkflowPublishedVersionId = async (
...
+  const executionContext = bypassDraftContext(context);
+  const executionUser = bypassDraftUser(user);
+  const statuses = await fullEntitiesList<BasicWorkflowStatus>(executionContext, executionUser, [ENTITY_TYPE_STATUS], {
+    filters: {
+      mode: FilterMode.And,
+      filters: [{ key: ['type'], values: [entityType] }],
+      filterGroups: [],
+    },
+  });
+  const templates = await fullEntitiesList<BasicWorkflowTemplateEntity>(executionContext, executionUser, [ENTITY_TYPE_STATUS_TEMPLATE]);
```
delemaf, line 437: This does an unscoped `fullEntitiesList` over **all** `StatusTemplate` entities on every preview request. The needed template ids are already known from `statuses` (`status.template_id`), so this could be scoped the same way other call sites in this module do it (e.g. `storeLoadByIds` on the specific ids) instead of a full-table scan.

opencti-platform/opencti-graphql/src/modules/workflow/migration/status-to-definition-converter.ts
```diff
@@ -0,0 +1,182 @@
...
+ */
+const synthesizeFullyConnectedTransitions = (
+  states: WorkflowMigrationState[],
+  templatesById: Map<string, BasicWorkflowTemplateEntity>,
+): WorkflowSerializedTransition[] => {
+  const stateIds = states.map((state) => state.statusId);
+  if (stateIds.length < 2) return [];
+  return stateIds.map((to) => {
+    const rawName = templatesById.get(to)?.name;
+    const name = rawName && rawName.trim().length > 0 ? rawName : to;
```
delemaf, line 63: Blank/whitespace-only template names are treated as "blank" here (falls back to `to`), but the name-conflict check below (~L117/L121) uses `?? status.template_id` directly without trimming — a whitespace-only name is treated as a real, distinct name there. Worth extracting a shared `resolveDisplayName` helper so both call sites agree on what counts as "blank".

Separately: when `templatesById.get(to)` is `undefined` (an orphaned/broken `template_id` reference), this silently falls back to the id with no diagnostic — given this preview's job is to surface migration issues, a broken template reference seems like exactly the kind of thing that should produce a diagnostic entry (like `MISSING_ORDER`/`NAME_CONFLICT` do), not a silent fallback.

opencti-platform/opencti-graphql/src/modules/workflow/migration/status-to-definition-converter.ts
```diff
@@ -0,0 +1,182 @@
...
+
+  // Name-conflict diagnostic: two distinct states (different template_id) resolving to the same
+  // display name would be indistinguishable to a human reviewing the migrated definition.
+  const nameCounts = new Map<string, number>();
+  ordered.forEach((status) => {
+    const name = templatesById.get(status.template_id)?.name ?? status.template_id;
+    nameCounts.set(name, (nameCounts.get(name) ?? 0) + 1);
+  });
+  ordered.forEach((status) => {
+    const name = templatesById.get(status.template_id)?.name ?? status.template_id;
```
delemaf, line 121: Same silent-fallback pattern as the transition-name resolution at L63, but without the blank-check: `?? status.template_id` here doesn't treat a whitespace-only name as blank. See the note on L63 — consider sharing one `resolveDisplayName` helper between both spots so the two diagnostics agree on behavior.

opencti-platform/opencti-graphql/src/modules/workflow/migration/status-to-definition-converter.ts
```diff
@@ -0,0 +1,182 @@
+import { StatusScope } from '../../../generated/graphql';
+import type { BasicWorkflowStatus, BasicWorkflowTemplateEntity } from '../../../types/store';
+import type { WorkflowSerializedTransition } from '../types/workflow-types';
+
+export interface WorkflowMigrationDiagnostic {
+  type: string;
```
delemaf, line 6: `type` is a raw `string` here (and in the GraphQL schema's `WorkflowMigrationDiagnostic.type: String!`), but only two literal values are ever produced (`'MISSING_ORDER'`, `'NAME_CONFLICT'`). A small union type / GraphQL enum would give compile-time exhaustiveness instead of loose string literals, and make it harder to typo a diagnostic type.

opencti-platform/opencti-graphql/src/modules/workflow/migration/status-to-definition-converter.ts
```diff
@@ -0,0 +1,182 @@
...
+    if (!seenTemplateIds.has(status.template_id)) {
+      seenTemplateIds.add(status.template_id);
+      uniqueStatuses.push(status);
+    }
+  });
+
+  uniqueStatuses.forEach((status) => {
+    if (!hasOrder(status)) {
+      diagnostics.push({
+        type: 'MISSING_ORDER',
```
delemaf, line 93: Only *missing* `order` produces a diagnostic — duplicate `order` values across statuses in the same scope group (which would make the resulting sort non-deterministic-looking to a reviewer) go unflagged. Might be worth a `DUPLICATE_ORDER` diagnostic alongside `MISSING_ORDER` for completeness.

opencti-platform/opencti-graphql/src/modules/workflow/api/workflow-resolvers.ts
```diff
@@ -28,6 +29,23 @@ const workflowResolvers = {
     workflowInstance: (_: any, { entityId }: { entityId: string }, context: AuthContext) => {
       return getWorkflowInstance(context, context.user!, entityId);
     },
+    workflowMigrationPreview: async (_: any, { entityType }: { entityType: string }, context: AuthContext) => {
```
delemaf, line 32: This resolver does non-trivial shaping (mapping `byScope` entries, hardcoding `published`/`hasPublishedVersion`) rather than delegating to the domain layer — the module-architecture doc describes resolvers as a thin bridge to domain logic. Moving this shaping into `getWorkflowMigrationPreview` (or a small domain-level helper) would also make it unit-testable without needing to spin up the resolver, and avoid the `result!` non-null assertions below that work around `Object.entries`' loose typing on the `Partial<Record<...>>` map.

delemaf (changes requested):

PR #18143, opened 2026-09-06
feat(fintel): column selection for entity attributes (#15604)

### LINKED ISSUE #15604: feat(fintel): column selection for entity attributes
## Use case

When building Fintel templates to generate intelligence reports (newsletters, 
executive summaries, case management exports), users insert table widgets that 
display entities contained in a container (e.g. Intrusion Sets, Malware, Tasks).

Currently, the columns displayed in those tables are fixed — users cannot choose 
which attributes to show. This forces them to accept a generic output or manually 
post-process the exported PDF, which defeats the purpose of automated report generation.

When tables exceed 8 columns the fintel report will switch to export in landscape.

Gwendoline-FAVRE-FELIX:
### Proposed changes

*  Added column selection for Fintel table/list widgets (add, remove, reorder).
* Reused existing attribute/column logic to list available attributes based on the selected entity type.
* Saved column configuration per widget, while keeping backward compatibility with existing templates.
* Applied selected columns in generated outputs (HTML/PDF) with fallback to default columns when none are configured.
* Improved PDF export for wide tables (compact/ultra-compact layout, adaptive orientation/page size) to prevent column clipping.
* Restored template UI behavior (horizontal scrolling kept, column header readability preserved).

### Related issues

* closes https://github.com/OpenCTI-Platform/opencti/issues/15604?reload=1#event-29562997372

### How to test this PR

* Go in Settings > Customization > Report > Templates
* Create a template and change which data to retrieve in this widget.
* In step 4, customize columns.
* Use it in your template and save it.
* Go in Analyses > Reports > Contente
* Generate an export based on a template
* Generate a PDF export

### Checklist


opencti-platform/opencti-front/src/private/components/widgets/WidgetCreationParameters.tsx
```diff
@@ -346,6 +346,24 @@ const WidgetCreationParameters = () => {
     );
   };
 
+  const getEntityTypeFromFilters = (filterGroup?: FilterGroup | null): string | undefined => {
```
Archidoit, line 349: this function from line 354 to line 362 is useless since it's already done in getEntityTypeThreeFirstLevelsFilterValues... You can directly use getEntityTypeThreeFirstLevelsFilterValues instead of your function getEntityTypeFromFilters

Archidoit (dismissed):

Archidoit:
There is a difference of layout between coumns customization in widgets of dashboards : 
image


and between the ones of fintel templates: 
image

--> The 'Selected columns' part should be half the total width, ie same width as the 'Available column' part

Archidoit:
Default widget edition: columns are not compatible with effectively displayed columns.

Steps to reproduce:
In the defaut fintel template 'Executive Summary' (of a Report for instance), open the edition form for the '$indicators' list widget 
image


In the 'parameters' tab of its edition form, I see no column selected : 
image

But when I generate a fintel from a report, using the 'executive summary' template, there are 2 columns : 
image


--> The columns displayed for widgets in Executive summary should be correctly displayed in their edition form.

Archidoit:
In a fintel template widget edition form, select the 'entities' perspective, than set no filters
image


In the 'available columns', we have the attributes of the entities for which we create a fintel template (reports here). 
image


That is no correct. Since there are no filters, the widget list is applied on the whole platform, not only reports.
If no filters, no entity type is selected, thus we should have the available columns by default for every stix core objects.

opencti-platform/opencti-front/src/private/components/widgets/WidgetCreationParameters.tsx
```diff
@@ -943,43 +961,45 @@ const WidgetCreationParameters = () => {
...
-              );
-            }
+        {type === 'list' && dataSelection.map(({ perspective, columns, filters }, index) => {
+          if (perspective !== 'relationships' && perspective !== 'entities') {
             return null;
-          })}
+          }
+
+          const entityTypeFromFilters = getEntityTypeFromFilters(filters);
+          const entityType = entityTypeFromFilters
```
Archidoit, line 970: setting the available columns type to the type of the host in case there are no filters leads to this problem (see my comment) : https://github.com/OpenCTI-Platform/opencti/pull/18143#issuecomment-5567695448

opencti-platform/opencti-front/src/private/components/widgets/WidgetCreationParameters.tsx
```diff
@@ -961,43 +971,48 @@ const WidgetCreationParameters = () => {
...
-                <WidgetColumnsCustomizationInput
-                  key={index}
-                  availableColumns={getWidgetColumns(perspective, entityType || undefined, metricsDefinition || undefined)}
-                  defaultColumns={defaultWidgetColumnsByType}
-                  value={[...(columns ?? defaultWidgetColumnsByType)]}
-                  onChange={(newColumns) => setColumns(index, newColumns)}
-                />
-              );
-            }
+        {type === 'list' && dataSelection.map(({ perspective, columns, filters }, index) => {
```
esrevi, line 974: there are many changes here, are you sure the current behaviour is still correct? 

otherwise, maybe we could have less changes? there is no tests on that part that can guarantee us we don't introduce regression

opencti-platform/opencti-front/src/private/components/widgets/WidgetCustomAttributesColumnsInput.tsx
```diff
@@ -88,10 +94,11 @@ const SingleColumnLayout: FunctionComponent<ColumnLayoutProps> = ({
   formatColumnName,
   t_i18n,
   listSx,
+  selectedLabel,
 }) => (
-  <Box sx={{ flex: 2 }}>
+  <Box sx={{ flex: 1 }}>
     <Typography variant="h4">
-      {`${t_i18n('Selected attributes')} (${value.length})`}
+      {`${t_i18n(selectedLabel)} (${value.length})`}
```
esrevi, line 101: are the selectedLabel values translated this way ?

opencti-platform/opencti-front/src/utils/htmlToPdf/utils/pdfTableWidth.ts
```diff
@@ -10,19 +21,36 @@ const setTableFullWidth = (content: string) => {
...
-      const nbColumns = element.querySelectorAll(header ? 'th' : 'td').length;
-      if (nbColumns) {
-        table.setAttribute('data-pdfmake', `{'layout':'default', 'widths':[${Array(nbColumns).fill("'*'").join()}]}`);
-      }
+    const nbColumns = getTableColumnCount(table);
+    if (nbColumns) {
+      const isWideTable = nbColumns >= WIDE_TABLE_COLUMN_THRESHOLD;
+      const isVeryWideTable = nbColumns >= VERY_WIDE_TABLE_COLUMN_THRESHOLD;
+      const layout = isVeryWideTable ? 'ultraCompact' : isWideTable ? 'compact' : 'default';
+      const fontSize = isVeryWideTable ? 8 : isWideTable ? 9 : undefined;
```
esrevi, line 29: if possible, avoid nested ternary operations :)

opencti-platform/opencti-front/src/utils/htmlToPdf/utils/pdfTableWidth.ts
```diff
@@ -34,6 +62,26 @@ export const defaultTableLayout: { [p: string]: CustomTableLayout } = {
     hLineWidth: () => 1,
     vLineWidth: (i, { table }) => ((i === 0 || i === (table.widths ?? []).length) ? 1 : 0),
   },
+  compact: {
+    hLineColor: '#dcdde4',
```
esrevi, line 66: you can create a common options all except the properties that are specific to a size i think

opencti-platform/opencti-front/src/utils/outcome_template/engine/stix_core_objects/useBuildListOutcome.tsx
```diff
@@ -13,6 +13,45 @@ import { RELATIONSHIP_WIDGETS_TYPES } from '../../../widget/widgetUtils';
...
 type ListItem = object & { id: string };
 
+const resolveWorkflowPath = (attribute?: string | null) => {
+  if (attribute === 'x_opencti_workflow_id') {
+    return 'status.template.name';
+  }
+  return attribute ?? '';
+};
+
+const hasDisplayableValue = (value: unknown) => {
```
esrevi, line 23: can we avoid unknown type here ?

opencti-platform/opencti-front/src/utils/outcome_template/engine/stix_core_objects/useBuildListOutcome.tsx
```diff
@@ -13,6 +13,45 @@ import { RELATIONSHIP_WIDGETS_TYPES } from '../../../widget/widgetUtils';
...
+  if (Array.isArray(value)) return value.length > 0;
+  return value !== '' && value !== null && value !== undefined;
+};
+
+const normalizeObjectForDisplay = (value: unknown): unknown => {
```
esrevi, line 28: same question here

opencti-platform/opencti-front/src/private/components/widgets/WidgetCustomAttributesColumnsInput.tsx
```diff
@@ -88,10 +96,12 @@ const SingleColumnLayout: FunctionComponent<ColumnLayoutProps> = ({
   formatColumnName,
   t_i18n,
   listSx,
+  selectedLabel,
+  selectedPaneFlex,
 }) => (
-  <Box sx={{ flex: 2 }}>
+  <Box sx={{ flex: selectedPaneFlex }}>
```
SarahBocognano: just a small typo error here: selectedPanelFlex

opencti-platform/opencti-front/src/private/components/widgets/WidgetCreationParameters.tsx
```diff
@@ -961,43 +961,51 @@ const WidgetCreationParameters = () => {
...
+            return (
+              <WidgetCustomAttributesColumnsInput
+                key={index}
+                availableColumns={availableColumns}
+                defaultColumns={defaultWidgetColumnsByType}
+                value={selectedColumns}
+                onChange={(newColumns) => setColumns(index, newColumns)}
+                selectedPanelFlex={1}
+                landscapeWarningThreshold={8}
+                landscapeWarningMessage={t_i18n('More than 8 columns selected — when applied the export will switch to landscape format.')}
```
esrevi: if this value landscapeWarningThreshold changes, the warning message more than 8 columns will be incorrect, can we use a constant here to contain that number ?

esrevi: 'More than 8 columns' is still not using the constant you added :p

esrevi (approved):
lgtm !

PR #18147, opened 2026-09-07
fix(backend): preserve standard id when building a restricted entity (#18026)

### LINKED ISSUE #18026: fix(pycti): Invalid STIX bundle produced by pycti get_stix_bundle_or_object_from_entity_id when entity creator organization is TLP-restricted for the requesting user
## Description

When exporting an entity as a STIX bundle via the pycti method get_stix_bundle_or_object_from_entity_id, if the entity's created_by_ref (the author organization/identity) is TLP-marked and the requesting user does not have access to that TLP, the resulting bundle is invalid STIX 2.1: the redacted author identity object has id: "Restricted" (not a valid STIX identifier) instead of a properly formatted identity--<uuid> or None.

This invalid bundle subsequently causes an exception on the import-document connector side, which expects a strictly valid STIX bundle and fails to process it. (https://github.com/OpenCTI-Platform/connectors/issues/7393)

## Reproducible steps

1. Create an Organization with a TLP marking (e.g. TLP:TEST / restricted marking).
2. Create a Sector (or any entity) with that Organization set as its author (created_by_ref).
3. Create a user who has access to the Sector but not to the Organization (TLP restriction on the author).
4. As that user, export the Sector entity via pycti: get_stix_bundle_or_object_from_entity_id.

## Expected output

Valid STIX bundle. 

## Actual output

```
{
  "id": "bundle--f067e003-8d22-4607-8578-db0f076157f5",
  "type": "bundle",
  "objects": [
    {
      "id": "Restricted",
      "type": "identity",
      "spec_version": "2.1",
      "created": "1970-01-01T00:00:00.000Z",
      "modified": "1970-01-01T00:00:00.000Z",
      "name": "Restricted",
      "description": "Restricted",
      "identity_class": "Restricted",
      "x_opencti_id": "79d14387-751e-4608-9807-2e8ed9c8711b",
      "x_opencti_organization_type": "Restricted",
      "x_opencti_type": "Organization"
    },
    {
      "id": "identity--d4551de9-4b9c-570e-a51c-d3c321eb9a8d",
      "type": "identity",
      "spec_version": "2.1",
      "created": "2026-08-25T07:44:55.797Z",
      "modified": "2026-08-25T07:44:55.833Z",
      "created_by_ref": "Restricted",
      "name": "Nutrition",
      "identity_class": "class",
      "confidence": 100,
      "revoked": false,
      "x_opencti_id": "e8566c5f-e0c5-4b99-921b-a818e5856ffc",
      "x_opencti_type": "Sector"
    }
  ]
}
```

gabrielebonadiman:
### Proposed changes

* `buildRestrictedEntity` in `opencti-platform/opencti-graphql/src/database/middleware.ts` overwrote every string attribute of a restricted entity with the literal `"Restricted"`, including `standard_id`. This produced an invalid STIX identifier when the restricted entity was later exported (e.g. as `created_by_ref` in a STIX bundle).
* Fix: `standard_id` is now preserved on the restricted entity, alongside the already-preserved `id`, `entity_type`, and `parent_types`. All other attributes (`name`, `description`, `representative`, etc.) still resolve to `"Restricted"`.

### Related issues
* closes #18026

### How to test this PR

`cd opencti-platform/opencti-graphql && npx vitest --config vitest.config.ci-unit.ts run tests/01-unit/database/middleware-restricted-entity-unit-test.ts tests/01-unit/database/middleware-distribution-unit-test.ts tests/01-unit/domain/stixCoreObject-test.ts`

This runs:
* `middleware-restricted-entity-unit-test.ts` — direct unit test of `buildRestrictedEntity`, asserting `standard_id` stays a valid `identity--<uuid>`.
* `middleware-distribution-unit-test.ts` — added case for `convertAggregateDistributions`'s access-denied branch.
* `stixCoreObject-test.ts` — added cases for `batchInternalRels` (the GraphQL `createdBy` resolver), covering single-ref and mixed multi-ref access-denied scenarios.

All three fail against the pre-fix code (reproducing the issue) and pass after the fix.

### Checklist

### Further comments

Full unit suite (`vitest.config.ci-unit.ts`) run before/after: no regressions (2387/2389 tests pass, 2 skipped; 35 file-level failures pre-exist on the base branch, unrelated — caused by a missing `sanitize-html` package).

### DIFF
--- a/opencti-platform/opencti-graphql/src/database/middleware.ts
+++ b/opencti-platform/opencti-graphql/src/database/middleware.ts
@@ -754,6 +754,7 @@ export const buildRestrictedEntity = (resolvedEntity: BasicStoreEntity): BasicSt
   return {
     ...restrictedEntity,
     id: resolvedEntity.internal_id,
+    standard_id: resolvedEntity.standard_id, // keep the real standard_id: it must stay a valid STIX id
     name: 'Restricted',
     entity_type: resolvedEntity.entity_type,
     parent_types: resolvedEntity.parent_types,
--- a/client-python/pycti/utils/opencti_stix2.py
+++ b/client-python/pycti/utils/opencti_stix2.py
@@ -2143,6 +2143,26 @@ class OpenCTIStix2:
                 ],
             }
 
+    @staticmethod
+    def _is_restricted_identity(identity: Dict) -> bool:
+        """Detect a createdBy identity obfuscated by the platform because the
+        user has no access to it (see buildRestrictedEntity on the backend).
+
+        Both name and identity_class must match: identity_class is a
+        platform-controlled field and is never legitimately set to the
+        literal string "Restricted", so requiring both avoids false
+        positives on a real entity that happens to be named "Restricted".
+
+        :param identity: createdBy entity as returned by the API
+        :type identity: Dict
+        :return: True if the identity is a restricted placeholder
+        :rtype: bool
+        """
+        return (
+            identity.get("name") == "Restricted"
+            and identity.get("identity_class") == "Restricted"
+        )
+
     def prepare_export(
         self,
         entity: Dict,
@@ -2172,6 +2192,7 @@ class OpenCTIStix2:
             not no_custom_attributes
             and "createdBy" in entity
             and entity["createdBy"] is not None
+            and not self._is_restricted_identity(entity["createdBy"])
         ):
             created_by = self.generate_export(entity=entity["createdBy"])
             if entity["type"] in STIX_CYBER_OBSERVABLE_MAPPING:
--- a/client-python/pycti/utils/opencti_stix2.py
+++ b/client-python/pycti/utils/opencti_stix2.py
@@ -2145,19 +2145,7 @@ class OpenCTIStix2:
 
     @staticmethod
     def _is_restricted_identity(identity: Dict) -> bool:
-        """Detect a createdBy identity obfuscated by the platform because the
-        user has no access to it (see buildRestrictedEntity on the backend).
-
-        Both name and identity_class must match: identity_class is a
-        platform-controlled field and is never legitimately set to the
-        literal string "Restricted", so requiring both avoids false
-        positives on a real entity that happens to be named "Restricted".
-
-        :param identity: createdBy entity as returned by the API
-        :type identity: Dict
-        :return: True if the identity is a restricted placeholder
-        :rtype: bool
-        """
+        """Detect the platform's restricted-access placeholder identity."""
         return (
             identity.get("name") == "Restricted"
             and identity.get("identity_class") == "Restricted"
--- a/client-python/tests/01-unit/utils/test_opencti_stix2.py
+++ b/client-python/tests/01-unit/utils/test_opencti_stix2.py
@@ -2,12 +2,26 @@ import datetime
 
 import pytest
 
+from pycti import OpenCTIApiClient
 from pycti.utils.opencti_stix2 import OpenCTIStix2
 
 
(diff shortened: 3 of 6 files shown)


JeremyCloarec:
I think the solution needs to be checked with product. With this solution, we'll still get a 'Restricted' individual added to the bundle, which in my opinion we do not want. 
In my opinion, we want to filter out Restricted entities from exports entirely, i.e. remove the created_by_ref from the entity in the bundle if the created_by_ref is linked to a 'Restricted' entity. This also looks to be the current solution implemented for a workaround PR here https://github.com/OpenCTI-Platform/connectors/pull/7479

fellowseb:
I tested the scenario shared by @romain-filigran. With this fix we still export the Organization entity but with some values still marked as "Restricted". Indeed the `created_by_ref` is fixed; but I believe what we really wanted is to drop the `created_by_ref` property and the restricted Organization entitye entirely instead.

EDIT: I just saw your comment Jeremy and I believe we're aligned.

EDIT2: the specs in the issue are not super clear on that point `the redacted author identity object has id: "Restricted" (not a valid STIX identifier) instead of a properly formatted identity-- or None.`; but I think in that case that the value `Restricted` for the `identity_class` is not valid for instance. The possible values are listed here: https://github.com/oasis-open/cti-stix2-json-schemas/blob/master/schemas/sdos/identity.json#L63. (`unknown` is one of them).
Anyways I still think it does make more sense to drop `created_by_ref` (It's not a mandatory attribute https://github.com/oasis-open/cti-stix2-json-schemas/blob/master/schemas/common/core.json) and the referenced restricted entity.

gabrielebonadiman:
@JeremyCloarec @fellowseb Thanks for the feedback, I've implemented the suggested approach: dropping  `created_by_ref`  and the restricted identity object entirely from the export, rather than exporting an obfuscated placeholder.

Change is in  `prepare_export()`  ( client-python/pycti/utils/opencti_stix2.py ):

```
entity["createdBy"] present?
  │
  ├─ NO  → nothing added
  │
  └─ YES → _is_restricted_identity(createdBy)?
             (name == "Restricted" AND identity_class == "Restricted")
             │
             ├─ TRUE  → skip entirely: no created_by_ref on the entity,
             │           identity object not appended to the bundle
             │
             └─ FALSE → unchanged: created_by_ref = identity["id"],
                         identity object appended to the bundle
```

The check requires both  name  and  identity_class  to match, so a real identity legitimately named "Restricted" isn't mistaken for the placeholder. Added tests covering the restricted case, the normal case, and that edge case.


Note for visibility:  `_is_restricted_identity` matches on the literal  `"Restricted"`  strings from  buildRestrictedEntity. no real contract between the two repos, so if that placeholder text ever changes, this silently stops working. Same pattern as the existing connectors workaround (#7479), not proposing to change it now, just flagging it.

opencti-platform/opencti-graphql/src/database/middleware.ts
```diff
@@ -754,6 +754,7 @@ export const buildRestrictedEntity = (resolvedEntity: BasicStoreEntity): BasicSt
   return {
     ...restrictedEntity,
     id: resolvedEntity.internal_id,
+    standard_id: resolvedEntity.standard_id, // keep the real standard_id: it must stay a valid STIX id
```
JeremyCloarec: we want to revert this change: now that pycti handles restricted entities there, we need to remove standard_id display here to prevent leaking information

gabrielebonadiman: 👍

JeremyCloarec (approved):

PR #18149, opened 2026-09-07
feat(vulnerability): add SSVC attributes in UI (#17566)

### LINKED ISSUE #17566: Chunk 2: Add in UI SSVC (in vulnerability overview & edition/creation forms)
### TODO

marieflorescontact:
### Proposed changes

* add SSVC tab in details view
* add SSVC tab in update form
* add SSVC attributes in creation form

### Related issues

* closes #https://github.com/OpenCTI-Platform/opencti/issues/17566

### Checklist


opencti-platform/opencti-front/src/private/components/arsenal/vulnerabilities/VulnerabilityDetails.tsx
```diff
@@ -505,6 +511,37 @@ const VulnerabilityDetailsComponent = ({ vulnerability }: VulnerabilityDetailsPr
...
             </Grid>
           )}
+          {value === 3 && isSsvcEnabled && (
+            <Grid container spacing={2} style={{ margin: 10, alignSelf: 'flex-start' }}>
+              <Grid item xs={6}>
+                <Label>
+                  {t_i18n('SSVC Exploitation')}
+                </Label>
+                <FieldOrEmpty source={vulnerability.x_opencti_ssvc_exploitation}>
+                  <Tag label={vulnerability.x_opencti_ssvc_exploitation}>
```
ludovic, line 521: Don't we want to translate these labels?

opencti-platform/opencti-front/src/private/components/arsenal/vulnerabilities/VulnerabilityEditionContainer.jsx
```diff
@@ -72,6 +79,14 @@ const VulnerabilityEditionContainer = (props) => {
             handleClose={handleClose}
           />
         )}
+        {currentTab === 4 && isSsvcEnabled && (
+          <VulnerabilityEditionSsvc
+            vulnerabilityRef={vulnerability}
+            enableReferences={useIsEnforceReference('Vulnerability')}
```
ludovic: I know you replicated the existing behavior, but it's bad practice. It is preferable to call hooks at the beginning and unconditionally.

`const enableReferences = useIsEnforceReference('Vulnerability');`

https://react.dev/reference/rules/rules-of-hooks#only-call-hooks-at-the-top-level

opencti-platform/opencti-front/src/private/components/arsenal/vulnerabilities/VulnerabilityEditionContainer.jsx
```diff
@@ -38,6 +42,9 @@ const VulnerabilityEditionContainer = (props) => {
             <Tab label={t_i18n('CVSS V2')} style={{ textTransform: 'none' }} />
             <Tab label={t_i18n('CVSS V3')} style={{ textTransform: 'none' }} />
             <Tab label={t_i18n('CVSS V4')} style={{ textTransform: 'none' }} />
+            {isSsvcEnabled && (
+              <Tab label={t_i18n('SSVC')} style={{ textTransform: 'none', minWidth: 'auto' }} />
```
ludovic: Quick question regarding the `minWidth: 'auto'` on the SSVC tab: since only this tab has it, it creates a slight asymmetry with the other tabs (which still use MUI's default `min-width`). 
If the goal is to make the spacing feel more balanced or prevent drawer overflow, maybe we could apply it consistently to all tabs via the parent `<Tabs>` component instead:
```jsx
<Tabs
  value={currentTab}
  onChange={handleChangeTab}
  sx={{ '& .MuiTab-root': { minWidth: 'auto', px: 2 } }}
>
```
(or use variant="scrollable" scrollButtons="auto" if drawer width is tight).

If we do that, we can also remove the unit test asserting toHaveStyle({ minWidth: 'auto' }), as it tests an inline style detail rather than component behavior.

What do you think?

opencti-platform/opencti-front/src/private/components/arsenal/vulnerabilities/VulnerabilityEditionContainer.test.tsx
```diff
@@ -0,0 +1,115 @@
...
+      />,
+    );
+
+    await user.click(screen.getByText('SSVC'));
+
+    expect(screen.getByTestId('tab-ssvc')).toBeInTheDocument();
+    expect(screen.queryByTestId('tab-overview')).not.toBeInTheDocument();
+  });
+
+  it('applies minWidth: auto to the SSVC tab so its spacing matches the other tabs', () => {
```
ludovic: Nit: We can probably safely drop this test!
Asserting directly on inline CSS (`toHaveStyle`) checks an internal styling detail rather than actual user-facing behavior, which might make the test brittle if we ever tweak the drawer or tab layout later. What do you think?

opencti-platform/opencti-front/src/private/components/arsenal/vulnerabilities/VulnerabilityDetails.tsx
marieflorescontact: The label is indeed translated in the `<Label> `component above. The `<Tag>` label is actually the raw field value returned by the API. It is never translated anywhere in the file .I followed the existing pattern.

opencti-platform/opencti-front/src/private/components/arsenal/vulnerabilities/VulnerabilityEditionContainer.jsx
marieflorescontact: fixed

opencti-platform/opencti-front/src/private/components/arsenal/vulnerabilities/VulnerabilityEditionContainer.test.tsx
marieflorescontact: ahah 😅

marieflorescontact: fixed

opencti-platform/opencti-front/src/private/components/arsenal/vulnerabilities/VulnerabilityEditionContainer.jsx
marieflorescontact: minWidth: 'auto' was applied only to the SSVC tab to compensate for its shorter label, which otherwise left a larger visual gap than between the other tabs (whose labels already fill MUI's default 90px width). but your proposition works well. thanks

ludovic (approved):

PR #18198, opened 2026-09-09
fix(knowledge-view): table and bar counters don't refresh on add/delete element in knowledge tab (#9838)

### LINKED ISSUE #9838: fix(knowledge-view): table and bar counters don't refresh on add/delete element in knowledge tab
## Environment

1. OS (where OpenCTI server runs): N/A
2. OpenCTI version: All
3. OpenCTI client: frontend 
4. Other environment details:

## Reproducible Steps

1. Go in any Knowledge tab that contains the right bar with entities & observables (ex: Threat Actors Group)
2. Go to a section (ex: Malwares)
3. Add an element

## Expected Output

The table counter should increment and also the counter in the bar

## Actual Output

The table counter doesn't increment, same for the counter in the bar

## Screenshots (optional)
https://github.com/user-attachments/assets/30b09858-ea09-4376-9e7d-a56410806389

delemaf:
### Proposed changes

- Refresh the `StixCoreObjectKnowledgeBar` distribution counters (Malwares, Threats, etc.) whenever a relationship is added or deleted from a knowledge tab, via a dedicated `KnowledgeBarRefreshEvent` window event and an imperative refetch through `Query.stixCoreObject(id)`.
- Dispatch `KnowledgeBarRefreshEvent` from the two relationship creation flows (`StixCoreRelationshipCreationFromEntity` and `StixCoreRelationshipCreationFormStage`) and from the relationship deletion flow (`StixCoreRelationshipPopover`).
- Kept this new event separate from the existing `ForceUpdateEvent` (used by bulk relation actions to remount the table via `key={forceUpdate}`) so refreshing the bar counters no longer causes the entities table to flash empty on every single add/delete.

### Related issues

- closes #9838

### How to test this PR

- Open any entity's Knowledge tab that has the right-side bar with entity/observable counts (e.g. a Threat Actor Group).
- Go to a section (e.g. Malwares) and add a relationship to an entity of that type, either via the "Create Relationship" header button or the classic add dialog.
- Confirm the table counter and the corresponding bar counter both increment immediately, without the table flashing empty.
- Delete the relationship and confirm both counters decrement immediately.

### Checklist


opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectKnowledgeBar.tsx
```diff
@@ -93,6 +105,7 @@ interface StixCoreObjectKnowledgeBarProps {
   availableSections: string[];
   data: StixCoreObjectKnowledgeBar_stixCoreObject$key;
   attribution?: string[];
+  relatedRelationshipTypes?: string[];
```
OctaveLaventure: This new `relatedRelationshipTypes` prop is only wired up as a *fallback default* for the refresh query below — it isn't passed by every caller that queries the fragment with a non-default `relatedRelationshipTypes` argument.

Concretely: `threat_actors_group/Root.tsx` and `threat_actors_individual/Root.tsx` both spread `...StixCoreObjectKnowledgeBar_stixCoreObject @arguments(relatedRelationshipTypes: $relatedRelationshipTypes)` with `$relatedRelationshipTypes = ['related-to', 'part-of', 'impersonates', 'known-as']`, but their `<StixCoreObjectKnowledgeBar>` element doesn't forward that same list as a prop (it's only forwarded to `ThreatActor*Knowledge`). So after this PR, adding/deleting a relationship on those two entity types will trigger a refresh that falls back to `['related-to']` only, and the bar's "related" counter will refetch with a narrower type set than the one it was originally rendered with — showing a lower count than reality. Worth passing `THREAT_ACTOR_GROUP_RELATED_RELATIONSHIP_TYPES` / `THREAT_ACTOR_INDIVIDUAL_RELATED_RELATIONSHIP_TYPES` into the `<StixCoreObjectKnowledgeBar>` instances in those two `Root.tsx` files.

opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectKnowledgeBar.tsx
```diff
@@ -134,19 +147,34 @@ const StixCoreObjectKnowledgeBar = ({
...
   } = useFragment(stixCoreObjectKnowledgeBarFragment, data);
 
+  useEffect(() => {
+    const onKnowledgeBarRefresh = () => {
+      fetchQuery<StixCoreObjectKnowledgeBarRefreshQuery>(
+        environment,
+        stixCoreObjectKnowledgeBarRefreshQuery,
+        { id, relatedRelationshipTypes: relatedRelationshipTypes ?? ['related-to'] },
+        { fetchPolicy: 'network-only' },
+      ).subscribe({});
```
OctaveLaventure: `.subscribe({})` has no `error` handler, so a failed refresh (network blip, GraphQL error) is silently swallowed and the bar just stays stale with no signal. Consider adding at least an `error` callback (e.g. console warning) here.

opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectKnowledgeBar.tsx
OctaveLaventure, line 108: **Bug:** the fallback here (`relatedRelationshipTypes ?? ['related-to']`) doesn't match what two pages actually query.

`threat_actors_group/Root.tsx` and `threat_actors_individual/Root.tsx` fetch this fragment with `relatedRelationshipTypes = ['related-to', 'part-of', 'impersonates', 'known-as']`, but they never pass that same list as a prop to `<StixCoreObjectKnowledgeBar>` (only to the separate `ThreatActor*Knowledge` component). So on those two pages, once someone adds/removes a relationship and the bar refreshes, it will use `['related-to']` only — and the count will drop below what's actually showing on screen.

**Fix:** in both `Root.tsx` files, add `relatedRelationshipTypes={THREAT_ACTOR_GROUP_RELATED_RELATIONSHIP_TYPES}` (or the `..._INDIVIDUAL_...` equivalent) to the `<StixCoreObjectKnowledgeBar>` element.

opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectKnowledgeBar.tsx
OctaveLaventure, line 172: **Nit:** `.subscribe({})` has no `error` handler, so if this refresh call fails (network hiccup, GraphQL error) it fails silently and the bar just stays stale with nothing telling you why. Worth adding e.g. `.subscribe({ error: (err) => console.error('KnowledgeBar refresh failed', err) })`.

OctaveLaventure (approved):
Both previously flagged issues are fixed in 632f6463: `relatedRelationshipTypes` is now passed to `<StixCoreObjectKnowledgeBar>` on the Threat Actor Group/Individual pages, and the refresh subscription now has an error handler. LGTM.

PR #18213, opened 2026-09-10
fix(dashboards): anchor time series buckets and axis labels on UTC (#12150)

### LINKED ISSUE #12150: fix(dashboards): relationships chart on home not displaying July after the month is over
## Description

Relationships chart on home is not displaying July after month is over

Image

Image

Image

I confirm 900K relationships created in both instances:

Image

frapuks:
### Proposed changes

* Anchor `fillTimeSeries` on UTC instead of the platform process time zone. `moment.parseZone()` was called on a JS `Date` — the client offset is already lost at that point by the `DateTime` GraphQL scalar — so it silently fell back to the server zone. Bucket instants therefore carried the server offset while Elasticsearch had aggregated on UTC calendar boundaries all along (`time_zone` is not set on the `date_histogram`). Counting was never affected; only the label attached to the counts was.
* Label time series axes in UTC on the front-end via a new shared `useTimeSeriesAxisFormatter` hook. Fixing only one end would have moved the bug rather than removed it. The hook also replaces an identical formatter-selection block triplicated across `WidgetMultiAreas`, `WidgetMultiLines` and `WidgetVerticalBars`, and covers `WidgetMultiHeatMap` plus the two charts calling `areaChartOptions` directly (`User.tsx`, `StixDomainObjectAuthorKnowledge.jsx`).
* Align `lastDayOfThePreviousMonth()` on UTC. Computed in browser local time it spilled into the next UTC bucket for negative-offset clients, appending a 13th near-empty partial bar. This change and the first one only work combined.
* Fix two adjacent defects of the same family found along the way: `elHistogramCount` used `hh` (12-hour clock in Java `DateTimeFormatter`) for the `hour` interval where `fillTimeSeries` formats with `HH`, so keys never matched in the afternoon and `keyed: true` made 03:00 and 15:00 collide; and weekly buckets were labelled with moment's `startOf('week')`, which follows the locale (Sunday for `en`) while Elasticsearch's `calendar_interval: week` uses the ISO Monday.
* Add unit coverage for `fillTimeSeries`, which had none, and for `lastDayOfThePreviousMonth()`.

### Related issues

* closes #12150

### How to test this PR

The bug only shows when the browser offset is lower than the server offset. The generated axis does not depend on stored data, so no seeding is required.

Try this query on the playground :
```
query LastCompletedMonth($start: DateTime!, $end: DateTime!) {
  stixRelationshipsTimeSeries(
    field: "created_at"
    operation: count
    startDate: $start
    endDate: $end
    interval: "month"
    relationship_type: ["stix-core-relationship"]
  ) {
    date
    value
  }
}
```
Variables :
```
{
  "start": "2025-09-10T00:00:00.000Z",
  "end": "2026-08-31T23:59:59.999Z"
}
```

Results :
- Master : 13 months
- This branch : Only 12 months

### Checklist

### Further comments
Why UTC rather than the reader's zone. The alternative was to propagate the client time zone down to the date_histogram's time_zone option. That was rejected: it is a product decision rather than a bug fix, since it would mean two users on two continents see different numbers for the same dashboard. The consequence to keep in mind is that a bucket labelled "July" spans 1 July 00:00 UTC to 31 July 23:59 UTC — for a reader in Tokyo that is 09:00 to 09:00 in local time. That offset already exists today; this PR only makes the label agree with the bucket's contents.
Behavioural delta worth flagging at review. lastDayOfThePreviousMonth() now returns ...T23:59:59.999Z instead of ...T23:59:59+02:00. Beyond the UTC alignment, this recovers the last 999 ms of the month that .format() was truncating, so the queried range widens very slightly. Counts may differ by a hair against the previous behaviour.


opencti-platform/opencti-front/src/components/dashboard/WidgetMultiAreas.tsx
```diff
@@ -43,7 +35,7 @@ const WidgetMultiAreas = ({
       isStacked,
       hasLegend,
     ) as ApexOptions;
-  }, [theme, interval, isStacked, hasLegend]);
+  }, [theme, interval, formatter, isStacked, hasLegend]);
```
OctaveLaventure: Minor nit: `interval` is redundant in this deps array now — `formatter` already changes whenever `interval` does (it's derived from it via `useTimeSeriesAxisFormatter(interval)`), so `interval` can be dropped here without changing behavior.

opencti-platform/opencti-front/src/components/dashboard/WidgetMultiLines.tsx
```diff
@@ -41,7 +33,7 @@ const WidgetMultiLines = ({
       false,
       hasLegend,
     ) as ApexOptions;
-  }, [theme, interval, hasLegend]);
+  }, [theme, interval, formatter, hasLegend]);
```
OctaveLaventure: Same nit as in `WidgetMultiAreas`/`WidgetVerticalBars`: `interval` is redundant here since `formatter` already tracks it.

opencti-platform/opencti-front/src/components/dashboard/WidgetVerticalBars.tsx
```diff
@@ -50,7 +42,7 @@ const WidgetVerticalBars = ({
       hasLegend,
       'dataPoints',
     ) as ApexOptions;
-  }, [theme, interval, isStacked, hasLegend]);
+  }, [theme, interval, formatter, isStacked, hasLegend]);
```
OctaveLaventure: Same nit as in `WidgetMultiAreas`/`WidgetMultiLines`: `interval` is redundant here since `formatter` already tracks it.

OctaveLaventure (approved):

PR #18221, opened 2026-09-10
fix(report): preserve published date when set to epoch placeholder (#17947)

### LINKED ISSUE #17947: fix(report): Report updates silently dropped when `published` is the epoch placeholder (1970-01-01)
## Description

Reports ingested with `published = 1970-01-01T00:00:00.000Z` (or missing) are accepted at creation, but every subsequent update to them is silently lost.

The chain is entirely inside the platform:

1. The built-in TAXII 2.1 ingester accepts a report whose `published` is the epoch placeholder. In our case the source is a national CTI sharing feed (MISP→STIX converted upstream, delivered over TAXII 2.1); MISP events that were never published produce reports with a missing/epoch `published` date.
2. When any later update targets that report (a playbook adding a label, an enrichment, an edit routed through ingestion), OpenCTI strips the epoch placeholder during STIX serialization, so the outgoing bundle has no `published` field at all.
3. The worker rejects the bundle because `published` is mandatory for Reports:

   [opencti_report] Missing parameters: name and published

4. The rejection is invisible in the UI. The playbook execution shows green/successful, no error is surfaced anywhere in the trace, and the change simply never appears on the report.

This is a silent data-loss class bug: on one production instance we observed 752 occurrences of the worker error over 7 days, all tied to reports the platform itself ingested, with zero user-visible signal. On a lab instance ingesting the same live feed, 314 reports are affected; the failure is deterministic for every one of them.

## Reproducible Steps

1. Import (or ingest via the built-in TAXII 2.1 ingester) a report with `published = 1970-01-01T00:00:00.000Z`, or with `published` absent. The report is created; the UI shows "-" in the Date column.
2. Create a minimal playbook: Listen knowledge (entity type Report) → Add label → Send for ingestion. Manually enroll the report.
3. Inspect the playbook execution trace and the worker logs.

## Expected Output

The label is applied to the report, or, if the update cannot be applied, the failure is surfaced in the playbook execution trace / UI.

## Actual Output

- Step 1 trace: the incoming bundle already has no `published` field (stripped during serialization).
- Step 2: the label patch is applied cleanly to the in-flight bundle.
- Step 3: the outgoing bundle carries the new label but still lacks `published`. The worker drops it with `[opencti_report] Missing parameters: name and published`.
- The playbook execution shows green/complete. The report never receives the label. No error in the UI.

## Additional Information

- [(https://github.com/OpenCTI-Platform/opencti/issues/3238) (closed, fixed in 5.7.3) is where the epoch-placeholder stripping originates; that behavior is presumably intentional, but combined with `published` being mandatory for Reports on update, it makes every epoch-dated report permanently un-updatable through ingestion.
- The asymmetry that makes this hard to spot: observables and relationships with no publication date update fine, because `published` is mandatory only for Reports. Manual UI edits to the affected reports also work; only ingestion-path updates are dropped.
- Suggested directions, in rough order of preference:
  1. Tolerate a missing `published` on update by preserving the existing stored value (upsert semantics).
  2. Surface the worker rejection in the playbook execution trace / UI instead of showing green.
- Also reproduced locally with synthetic bundles; the worker log line timestamp-matches the drop exactly.

ludovic:
### Proposed changes

* Add an `allowEpoch` option to `convertToStixDate` (defaulting to `false`) to allow preserving Unix epoch dates (`1970-01-01T00:00:00.000Z` / `FROM_START`) when required.
* Update STIX 2.0 and STIX 2.1 `convertReportToStix` converters to pass `{ allowEpoch: true }` for the mandatory `published` date.
* Add unit tests for `convertToStixDate` covering edge cases, as well as STIX 2.0 and STIX 2.1 regression tests for report serialization with epoch published dates.

### Related issues

* closes #17947

### How to test this PR

1. **Automated unit tests**:
   ```bash
   yarn test:ci-unit tests/01-unit/utils/stix-converter-test.ts
   yarn test:ci-unit tests/01-unit/database/stix-2-0-converter-test.ts
   yarn test:ci-unit tests/01-unit/database/stix-2-1-converter-report-test.ts
   ```
   
 2. **Functional testing**:
  Create or ingest a Report with its `published` date set to `1970-01-01T00:00:00.000Z` (e.g., from an unpublished MISP event or an RSS feed fallback).
  Update the report or trigger a playbook / draft workspace approval involving it.
  Verify that python workers process the update bundle without throwing `[opencti_report] Missing parameters: name and published.`

### Checklist

### Further comments

`convertToStixDate` strips `FROM_START` (`1970-01-01T00:00:00.000Z`) and `UNTIL_END` (`5138-11-16T09:46:40.000Z`) by default to avoid exporting internal sentinel dates on optional interval fields (`first_seen`, `start_time`, `stop_time`, etc.).

However, for `Report`, `published` is an OASIS STIX-mandatory attribute. Stripping it converted the value to `undefined`, which was then removed by `cleanObject`, generating an invalid STIX report and causing python workers to fail with `Missing parameters: name and published`.

Using an optional `allowEpoch: true` flag preserves valid epoch publication dates on reports while guaranteeing no regression on existing sentinel stripping for other interval fields.

### DIFF
--- a/opencti-platform/opencti-graphql/src/database/stix-2-0-converter.ts
+++ b/opencti-platform/opencti-graphql/src/database/stix-2-0-converter.ts
@@ -436,7 +436,7 @@ export const convertReportToStix = (instance: StoreEntity): SDO.StixReport => {
     name: instance.name,
     description: instance.description,
     report_types: instance.report_types,
-    published: convertToStixDate(instance.published),
+    published: convertToStixDate(instance.published, { allowEpoch: true }),
     object_refs: convertObjectReferences(instance),
     x_opencti_reliability: instance.x_opencti_reliability,
   };
--- a/opencti-platform/opencti-graphql/src/database/stix-2-1-converter.ts
+++ b/opencti-platform/opencti-graphql/src/database/stix-2-1-converter.ts
@@ -569,7 +569,7 @@ const convertReportToStix = (instance: StoreEntity, type: string): SDO.StixRepor
     name: instance.name,
     description: instance.description,
     report_types: instance.report_types,
-    published: convertToStixDate(instance.published),
+    published: convertToStixDate(instance.published, { allowEpoch: true }),
     object_refs: convertObjectReferences(instance),
     extensions: {
       [STIX_EXT_OCTI]: cleanObject({
--- a/opencti-platform/opencti-graphql/src/database/stix-converter-utils.ts
+++ b/opencti-platform/opencti-graphql/src/database/stix-converter-utils.ts
@@ -13,20 +13,30 @@ export const assertType = (type: string, instanceType: string) => {
     throw UnsupportedError('Incompatible type', { instanceType, type });
   }
 };
-export const convertToStixDate = (date: Date | string | undefined): S.StixDate | S2.StixDate => {
+export const convertToStixDate = (
+  date: Date | string | undefined,
+  options: { allowEpoch?: boolean } = {},
+): S.StixDate | S2.StixDate => {
   if (date === undefined) {
     return undefined;
   }
+  const { allowEpoch = false } = options;
   // date type from graphql
   if (date instanceof Date) {
     const time = date.getTime();
-    if (time === FROM_START || time === UNTIL_END) {
+    if (!allowEpoch && time === FROM_START) {
+      return undefined;
+    }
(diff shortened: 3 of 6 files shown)


esrevi (approved):
lgtm !

PR #18290, opened 2026-09-15
fix(frontend): render both sides of a relationship nested on both ends in the knowledge graph (#18288)

### LINKED ISSUE #18288: fix(frontend): investigation graph crashes when a relationship is nested on both ends
The knowledge graph (grouping/report/case/investigation) can crash and render blank, or show entities for a fraction of a second before they disappear. The browser console shows repeated `Uncaught Error: node not found: <id>` thrown from `react-force-graph`'s `link.js`, followed by `Cannot create property 'vx' on string '<id>'`.

## Environment

1. OS: N/A (browser-side)
2. Version: current `main` (opencti-front)
3. Other environment details: affects `GraphContainerKnowledge` / `useGraphParser.ts`, used by Grouping, Report, Case, and Investigation knowledge graphs

## Reproducible steps

1. Create a relationship `R` where **both** `R.from` and `R.to` are themselves relationships (a relationship nested on both ends).
2. Add `R` (and its nested endpoints) to a grouping/report's objects.
3. Open the container's "Knowledge" graph view.

## Expected output

All entities and relationships render normally in the graph and stay rendered.

## Actual output

The graph nodes/links briefly flash then disappear; the console repeatedly throws `node not found: <id>` and the force-graph simulation crashes.

## Additional information

None

gabrielebonadiman:
### Proposed changes

* Fix `buildGraphData` in `useGraphParser.ts` so both sides of a relationship are promoted to graph nodes when that relationship is nested on both ends (i.e. both `from` and `to` are themselves relationships). Previously only the `from` side was recorded, leaving the `to` side excluded from the node list while still being referenced by a link.
* Add a regression test in `useGraphParser.test.ts` that builds a doubly-nested relationship and asserts every link's `source_id`/`target_id` corresponds to an existing node.

### Related issues
* closes #18288

### How to test this PR

1. Create two "leaf" relationships, each connecting two plain entities (e.g. `entity-A` → `entity-B`, `entity-C` → `entity-D`).
2. Create a third relationship whose `from` is the first relationship and whose `to` is the second (relationship nested on both ends). This requires the GraphQL API/Playground, since the standard relationship-creation UI only allows entities as endpoints.
3. Add all entities and relationships to a Grouping (or Report/Case) and open its Knowledge tab.
4. Before the fix: nodes/links render briefly then disappear, with repeated `Uncaught Error: node not found: <id>` in the browser console (from `react-force-graph`'s `link.js`). After the fix: all entities and relationships render and remain stable, with no console errors.
5. Unit test: `yarn vitest run src/components/graph/utils/useGraphParser.test.ts` (12/12 passing).

### Checklist

### Further comments

The bug only manifests when a relationship references two other relationships as both its `from` and `to` (nested on both ends) less common but valid graph shape, which is why it wasn't caught by prior tests.


esrevi (approved):
thx for the clear reproduction steps, tested ok

LGTM

PR #18313, opened 2026-09-16
feat(fintel): simply the export HTML to PDF so that it generates directly the Fintel in PDF (#18307)

### LINKED ISSUE #18307: feat(Fintel): Simply the ExportHTML to PDF so that it generates directly the Fintel in PDF
## Use case

When exporting via HTML to PDF directly in a vulnerability or any entity that does not a “mappable content”, the PDF export is blank. This is due to the fact that content file not present in entities different from container

## Proposed solution

- When using HTML content file to PDF export, have another option "export using fintel template/export dedicated file in PDF"
- If option export using fintel template is selected, then user selects the template of its choice AND the fintel design.
- When the user validates, behind the scene:
    - The Generate Fintel From Template Connector is triggered to generate the HTML file
    - Then the real HTML content file to PDF export is triggered, to transform the generated PDF into a fintel.
   
delemaf:
### Proposed changes

* Add an "Export as fintel" toggle to the HTML-to-PDF export form, enabled by default when the entity has available FINTEL templates.
* When enabled: Template starts empty and is required, "File to export" is disabled and shows "Generated file", Fintel design defaults to the default design if one exists, export filename/markings start empty, and page options (cover/back) start ON.
* On create: generates the HTML from the selected template, saves it as `<name>.html`, then renders and saves `<name>.pdf` from that HTML with the selected design/markings/page options. Both files carry the same file markings and template metadata.
* Form stays busy until both uploads complete. If the HTML save fails, the PDF step never starts. If the PDF render/upload fails, the HTML file is kept, an error is shown, and the form stays open.
* Disabling the toggle (or having no available templates) preserves today's existing-file export behavior unchanged.
* Fixed switching directly between "Export as fintel" and other connectors leaking cleared markings, and prevented dismissing the dialog (Escape/backdrop) while an export upload is still in flight.

### Related issues

* closes #18307

### How to test this PR

- On a Report/Case/Grouping with at least one FINTEL template, open **Export > HTML content files to PDF**.
- Confirm the "Export as fintel" toggle is ON by default, Template is empty/required, File to export shows disabled "Generated file".
- Fill Template + Export file name, click Create: confirm an HTML file then a PDF file are saved (in that order) with matching markings.
- Toggle FINTEL off: confirm the form reverts to today's existing-file export fields/behavior.
- On an entity with no FINTEL templates, confirm the toggle is hidden and legacy behavior is untouched.

### Checklist

**Note:** opened as draft — additional tests around real marking/design-field interactions are still being finalized, and UI translations for the new strings haven't been added yet.


opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectFileExport.tsx
```diff
@@ -282,29 +282,49 @@ const StixCoreObjectFileExportComponent = ({
...
+            }
+            resetForm();
+            close();
+          }
+          resolve();
         },
-        onError: () => {
-          resetForm();
-          close();
+        onError: (error) => {
```
OctaveLaventure, line 316: **Bug:** this `onError` handler isn't the only thing that runs on failure — `useApiMutation` fires a second, unguarded error toast on top of it.

- `src/utils/hooks/useApiMutation.ts` always calls `MESSAGING$.notifyRelayError(error)` right after invoking this custom `onError`, since no `errorMessage`/`errorMessageMap` option is passed here.
- For a plain network error (no `.res.errors`, e.g. connectivity failure), `notifyRelayError` throws `TypeError: Cannot read properties of undefined (reading 'errors')`.
- Reproduced via the new tests: `rejectMostRecentOperation(new Error(...))` triggers 2 uncaught exceptions in vitest (tests still pass since the crash happens post-assertion).

**Impact:** the friendly "HTML file was saved, but the PDF export failed" toast fires, then crashes on a second, unintended toast attempt.

**Suggestion:** pass an `errorMessage` option to `useApiMutation` here (to suppress the generic Relay toast), or guard `notifyRelayError` against errors without `.res`.

opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectFileExport.tsx
```diff
@@ -282,29 +282,49 @@ const StixCoreObjectFileExportComponent = ({
...
+            close();
+            resolve();
+          }
         },
       });
-    };
+    });
 
     try {
+      if (isFintelPdf && !values.template) {
```
OctaveLaventure, line 330: **Nit:** this guard looks unreachable in practice.

- The Yup schema in `StixCoreObjectFileExportForm.tsx` already requires `template` whenever `exportAsFintel` is true.
- The UI can't submit with `isFintelPdf && !values.template`.

Not harmful — consider removing it, or adding a one-line comment noting it's purely defensive (e.g. in case this function is ever called without going through the form's validation).

opencti-platform/opencti-front/src/private/components/common/form/StixCoreObjectFileExportForm.tsx
```diff
@@ -439,7 +469,16 @@ const StixCoreObjectFileExportForm = ({
                   />
                   {values.connector && (
                     <>
-                      {values.connector.value === BUILT_IN_FROM_TEMPLATE.value && (
+                      {values.connector.value === BUILT_IN_HTML_TO_PDF.value && (templates?.length ?? 0) > 0 && (
```
OctaveLaventure, line 475: **UX gotcha found during manual testing:** the new toggle can be unreachable depending on entry point.

1. Open the export dialog via the pre-existing "Generate an export based on a template" shortcut (Content > Files panel).
2. It defaults to `format: text/html` + the `Generate FINTEL from template` connector.
3. Switching the connector combobox to `HTML content files to PDF` is **disabled** (`aria-disabled=true`) in that state, because `isConnectorValid` filters by `connectorScope.includes(values.format)`, and this connector only supports `application/pdf`.

**Result:** a user starting from that shortcut can't reach this toggle unless they first go back to the Format step and pick PDF themselves.

Not a regression from this PR, but since this toggle is meant to be the new default/primary path for fintel entities, it may be worth defaulting format to PDF (or surfacing PDF-compatible connectors more proactively) when opened from that shortcut with templates available.

opencti-platform/opencti-front/src/private/components/common/form/StixCoreObjectFileExportForm.tsx
```diff
@@ -221,6 +225,12 @@ const StixCoreObjectFileExportForm = ({
       onSubmit={onSubmit}
     >
       {({ submitForm, handleReset, isSubmitting, setFieldValue, values }) => {
+        const isFintelPdf = values.connector?.value === BUILT_IN_HTML_TO_PDF.value && values.exportAsFintel;
+
+        useEffect(() => {
```
OctaveLaventure, line 230: **Observation:** this effect resets `exportAsFintel` on every connector change, not just HTML-to-PDF transitions.

- Runs whenever `values.connector?.value` changes, resetting `exportAsFintel` to `true` (if templates exist) — even when switching to an unrelated connector like `fromTemplate`.
- Currently harmless since `isFintelPdf` re-checks the connector value too.
- Still, the field silently flips back to `true` behind the scenes on unrelated connector switches, which could surprise a future reader.

**Suggestion:** scope the effect/condition to only fire when the connector actually becomes `BUILT_IN_HTML_TO_PDF`.

opencti-platform/opencti-front/src/private/components/common/form/StixCoreObjectFileExportForm.tsx
```diff
@@ -242,19 +252,37 @@ const StixCoreObjectFileExportForm = ({
           const connector = values.connector?.value;
```
OctaveLaventure, line 255: **Readability suggestion:** this effect now carries a fair amount of branching for fintel-vs-legacy state resets, alongside:

- the `wasFintelPdf` ref,
- the template-defaults effect,
- the file-based-defaults effect below.

A short comment above this effect summarizing the intent would help future maintainers, e.g.:

> Reset dependent fields when switching connector/fintel-toggle; `wasFintelPdf` remembers the prior fintel state so markings can be restored when toggling off.

opencti-platform/opencti-front/src/private/components/common/stix_core_objects/StixCoreObjectFileExport.tsx
```diff
@@ -282,29 +285,51 @@ const StixCoreObjectFileExportComponent = ({
...
+            resolve();
+          }
         },
       });
-    };
+    });
 
     try {
+      // Guard callers that bypass Formik validation.
+      if (isFintelPdf && !values.template) {
```
OctaveLaventure: **Nit: this guard looks unreachable**

- `template` is already required by the Yup schema whenever `isFintelPdf` is true (see the `template: Yup.object().nullable().when(...)` rule in `StixCoreObjectFileExportForm.tsx`).
- Formik won't call `onSubmit` unless validation passes, so by the time we're here `values.template` should never be null in the fintel path.

**Suggestion:** either remove this guard, or — if it's meant as a defensive check for callers that bypass Formik validation — say so in the comment (similar to what was just added above it) so future readers know it's intentional, not dead code.

opencti-platform/opencti-front/src/private/components/common/form/StixCoreObjectFileExportForm.tsx
```diff
@@ -439,7 +472,16 @@ const StixCoreObjectFileExportForm = ({
                   />
                   {values.connector && (
                     <>
-                      {values.connector.value === BUILT_IN_FROM_TEMPLATE.value && (
+                      {values.connector.value === BUILT_IN_HTML_TO_PDF.value && (templates?.length ?? 0) > 0 && (
```
OctaveLaventure: **UX gotcha: the toggle can be invisible on first open**

**Repro:**
1. Open this dialog via the pre-existing "Generate an export based on a template" shortcut (defaults `format: text/html`).
2. Go to the Connector step.

**Result:** "HTML content files to PDF" is `aria-disabled` (via `isConnectorValid` filtering on `connectorScope`, which only includes `application/pdf`) — so this whole "Export as fintel" block is unreachable until the user manually goes back to the Format step and picks PDF.

**Suggestion:** consider either defaulting `format` to `application/pdf` when this shortcut is used, or surfacing a hint near the disabled option explaining why it's disabled.

opencti-platform/opencti-front/src/private/components/common/form/StixCoreObjectFileExportForm.tsx
```diff
@@ -238,23 +250,42 @@ const StixCoreObjectFileExportForm = ({
           }
         }, [values.format, connectors, values.connector, setFieldValue]);
 
+        // Reset dependent fields on mode changes; wasFintelPdf restores markings when leaving FINTEL mode.
         useEffect(() => {
```
OctaveLaventure: **Readability: dense multi-purpose effect**

Thanks for the summary comment added above this effect — it helps! One more suggestion: this effect still handles several distinct concerns in one block:
- resetting `fileToExport`/`template` on connector change
- entering fintel mode (lines 258-268)
- leaving fintel mode / restoring markings (lines 269-273)
- legacy connector defaults (lines 274-284)

**Suggestion:** a short inline comment above each branch (e.g. `// Entering FINTEL mode`, `// Leaving FINTEL mode`, `// Legacy connector defaults`) would make it much faster to scan without needing to mentally simulate the state machine.

opencti-platform/opencti-front/src/private/components/common/form/StixCoreObjectFileExportForm.tsx
OctaveLaventure, line 475: **Non-blocking UX note**

If a user opens this dialog via the existing "Generate an export based on a template" shortcut, the format defaults to `text/html`. That makes the "HTML content files to PDF" connector show up disabled in the dropdown, which hides this whole "Export as fintel" toggle until they manually go back and pick PDF.

Not a bug, just easy to miss on first use. Feel free to leave for a follow-up if not worth addressing now.

OctaveLaventure (approved):

opencti-platform/opencti-front/src/private/components/common/form/StixCoreObjectFileExportForm.tsx
delemaf: Fixed, the shortcut now opens directly in PDF FINTEL mode, making the toggle immediately visible.

PR #18314, opened 2026-09-16
refactor(form-intakes): improve form intakes architecture (#18274)

### LINKED ISSUE #18274: refactor(form-intakes): improve form intakes architecture
## Description

The Form Intake feature (form definition editing in the frontend, and form ingestion/submission in the backend, both draft and non-draft) had accumulated architectural debt:

1. `formSubmit` mixed pure decision logic (what bundle/draft to build) with I/O side effects (work creation, queue push, patch, telemetry) in one large function.
2. `buildMainStixEntities`'s 4 source-mode branches (lookup, multiple, parsed, default) were inlined as a long if/else chain, with per-branch materialization option asymmetries that were easy to silently break.
3. Draft-related field visibility/required/validation rules were duplicated 3x across `FormView.tsx` (init values, Yup validation, render conditions).
4. `FormSchemaEditor.tsx`'s state mutations went through a single generic `handleFieldChange(path, value)` string-path walker, even for semantically stable operations (rename field, toggle parsed mode, add entity, edit relationship), making the code hard to follow and test in isolation.
5. `FormSchemaEditor.tsx` itself had grown into a very large single file mixing state, policy and JSX composition for several structurally distinct sections.

## Goal

Deepen the module boundaries without changing any observable behavior:

- Backend: split `formSubmit` into a pure `planSubmission` (decision) + `commitSubmission` (I/O) composition, and extract draft choreography into a `buildDraftPlan` helper.
- Backend: extract `buildMainStixEntities`'s branches into an `EntitySourceAdapter` map, preserving all per-branch option asymmetries exactly.
- Frontend: centralize draft field policy computation into a single `computeDraftPolicy` function/hook, replacing the 3x-duplicated logic in `FormView.tsx`.
- Frontend: introduce named `FormBuilderData` operations (`renameField`, `toggleParsedMode`, `addAdditionalEntity`, relationship operations, ...) via a `useFormBuilderOperations` hook, replacing the primary uses of the generic `handleFieldChange` path-walker.
- Frontend: continue extracting `FormSchemaEditor.tsx` into focused section components.

## Acceptance criteria

- No observable behavior change (verified by full backend + frontend test suites and manual equivalence review).
- Existing test suites pass; new unit tests cover the extracted units.
- `tsc`/typecheck clean on both packages.

OctaveLaventure:
## Description

Deepens the Form Intake feature's architecture (both frontend form editing and backend form ingestion, draft and non-draft modes) per the 5 candidates identified in an `/improve-codebase-architecture` review. Pure refactor, no observable behavior change.

Closes #18274

### Backend

- **`buildDraftPlan`**: centralizes the draft-mode choreography previously inlined in `formSubmit`'s draft branch (`form-domain.ts`).
- **`planSubmission` / `commitSubmission`**: splits `formSubmit` into a pure decision step (`planSubmission`: builds the bundle/draft plan, no I/O) and an execution step (`commitSubmission`: work creation, queue push, patch, telemetry). `formSubmit` is now a thin 2-call composition.
- **`EntitySourceAdapter` map**: extracts `buildMainStixEntities`'s 4 source-mode branches (lookup, multiple, parsed, default) into a `Record<Mode, EntitySourceAdapter>` map + a mode resolver, preserving every per-branch `buildMaterializeOptions` asymmetry exactly.

### Frontend

- **`computeDraftPolicy`**: centralizes draft field visibility/required/validation rules that were previously duplicated 3x in `FormView.tsx` (init values, Yup validation, render conditions) into a single computed `DraftPolicy`.
- **`useFormBuilderOperations`**: introduces named `FormBuilderData` operations (`renameField`, `toggleParsedMode`, `addAdditionalEntity`, `updateRelationshipEntity`, `updateRelationshipType`, `toggleRelationshipRequired`) replacing the primary uses of the generic `handleFieldChange(path, value)` string-path walker. Genuinely dynamic/generic settings paths remain on `handleFieldChange` by design.

## How to test this PR

Backend:
```bash
yarn test:ci-unit
yarn check-ts
```

Frontend:
```bash
cd opencti-platform/opencti-front
yarn vitest run src/private/components/data/forms
yarn check-ts
```

Manually, verify form creation/edition and form submission (draft and non-draft) still behave identically:
- Create/edit a form schema in the Form Intake builder (rename a field, toggle parsed mode, add an additional entity, edit a relationship).
- Submit a public form both as a draft and as a direct (non-draft) submission, for each main-entity source mode (lookup, multiple observables, parsed/pattern, default single entity).


opencti-platform/opencti-front/src/private/components/data/forms/AdditionalEntitiesSection.tsx
```diff
@@ -0,0 +1,374 @@
...
+                ? t_i18n('Required (auto-set due to default values)')
+                : t_i18n('Required')}
+            />
+          );
+        })()}
+
+        {entity.multiple && !entity.lookup && (
+          <Select
+            value={entity.fieldMode}
+            onValueChange={() => toggleParsedMode(entity.id)}
```
delemaf: #### Preserve the selected target mode instead of toggling

The issue requires no observable behavior change, but this callback now discards the selected value. `AdditionalEntity.fieldMode` is optional: when an existing/imported entity omits it, selecting 'Multiple fields' invokes `toggleParsedMode`, which changes undefined to 'parsed'. The previous callback assigned the selected 'multiple' value. This unexpectedly displays parsing controls and can require a mapping before saving. I reproduced the resulting state with the actual hook in memory. Could we pass the selected target mode into the named operation and add a regression test for an omitted fieldMode? The binary-mode comment does not account for this valid unset state.

opencti-platform/opencti-graphql/src/modules/form/form-domain.ts
```diff
@@ -450,17 +514,43 @@ export const formSubmit = async (
   };
```
delemaf, line 523: #### The extracted planner still performs I/O

Issue #18274 explicitly asks for a pure planSubmission decision step and a commitSubmission I/O step. These builders still reach external dependencies: lookup materialization calls loadFormEntity/storeLoadById, and pattern construction can call the Python bridge. The planner also generates UUIDs and writes logs. This is a partial architecture requirement rather than a newly introduced runtime regression: planning still needs services and cannot deterministically replay identical inputs. Could we resolve external inputs before entering the pure planner, inject generated IDs, and keep logging and service calls in orchestration?

opencti-platform/opencti-front/src/private/components/data/forms/RelationshipsSection.tsx
```diff
@@ -0,0 +1,392 @@
...
+  updateRelationshipEntity,
+  updateRelationshipType,
+  toggleRelationshipRequired,
+  handleRemoveRelationship,
+  handleAddRelationship,
+}) => {
+  const classes = useStyles();
+  const { t_i18n } = useFormatter();
+  const { schema } = useAuth();
+  const updateEntity = updateRelationshipEntity || ((relationshipId, side, entityId) => {
```
delemaf: #### Remove duplicate relationship-operation fallbacks

Possible Duplicated Code / Speculative Generality: the fallback here and the adjacent type/required fallbacks maintain a second implementation of the operations hook. The production caller supplies all three callbacks, while the section tests exercise the fallback path. Could these callbacks be required, with tests supplying them and covering the production wiring? That would remove redundant mutation logic and avoid tests validating a different path from the one users run.

opencti-platform/opencti-front/src/private/components/data/forms/useFieldRenderer.tsx
```diff
@@ -0,0 +1,511 @@
...
+      .map((f) => f.attributeMapping.attributeName);
+    allAttributes = allAttributes.filter((attr) => !existingFields.includes(attr.value));
+
+    // Determine available field types based on selected attribute
+    let availableFieldTypes: typeof FIELD_TYPES = [];
+    if (field.attributeMapping.attributeName) {
+      const selectedAttribute = allAttributes.find((attr) => attr.value === field.attributeMapping.attributeName);
+
+      // Check if it's a special attribute first
+      if (field.attributeMapping.attributeName === 'createdBy') {
```
delemaf: #### Share special-attribute compatibility rules

Possible Repeated Switches: the special-attribute branches here recur in the selected-type handling around line 210, including createdBy and sibling attributes. The two lists already handle x_opencti_main_observable_type differently. This duplication predates the extraction, so it is not a new regression, but a shared attribute-to-type mapping and compatibility helper would make the extracted renderer easier to read and maintain without the lists drifting apart.

opencti-platform/opencti-front/src/private/components/data/forms/MainEntitySection.tsx
```diff
@@ -0,0 +1,323 @@
...
+                <SelectItem value="line">{t_i18n('One per line')}</SelectItem>
+              )}
+            </SelectContent>
+          </Select>
+          <Select
+            value={formData.mainEntityParseFieldMapping || ''}
+            onValueChange={(value) => {
+              const newMapping = value;
+              updateFormData((prev) => {
+                const wasFirstSelection = !prev.mainEntityParseFieldMapping;
```
delemaf: #### Share the parsed-mapping field-removal operation

Possible Duplicated Code: the wasFirstSelection logic and subsequent field-removal algorithm are repeated in AdditionalEntitiesSection.tsx around line 246. This is inherited logic, not an introduced behavior regression. As a follow-up cleanup, could a shared parsed-mapping operation own the first-selection versus replacement behavior, parameterized by the entity ID? That would keep fixes consistent between the main and additional entity sections.

opencti-platform/opencti-front/src/private/components/data/forms/view/fieldRenderers/types.ts
```diff
@@ -0,0 +1,47 @@
+import type React from 'react';
+import type { FieldOption } from '../../../../../../utils/field';
+import type { FormFieldDefinition } from '../../Form.d';
+
+export interface FieldRendererContext {
+  field: FormFieldDefinition;
+  values: Record<string, unknown>;
+  errors: Record<string, string>;
+  touched: Record<string, boolean>;
+  setFieldValue: (
```
delemaf, line 13: #### Reuse the renderer input contract

Possible Data Clumps / Duplicated Code: the long setFieldValue value union and entitySettings structure repeat the inputs declared in FormFieldRenderer.tsx. Could both interfaces compose a shared renderer-input type instead? Adding a supported field value or changing entity settings would then require one contract update rather than maintaining two matching declarations.

opencti-platform/opencti-graphql/tests/01-unit/domain/form-plan-submission-test.ts
```diff
@@ -0,0 +1,68 @@
...
+  buildAdditionalEntities: vi.fn().mockResolvedValue(new Map()),
+  buildRelationships: vi.fn().mockResolvedValue(undefined),
+  wrapInContainerOrPush: vi.fn(),
+}));
+
+const mockContext: any = {};
+const mockUser: any = { id: 'user-1', capabilities: [] };
+
+describe('planSubmission', () => {
+  it('builds a bundle with objects and no draft plan when isDraft is false', async () => {
```
delemaf: #### Make the bundle-construction test match its claim

This test says it builds a bundle 'with objects', but all bundle-populating builders are mocked, wrapInContainerOrPush is a no-op, and bundle.objects is never asserted. It would still pass if the planner stopped wiring the builders together correctly. Could we either assert the builder arguments and rename this as an orchestration test, or exercise real builders and assert the resulting objects? This would make coverage of the extracted unit meaningful and the test name accurate.

opencti-platform/opencti-graphql/tests/01-unit/domain/form-draft-plan-test.ts
```diff
@@ -0,0 +1,44 @@
+import { describe, expect, it } from 'vitest';
+import { buildDraftPlan } from '../../../src/modules/form/form-domain';
+import { BYPASS } from '../../../src/utils/access';
+
+const mockUser: any = { id: 'user-1', capabilities: [{ name: BYPASS }] };
+const nonBypassUser: any = { id: 'user-2', capabilities: [] };
+
+describe('buildDraftPlan', () => {
```
delemaf: This unit test only covers the `author` field. It never exercises `canOverrideAuthorizedMembers`/`resolveAuthorizedMembersForDraft` (the privilege-gating logic for non-bypass users), nor the `main_entity_author`/`none` author types. `form-submit-test.ts` still covers this end-to-end today, so nothing is broken, but `buildDraftPlan` in isolation has no regression protection for its security-sensitive branches — a future edit to this function alone wouldn't fail any test scoped to it. Worth adding a non-bypass + `authorizedMembers.isEditable: false` denial case, plus `main_entity_author`/`none` cases.

opencti-platform/opencti-front/src/private/components/data/forms/view/FormViewUtils.ts
```diff
@@ -1,6 +1,80 @@
...
+        author?.isEditable
+        && author?.isRequired
+        && author.type === 'none'
+      ),
+      clearable: author?.type === 'main_entity_author',
+    },
+    authorizedMembers: {
+      visible: !!(authorizedMembers?.enabled && (isBypass || authorizedMembers.isEditable)),
+      initialized: !!authorizedMembers?.enabled,
+      required: !isBypass && !!(authorizedMembers?.enabled && authorizedMembers?.isRequired),
```
delemaf: `required` here is gated with `!isBypass &&`, unlike every sibling field's `required` (unconditional), and it's never consumed — `AuthorizedMembersField` only reads `.disabled`. Harmless (pre-existing pattern), but it breaks the `DraftFieldPolicy` interface's consistency and could mislead a future reader into thinking it's wired up. Either remove it or use it.

opencti-platform/opencti-graphql/src/modules/form/form-domain.ts
```diff
@@ -410,18 +410,91 @@ export const resolveAuthorizedMembersForDraft = (
...
+  if (createdBy) draftInput.createdBy = createdBy;
+  if (authorized_members.length > 0) draftInput.authorized_members = authorized_members;
+  // Form intake configuration must override customization mandatory attributes.
+  draftInput.bypassMandatoryAttributes = true;
+
+  return { draftInput };
+};
+
+export interface SubmissionPlan {
+  bundle: any;
```
delemaf: `bundle: any` (and the adapter results typed `any[]` in form-bundle-builder.ts) carries forward pre-existing `any` typing. Not a regression, but since `SubmissionPlan` is a new interface introduced in this PR, it was a low-cost opportunity to type this against the existing STIX bundle/container type instead of `any`.

opencti-platform/opencti-front/src/private/components/data/forms/useFormBuilderOperations.ts
```diff
@@ -0,0 +1,118 @@
...
+    entityId: string,
+  ) => {
+    updateFormData((prev) => ({
+      ...prev,
+      relationships: prev.relationships.map((relationship) => (
+        relationship.id === relationshipId
+          ? {
+              ...relationship,
+              [side]: entityId,
+              relationshipType: relationship.relationshipType ? '' : relationship.relationshipType,
```
delemaf: `relationship.relationshipType ? '' : relationship.relationshipType` is correct (behavior-preserving "clear only if set" logic, verified against the original two-call pattern) but reads as a no-op flip at a glance. A short comment (e.g. `// clear the type when switching entities`) would save the next reader a double-take.

opencti-platform/opencti-graphql/src/modules/form/form-domain.ts
```diff
@@ -450,17 +522,43 @@ export const formSubmit = async (
...
-  logApp.info('[FORM] STIX Bundle generated', { bundleId: bundle.id, objectCount: bundle.objects.length, bundle });
 
+  let draftPlan: DraftPlan | null = null;
+  if (finalIsDraft) {
+    try {
+      draftPlan = buildDraftPlan(form.name, schema, values, user, isBypass);
+    } catch (error) {
+      // Preserve the same error surface as commitSubmission's queue-push failures below,
+      // since draft-plan building was part of that single try/catch before the split.
+      logApp.error('[FORM] Error sending bundle to connector queue', { error });
```
delemaf: This catch logs `'[FORM] Error sending bundle to connector queue'` for a `buildDraftPlan` failure, before any queue interaction has happened. Pre-existing mismatch (the original code shared one try/catch for draft-plan building and the queue push), not a regression — but since this call site was already being restructured for the plan/commit split, it was a nearly-free opportunity to give this branch its own accurate message (e.g. `'[FORM] Error building draft plan'`).

opencti-platform/opencti-graphql/tests/01-unit/domain/form-draft-plan-test.ts
delemaf, line 8: This test only covers the `author` field. It never tests `canOverrideAuthorizedMembers`/`resolveAuthorizedMembersForDraft` (the permission checks for non-bypass users), or the `main_entity_author`/`none` author types. `form-submit-test.ts` still covers these cases end-to-end, so nothing is broken today. But `buildDraftPlan` has no dedicated test for its permission logic, so a future change to it alone could break silently. Suggest adding a case for a non-bypass user with `authorizedMembers.isEditable: false`, plus the two missing author types.

opencti-platform/opencti-front/src/private/components/data/forms/view/FormViewUtils.ts
delemaf, line 72: `required` is set with `!isBypass &&`, but every other field here sets `required` unconditionally. It's also never used — `AuthorizedMembersField` only reads `.disabled`. Harmless dead code (pre-existing pattern), but it makes this interface look wired up when it isn't. Remove it or use it.

opencti-platform/opencti-graphql/src/modules/form/form-domain.ts
delemaf, line 477: `bundle: any` (and the adapter results typed `any[]` in form-bundle-builder.ts) keeps the old loose typing. Not a regression, but `SubmissionPlan` is a new interface added in this PR, so this was a cheap chance to use the existing STIX bundle type instead of `any`.

opencti-platform/opencti-front/src/private/components/data/forms/useFormBuilderOperations.ts
delemaf, line 75: `relationship.relationshipType ? '' : relationship.relationshipType` is correct — it clears the type only when one is set, matching the original logic. But it reads like a no-op at first glance. A short comment like `// clear type when switching entities` would help.

opencti-platform/opencti-graphql/src/modules/form/form-domain.ts
delemaf, line 537: This catch logs `'[FORM] Error sending bundle to connector queue'` for a `buildDraftPlan` failure, but no queue call has happened yet. Pre-existing issue, not a regression — the original code shared one try/catch for both steps. Since this part was already rewritten, a more accurate message like `'[FORM] Error building draft plan'` would help.

delemaf (approved):
Minor change, non blocking 🪨

PR #18330, opened 2026-09-17
docs(exclusion-lists): clarify CIDR usage and indicator-only scope (#18333)

### LINKED ISSUE #18333: docs: exclusion lists don't clarify CIDR usage or that Observables are unaffected
### Documentation request

The exclusion lists page does not explain how to format IPv4/IPv6 values. It's not clear that a single entry like `192.168.0.0/16` is a valid CIDR range that excludes every address within it (e.g. `192.168.20.10`), as opposed to only supporting single addresses.

The page also doesn't state that exclusion lists are only evaluated when an Indicator is created from a STIX pattern. It does not filter standalone Observables created directly (for example by a connector or feed ingestor that produces an Observable without a matching Indicator). This scope limitation caused confusion in the community, since users assumed exclusion lists would prevent any matching value from being imported at all.

### Additional context

_No response_

gabrielebonadiman:
### Proposed changes

* Document that IPv4-Addr/IPv6-Addr exclusion list entries accept CIDR ranges (e.g. `192.168.0.0/16`) in addition to single addresses.
* Add a note clarifying that exclusion lists are evaluated only against Indicator STIX patterns, not standalone Observables created directly by connectors or feed ingestors.

### Related issues
* closes  #18333 

### How to test this PR

Review the rendered page for `docs/docs/administration/exclusion-lists.md` (`mkdocs serve` from `docs/`) and confirm the new IP formatting examples and the indicator-only scope note read clearly in context.

### Checklist

### Further comments

This is a documentation-only change addressing the CIDR/scope confusion; it does not close the broader wildcard/regex/normalization request in that issue, which remains open for Domain-Name, URL, and Hostname matching.


aHenryJard (approved):
thanks

PR #18341, opened 2026-09-17
fix(export): wide tables embedded in report content are cut off in PDF exports (#15021)

### LINKED ISSUE #15021: fix(export): wide tables embedded in report content are cut off in PDF exports
## Description

Tables generated in a PDF of a report are cut off losing critical data relevant to the report.
This happens on nearly all PDFs with a wide table, See some images attached to generate a PDF, these are cut off so half the table is missing, if a table is small it does not happen. 

## Environment
1. SaaS 7.260309.0

## Reproducible Steps
See in CSM platform
https://csm.octi.staging.filigran.io/dashboard/analyses/reports/9de45d1e-f1dc-4446-a908-c7e61281ada5/content?currentFileId=null&contentSelected=true&pdfViewerZoom=1.2&markdownSelectedTab=write&readMode=false&changed=false&isLoading=false

1. Have a report with a long **Description** (mardkown) ex: https://testing.octi.staging.filigran.io/dashboard/analyses/reports/4663b98f-8362-4220-afe0-9ab93142ea17/content?currentFileId=null&contentSelected=true&pdfViewerZoom=1.2&markdownSelectedTab=write&readMode=false&changed=false&isLoading=false&onProgressExportFileName=undefined 
2. Export using ExportReportPDF

## Expected output
The description is properly parsed, including the columns

## Current output
The columns are cropped

## Screenshots (optional)

Image

Image

delemaf:
### Proposed changes

Prevent tables containing long hashes or URLs from overflowing the printable page in built-in HTML/Markdown and Fintel PDF exports.

* Replace pdfmake flexible (`*`) column widths with equal percentage widths. Flexible columns respect long tokens' minimum widths and can exceed the page; constrained widths allow those tokens to wrap.
* Apply table preparation after Markdown conversion in the built-in export path, as well as in Fintel exports.
* Count logical columns using `colSpan` so merged headers do not cause rendering failures.
* Constrain table images to their cell content width while preserving aspect ratio, accounting for padding and each export's page margins.
* Preserve source content, link targets, font size, and page orientation.
* Add real rendered-PDF regression tests and declare the already-used PDF.js parser as a direct development dependency for PDF text, link, and image-bounds assertions.

### Related issues

* closes #15021

### How to test this PR

1. Add a four-column table to report content with ordinary prose, a 64-character SHA-256 hash, a long linked URL, and a final status column.
2. Export using the built-in PDF conversion. Repeat with a Markdown file containing the same table and with a Fintel template containing the table.
3. Verify that every column is visible inside the page margins, the full hash and URL are present, ordinary words wrap naturally, and the URL remains clickable with its original target.
4. Add a merged header spanning all four columns and repeat the export. Confirm that PDF generation succeeds.
5. Add a wide image to the last column. Confirm that it stays inside the cell without aspect-ratio distortion.
6. Confirm that page orientation and font size remain unchanged.

### Checklist

Rendered-PDF regression tests are included; no new browser E2E tests or documentation changes are included.


opencti-platform/opencti-front/src/utils/htmlToPdf/utils/pdfTableWidth.ts
```diff
@@ -1,22 +1,38 @@
...
+        }));
+        const parentCell = table.parentElement?.closest('td, th');
+        const tableWidth = (parentCell && cellWidths.get(parentCell)) || contentWidth;
+        Array.from(table.rows).forEach((row) => {
+          Array.from(row.cells).forEach((cell) => {
+            const cellWidth = Math.max(1, tableWidth * cell.colSpan / columnCount - 2 * TABLE_PADDING - 2 * TABLE_BORDER);
+            cellWidths.set(cell, cellWidth);
+            cell.querySelectorAll('img').forEach((image) => {
+              if (image.closest('td, th') === cell) image.style.maxWidth = `${cellWidth}pt`;
+            });
```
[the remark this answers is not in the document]
delemaf: Checked against the installed pdfmake 0.3.11 renderer: DocMeasure.measureImageWithDimensions recomputes height from the intrinsic image ratio whenever maxWidth clamps the width, even when height was explicitly supplied. This differs from browser CSS behavior. Added local rendered-PDF regressions for height attributes and inline heights across the helper, HTML, Markdown and Fintel paths; image bounds and intrinsic aspect ratios pass without removing height.

opencti-platform/opencti-front/src/utils/htmlToPdf/utils/pdfTableWidth.ts
```diff
@@ -1,22 +1,38 @@
...
-    const header = table.querySelector('thead tr');
-    const body = table.querySelector('tbody tr');
-    const element = header ?? body;
-    if (element) {
-      const nbColumns = element.querySelectorAll(header ? 'th' : 'td').length;
-      if (nbColumns) {
-        table.setAttribute('data-pdfmake', `{'layout':'default', 'widths':[${Array(nbColumns).fill("'*'").join()}]}`);
+    const firstRow = table.rows[0];
+    if (firstRow) {
+      const columnCount = Array.from(firstRow.cells).reduce((total, cell) => total + cell.colSpan, 0);
```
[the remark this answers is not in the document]
delemaf: Confirmed: a grouped two-column header followed by wider body rows was incorrectly counted as two columns. Fixed locally by taking the maximum sum of colSpan values across table.rows. Added regressions for 4-, 8- and 12-column bodies to cover width allocation and adaptive layout detection; these failed before the change and pass now.

opencti-platform/opencti-front/src/utils/htmlToPdf/utils/pdfTableWidth.ts
```diff
@@ -1,14 +1,105 @@
...
+const TABLE_BORDER = 1;
+
+const getTableCells = (table: HTMLTableElement) => {
+  const occupiedUntil: number[] = [];
+  return Array.from(table.rows).flatMap((row, rowIndex) => {
+    let column = 0;
+    return Array.from(row.cells).map((cell) => {
+      while (occupiedUntil[column] > rowIndex) column += 1;
+      const position = { cell, column };
+      const remainingRows = (row.parentElement?.children.length ?? 1) - row.sectionRowIndex;
```
OctaveLaventure, line 33: Possible bug: rowspan across thead/tbody boundary.

`remainingRows` is calculated only within the current section (thead or tbody), but it's then compared against `rowIndex`, which counts rows across the *whole* table. So if a cell has `rowspan` that would logically reach into the next section, this code won't account for it correctly, and columns after it could shift/misalign.

Suggestion: add a test with a `rowspan` cell near the end of `<thead>` or `<tbody>` to confirm this works as expected, and adjust the calculation if it doesn't.

opencti-platform/opencti-front/src/utils/htmlToPdf/utils/pdfTableWidth.ts
```diff
@@ -1,14 +1,105 @@
...
+    });
+  });
+};
+
+const getTableColumnCount = (table: HTMLTableElement) => getTableCells(table)
+  .reduce((maxColumns, { cell, column }) => Math.max(maxColumns, column + cell.colSpan), 0);
+
+const getColumnMinimumWidths = (table: HTMLTableElement): number[] => {
+  const columnCount = getTableColumnCount(table);
+  const padding = columnCount >= VERY_WIDE_TABLE_COLUMN_THRESHOLD ? 1 : columnCount >= WIDE_TABLE_COLUMN_THRESHOLD ? 2 : TABLE_PADDING;
```
OctaveLaventure: This block re-implements the same "how wide/small should this table's font and padding be" logic that already exists a bit further down in `setTableFullWidth` (the `isVeryWideTable`/`isWideTable` if/else). It's written differently here but does the same job.

Suggestion: if possible, pull this into one small shared function (e.g. `getTableRenderConfig(columnCount)` returning `{ padding, fontSize, layout }`) and call it from both places. That way, if these thresholds/values ever need to change, there's only one place to update instead of two.

opencti-platform/opencti-front/src/utils/htmlToPdf/utils/pdfTableWidth.ts
```diff
@@ -1,14 +1,105 @@
...
+      const nestedWidth = getColumnMinimumWidths(nested).reduce((total, width) => total + width, 0);
+      for (let offset = 0; offset < cell.colSpan; offset += 1) {
+        minimums[column + offset] = Math.max(minimums[column + offset], (nestedWidth + spacing) / cell.colSpan);
+      }
+    });
+  });
+  return minimums;
+};
+
+const allocateColumnWidths = (
```
OctaveLaventure, line 65: `allocateColumnWidths` takes 7 arguments. `padding`, `fontSize` and `fontFamily` always seem to travel together wherever this function is used.

Suggestion: group these three into a single object (e.g. `renderConfig: { padding, fontSize, fontFamily }`) to make the function signature easier to read and call. Not blocking, just a readability improvement.

opencti-platform/opencti-front/src/utils/htmlToPdf/utils/pdfTableWidth.ts
```diff
@@ -1,14 +1,105 @@
...
+) => {
+  const equalWidth = tableWidth / columnCount;
+  if (!context) return Array<number>(columnCount).fill(equalWidth);
+
+  const spacing = 2 * padding + 2 * TABLE_BORDER;
+  const minimum = Math.min(equalWidth / 2, spacing + 4 * fontSize);
+  const maximum = tableWidth * Math.max(1 / columnCount, Math.min(0.6, 2 / columnCount));
+  const desired = Array<number>(columnCount).fill(minimum);
+  getTableCells(table).forEach(({ cell, column }) => {
+    context.font = `${cell.tagName === 'TH' ? 'bold ' : ''}${fontSize}pt ${fontFamily}`;
```
OctaveLaventure, line 81: This might undermine the fix for CJK (Japanese/Korean) tables specifically.

Here we measure text width in a `<canvas>` using `fontFamily` (which can be `'NotoSansJp'` or `'NotoSansKr'`, see `detectLanguage()`). The problem: those font names are only registered inside pdfMake's own font setup (`pdfFonts.ts`) for generating the actual PDF — they're never loaded as a real browser font (no `@font-face`/`FontFace` for them anywhere in the codebase).

So in a real browser, `context.font = '...NotoSansJp'` will silently fall back to some default font the browser already has, and `measureText` will return a width based on *that* font — not the actual (usually much wider) CJK glyphs pdfMake will render. That means column widths calculated here could end up too narrow for CJK text, which risks bringing back the exact "columns get cropped" bug from #15021, just for CJK reports.

Worth double-checking with an actual Japanese/Korean wide table export. The current tests don't catch this because the mocked canvas always reports Roboto-style widths, regardless of `fontFamily`.

Possible fix: load the Noto fonts as real browser fonts before measuring (e.g. via the CSS Font Loading API), or pad the estimated width more generously when the font isn't Roboto.

opencti-platform/opencti-front/src/utils/htmlToPdf/utils/pdfTableWidth.ts
```diff
@@ -1,14 +1,105 @@
...
+    const textWidth = context.measureText(text).width * 0.75;
+    const demand = (textWidth + spacing) / cell.colSpan;
+    for (let offset = 0; offset < cell.colSpan; offset += 1) {
+      desired[column + offset] = Math.max(desired[column + offset], demand);
+    }
+  });
+
+  const widths = getColumnMinimumWidths(table).map((width) => Math.max(minimum, width));
+  let remaining = tableWidth - widths.reduce((total, width) => total + width, 0);
+  if (remaining < 0) return Array<number>(columnCount).fill(equalWidth);
```
OctaveLaventure: Edge case: if there truly isn't enough room for every column's minimum width (`remaining < 0`), this throws away all the minimum-width work above and just splits the space evenly instead. That fallback doesn't respect the minimums computed earlier, so it's possible for some columns to end up even narrower than what was calculated as their minimum — which is the exact kind of cropping/overflow this PR is trying to fix, just in a very dense-table edge case.

Suggestion: instead of jumping straight to equal widths, shrink each column proportionally starting from its minimum (i.e., still respect relative sizing, just scale everything down a bit).

opencti-platform/opencti-front/src/utils/htmlToPdf/htmlToPdf.ts
```diff
@@ -79,6 +79,7 @@ export const htmlToPdf = (
   if (fileName && fileName.endsWith('.md')) {
     htmlData = renderToString(compiler(htmlData, { wrapper: null }));
   }
+  htmlData = setTableFullWidth(htmlData);
```
OctaveLaventure, line 82: Quick question on scope: this is the plain `htmlToPdf` export (used for Markdown/HTML), and it calls `setTableFullWidth(htmlData)` with the default fixed width (515.28) — it never adjusts page size or orientation for wide tables.

Compare with `htmlToPdfReport` further down, which switches to landscape/A3 and uses tighter margins when a table is wide.

So a wide table exported through this path won't overflow anymore, but it'll be squeezed into a portrait A4 page at 8pt font / minimal padding, which is much less readable than the same table exported via the Report path.

Is that intentional (e.g. this path isn't used for reports with big tables), or should it get the same page-size/orientation treatment as `htmlToPdfReport`?

delemaf: we’re preserving the behavior introduced in #18143.

Plain HTML/Markdown exports remain A4 portrait; 515.28pt is the printable width after the default margins. Fintel report exports retain automatic A4 landscape at 8–11 columns and A3 landscape at 12+, including their tighter margins.

This PR fixes overflow and improves column allocation without extending that page-orientation policy to plain exports. The readability tradeoff is real, but applying automatic landscape/A3 there would be a separate behavior change. Regression tests cover both paths and the threshold boundaries.

OctaveLaventure (approved):
Re-reviewed after latest commit (c190ab6). All prior comments fixed or clarified. 99/99 tests pass.

